PR #44, opened 2015-07-20
CMake: Add Boost >= 1.56.0 requirement

ax3l:
The header `<boost/core/ignore_unused.hpp>` requires at least boost 1.56.0 to be compiled
successfully.

The `find_package(Boost 1.56.0 QUIET [...])` calls are now using QUIET since `alpakaConfig.cmake` itself creates a "Boost not found" warning and this "option disables [additional] messages if the package
cannot be found."
  http://www.cmake.org/cmake/help/v3.0/command/find_package.html

P.S.: not sure which branch you prefer for pull requests, so I choose `develop`.

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -93,20 +93,21 @@ SET_PROPERTY(CACHE ALPAKA_DEBUG PROPERTY STRINGS "0;1;2")
 #-------------------------------------------------------------------------------
 # Find Boost.
 #-------------------------------------------------------------------------------
+SET(_ALPAKA_BOOST_MIN_VER "1.56.0") # minimum version for basic features
 IF(${ALPAKA_DEBUG} GREATER 1)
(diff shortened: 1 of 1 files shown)


BenjaminW3:
Thanks for this enhancement.
If you could rename ```Boost_MINV``` to ```_ALPAKA_BOOST_MIN_VER``` I will merge it.
Because CMake does not provide a namespace concept and those variables will pass through to calling/including files, alpaka specific CMake variables are prefixed with ```ALPAKA``` to prevent name clashes and confusion.
Furthermore I distinguish variables for public consumption from those for internal usage by a underscore prefix.

P.S.: Yes, ```develop``` is correct.

ax3l:
sure, thanks! shall I generally `unset` the var again? I am not sure if `mark_as_advanced` would be useful for `_ALPAKA_BOOST_MIN_VER`

PR #68, opened 2015-10-15
fix race condition in `EventCpu.hpp`

psychocoderHPC:
- take care whether a event we wait on is already finished
- fix `WaiterWaitFor` for async and sync cpu streams

This bug is triggered if event finished while the main thread enter `WaiterWaitFor<>::waiterWaitFor`

### DIFF
--- a/include/alpaka/event/EventCpu.hpp
+++ b/include/alpaka/event/EventCpu.hpp
@@ -481,17 +481,17 @@ namespace alpaka
                     // This is forwarded to the lambda that is enqueued into the stream to ensure that the event implementation is alive as long as it is enqueued.
                     auto spEventCpuImpl(event.m_spEventCpuImpl);
 
+                    std::lock_guard<std::mutex> lk(spEventCpuImpl->m_Mutex);
+                    if(!spEventCpuImpl->m_bIsReady)
                     {
-                        std::lock_guard<std::mutex> lk(spEventCpuImpl->m_Mutex);
                         spEventCpuImpl->m_bIsWaitedFor = true;
+                        // Enqueue a task that waits for the given event.
+                        spStreamImpl->m_workerThread.enqueueTask(
+                            [spEventCpuImpl]()
+                            {
+                                wait::wait(spEventCpuImpl);
+                            });
                     }
-
-                    // Enqueue a task that waits for the given event.
-                    spStreamImpl->m_workerThread.enqueueTask(
-                        [spEventCpuImpl]()
-                        {
-                            wait::wait(spEventCpuImpl);
-                        });
                 }
             };
             //#############################################################################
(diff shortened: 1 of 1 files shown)


BenjaminW3:
How did you find the solution to this hard to track down bug?

erikzenker:
Pair programming :-)

BenjaminW3:
Congratulations!

The CI tests will fail because the boost.fiber libraries develop branch that is used for testing has now raised its dependency to the current boost develop branch (all the CPU tests except the ones using boost develop fail).

BenjaminW3:
Could you please create another pull request removing the `&& [ ... 1.59.0 ... ]` from [here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/master/.travis.yml#L328) and changing `boost-1.59.0` to `develop` [here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/master/.travis.yml#L330) and in the following echo line?
This hopefully allows to fix the CI and re-execute the tests for this branch.

BenjaminW3:
Have you tested the functionality for the synchronous stream?

psychocoderHPC:
No I check it tomorrow.

psychocoderHPC:
Tested with PIConALPAKA sync and async cpu streams.

include/alpaka/event/EventCpu.hpp
```diff
@@ -534,14 +534,13 @@ namespace alpaka
...
                     // This is forwarded to the lambda that is enqueued into the stream to ensure that the event implementation is alive as long as it is enqueued.
                     auto spEventCpuImpl(event.m_spEventCpuImpl);
-
+                    std::lock_guard<std::mutex> lk(spEventCpuImpl->m_Mutex);
+                    if(!spEventCpuImpl->m_bIsReady)
                     {
-                        std::lock_guard<std::mutex> lk(spEventCpuImpl->m_Mutex);
                         spEventCpuImpl->m_bIsWaitedFor = true;
+                        // NOTE: Difference to async version: directly wait for event.
+                        wait::wait(spEventCpuImpl);
```
BenjaminW3: I am not sure if the synchronous case will work on all platforms.
This `WaiterWaitFor` specialization locks the mutex and directly calls `wait::wait(...)` on the event. This in turn calls `CurrentThreadWaitFor` (line 443) which tries to lock the mutex again leading to a deadlock. This would require a recursive mutex which is not used currently.

I would say that we can delete lines 534-540 as well as line 543 and keep only this line and its comment. This would solve the recursive mutex call. And remove duplicate code that is executed in the  `wait::wait(...)` -> `CurrentThreadWaitFor` anyhow.

erikzenker:
@BenjaminW3 to not lock the mutex again makes totally sense. Thus, this code would lock in each run? Just ask because @psychocoderHPC said he also tested the sync cpu streams. Nevertheless, I think we can remove the mentioned lines.

psychocoderHPC:
ready to merge. Tested sync stream with `PIConALPAKA`

PR #69, opened 2015-10-15
Fix race condition

erikzenker:
Same as #68 but with fixed CI


erikzenker:
travis in particular pr -> close this one !

PR #71, opened 2015-10-15
Fix boost fiber ci

BenjaminW3:
Because we should not rely on the compilability of the boost develop branch, the boost.fiber tests should not enforce this branch.
The boost.fiber tests will be disabled instead for non boost develop branch builds.


psychocoderHPC:
I think there is no need to push to master, all changes goes ofer develop and later on as release to master.

BenjaminW3:
Oh, this was my mistake. Wrong branch...

PR #79, opened 2015-10-26
fix shared memory allocation

psychocoderHPC:
- change the interface of `allocVar` and `allocArr` (add unique id)
- add unique id to `allocArr` *example/sharedMem* code

If shared memory is allocated on `alpaka::acc::AccGpuCudaRt<...>` more than for one array or variable with the same type than all returned memory references points to the same memory address.
e.g.
```C++
...
auto& a = alpaka::block::shared::allocVar<std::uint32_t>(acc);
auto& b = alpaka::block::shared::allocVar<std::uint32_t>(acc);
assert( &a != &b ); // error both memory addresses are equal 
```

Shared memory allocations on CUDA inside a class are similar to static class variables. All classes with the same template signature has the same static variable.
```C++
template<typename T>
struct Foo {
static T mem;
};
```
`mem` is equal for all instances of Foo<T>  with the the same `T`.
The way to avoid this bug is to create a unique template signature for each in kernel shared memory allocation. Thus the problem that it is not possible in C++ to generate unique id's at compile time  the user must take care about this.

With this pull request the example looks like this:
```C++
...
auto& a = alpaka::block::shared::allocVar<std::uint32_t,0>(acc); //instead of manual counter __CONTER__ can used
auto& b = alpaka::block::shared::allocVar<std::uint32_t,1>(acc);
assert( &a != &b ); // OK!!
```

### DIFF
--- a/test/integ/sharedMem/src/main.cpp
+++ b/test/integ/sharedMem/src/main.cpp
@@ -68,8 +68,8 @@ public:
         std::uint32_t * const pBlockShared(acc.template getBlockSharedExternMem<std::uint32_t>());
 
         // Get some shared memory (allocate a second buffer directly afterwards to check for some synchronization bugs).
-        //std::uint32_t * const pBlockShared1(alpaka::block::shared::allocArr<std::uint32_t, TnumUselessWork::value>());
-        //std::uint32_t * const pBlockShared2(alpaka::block::shared::allocArr<std::uint32_t, TnumUselessWork::value>());
+        //std::uint32_t * const pBlockShared1(alpaka::block::shared::allocArr<std::uint32_t, TnumUselessWork::value,__COUNTER__>());
+        //std::uint32_t * const pBlockShared2(alpaka::block::shared::allocArr<std::uint32_t, TnumUselessWork::value,__COUNTER__>());
 
         // Calculate linearized index of the thread in the block.
         std::size_t const blockThreadIdx1d(alpaka::idx::getIdx<alpaka::Block, alpaka::Threads>(acc)[0u]);
--- a/include/alpaka/block/shared/BlockSharedAllocCudaBuiltIn.hpp
+++ b/include/alpaka/block/shared/BlockSharedAllocCudaBuiltIn.hpp
@@ -1,6 +1,6 @@
 /**
  * \file
- * Copyright 2014-2015 Benjamin Worpitz
+ * Copyright 2014-2015 Benjamin Worpitz, Rene Widera
  *
  * This file is part of alpaka.
  *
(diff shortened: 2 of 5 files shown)


erikzenker:
I ask for a solution of this problem on stackoverflow [here](http://stackoverflow.com/questions/33345498/cuda-shared-memory-wrapped-in-class-points-to-same-memory?noredirect=1#comment54487402_33345498). Maybe there is somebody out there who has some idea to solve it.

BenjaminW3:
Thanks for reporting and 'fixing' it. I always feared that this will be an issue.
I will wait some days for answers in the SO post before merging. Maybe there is a compile-time workaround available somewhere ;-)

psychocoderHPC:
What I found was: http://b.atch.se/posts/constexpr-counter/#solution but the problem was that it not run with nvcc? (I have not tested other compilers)

BenjaminW3:
In #85 I have proposed a version that hides the explicit parameter and uses the trick mentioned in your link to generate compile-time IDs.

ax3l:
ha, finally `constexpr` ;P

psychocoderHPC:
Since another two days I tested how we can get #85  compiling on gpu. It looks like it is currently not possible.
Therefore I vote that 
 - I rebase this branch and we change the alpaka interface for shared memory allocation
 - @BenjaminW3 accept this changes

I can move the ID that it is the last parameter thus we can easily add a `unique()` if there come the day that we found a way for it.

psychocoderHPC:
I updated this pull request and use the naming from #85

test/integ/sharedMem/src/main.cpp
```diff
@@ -68,8 +68,8 @@ class SharedMemKernel
         std::uint32_t * const pBlockShared(acc.template getBlockSharedExternMem<std::uint32_t>());
 
         // Get some shared memory (allocate a second buffer directly afterwards to check for some synchronization bugs).
-        //std::uint32_t * const pBlockShared1(alpaka::block::shared::allocArr<std::uint32_t, TnumUselessWork::value>());
-        //std::uint32_t * const pBlockShared2(alpaka::block::shared::allocArr<std::uint32_t, TnumUselessWork::value>());
+        //std::uint32_t * const pBlockShared1(alpaka::block::shared::allocArr<__COUNTER__,std::uint32_t, TnumUselessWork::value>());
+        //std::uint32_t * const pBlockShared2(alpaka::block::shared::allocArr<__COUNTER__,std::uint32_t, TnumUselessWork::value>());
```
BenjaminW3: Shouldn`t the unique id be the last template argument?

psychocoderHPC: right my mistake

psychocoderHPC:
updated

btw: I tested the new shared memory implementation on GPU (it runs)

PR #84, opened 2015-10-26
add CUDA 7.5 to travis

BenjaminW3:
CUDA 7.5 requires at least Ubuntu 14.04. Travis now allows to use a beta 14.04 image that we can use to test.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,9 @@ language:
 os:
     - linux
 
+sudo: required
+dist: trusty
+
 ################################################################################
 #
 ################################################################################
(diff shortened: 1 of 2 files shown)


bussmann:
Back to the future.

BenjaminW3:
As my tests have shown, CUDA 7.5 does not correctly compile alpaka when using clang, even so the documentation states the added support for this compiler on linux (see the 2 failed tests [here](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/87515937)). Therfore, CUDA will be disabled in the CI tests for clang for now.

psychocoderHPC:
Currently we can not switch to cuda 7.5. We had no system with cuda 7.5. The possibilities to break cuda 7.0 support is to high.

BenjaminW3:
This pull request does NOT raise the minimum CUDA version that is required. It only adds CUDA 7.5 to the travis tests, because this version was previously untested. Therefore, it is safe to merge ;-)

psychocoderHPC:
Oh thanks than everything is fine

BenjaminW3:
Is there anything preventing a merge of this?

Issue #77, opened 2015-10-22
create simple examples


psychocoderHPC:
We need to add a section with `simple` examples to our example folder.
The current examples are hard to understand for beginners.

**simple example means:**
 - only one file
 - can only use one accelerator 
 - used no extern defined helper functions but *alpaka*
 - no time messurement

erikzenker:
I would suggest to create a 'test' folder where we put the current examples, since these examples are more integration tests than example to have a quick start. Then I would fill the example folder with a bunch of simple examples like @psychocoderHPC suggested.

BenjaminW3:
I agree with both of you.

BenjaminW3:
Closed because the first few examples are already merged.

PR #85, opened 2015-10-28
Use implicit id counter for shared memory allocation

BenjaminW3:
This branch implements a compile time counter for the shared memory allocation that is compiling on all platforms.
It is based on the code from [Filip Roséen](http://b.atch.se/posts/constexpr-counter/).
At least in C++11 and C++14 this code is valid. The C++ Standards Committee is currently debating on wheter this should be legal or not in C++17. See [here](https://groups.google.com/a/isocpp.org/forum/#!msg/std-discussion/M6aJMH_ewoM/-cV4Rq0ou5QJ).

- [ ] [Suppress](https://gcc.gnu.org/onlinedocs/gcc/Diagnostic-Pragmas.html) the high number of warnings. (-Wundefined-internal / -Wnon-template-friend)
- [ ] Increase the maximum counter value from 64 to something reasonable. However, this could exceed the maximum template instantiation depth.
- [ ] Add the solution to the [Stackoverflow thread](http://stackoverflow.com/questions/33345498/cuda-shared-memory-wrapped-in-class-points-to-same-memory).

### DIFF
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -188,6 +188,7 @@
 #include <alpaka/meta/ForEachType.hpp>
 #include <alpaka/meta/Metafunctions.hpp>
 #include <alpaka/meta/NdLoop.hpp>
+#include <alpaka/meta/UniqueId.hpp>
 #include <alpaka/meta/StdTupleAsMplSequence.hpp>
 #include <alpaka/meta/Transform.hpp>
 
--- a/include/alpaka/block/shared/st/Traits.hpp
+++ b/include/alpaka/block/shared/st/Traits.hpp
@@ -23,6 +23,8 @@
 
 #include <alpaka/core/Common.hpp>   // ALPAKA_FN_HOST_ACC
 
+#include <alpaka/meta/UniqueId.hpp> // meta::uniqueId
+
 #include <type_traits>              // std::enable_if, std::is_base_of, std::is_same, std::decay
 
 namespace alpaka
@@ -86,8 +88,8 @@ namespace alpaka
                 ALPAKA_NO_HOST_ACC_WARNING
                 template<
                     typename T,
-                    std::size_t TuniqueId,
-                    typename TBlockSharedMemSt>
+                    typename TBlockSharedMemSt,
+                    std::size_t TuniqueId = meta::uniqueId()>
                 ALPAKA_FN_HOST_ACC auto allocVar(
                     TBlockSharedMemSt const & blockSharedMemSt)
                 -> T &
@@ -114,8 +116,8 @@ namespace alpaka
                 template<
                     typename T,
                     std::size_t TnumElements,
-                    std::size_t TuniqueId,
-                    typename TBlockSharedMemSt>
+                    typename TBlockSharedMemSt,
+                    std::size_t TuniqueId = meta::uniqueId()>
                 ALPAKA_FN_HOST_ACC auto allocArr(
                     TBlockSharedMemSt const & blockSharedMemSt)
                 -> T *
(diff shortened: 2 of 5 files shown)


psychocoderHPC:
Thanks for this workaround I will test it next week.

psychocoderHPC:
There is no need to add the id to the CPU alloc structs. The use of an id for shared memory can fully hidden in the GPU specialization.

psychocoderHPC:
I tested you current branch with `cuda7.0` and two shared memory variables. Both are equal in the kernel which means the current implementation is not working.

BenjaminW3:
I might have an Idea how to fix this and will test it in the next days.

psychocoderHPC:
btw: I tested the original solution of [Filip Roséen](http://b.atch.se/posts/constexpr-counter/) with g++4.9 on the host and always get the some ID.

psychocoderHPC:
OK the original id generation on CPU is not running because of a problem that always the path for `BOOST_COMP_MSVC` is used also in linux with gcc.

include/alpaka/core/UniqueId.hpp
```diff
@@ -0,0 +1,152 @@
...
+                //-----------------------------------------------------------------------------
+                friend constexpr std::size_t adl_flag(flag<N>)
+                {
+                    return N;
+                }
+
+                static constexpr std::size_t value = N;
+            };
+
+#ifdef BOOST_COMP_MSVC
```
psychocoderHPC: This must be  `#if` else always the first code path is used

psychocoderHPC:
I pushed a small update to your pull request that fix some small mistakes. https://github.com/BenjaminW3/alpaka/pull/1

Shared memory id generation on is working but not with cuda 7.0

BenjaminW3:
Thanks, moving the `uniqueId()` call to the default template parameter was the same idea I had.

include/alpaka/core/UniqueId.hpp
BenjaminW3: My mistake...

psychocoderHPC:
Yes if it is not a default parameter at this place no new ids are generated.

BenjaminW3:
Could you please apply #84 so that the tests are executed with CUDA 7.0 and 7.5?

BenjaminW3:
I will do a part of what #77 does and will move the current integration test *examples* to `test/integ` and will add a unit test into `test/unit` that directly tests the unique shared memory allocation to return unique memory adresses.

include/alpaka/core/UniqueId.hpp
```diff
@@ -0,0 +1,152 @@
...
+            //-----------------------------------------------------------------------------
+            //! The matcher.
+            //-----------------------------------------------------------------------------
+            template<
+                std::size_t N,
+                class = char[noexcept(adl_flag(flag<N>{})) ? +1u : -1u]>
```
psychocoderHPC: The current problem with NVCC is that it not supports `noexcept` on this way we need it here. NVCC implements [this](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2010/n3050.html#operator-noexcept-expression).
The second pass in line 92 is also not working with nvcc :-(

BenjaminW3:
I tried a lot of different things but did not get it working with nvcc (neither 7.0 nor 7.5). Therefore I will close this for now. However, clang, MSVC and gcc did work as expected.

ax3l:
> The C++ Standards Committee is currently debating on wheter this should be legal or not in C++17.

@BenjaminW3 btw, do you know how the discussion ended? ;)

BenjaminW3:
@ax3l You can have a look at the C++ core issue [here](http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_active.html#2118)

PR #96, opened 2015-11-10
Hello World example

erikzenker:
I wrote a Hello world example, just to give a start for the user friendly examples.
I think we should provide as easy as possible examples and each example
should only show a small number of alpaka features. Furthermore, the examples
should provide some documentation, which explaines why something must be
used this or that way.

 I don't know if my code is best practice, but I would be glad if you have some
advices. With good examples you give the users the best starting point into
your library.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -497,7 +497,9 @@ script:
     #-------------------------------------------------------------------------------
 
     # Build and execute all examples.
+    - ./travis/compileExec.sh "example/helloWorld" ./helloWorld    
     - ./travis/compileExec.sh "example/vectorAdd" ./vectorAdd
+
 
 ################################################################################
 #
--- a/example/helloWorld/CMakeLists.txt
+++ b/example/helloWorld/CMakeLists.txt
@@ -1,0 +1,45 @@
+#
+# \file
+# Copyright 2014-2015 Erik Zenker
+#
+# This file is part of alpaka.
+#
+# alpaka is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# alpaka is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
+# along with alpaka.
+# If not, see <http://www.gnu.org/licenses/>.
+#
+#
+
+project(alpaka-helloworld)
+cmake_minimum_required(VERSION 3.0.1)
+
+################################################################################
+# Find alpaka
+################################################################################
+set(ALPAKA_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../../" CACHE STRING  "The location of the alpaka library")
+list(APPEND CMAKE_MODULE_PATH "${ALPAKA_ROOT}")
+
+find_package("alpaka" REQUIRED)
+include_directories(SYSTEM ${alpaka_INCLUDE_DIRS})
+add_definitions(${alpaka_DEFINITIONS})
+add_definitions(${ALPAKA_DEV_COMPILE_OPTIONS})
+set(LIBS ${LIBS} ${alpaka_LIBRARIES})
+
+###############################################################################
+# Executable
+###############################################################################
+set(helloWorld "helloWorld")
+file(GLOB SRCFILES "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp")
+
+alpaka_add_executable(${helloWorld} ${SRCFILES})
+target_link_libraries(${helloWorld} ${LIBS})
--- a/example/helloWorld/src/helloworld.cpp
+++ b/example/helloWorld/src/helloworld.cpp
@@ -1,0 +1,205 @@
+/**
+ * \file
+ * Copyright 2014-2015 Erik Zenker
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * alpaka is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with alpaka.
+ * If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+// Alpaka
+#include <alpaka/alpaka.hpp>
+
+// STL
+#include <iostream>
+
+
+/**
+ * Hello World Kernel
+ *
+ * Prints "[x, y, z][gtid] Hello World" where tid is
+ * the global thread number.
+ */
+struct HelloWorldKernel {
+    template <typename T_Acc>
+    ALPAKA_FN_ACC void operator()( T_Acc const &acc) const {
+        /**
+         * In the most cases the parallel work distibution depends
+         * on the current index of a thread and how many threads
+         * exist overall. These information can be obtained by
+         * getIdx() and getWorkDiv(). In this example these
+         * values are obtained for a global scope.
+         */
+        auto globalThreadIdx    = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+        auto globalThreadExtent = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+
+        /**
+         * Map the three dimensional thread index into a
+         * one dimensional thread index space. We call it
+         * linearize the thread index.
+         */
+        auto linearizedGlobalThreadIdx = alpaka::core::mapIdx<1u>(globalThreadIdx,
+                                                                  globalThreadExtent);
+
+        /**
+         * Each thread prints a hello world to the terminal
+         * together with the global index of the thread in
+         * each dimension and the linearized global index.
+         * Mind, that alpaka uses the mathematical index
+         * order [z][y][x] where the last index is the fast one.
+         */
+        printf("[z:%u, y:%u, x:%u][linear:%u] Hello World\n",
+               static_cast<unsigned>(globalThreadIdx[0]),
+               static_cast<unsigned>(globalThreadIdx[1]),
+               static_cast<unsigned>(globalThreadIdx[2]),
+               static_cast<unsigned>(linearizedGlobalThreadIdx[0]));
+
+    }
+
+};
+
+int main() {
+
+    /**
+     * Define accelerator types
+     *
+     * It is possible to choose from a set of accelerators
+     * that are defined in the alpaka::acc namespace e.g.:
+     * - AccGpuCudaRt
+     * - AccCpuThreads
+     * - AccCpuFibers
+     * - AccCpuOmp2Threads
+     * - AccCpuOmp2Blocks
+     * - AccCpuOmp4
+     * - AccCpuSerial
+     *
+     * Each accelerator has strengths and weaknesses. Therefore,
+     * they need to be choosen carefully depending on the actual
+     * use case. Furthermore, some accelerators only support a
+     * particular workdiv, but workdiv can also be generated
+     * automatically.
+     */
+    using Dim     = alpaka::dim::DimInt<3>;
+    using Size    = std::size_t;
+    using Host    = alpaka::acc::AccCpuSerial<Dim, Size>;
+    using Acc     = alpaka::acc::AccCpuSerial<Dim, Size>;
+    using Stream  = alpaka::stream::StreamCpuSync;
+    using DevAcc  = alpaka::dev::Dev<Acc>;
+    using DevHost = alpaka::dev::Dev<Host>;
+    using WorkDiv = alpaka::workdiv::WorkDivMembers<Dim, Size>;
+
+
+    /**
+     * Get the first devices
+     *
+     * The accelerator only defines how something should be
+     * parallized, but a device is the real entity which will
+     * run the parallel programm. The device can be choosen
+     * by id (0 to the number of devices minus 1) or you
+     * can also retrieve all devices in a vector (getDevs()).
+     * In this example the first devices is choosen.
+     */
+    DevAcc  devAcc  (alpaka::dev::DevMan<Acc>::getDevByIdx(0));
+    DevHost devHost (alpaka::dev::DevMan<Host>::getDevByIdx(0));
+
+    /**
+     * Create a stream to the accelerator device
+     *
+     * A stream can be interpreted as the work queue
+     * of a particular device. Streams are filled with
+     * executors and alpaka takes care that these
+     * executors will be executed. Streams are provided in
+     * async and sync variants.
+     * The example stream is a sync stream to a cpu device,
+     * but it also exists an async stream for this
+     * device (StreamCpuAsync).
+     */
+    Stream  stream  (devAcc);
+
+
+    /**
+     * Init workdiv
+     *
+     * A kernel is executed for each element of a
+     * n-dimensional grid distinguished by the element indices.
+     * The work division defines the number of kernel instantiations as
+     * well as the type of parallelism used by the executor.
+     * Different accelerators have different requirements on the work
+     * division. For example, the sequential accelerator can not
+     * provide any thread level parallelism (synchronizable as well as non synchronizable), 
+     * whereas the CUDA accelerator can spawn hundreds of synchronizing
+     * and non synchronizing threads at the same time.
+     *
+     * The workdiv is divided in three levels of parallelization:
+     * - grid-blocks:      The number of blocks in the grid (parallel, not synchronizable)
+     * - block-threads:    The number of threads per block (parallel, synchronizable). 
+     *                     Each thread executes one kernel invocation.
+     * - thread-elements:  The number of elements per thread (sequential, not synchronizable). 
+     *                     Each kernel has to execute its elements sequentially.
+     *
+     * - Grid     : consists of blocks
+     * - Block    : consists of threads
+     * - Elements : consists of elements
+     *
+     * Threads in the same grid can access the same global memory,
+     * while threads in the same block can access the same shared
+     * memory. Elements are supposed to be used for vectorization.
+     * Thus, a thread can process data element size wise with its
+     * vector processing unit.
+     */
+    alpaka::Vec<Dim, Size> const elementsPerThread(static_cast<Size>(1),
+                                                   static_cast<Size>(1),
+                                                   static_cast<Size>(1));
+
+    alpaka::Vec<Dim, Size> const threadsPerBlock(static_cast<Size>(1),
+                                                 static_cast<Size>(1),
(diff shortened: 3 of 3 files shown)


psychocoderHPC:
I merged #89, please add your example to `.travis.yml`

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+     * Thus, a thread can process data in a element wise with its
+     * vector processing unit.
+     *
+     */
+    alpaka::Vec<Dim, Size> const elem (static_cast<Size>(1), 
+					static_cast<Size>(1), 
+					static_cast<Size>(1)); 
+    
+    alpaka::Vec<Dim, Size> const blocks (static_cast<Size>(128),
+					 static_cast<Size>(1),
```
psychocoderHPC: not the same alignment like in line 138

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+#include <iostream>
+
+// Alpaka
+#include <alpaka/alpaka.hpp>
+
+
+/**
+ * Helper functions that calculates the global thread id depending
+ * on the number of threads set by the workdiv.
+ *
```
psychocoderHPC: should we remove this empty line in all comments?

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
+/**
+ * \file
+ * Copyright 2014-2015 Benjamin Worpitz
```
psychocoderHPC: I think your name is Erik^^

erikzenker: I Added me as author ^^

example/helloWorld/src/helloworld.cpp
erikzenker: We can, I dont mind.

example/helloWorld/src/helloworld.cpp
erikzenker: dare you emacs :triumph:

erikzenker:
Well thx for the review, an update is available

example/helloWorld/src/helloworld.cpp
ax3l: better keep it in the top line, we are all lazy people and will forget to update multiple positions ;)

ax3l: there is also a \copyright in doxygen, if you feel fancy. but that's more for a per-project level

erikzenker:
Looks like cmake does not find openMP support in the clang cases ...therefore the test fails. I could use the sequential accelerator since this one is supported by all compilers.

erikzenker:
@psychocoderHPC CI says yes, but only for the AccCpuSerial accelerator. The vectorAdd example also just uses this simple accelerator. 

I think our example should not all base on this accelerator (since it is no)

BenjaminW3:
clang does not (currently) support OpenMP out of the box.
Therefore, you can not unconditionally use the OpenMP accelerator in such an example. However the same is the case for the CUDA accelerator because it is not available on all platforms.

example/helloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+################################################################################
+# Find alpaka
+################################################################################
+SET(ALPAKA_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../../" CACHE STRING  "The location of the alpaka library")
+LIST(APPEND CMAKE_MODULE_PATH "${ALPAKA_ROOT}")
+
+find_package("alpaka" REQUIRED)
+include_directories(SYSTEM ${alpaka_INCLUDE_DIRS})
+add_definitions(${alpaka_DEFINITIONS})
+ADD_DEFINITIONS(${alpaka_DEFINITIONS} ${ALPAKA_DEV_COMPILE_OPTIONS})
```
BenjaminW3: You have added ${alpaka_DEFINITIONS} twice.
Moreover, you should use either upper or lower case consistently for the built-in functions.

example/helloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,35 @@
...
+set(LIBS ${LIBS} ${alpaka_LIBRARIES})
+
+
+################################################################################
+# Compiler Flags
+################################################################################
+# GNU
+if(CMAKE_COMPILER_IS_GNUCXX)
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wall")
+  set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -Wextra")
```
BenjaminW3: These flags (Wall, Wextra) are part of `${ALPAKA_DEV_COMPILE_OPTIONS}` as well as others for other platforms.
Therefore you can remove this part.

example/helloWorld/src/helloworld.cpp
BenjaminW3: Right, please use the copyright line.
If the author flag should be used, this should be done for the whole project.

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+ * Kernel
+ *
+ * Prints "[tid] Hello World" where tid is
+ * the global thread number.
+ *
+ */
+struct HelloWorldKernel {
+    template <typename T_Acc>
+    ALPAKA_FN_ACC void operator()( T_Acc const &acc) const {
+	printf("[%u] Hello World\n", static_cast<unsigned>(globalThreadIdx(acc)));
```
BenjaminW3: Sidenote: This works on all current platforms but may not be possible on future ones (e.g. Xeon Phi).
An abstract printf like method should be implemented as seperate task.

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+     * automatically.
+     *
+     */
+    using Dim     = alpaka::dim::DimInt<3>;  
+    using Size    = std::size_t;
+    using Host    = alpaka::acc::AccCpuSerial<Dim, Size>;
+    using Acc     = alpaka::acc::AccCpuSerial<Dim, Size>;
+    using Stream  = alpaka::stream::StreamCpuSync;
+    using DevAcc  = alpaka::dev::Dev<Acc>;
+    using DevHost = alpaka::dev::DevCpu;
```
BenjaminW3: For consistency I would prefer `alpaka::dev::Dev<Host>` even though it maps to `DevCpu` in the end.

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+     * parallized, but a device is the real entity which will
+     * run the parallel programm. The device can be choosen
+     * by id (0 to the number of devices minus 1) or you
+     * can also retrieve all devices in a vector (getDevs()).
+     * In this example the first devices is choosen.
+     *
+     */
+    DevAcc  devAcc  (alpaka::dev::DevMan<Acc>::getDevByIdx(0));
+    DevHost devHost (alpaka::dev::DevMan<Host>::getDevByIdx(0));    
+    Stream  stream  (devAcc);
```
BenjaminW3: Could you additionally add a note about the stream?

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+ * on the number of threads set by the workdiv.
+ *
+ */
+template <typename T_Acc>
+size_t globalThreadIdx(T_Acc const &acc){
+    auto threadsExtent = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+    auto threadIdx = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+    
+    auto globalThreadIdx =
+	threadIdx[0]
```
BenjaminW3: I see tabs in here. Personally I would prefer them, but for consistency spaces are used ;-)

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+
+
+    /**
+     * Run kernel
+     *
+     * Kernels need to be provided as classes or structs
+     * which provide a public operator(). This operator is
+     * the actual method that should be accelerated. An
+     * object of the kernel is used to create an execution
+     * unit and this unit is finally enqueue into an
```
BenjaminW3: enqueue**d**

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+    /**
+     * Run kernel
+     *
+     * Kernels need to be provided as classes or structs
+     * which provide a public operator(). This operator is
+     * the actual method that should be accelerated. An
+     * object of the kernel is used to create an execution
+     * unit and this unit is finally enqueue into an 
+     * accelerator stream. The enqueuing can be done 
+     * synchronous or asynchronous depending on the choosen
```
BenjaminW3: synchronous**ly** or asynchronous**ly**

BenjaminW3:
Please do not be afraid by the number of comments. I am happy about the very good introductoral example!

.travis.yml
```diff
@@ -498,6 +498,7 @@ script:
 
     # Build and execute all examples.
     - ./travis/compileExec.sh "example/vectorAdd" ./vectorAdd
+    - ./travis/compileExec.sh "example/helloWorld" ./helloWorld
```
BenjaminW3: I would sort them alphabetically or by difficulty. In either case yours would be the first!

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with alpaka.
+ * If not, see <http://www.gnu.org/licenses/>.
+ *
+ * \author Erik Zenker
+ *
+ */
+
+// STL
+#include <iostream>
```
BenjaminW3: Could you please move this header under the alpaka inclusion for consistent inclusion in order of increasing generality?

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,177 @@
...
+ */
+template <typename T_Acc>
+size_t globalThreadIdx(T_Acc const &acc){
+    auto threadsExtent = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+    auto threadIdx = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+    
+    auto globalThreadIdx =
+	threadIdx[0]
+	+ threadIdx[1] * threadsExtent[0]
+	+ threadIdx[2] * threadsExtent[0] * threadsExtent[1];
```
BenjaminW3: alpaka already provides a utility for index mapping of arbitrary dimensions (currently only specialized for up to 3 but extensible for more dimensions):
```
auto globalThreadIdx =
    alpaka::core::mapIdx<1u>(
        threadIdx,
        threadsExtent);
```

example/helloWorld/src/helloworld.cpp
erikzenker: It also looks more uniform.

example/helloWorld/src/helloworld.cpp
erikzenker: thx

.travis.yml
erikzenker: ok

example/helloWorld/src/helloworld.cpp
erikzenker: Ohh, that is interesting, because I usually go from the more general to the less general libraries in my projects. I will switch that behaviour in alpaka. Is it common in c++ projects ?

example/helloWorld/src/helloworld.cpp
erikzenker: Cool, I learned something new. I need to write more examples!

erikzenker:
@BenjaminW3 thank you for the review. I fixed all of your suggestions and will take your advices to tune the other examples I already wrote.

example/helloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,22 @@
...
+
+################################################################################
+# Find alpaka
+################################################################################
+set(ALPAKA_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../../" CACHE STRING  "The location of the alpaka library")
+list(APPEND CMAKE_MODULE_PATH "${ALPAKA_ROOT}")
+
+find_package("alpaka" REQUIRED)
+include_directories(SYSTEM ${alpaka_INCLUDE_DIRS})
+add_definitions(${alpaka_DEFINITIONS})
```
BenjaminW3: Could you please re-add `${ALPAKA_DEV_COMPILE_OPTIONS}`?

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Create a stream to the accelerator device
+     *
+     * A stream can be interpreted as the work queue
+     * of a particular device. Streams are filled with
+     * execution units and alpaka takes care that these
+     * units will be executed. Streams are provided in
+     * async and sync variants. 
+     * The example stream is a sync stream to a Cpu device, 
+     * but it also exist an async stream for this
```
BenjaminW3: exist**s**

psychocoderHPC:
I assign this pull request to @BenjaminW3

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+    DevAcc  devAcc  (alpaka::dev::DevMan<Acc>::getDevByIdx(0));
+    DevHost devHost (alpaka::dev::DevMan<Host>::getDevByIdx(0));
+
+    
+    /**
+     * Create a stream to the accelerator device
+     *
+     * A stream can be interpreted as the work queue
+     * of a particular device. Streams are filled with
+     * execution units and alpaka takes care that these
```
BenjaminW3: In the code the are called executor and not execution unit. Maybe it is better to use the library internal naming here and in the following comments.

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+    /**
+     * Init workdiv
+     *
+     * The workdiv defines how many instances of a
+     * kernel can be spawned in parallel. How many kernels
+     * are really spawned finally depends on the choosen
+     * accelerator e.g. the sequential accelerator will
+     * only spawn one kernel at a time, whereas a CUDA
+     * accelerator can spawn hundreds of kernels at the
+     * same time.
```
BenjaminW3: I have reworded it a bit:
A kernel is executed for each element of a n-dimensional grid distinguished by the element indices.
The work division defines the number of instantiations as well as the type of parallelism used.
Different accelerators have different requirements on the work division. For example, the seuential accelerator can not provide any thread level parallelism (synchronizable as well as non synchronizable), whereas the CUDA accelerator can spawn hundreds of synchronizing and non synchronizing threads at the same time.

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * The workdiv is divided in three levels of parallelization:
+     * - Grid     : consists of blocks
+     * - Block    : consists of threads
+     * - Elements : consists of elements
```
BenjaminW3:
```
grid-blocks:      The number of blocks in the grid (parallel, not synchronizable)
block-threads:    The number of threads per block (parallel, synchronizable). Each thread executes one kernel invocation.
thread-elements:  The number of elements per thread (sequential, not synchronizable). Each kernel has to execute its elements sequentially.
```

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+     *
+     * Threads in the same grid can access the same global memory,
+     * while threads in the same block can access the same shared
+     * memory. Elements are supposed to be used for vectorization.
+     * Thus, a thread can process data element size wise with its
+     * vector processing unit.
+     *
+     */
+    alpaka::Vec<Dim, Size> const elem (static_cast<Size>(1),
```
BenjaminW3: elemsPerBlock

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+					static_cast<Size>(1),
+					static_cast<Size>(1));
+
+    alpaka::Vec<Dim, Size> const blocks (static_cast<Size>(1),
```
BenjaminW3: `blockThreads` or `threadsPerBlock`

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+					 static_cast<Size>(1),
+					 static_cast<Size>(1));
+
+    alpaka::Vec<Dim, Size> const grid (static_cast<Size>(128),
```
BenjaminW3: `gridBlocks` or `blocksPerGrid`

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+// STL
+#include <iostream>
+
+
+/**
+ * Helper functions that calculates the global thread id depending
+ * on the number of threads set by the workdiv.
+ */
+template <typename T_Acc>
+size_t globalThreadIdx(T_Acc const &acc){
```
BenjaminW3: Could you rename this function to `linearizedGlobalThreadIndex` because what you call `threadIdx` inside is already the real  `globalThreadIdx` that is identical to what I call `gridThreadIdx`.

example/helloWorld/src/helloworld.cpp
```diff
@@ -0,0 +1,190 @@
...
+/**
+ * Hello World Kernel
+ *
+ * Prints "[tid] Hello World" where tid is
+ * the global thread number.
+ */
+struct HelloWorldKernel {
+    template <typename T_Acc>
+    ALPAKA_FN_ACC void operator()( T_Acc const &acc) const {
+        printf("[%u] Hello World\n", static_cast<unsigned>(globalThreadIdx(acc)));
```
BenjaminW3: Could you please log the 3 dimensional as well as the linearized thread index? Because else there is little sense in calling the kernel with 3 dimensions. This would also show that alpaka uses the mathematical index order [z][y][x] where the last index is the fast one (you are calling the kernel with 128 blocks/threads in the z dimension).

example/helloWorld/src/helloworld.cpp
erikzenker: You are right, I was not sure how to call these puppets.

example/helloWorld/src/helloworld.cpp
psychocoderHPC: In this example we should only start a 1D kernel. This example should as simple as posible.

erikzenker: I am not sure. I think we don't expect to much if we provide a three dimensional kernel here. Because multi dimensional kernel is a fundamental feature of alpaka.

example/helloWorld/src/helloworld.cpp
erikzenker: That makes sense, since the user will see how to obtain the idx in several dimensions.

erikzenker:
Update, thx for the formulation help! :smiley:

example/helloWorld/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+project(alpaka-helloworld)
```
BenjaminW3: The license header is missing. Just copy it from the other examples and replace the name.

BenjaminW3:
When you have:
- [ ] added the license header to the CMake file
- [ ] replaced my name by yours in the source and the CMake license header
- [ ] removed the \author tag
- [ ] and finally squashed all commits together to a single one
we are ready!

erikzenker:
Ready to merge :sunny:

PR #98, opened 2015-11-17
buffer copy example

erikzenker:
This example creates several buffers, inits these buffers and performs
copy operations.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -499,7 +499,8 @@ script:
     #-------------------------------------------------------------------------------
 
     # Build and execute all examples.
-    - ./travis/compileExec.sh "example/helloWorld" ./helloWorld    
+    - ./travis/compileExec.sh "example/bufferCopy" ./bufferCopy        
+    - ./travis/compileExec.sh "example/helloWorld" ./helloWorld
     - ./travis/compileExec.sh "example/vectorAdd" ./vectorAdd
 
 
--- a/example/bufferCopy/CMakeLists.txt
+++ b/example/bufferCopy/CMakeLists.txt
@@ -1,5 +1,6 @@
 #
-# Copyright 2015 Benjamin Worpitz
+# \file
+# Copyright 2014-2015 Erik Zenker
 #
 # This file is part of alpaka.
 #
@@ -17,19 +18,28 @@
 # along with alpaka.
 # If not, see <http://www.gnu.org/licenses/>.
 #
+#
+
+project(alpakaBufferCopy)
+cmake_minimum_required(VERSION 3.0.1)
 
 ################################################################################
-# Required CMake version.
+# Find alpaka
 ################################################################################
+set(ALPAKA_ROOT "${CMAKE_CURRENT_SOURCE_DIR}/../../" CACHE STRING  "The location of the alpaka library")
+list(APPEND CMAKE_MODULE_PATH "${ALPAKA_ROOT}")
 
-CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
+find_package("alpaka" REQUIRED)
+include_directories(SYSTEM ${alpaka_INCLUDE_DIRS})
+add_definitions(${alpaka_DEFINITIONS})
+add_definitions(${ALPAKA_DEV_COMPILE_OPTIONS})
+set(LIBS ${LIBS} ${alpaka_LIBRARIES})
 
-PROJECT("alpakaExamples")
+###############################################################################
+# Executable
+###############################################################################
+set(bufferCopy "bufferCopy")
+file(GLOB SRCFILES "${CMAKE_CURRENT_SOURCE_DIR}/src/*.cpp")
 
-################################################################################
-# Add subdirectories.
-################################################################################
-
-ADD_SUBDIRECTORY("bufferCopy/")
-ADD_SUBDIRECTORY("helloWorld/")
-ADD_SUBDIRECTORY("vectorAdd/")
+alpaka_add_executable(${bufferCopy} ${SRCFILES})
+target_link_libraries(${bufferCopy} ${LIBS})
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -1,0 +1,352 @@
+/**
+ * \file
+ * Copyright 2014-2015 Erik Zenker
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * alpaka is distributed in the hope that it will be useful,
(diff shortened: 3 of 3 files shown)


BenjaminW3:
You are exceeding the maximum log length of the CI ;-)
Please reduce the buffer size!

.travis.yml
```diff
@@ -497,8 +497,9 @@ script:
     #-------------------------------------------------------------------------------
 
     # Build and execute all examples.
-    - ./travis/compileExec.sh "example/helloWorld" ./helloWorld    
-    - ./travis/compileExec.sh "example/vectorAdd" ./vectorAdd
+    - ./travis/compileExec.sh "example/copy"       ./copy        
+    - ./travis/compileExec.sh "example/helloWorld" ./helloWorld
+    - ./travis/compileExec.sh "example/vectorAdd"  ./vectorAdd
```
BenjaminW3: Could you please remove the alignment and just use the single space in between?
alpaka does not use alignment anywhere, only indendation.

erikzenker: Ohhh :joy:

erikzenker:
Ohh, you are right. I did not think of the CI :joy:

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+
+    /**
+     * Create host and device buffers
+     *
+     * A buffer is an n-dimensional structure with a
+     * particular data type and size which corresponds
+     * to memory on the desired device. Buffers can be
+     * allocated on the device or can be obtained from
+     * allready existing allocations e.g. std::array or
```
BenjaminW3: al~~**l**~~ready

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,282 @@
...
+     * std::vector.
```
BenjaminW3: ... or a simple call to new

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,282 @@
...
+
+    /**
+     * Copy host to device Buffer
+     *
+     * A copy operation of one buffer into
+     * another buffer is enqueued into a stream
+     * like it is done for kernel execution, but
+     * more automatically. Copy is only possible
+     * from host to host, host to device and
+     * device to host. Device to device copies
```
BenjaminW3: I think that this is not true.
I have never tested it, but you can copy from GPU to GPU. The GPU buffer holds the device handle to decide if it is a peer copy or a copy within a device.
Because there is only ever one CPU device shared by all CPU accelerators (at the moment), all CPU copie versions are the same.

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,282 @@
...
+     * are not possible. In this example the
+     * buffer with the plain pointer is first copied
+     * into the host buffer and then is the host
+     * buffer copied into the device buffer.
+     */
+    alpaka::mem::view::copy(stream, hostBuffer, hostBufferPlain, extents);
```
BenjaminW3: I can see a reason why all versions (native host buffer, native device buffer and plain pointer host buffer wrapper) should be showcased.
However, the detour to copy hostBufferPlain -> hostBuffer -> deviceBuffer instead of the direct copy hostBufferPlain -> deviceBuffer looks very artificial.

example/copy/src/copy.cpp
erikzenker: Maybe we should write somethink like : 
```
Some devices also support device to device copy, but this is not true in general.
```

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,282 @@
...
+    alpaka::stream::enqueue(stream, test);
+
+
+    /**
+     * Print device Buffer
+     *
+     * Because we really like to flood our
+     * terminal with numbers, the following
+     * kernel prints all numbers of the
+     * device buffer to stdout on the terminal.
```
BenjaminW3:
```
Because this possibly is a parallel operation, the output can appear in any order or even completely distorted.
```

example/copy/src/copy.cpp
BenjaminW3: However, currently all (both CPU and GPU) devices support it.

BenjaminW3: `As always within alpaka, you will get a compile time error if the desired copy coperation (e.g. between various accelerator devices) is not currently supported.`

example/copy/src/copy.cpp
erikzenker: My intention was to show that copies between host buffers are the same than copies from host to device buffers.  It is really very artificial, but I would not remove it, because otherwise the plain pointer host buffer would be absolute useless. Do you have an Idea what to do with the plain pointer buffer ?

BenjaminW3: Maybe we should just create a second device buffer and copy both buffers to the device and print both individually to show that they work. In the current version, the values within the hostBuffer are discarded.

erikzenker: Okay, I added a second device buffer and added a particular kernel for value initialization. So we have some variation in the example.

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,352 @@
...
+    PrintBufferKernel printBufferKernel;
+    auto const print1 (alpaka::exec::create<Acc> (workdiv,
+                                                printBufferKernel,
+                                                alpaka::mem::view::getPtrNative(deviceBuffer1), // 1st kernel argument
+                                                extents));                                     // 2nd kernel argument
+
+    auto const print2 (alpaka::exec::create<Acc> (workdiv,
+                                                  printBufferKernel,
+                                                  alpaka::mem::view::getPtrNative(deviceBuffer2), // 1st kernel argument
+                                                  extents));                                     // 2nd kernel argument
```
BenjaminW3: The alignment got destroyed here and above.

example/copy/src/copy.cpp
```diff
@@ -0,0 +1,352 @@
...
+
+
+    /**
+     * Print device Buffer
+     *
+     * Because we really like to flood our
+     * terminal with numbers, the following
+     * kernel prints all numbers of the
+     * device buffer to stdout on the terminal.
+     * Since, this possibly is a parallel operation,
```
BenjaminW3: I do not think there has to be a comma after since.

BenjaminW3:
I am unsure about the name. Do you think `copy` is descriptive enough? Maybe `bufferCopy` or `bufCopy` is better.

example/copy/src/copy.cpp
erikzenker: Well spotted, thx.

example/copy/src/copy.cpp
erikzenker: I am never sure about that, thx for spotting.

erikzenker:
lets call it bufferCopy :-)

BenjaminW3:
When you have squashed all commits together, I will merge it.

erikzenker:
Rebase finished, waiting for CI :hourglass:

PR #99, opened 2015-11-17
clean up tests

BenjaminW3:
This performs some minor clean ups to the CMake and test configurations:

- adds some currently missing files to the Visual Studio Project
- renames the test projects to use camelCase
- moves the test helpers from `test/common/alpaka/integ/` to `test/common/alpaka/test` to make them accessible by future unit tests.
- accordingly moved those functions to `alpaka::test` namespace
- adds a `_PROJECT_NAME` variable to the test CMakeLists to reduce the number of places where the name has to be changed


test/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
 
 SET_PROPERTY(GLOBAL PROPERTY USE_FOLDERS ON)
 
-PROJECT("alpaka_test")
+PROJECT("alpakaTest")
```
psychocoderHPC: Could we rename it ti `alpaka-test`?

test/integ/CMakeLists.txt
```diff
@@ -26,7 +26,7 @@ CMAKE_MINIMUM_REQUIRED(VERSION 2.8.12)
 
 SET_PROPERTY(GLOBAL PROPERTY USE_FOLDERS ON)
 
-PROJECT("alpaka_integ_test")
+PROJECT("alpakaIntegTest")
```
psychocoderHPC: Might be `alpaka-integTest`

erikzenker:
Should the examples also be renamed to camel case projects names ?

BenjaminW3:
Currently all of the examples and tests already have a camelCase naming scheme. Therefore I tried to be consistent. However, we can always talk about a switch to PascalCase or other styles.

PR #100, opened 2015-11-17
add check for self contained headers

BenjaminW3:
This adds an integration test that checks that all the headers in the project are self contained.

To do this the test creates a cpp file per header file.
It includes the corresponding header twice:
- The first inclusion triggers the missing inclusion of dependency headers as well as missing guards (e.g. if CUDA is available).
- The second inclusion finds missing include guards (`#pragma once`).

Even though alpaka provides a main `alpaka.hpp` file, you are allowed to include individual files manually.
However, currently only the main header file guards the inclusion of the other files with `#ifdef`'s.
Therefore, those guards have to be moved to the individual headers.


bussmann:
100! Yeah!

psychocoderHPC:
Nice feature!

erikzenker:
I like it

BenjaminW3:
Now it seems to work correctly in all configurations.
However, this test adds 2-3 minutes to the CPU builds and up to 15 minutes for the nvcc builds!
Should I enable this test only for some jobs? Should I add a special job that only builds this test and nothing else for both, CPU and GPU? Both solutions would reduce the number of combinations that are tested.

bussmann:
Only fifteen minutes? That's still too short, make compilation times longer!

BenjaminW3:
Now the extended analysis (headerCheck, sloc, search for FIXME/HACK/TODO) is only executed for two jobs (one for CUDA, one without) early in the job list so that the 25 minute nvcc build is not the last one running but there are always 4 other jobs executed in parallel.
So this can be merged now!

bussmann:
I take you didn't read my comment.

bussmann:
Btw: Great work!

erikzenker:
Good Job, review says ready to merge !

BenjaminW3:
ping

psychocoderHPC:
pong

include/alpaka/dev/cpu/Wait.hpp
```diff
@@ -0,0 +1,78 @@
...
+                    // Now wait for all the events.
+                    for(auto && event : vEvents)
+                    {
+                        wait::wait(event);
+                    }
+                }
+            };
+        }
+    }
+}
```
psychocoderHPC: please add an empty line

include/alpaka/dev/DevCpu.hpp
```diff
@@ -462,51 +463,4 @@ namespace alpaka
             }
         }
     }
-    namespace wait
```
psychocoderHPC: **Only a note:** moving this code to a new file and add the new feature that all include are verified make the review hard. Two pull request helps.

BenjaminW3:
The newline has been added.
The new file was necessary due to cyclic dependencies and is necessary for self-contained headers.

PR #101, opened 2015-11-18
Add the missing helloWorld example to the CMake alpakaExamples project

BenjaminW3:
The helloWorld example was missing from the project that includes all examples.
By the way, the USE_FOLDERS property is useless in this CMakeLists.txt and the project name is not in the alpaka default camelCase.


erikzenker:
I think that I have to do the same for the copy example in #98

erikzenker:
I think, that this can be merged, but I am no maintainer. @psychocoderHPC could you be so kind and merge this PR ?

PR #104, opened 2015-11-22
convert vectorAdd integ test to axpy integ test

BenjaminW3:
Currently there is a `vectorAdd` example and a `vectorAdd` integration test.
It is not possible to include both targets within a single project due to the name ambiguity. To solve this, I have converted the integration test to calculate `axpy` instead of a simple vector addition.

### DIFF
--- a/test/integ/axpy/src/main.cpp
+++ b/test/integ/axpy/src/main.cpp
@@ -31,18 +31,19 @@
 //! A vector addition kernel.
 //! \tparam TAcc The accelerator environment to be executed on.
 //#############################################################################
-class VectorAddKernel
+class AxpyKernel
 {
 public:
     //-----------------------------------------------------------------------------
-    //! The kernel entry point.
+    //! Vector addition Y = alpha * X + Y.
     //!
+    //! \tparam TAcc The type of the accelerator the kernel is executed on..
     //! \tparam TElem The matrix element type.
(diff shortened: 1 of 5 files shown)


test/integ/axpy/src/main.cpp
```diff
@@ -97,7 +98,7 @@ struct VectorAddKernelTester
         using Val = float;
 
         // Create the kernel function object.
-        VectorAddKernel kernel;
+        AxpyKernel kernel;
 
         // Get the host device.
         auto devHost(alpaka::dev::cpu::getDev());
```
erikzenker: Should we use getDevByIdx(0) here ? I know that it is the same, but it is more explicit (we did the same in the examples).

BenjaminW3: I can do this. While I am at it, should I directly remove the `cpu::getDev()` method because it only forwards to `getDevByIdx(0u)`?

erikzenker: I would suggest it, because when I was writing a alpaka code for the first time it was somehow confusing why the cpu device is treated in a special way. Same behavior vor all accelerators is less confusing I think. Furthermore, two methods for the same functionality raises the question which one to use. Is one better then the other ...

erikzenker:
So I will merge this PR an we will work on the device retrieval separately

PR #107, opened 2015-11-24
remove cpu::getDev()

### LINKED ISSUE #105: Remove cpu::dev and replace by getDevByIdx()
See discussion in #104

BenjaminW3:
This closes #105.

### DIFF
--- a/example/vectorAdd/src/main.cpp
+++ b/example/vectorAdd/src/main.cpp
@@ -100,7 +100,7 @@ auto main()
         VectorAddKernel kernel;
 
         // Get the host device.
-        auto devHost(alpaka::dev::cpu::getDev());
+        auto devHost(alpaka::dev::DevManCpu::getDevByIdx(0u));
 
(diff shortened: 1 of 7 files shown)


BenjaminW3:
Correctly compiles now ;-)

erikzenker:
Well done, I would merge this PR. But just one thing ...is there a special reasion why there is DevManCpu used instead of DevMan<Host> ?

BenjaminW3:
There is a device manager for each device type. So there is one for CPU devices and one for GPU devices. The CPU device manager currently has only one single device.
We could also have defined a `DevHost` as `DevCpu` and then retrieved the device manager of the CPU device by calling `DevMan<DevHost>` as you have done in your examples but I think this detour is not always necessary.

erikzenker:
I just want to mention that when `DevMan<DevCpu>` is equal to `DevManCpu` than a user could wonder which one to use.

Issue #78, opened 2015-10-22
alpaka kernel without templates


psychocoderHPC:
It is not possible to use a alpaka kernel without templates for the `operator()` in a cuda devices.

```C++
struct Foo {
    ALPAKA_FN_ACC
    void operator()(const alpakacu::Acc& acc, const float *A, const float *B, float *C, int numElements) const
    {
        auto gridDimension = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Blocks>(acc);
        ...
    }
 };
```

Compiling for a CPU device works fine.
Compiling with nvcc for cuda trigger an error:
```C++
/bigdata/hplsim/scratch/widera/alpaka/include/alpaka/workdiv/Traits.hpp(69): error: calling a __device__ function("getWorkDiv") from a __host__ __device__ function("getWorkDiv") is not allowed
          detected during:
            instantiation of "auto alpaka::workdiv::getWorkDiv<TOrigin,TUnit,TWorkDiv>(const TWorkDiv &)->alpaka::Vec<alpaka::dim::Dim<TWorkDiv>, alpaka::size::Size<TWorkDiv>> [with TOrigin=alpaka::origin::Block, TUnit=alpaka::unit::Threads, TWorkDiv=alpaka::workdiv::WorkDivCudaBuiltIn<alpaka::dim::DimInt<3UL>, alpakacu::IdxSize>]" 
(134): here
            instantiation of "auto alpaka::workdiv::traits::GetWorkDiv<TWorkDiv, alpaka::origin::Block, alpaka::unit::Threads, std::enable_if<<expression>, void>::type>::getWorkDiv(const TWorkDiv &)->alpaka::Vec<alpaka::dim::Dim<TWorkDiv::WorkDivBase>, alpaka::size::Size<TWorkDiv::WorkDivBase>> [with TWorkDiv=alpakacu::Acc]" 
(69): here
            instantiation of "auto alpaka::workdiv::getWorkDiv<TOrigin,TUnit,TWorkDiv>(const TWorkDiv &)->alpaka::Vec<alpaka::dim::Dim<TWorkDiv>, alpaka::size::Size<TWorkDiv>> [with TOrigin=alpaka::origin::Block, TUnit=alpaka::unit::Threads, TWorkDiv=alpakacu::Acc]" 
/bigdata/hplsim/scratch/widera/cudaExample/./src//main.cpp(19): here
```

The reason for the error is:
[...]

Issue #73, opened 2015-10-16
`ALPAKA_FN_ACC ` compatibility with CUDA only libraries


psychocoderHPC:
@BenjaminW3 is there a reason why `ALPAKA_FN_ACC` is defined as `__device__ __host__ __forceinline__`  [see here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/core/Common.hpp#L42)

The problem is that this definition breaks the compatibility to all non alpaka gpu libraries. 
If inside a alpaka function/kernel  which is defined as `ALPAKA_FN_ACC` a non alpaka library call is used (which is defined with `__device__`) always a compiler **error** `calling a device function from host/device` is triggered.

psychocoderHPC:
A solotion can be to change the define to:
`#define ALPAKA_FN_ACC __device__  __forceinline__`

BenjaminW3:
When the `__host__` attribute is removed, it would not be possible to have a kernel that is simultaneously used with CUDA and on the CPU.

psychocoderHPC:
@BenjaminW3  You are right, this was the reason I forgot. 

The question now is, how we can solve the compatibility to extern libraries? Else we can not write specialized fast versions of a kernel for a specific hardware.

BenjaminW3:
That is a really good question. I do not see a way to directly use such CUDA only libraries.

It might be possible to define a wrapper for each such library. This wrapper uses `ALPAKA_FN_ACC` and internally dispatches to the CUDA only version within a `__CUDA_ARCH__` ifdef. The host version does not have to be implemented as long as it is only used in a non heterogeneous CUDA only simulation.

psychocoderHPC:
Why you are closing this issue? I think alpaka is in the current state not usable with other libraries cuda libraries.

psychocoderHPC:
A solition for this problen can be to add an flag to alpaka called `compatibility mode` which set `ALPAKA_FN_ACC` for coda only to `__device__`

BenjaminW3:
I have started implementing an `ALPAKA_ACC_GPU_CUDA_ONLY_MODE`. The solution depends on #100 because in such a mode where `ALPAKA_FN_ACC` is defined to `__device__` only, the other accelerators can not be built at all (because they would use OpenMP and other things within a then `__device__` only function).

psychocoderHPC:
[...]

Issue #47, opened 2015-07-30
enable usage of lambdas as kernels


BenjaminW3:
CUDA 7.5 allows to pass lambdas into the global kernel with the flag `--expt-extended-lambda`. See [here](http://on-demand.gputechconf.com/gtc/2015/webinar/gtc-express-everything-you-need-to-know-cuda-7-5-webinar.pdf).

``` C++
__device__ int result;
template <typename T>
__global__ void kernel(T in) { result = in(4); }

void foo(void) {
    int x = 30;
    kernel<<<1,1>>>(
        [=] __device__ (int in) {
            return in * x; // x captured by value
        })
}
```

We have to make sure that ALPAKA_FN_ACC can be used in the place of `__device__` especially in non-cuda code.

BenjaminW3:
CUDA 8.0 supports Heterogeneous C++ Lambda. We could add a test now.

erikzenker:
Nice ! I was looking forward to this.

bussmann:
Awesome!

erikzenker:
Let me prepare an example and test for it.

PR #108, opened 2015-11-25
Example that describes lambdas and std::function as kernel functions

erikzenker:
I did not know that lambdas and std::function can be used within alpaka
until I read your thesis more precisely. So I tried it and it worked !
Finally I wrote this example which is a small modification of the
Hello World example.

### DIFF
--- a/example/helloWorldLambda/src/helloWorldLambda.cpp
+++ b/example/helloWorldLambda/src/helloWorldLambda.cpp
@@ -1,0 +1,175 @@
+/**
+ * \file
+ * Copyright 2014-2015 Erik Zenker
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * alpaka is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with alpaka.
+ * If not, see <http://www.gnu.org/licenses/>.
+ *
+ */
+
+// Alpaka
+#include <alpaka/alpaka.hpp>
+
+// STL
+#include <iostream>
+#include <functional>
+
+/**
+ * This functions says hi to the world and
+ * can be encapsulated into a std::function
+ * and used as a kernel function. It is 
+ * just another way to define alpaka kernels
+ * and might be useful when it is necessary
+ * to lift an existing function into a kernel
+ * function.
+ */
+template<typename Acc>
+void hiWorldFunction(Acc& acc, size_t const nExclamationMarks){
+    auto globalThreadIdx    = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+    auto globalThreadExtent = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+    auto linearizedGlobalThreadIdx = alpaka::core::mapIdx<1u>(globalThreadIdx,
+                                                              globalThreadExtent);
+                                                          
+    printf("[z:%u, y:%u, x:%u][linear:%u] Hi world from a std::function",
+           static_cast<unsigned>(globalThreadIdx[0]),
+           static_cast<unsigned>(globalThreadIdx[1]),
+           static_cast<unsigned>(globalThreadIdx[2]),
+           static_cast<unsigned>(linearizedGlobalThreadIdx[0]));
+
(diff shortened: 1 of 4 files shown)


BenjaminW3:
Cool that this is really working ;-)

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,175 @@
...
+
+                                                          for(size_t i = 0; i < nExclamationMarks; ++i){
+                                                              printf("!");
+                                                          }
+                                                          
+                                                          printf("\n");
+
+                                                      },
+                                                      nExclamationMarks // 1st real kernel argument, but 2nd argument in the lambda function,
+                                                                        // since, the first argument as allways the accelerator!
```
BenjaminW3: ~~as~~ -> is

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,175 @@
...
+                                                      ));
+
+    alpaka::stream::enqueue(stream, helloWorld);
+
+    
+    /**
+     * Run kernel with std::function
+     *
+     * This kernel says hi to world by using
```
BenjaminW3: to **the** world

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,175 @@
...
+     * Alpaka accepts these std::functions 
+     * as kernel functions. Therefore, it is easy
+     * to wrap allready existing code into a
+     * std::function and provide it to the alpaka 
+     * library.
+     */
+    auto const hiWorld (alpaka::exec::create<Acc> (workdiv,
+                                                   std::function<void(Acc&, size_t)>( hiWorldFunction<Acc> ),
+                                                   nExclamationMarks)); // 1st real kernel argument, but 2nd argument in the wrapped function
+                                                                        // since, the first argument as allways the accelerator!
```
BenjaminW3: ~~as~~ -> is

BenjaminW3:
When you have corrected the typos and squashed the commits this is ready to merge.

BenjaminW3:
Oh no, wait. The CI tells me that CUDA 7.0 does not correctly support C++11 lambda functions even in Host code...
Furthermore, the test is only using the `DevCpu` which means that it does not show that it is also woking in CUDA. I had already written a possible solution in #47 where `ALPAKA_FN_ACC` has to be added after the lambda capture list as well as a specific nvcc compiler option. Because the host only definition of `ALPAKA_FN_ACC` is currently not empty, this could possibly require further changes.

erikzenker:
Mmh I am not able to compile when the cuda acc is enabled at all, also not the other examples

erikzenker:
"Falscher alarm" just had a fix on the cluster so piconalpaka can compile and run

    -    #define ALPAKA_FN_ACC __device__ __host__ __forceinline__
    +    #define ALPAKA_FN_ACC __device__ __forceinline__

erikzenker:
I used the following lambda : 

    auto foo = [] __device__ (Acc& acc) {
        auto globalThreadIdx    = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
    };
    
This lambda function is just written in the code and not used, but produces the following error
during compilation. So we also have a problem with host and device functions of cuda.

`error: calling a __device__ function("getIdx") from a __host__ __device__ function("getIdx") is not allowed`

erikzenker:
Why does this 
   global -> host device -> device

call does work for non lambda kernels ?

erikzenker:
My last comment is related to #78  ...lambdas are also functions without further templates

erikzenker:
This PR is also related to #73.

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,175 @@
...
+     * lambda function. This kind of kernel function
+     * declaration might be useful when small kernels
+     * are written for testing or lambda functions
+     * allready exist.
+     *
+     */
+    const size_t nExclamationMarks = 10;
+    
+    auto const helloWorld (alpaka::exec::create<Acc> (workdiv,
+                                                      [](Acc& acc, size_t const nExclamationMarks) -> void {
```
psychocoderHPC: Please use indention instead of alignment

erikzenker:
After a further discussion with @psychocoderHPC we agree that it is not possible to use lambdas in alpaka in combination with the cuda accelerator. It might be possible if we change the ALPAKA_FN... defintion
but this leads to the problem that we can not use a cpu acc at the same time. I can live without lambdas, but it would be a nice to have.

BenjaminW3:
We could maybe use C++14 generic lambdas with auto as parameter type.

```C++
auto l = [] (auto const &  TAcc) { ...; };
```

So that l is an instance of the type:

```C++
class /* unnamed */
{
public:
    template<
        typename TAcc>
    auto operator() (auto const & TAcc) const
    -> void
    { ...; }
};
```

erikzenker:
In world were nvcc accepts C++14 lambdas this could be a solution (but we do not know if this does the trick). From my point of view we can close this PR.

Issue #111, opened 2015-12-07
vectorAdd with CUDA: Assertion failed


tu-maurice:
Due to a project assigment by Mr. Bussmann, my collegue and I need to use the Alpaka library on the Taurus of the TUD. Today I used the example program vectorAdd in the develop as well as the master branch to test if I can get Alpaka to run on the Taurus.

Unfortunately I get an Assertion Error whenever either version uses the GPU.
```
// develop branch. Should be similar to the one from the master branch (see attachments)
vectorAdd: /home/s7745593/alpaka/include/alpaka/exec/ExecGpuCudaRt.hpp:471: static void 
alpaka::stream::traits::Enqueue<alpaka::stream::StreamCudaRtSync, 
alpaka::exec::ExecGpuCudaRt<TDim, TSize, TKernelFnObj, TArgs ...>, 
void>::enqueue(alpaka::stream::StreamCudaRtSync&, const alpaka::exec::ExecGpuCudaRt<TDim, 
TSize, TKernelFnObj, TArgs ...>&) [with TDim = std::integral_constant<long unsigned int, 1ul>; TSize = 
long unsigned int; TKernelFnObj = VectorAddKernel; TArgs = {float*, float*, float*, long unsigned int}]: 
Assertion `threadElemExtent[TDim::value-1u-i] == 1' failed.
```
I'm not sure if the assertion failed because of an error in the example or if the assertion is wrong.

To run the vectorAdd example of the develop branch on the GPU, I applied some changes to the main.cpp (a diff is attached).

The environment we used to run this example:
```
Compiler    : gcc 4.9.1 (No gcc 4.9.2 available, might also be the cause of the problem)
Boost       : 1.56
Boost Fiber : N/A
CUDA        : 7.0.28
CMake       : 3.3
```

I tried to gather as much information as possible so here it is:
[develop-stderr.txt](https://github.com/ComputationalRadiationPhysics/alpaka/files/54224/develop-stderr.txt) - Error of vectorAdd in the develop branch
[develop-stdout.txt](https://github.com/ComputationalRadiationPhysics/alpaka/files/54227/develop-stdout.txt) - Output of vectorAdd in develop branch
[develop-main.cpp.txt](https://github.com/ComputationalRadiationPhysics/alpaka/files/54222/develop-main.cpp.txt) - "diff" between the repo version and "my" version of vectorAdd/src/main.cpp to run the Kernel on the GPU
[...]

PR #112, opened 2015-12-08
Moved assert for elems layer into correct loop

erikzenker:
The GPU Executor asserted that the elements vector is set to 1 in all dimensions,
but I assume that this is not correct. Because, also a GPU should be able to 
process multiple elements.

I moved the assert into the loop that checks if necessary dimensions are set to 1.

This PR is related to issue #111


BenjaminW3:
Thanks for fixing this!

PR #114, opened 2015-12-10
fix wrong guards

psychocoderHPC:
rename guard `ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED` to
`ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED` for `AccCpuOmp2Blocks`

This bug was introduced with the issue #100


BenjaminW3:
Could find!

PR #109, opened 2015-11-27
correct view naming and usage

BenjaminW3:
This cleans up some of the parts that are mentioned in #106.

* `buf\BufPlainPtrWrapper` -> `view\ViewPlainPtr` 
* `buf\BufStdContainers` -> `view\ViewStdContainers` 
* `view\ViewBasic` -> `view\ViewSubView`
* renaming many occurences of `TBuf`, `buf`, ... to `TView`, `view`, ...
* remove the currently useless `createView` method
* Adds a `CMakeLists.txt` that builds all examples and tests at once

PR #115, opened 2015-12-13
fix wrong view namespace

BenjaminW3:
In the course of moving `ViewPlainPtr` in #109 it was missed to change the namespace.
Furthermore, views can not be required to provide an underlying buffer because only `ViewSubView` has one but all others do not. Therefore `mem::view::GetBuf` has been removed.


erikzenker:
Good point, I support the namespace moving. I think there is a lot source code using the old namespace. Do we want to change their namespace in an extra pull request ?

BenjaminW3:
Which code do you mean? I have already adapted all the examples and tests (haven`t been many). Due to all examples and tests being CI checked, there should not be anything within the repository missing.

erikzenker:
You are right, I missed one file in the changed files log. This can be merged.

PR #117, opened 2015-12-13
work around boost 1.57 MSVC bug

BenjaminW3:
`boost::lockfree::queue` is buggy in versions lower then 1.58 on MSVC.


erikzenker:
We should definitely raise the boost version in future to get rid of such workarounds, but this is okay for the moment. Thank you very much !

PR #118, opened 2015-12-13
remove previously generated sources in headerCheck test

BenjaminW3:
When iteratively working on the same machine and alpaka includes are moved or removed, the `headerCheck` test source files currently reside within the folders and therefore within the projects but are referencing invalid headers. Therefore, they will now be removed on each cmake execution.

`headerCheck` does not require any specific boost version, therefore the version requirement is removed.


erikzenker:
Well done :-)

PR #121, opened 2015-12-17
provide blockThread and threadElem extents independently to getBlockSharedExternMemSizeBytes

### LINKED ISSUE #110: Add elements level to extern shared memory calculation
The interface to calculate the amount of extern shared memory only provides the blockExtent:

                ALPAKA_NO_HOST_ACC_WARNING
                       template<
                        typename TDim,
                        typename... TArgs>
                    ALPAKA_FN_HOST_ACC static auto getBlockSharedExternMemSizeBytes(
                        Vec<TDim, size::Size<TAcc>> const & blockElemExtent,
                        TArgs const & ... args)
                    -> size::Size<TAcc>
                    {
                        boost::ignore_unused(blockElemExtent);
                        boost::ignore_unused(args...);
    
                            return 0;
                    }

But this shared memory could also be depdendent on the element level.

BenjaminW3:
This fixes #110 and additionally removes the never used `SupportsVectorization` trait that had been forgotten to remove.

### DIFF
--- a/include/alpaka/exec/ExecCpuFibers.hpp
+++ b/include/alpaka/exec/ExecCpuFibers.hpp
@@ -146,8 +146,8 @@ namespace alpaka
                     workdiv::getWorkDiv<Grid, Blocks>(*this));
                 auto const blockThreadExtent(
                     workdiv::getWorkDiv<Block, Threads>(*this));
-                auto const blockElemExtent(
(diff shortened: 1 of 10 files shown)


psychocoderHPC:
Well done and big thanks for this pull request.

PR #122, opened 2015-12-18
Remove dependency on Boost.MPL

BenjaminW3:
While trying to extend the testing infrastructure I found the need to test all accelerators (already done in the integration tests) but instantiated with many combinations of `TSize` and `TDim` types. (see `test/unit/acc/src/AccNameTest.cpp`)
Currently there are 7 Accelerators, which I wanted to test with 4 dimensionalities and 9 size types.
`7 * 4 * 9 = 252` combinations (cartesian product)
However, `Boost.MPL` is very limited in its capabilities and can by default only support up to 50 types in a list. Due to it's preprocessing and header generation nature one can manually generate headers supporting more elements but this is no solution that can be applied on user systems.
Therefore, I replaced the usage of `Boost.MPL` with C++11 constructs. This decreases the compilation time, supports (theoretically) unlimited type list lengths and is easier to write and understand.
The current template meta programming constructs have been moved to a `meta` folder. Additionally tests have been added for most of the basic meta-programming functions.


erikzenker:
Woah, what a big commit. Its too much to have a look all in detail. There was nothing, I have seen and CI says okay.

erikzenker:
Just a question:
Is it really necessary to create each combination in advance ? Couldn't you do it in several nested for loops ?

BenjaminW3:
Yes, it would be possible to do the iteration myself within the test function as you have done in libPMacc. However, it is nicer this way because the test framework will tell me exactly which combination of flags failed because each combination is tested in isolation. So I do not have to decode cryptic compiler errors first.

BenjaminW3:
**NOTE**: You have merged this even though it was marked as WIP - Work In Progress. It is no Problem in this case and you additionally assured the CI is green. I just wanted to wait for your answer regarding the nvcc `BOOST_AUTO_TEST_CASE_TEMPLATE` bug :smile:

erikzenker:
Ahh totally ignored the WIP flag

erikzenker:
Lets keep the nvcc bug in mind

PR #124, opened 2015-12-20
fix boost fibers file renaming

BenjaminW3:
Within the Boost.Fiber repository the `condition.hpp` header has been renamed to `condition_variable.hpp`.
Furthermore the library requirements have been relaxed to C++11 instead of C++14.


erikzenker:
Thank you !

PR #126, opened 2015-12-21
Remove exception handling from vectorAdd example

BenjaminW3:
I have removed the exception handling and the initial console printing from the vectorAdd example for simplification.


erikzenker:
thx for fixing the examples :smiley:

PR #127, opened 2015-12-28
add travis_retry to reduce the number of network related errors

BenjaminW3:
Sometimes network related commands like `wget` or `apt-get install` fail spuriously and require a manual restart of the job. By using `travis_retry` this could possibly be reduced.


erikzenker:
Good to know that something like this exists

PR #128, opened 2016-01-03
Refactor block shared dynamic memory

BenjaminW3:
The allocation of block shared dynamic memory is identical for all current CPU accelerators but the implementing code is duplicated. Furthermore, it is the last accelerator method that requires the old and hard to use syntax `acc.template getExternSharedMem<Type>()`.

This Pull Request conceptifies the dynamic memory allocation and unifies the implementations.
Now it can be accessed via `alpaka::block::shared::dyn::getMem<Type>(acc)` in the same way all other accelerator methods work.
The static versions are moved from `alpaka::block::shared` to `alpaka::block::shared::st`.
To match the CUDA naming the `BlockSharedExternMemSizeBytes` trait has been renamed to `BlockSharedMemDynSizeBytes`.


erikzenker:
Good work !

Issue #129, opened 2016-01-05
Example VectorAdd :unsigned leads to wrong calculation


vincentridder:
https://github.com/ComputationalRadiationPhysics/alpaka/blob/master/examples/vectorAdd/src/main.cpp#L74
is always zero.
so the last thread will run out of bounds.

tu-maurice:
Correction: The min() always returns zero because both parameters are unsigned.

BenjaminW3:
I do not know how why this `static_cast` was necessary but you are right, it is wrong.
The code where this example was copied from does not have these `static_cast`.

The `static_cast` of a negative integral value to an unsigned integral type leads to a very large positive value. The minimum of this large value and zero is always zero.
The last thread will then try to calculate `(n/threads)-0` elements which is possibly out of bounds.

tu-maurice:
@BenjaminW3 thank you.

PR #130, opened 2016-01-05
fix wrong extent calculation for last axpy thread

BenjaminW3:
Details of the bug are described in #129.


BenjaminW3:
The buggy `static_cast` had been introduced because the argument `0u` to `min` and the other value possibly have a different type leading to an ambiguous function call.
The new version does not use the `min` function anymore and uses a much easier calculation.

PR #131, opened 2016-01-11
clean-ups

BenjaminW3:
* introduce `test/common` project (folder and sources already existed) that shows up in IDEs and simplifies the other test CMake scripts
* pthread has to be linked unconditionally for the CPU Stream implementations
* `Boost.Fiber` does not require C++14 anymore but uses C++11 instead
* rename `_PROJECT_NAME` to `_TARGET_NAME`
* fix some warnings due to mixed signs in `meta::MakeIntegerSequenceOffset` by introducing a `meta::ConvertIntegerSequence` cast


erikzenker:
I had the look at your changes and found nothing to worry about :sun_with_face:

PR #132, opened 2016-01-12
extend map idx specializations to 4D

BenjaminW3:
This adds `MapIdx` specializations for 4D cases and removes type conversion warnings.

PR #125, opened 2015-12-21
Add tests for block shared memory

BenjaminW3:
This test application can be used to test if #85 is supported by all compilers, platforms and runtimes.

PR #133, opened 2016-01-12
fix gcc incorrect assembly bug

BenjaminW3:
gcc seems to be overly optimistic in release mode and misoptimizes some accesses to members of the `AccDevProps` struct for some TDim/TSize combinations.
The resulting assembly incorrectly tries to access the memory adress `0x0000000000000080`.
This bug is not triggered when the members are reordered.
As I was unable to build a minimal repro, there is no ticket for it.

### DIFF
--- a/include/alpaka/acc/AccDevProps.hpp
+++ b/include/alpaka/acc/AccDevProps.hpp
@@ -55,22 +55,25 @@ namespace alpaka
                 TSize const & blockThreadCountMax,
                 Vec<TDim, TSize> const & threadElemExtentMax,
                 TSize const & threadElemCountMax) :
-                    m_multiProcessorCount(multiProcessorCount),
                     m_gridBlockExtentMax(gridBlockExtentMax),
-                    m_gridBlockCountMax(gridBlockCountMax),
                     m_blockThreadExtentMax(blockThreadExtentMax),
-                    m_blockThreadCountMax(blockThreadCountMax),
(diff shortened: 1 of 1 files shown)


BenjaminW3:
push!

psychocoderHPC:
yes I check your commit tomorrow.

psychocoderHPC:
Which gcc version triggered the bug?

BenjaminW3:
This bug occurs with gcc 4.9.2 as well as with 5.2. However, this only occurs when it is used as stand-alone compiler (no nvcc). Furthermore, this is only triggered in Release mode and not in Debug mode.
It was discovered with the tests written in #125 (they additionally require #132). When this fix is not applied the error can be seen e.g. in [this build](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/101390796) in jobs 10 and 11.
`unknown location(0): fatal error: in "blockSharedMemSt/nonNull<N6alpaka3acc12AccCpuSerialISt17integral_constantImLm4EEiEE>": signal: SIGSEGV, si_code: 0 (memory access violation at address: 0x00000080)`
The error occurs in [those lines](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/workdiv/WorkDivHelpers.hpp#L222-L230).

psychocoderHPC:
thanks for the fix

PR #134, opened 2016-01-12
raise the boost minimum required version to 1.59

BenjaminW3:
This is necessary because Boost.Test is buggy when used with nvcc in earlier boost versions.


psychocoderHPC:
I will check this pull request.

**Note:** We should be more conservative with changes to new boost versions. The IBM XL compiler we should support in the near future is not the best for compiling new boost versions. This compiler will play a bigger role when openPower CPUs combined with NVIDIA GPUs enter the HPC market.

psychocoderHPC:
Could you please add a short note what buggy means. From you pull request it looks like a problem with the Microsoft C++ compiler and boost.

README.md
```diff
@@ -55,7 +55,7 @@ Supported Compilers
 
 This library uses C++11 (or newer when available).
 
-|Accelerator Back-end|gcc 4.9.2|gcc 5.2|clang 3.5/3.6|clang 3.7|MSVC 2015|
+|Accelerator Back-end|gcc 4.9.2|gcc 5.3|clang 3.5/3.6|clang 3.7|MSVC 2015|
```
psychocoderHPC: Is there a reason for the gcc change?

BenjaminW3: I have updated this, as the Travis CI Ubuntu now provides gcc 5.3. Therefore, gcc 5.2 is not tested anymore. This is no problem as it is only a bugfix gcc release.

BenjaminW3:
No, this is no MSVC issue. The current MSVC 2015 Update 1 does not even compile alpaka anymore. They seem to have broken some things.
Boost.Test uses some static variables that are not decorated with `__device__` and are therefore inaccessible. This has been fixed in boost 1.59.

Issue #136, opened 2016-01-13
Endless recusion in GetPitchBytes for ViewSubView


psychocoderHPC:
The trait specialization `GetPitchBytes` of `ViewSubView` results in an endless recursion.

The specialization calls the trait `getPitchBytes()` for the some input view which results in a call to the same method.

@BenjaminW3 Could you look over if I missed something

psychocoderHPC:
Possible solution is to call the trait with the Buffer which is contained in `ViewSubView`.

```C++
return
                            mem::view::getPitchBytes<TIdx::value>(
                                view.m_Buf);
```

BenjaminW3:
Yes, this solution seems to be correct. Please write a Pull Request.

erikzenker:
Fixed by #143

PR #140, opened 2016-01-13
Test for copy views

erikzenker:
This PR creates a test for copy data between views.
It will prove the correct usage of pitch and extents during copy operations.
It will also prove copy operations between views of different dimensionality.

- [ ] Merge after #136 
- [ ] Merge after #141

### DIFF
--- a/test/integ/CMakeLists.txt
+++ b/test/integ/CMakeLists.txt
@@ -31,6 +31,7 @@ PROJECT("alpakaIntegTest")
 ################################################################################
 
 ADD_SUBDIRECTORY("axpy/")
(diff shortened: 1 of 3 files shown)


erikzenker:
This PR still fails because of #136

erikzenker:
Fails also because of #141

erikzenker:
Copy a view seems to another buffer seems to compromise the original buffer:

    alpaka::mem::view::copy(stream, viewBuffer, destBuffer, viewExtents);

erikzenker:
A more complex unit test is given in #143

Issue #138, opened 2016-01-13
`copy` and `set` of views are boken


psychocoderHPC:
If someone copies a view of a three dimensional buffer to another one than the wrong extents/ piches are used for the operation.
The some problem is alive for the operation `set`.

e.g. The [cuda operation set](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/buf/cuda/Set.hpp#L440) is an good example 
The trait `extent::getHeight()` [here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/buf/cuda/Set.hpp#L461) results the height of the a view and not of the physical  memory behind the view. This results in a problem that the wrong height [here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/buf/cuda/Set.hpp#L480) is given to the cuda copy.
The some issue exists in the [cpu implementation](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/buf/cpu/Set.hpp#L99)

The reason for that issue is that we not differ between physical memory extents and view extents.

@erikzenker is writing a test case for that problem

BenjaminW3:
Is this only true for 3-dimensional views?

psychocoderHPC:
A possible solution is to use the trait `getPitchedBytes<>` and calculate of of the pitch the physical memory sizes. 

e.g.
```C++
auto phyHeight = view::getPitchedBytes<1>/view::getPitchedBytes<2>;
```
but issue #135 and #134 must solved before.

Nevertheless the usage of this magic is not good IMO a better way is to add a trait `getPhysicalExtent` which should defined as the extents of the physical memory in the deepest parent view (the buffer or `ViewPlainPtrView`)

@BenjaminW3 Could you please participate

BenjaminW3:
I will have a look over all of those issues.

psychocoderHPC:
There is no need that you fix all of this, we will involve to be sure that we not interpret something wrong or misunderstood something.

BenjaminW3:
`mem::view::getPitchBytes` always returns the pitch of the physical underlying buffer irrespective of the number of views stacked on top. If this is not the case there might be a bug (as #137 shows, there is one).

erikzenker:
Currently, I am working on a fix for the cpu.

erikzenker:
[...]

PR #144, opened 2016-01-18
Add pitch for y dimension in cpu copy

erikzenker:
Fixes the copy bug mentioned in #138 for the cpu accelerators.
The copy algorithm used wrong pitches to "jump" into the next z slice.
Y pitch was added as member and used for next z slice calculation.

### DIFF
--- a/include/alpaka/mem/buf/cpu/Copy.hpp
+++ b/include/alpaka/mem/buf/cpu/Copy.hpp
@@ -97,8 +97,10 @@ namespace alpaka
                                 m_dstDepth(static_cast<Size>(extent::getDepth(viewDst))),
                                 m_srcDepth(static_cast<Size>(extent::getDepth(viewSrc))),
 #endif
-                                m_dstPitchBytes(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewDst>::value - 1u>(viewDst))),
-                                m_srcPitchBytes(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewSrc>::value - 1u>(viewSrc))),
+                                m_dstPitchBytesX(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewDst>::value - 1u>(viewDst))),
+                                m_srcPitchBytesX(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewSrc>::value - 1u>(viewSrc))),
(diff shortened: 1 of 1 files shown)


erikzenker:
A unit test, verifying various kinds of copies, should follow.

BenjaminW3:
Thanks for fixing this!
I would prefer to have unit tests on a per concept implementation base. E.g., the `ViewSubViewTest` I have just added should be extended to test its copy and set implementations. All of the tests then should be done equally for `BufCpu`, `BufGpu` and `ViewPlainPtr`.

include/alpaka/mem/buf/cpu/Copy.hpp
```diff
@@ -97,8 +97,10 @@ namespace alpaka
                                 m_dstDepth(static_cast<Size>(extent::getDepth(viewDst))),
                                 m_srcDepth(static_cast<Size>(extent::getDepth(viewSrc))),
 #endif
-                                m_dstPitchBytes(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewDst>::value - 1u>(viewDst))),
-                                m_srcPitchBytes(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewSrc>::value - 1u>(viewSrc))),
+                                m_dstPitchBytesX(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewDst>::value - 1u>(viewDst))),
+                                m_srcPitchBytesX(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewSrc>::value - 1u>(viewSrc))),
+                                m_dstPitchBytesY(static_cast<Size>(mem::view::getPitchBytes<dim::Dim<TViewDst>::value - 2u>(viewDst))),
```
BenjaminW3: Currently you are failing here beause for a 1D Buffer this is asking for the Pitch of the dimension -1.
To fix the current bug, it should be enough to specialize a copy for the 1D case.

psychocoderHPC: dim::Dim<TViewDst>::value
Maybe we can write this to solve the 1D problem
```C++
constexpr dim = dim::Dim<TViewDst>::value;
m_dstPitchBytesY(static_cast<Size>(
    mem::view::getPitchBytes<
    dim::Dim<
            TViewDst
        >::value - ( 2u % dim )
    >(
        viewDst
    )
)
```

erikzenker:
Fixed :-)

BenjaminW3:
Could you make this a single commit, please?

erikzenker:
I did a rebase against the current develop branch

PR #145, opened 2016-01-18
Fix bug in set views

erikzenker:
Fixes the set bug mentioned in #138 for the cpu accelerator. Fix similar to #144 .

### DIFF
--- a/include/alpaka/mem/buf/cpu/Set.hpp
+++ b/include/alpaka/mem/buf/cpu/Set.hpp
@@ -1,6 +1,6 @@
             /**
 * \file
-* Copyright 2014-2015 Benjamin Worpitz
+* Copyright 2014-2015 Benjamin Worpitz, Erik Zenker
 *
 * This file is part of alpaka.
 *
@@ -56,6 +56,8 @@ namespace alpaka
                         typename TExtent>
                     struct TaskSet
                     {
+                        using Size = size::Size<TExtent>;
+                        
                         static_assert(
                             dim::Dim<TView>::value == dim::Dim<TExtent>::value,
                             "The destination view and the extent are required to have the same dimensionality!");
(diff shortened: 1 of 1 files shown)


BenjaminW3:
Very good that you have fixed the missing copy of the input arguments as well as the missing Y pitch.
However, you have removed the fast path for memsets where the whole buffer or only planes have the same size and can be handled at once. Even though this is currently buggy, this should not be removed. The correct solution should look nearly identical to what is done in `include/alpaka/mem/buf/cpu/Copy.hpp`.

include/alpaka/mem/buf/cpu/Set.hpp
```diff
@@ -67,10 +69,27 @@ namespace alpaka
...
+                                m_extentHeight(static_cast<Size>(extent::getHeight(extent))),
+                                m_extentDepth(static_cast<Size>(extent::getDepth(extent))),
+                                m_dstWidth(static_cast<Size>(extent::getWidth(view))),
+                                m_dstHeight(static_cast<Size>(extent::getHeight(view))),
+#if (!defined(NDEBUG)) || (ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL)
+                                m_dstDepth(static_cast<Size>(extent::getDepth(view))),
+#endif
+                                m_extentWidthBytes(m_extentWidth * sizeof(elem::Elem<TView>)),
+                                m_dstPitchBytesX(mem::view::getPitchBytes<dim::Dim<TView>::value - 1u>(view)),
+                                m_dstPitchBytesY(mem::view::getPitchBytes<dim::Dim<TView>::value - 2u>(view)),
```
BenjaminW3: Currently you are failing here beause for a 1D Buffer this is asking for the Pitch of the dimension -1.
To fix the current bug, it should be enough to specialize a copy for the 1D case.

erikzenker:
I will give it a try

erikzenker:
To be able to set the whole buffer a once I need to notice if the the extent of the view I want to set is equal to the origin buffer. Is there a way to retrieve the very "first" buffer of a view/buf?

BenjaminW3:
For now this solution is ok. Please, sqash the commits into one.

erikzenker:
Rebased

PR #148, opened 2016-01-26
fix invalid call of `vec::subVecEnd<>()`

psychocoderHPC:
`ViewPlainPtr` calls `vec::subVecEnd()` without explicit cast of the templated type to an alpaka vector

If a user defined vector for `pitchBytes` is used than no function signature for `vec::subVecEnd<TDim>` matches.
`ViewPlainPtr` allows any type for `pitchBytes` therefor a static cast is needed. After this pull request all user defined vector types with a well defined cast operator can be used.


BenjaminW3:
For now this is correct. The best solution would be to define some traits for vector data types that can be specialized for user types to avoid the full copy. I may have a look at this.

PR #150, opened 2016-01-27
Fix set gpu for 3D buffers/views

erikzenker:


include/alpaka/mem/buf/cuda/Set.hpp
```diff
@@ -470,14 +474,14 @@ namespace alpaka
...
+                            dstPitchBytesX,
+                            dstWidth * sizeof(Elem),
+                            dstPitchBytesY/dstPitchBytesX));
 
                     cudaExtent const cudaExtentVal(
                         make_cudaExtent(
-                            extentWidth,
-                            extentHeight,
+                            extentWidth * sizeof(Elem),
+                            extentHeight ,
```
BenjaminW3: What is the space doing there? :smile:

BenjaminW3:
Have you tested this with a 3 dimensional buffer? The integration tests currently only have 1D and 2D tests.

erikzenker:
I have some simple test cases for 1-3 dim copy and set, but i would like to improve them before I add them to alpaka.

erikzenker:
PS: I removed the space ^^

PR #152, opened 2016-01-28
Fixed copy paste failure for mapidx 4D case

erikzenker:
I think this bug occurred because of copy and paste from the 3D case.


BenjaminW3:
You are right, thanks!

PR #154, opened 2016-01-30
fix CUDA memcpy

BenjaminW3:
* fixes some possible copy-paste errors mixing dst and src values
* fixes some unused variable warnings
* fixes missing camelCase


erikzenker:
Thank you for fixin that !

PR #155, opened 2016-01-31
Added getExtents method

erikzenker:
This method returns the vector of extents in all dimensions. The method is necessary for unspecialized n-dimensional algorithms.


BenjaminW3:
This is already available as `alpaka::extent::getExtentVec` and `alpaka::extent::getExtentVecEnd`  to get only the given number of dimensions (although this is implemented in `Vec.hpp` to prevent cyclic dependencies)

erikzenker:
Okay, sorry I have not seen it.

PR #157, opened 2016-02-01
Iterator fix

erikzenker:
* Replace `getExtents` by `getExtentVec`
* Cast `Dim::value` to `Size` type
* Set correct alpaka fn prefix


BenjaminW3:
Is this ready? Why Work-In-Progress?

erikzenker:
Is ready now ...it was work-in-progress during the day

PR #161, opened 2016-02-02
Fixed construction of vec for typecast operator of non alpaka vectors

erikzenker:
When some non alpaka vector was casted to `alpaka::Vec` than the value constructor is used for the 1D case instead of the non alpaka vector typecast operator. 

I fixed that behaviour and added an unit test to prevent the error in the future.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -40,7 +40,7 @@
 #include <cstdint>                          // std::uint32_t
 #include <ostream>                          // std::ostream
 #include <cassert>                          // assert
-#include <type_traits>                      // std::enable_if
+#include <type_traits>                      // std::enable_if, std::decay
 #include <algorithm>                        // std::min, std::max, std::min_element, std::max_element
 
 // The nvcc compiler does not support the out of class version.
@@ -169,11 +169,8 @@ namespace alpaka
             typename = typename std::enable_if<
                 // There have to be dim arguments.
                 (sizeof...(TArgs)+1 == TDim::value)
-                && (
-                    // And there is either more than one argument ...
-                    (sizeof...(TArgs) > 0u)
-                    // ... or the first argument is not applicable for the copy constructor.
-                    || (!std::is_base_of<Vec<TDim, TSize>, typename std::remove_reference<TArg0>::type>::value))
+                &&
+                (std::is_same<TSize, typename std::decay<TArg0>::type>::value)
                 >::type>
         ALPAKA_FN_HOST_ACC Vec(
             TArg0 && arg0,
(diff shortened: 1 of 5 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -169,6 +169,8 @@ namespace alpaka
             typename = typename std::enable_if<
                 // There have to be dim arguments.
                 (sizeof...(TArgs)+1 == TDim::value)
+                &&
+                (std::is_same<TSize, TArg0>::value)
```
BenjaminW3: This will not work if `TArg0` is a reference type. You should use `std::is_same<TSize, typename std::decay<TArg0>::type>::value` instead.
Furthermore, I think it would be better to replace the current line 176 with this line because the new line is just a stricter version of the current line 176.

test/unit/vec/src/VecTest.cpp
```diff
@@ -0,0 +1,71 @@
...
+
+        for (TSize d = 0; d < TDim::value; d++)
+        {
+            result[TDim::value - 1 - d] = (*this)[d];
+        }    
+
+        return result;
+    }
+
+    TSize operator [](TSize idx) const
```
BenjaminW3: Could you please use a trailing return type (auto ... -> TSize)?

include/alpaka/vec/Vec.hpp
erikzenker: I think we can not replace the line 176 with the new line, because line 176 is in a disjunction with line 178 and line 178 would return true for non alpaka vectors. But we might remove line 176.

test/unit/vec/src/VecTest.cpp
erikzenker: sure

erikzenker:
Do you want to provide a more general fix in PR #163 ?

erikzenker:
I fixed your mentioned lines, but I think that your fix #163 is more general. We could merge this PR and #163 will be merge above.

BenjaminW3:
I have tried to implement a different solution in #163. However, nvcc does not compile this currently.

BenjaminW3:
I think you can completely remove the `std::is_base_of` line.

erikzenker:
Okay, I removed `std::is_base_of`

test/unit/vec/src/VecTest.cpp
```diff
@@ -0,0 +1,72 @@
...
+        TDim,
+        TSize>() const
+    {
+        using AlpakaVector = ::alpaka::Vec<
+            TDim,
+            TSize
+        >;
+        AlpakaVector result( AlpakaVector::zeros() );
+
+        for (TSize d = 0; d < TDim::value; d++)
```
BenjaminW3: If you remove the space after the `for`, make the `d++` a `++d` and sqash all commits together, I will merge it :smile:

erikzenker:
Style fix and rebase

PR #163, opened 2016-02-02
Only allow Vec creation with correct types

BenjaminW3:


BenjaminW3:
Closed for now because nvcc does not correctly compile this.

PR #165, opened 2016-02-03
enable clang native CUDA compilation

BenjaminW3:
This branch tests the capabilities of the clang built-in native CUDA compilation without the usage of the nvcc compiler wrapper.


BenjaminW3:
Things learned by now:
* Boost is not yet there: It does not yet know clang as a CUDA compiler and thinks it is nvcc and therefore disables variadic templates and uses other wrong configs
* clang for CUDA does not yet support cuRand
* The `std::exception_ptr` problem has to be researched

PR #166, opened 2016-02-04
allow object for `getBlockSharedMemDynSizeBytes`

psychocoderHPC:
Currently there is only the kernel type used to define external shared memory. 
If the user defined a kernel where the needed information for the external shared memory size is stored as member attribute than this can't mapped with the current interface.
The upper case can be done for example with a own `bind()` command where all functor parameters were stored as const members of a functor.
By passing the function kernel object to the shared memory trait this can be handled. 

Allow that external shared memory can be defined with the help of an function object instead only by a function type.

- change template signature for the trait `getBlockSharedMemDynSizeBytes()`
- add one  parameter to `traits::BlockSharedMemDynSizeBytes<>::getBlockSharedMemDynSizeBytes()`
- update examples and integration tests

### DIFF
--- a/include/alpaka/exec/ExecCpuOmp4.hpp
+++ b/include/alpaka/exec/ExecCpuOmp4.hpp
@@ -1,6 +1,6 @@
 /**
 * \file
-* Copyright 2014-2015 Benjamin Worpitz
+* Copyright 2014-2016 Benjamin Worpitz, Rene Widera
 *
 * This file is part of alpaka.
 *
@@ -130,8 +130,8 @@ namespace alpaka
                         {
                             return
                                 kernel::getBlockSharedMemDynSizeBytes<
-                                    TKernelFnObj,
                                     acc::AccCpuOmp4<TDim, TSize>>(
+                                        m_kernelFnObj,
                                         blockThreadExtent,
                                         threadElemExtent,
                                         args...);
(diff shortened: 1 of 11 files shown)


include/alpaka/kernel/Traits.hpp
```diff
@@ -90,11 +93,12 @@ namespace alpaka
         //-----------------------------------------------------------------------------
         ALPAKA_NO_HOST_ACC_WARNING
         template<
+             typename TAcc,
```
BenjaminW3: Here is one space too much :smile:

include/alpaka/exec/ExecCpuOmp4.hpp
```diff
@@ -188,7 +188,7 @@ namespace alpaka
                         {
                             Vec<dim::DimInt<1u>, TSize> const gridBlockIdx(b);
                             // When this is not repeated here:
-                            // error: �gridBlockExtent� referenced in target region does not have a mappable type
+                            // error: �gridBlockExtent� referenced in target region does not have a mappable type
```
BenjaminW3: What are those � here and in ExecCpuOmp2Threads.hpp?

BenjaminW3:
Could you finally squash both commits into a single one?

include/alpaka/exec/ExecCpuOmp4.hpp
psychocoderHPC: This file was utf8 encoded. I will fix the wrong symbols. A plain ASCII will is better, some compiler had problems with UTF8 (welcome to the 21th century :-)

psychocoderHPC:
Updated and quashed, please wait that all tests were finished.

Issue #48, opened 2015-08-11
conceptify the device managers


BenjaminW3:
-> they should not provide members

BenjaminW3:
Implemented in f85e233b321f0f7aa38aca18dca52e0dee41f4f2.

PR #167, opened 2016-02-06
refactor device managers

BenjaminW3:
This implements the long waiting issue #48 to unify the device enumeration process with all other alpaka methods.

* renames the device managers to platforms because they do not really manage anything, it is a shorter name and this is the wording used by OpenCL and SYCL
* moves the implementation of the DeviceManagers away from the devices themselves (`dev` -> `pltf`)
* makes acccess to the platforms uniform to all other types in alpaka by introducing traits (`DevManCpu::getDevByIdx(0)` -> `alpaka::pltf::getDevByIdx<pltf::PltfCpu>(0u)`)


psychocoderHPC:
I am on it, I need some time to think about the changes.

psychocoderHPC:
Thanks for this pull request.

PR #169, opened 2016-02-07
fix some icc incompatibilities

BenjaminW3:
While using icc 16.0.1 on Windows I have found some incompatibilities:
* the `ViewSubViewTest.cpp` is missing an include
* ICC emulates GCC on linux and MSVC on Windows and always provides the same `cpuid` function as the emulated compiler.


erikzenker:
Well spotted :-)

Issue #18, opened 2014-11-26
add special kernel methods


BenjaminW3:
- [x] ~~clock_t clock();
  long long int clock64();~~
- [ ] T __ldg(const T\* address);
  The read-only data cache load function is only supported by devices of compute capability 3.5 and higher.
- [x] __shfl_sync, __shfl_up_sync, __shfl_down_sync, __shfl_xor_sync
  Exchange a variable between threads within a warp.

felicepantaleo:
just checking if there is any plan on this

psychocoderHPC:
> just checking if there is any plan on this

You increased the priority for this features by +1

BenjaminW3:
Are you more interested in ldg or shfl?

psychocoderHPC:
The request in our internal Mattermost was for `__ballot_sync()` and `__shfl_up_sync()`. I also see  shuffle as more important as `__ldg()`

Note: CUDA has `__ballot` (asynchronous) and `__ballot_sync()`

j-stephan:
We will implement these intrinsics for alpaka 0.7.0.

* Implementing `__ldg` assigned to @j-stephan (as part of #38).
* Implementing `__shfl` assigned to @sbastrakov.

sbastrakov:
The shuffle part was implemented in #1273 (not by me, but it's now in).

j-stephan:
2/3 of the methods mentioned here have already been implemented. The remaining functionality will be tracked in #1312 (for easier project management).

PR #171, opened 2016-02-09
add accelerator clock method

BenjaminW3:
This adds the missing `clock64()` instruction returning a wall clock time in ticks which was noted in issue #18 as `alpaka::time::clock(acc)`.

### DIFF
--- a/include/alpaka/acc/AccCpuFibers.hpp
+++ b/include/alpaka/acc/AccCpuFibers.hpp
@@ -1,6 +1,6 @@
 /**
 * \file
-* Copyright 2014-2015 Benjamin Worpitz
+* Copyright 2014-2016 Benjamin Worpitz
 *
 * This file is part of alpaka.
 *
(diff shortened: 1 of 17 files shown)


test/unit/block/shared/src/BlockSharedMemDyn.cpp
```diff
@@ -37,7 +37,7 @@
 //#############################################################################
 //!
 //#############################################################################
-class BlockSharedMemDyn
+class BlockSharedMemDynTestKernel
```
erikzenker: Why the renaming in this PR?

erikzenker:
Nice Work :+1:  but you did also a lot of replacement of comments. Could you separate renamings and replacements which are unrelated to the time feature into a new pull request ?

test/unit/block/shared/src/BlockSharedMemDyn.cpp
BenjaminW3: Because I wrote a new test kernel, looked for the current naming convention and found that this was the only test kernel that was named inconsistently :smile:

BenjaminW3:
Renamings, comment corrections and the time feature are in 3 independent commits, so merging it at once would not introduce any revert problems. However, if you insist on it, I could seperate them when I have time again.

erikzenker:
It would be nice to split this PR up into a "new feature" and a "comment" PR. Because, if we search for some changes of code in the future it is easier to understand what happened in this PR.

BenjaminW3:
The two other commits have been proposed independently as #177 and #178.

BenjaminW3:
push!

erikzenker:
Thank you for separating this PR

PR #172, opened 2016-02-11
Use reference to device in ViewPlainPtr

erikzenker:
If no reference is used to provide the device to the ViewPlainPtr constructor, then
the device will destruct on ViewPlainPtr destruction. This error occured only
in CMake Release mode (CI was not able to find it, right?).

### DIFF
--- a/include/alpaka/mem/view/ViewPlainPtr.hpp
+++ b/include/alpaka/mem/view/ViewPlainPtr.hpp
@@ -74,7 +74,7 @@ namespace alpaka
                     typename TPitch>
                 ALPAKA_FN_HOST_ACC ViewPlainPtr(
                     TElem * pMem,
-                    TDev const dev,
+                    TDev const & dev,
                     TExtent const & extent,
                     TPitch const & pitchBytes) :
                         m_pMem(pMem),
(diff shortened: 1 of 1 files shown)


BenjaminW3:
CI says: 
```
unknown location(0): fatal error: in "acc/basicViewSubViewOperations<N6alpaka3acc12AccCpuSerialISt17integral_constantImLm1EEmEE>": memory access violation at address: 0x017a513d: invalid permissions
/home/travis/build/ComputationalRadiationPhysics/alpaka/test/unit/mem/view/src/ViewSubViewTest.cpp(229): last checkpoint
```

As it is now, the device is only a handle with shared ownership. Therfore it can be copied without a problem. Taking it by reference alone is only an optimization. A real problem might be that `ViewPlainPtr` does not hold the shared device itself, but only a reference.

erikzenker:
Okay, lets mark this PR as WIP. I try to setup a test case for the problem I tried to fix.

erikzenker:
Okay, the test case said that there is no problem ^^.
I will copy the gdb backtrace stack into this conversation to let you know what I mean.

PR #173, opened 2016-02-13
work around MSVC 2015 Update 1&2 subVecEnd ICE

BenjaminW3:
MSVC 2015 Update 1 introduced a bug into the compiler that leads to an ICE (Internal Compiler Error) when compiling the `vec::subVecEnd` method. My hope was that this would be fixed in Update 2. However, this was not the case.
I have now worked around this by refactoring and reworking the affected parts of the `Vec` class. Furthermore, the affected methods have been converted to traits and have been transfered to the `vec/Traits.hpp` header. The previous functionality should be fully preserved and a new unit test has been added.


erikzenker:
Thx for providing a unit test for the PR. I like it when a PR tests itself :smile:

PR #174, opened 2016-02-13
use BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION for decltype return types

BenjaminW3:
Defining the return type of forwarding functions is not only cumbersome but leads to harder error messages. Using `BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION` to only include those `decltype` return type declarations for non C++14 capable compilers has many advantages:
* The compile time decreases by a factor of 2 in each template instantiation.
* Error messages for users specializing the traits get much better. Currently a wrong specialization leads to a message that the forwarding function itself could not be instantiated. However, there is no note about the user template that was tried to compile as the compilation already failed within the return type. By failing within the function, now the correct error messages get printed out to the user.

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -24,6 +24,8 @@
 #include <boost/predef.h>   // workarounds
 #include <boost/version.hpp>// workarounds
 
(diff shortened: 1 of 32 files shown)


erikzenker:
Code revied finished ! Well done

PR #177, opened 2016-02-16
fix comments.

BenjaminW3:
As the title says...


erikzenker:
Refining comments is always fine

PR #178, opened 2016-02-16
unify test kernel names

BenjaminW3:
Not much to say...


bussmann:
I like the way you comment ; -)

Issue #183, opened 2016-02-27
Support for Library Builds


mxmlnkn:
[imresh](https://github.com/ComputationalRadiationPhysics/imresh) makes heavy use of interim static libraries to reduce compile times and also the finished product is supposed to be a [shared] library which can be linked against. There are some things which are problematic when using alpaka. I'm not finished porting yet, so there may be still more problems:

 - ~~`ALPAKA_FN_ACC` which e.g. replaces `__global__` when porting from CUDA adds a `__force_inline__` or `inline` which makes it impossible to compile a library, see commit 0564c8428bb9cfb2fb43b1d21770cee630464a1c~~ fixed in #194
 - to compile a library in the first place explicit instantiations of the accelerator type are necessary and add boiler plate could, which maybe could be prevented with some kind of macro provided by alpaka, see [here](https://github.com/ComputationalRadiationPhysics/imresh/commit/fa703de79a3501d8c119bc23976be61561405799#diff-3bf5faeb9a2e700f0ac9de222a9fbeb4R404) and [here](https://github.com/ComputationalRadiationPhysics/imresh/blob/fa703de79a3501d8c119bc23976be61561405799/src/imresh/libs/alpaka_T_ACC.hpp)  They were necessary anyway in imresh, but now they become nested.
 - ~~I couldn't find a `ALPAKA_ADD_LIBRARY` `CMake` function, only `ALPAKA_ADD_EXECUTABLE`, see commit 67e9554158212e76a895ba135b20feb1c7768f28~~ added in #184

BenjaminW3:
You are talking about multiple issues, therefore I will start adressing the last one.
Issue #64 already mentions the missing support for `ALPAKA_ADD_LIBRARY`.
Your implementation can be merged into alpaka after adding the new CMake file to [this line](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/alpakaConfig.cmake#L363). Could you please create a Pull Request?

BenjaminW3:
To adress your first point: I do not remember any strict reason for the `__forceinline__` and `inline` declarations. They are only there because the [original libPMacc declarations](https://github.com/ComputationalRadiationPhysics/picongpu/blob/release-0.1.2/src/libPMacc/include/types.h#L59-L61) used them. Maybe @psychocoderHPC knows more about the reasoning behind those.
I can imagine some performance advantages with those declarations, but without reliable tests showing this, I would agree to remove them.

[...]

Issue #64, opened 2015-09-20
add ALPAKA_ADD_LIBRARY


BenjaminW3:
The implementation should be similar to `ALPAKA_ADD_EXECUTABLE`.

PR #184, opened 2016-02-27
Issue64

mxmlnkn:
This should add `alpakaAddLibrary` support in `CMake`. I successfully compiled a static library with this on linux using `-DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=ON` and `nm -C libimresh.a | grep imresh` showed the functions appearing. But I didn't try running anything yet and other accelerators especially CUDA which could be a bit more finicky about compiling to a library (e.g. see [here](http://stackoverflow.com/questions/35073759/why-do-i-get-a-segmentation-fault-when-calling-a-cuda-kernel-directly-from-a-sha/35103383#35103383) and [here](http://stackoverflow.com/questions/35106360/why-does-this-cuda-program-crash-when-omitting-the-const-qualifier/35183878#35183878)) still need to be tested. See #64 and #183


erikzenker:
:+1:

Issue #182, opened 2016-02-26
alpaka no allows to allocate 0 byte


psychocoderHPC:
Is it necessary that alpaka not allow to allocate zero byte?
 - zero byte allocations are disallowed by [this](https://github.com/ComputationalRadiationPhysics/alpaka/blob/master/include/alpaka/mem/buf/BufCudaRt.hpp#L393) asserts
Also to free a NULL pointer is not allowed. (https://github.com/ComputationalRadiationPhysics/alpaka/blob/master/include/alpaka/mem/buf/BufCudaRt.hpp#L109)

[This source](http://stackoverflow.com/a/4133555) points to the C++ standard which explicit allowes zero byte allocations.
This (source)[http://dietmar-kuehl.de/mirror/c++-faq/freestore-mgmt.html#faq-16.8] say that free of a NULL pointer is allowed.

@BenjaminW3 Did I miss something that avoid use to remove the asserts and allow zero byte allocations?

BenjaminW3:
Yes, allocating zero size buffers and freeing a nullptr are both valid.
I do not see a reason for the assertions. I may have added them because such a usage is an indication of a bug in most cases (double free, wrong buffer size, etc...).
I will remove those assertions.

psychocoderHPC:
thx for this fast answer!

I can remove this, because if we remove this we need to disallow NULL pointer pinning in the cuda back-end. 
My inspiration for this question was to avoid double work if there is a reason for this asserts.

I create a pull request next week.

PR #185, opened 2016-02-27
zero size allocations

BenjaminW3:
This removes the assertions that did not allow to create zero size CPU and CUDA buffers and to delete a nullptr via `cudaFree` (issue #182).
Furthermore, a test verifying that the basic view operations are valid on such a (multidimensional) zero size buffer has been added.


psychocoderHPC:
Please try to pin a zero alloced host buffer! This should be fail because cuda not allow to pin zero byte memory.
A solution is to check the size while the memory should be  pinned.

BenjaminW3:
I have added an additional check.

BenjaminW3:
I will resolve the merge conflicts this evening.

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -519,6 +518,11 @@ namespace alpaka
                         if(!mem::buf::isPinned(buf))
                         {
 #if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && defined(__CUDACC__)
+                            if(buf.m_extentElements.prod() == 0u)
+                            {
+                                throw std::runtime_error("Zero size buffers can not be pinned!");
```
psychocoderHPC: Is this necessary to throw an error? We can ignore the pinning call if the memory contains zero elements.
The reason for that is that the memory behind a address is pinned. But there is no memory (zero elements) therefore not pinning is a clean and valid solution.

BenjaminW3: You are right. I will remove the throw.

BenjaminW3:
The AppVeyor build hit the time limit...

psychocoderHPC:
Big Thanks for this nice pull request!

Issue #187, opened 2016-02-28
::alpaka::atomic::atomicOp<::alpaka::atomic::op::Cas> only works for CUDA


mxmlnkn:
`::alpaka::atomic::atomicOp<::alpaka::atomic::op::Cas` is only defined `alpaka/include/alpaka/atomic/AtomicCudaBuiltIn.hpp`. When trying to compile with `-DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=ON` I get pages of error messages which basically boil down to this

    imresh/src/imresh/algorithms/alpaka/vectorReduce.tpp:80:195: error: no matching function for call to ‘atomicOp(const alpaka::acc::AccCpuOmp2Threads<std::integral_constant<long unsigned int, 3ul>, unsigned int>&, int*, int&, int)’
    imresh/cupla/alpaka/include/alpaka/atomic/Traits.hpp:66:33: note:   template argument deduction/substitution failed:
    In file included from imresh/src/imresh/algorithms/alpaka/vectorReduce.cpp:27:0:
    imresh/src/imresh/algorithms/alpaka/vectorReduce.tpp:80:195: note:   candidate expects 3 arguments, 4 provided
                 old = ::alpaka::atomic::atomicOp<::alpaka::atomic::op::Cas, decltype(acc), decltype(assumed)>( acc, (int*) rdpTarget, assumed, __float_as_int( f( __int_as_float(assumed), rValue ) ) );
                                                                                                                                                                                                       ^

BenjaminW3:
This is an oversight. The non-cuda back-ends only implement the two argument atomicOp version. I will add the 3 argument one.

mxmlnkn:
> EDIT: Has atomic::op::Cas ever worked on CUDA? It does not look like it should have worked on any accelerator at all.

I couldn't test on CUDA because of #186, I just assumed it would work, because the code was there

I did a `git-cherry-pick` on your commit, because `cupla` doesn't work with alpaka HEAD yet, and I get the following error:
    
    imresh/cupla/alpaka/include/alpaka/atomic/Traits.hpp:115:26: error: incomplete type 'alpaka::atomic::traits::AtomicOp<alpaka::atomic::op::Cas, const alpaka::acc::AccCpuOmp2Threads<std::integral_constant<long unsigned int, 3ul>, unsigned int>&, int, void>' used in nested name specifier
                         value);
                              ^
[...]

Issue #186, opened 2016-02-28
redefinition of argument 'std'


mxmlnkn:
When trying to compile with `CUDA` I get the following error:

    cmake .. -DIMRESH_DEBUG=ON -DCMAKE_CXX_COMPILER=$(which g++-4.9) -DCMAKE_C_COMPILER=$(which gcc-4.9) -DBUILD_DOC=OFF -DALPAKA_ACC_GPU_CUDA_ENABLE=ON
    make
        [ 16%] Built target alpaka
        Scanning dependencies of target cupla
        [ 25%] Linking CXX static library libcupla.a
        [ 25%] Built target cupla
        [ 33%] Building NVCC (Device) object CMakeFiles/imresh.dir/cupla/src/imresh_generated_stream.cpp.o
        nvcc fatal   : redefinition of argument 'std'
        CMake Error at imresh_generated_stream.cpp.o.cmake:207 (message):
          Error generating ./imresh/build/CMakeFiles/imresh.dir/cupla/src/./imresh_generated_stream.cpp.o

        CMakeFiles/imresh.dir/build.make:112: recipe for target 'CMakeFiles/imresh.dir/cupla/src/imresh_generated_stream.cpp.o' failed
        make[2]: *** [CMakeFiles/imresh.dir/cupla/src/imresh_generated_stream.cpp.o] Error 1
        CMakeFiles/Makefile2:141: recipe for target 'CMakeFiles/imresh.dir/all' failed
        make[1]: *** [CMakeFiles/imresh.dir/all] Error 2
        Makefile:83: recipe for target 'all' failed
        make: *** [all] Error 2

I'm using `CMake 3.4.1`. I guess the problem is related to http://stackoverflow.com/questions/29121211/cuda-compilation-issue-with-cmake 
`CUDA_PROPAGATE_HOST_FLAGS` is explicitly set to `ON` in `alpakaConfig.cmake` suggesting turning off won't be a good idea. It really isn't; when turning it off the following error occurs

    [ 16%] Built target alpaka
    [ 25%] Built target cupla
    [ 33%] Building NVCC (Device) object CMakeFiles/imresh.dir/cupla/src/imresh_generated_stream.cpp.o
    nvcc fatal   : redefinition of argument 'gpu-architecture'
    CMake Error at imresh_generated_stream.cpp.o.cmake:203 (message):
[...]

PR #188, opened 2016-02-28
fix atomic::op::Cas

BenjaminW3:
Fixes issue #187 and additionally adds a test for all atomic operations with different types (if supported by the accelerator).

### DIFF
--- a/include/alpaka/atomic/Traits.hpp
+++ b/include/alpaka/atomic/Traits.hpp
@@ -21,9 +21,11 @@
 
 #pragma once
 
-#include <alpaka/core/Common.hpp>   // ALPAKA_FN_HOST_ACC
+#include <alpaka/core/Common.hpp>       // ALPAKA_FN_HOST_ACC
 
-#include <type_traits>              // std::enable_if, std::is_base_of, std::is_same, std::decay
+#include <alpaka/meta/IsStrictBase.hpp> // meta::IsStrictBase
+
+#include <type_traits>                  // std::enable_if
 
 namespace alpaka
 {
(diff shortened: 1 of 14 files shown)


mxmlnkn:
> EDIT: Has atomic::op::Cas ever worked on CUDA? It does not look like it should have worked on any accelerator at all.

I couldn't test on CUDA because of #186, I just assumed it would work, because the code was there

I did a `git-cherry-pick` on your commit, because `cupla` doesn't work with alpaka HEAD yet, and I get the following error:
    
    imresh/cupla/alpaka/include/alpaka/atomic/Traits.hpp:115:26: error: incomplete type 'alpaka::atomic::traits::AtomicOp<alpaka::atomic::op::Cas, const alpaka::acc::AccCpuOmp2Threads<std::integral_constant<long unsigned int, 3ul>, unsigned int>&, int, void>' used in nested name specifier
                         value);
                              ^
The only merge conflicts were in `.travis.yml` and `test/unit/atomic/CMakeLists.txt` so I don't think I changed anything in a wrong way. See this commit https://github.com/mxmlnkn/alpaka/commit/1ba560329cdbc30ca581d0c403ff841f5927571e

Note that I had the exact same bug when I tried to implement this fix myself, but I deleted my CMakeCache and everything.

BenjaminW3:
Yes, I know. The CI tells me the same. I am trying to fix it...

psychocoderHPC:
@BenjaminW3 thanks for the fast fixes. 

Please wait that the tests are finished before merging! Do not merge your own pull request this breaks the pear reviewing.

include/alpaka/atomic/Traits.hpp
```diff
@@ -80,6 +82,41 @@ namespace alpaka
...
+        //! \param addr The value to change atomically.
+        //! \param compare The comparison value used in the atomic operation.
+        //! \param value The value used in the atomic operation.
+        //-----------------------------------------------------------------------------
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<
+            typename TOp,
+            typename TAtomic,
+            typename T>
+        ALPAKA_FN_HOST_ACC auto atomicOp(
```
psychocoderHPC: Why is this a fixed 4 elements signature and not something like.

```
ALPAKA_NO_HOST_ACC_WARNING
template<
    typename TOp,
    typename TAtomic,
    typename... TArgs>
ALPAKA_FN_HOST_ACC auto atomicOp(
    TAtomic const & atomic,
    TArgs & args...);
```

BenjaminW3: There is no hard reason against your solution. The current version is just easier to debug if it is called with too many arguments as it shows the error at an earlier stage.

psychocoderHPC: My point goes in the direction that not all atomics had a address, compare and value.
But I am fine with this implementation if this is still valid for all alpaka provided atomic.

PR #189, opened 2016-02-28
more intricate argument evaluation for alpaka_add_library

mxmlnkn:
The problem here is the following positional argument difference:

    CUDA_ADD_LIBRARY( <name> file0 file1 ...  [STATIC | SHARED | MODULE]
    add_library( <name> [STATIC | SHARED | MODULE]  source1 [source2 ...] )

If one works, then the other won't work. E.g. the error message would be:

    CMake Error at cupla/alpaka/cmake/addLibrary.cmake:35 (ADD_LIBRARY):
      Cannot find source file:

        STATIC

      Tried extensions .c .C .c++ .cc .cpp .cxx .m .M .mm .h .hh .h++ .hm .hpp
      .hxx .in .txx
      CMakeLists.txt:146 (alpaka_add_library)

This commit automatically reorders `STATIC` and other arguments as needed. Of course this means, that for more complex library calls this function may need to be extended. E.g. `add_library` supports an `ALIAS` subcommand which `ADD_CUDA_LIBRARY` doesn't and so on.

### DIFF
--- a/cmake/addLibrary.cmake
+++ b/cmake/addLibrary.cmake
@@ -18,22 +18,107 @@
 CMAKE_MINIMUM_REQUIRED(VERSION 3.3.0)
 
 #------------------------------------------------------------------------------
-# Calls CUDA_ADD_EXECUTABLE or ADD_EXECUTABLE depending on the enabled alpaka accelerators.
+# Calls CUDA_ADD_LIBRARY or ADD_LIBRARY depending on the enabled alpaka
+# accelerators.
+#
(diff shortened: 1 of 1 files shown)


BenjaminW3:
This looks really cool!
Could you please use upper case for all the CMake commands for consistency?

mxmlnkn:
I force pushed something

BenjaminW3:
Thanks!

bussmann:
Just wanted to tell you two that you're doing a great job! Greatly appreciate what you're doing.

PR #190, opened 2016-02-29
fix `getFreeGlobalMemSizeBytes`

psychocoderHPC:
`/proc/meminfo` defines memory in `KiB` and not byte.
add missing factor `1024` to fix the issue


BenjaminW3:
The appveyor tests take too long and get stopped automatically. I will have to disable some things. For now you can ignore it.

erikzenker:
The tests really take a lot of time (5,5 h on travis) ...

PR #191, opened 2016-02-29
add difference for `Vec`

psychocoderHPC:
- add struct `CreateSub`
- add `operator-(Vec,Vec)`


BenjaminW3:
The appveyor tests take too long and get stopped automatically. I will have to disable some things. For now you can ignore it.

erikzenker:
Thx for that PR ...I just want to mention that the name CreateSub sound like creating a sub vector, but the name is consistent with the other methods in the file.

PR #193, opened 2016-02-29
reduce the CI compilation times

BenjaminW3:
This Pull Request reduces the compilation times (issue #192).
The atomic operations unit test increased the gcc and MSVC compilation times dramatically. However, clang was not affected that much.

PR #195, opened 2016-02-29
consistently use meta::IsStrictBase

BenjaminW3:
By consistently using the `alpaka::meta::IsStrictBase` trait instead of the inconveniant `std::is_base_of && !std::is_same`, the code is better understandable.


BenjaminW3:
CI fails because AppVeyor build times exceed the maximum compilation time (fixed in #193)

erikzenker:
Well Done :-)

PR #194, opened 2016-02-29
remove inline and __forceinline__ from ALPAKA_FN_XXX

BenjaminW3:
Solves a part of #183.

PR #196, opened 2016-02-29
adding ALPAKA_FN_NO_INLINE variants

mxmlnkn:
My current fix / suggestion for working with explicit template instantiations, see #183. I think this is the least intrusive.


BenjaminW3:
As it was shown in #194 that we do not need `inline`by default, I will close this.

PR #198, opened 2016-03-03
BufCudaRT allow zero size memory allocation

psychocoderHPC:
- remove asserts those avoid to allocate zero myte memory
- this is a follow up of #185

### DIFF
--- a/include/alpaka/mem/buf/BufCudaRt.hpp
+++ b/include/alpaka/mem/buf/BufCudaRt.hpp
@@ -378,7 +378,6 @@ namespace alpaka
                             cudaMalloc(
                                 &memPtr,
                                 widthBytes));
-                        assert(memPtr);
 
 #if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
                         std::cout << BOOST_CURRENT_FUNCTION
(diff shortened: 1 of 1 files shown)


include/alpaka/mem/buf/BufCudaRt.hpp
```diff
@@ -436,8 +435,7 @@ namespace alpaka
                                 &pitchBytes,
                                 widthBytes,
                                 height));
-                        assert(memPtr);
-                        assert(pitchBytes>=widthBytes);
+                        assert(pitchBytes>=widthBytes||(width*height)==0);
```
BenjaminW3: Do you know which pitch cudaMallocPitch returns if the width = 0?
If the pitch is also 0, the original assert would be enough and would not need to be extended.

BenjaminW3:
Thanks for finding those!
If we could execute the CUDA enabled test in the CI, those omissions would have been found by the unit tests.

include/alpaka/mem/buf/BufCudaRt.hpp
psychocoderHPC: I am not sure. If the with is zero the old assert is enough but what is if `height` is zero. Therefore I added the second check.

psychocoderHPC:
Could we remove the experimentel fiber checks from `appveyor`?

BenjaminW3:
Yes we can do this. As the compilation is currently failing, the error message printing slows down compilation immensly. You can issue a new PR and set [this line](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/appveyor.yml#L72) to `OFF`.

PR #199, opened 2016-03-04
close #197: fix race condition

### LINKED ISSUE #197: Reference to already destroyed objects
I ported PIConGPU with the help of cupla to alpaka and get a deadlock that a kernel never finished ( the event after the kernel is not triggered).
I am not 100% sure if this is an alpaka bug but I think so.

If I use `::alpaka::stream::StreamCpuAsync` and `::alpaka::acc::AccCpuOmp2Threads<>` I can trigger a deadlock in the most runs. With `::alpaka::stream::StreamCpuSync` the deadlock is never triggered.

I think the problem is trigger by some code part like [this](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/core/ConcurrentExecPool.hpp#L362) where a functor is created and some arguments are [bind as reference](http://de.cppreference.com/w/cpp/utility/functional/bind). If the functor which is created with bind is passed to a thread than it can be that the (internal stored) references are not valid anymore.
This can be the case if an alpaka kernel is created in a local scope and passed to `::alpaka::stream::enqueue()`

```C++
{
 MyKernel kernel;
 auto dev( manager::Device<AccDev>::get().current() );
  ::alpaka::workdiv::WorkDivMembers<
    KernelDim,
    IdxType
  > workDiv(
      static_cast<IdxVec3>(gridSize),
      static_cast<IdxVec3>(blockSize),
      static_cast<IdxVec3>(elemPerThread)
  );
  auto const exec(::alpaka::exec::create<Acc>(workDiv, kernel, args...));
  ::alpaka::stream::enqueue(stream, exec);
} // kernel, workDiv and args are now destroyed and get invalid if references are used in a different thread which is already running
```

@BenjaminW3 Currently I can't find the place were we must definitely create a copy of the kernel and the given parameter to avoid that references are stored in [TaskPackages](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/core/ConcurrentExecPool.hpp#L107). If my description of the problem is not clear enough could you please call me by phone (at work). I need some help to locate the broken code part.

psychocoderHPC:
- avoid that references are given to threads
- guard `notify_one` to avoid race conditions
- close #197

This branch removes also some `forward<>()` to avoid that a thread get the functors and parameters by reference to temporary resources of an other tread.

This pull request is rebased against #211 to avoid that appveyor fiber test always fails.

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -184,7 +184,9 @@ namespace alpaka
 
                 TPromise<TFnObjReturn> m_Promise;
             private:
-                TFnObj m_FnObj;
+                // to avoid out of memory access `std::remove_reference` forbid
+                // that local references of an other thread can be stored as member 
+                typename std::remove_reference<TFnObj>::type m_FnObj;
             };
 
             //#############################################################################
@@ -235,7 +237,9 @@ namespace alpaka
 
                 TPromise<void> m_Promise;
             private:
-                TFnObj m_FnObj;
+                // to avoid out of memory access `std::remove_reference` forbid
+                // that local references of an other thread can be stored as member
+                typename std::remove_reference<TFnObj>::type m_FnObj;
             };
 
             //#############################################################################
@@ -359,7 +363,7 @@ namespace alpaka
                 -> typename std::result_of< decltype(&TPromise<typename std::result_of<TFnObj(TArgs...)>::type>::get_future)(TPromise<typename std::result_of<TFnObj(TArgs...)>::type> *) >::type
 #endif
                 {
-                    auto boundTask(std::bind(std::forward<TFnObj>(task), std::forward<TArgs>(args)...));
+                    auto boundTask(std::bind(task, args...));
 
                     // Return type of the function object, can be void via specialization of TaskPkg.
                     using FnObjReturn = typename std::result_of<TFnObj(TArgs...)>::type;
@@ -578,7 +582,7 @@ namespace alpaka
                 -> typename std::result_of< decltype(&TPromise<typename std::result_of<TFnObj(TArgs...)>::type>::get_future)(TPromise<typename std::result_of<TFnObj(TArgs...)>::type> *) >::type
 #endif
                 {
-                    auto boundTask(std::bind(std::forward<TFnObj>(task), std::forward<TArgs>(args)...));
+                    auto boundTask(std::bind(task, args...));
 
                     // Return type of the function object, can be void via specialization of TaskPkg.
                     using FnObjReturn = typename std::result_of<TFnObj(TArgs...)>::type;
@@ -589,9 +593,12 @@ namespace alpaka
 
                     auto future(pTaskPackage->m_Promise.get_future());
 
-                    m_qTasks.push(std::move(upTaskPackage));
+                    {
+                        std::lock_guard<TMutex> lock(m_mtxWakeup);
+                        m_qTasks.push(std::move(upTaskPackage));
 
-                    m_cvWakeup.notify_one();
+                        m_cvWakeup.notify_one();
+                    }
 
                     return future;
                 }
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
The result of the race condition was that a thread add new work to the queue of the worker thread and notify the worker. The worker missed the notify if the thread is not in the conditional waiting function. If the worker than enter the waiting and no new work is added to the queue the thread never wake up.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -73,7 +73,7 @@ namespace alpaka
                 {
                     std::lock_guard<std::mutex> lk(m_Mutex);
 
-                    std::queue<T>::push(std::forward<T>(t));
+                    std::queue<T>::push(t);
```
BenjaminW3: I am very sure that this forward is no problem but a real optimization.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -159,7 +159,7 @@ namespace alpaka
                 TaskPkg(
                     TFnObj && func) :
                         m_Promise(),
-                        m_FnObj(std::forward<TFnObj>(func))
+                        m_FnObj(func)
```
BenjaminW3: The same as above should be true here. Therefore we should use `typename std::remove_reference<TFnObj>::type m_FnObj;` as the member to allow only copies and moves as the current version (even without forward) allows to hold a reference.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -209,7 +209,7 @@ namespace alpaka
                 TaskPkg(
                     TFnObj && func) :
                         m_Promise(),
-                        m_FnObj(std::forward<TFnObj>(func))
```
BenjaminW3: Same here.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -589,6 +589,7 @@ namespace alpaka
 
                     auto future(pTaskPackage->m_Promise.get_future());
 
+                    std::lock_guard<TMutex> lock(m_mtxWakeup);
                     m_qTasks.push(std::move(upTaskPackage));
 
                     m_cvWakeup.notify_one();
```
BenjaminW3: Could you please add `{` and `}` around lines 592 to 595 to make it stand out better?

include/alpaka/core/ConcurrentExecPool.hpp
psychocoderHPC: But `t` is a reference to a function local shared pointer of a thread different to the worker thread. Therefore it is not allowed to hold a reference.

include/alpaka/core/ConcurrentExecPool.hpp
psychocoderHPC: Yes I will add this change to this pull request

include/alpaka/core/ConcurrentExecPool.hpp
psychocoderHPC: ok

include/alpaka/core/ConcurrentExecPool.hpp
BenjaminW3: No, because `push` is no template, `T && t` is no 'universal reference' (representing either `&` or `&&`) but is forced to be a rvalue reference (`&&`). This means that `t` can only ever be moved into the `push` function. This means that the caller will always give up his reference and will transfer the ownership to the `push` function which will then transfer it to the internal queue via `forward`. At the call side, this can be reognized by the `std::move(upTaskPackage)`. It will not compile without the `std::move`.

psychocoderHPC: OK thanks for clarifing

include/alpaka/core/ConcurrentExecPool.hpp
psychocoderHPC: I add the last changes after the GTC2016.

psychocoderHPC:
updated

BenjaminW3:
Thanks!

PR #207, opened 2016-03-22
fix comparison of EventCudaRt

### LINKED ISSUE #204: EventCudaRt.hpp:158:65: warning: unused parameter ‘rhs’
```
alpaka/include/alpaka/event/EventCudaRt.hpp: At global scope:
alpaka/include/alpaka/event/EventCudaRt.hpp:158:65: warning: unused parameter ‘rhs’ [-Wunused-parameter]
             ALPAKA_FN_HOST auto operator==(EventCudaRt const & rhs) const
                                                                 ^
alpaka/include/alpaka/core/Cuda.hpp: In instantiation of ‘void alpaka::cuda::detail::cudaRtCheckIgnore(const cudaError_t&, const char*, const char*, const int&, TErrors&& ...) [with TErrors = {}; <template-parameter-1-2> = void; cudaError_t = cudaError]’:
alpaka/include/alpaka/dev/DevCudaRt.hpp:123:241:   required from here
```
The problem are these source code lines:
```C++
//! Equality comparison operator.
ALPAKA_FN_HOST auto operator==(EventCudaRt const & rhs) const
-> bool
{
    return (m_spEventCudaImpl->m_CudaEvent == m_spEventCudaImpl->m_CudaEvent);
}
```
I'm amazed there is no warning about comparison of a variable with itself, this should only throw no compiler warning in my opinion, if the variable was volatile. Either it wants to return true every time, or it should actually be:
```C++
    return (m_spEventCudaImpl->m_CudaEvent == rhs.m_CudaEvent);
```
In my version this leads to

    EventCudaRt.hpp(161): error: class "alpaka::event::EventCudaRt" has no member "m_CudaEvent"

So I just commented out the `rhs` parameter name to quelch this warning:
```C++
ALPAKA_FN_HOST auto operator==(EventCudaRt const & /*rhs*/) const
```

BenjaminW3:
fixes #204

### DIFF
--- a/include/alpaka/event/EventCudaRt.hpp
+++ b/include/alpaka/event/EventCudaRt.hpp
@@ -158,7 +158,7 @@ namespace alpaka
             ALPAKA_FN_HOST auto operator==(EventCudaRt const & rhs) const
             -> bool
             {
-                return (m_spEventCudaImpl->m_CudaEvent == m_spEventCudaImpl->m_CudaEvent);
+                return (m_spEventCudaImpl->m_CudaEvent == rhs.m_spEventCudaImpl->m_CudaEvent);
             }
             //-----------------------------------------------------------------------------
             //! Equality comparison operator.


erikzenker:
thx for fixing this so fast :-)

PR #211, opened 2016-04-12
disable appveyor fiber test case

psychocoderHPC:
Disable boost fibers test.

To disable the broken fiber test were suggested [here](https://github.com/ComputationalRadiationPhysics/alpaka/pull/198#issuecomment-192217032)


psychocoderHPC:
@BenjaminW3 There is something wrong with the appveyor checks while some git commands are running. Could you see whats wrong?

BenjaminW3:
This seems to be a problem of appveyor (maybe traffic / connection to github?). Other builds of today have been green.

PR #216, opened 2016-04-13
test cmake 3.5 and clang OpenMP

BenjaminW3:
This extents the CI tests to use CMake 3.5.2 and enables OpenMP tests with clang. OpenMP 4.0 is supported starting with clang 3.8.


BenjaminW3:
@ComputationalRadiationPhysics/alpaka-maintainers Ready to merge CMake 3.5.2 tests ;-)

alpakaConfig.cmake
```diff
@@ -184,6 +184,21 @@ ENDIF()
 #-------------------------------------------------------------------------------
 IF(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR ALPAKA_ACC_CPU_BT_OMP4_ENABLE)
     FIND_PACKAGE(OpenMP)
+
+    # Manually find OpenMP for the clang compiler if it was not already found.
+    # Even CMake 3.5 is unable to find libiomp and provide the correct OpenMP flags.
+    IF(NOT OPENMP_FOUND)
+        IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
psychocoderHPC: Should we also check that the clang version is `>=3.8`, to be sure that the clang version supports `OpenMP`?

BenjaminW3: Yes, we could check for clang `>=3.7` but it is no problem, as the old versions just ignore the fopenmp flag.

psychocoderHPC: Should I wait for an update of this pull request or would shift it to a new pull request.

BenjaminW3: I would do it in a follow-up pull request.

PR #218, opened 2016-04-19
CUDA only mode / CUDA compatibility

BenjaminW3:
This adds a check if the define `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is set in which case `ALPAKA_FN_ACC` is defined to `__device__` only instead of `__host__ __device__` which is used in the default mode. In the CUDA only case alpaka is compatible with CUDA code and CUDA intrinsics can be used within alpaka kernels.

- [x] add a test to the CI


psychocoderHPC:
Note: This is a feature which we need before we release alpaka 0.1.0

BenjaminW3:
I have added a test for the CI and everything works now as expected.

As a follow up, the usage of BOOST_REQUIRE_EQUAL has to be removed from the kernel functions as this prints a warning (calling `__host__` function from a `__host__ __device__` function) in the default mode, but will not compile at all in `ALPAKA_ACC_GPU_CUDA_ONLY_MODE`. I have opened ticket #230 for this.

psychocoderHPC:
big thanks for this change.

I will merge this pull request but have a question (maybe for a follow up):
Should we set `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` automatically if only the cuda backend is enabled?

BenjaminW3:
I do not think that this would be correct. If the user has implemented an additional CPU back-end, does not want to use any of our back-ends, and wants to use our CUDA back-end, this would conflict.

PR #220, opened 2016-04-22
enable clang native CUDA compilation take two

### LINKED ISSUE #214: Clang CUDA backend
It would be great if we could test the open source [clang 3.8+ CUDA](http://llvm.org/releases/3.8.0/tools/clang/docs/ReleaseNotes.html#cuda-support-in-clang) end-to-end compilation with our CUDA backend in our [compiler list](https://github.com/ComputationalRadiationPhysics/alpaka#supported-compilers) and report problems that occur to [`cfe-dev` list](http://lists.llvm.org/mailman/listinfo/cfe-dev).

BenjaminW3:
Due to the force push while the Pull Request was closed, github was unable to reopen #165. See there for a description.
See [here](http://llvm.org/docs/CompileCudaWithLLVM.html) for the CUDA with LLVM documentation.
This fixes #214.


BenjaminW3:
@ComputationalRadiationPhysics/alpaka-maintainers After a lot of work I have finally finished to get clang-3.8 to natively compile alpaka with enabled CUDA accelerator without nvcc!
You simply have to set `ALPAKA_CUDA_COMPILER=clang` in cmake!

For now, there are some restrictions:
- 3 of the unit tests fail to compile (clang crashes)
- cuRAND is not supported
- I have not tested if the binaries are executable or if the execution results are correct! I would be happy if someone could verify this!

I do not intent to make everything perfect within this Pull Request. This only enables compilation. If there are more things to do I would like do open follow-up Pull Requests. If reasonable, I would like to have this merged as soon as possible as the possibility of future merge conflicts is very high.

ax3l:
I usually try to avoid spam but this is amazing! Wow Benjamin! :sparkles: 

Let us just send a request for enhancement upstream about cuRAND so they know users want it :)

bussmann:
Amazing work! Big 'Thank you'!

psychocoderHPC:
Thanks for this nice pull request. The review will take some time do to the number of changed files and the upcoming public holidays .

alpakaConfig.cmake
```diff
@@ -253,47 +253,76 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+            IF(ALPAKA_CUDA_COMPILER MATCHES "clang")
+                LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "--cuda-gpu-arch=${ALPAKA_CUDA_ARCH}")
 
-            OPTION(ALPAKA_CUDA_FAST_MATH "Enable fast-math" ON)
-            IF(ALPAKA_CUDA_FAST_MATH)
-                LIST(APPEND CUDA_NVCC_FLAGS "--use_fast_math")
-            ENDIF()
+                # This flag silences the warning produced by the Dummy.cpp files:
+                # clang: warning: argument unused during compilation: '--cuda-gpu-arch=sm_20'
+                LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-Qunused-arguments")
```
psychocoderHPC: Could we only enable this if `sm_20` is used?

include/alpaka/core/Common.hpp
```diff
@@ -23,20 +23,75 @@
...
+#define BOOST_PREDEF_MAKE_10_VVRRP(V) BOOST_VERSION_NUMBER(((V)/1000)%100,((V)/10)%100,(V)%10)
+
+//-----------------------------------------------------------------------------
+// CUDA language detection
+// - clang defines __CUDA__ when compiling CUDA code ('-x cuda')
+// - nvcc defines __CUDACC__ when compiling CUDA code
+//-----------------------------------------------------------------------------
+#if defined(__CUDA__) || defined(__CUDACC__)
+    #include <cuda.h>
+    #define BOOST_LANG_CUDA BOOST_PREDEF_MAKE_10_VVRRP(CUDA_VERSION)
```
psychocoderHPC: Why we need to set `BOOST_LANG_CUDA`? Is this a trigger for the library boost?

alpakaConfig.cmake
BenjaminW3: This is not the problem. Clang warns for all sm's. The problem is that the Dummy.cpp file does not contain any CUDA code and therefore the flag is useless for the compilation of this file.

include/alpaka/core/Common.hpp
BenjaminW3: I am exending the Boost Predef library here. To make it more clear that the values of `BOOST_LANG_CUDA` and the other macros follow the rules of the Predef library, I use the same 'macro namespace'. Finally, if they prove beneficial, I intent to propose those extensions to Boost Predef itself.

BenjaminW3:
I have added comments into the source for both of your annotations.

psychocoderHPC:
I will check this pull request on Tuesday again and merge it if all is ok.

**Note:** Please split future pull requests if they contain several  independent changes e.g., this pull request contains renaming of defines, new handling of version numbers, cmake changes and the clang integration

psychocoderHPC:
Thanks for the cuda clang support integration

PR #223, opened 2016-04-29
fix that atomic operations are not grid atomics

### LINKED ISSUE #222: AtomicStlLock only atomic inside a block?
Is the implementation of `AtomicStlLock` only atomic for threads inside a block?
https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/atomic/AtomicStlLock.hpp#L73
The used mutex is part of the accelerator. Each block gets its own accelerator. This looks like that a atomic operation is **not** atomic between different blocks within a grid.

psychocoderHPC:
close #222

- AccCpuFiber: inherit now from `AtomicStlLock`
- AtomicStlLock: use static mutex member to allow grid global atomics

With this fix all atomics are know grid atomic. In difference to Cuda atomics those are device atomics.

### DIFF
--- a/include/alpaka/atomic/AtomicStlLock.hpp
+++ b/include/alpaka/atomic/AtomicStlLock.hpp
@@ -69,8 +69,11 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             ALPAKA_FN_ACC_NO_CUDA /*virtual*/ ~AtomicStlLock() = default;
 
-        private:
-            std::mutex mutable m_mtxAtomic; //!< The mutex protecting access for a atomic operation.
+            std::mutex & getMutex() const
+            {
+                static std::mutex m_mtxAtomic; //!< The mutex protecting access for a atomic operation.
+                return m_mtxAtomic;
+            }
         };
 
         namespace traits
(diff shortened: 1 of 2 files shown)


include/alpaka/atomic/AtomicStlLock.hpp
```diff
@@ -70,9 +70,12 @@ namespace alpaka
             ALPAKA_FN_ACC_NO_CUDA /*virtual*/ ~AtomicStlLock() = default;
 
         private:
-            std::mutex mutable m_mtxAtomic; //!< The mutex protecting access for a atomic operation.
+            static std::mutex m_mtxAtomic; //!< The mutex protecting access for a atomic operation.
         };
 
+        // initialize the static member of the class AtomicStlLock
+        std::mutex AtomicStlLock::m_mtxAtomic;
```
BenjaminW3: This does not compile.
You might have a look at [this Stackoverflow question](http://stackoverflow.com/questions/11709859/how-to-have-static-data-members-in-a-header-only-library).

psychocoderHPC:
Need to change [this](https://github.com/ComputationalRadiationPhysics/alpaka/pull/223/files#diff-225a98a657f7401ea7b37cf7e7f170dfR114) line, too.

psychocoderHPC:
This pull is ready to merge (I removed the WIP lable)

PR #225, opened 2016-05-05
disable infunctional clang-3.8 OpenMP4 accelerator

BenjaminW3:
This enables the OpenMP 4.0 accelerator within the CI tests.
Clang 3.8 OpenMP 4.0 support had to be removed, as clang 3.8 can parse OpenMP 4.0 correctly but is not able to generate code for all pragmas and therefore does not produce correct results in the tests (without any compiler warning!).
Furthermore, this adds some code that should make the debugging of such cases (clang-3.9?) easier.


erikzenker:
Thank you for this fix

Issue #209, opened 2016-04-12
Adding support for constant memory


theZiz:
I noticed with Isaac, that Cuda code with `__constant__` variables cannot be directly ported to Alpaka. Passing a const parameter to a kernel is slower than directly using `__constant__` variables.

So I would like to add a way for using `__constant__` variables with Alpaka. As `__constant__` variables cannot be member of a struct or class I would suggest to define a `#define` like `ALPAKA_CONST_VAR`, which resolves to `__constant__` for Cuda, but is empty for CPU accelerators. However I am not sure, how global variables on xeon phi devices work. Imho this should be considered before adding this feature.

Only the host can change the content of a `__constant__` variable in Cuda, either with copying from host to the variable (`cudaMemcpyToSymbol`) or directly from device memory to the constant memory (`cudaMemcpy`+ `cudaGetSymbolAddress`).

I am not sure, how to implement this best in an _Alpaka way_. Normally you would use `alpaka::mem::view::copy` for copying stuff. But as constant memory is very special, maybe a `alpaka::mem::const::copy` (which throughs exception if used in the wrong direction) would be more suitable?

So, please tell me your thoughts!

BenjaminW3:
Not being able to have a `__constant__` variable as a member is a hard limitation.
Adding a `ALPAKA_VAR_CONST` defined to `__constant__` for CUDA seems to be the way to go (at a first look).
Please note that each device in a multi-GPU setup has a seperate instance of the `__constant__` memory. Therefore the copy needs to know the specific device.

I would suggest to create a method `alpaka::mem::const::getView(dev, ptr, extent)` which creates a `ViewConstVar` given the device, the constant variable pointer and its extent.
Internally it calls `cudaGetSymbolAddress` to get the pointer on the device.
Later on, this `ViewConstVar` can be used like any other view in a standard `alpaka::mem::view::copy` call that automatically calls the correct memcpy. This should even support inter-GPU `__constant__` memory copies.  

However, as you said, it would be preferable if someone could find out how something similar could work with offloaded OpenMP code or if this is not possible at all. As we do not support OpenMP offloading at the moment, this would not be a blocker.

theZiz:
As I got a bit more spare time I had the possibility to think about this feature request. I think the most important thing you mentioned is, that every cuda device got it's own constant memory. So the same has to be the case for any other accelerator. I am right, that this is not a problem at all, as every not-gpu device count is hard coded to 1 atm?

[...]

Issue #60, opened 2015-09-13
add kernel static global memory allocation


BenjaminW3:
Currently only dynamic memory allocation via ```new``` / ```malloc``` is supported within a kernel. Static memory allocations with the ```__device__``` attribute are currently not available.
This can be implemented in the same way as the shared memory allocations are done.

erikzenker:
Fixed by #226

PR #226, opened 2016-05-08
CUDA static device memory

BenjaminW3:
This implements #209 (\__constant\__) and #60 (\__device\__).

### DIFF
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -146,3 +146,21 @@
 #else
     #define ALPAKA_FN_INLINE inline
 #endif
+
+//-----------------------------------------------------------------------------
+//! This macro defines a variable lying in global accelerator device memory.
+//-----------------------------------------------------------------------------
+#if BOOST_LANG_CUDA && BOOST_ARCH_CUDA_DEVICE
+    #define ALPAKA_STATIC_DEV_MEM_GLOBAL __device__
+#else
+    #define ALPAKA_STATIC_DEV_MEM_GLOBAL
+#endif
+
+//-----------------------------------------------------------------------------
+//! This macro defines a variable lying in constant accelerator device memory.
+//-----------------------------------------------------------------------------
+#if BOOST_LANG_CUDA && BOOST_ARCH_CUDA_DEVICE
+    #define ALPAKA_STATIC_DEV_MEM_CONSTANT __constant__
+#else
+    #define ALPAKA_STATIC_DEV_MEM_CONSTANT
+#endif
--- a/include/alpaka/mem/view/Traits.hpp
+++ b/include/alpaka/mem/view/Traits.hpp
@@ -191,6 +191,14 @@ namespace alpaka
                     typename TDevSrc,
                     typename TSfinae = void>
                 struct TaskCopy;
+
+                //#############################################################################
+                //! The static device memory view creation trait.
+                //#############################################################################
+                template<
+                    typename TDev,
+                    typename TSfinae = void>
+                struct CreateStaticDevMemView;
             }
 
             //-----------------------------------------------------------------------------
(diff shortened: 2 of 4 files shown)


test/unit/mem/view/src/ViewConstantMem.cpp
```diff
@@ -0,0 +1,139 @@
...
+        alpaka::mem::view::ViewPlainPtr<DevAcc, const Elem, Dim, Size> bufHost(data.data(), devAcc, extent);
+
+        auto viewConstantMemUninitialized(
+            alpaka::mem::view::createConstantMemView(
+                &g_constantMemory2DUninitialized[0u][0u],
+                devAcc,
+                extent));
+
+        alpaka::mem::view::copy(streamAcc, viewConstantMemUninitialized, bufHost, extent);
+
```
erikzenker: I think there is an `alpaka::wait::wait(streamAcc)` missing here.

test/unit/mem/view/src/ViewConstantMem.cpp
```diff
@@ -0,0 +1,139 @@
...
+    }
+
+    //-----------------------------------------------------------------------------
+    // uninitialized global memory
+    {
+        using StreamAcc = alpaka::test::stream::DefaultStream<DevAcc>;
+        StreamAcc streamAcc(devAcc);
+
+        std::vector<Elem> const data{0, 1, 2, 3, 4, 5};
+        alpaka::mem::view::ViewPlainPtr<DevAcc, const Elem, Dim, Size> bufHost(data.data(), devAcc, extent);
```
erikzenker: Doesn`t this needs to be a host buffer ?

test/unit/mem/view/src/ViewConstantMem.cpp
```diff
@@ -0,0 +1,139 @@
...
+    {
+        auto const gridThreadExtent =
+            alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+        auto const gridThreadIdx =
+            alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+
+        auto const offset = gridThreadExtent[1u] * gridThreadIdx[0u] + gridThreadIdx[1u];
+        auto const val = offset;
+
+        BOOST_REQUIRE_EQUAL(
```
erikzenker: Cuda does not like this macro, but I think that this is an known issue

test/unit/mem/view/src/ViewConstantMem.cpp
BenjaminW3: Yes, you are right.

test/unit/mem/view/src/ViewConstantMem.cpp
BenjaminW3: Yes, you are right.

erikzenker:
Basically it works, but we have some problems:

* If we are using CPU and CUDA accelerator at the same time then all constant memory declarations are prefixed with `__constant__`, therefore this constant memory will live within the CUDA device. However, there is some [undocumented behaviour](http://stackoverflow.com/questions/34041372/access-cuda-global-device-variable-from-host) that for each constant memory on the device exist a constant memory on the host. This corresponding host memory is not initialized with the values from the initializer list (thats why the first test case will fail). Thus, the user has to initialize its constant memory by the copy all if it should work in all cases.

* Clang does not implement this undocumented behaviour mentioned above. Accordingly, no one know what happens when the CPU tries to access constant memory while GPU is also enabled.

BenjaminW3:
I have fixed the bugs you noted and added support not only for `\__constant\__`, but also for `\__device\__` memory.
Could you please have a look at it again?

test/unit/mem/view/src/ViewConstantMem.cpp
BenjaminW3: I have added issue #230 to address this problem.

erikzenker:
Thank you very much for that quick fix. The tests run through now (for cpu and gpu). I found no other issues while reading through the code. If there is no functionality left (WIP), then I think this is ready to merge.

BenjaminW3:
I have removed the work-in-progress label.

PR #221, opened 2016-04-25
TBB block parallel back-end

erikzenker:
This PR introduces a new alpaka back-end based on Intel TBB. It should be seen as a first draft using TBB in alpaka and follows the need specified in issue #65. The back-end is very similar to the OpenMP 2.0 block parallel back-end. Therefore, it supports blocks with one thread.

PR #228, opened 2016-05-11
add hierarchy atomics

### LINKED ISSUE #200: atomic operations
Atomic operations in **alpaka** differ from the cuda atomic operations restrictions.
In cuda atomics are valid between concurrent running kernel.
Thus alpaka used in openmp accelerators `omp critical section` to perform atomics they are only conflict free inside a grid (the running kernel). If two kernels were running  concurrently atomics are not conflict free between both.

- [ ] add a section to the documentation

### LINKED ISSUE #201: atomic operations with scope
Based on the fact in #200 I realized that we need a scope to a alpaka layer for atomic operations where they are atomic.
Currently the atomic operations are very slow if they are emulated (e.g., with `omp  critical section`) this result in code which is not performance portable.
Alpaka knows best which layer (grid, block, thread) needs which type of atomics.

We should add to all alpaka atomics a hint from the user in which scope the operation must be atomic.

``` C++
::alpaka::atomic::atomicOp<::alpaka::atomic::op::Add>(acc, ppPointer, ppValue, ::alpaka::Threads )
```

Scopes can be:
alpaka::Threads - operation is atomic between threads in a block
alpaka::Blocks - operation is atomic between blocks in a grid
alpaka::Grids - operation is atomic between grids on a device

Why the scope should not be based on memory (e.g., shared memory, global memory)?
The reason is that global memory can used by the user in that way that the domains of the blocks are independent. With this strategy and if the user defines  that the scope is `::alpaka::Threads` the atomic inside alpaka can be `atomic::AtomicNoOp` if the accelerator is serialized on thread level ( e.g. in `AccCpuSerial`, `AccCpuOmp2Blocks`)

psychocoderHPC:
close #200
close #201 

- add namespace hierarchy and add
  - class Grids
  - class Blocks
  - class Threads
- add class `AtomicHierarchy`
- add the needed hierarchies to the accelerators

This pull request solve #201 and should be a base for discussion.

There is a new namespace introduced named `hierarchy` and contains `Grids`,`Blocks` and `Threads`.
Each atomic operation can have a *optional* parameter which describe the scope were the atomic operation is guaranteed atomic (default: `Grids()`).

**The meaning of the hierarchies:**
- `Grids` -> the operation is atomic between independent grids in a single device and one accelerator type (**attention:** not between different accelerator types,` AccCpuOmp2Blocks` operations are not atomic to `AccCpuOmp2Threads` operations)
- `Blocks` -> the operation is atomic between blocks in a grid
- `Threads` -> the operation is atomic between threads in a block

There is no guarantee that a atomic operation in the `Grids` hierarchy to a address `x` is atomic to a operation to the same address in the `Blocks` hierarchy.
It is not possible to write generic concepts that atomics between different accelerators are atomic therefore all atomics are restricted to one accelerator type.
Currently atomic operations are atomic to the address which means that a atomic `Add` and a `Exch` to the same address  can be mixed.

```C++
//...
template< typename TAcc>
ALPAKA_FN_ACC void operator()( const TAcc & acc, int* x)
{
    // atomic between grids in a device of the same accelerator
    atomic::atomicOp<atomic::op::Add>(acc, x, 1, hierarchy::Grids()); 
    // atomic between blocks in a grid
    // there is no guarantee that the next operation is atomic to the first
    atomic::atomicOp<atomic::op::Add>(acc, x, 1, hierarchy::Blocks()); 
}
```

Maybe the namespace `hierarchy` should be renamed to `between` than a atomic operation is human readable `atomic::atomicOp<atomic::op::Add>(acc, x, 1, between::Blocks()); `

**TODO:**
- [x] add documentation to `class AtomicHierarchy`
- [x] rebase against #221 and modify the TBB backend

@ComputationalRadiationPhysics/alpaka-maintainers please participate the discussion

### DIFF
--- a/include/alpaka/acc/AccCpuThreads.hpp
+++ b/include/alpaka/acc/AccCpuThreads.hpp
@@ -1,6 +1,6 @@
 /**
 * \file
-* Copyright 2014-2016 Benjamin Worpitz
+* Copyright 2014-2016 Benjamin Worpitz, Rene Widera
 *
 * This file is part of alpaka.
 *
@@ -28,6 +28,7 @@
 #include <alpaka/idx/gb/IdxGbRef.hpp>               // IdxGbRef
 #include <alpaka/idx/bt/IdxBtRefThreadIdMap.hpp>    // IdxBtRefThreadIdMap
 #include <alpaka/atomic/AtomicStlLock.hpp>          // AtomicStlLock
+#include <alpaka/atomic/AtomicHierarchy.hpp>    // AtomicHierarchy
 #include <alpaka/math/MathStl.hpp>                  // MathStl
 #include <alpaka/block/shared/dyn/BlockSharedMemDynBoostAlignedAlloc.hpp>   // BlockSharedMemDynBoostAlignedAlloc
 #include <alpaka/block/shared/st/BlockSharedMemStMasterSync.hpp>            // BlockSharedMemStMasterSync
@@ -79,7 +80,11 @@ namespace alpaka
             public workdiv::WorkDivMembers<TDim, TSize>,
             public idx::gb::IdxGbRef<TDim, TSize>,
             public idx::bt::IdxBtRefThreadIdMap<TDim, TSize>,
-            public atomic::AtomicStlLock,
+            public atomic::AtomicHierarchy<
+                atomic::AtomicStlLock, // grid atomics
+                atomic::AtomicStlLock, // block atomics
+                atomic::AtomicStlLock  // thread atomics
+            >,
             public math::MathStl,
             public block::shared::dyn::BlockSharedMemDynBoostAlignedAlloc,
             public block::shared::st::BlockSharedMemStMasterSync,
@@ -108,7 +113,11 @@ namespace alpaka
                     workdiv::WorkDivMembers<TDim, TSize>(workDiv),
                     idx::gb::IdxGbRef<TDim, TSize>(m_gridBlockIdx),
                     idx::bt::IdxBtRefThreadIdMap<TDim, TSize>(m_threadToIndexMap),
-                    atomic::AtomicStlLock(),
+                    atomic::AtomicHierarchy<
+                        atomic::AtomicStlLock, // atomics between grids
+                        atomic::AtomicStlLock, // atomics between blocks
+                        atomic::AtomicStlLock  // atomics between threads
+                    >(),
                     math::MathStl(),
                     block::shared::dyn::BlockSharedMemDynBoostAlignedAlloc(static_cast<std::size_t>(blockSharedMemDynSizeBytes)),
                     block::shared::st::BlockSharedMemStMasterSync(
--- a/include/alpaka/atomic/AtomicCudaBuiltIn.hpp
+++ b/include/alpaka/atomic/AtomicCudaBuiltIn.hpp
@@ -1,6 +1,6 @@
 /**
  * \file
- * Copyright 2014-2015 Benjamin Worpitz
+ * Copyright 2014-2016 Benjamin Worpitz, Rene Widera
  *
  * This file is part of alpaka.
  *
@@ -38,11 +38,13 @@ namespace alpaka
     {
         //#############################################################################
         //! The GPU CUDA accelerator atomic ops.
+        //
+        //  Atomics can used in the hierarchy level grids, blocks and threads.
+        //  Atomics are not guaranteed to be save between devices
         //#############################################################################
         class AtomicCudaBuiltIn
         {
         public:
-            using AtomicBase = AtomicCudaBuiltIn;
 
             //-----------------------------------------------------------------------------
             //! Default constructor.
@@ -82,11 +84,13 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             //! The GPU CUDA accelerator atomic operation.
             //-----------------------------------------------------------------------------
-            template<>
+            template<
+                typename THierarchy>
             struct AtomicOp<
                 op::Add,
                 atomic::AtomicCudaBuiltIn,
-                int>
+                int,
+                THierarchy>
             {
                 //-----------------------------------------------------------------------------
                 //
@@ -103,11 +107,13 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             //! The GPU CUDA accelerator atomic operation.
             //-----------------------------------------------------------------------------
-            template<>
+            template<
+                typename THierarchy>
             struct AtomicOp<
                 op::Add,
                 atomic::AtomicCudaBuiltIn,
-                unsigned int>
+                unsigned int,
+                THierarchy>
             {
                 //-----------------------------------------------------------------------------
                 //
@@ -124,11 +130,13 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             //! The GPU CUDA accelerator atomic operation.
             //-----------------------------------------------------------------------------
-            template<>
+            template<
+                typename THierarchy>
             struct AtomicOp<
                 op::Add,
                 atomic::AtomicCudaBuiltIn,
-                unsigned long long int>
+                unsigned long long int,
+                THierarchy>
             {
                 //-----------------------------------------------------------------------------
                 //
@@ -145,11 +153,13 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             //! The GPU CUDA accelerator atomic operation.
             //-----------------------------------------------------------------------------
-            template<>
+            template<
+                typename THierarchy>
             struct AtomicOp<
                 op::Add,
                 atomic::AtomicCudaBuiltIn,
-                float>
+                float,
+                THierarchy>
             {
                 //-----------------------------------------------------------------------------
                 //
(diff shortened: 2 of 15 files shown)


include/alpaka/core/Positioning.hpp
```diff
@@ -24,6 +24,17 @@
 namespace alpaka
 {
     //-----------------------------------------------------------------------------
+    //! Defines the parallelism hierarchy levels of Alpaka
+    //-----------------------------------------------------------------------------
+    namespace hierarchy
+    {
+        struct Grids{};
```
BenjaminW3: Could you please add the `//#########...` comment above the struct definitions?
I think you could remove the braces as they are optional.

include/alpaka/atomic/AtomicCudaBuiltIn.hpp
```diff
@@ -82,11 +85,12 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             //! The GPU CUDA accelerator atomic operation.
             //-----------------------------------------------------------------------------
-            template<>
+            template<typename THierarchy>
```
BenjaminW3: Could you please add a line break after the `template<`?

include/alpaka/atomic/AtomicCudaBuiltIn.hpp
```diff
@@ -82,11 +85,12 @@ namespace alpaka
...
             struct AtomicOp<
                 op::Add,
                 atomic::AtomicCudaBuiltIn,
-                int>
+                int,
+                THierarchy>
```
BenjaminW3: You will have to do the same for all of the `struct AtomicOp<` in this file to compile correctly.

include/alpaka/atomic/AtomicOmpCritSec.hpp
```diff
@@ -33,6 +33,9 @@ namespace alpaka
     {
         //#############################################################################
         //! The OpenMP accelerator atomic ops.
+        //
+        //  Atomics can used in the hierarchy level blocks and threads.
```
BenjaminW3: Here and in the other classes the following would be more correct:
`Atomics can be used in the blocks and threads hierarchy levels.`

include/alpaka/atomic/AtomicHierarchy.hpp
```diff
@@ -0,0 +1,125 @@
...
+        //! build a single class to inherit from different atomic implementations
+        //
+        //  \tparam TGridAtomic  atomic implementation for atomics in a single grid
+        //  \tparam TBlockAtomic atomic implementation for atomics in a single block
+        //  \tparam TThreadAtomic atomic implementation for atomics in a single thread
+        //#############################################################################
+        template<
+            typename TGridAtomic,
+            typename TBlockAtomic = TGridAtomic,
+            typename TThreadAtomic = TBlockAtomic
```
BenjaminW3: Are those defaults necessary as they are not currently used anywhere?

include/alpaka/acc/AccCpuThreads.hpp
```diff
@@ -108,7 +113,11 @@ namespace alpaka
                     workdiv::WorkDivMembers<TDim, TSize>(workDiv),
                     idx::gb::IdxGbRef<TDim, TSize>(m_gridBlockIdx),
                     idx::bt::IdxBtRefThreadIdMap<TDim, TSize>(m_threadToIndexMap),
-                    atomic::AtomicStlLock(),
+                    atomic::AtomicHierarchy<
+                        atomic::AtomicStlLock,    // atomics between grids
```
BenjaminW3: The alignment is wrong here. Are those comments necessary at all?

include/alpaka/atomic/AtomicHierarchy.hpp
```diff
@@ -0,0 +1,125 @@
...
+            class Empty0{};
+            class Empty1{};
+            class Empty2{};
+        }
+        //#############################################################################
+        //! build a single class to inherit from different atomic implementations
+        //
+        //  \tparam TGridAtomic  atomic implementation for atomics in a single grid
+        //  \tparam TBlockAtomic atomic implementation for atomics in a single block
+        //  \tparam TThreadAtomic atomic implementation for atomics in a single thread
```
BenjaminW3: Is `in a single` correct? Wouldn't `between (Thread|Block|Grid)s within a (Block|Grid|Device)` be better?

include/alpaka/atomic/AtomicHierarchy.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+namespace alpaka
+{
+    namespace atomic
+    {
+        namespace atomicHierarchy
+        {
+            class Empty0{};
+            class Empty1{};
+            class Empty2{};
```
BenjaminW3: I think the braces are not necessary.

BenjaminW3:
This is really cool! This will speed up lots of code!

BenjaminW3:
Does anything argue against making the hierarhcy argument a template parameter as it is in the `getIdx` and `getWorkDiv` calls?
```
atomic::atomicOp<atomic::op::Sub, hierarchy::Grids>(acc, x, 5);
```

include/alpaka/atomic/AtomicStlLock.hpp
```diff
@@ -31,6 +31,9 @@ namespace alpaka
     {
         //#############################################################################
         //! The CPU threads accelerator atomic ops.
+        //
+        //  Atomics can used in the hierarchy level grids, blocks and threads.
+        //  Atomics are not guaranteed to be save between devices.
         //#############################################################################
         class AtomicStlLock
```
BenjaminW3: The friend declaration at line 46 needs to be extended by `typename THierarchy`.

include/alpaka/core/Positioning.hpp
psychocoderHPC: Yes I will add some comments.
The braces were needed because we need to instantiate the structs to give it to the the atomic functions.

include/alpaka/atomic/AtomicHierarchy.hpp
psychocoderHPC: There is no need for the defaults but simplify the interface usage. If you prefer  I can remove the defaults.

include/alpaka/acc/AccCpuThreads.hpp
psychocoderHPC: The comments should help someone who read the code in is not fully familiar with the interfaces

include/alpaka/atomic/AtomicHierarchy.hpp
psychocoderHPC: Maybe `within a Device` suggest that a atomic operation in a `Device` is also atomic to the `Block` atomics but this is not the case. We can only guarantee that atomics in one hierarchy are atomic.

include/alpaka/atomic/AtomicHierarchy.hpp
psychocoderHPC: I can't inherit if this is only a forward definition without a implementation.

psychocoderHPC:
> Does anything argue against making the hierarhcy argument a template parameter as it is in the getIdx > and getWorkDiv calls?

If I add the `hierarchy` as a pure template parameter I can't use the most save atomic placement as default parameter. The result is that I need to change all current codes those uses atomics (The current implementation is backward compatible).

include/alpaka/acc/AccCpuThreads.hpp
erikzenker: At the moment I like the comments, because it explains the hierarchy to me. However, it should be very well documented within the `AtomicHierarchy` class.

include/alpaka/atomic/AtomicHierarchy.hpp
erikzenker: I also think, that we should make this very clear at this point. Thus, there is no confusion. Does it make sense to specify which memory is used for the atomics?

erikzenker:
Very nice work! We should prepare unit tests for the several atomic levels in an additional PR.

include/alpaka/atomic/AtomicHierarchy.hpp
psychocoderHPC: The atomics are not scoped to memory hierarchies, only to parallelism hierarchies.
You can find more information in #201

psychocoderHPC:
> Very nice work! We should prepare unit tests for the several atomic levels in an additional PR.

It is possible to test if all hierarchies not create any compile errors but checking if a atomic is really atomic operation is hard. It is something like checking race conditions.

include/alpaka/atomic/AtomicHierarchy.hpp
BenjaminW3: For now, I would prefer the explicit version.

include/alpaka/atomic/AtomicHierarchy.hpp
psychocoderHPC: @BenjaminW3 Yes you are right your description is better, I misunderstood your comment. Know I am with you^^

psychocoderHPC:
I updated this pull request:
- rebase against the newest `develeop`
- add more documentation to `AtomicHierarchy`
- add `AtomicHierarchy` to the tbb backend
- remove the default from the template signature of `AtomicHierarchy`
- include annotations from github

Issue #80, opened 2015-10-26
remove `alpaka::block::shared::allocArr`


psychocoderHPC:
Currently `ALPAKA` supports two ways to allocate static sized shared memory in a kernel `allocArr` and `allocVar` 
IMO we can remove `allocArr` because the same memory allocation can described with `alpaka::block::shared::allocVar<std::array<int,10>>`.
The problem is that `std::array` is not supported by cuda 7.0. If we ship a light implementation for static sized array with alpaka we can remove the trait for `allocArr`.

a implementation for an array can be:
```C++
template<typename T_Type, size_t T_size>
struct Array
{
    T_Type m_data[T_size];

    template<typename T_Idx>
    ALPAKA_FN_HOST_ACC const T_Type& operator[](const T_Idx idx ) const
    {
        return m_data[idx];
    }

    template<typename T_Idx>
    ALPAKA_FN_HOST_ACC T_Type& operator[](const T_Idx idx )
    {
        return m_data[idx];
    }
};
```

btw with an array we can also define easily 2D arrays:
```C++
alpaka::block::shared::allocVar<std::array<std::array<int,32>,10>>
```

erikzenker:
After thinking a little bit about it, I like the solution :-)

BenjaminW3:
After initial scepticism I am now convinced that this is the correct solution. However, I would prefer a native `std::array` support instead of a self developed version.

psychocoderHPC:
I am also prefer a native std::array but it is not available in cuda. We should implenent it that it looks like a native std::array. If it is available in CUDA we can create a typedef which points to std::array.

BenjaminW3:
Could someone check if `boost::array` is supported by CUDA instead?

psychocoderHPC:
Not tested but it can not supported because the method prefix  __device__ is missing. see http://www.boost.org/doc/libs/1_59_0/doc/html/boost/array.html

PR #229, opened 2016-05-12
Removed allocArr

erikzenker:
This PR fixes issue #80. The free function `alpaka::block::shared::st::allocArr<type, nElements, __COUNTER__>` was simply removed and can now be "emulated" by `alpaka::block::shared::st::allocVar<Array<type, nElements>, __COUNTER__>`. The kind of array type to use is left open to the user.

### DIFF
--- a/test/unit/block/shared/src/BlockSharedMemSt.cpp
+++ b/test/unit/block/shared/src/BlockSharedMemSt.cpp
@@ -39,6 +39,32 @@ BOOST_AUTO_TEST_SUITE(blockSharedMemSt)
 //#############################################################################
 //!
 //#############################################################################
+template<
+    typename TType,
+    size_t TSize>
+struct Array {
+    TType m_data[TSize];
+
+    template<
+        typename T_Idx>
+    ALPAKA_FN_HOST_ACC const TType &operator[](
+        const T_Idx idx) const
+    {
+        return m_data[idx];
+    }
+
+    template<
+        typename TIdx>
+    ALPAKA_FN_HOST_ACC TType & operator[](
+        const TIdx idx)
+    {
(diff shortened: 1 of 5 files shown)


BenjaminW3:
Thanks, this is really cool. The user provided array is by far the best solution right now!

BenjaminW3:
I have merged this even though some of the CI builds are red. This is due to the log length being exceeded because of the extended tests and the new accelerator type. I will have a look at what we can do to further reduce the CI log output.

psychocoderHPC:
This pull request removed the possibility for the user to create an shared memory array.
The current implementation of the new class `Array` is only avail in the test suite.
We need to ship this type with the core of **Alpaka** else each user must create an own array implementation.

psychocoderHPC:
I added #234 to remove all comments with `allocArr`

test/unit/block/shared/src/BlockSharedMemSt.cpp
```diff
@@ -39,6 +39,32 @@ BOOST_AUTO_TEST_SUITE(blockSharedMemSt)
 //#############################################################################
 //!
 //#############################################################################
+template<
+    typename TType,
+    size_t TSize>
+struct Array {
+    TType m_data[TSize];
```
psychocoderHPC: This should be named `data` to be more equeal to a C++11 `std::Array` [link](http://de.cppreference.com/w/cpp/container/array)
This also avoid not nice operations like `&array[0]` to get the pointer of the array.

BenjaminW3:
I do not currently see a huge problem for the user to implement his own array. However, if an array should be added to alpaka then it has to comply with the C++11 `std::array` in as many aspects as possible.

PR #238, opened 2016-05-21
fix the clang CUDA fast-math and show register options

BenjaminW3:
When natively building CUDA code with clang, the fast-math and show-register flags were wrongly written to the `CUDA_NVCC_FLAGS` variable. This resulted in slower code when only building with alpaka, but in compile errors when having native `.cu` files within the same project (as is the case in `matmul`), as the clang flags are unknown to nvcc.


psychocoderHPC:
thx

PR #240, opened 2016-05-28
add support for CUDA 8.0

BenjaminW3:
Everything works out-of-the-box. No changes required.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ compiler:
 # ALPAKA_ACC_CPU_BT_OMP4_ENABLE                 : {ON, OFF}
 #   [ON] OMP_NUM_THREADS                        : {1, 2, 3, 4}
 # ALPAKA_ACC_GPU_CUDA_ENABLE                    : {ON, OFF}
-#   [ON] ALPAKA_CUDA_VER                        : {7.0, 7.5}
+#   [ON] ALPAKA_CUDA_VER                        : {7.0, 7.5, 8.0}
 #   [ON] ALPAKA_CUDA_COMPILER                   : {nvcc, [CXX==clang++]:clang}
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
note.: we can use native double atomic add with the new pascal gpus.

BenjaminW3:
Yes, this could be a further optimization. Do you want to create an issue for this?

BenjaminW3:
This can be merged now, as #257 conflicts with it and has to be rebased before it can be merged next.

.travis.yml
```diff
@@ -191,21 +191,20 @@ before_install:
           && ALPAKA_CUDA_VER_MINOR=${ALPAKA_CUDA_VER:2:1}
           && echo ${ALPAKA_CUDA_VER_MINOR}
           && ALPAKA_CUDA_CACHE_DIR=${HOME}/cache/CUDA/CUDA-${ALPAKA_CUDA_VER}
-          && if [ ${ALPAKA_CUDA_VER_MAJOR} == 7 ]
+          && if (( (( ${ALPAKA_CUDA_VER_MAJOR} == 7 )) && (( ${ALPAKA_CUDA_VER_MINOR} == 0 )) ))
```
psychocoderHPC: Why there are always double braket's `((` instead of `(`

BenjaminW3: I am no bash pro and so I do not know the specific reasons. 😄 It just works. If there is a better syntax, please create a Pull Request.

psychocoderHPC: `if [ ${ALPAKA_CUDA_VER_MAJOR} -eq 7 ] && [ ${ALPAKA_CUDA_VER_MINOR} -eq 0 ]` is the version wish is also `sh` compatible.
But I am fine with `(( ... ))`, I was only confused because I have never seen this keyword before.

More information: http://mywiki.wooledge.org/BashFAQ/031

PR #242, opened 2016-05-30
fix the minimum boost version

BenjaminW3:


erikzenker:
Thx for updating the minimal Boost version

PR #243, opened 2016-06-09
add gcc 6.1 to CI and supported compilers

BenjaminW3:
This adds support for gcc 6.1 (nothing had to be done, works out-of-the-box). It has only been added to the CI.


BenjaminW3:
As has been written in #244, the clang apt repo is offline and therefore all clang builds fail at the moment.

erikzenker:
Thank you for adding the ci build of gcc 6.1

Issue #244, opened 2016-06-09
clang apt-repo offline


BenjaminW3:
The llvm apt-repository has been taken offline due to excessive load until a solution can be found.
So all the clang builds are currently failing.
See [here](http://lists.llvm.org/pipermail/llvm-dev/2016-May/100303.html).
I have temporarily disabled the status check required before a merge, as all our Travis builds will be marked as failed as long as this has not been resolved and this would hinder us from merging anything for an unknown time.

BenjaminW3:
The apt repo has temporarily been moved to [here](http://llvm-apt.ecranbleu.org/apt/). This would enable us to fix the CI temporarily.
I will not be able to test this until July.

erikzenker:
Fixed temporarily by #245

PR #245, opened 2016-06-16
fix clang CI tests

BenjaminW3:
Fixes the urgent CI testing bug #244.

Furthermore, the current boost git repo can not be cloned with --depth 1 as the super-project seems to track a non-HEAD commit of one of the submodules which is not available when --depth 1 is recursively applied.

Additionally, this commit introduces caching of the CMake, clang and CUDA downloads reducing the build times by 1 min, 1 min and 2 min respectively for jobs using those.
Old: Total time 4 hrs 8 min 44 sec
New: Total time 3 hrs 23 min 21 sec
This could additionally be used to cache the boost build later on saving much more time.

The boost.fiber additional download has been removed, as it is now part of boost itself.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -41,11 +41,12 @@ compiler:
     - clang
 
 ################################################################################
-# \TODO: Test full matrix:
+# NOTE: Testing the full matrix is not practical.
+# Therefore we aim to have each value been set in at lest one job.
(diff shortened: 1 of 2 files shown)


BenjaminW3:
Ready to merge!

erikzenker:
Thank you for fixing the clang issue :-)

Issue #230, opened 2016-05-15
boost unit test framework functions can not be used within kernels


BenjaminW3:
The usage of `BOOST_REQUIRE_EQUAL` within kernel functions is problematic.
They print a warning (calling `__host__` function from a `__host__ __device__` function) in the default mode, but will not compile at all in `ALPAKA_ACC_GPU_CUDA_ONLY_MODE`.

There are multiple possible solutions:
- The easiest method would be to replace them by `BOOST_VERIFY`. The whole unit test will then crash at the first error instead of printing all failed expectations. However, this is not a big problem.
- A much harder solution would be to replace the check macros by some method to return a success value back to the host (maybe via global device memory) which can then be checked with `BOOST_REQUIRE_EQUAL` on the host. However, this would additionally require a way to distinguish between all failed checks on the host (maybe via a bitset) to have any advantage over the assert soultion

BenjaminW3:
I will work on this after #218 and #226 are merged.

PR #246, opened 2016-06-20
remove boost test usage from kernels

BenjaminW3:
This solves #230.

Removes the usage of the `__host__` only  `BOOST_REQUIRE_XXX` macros from the test kernels which produce warnings from nvcc and can neither compile in CUDA only mode nor with native clang CUDA support.
It has been replaced by `BOOST_VERIFY` which does not rely on host memory.


erikzenker:
#245 was merge. Therefore rebase can be performed now.

BenjaminW3:
ready to merge!

erikzenker:
Nice workaround, thx for this PR :-)

Issue #248, opened 2016-06-20
Add an OpenACC backend


bussmann:
According to [this paper on the Sunway Taihulight](http://www.netlib.org/utk/people/JackDongarra/PAPERS/sunway-report-2016.pdf), the new Chinese HPC system that topped the TOP 500 at ISC'2016 uses a custom OpenACC implementation. So, in order to use it we would surely need an OpenACC backend. How much work would that be and can work be copied from the OpenMP 4 implementation?

BenjaminW3:
As gcc 6.1 supports OpenACC 2.0a we have already support for a OpenACC capable compiler and do not need to add e.g. the PGI compiler (especially the integration into the CI could be problematic). => Nothing to do
Is there any information about the compiler they use?
I do not have any information and have not searched for it, but I expect it to be a non-offloading many-core CPU. This would allow an implementation very similar to our current OpenMP implementation, as it does not support offloading either, because we directly use host allocated memory within OpenMP regions => OpenMP adaption should be possible.
Is this non-offloading assumption true?

erikzenker:
I saw today updated slides about the software stack supported by the sunway taihulight on the isc. Next to OpenAcc there was also OpenMP and MPI 3.0 mentioned. Nevertheless, the speaker only mentioned porting efforts to OpenAcc.

There were no information about the supported compilers, used operating system or the architecture of the processor yet. But I think such information will follow soon.

ax3l:
Update: According to our latest meeting and information from @juckel, the chinese machines around Sunway Taihulight is moving away from OpenACC :'-S

j-stephan:
If I'm not mistaken we should be able to close this. @jkelling introduced OpenACC support with #1127.

psychocoderHPC:
> If I'm not mistaken we should be able to close this. @jkelling introduced OpenACC support with #1127.

Yes I think so, I will close it.

psychocoderHPC:
implemented with #1127

PR #252, opened 2016-06-26
[WIP] OpenACC back-end

BenjaminW3:
Some first tests for OpenACC support for #248.


theZiz:
So, what is missing right now before we can merge OpenACC?

BenjaminW3:
First of all, there are compile errors e.g. `ExecCpuOpenAcc2.hpp:194:79: error: ‘gridBlockExtent’ referenced in target region does not have a mappable type`.
Secondly, I was unable to find a way to explicitly synchronize threads within a block. If this is not possible, we would have to add the restriction, that the blockThreadExtend can only ever be 1 (similar to the TBB executor).
Last but not least, the current code directly accesses host memory so we can only use this as non-offloading accelerator (similar to the OpenMP4 back-end).

psychocoderHPC:
> Secondly, I was unable to find a way to explicitly synchronize threads within a block. If this is not possible, we would have to add the restriction, that the blockThreadExtend can only ever be 1 (similar to the TBB executor).

To set blockThreadExtend force to one is fine on a CPU but on a GPU

psychocoderHPC:
A short internal discussion with @theZiz pointed out that we should limit for the first kick of of this back-end the number of threads per block to one.

> Last but not least, the current code directly accesses host memory so we can only use this as non-offloading accelerator (similar to the OpenMP4 back-end)

We need to check if we can workaround this, else we will not fulfill our goal to have a single source code.

> First of all, there are compile errors e.g. ExecCpuOpenAcc2.hpp:194:79: error: ‘gridBlockExtent’ referenced in target region does not have a mappable type.

This needs further investigation

BenjaminW3:
CMake 3.10 adds his own `FindOpenACC.cmake`. So the one currently included can be removed.

cmake/modules/FindOpenACC.cmake
```diff
@@ -0,0 +1,104 @@
+#.rst:
```
BenjaminW3: CMake 3.10 has added a FindOpenACC script so this should not be necessary anymore.

ax3l:
From the PIConGPU side it's ok if Alpaka moves to CMake 3.10+ as well.

ax3l:
We are members in the OpenACC org and regularly at user group meetings (e.g. the next at GTC). Can we formulate the two main points,

> Secondly, I was unable to find a way to explicitly synchronize threads within a block.

and

> the current code directly accesses host memory so we can only use this as non-offloading accelerator (similar to the OpenMP4 back-end).

as a proposal to the [OpenACC specification](https://www.openacc.org/resources)? What would you like to have added? I think we did let these two points to long at rest without actively proposing changes we need to the OpenACC community and should do this.

cmake/modules/FindOpenACC.cmake
ax3l: From the PIConGPU side it's ok if Alpaka moves to CMake 3.10+ as well.

ax3l:
> the current code directly accesses host memory so we can only use this as non-offloading accelerator (similar to the OpenMP4 back-end).

@BenjaminW3 is this a limitation of OpenACC or just not implemented in the PR?

BenjaminW3:
@ax3l I do not remember anymore. It is too long ago.
If OpenACC has a method to allocate memory on a device from host code, it should be possible. I do not know if this exists.

include/alpaka/acc/AccCpuOpenAcc2.hpp
```diff
@@ -0,0 +1,307 @@
+/**
+* \file
+* Copyright 2018 Benjamin Worpitz
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
```
ax3l, line 7: With regards to #637, this new file's header will need an update :)

include/alpaka/atomic/AtomicOpenAcc.hpp
```diff
@@ -0,0 +1,371 @@
+/**
+* \file
+* Copyright 2018 Benjamin Worpitz
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
```
ax3l, line 7: With regards to #637, this new file's header will need an update :)

include/alpaka/block/shared/dyn/BlockSharedMemDynRef.hpp
```diff
@@ -0,0 +1,89 @@
+/**
+* \file
+* Copyright 2018 Benjamin Worpitz
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
```
ax3l, line 7: With regards to #637, this new file's header will need an update :)

include/alpaka/exec/ExecCpuOpenAcc2.hpp
```diff
@@ -0,0 +1,337 @@
+/**
+* \file
+* Copyright 2018 Benjamin Worpitz
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
```
ax3l, line 7: With regards to #637, this new file's header will need an update :)

include/alpaka/idx/bt/IdxBtRef.hpp
```diff
@@ -0,0 +1,129 @@
+/**
+* \file
+* Copyright 2018 Benjamin Worpitz
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
```
ax3l, line 7: With regards to #637, this new file's header will need an update :)

cmake/modules/FindOpenACC.cmake
ax3l: With regards to #637, this new file will need an MPL-2.0 header :)

Issue #67, opened 2015-10-04
fix random number generators in CUDA


BenjaminW3:
When calling the `__device__ operator()` of the random number distributions from a `__host__ __device__` kernel function, nvcc errors.
In the same way all other methods are implemented, there should be a function forwarding to a trait that can be specialized per implementation.

PR #255, opened 2016-06-30
add default constructor to `Xor` RNG

psychocoderHPC:
add default constructor

This change allows that Xor can used in an `std::Array` or plain array.

Not allowed before`
```C++
Xor myArray[5]; // call the default constructor which not exists
for(int i=0; i<5; ++i)
    myArray[i] = Xor(i);
```

With this commit the example is allowed. The `RandStl` supported the example without changes because there is a default constructor avail.
This feature is enhancement is needed to use the random number generator inside the element level.

### DIFF
--- a/include/alpaka/rand/RandCuRand.hpp
+++ b/include/alpaka/rand/RandCuRand.hpp
@@ -65,6 +65,15 @@ namespace alpaka
                 class Xor
                 {
                 public:
+
+                    //-----------------------------------------------------------------------------
+                    //! Constructor.
+                    //
+                    // After calling this constructor the instance is not valid initialized and
+                    // need to be overwritten with a valid object
+                    //-----------------------------------------------------------------------------
+                    ALPAKA_FN_ACC_CUDA_ONLY Xor() = default;
+
                     //-----------------------------------------------------------------------------
                     //! Constructor.
                     //-----------------------------------------------------------------------------


BenjaminW3:
Thanks for fixing this!
Do the RNGs work at all? In #67 there is an issue documented that this might not be the case. Do you compile in CUDA only and CPU only or mixed mode?

psychocoderHPC:
I trait both CUDA only and CPU only. Thats the reason why I found the "missing" default constructor behavior.

psychocoderHPC:
@BenjaminW3: Could you see why the clang compile with travis crashed? Maybe it is an travis issue?

psychocoderHPC:
OK I think the verbosity is to high we need to fix first our travis tests. :-(

BenjaminW3:
The clang build 1660.13 failed because it could not install git. This have most probably been network issues.
The build 1660.8 failed with
```
./memView
Running 444 test cases...
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_S_create
unknown location(0): fatal error: in "viewStaticAccMem/staticDeviceMemoryGlobal<N6alpaka3acc10AccCpuOmp4ISt17integral_constantImLm2EEjEE>": signal: SIGABRT (application abort requested)
/home/travis/build/ComputationalRadiationPhysics/alpaka/test/unit/mem/view/src/ViewStaticAccMem.cpp(75): last checkpoint

*** 1 failure is detected in the test module "memView"
```
I have not seen this before. May be out of memory.

BenjaminW3:
I have restarted the faild jobs.

BenjaminW3:
I have merged it as the restarted jobs succeeded.
We should have an eye on the failing clang test if this happens again.

@psychocoderHPC Do you have a small test case? Could you please try to compile your code in mixed mode with both a CPU and a GPU accelerator enabled if possible? It could be that #67 is only a problem in this setup. If there is no problem at all, this ticket could be closed.

PR #258, opened 2016-07-06
enable Werror in CI

BenjaminW3:
This enables -Werror for clang, gcc and MSVC in the CI.

### DIFF
--- a/test/analysis/headerCheck/src/main.cpp
+++ b/test/analysis/headerCheck/src/main.cpp
@@ -20,7 +20,16 @@
  */
 
 #define BOOST_TEST_MODULE headerCheck
+
+#include <boost/predef.h>   // BOOST_COMP_CLANG
+#if BOOST_COMP_CLANG
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wunused-parameter"
+#endif
 #include <boost/test/unit_test.hpp>
+#if BOOST_COMP_CLANG
+    #pragma clang diagnostic pop
+#endif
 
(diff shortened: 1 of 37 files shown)


BenjaminW3:
Ready to merge!

test/analysis/headerCheck/src/main.cpp
```diff
@@ -20,7 +20,16 @@
...
 #define BOOST_TEST_MODULE headerCheck
+
+#include <boost/predef.h>   // BOOST_COMP_CLANG
+#if BOOST_COMP_CLANG
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wunused-parameter"
+#endif
 #include <boost/test/unit_test.hpp>
+#if BOOST_COMP_CLANG
+    #pragma clang diagnostic pop
```
psychocoderHPC: Is there a need to disable all this pragma's. 
I know that unknown pragmas result in a warning but maybe we should disable the warning for unknown pragmas.

BenjaminW3: Most of the time the warning about unknown pragmas is not useful. However, in the case of alpaka that defines many macros containing pragmas, the warning most probably shows a bug within alpaka. Therefore I would not want to fully disable it.

psychocoderHPC: ok I am fine with it

BenjaminW3: Is there anything else to do before it can be merged?

psychocoderHPC: merged

Issue #261, opened 2016-07-08
CI tests are flaky


BenjaminW3:
Some tests hang randomly. This seems to be the case since `BOOST_REQUIRE_EQ` has been replaced by `BOOST_VERIFY` in #246. The origin of the hangs has to be researched as it happens on both gcc and clang in different jobs randomly.

BenjaminW3:
As all tests succeeded (without any manual reexecutions) since #263, I will close this ticket.

PR #263, opened 2016-07-09
fix flaky tests

BenjaminW3:
This hopefully fixes the flaky tests (#261) by removing an invalid `BOOST_REQUIRE_XXX` call from within a test kernel and fixing the non-deterministic time test.


BenjaminW3:
This should be merged before the release.

PR #265, opened 2016-07-11
use Weverything for clang

BenjaminW3:
This enables `-Weverything` for clang which includes `-Wall` and all other warnings that clang can provide (only within CI builds).
Most warnings have been fixed (member order for better packing, explicit casts, invalid documentation) or disabled locally.
Only 3 warnings had to be disabled globally.
Furthermore, warnings within boost had to be silenced.


bussmann:
Yeah!

BenjaminW3:
Ready!

psychocoderHPC:
I review it next week.

psychocoderHPC:
thanks for removing all the implicit cast and adding more comments!

PR #272, opened 2016-07-23
move Vec to vec:: and enhance documentation

BenjaminW3:
enhance documentation by:
* adding documentation
* moving Vec and Vec specific functions to `detail` and `vec` namespaces


BenjaminW3:
push

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -81,10 +81,11 @@ static auto createVecFromIndexedFn()
 -> alpaka::Vec<TDim, TSize>
 {
     return
+        alpaka::
```
psychocoderHPC: IMO this is very hard to read if the code is spitted by to many `#ifdef`

Could we change it back to the old style and only add `vec::` to the secound `#ifdef` part
```C++
  #ifdef ALPAKA_CREATE_VEC_IN_CLASS
         alpaka::Vec<TDim, TSize>::template
  #else
         vec::alpaka::
  #endif
```

test/unit/mem/view/src/ViewSubViewTest.cpp
```diff
@@ -106,10 +106,11 @@ static auto createVecFromIndexedFn()
 -> alpaka::Vec<TDim, TSize>
 {
     return
+        alpaka::
```
psychocoderHPC: Could we change it back and add only `vec::` to the second if body

include/alpaka/vec/Vec.hpp
```diff
@@ -988,7 +1014,9 @@ namespace alpaka
             using IdxOffset = std::integral_constant<std::intmax_t, static_cast<std::intmax_t>(dim::Dim<TExtent>::value) - static_cast<std::intmax_t>(TDim::value)>;
             return
 #ifdef ALPAKA_CREATE_VEC_IN_CLASS
-            Vec<TDim, size::Size<TExtent>>::template
+                Vec<TDim, size::Size<TExtent>>::template
```
psychocoderHPC: Could we move `alpaka::Vec<>` also to the namepsace `alpaka::vec`.
If we move `alpaka::Vec<>` to the namespace `alpaka::vec::` than we avoid a lot of the not so nice precompiler `#ifdef`'s

psychocoderHPC:
Thanks for this pull request. I think we should move `alpaka::Vec` to the correct namespace to avoid to much `#ifdef`'s in the code.

include/alpaka/vec/Vec.hpp
BenjaminW3: This would be my preferred solution but this will require changes to existing code using alpaka.

BenjaminW3:
I will do it.

include/alpaka/vec/Vec.hpp
psychocoderHPC: We should do this change now even if it need a change in the user code. This is only a change in the current dev which goes to the next release. We should add a no with interface changes to the release notes of the next release.

BenjaminW3:
Done.

PR #274, opened 2016-08-10
fix warning unused variable

psychocoderHPC:
fix `warning: variable "iBlockThreadCount" was declared but never referenced`

This warning is triggered if no OpenMP accelerator is used but enabled in `cmake`


BenjaminW3:
Thanks!

PR #275, opened 2016-08-10
fix warning call `__host__` function

psychocoderHPC:
fix `warning: calling a __host__ function from a __host__ __device__ function is not allowed`


BenjaminW3:
Thanks for fixing this!

Issue #277, opened 2016-09-06
Check if CMake 3.7.0 allows to simplify build scripts (when released)


BenjaminW3:
With CMake 3.7.0, the FindCUDA script will be able to get the `TARGET_PROPERTY`(s) `INCLUDE_DIRECTORIES` and `COMPILE_DEFINITIONS` directly from the target linked with `TARGET_LINK_LIBRARIES` see [here](https://gitlab.kitware.com/cmake/cmake/issues/14201). This has long been possible with normal targets, but FindCUDA did not support it. 
This should allow us to remove all `INCLUDE_DIRECTORIES` and `ADD_DEFINITIONS` in the tests and only use `TARGET_INCLUDE_DRECTORIES` and `TARGET_COMPILE_DEFINITIONS` for target local properties.

PR #278, opened 2016-09-11
CMake 3.7.0 support

BenjaminW3:
This adds CMake 3.7.0 to the tested versions and uses TARGET_INCLUDE_DIRECTORIES and TARGET_COMPILE_OPTIONS even for CUDA targets (#277).
For the moment this results in more complex CMake scripts as both the old and the new versions have to be supported but in a distant future when we only have to support CMake 3.7.0+ this will make things easier 😃


erikzenker:
What are the reasons to use these Variables on newer cmake versions ? just interested

BenjaminW3:
The old `INCLUDE_DIRECTORIES` etc. added those includes to all targets (libs/executables) defined within the current or contained directories. This is most probably not what you want. By using `TARGET_INCLUDE_DIRECTORIES` you can add those headers only for a specific target and specify if those headers are transitively added to all targets linking to this one or if they should be for private use only. The target versions should generally always be preferred over the global versions in all cases where it is supported (CUDA_ADD_XXX did not support it before CMake-3.7.0).

erikzenker:
thx

test/integ/mandelbrot/CMakeLists.txt
```diff
@@ -51,11 +51,12 @@ SET(_SOURCE_DIR "src/")
 # Add all the source files in all recursive subdirectories and group them accordingly.
 append_recursive_files_add_to_src_group("${_SOURCE_DIR}" "${_SOURCE_DIR}" "cpp" _FILES_SOURCE)
 
-INCLUDE_DIRECTORIES(
-    ${_INCLUDE_DIRECTORIES_PRIVATE}
```
erikzenker: There are a lot of there `*_PRIVATE` variables disappearing. They are not needed anymore ?

BenjaminW3: No, they are not needed anymore.

PR #280, opened 2016-09-15
remove std::err messages before exceptions

psychocoderHPC:
Disable error message before exceptions if debug mode of alpaka is not activated.

If something inside alpaka goes wrong an exception is triggered. 
The exception can be handled inside the user code. 
The problem is that in some cases a error message is written directly to `std::err` even if the user catch the exception.

This pull request removes the error message before an exception if the debug mode for alpaka is disabled.

### DIFF
--- a/include/alpaka/core/Cuda.hpp
+++ b/include/alpaka/core/Cuda.hpp
@@ -107,7 +107,9 @@ namespace alpaka
                     if(std::find(aIgnoredErrorCodes.cbegin(), aIgnoredErrorCodes.cend(), error) == aIgnoredErrorCodes.cend())
                     {
                         std::string const sError(std::string(file) + "(" + std::to_string(line) + ") '" + std::string(cmd) + "' returned error: '" + std::string(cudaGetErrorString(error)) + "'!");
+#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
                         std::cerr << sError << std::endl;
+#endif
                         ALPAKA_DEBUG_BREAK;
                         throw std::runtime_error(sError);
                     }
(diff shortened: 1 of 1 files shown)


include/alpaka/core/Cuda.hpp
```diff
@@ -177,13 +181,15 @@ namespace alpaka
...
-            }
-        }
-    }
+#endif
+                  ALPAKA_DEBUG_BREAK;
+                  throw std::runtime_error(sError);
+              }
+           }
+       }
+   }
```
BenjaminW3: The indendation seems to be wrong in those lines as there are no real changes ;-)

include/alpaka/core/Cuda.hpp
```diff
@@ -177,13 +181,15 @@ namespace alpaka
                 if(error != CUDA_SUCCESS)
                 {
                     std::string const sError(std::to_string(file) + "(" + std::to_string(line) + ") '" + std::to_string(cmd) + "' returned error: '" + std::to_string(error) + "' (possibly from a previous CUDA call)!");
+#if ALPAKA_DEBUG != ALPAKA_DEBUG_DISABLED
```
BenjaminW3: In all other places we currently use `ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL`. I would prefer this for consistency and because it makes it easier to increase the level of this debug message later.

include/alpaka/core/Cuda.hpp
psychocoderHPC: This indention was wrong before. I fixed that the brackets are now well aligned but not 4 spaces indented.
Should I remove the changes and keep the wrong indention or should I fix it fully?

BenjaminW3: Please fix it fully ;-)

psychocoderHPC:
I will add the changes next week, currently I am sick.

psychocoderHPC:
I debug guard to `ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL`

PR #269, opened 2016-07-21
add TBB to the windows CI tests

BenjaminW3:
Appveyor has changed its build farm and the build times have been reduced to 1/4. This allows us to extend the tests.

PR #281, opened 2016-09-26
fix broken `alpakaConfig.cmake`

psychocoderHPC:
move debug output of common variables after debug level definition

Error message before this pull request:
```
CMake Error at /bigdata/hplsim/scratch/widera/cupla/alpaka/alpakaConfig.cmake:88 (IF):
  if given arguments:

    "GREATER" "1"

  Unknown arguments specified
```

This bug was introduced with #269


psychocoderHPC:
The reason for this error is that `ALPAKA_DEBUG` was used before it was defined.
Tested with: cmake/3.3.0

PR #283, opened 2016-10-07
Lambda kernel example

erikzenker:
Shows how to use lambda function, std::function, and std::bind to construct an alpaka kernel.
It needs to be tested if this works for CUDA 8.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -728,8 +728,8 @@ script:
     # Build and execute all examples.
     - if [ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" ] ;then ./travis/compileExec.sh "example/bufferCopy" ./bufferCopy ;fi
     - if [ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" ] ;then ./travis/compileExec.sh "example/helloWorld" ./helloWorld ;fi
+    - if ([ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" ]) && (([ ${ALPAKA_CUDA_VER_MAJOR} -ge 8 ]) || ([ "${ALPAKA_CUDA_VER_MAJOR}" == "7" -a  "${ALPAKA_CUDA_VER_MINOR}" == "5" ]));then ./travis/compileExec.sh "example/helloWorldLambda" ./helloWorldLambda ;fi
     - if [ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" ] ;then ./travis/compileExec.sh "example/vectorAdd" ./vectorAdd ;fi
-
 
 ################################################################################
 #
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -333,6 +333,10 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                     LIST(APPEND CUDA_NVCC_FLAGS "-Wno-deprecated-gpu-targets")
                 ENDIF()
 
+                IF(NOT ALPAKA_CUDA_VERSION VERSION_LESS 7.5)
+                    LIST(APPEND CUDA_NVCC_FLAGS "--expt-extended-lambda")
+                ENDIF()
+
                 LIST(APPEND CUDA_NVCC_FLAGS "-arch=${ALPAKA_CUDA_ARCH}")
 
                 IF(NOT MSVC)
--- a/example/helloWorldLambda/src/helloWorldLambda.cpp
+++ b/example/helloWorldLambda/src/helloWorldLambda.cpp
@@ -1,0 +1,214 @@
+/**
+ * \file
+ * Copyright 2014-2016 Erik Zenker
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * alpaka is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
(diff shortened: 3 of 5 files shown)


BenjaminW3:
/home/travis/build/ComputationalRadiationPhysics/alpaka/example/helloWorldLambda/src/helloWorldLambda.cpp:136:71: error: 
      declaration shadows a local variable [-Werror,-Wshadow]
  ...ALPAKA_FN_ACC (Acc & acc, size_t const nExclamationMarks) -> void {
/home/travis/build/ComputationalRadiationPhysics/alpaka/example/helloWorldLambda/src/helloWorldLambda.cpp:131:18: note: 
      previous declaration is here
    const size_t nExclamationMarks = 10;

BenjaminW3:
Aaaahhh... all GCC builds were green, even for CUDA 7.0. There was an error in the script executing CMake in the CI so the CUDA version was not set and the CUDA back-end was simply disabled. I am fixing this currently.

erikzenker:
There is also missing the nvcc compiler flag `--expt-extended-lambda`. Where is the best place to put it ? Maybe [here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/alpakaConfig.cmake#L332) ?

BenjaminW3:
Is this this flag required in CUDA 8.0? This flag should b conditionally be aded for CUDA 7.5+ then.

erikzenker:
At least on my arch linux system I need to use this flag:
```
nvcc --version
Copyright (c) 2005-2016 NVIDIA Corporation
Built on Sun_Sep__4_22:14:01_CDT_2016
Cuda compilation tools, release 8.0, V8.0.44
```
error: __host__ or __device__ annotation on lambda requires --expt-extended-lambda nvcc flag
``

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,213 @@
...
+     * declaration might be useful when small kernels
+     * are written for testing or lambda functions
+     * allready exist.
+     */
+    const size_t nExclamationMarks = 10;
+    const size_t exclamationMarksMultiplier = 3;    
+    
+    auto const helloWorld(alpaka::exec::create<Acc>(
+        workdiv,
+	[&exclamationMarksMultiplier] ALPAKA_FN_ACC (Acc & acc, size_t const nExclamationMarks) -> void {
```
erikzenker: error: An extended __host__ __device__ lambda cannot capture variables by reference

BenjaminW3: I think that this is correct, as the reference to a host variable is not valid nor accessible on the device. You can only capture by copy (if it is allowed at all).

psychocoderHPC: Yes only copy by value or a this pointer is allowed. As I know in the case of a this pointer the full class is copied to the device.

erikzenker:
I addressed the code review issues.

alpakaConfig.cmake
```diff
@@ -329,11 +329,12 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 SET(CUDA_PROPAGATE_HOST_FLAGS ON)
 
+
                 IF(ALPAKA_CUDA_VERSION VERSION_EQUAL 8.0)
                     LIST(APPEND CUDA_NVCC_FLAGS "-Wno-deprecated-gpu-targets")
                 ENDIF()
 
-                LIST(APPEND CUDA_NVCC_FLAGS "-arch=${ALPAKA_CUDA_ARCH}")
+                LIST(APPEND CUDA_NVCC_FLAGS "-arch=${ALPAKA_CUDA_ARCH} --expt-extended-lambda")
```
BenjaminW3: Please add this flag in an extra line and only 'IF(NOT ALPAKA_CUDA_VERSION VERSION_LESS 7.5)'

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,215 @@
...
+	    auto linearizedGlobalThreadIdx = alpaka::idx::mapIdx<1u>(globalThreadIdx,
+								      globalThreadExtent);
+                                                          
+	    printf("[z:%u, y:%u, x:%u][linear:%u] Hello world from a lambda",
+		   static_cast<unsigned>(globalThreadIdx[0]),
+		   static_cast<unsigned>(globalThreadIdx[1]),
+		   static_cast<unsigned>(globalThreadIdx[2]),
+		   static_cast<unsigned>(linearizedGlobalThreadIdx[0]));
+
+	    for(size_t i = 0; i < nExclamationMarks * exclamationMarksMultiplier; ++i){
```
BenjaminW3: This should be nExclamationMarksAsArg

.travis.yml
```diff
@@ -728,6 +728,7 @@ script:
     # Build and execute all examples.
     - if [ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" ] ;then ./travis/compileExec.sh "example/bufferCopy" ./bufferCopy ;fi
     - if [ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" ] ;then ./travis/compileExec.sh "example/helloWorld" ./helloWorld ;fi
+    - if [ "${ALPAKA_CI_ANALYSIS}" == "OFF" -a "${ALPAKA_ACC_GPU_CUDA_ONLY_MODE}" == "OFF" -a ${ALPAKA_CUDA_VER_MAJOR} > 7 -a ${ALPAKA_CUDA_VER_MINOR} > 0] ;then ./travis/compileExec.sh "example/helloWorldLambda" ./helloWorldLambda ;fi
```
BenjaminW3: This is not correct because it excludes cuda 8.0. (maj>7) || ((maj == 7) && (min >5)) should be better.

BenjaminW3: I am no bash expert but the syntax is not supported: The builds say `/home/travis/build.sh: line 45: [: missing `]'` but did not get red (I do not know why).
You may need to use double parenthesis like I did in other places to make the arithmetic comparison work.

erikzenker: There is a space missing in front of `]` ...I try to fix this in the next days.

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -0,0 +1,215 @@
...
+     * to wrap allready existing code into a
+     * std::function and provide it to the alpaka 
+     * library.
+     */
+    auto const hiWorld (alpaka::exec::create<Acc> (
+        workdiv,
+        std::function<void(Acc&, size_t)>( hiWorldFunction<Acc> ),
+        nExclamationMarks));
+
+
```
BenjaminW3: please remove this additional newline

BenjaminW3:
Looks good! Please squash all the commits when everything is green.

PR #285, opened 2016-10-29
Add support for CC 6.x double atomicAdd

BenjaminW3:


include/alpaka/exec/ExecGpuCudaRt.hpp
```diff
@@ -114,8 +114,6 @@ namespace alpaka
 #if (!__GLIBCXX__) // libstdc++ even for gcc-4.9 does not support std::is_trivially_copyable.
             static_assert(
                 meta::Conjunction<
-                    // This true_ is required for the zero argument case because and_ requires at least two arguments.
```
psychocoderHPC: Is this change related to the atomicAdd support for sm_60?

BenjaminW3: Not really. It just removes some useless lines.

psychocoderHPC: Could you please move this change to a separate pull request. It removes a argument from a class and we should have a documentaion why we now allowed to remove it.

BenjaminW3: I will move it. It is not necessary because there is already at least one argument. The removal of the true_type has been forgotten long ago.

PR #287, opened 2016-10-30
extend the list of warnings enabled in CI for gcc

### LINKED ISSUE #262: increase the number of warnings enabled in CI
See for example [here](https://github.com/kaidokert/cpp-travis-test/blob/master/compiler-flags.mk)
The clang side is handled in #265.

BenjaminW3:
This enables most of all available GCC warnings for the CI builds.
This closes #262.

### DIFF
--- a/cmake/dev.cmake
+++ b/cmake/dev.cmake
@@ -38,6 +38,78 @@ ELSE()
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wextra")
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-pedantic")
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Werror")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wdouble-promotion")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wmissing-include-dirs")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wunknown-pragmas")
+        # Higher levels (max is 5) produce some strange warnings
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wstrict-overflow=2")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wtrampolines")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wfloat-equal")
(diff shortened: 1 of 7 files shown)


cmake/dev.cmake
```diff
@@ -38,6 +38,78 @@ ELSE()
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wextra")
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-pedantic")
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Werror")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wdouble-promotion")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wmissing-include-dirs")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wunknown-pragmas")
+        # Higher levels (max is 5) produce some strange warnings
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wstrict-overflow=2")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wtrampolines")
+        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wfloat-equal")
```
psychocoderHPC: This flag should not enabled by default in alpaka. alpaka uses no floats where is flag is useful.
This means if the user need to fulfill this strict rule the flag can be enabled by the user.

ax3l:
@BenjaminW3 is it ok if you do not self-merge your PRs? :)
That would avoid over-looking open questions like the one above from @psychocoderHPC :)

cmake/dev.cmake
BenjaminW3: Yes, those flags are only used for the tests and examples and are not provided to alpaka users.

psychocoderHPC: Thx, it was not obvious that `ALPAKA_DEV_COMPILE_OPTIONS` is only used for the test cases.

BenjaminW3:
As it was already approved, I do not see a reason to not merge the PullRequest myself. To my understanding the approval system is meant exactly to separate the reviewer(s) from the merger. Therefore i think it is ok to self-merge if a PullRequest is approved.
I saw the question but had no time to answer until now.

ax3l:
yes makes perfect sense, @psychocoderHPC still has to get used to it. We just checked out the new options again :)

If you guys agree, you can also just make "requires approved review" a requirement for merge the same as "pass CI" in the setting and then the person who merges does not matter any more :)

PR #289, opened 2016-10-30
add unit tests for (generic) (variadic) lambda kernels

### LINKED ISSUE #47: enable usage of lambdas as kernels

BenjaminW3:
This will close #47.

### DIFF
--- a/test/unit/kernel/src/KernelLambda.cpp
+++ b/test/unit/kernel/src/KernelLambda.cpp
@@ -1,0 +1,216 @@
+/**
+ * \file
+ * Copyright 2015 Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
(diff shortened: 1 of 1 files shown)


erikzenker:
Looks nice, I like that you already test generic lambdas if they are available on C++14 capable compilers. When the ci approves the commit, I would merge.

BenjaminW3:
all green now!

PR #300, opened 2017-01-03
update dependencies in CI

BenjaminW3:
Raises the CMake versions used.
Adds the new boost-1.63.0 to the CI which produces new warnings with MSVC with boost.fiber.


psychocoderHPC:
thx for this update

Issue #159, opened 2016-02-01
alpaka::mem::view::copy for n-dimensional buffers


erikzenker:
Currently buffer copies are only defined for up to three dimensions. But coming unit tests will use at least four dimensional buffers. Therefore, we need to add support for n-dimensional copy operations.

BenjaminW3:
#290 implements full n-dimensional support for all missing CPU operations (set and copy). N-Dimensional CPU execution support was already there.

Still missing is the support for more then 3 dimensions for CUDA kernel execution (`ExecGpuCudaRt::Enqueue`, `AccGpuCudaRt::getAccDevProps`, `IdxBtCudaBuiltIn`, `IdxGbCudaBuiltIn`) as well as all cuda operations (`BufCudaRt::copy`, `BufCudaRt::set`, `BufCudaRt::alloc`) and maybe more.

PR #307, opened 2017-01-09
add viewTestMutable

BenjaminW3:
This adds the generic test function `viewTestMutable` which tests the operation `set` and the `copy` operation on the given view/buffer. The view/buffer contents are checked for correctness with a kernel.

* adds static asserts for dim<=3 for CPU copy and set
* disables 4D tests even for CPU accelerators as long as it is not fully supported

Prework for n-dimensional copy/set support (See #159 and #290).


BenjaminW3:
Ready!

PR #308, opened 2017-01-19
add more Vec functions

BenjaminW3:
* adds `alpaka::vec::concat` trait and implements it for `alpaka::vec::Vec`.
* adds `alpaka::vec::Vec<Dim, bool> alpaka::vec::Vec<Dim, Size>::operator<=`
* add specialization for reversing of one-dimensional Vec
* add specialization for casting Vec when Src and Dst types are equal

Prework for n-dimensional copy/set support (See #159 and #290).

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -1,23 +1,23 @@
 /**
-* \file
-* Copyright 2014-2015 Benjamin Worpitz
-*
-* This file is part of alpaka.
-*
-* alpaka is free software: you can redistribute it and/or modify
-* it under the terms of the GNU Lesser General Public License as published by
-* the Free Software Foundation, either version 3 of the License, or
-* (at your option) any later version.
-*
-* alpaka is distributed in the hope that it will be useful,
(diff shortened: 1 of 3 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -927,6 +1000,83 @@ namespace alpaka
...
+                Vec<TDimR, TSize>>
+            {
+                ALPAKA_NO_HOST_ACC_WARNING
+                ALPAKA_FN_HOST_ACC static auto concat(
+                    Vec<TDimL, TSize> const & vecL,
+                    Vec<TDimR, TSize> const & vecR)
+                -> Vec<dim::DimInt<TDimL::value + TDimR::value>, TSize>
+                {
+                    return
+#ifdef ALPAKA_CREATE_VEC_IN_CLASS
```
psychocoderHPC: only a note: we should check if we can skip this `#ifdef` asap, it is very hard to read this little peace of code.

psychocoderHPC:
thx for the new functions

PR #316, opened 2017-01-28
add syncBlockThreads test

BenjaminW3:
Adds a test that verifies that the current syncBlockThreads implementation works for all accelerators.


BenjaminW3:
Push!

psychocoderHPC:
What do you mean with `... implementation works for all accelerators`? That is is possible to use `syncBlockThreads` or is there any verification that it relay works. 
For a verification we need to trigger some load imbalances (e.g. with sleep or random busy loop) to hope that we get a race condition if some of the threads not wait on synchronize border.

BenjaminW3:
You are right, I could enhance the test and defer some of the threads intentionally.
I simply hoped that the scheduling variances will trigger such errors nevertheless. due to the very short run-time until the syncBlockThreads.

psychocoderHPC:
Should I merge the current state and you add this in a follow up?

BenjaminW3:
I will add the enhancements in a follow up.

Issue #13, opened 2014-11-26
add syncBlockThreadsPredicate template


BenjaminW3:
to support predicates:

int __syncthreads_count(int predicate)
evaluates predicate for all threads of the block and returns the number of threads for which predicate evaluates to non-zero.

int __syncthreads_and(int predicate)
 evaluates predicate for all threads of the block and returns non-zero if and only if predicate evaluates to non-zero for all of them.

int __syncthreads_or(int predicate)
evaluates predicate for all threads of the block and returns non-zero if and only if predicate evaluates to non-zero for any of them.

PR #317, opened 2017-01-28
add syncBlockThreadsPredicate (count, and, or)

BenjaminW3:
This implements #13.

In the current version, this adds some overhead to some accelerators in form of additional storage for member variables required per block even if the `syncBlockThreadsPredicate` method is never called.

* For `AccCpuFibers`, there are currently 5 additional variables per block which could be reduced to 2 (with some amount of work).
* For `AccCpuThreads`, there are currently 7 additional variables per block.
* For `AccCpuOmpThreads`, there are currently 2 additional int32 and 1 uint8 per block. I do not see a way to optimize this further.

### DIFF
--- a/include/alpaka/block/sync/BlockSyncBarrierFiber.hpp
+++ b/include/alpaka/block/sync/BlockSyncBarrierFiber.hpp
@@ -53,7 +53,8 @@ namespace alpaka
                 //-----------------------------------------------------------------------------
                 ALPAKA_FN_ACC_NO_CUDA BlockSyncBarrierFiber(
                     TSize const & blockThreadCount) :
-                        m_barrier(static_cast<std::size_t>(blockThreadCount))
+                        m_barrier(static_cast<std::size_t>(blockThreadCount)),
+                        m_threadCount(blockThreadCount)
                 {}
                 //-----------------------------------------------------------------------------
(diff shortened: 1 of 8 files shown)


include/alpaka/block/sync/BlockSyncBarrierFiber.hpp
```diff
@@ -99,6 +105,52 @@ namespace alpaka
...
+                    //-----------------------------------------------------------------------------
+                    ALPAKA_NO_HOST_ACC_WARNING
+                    ALPAKA_FN_ACC static auto syncBlockThreadsPredicate(
+                        block::sync::BlockSyncBarrierFiber<TSize> const & blockSync,
+                        int predicate)
+                    -> int
+                    {
+                        if(blockSync.m_curThreadCount == blockSync.m_threadCount)
+                        {
+                            blockSync.m_curThreadCount == 0;
```
BenjaminW3: Visual Studio 2015 is the only!!! compiler that has found this bug!

BenjaminW3:
Open for comments!

psychocoderHPC:
I need some time to check this

psychocoderHPC:
I hope that the impact of the new variable is not to large, never the less I will merge it. Big thanks!

PR #318, opened 2017-01-30
fix shared memory warnings

psychocoderHPC:
- fix warning (CUDA8) that shared memory objects are not allowed to have a constructor
- remove old documentation (not valid any more)
- add documentation to the trait `allocVar`

This pull request is equal to the PIConGPU shared memory allocation change [PIConGPU#1726](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1726) and [PIConGPU#1767](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1767)

### DIFF
--- a/include/alpaka/block/shared/st/BlockSharedMemStCudaBuiltIn.hpp
+++ b/include/alpaka/block/shared/st/BlockSharedMemStCudaBuiltIn.hpp
@@ -32,6 +32,7 @@
 #include <alpaka/block/shared/st/Traits.hpp>// AllocVar
 
 #include <type_traits>                      // std::is_trivially_default_constructible, std::is_trivially_destructible
+#include <cstdint>                          // uint8_t
 
 namespace alpaka
 {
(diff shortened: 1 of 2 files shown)


BenjaminW3:
Thanks for fixing this!

We should add some documentation to the `allocVar` method saying something like `The allocated object is uninitialized and not default constructed!`

psychocoderHPC:
Yes the documentation should be part of the free trait function `allocVar` in `Traits.hpp`

BenjaminW3:
Could you do this directly within this Pull Request?

psychocoderHPC:
If something in the CI fails than I will add the documentation to this pull request else I opened a new one.

BenjaminW3:
Everything compiles correctly except the version where clang natively compiles CUDA:
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/block/shared/st/BlockSharedMemStCudaBuiltIn.hpp:99:40: error: 
      'alignas' attribute cannot be applied to types
                            __shared__ alignas(alignof(T)) uint8_t shMem...
                                       ^
```

psychocoderHPC:
[This stackoverflow](http://stackoverflow.com/a/15912208) post solved the issue.

well intuitive solution ;-)
------------------------
```C++
__shared__ uint8_t shMem alignas(alignof(T)) [sizeof(T)];
```
test example
-----------------
```C++
#include <typeinfo>
#include <cuda_runtime.h>
#include <cstdint>

template< typename T>
__global__ void test(T* foo)
{
        __shared__  uint8_t shMem  alignas(alignof(T)) [ sizeof(T)];
        auto& sRef = *(reinterpret_cast<T*>(shMem));
        sRef.bar[0] = (*foo).bar[0];
        sRef.bar[0]++;
        (*foo).bar[0] = sRef.bar[0];
}

struct Foo
{
        alignas(16) int bar[3];
};

int main ()
{
    Foo foo;
    test<<<1,1>>>(&foo);
    return 0;
}
```

psychocoderHPC:
I hopefully fixed the clang compile and added a doc string to the trait.

Issue #321, opened 2017-01-31
heap data alignment required


psychocoderHPC:
To unleash the full performance of a processor we need to take care of the heap data alignments for global and shared memory.
Currently we align all dynamic allocated data with `boost::alignment::aligned_alloc(16u,...)`.
With 16 byte alignment we get in the most cases only SSE support from the compiler. To support authorized AVX, AVX2 and Phi KNL AVX2 code we need to check at compile time for the instruction set and used the corresponding alignment in `aligned_alloc(...)`

```
alignments needed
SSE = 16 byte
AVX/AVX2 = 32 byte
AVX-512 (Intel Phi KNL) = 64 byte
```

CC-ing: @theZiz

BenjaminW3:
You may want to use the macros defined by Boost.Predef for this purpose see [here](http://www.boost.org/doc/libs/1_60_0/libs/predef/doc/html/predef/reference/boost_hw_hardware_macros.html).
It is currently missing support for detecting AVX-512.

You could use the macros manually as it is done in alpaka:
```
#if defined(__AVX512BW__)
                64u
#elif defined(__AVX2__)
                32u
#elif defined(__SSE2__)
                16u
#elif defined(__ARM_NEON__)
                16u
#elif defined(__ALTIVEC__)
                16u
...
#endif
```

PR #322, opened 2017-01-31
use optimal heap allocation alignment

BenjaminW3:
First try to solve #321

### DIFF
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynBoostAlignedAlloc.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynBoostAlignedAlloc.hpp
@@ -21,6 +21,7 @@
 
 #pragma once
 
+#include <alpaka/core/Vectorize.hpp>            // defaultAlignment
 #include <alpaka/block/shared/dyn/Traits.hpp>   // AllocVar
 
 #include <alpaka/core/Common.hpp>               // ALPAKA_FN_ACC_NO_CUDA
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
Ohh I don_t know that boost had some magic defines for the SIMD length. 
Is the define `BOOST_HW_SIMD_X86` set on user program compile time or on boost compile time. 

If it works we need to test if we can see any speed improve with matmul or vector add. (HINT: we need to compile  with `-march=native and -tune=native`)

BenjaminW3:
Compilers define different macros e.g. `__AVX2__` depending on the target architecture. Boost.Predef only checks for the existance of the macros and defines its own macros which are much more consistent.
I will fix the clang warning in the next few days and will try to use the new alpaka version with the vecadd and matmul libs to perform some tests.

BenjaminW3:
I have remove the usage of Boost.Predef for now because it does not yet detect AVX512 but only the MIC instruction set.

psychocoderHPC:
Are the clang warnings fixed?

BenjaminW3:
Everything compiles now without warnings and the tests are green. However, I had no time to test if this has any positive performance impact until now and can not etimate if I will have time to do so in the near future.

psychocoderHPC:
@theZiz  tested the new alignments on a Intel KNL with single precision [the gemm](https://github.com/BenjaminW3/matmul) saw a speedup of `1.2` compared to the alpaka version with 16 byte alignment.

PR #323, opened 2017-02-13
MapIdx: 1u:ND and ND:1u

ax3l:
Adds general calculation for mappings from 1D to ND and vice versa. (Discussed in https://github.com/ComputationalRadiationPhysics/picongpu/pull/1805)

The current implementation is not optimized for minimal overhead, but gives the two general formulas. We probably need some more helper structs for that to build helper vars recursively, maybe you have suggestions how you would like those implemented?

Note: I left the "swap index order" change intentionally in the 2nd commit because hey, that makes it look complicated! :call_me_hand:

### DIFF
--- a/include/alpaka/idx/MapIdx.hpp
+++ b/include/alpaka/idx/MapIdx.hpp
@@ -106,13 +106,13 @@ namespace alpaka
                     MyVec idxnd( MyVec::all( 0u ) );
                     MyVec hyperPlanesBefore( MyVec::all( 1u ) );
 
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
offline discussed: current implementation missed to take care that the highest index is the fast moving one.

BenjaminW3:
Thank you for providing this implementation! I will fix the compilation and add a unit test.

ax3l:
You are welcome! Let's see if this already fixes the first part: compile :)

PR #325, opened 2017-02-16
implement optimized mapIdx for ND

BenjaminW3:
Based on #323 I implemented optimized versions of the alorithms proposed.
Both algorithms should only need the minimum required number of iterations/operations even for small vectors.
Furthermore, I added a unit test and assertions that we can not work with zero-dimensional vectors.


ax3l:
Thanks! I rebuild your branch to actually see what you changed compared to my version.

include/alpaka/idx/MapIdx.hpp
```diff
@@ -40,33 +40,35 @@ namespace alpaka
             template<
                 std::size_t TidxDimOut,
                 std::size_t TidxDimIn,
+                bool isSame = ( TidxDimOut == TidxDimIn ),
```
BenjaminW3: In my version, this line is not required

ax3l: in that case, my cherry pick was noch successful :-) will fix!

ax3l:
@BenjaminW3 @psychocoderHPC ping :)

psychocoderHPC:
I assigned my self for this pull request because both of you implemented the new version of the mapping.

psychocoderHPC:
thx @ax3l and @BenjaminW3 for this nice pull request

PR #330, opened 2017-03-04
update gcc and CMake versions used in CI

BenjaminW3:


psychocoderHPC:
I will check all pull requests next week, I am currently on a business trip.

PR #332, opened 2017-03-18
VS2017 support

BenjaminW3:
Boost adds support for VS2017 in `boost-1.64.0`. However, the current CMake version does not support finding `boost-1.64.0` correctly because it looks for the library file suffix `15.0` but `boost-1.64.0` creates files with `14.1`. This is fixed in CMake-3.8.1 which we download by hand.
Furthermore, `boost-1.64.0` does not support VS2017 fully because boost.uuid misses an include that was previously available by default. This is fixed by including it manually before including those boost headers.


BenjaminW3:
push

psychocoderHPC:
kk I review it next week.

BenjaminW3:
push

PR #333, opened 2017-03-23
support clang 4.0+ native CUDA

BenjaminW3:


psychocoderHPC:
The one test is not compiling because of a host function call.
Clang not allows to call a `__host__` from `__host__ __device__`, see [this documentation ](http://llvm.org/docs/CompileCudaWithLLVM.html#id13)
Please see also [this](https://github.com/ComputationalRadiationPhysics/picongpu/pull/1707) discussion about the hidden nvcc suppressor pragmas and the behavior with clang.

BenjaminW3:
We have some generic functions like those to create an `alpaka::vec::Vec` which are declared as `__host__ __device__`. They take the element creation function object as a template parameter. In some cases those element creation function objects are only `__host__` or `__device__` When called from the respective side, this is valid code. So `__host__` -> `__host__ __device__` -> `__host__` as well as `__device__` -> `__host__ __device__` -> `__device__` are legitimate use cases.
If clang does not support this anymore with version 4.0, this is a problem.
As far as I can see, the only possible work around is to use the overloading feature of clangfor those intermediate functions and declare identical copies of the `__host__ __device__` functions as `__host__` and `__device__`. However, nvcc does not support attribute overloading at all.

psychocoderHPC:
:-( that is a bad news. I thought this behavior is equal in clang 3.9 and 4.0. I can compile PIConGPU with clang 4.0pre release without those issue but maybe it is a result that we define everything as `__host__ __device__`

psychocoderHPC:
I will spend some time to look into this issue in two weeks after my vacation

psychocoderHPC:
sry for the long delay I will test this PR tomorrow. But I agree with you that we maybe need to add a overloaded function for host and device if clang is used as device compiler.

BenjaminW3:
@psychocoderHPC After thinking about it many times and having a look at the clang source code and unit tests regarding this, I think it is clangs desired behaviour to reject code that calls `__host__` -> `__host__ __device__` -> `__host__` or `__device__` -> `__host__ __device__` -> `__device__`.
For clang as well as nvcc , `__device__` or `__host__` functions are the same class of functions. They can call the same class as well as `__host__ __device__` functions.
For clang `__host__ __device__` is the most restricted class of functions. They can neither call `__device__` nor `__host__` functions. Only other `__host__ __device__` functions are allowed to be called. This means everything that a `__host__ __device__` function can ever call down the call stack can be executed on both sides. This can be verified before generating the code by pure syntax analysis.
However, for nvcc `__host__ __device__` functions are the least restricted class of functions. They can call other `__host__ __device__` functions, but also `__device__` or `__host__` functions. As long as they are called from the right side, there is no problem. This is not enforced at compile time and might lead to compile errors, misgenerated code or simply crashing code.

BenjaminW3:
@psychocoderHPC We could work around this by replacing all of the `__host__ __device__` functions with two overloaded identical versions for clang - one declared only as `__host__` and one only declared as `__device__`. This would result in two clean call stacks: `__host__` -> `__host__` -> `__host__` or `__device__` -> `__device__` -> `__device__` for the respective sides due to clangs overload resolution rules for those target attributes.
nvcc does not support such overloading and therefore needs the previous `__host__ __device__` methods. 

I am sceptical about what implications this would have for our heterogeneous kernels declared as `__host__ __device__`. Probably they would have to be defined twice for clang as well. However, this is exactly what we do not want to do. The only solution I can imagine is to have a macro for those `__host__ __device__` functions that declares them as `__host__ __device__` for nvcc but duplicates and overloads them with `__host__` and `__device__` for clang >= 4.0.

psychocoderHPC:
I tested clang 4.0 as native compiler with [cupla](https://github.com/ComputationalRadiationPhysics/cupla), the only thing missing is the compiler flag `--cuda-path=${CUDA_TOOLKIT_ROOT_DIR}`. I will add a pull request for the missing option today.

psychocoderHPC:
To avoid merge conflicts and waiting two times for the CI I will add the clang device compiler flag after #363.

BenjaminW3:
As long as you use only one accelerator (either a host or a CUDA one), the tests compile correctly. Only those using both with the same kernel within the same executable fail.

BenjaminW3:
Woohoo! I may have found a version that works now! I tried many things but in the end the main culprit were again the `Vec` factory methods. Maybe they are really too hard because nvcc and intel do not support them either. Possibly they are simply buggy...
As soon as I get a green build, I will remove the WIP tag and will ask for merging.

BenjaminW3:
ready to be merged (as well as the other two PRs)!

psychocoderHPC:
well done!!!!

PR #335, opened 2017-04-02
test cmake-3.8.0

BenjaminW3:
@ComputationalRadiationPhysics/alpaka-maintainers How should we handle the deprecation of older CMake versions? It would be desirable to only support 3.7+ because they support correct target handling for CUDA. This would allow to remove some cruft from the test CMakeLists.
Do your current  build environments already provide CMake3.7?


ax3l:
I understand your pain, but 7 days after the initial release I can not pop the latest CMake 3.8.0 as a requirement unto users (yet) :)

Let's get out the next PIConGPU release (0.4.0/1.0.0) with initial cupla support with the current CMake requirements (~summer) and after that, in PIConGPU's late summer/autum release, we can push the requirement up. Is that ok with you?

PR #336, opened 2017-04-20
test boost version 1.64.0

BenjaminW3:
@ComputationalRadiationPhysics/alpaka-maintainers Same question here: How should we handle the deprecation of older boost versions? Boost fibers were integrated into boost in 1.62.0. It would be easier to support less version and only the ones that have all features that alpaka provides.
It was and is not feasible to test all boost versions against all supported compiler combinations. But in the future it may not even be possible to test each boost version at least once in the build matrix at all.


ax3l:
Great to hear fiber is now officially included! Yes, Boost 1.62.0 is released half a year ago (September 2016)

If you are ok I would let @psychocoderHPC merge in the bunch of patches he has prepared for the mainline cupla/alpaka PIConGPU release that comes up next (maybe 2-3 months) and after that we can increase it.

If if hinders you too much go ahead here (Boost 1.62.0+) and with #335 (CMake 3.7+), the only thing I would wish for from the "downsteam" perspective is that you start making regular releases with versions that state their dependencies explicitly, since a lot of install documents and scripts will now depend on it.

BenjaminW3:
@ComputationalRadiationPhysics/alpaka-maintainers Thanks for your answer. It does not hinder me at the moment. The reliability of the CI decreases the more versions we pretent to support but have not been able to test in all combinations.
I am fine with waiting some months and then raise the minimums to boost-1.62 and CMake-3.7.
Last question: Do we still have to support CUDA 7.0? As far as I have found out, PIConGPU does not do so. Could we drop support for CUDA 7.0 in alpaka as well? It would allow me to remove some workarounds.

ax3l:
Sounds great!

> Last question: Do we still have to support CUDA 7.0? As far as I have found out, PIConGPU does not do so. Could we drop support for CUDA 7.0 in alpaka as well? It would allow me to remove some workarounds.

Yes, feel free to bump to CUDA 7.5+, Titan (ORNL) supports it and PIConGPU needs it, too. How about releasing a new version with the current (CUDA 7.0 compatible) state of alpaka first and then you can bump it and remove all the nasty workarounds for CUDA 7.0?

.travis.yml
```diff
@@ -49,7 +49,7 @@ compiler:
 #   [clang++] ALPAKA_CLANG_VER                  : {3.5.2, 3.6.2, 3.7.1, 3.8.1, 3.9.0}
 # CMAKE_BUILD_TYPE                              : {Debug, Release}
 # ALPAKA_CI                                     : {ON}
-# ALPAKA_CI_BOOST_BRANCH                        : {boost-1.59.0, boost-1.60.0, boost-1.61.0, boost-1.62.0, boost-1.63.0, develop}
+# ALPAKA_CI_BOOST_BRANCH                        : {boost-1.59.0, boost-1.60.0, boost-1.61.0, boost-1.62.0, boost-1.63.0, boost-1.64.0, develop}
```
psychocoderHPC: Boost `boost-1.60.0` must be removed from the list because it is not checked.

BenjaminW3: It is still tested, see line 90

psychocoderHPC: Ahh fine I haven't seen it.

psychocoderHPC:
With you last changes somethings breaks in appveyor

BenjaminW3:
Appveyor has updated its CMake version to 3.8.0 which is broken for Windows with newer boost versions. #332 fixes this by manually downloading CMake-3.8.1 which includes my fix and adds support for Visual Studio 2017.

PR #344, opened 2017-06-15
remove support for CUDA 7.0

BenjaminW3:
Currently only for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -65,7 +65,7 @@ compiler:
 # ALPAKA_ACC_CPU_BT_OMP4_ENABLE                 : {ON, OFF}
 #   [ON] OMP_NUM_THREADS                        : {1, 2, 3, 4}
 # ALPAKA_ACC_GPU_CUDA_ENABLE                    : {ON, OFF}
-#   [ON] ALPAKA_CUDA_VER                        : {7.0, 7.5, 8.0}
+#   [ON] ALPAKA_CUDA_VER                        : {7.5, 8.0}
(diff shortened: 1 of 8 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -46,13 +46,6 @@
 #include <type_traits>                      // std::enable_if, std::decay, std::is_same
 #include <algorithm>                        // std::min, std::max, std::min_element, std::max_element
 
-// Some compilers do not support the out of class versions:
-// - the nvcc 7.0 CUDA compiler
```
BenjaminW3: Testing showed, that this work around is required even for CUDA 7.5 and CUDA 8.0...

psychocoderHPC: Than we should only update the comment

BenjaminW3: This comment has already been added in a different PR.

BenjaminW3:
For now I do not see a hard reason to remove CUDA 7.0.
I could not find / remember places where CUDA 7.0 required workarounds due to compiler bugs.
The C++ support should be nearly identical for all 3 current versions.
Feature wise  CUDA 7.5 only brought experimental support for GPU lambdas and CUDA 8.0 brought support for heterogeneous lambdas. Both are not required by alpaka.
Does anybody else see a reason to remove CUDA 7.0 support? Else, I would close this Pull Request for now and keep it in the list of supported CUDA versions.

PR #346, opened 2017-06-15
remove support for boost-1.61.0 and older

BenjaminW3:
Furthermore, this removes testing the boost develop branch because this is not supported since the change to use imported boost targets. They are not available for the develop branch from FindBoost and those jobs currently always fail.


.travis.yml
```diff
@@ -49,7 +49,7 @@ compiler:
 #   [clang++] ALPAKA_CLANG_VER                  : {3.5.2, 3.6.2, 3.7.1, 3.8.1, 3.9.0}
 # CMAKE_BUILD_TYPE                              : {Debug, Release}
 # ALPAKA_CI                                     : {ON}
-# ALPAKA_CI_BOOST_BRANCH                        : {boost-1.59.0, boost-1.60.0, boost-1.61.0, boost-1.62.0, boost-1.63.0, boost-1.64.0, develop}
+# ALPAKA_CI_BOOST_BRANCH                        : {boost-1.62.0, boost-1.63.0, boost-1.64.0, develop}
```
psychocoderHPC: `develop` needs to be removed from the list

BenjaminW3: Thanks, I missed those.

PR #349, opened 2017-06-16
Release 0.2.0

BenjaminW3:
This is the second release of alpaka. It does not contain huge changes but only compatibility fixes and small enhancements:

* the documentation has been greatly enhanced
* adds support for CUDA 8.0
* adds support for CMake versions 3.6, 3.7 and 3.8
* adds support for Boost 1.62, 1.63 and 1.64
* adds support for clang-3.9
* adds support for Visual Studio 2017
* alpaka now compiles clean even with clang `-Weverything`
* re-enabled the `boost::fiber` accelerator backend which was disabled in the last release

API changes:
* `mapIdx` is moved from namespace `alpaka::core` to `alpaka::idx`
* `Vec` is moved from namespace `alpaka` to `alpaka::vec`
* `vec::Vec` is now allowed to be zero-dimensional (was previously forbidden)
* added `vec::concat`
* added element-wise `operator<` for `vec::Vec` which returns a vector of bool
* CPU accelerators now support arbitrary dimensionality (both kernel execution as well as memory operations)
* added support for `syncBlockThreadsPredicate` with `block::sync::op::LogicalOr`, `block::sync::op::LogicalAnd` and `block::sync::op::Count`
* memory allocations are now aligned optimally for the underlying architecture (16 bit for SSE, 32 bit for AVX, 64 bit for AVX512) instead of 16 bit for all architectures in the previous release

Thanks to all contributors!


BenjaminW3:
@ComputationalRadiationPhysics/alpaka-maintainers What has happened with the TBB back-end? It was always available but has been undocumented in a10d29402f0b0d86773f2d3e088b7242c928e833 but has never been documented again.

psychocoderHPC:
@BenjaminW3 It was removed before the first release, please see your mail form 07/07/2016

BenjaminW3:
Currently I do not see the necessity for such a changelog within the repository because the changelog can be found next to the github release and the full history is always available to everyone.

ax3l:
Sure, on the release page is fine, too

PR #351, opened 2017-06-18
unify Vec creation workarounds into single place

BenjaminW3:
This is a simple refactoring that unifies the workaround for the Vec factory methods into a single place. This allows to remove most of the #ifdef magic from all over the code.
This is prework for the clang-4.0 support.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -47,7 +47,7 @@
 #include <algorithm>                        // std::min, std::max, std::min_element, std::max_element
 
 // Some compilers do not support the out of class versions:
-// - the nvcc 7.0 CUDA compiler
+// - the nvcc CUDA compiler (at least 7.0, 7.5 and 8.0)
 // - the intel compiler
 #if BOOST_COMP_NVCC || BOOST_COMP_INTEL
     #define ALPAKA_CREATE_VEC_IN_CLASS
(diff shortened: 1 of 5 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -555,6 +555,66 @@ namespace alpaka
...
+        //! This is a conveniance method to have a out-of-class factory method even though the out-of-class version is not supported by all compilers.
+        //! This has the draw-back, that it requires the TSize parameter even though it should not be necessary.
+        //-----------------------------------------------------------------------------
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<
+            typename TDim,
+            typename TSize,
+            template<std::size_t> class TTFnObj,
+            typename... TArgs>
+        ALPAKA_FN_HOST_ACC auto createVecFromIndexedFn2(
```
psychocoderHPC: Could we find a better name for `createVecFromIndexedFn2` e.g. `indexedFnToVec`or only `vecFromIndexed`. 
The point that a vector is created should be clear from the result type.

BenjaminW3: For conforming compilers there are already the versions without the `2` at the end (`createVecFromIndexedFn`, ...). Non conformant compilers have a workaround where those functions are within the class (which is not the purpose of a factory function).
Previously at every place where one of the factory functions was called, the correct one was selected manually via `#ifdef`.
Those new functions (`createVecFromIndexedFn2`, ...) abstract this manual selection away by providing an out of class version which calls the correct version (out-of-class/ in-class) for the underlying compiler.
Those functions are only a temporary hack and should be removed as soon as all compilers support the native out-of-class version.
Maybe we should call them `createVecFromIndexedFnWorkaround` or `createVecFromIndexedFnHack`?

psychocoderHPC: I would say `createVecFromIndexedFnWorkaround` but what is with something like 
```
/** This function is equivalent to the class member createVecFromIndexedFn and the external function ....
Depending of the compiler the internal or external factory(creation function) is called.
*/ 
createVecFromIndexedFnAuto
```

BenjaminW3: I prefer the `createVecFromIndexedFnWorkaround` name. Will change it.

BenjaminW3: I have renamed the functions and extended the comments.

BenjaminW3:
push

Issue #352, opened 2017-06-22
Fix CUDA Installation on Ubuntu 14.04 after travis update


BenjaminW3:
```
E: Failed to fetch file:/var/cuda-repo-7-5-local/Release  No Hash entry in Release file /var/lib/apt/lists/partial/_var_cuda-repo-7-5-local_Release which is considered strong enough for security purposes
E: Some index files failed to download. They have been ignored, or old ones used instead.
```

BenjaminW3:
Fixed with #354

PR #354, opened 2017-06-24
fix travis CUDA installation

BenjaminW3:
This fixes the CI that is currently broken. See #352.

Travis CI updated their Ubuntu images.
With the updates, old packages [only providing SHA1](https://wiki.debian.org/Teams/Apt/Sha1Removal) are broken and can not be installed.
```E: Failed to fetch file:/var/cuda-repo-7-5-local/Release  No Hash entry in Release file /var/lib/apt/lists/partial/_var_cuda-repo-7-5-local_Release which is considered strong enough for security purposes```

Both, CUDA 7.0 and 7.5 packages are therefore broken.
By adding the SHA256 manually before installing and ignoring that they can not be verified, the installation is successfull again (see [this stackexchange question](https://askubuntu.com/questions/760242/how-can-i-force-16-04-to-add-a-repository-even-if-it-isnt-considered-secure-eno)).
```E: There were unauthenticated packages and -y was used without --allow-unauthenticated```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -290,7 +290,7 @@ before_install:
                   if [ ${ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE} == "ON" ]
                   ;then
                       export ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
something is still wrong with apt-get

Issue #362, opened 2017-08-04
CUDA: Multi-Arch Support


ax3l:
PIConGPU/PMacc allows to define multiple [PTX+SASS targets for CUDA](https://github.com/ComputationalRadiationPhysics/picongpu/blob/97988937ded92a395770316767341fc081b999e2/include/pmacc/PMaccConfig.cmake#L155-L159) device code.
In the process of putting PIConGPU on alpaka in mainline we need to port this feature to alpaka.

PR #363, opened 2017-08-06
add multi-arch support

BenjaminW3:
#362


alpakaConfig.cmake
```diff
@@ -301,7 +291,15 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
             OPTION(ALPAKA_CUDA_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps (folder: nvcc_tmp)" OFF)
 
             IF(ALPAKA_CUDA_COMPILER MATCHES "clang")
-                LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "--cuda-gpu-arch=${ALPAKA_CUDA_ARCH}")
+                LIST(
+                    LENGTH
+                    ALPAKA_CUDA_ARCH
+                    _ALPAKA_CUDA_ARCH_LENGTH)
+                IF(${_ALPAKA_CUDA_ARCH_LENGTH} GREATER 1)
+                    MESSAGE(WARNING "Multi-arch compilation with clang as native compiler not supported!")
```
ax3l: Maybe: "... with clang as native CUDA compiler ..." ?

psychocoderHPC:
The log of a travis CI test is larger than 4MB there the test can't pass.

```
The log length has exceeded the limit of 4 MB (this usually means that the test suite is raising the same exception over and over).
```

alpakaConfig.cmake
psychocoderHPC: I tried multi arch with clang as device compiler with PIConGPU and created the ptx code for two architectures within one compile run. With clang 4.0 multi architecture compile **is supported**.

.travis.yml
```diff
@@ -671,6 +671,18 @@ script:
     - if [ "${ALPAKA_CI_SANITIZERS}" == *"MSan"* ]; then export CMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS} -fsanitize=memory -fsanitize-memory-track-origins"; fi
 
     #-------------------------------------------------------------------------------
+    # gpu architecture
+    - if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" ]
+      ;then
+          if [ "${ALPAKA_CUDA_COMPILER}" == "clang" ]
+          ;then
+              export ALPAKA_CUDA_ARCH="20"
```
psychocoderHPC: This can be removed if [this](https://github.com/ComputationalRadiationPhysics/alpaka/pull/363/files#r131593736) limitation is removed.

alpakaConfig.cmake
BenjaminW3: You are right. I did not had a look at the documentation and simply expected that it is not supported right now. I stand corrected. It is even easier than for nvcc.

.travis.yml
BenjaminW3: done

psychocoderHPC:
@BenjaminW3  Can we somehow disable the log spamming for the CI?

BenjaminW3:
I will try to reduce it. However, as far as I can remember, nearly all of the spam are warnings about host functions called from host device functions within boost. Due to the device code being built again for each architecture, those warnings with huge call stacks are now duplicated.

ax3l:
Would it help to go to CMake 3.8+, which has native CUDA language support, so boost can be included as -isystem library again? 
This would kill all warnings originating from third party headers.

BenjaminW3:
I suppressed some installation and download logging and found found some HD warnings that could be suppressed. So the log should now be shorter than before even when the device code is compiled multiple times.

Issue #368, opened 2017-08-08
polling: low performance with CPU accelerator and async streams


psychocoderHPC:
If multiple asynchronous streams will be used on a CPU accelerator and the user program is using a lot of test calls to check the states of events and streams than we lose one cpu core for the polling action.

The overhead can be reduced if we add the possibility to add priorities to a device and/or stream.

**This is only a suggestion we need to evaluate the possibilities.**

BenjaminW3:
I do not know if I understand your problem correctly, could you please elaborate it some more?

BenjaminW3:
It may be possible to replace the mutex lock within the `alpaka::event::test` method for `EventCpu` with an atomic memory access. This should be by far faster.

psychocoderHPC:
The performance of the test call is ok. The issue is coming that one the master thread (thread which create kernel calls and records events) is highly frequently polling all eventa to find out which event is finished. 
This is some how a user code issue which can slidly solved by alpaka if we can set the priority of the master thread. This means that the polling thread is not so often activated by the operating system if compite threads have a higher priority.

A clean solution would be asynchron callbacks but this is not supported by cuda, or would be increase the cost to start a kernel.

This issue has no high priority but I opened it that we can work on it in the future.

BenjaminW3:
Asynchronous callbacks are supported via `cudaStreamAddCallback`. It would be no problem to add this to alpaka looking like this:
```
alpaka::stream::enqueue(stream, [&arbitrary](){do_something();});
```
CPU backends simply start a new thread with the user callback when it is reached.
The CUDA backend directly starts a thread that waits on a condition_variable. This condition_variable is activated by the CUDA callback and allows the thread to go on executing the user calback.
This would even work around the CUDA limitations `Callbacks must not make any CUDA API calls. Attempting to use CUDA APIs will result in cudaErrorNotPermitted.`

psychocoderHPC:
The problem with cuda calls back is that this means we need to start a kernel, add a call back add a event.
This will increase the time within the driver.
The event is needed to build dependencies between streams.
Currently we start a kernel and than add an event for status checks and dependency management.

A driver call like adding a event or a kernel call cost around 14us if I remember right. In the case of PIConGPU we start up to 10000 Events per  second.

If we will implement callback we need to take care of HIP and openACC.

BenjaminW3:
[...]

PR #373, opened 2017-08-11
add stream callback

BenjaminW3:
possible solution for #368


psychocoderHPC:
note: I am two weeks not available and will check all PRs after my holidays.

psychocoderHPC:
Could you please add a short description which concepts are new in this pull request and maybe a code snipped (or pointing to a code snipped) how to use the call backs.

How does the call back works. Is there a new thread opened and executes the call back code?

BenjaminW3:
There are no new concepts. The `alpaka::stream::traits::Enqueue` trait and its accessor method `alpaka::stream::enqueue` are used as is and specialized for callbacks as parameter.
Callback usage looks like this:
```
alpaka::stream::enqueue(
    stream,
    [&](){
        do_what_you_want_in_here();
    }
);
```
A usage example can also be found in the new unit test.

Callbacks are always executed within an independent thread.  The CPU streams always supported this (unintentionally ;-) ) because it uses a thread pool to execute the tasks. Only for CUDA streams there is new code. CUDA itself calls the callback from within its own thread which does not allow to call CUDA runtime API functions from within the callback. This limitation is lifted by starting a new thread per callback (for async CUDA streams).

BenjaminW3:
ready for review ;-)

include/alpaka/stream/StreamCudaRtAsync.hpp
```diff
@@ -238,7 +238,79 @@ namespace alpaka
...
+                {
+                    auto pCallbackSynchronizationData = std::make_shared<CallbackSynchronizationData>();
+
+                    ALPAKA_CUDA_RT_CHECK(cudaStreamAddCallback(
+                        stream.m_spStreamCudaRtAsyncImpl->m_CudaStream,
+                        cudaRtCallback,
+                        pCallbackSynchronizationData.get(),
+                        0u));
+
+                    std::thread t(
```
psychocoderHPC: To create a thread per callback can be very expensive in time. e.g. PIConGPU is spawning over 2000 kernel/memcpy per second and will have over 100 tasks waiting in streams. This means we need to spawn 2k threads/s and have over 100 active threads.

Is it possible to use on thread for all callbacks (waiting in the background), add callbacks to a list and than execute the callback always from the callback thread?

We can move this also to later pull request if it is currently not easy to do.

BenjaminW3: Currently, I would like to merge this as is. I would create a follow-up issue for optimizing this.
The most flexible solution would be a thread pool with a queue of ready callbacks. If there is only one thread, the latency would be highest but it would equal your single thread solution, if there are multiple threads, the latency/resource tradeoff can be adapted per use case.

BenjaminW3: Can we merge this and create the follow up ticket? My upcoming event unit tests are based on some of those stream test helpers.

psychocoderHPC: sry I was busy, yes I will merge it

PR #375, opened 2017-08-12
remove boost.context from ReadMe requirements

BenjaminW3:


ax3l:
oh, good to know. I just added that to PIConGPU ;)

Is context not a requirement of fibers anymore?

ax3l:
oh, good to know. I just added that to PIConGPU ;)

Is context not a requirement of fiber anymore?

ax3l:
@BenjaminW3 also, the `alpakaConfig.cmake` still seems to require a good ammount of components that need compiles: `FIND_PACKAGE(fiber context system thread atomic chrono date_time)`

BenjaminW3:
Yes, all of those are dependencies of boost fiber. However, I only mentioned one transitive dependeny boost context and not the others. This was inconsistent. So I preferred to remove this one. The dependencies of fiber could change with every boost version. So I do not want to mention them explicitly. Boost build should handle to build those dependencies automatically.

ax3l:
Ok that's fine. Unluckily CMake is always behind keeping the transitive deps up-to-date until they fully switch to CMake in the future, so we did mention transitive deps (e.g. system in filesystem) explicitly so far.

Nevertheless, what is the correct way to import alpaka via cupla with its pulled dependencies. It looks to me that a found boost fiber third-party dependency is not fully forwarded to PIConGPU yet.

ax3l:
Ah I think I found the cause. `PMaccConfig.cmake` does not yet forward the cupla libs...

PR #381, opened 2017-08-14
support CMake 3.9.0

BenjaminW3:
#380


cmake/addExecutable.cmake
```diff
@@ -41,7 +41,11 @@ FUNCTION(ALPAKA_ADD_EXECUTABLE In_Name)
                     SET_SOURCE_FILES_PROPERTIES(${_file} PROPERTIES CUDA_SOURCE_PROPERTY_FORMAT OBJ)
                 ENDIF()
             ENDFOREACH()
-            CMAKE_POLICY(SET CMP0023 OLD)   # CUDA_ADD_EXECUTABLE calls TARGET_LINK_LIBRARIES without keywords.
+            IF (CMAKE_VERSION VERSION_LESS 3.9.0)
+                CMAKE_POLICY(SET CMP0023 OLD)   # CUDA_ADD_EXECUTABLE calls TARGET_LINK_LIBRARIES without keywords.
+            ELSE()
+                SET(CUDA_LINK_LIBRARIES_KEYWORD "PUBLIC")
```
ax3l: is that a new, undocumented keyword or do I miss its usage?

ax3l: Ah it's a new keyword in the `FindCUDA.cmake` module, all right!

PR #385, opened 2017-08-21
support VS 2017.3

BenjaminW3:
Appveyor updated its Visual Studio 2017 installation. The new C++ compiler partially supports two-phase-name-lookup. However, it does not when openmp is enabled ;-) In this case, it does not disable two-phase-name-lookup but it disables OpenMP. So we have to disable two-phase-name-lookup manually for now.


appveyor.yml
```diff
@@ -147,7 +147,7 @@ before_build:
     - cmd: if "%CONFIGURATION%"=="Release" set ALPAKA_BOOST_VARIANT=release
     # Select the libraries required.
     - cmd: set ALPAKA_BOOST_B2=--with-test
-    - cmd: if "%ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE%"=="ON" set ALPAKA_BOOST_B2=%ALPAKA_BOOST_B2% --with-fiber --with-context --with-thread --with-system --with-atomic --with-chrono --with-date_time
```
ax3l: why don't you need atomic?
atomic is needed by thread and thread is needed by fiber.

BenjaminW3: You are right, I do not know where this came from.

PR #386, opened 2017-08-22
add support for boost-1.65

BenjaminW3:
* replaces usage of `Boost::unit_test_framework` with `Boost_Libraries` because CMake does not know boost-1.65.1 yet so it does not create this target.
* **boost-1.65.0 is not supported, only boost-1.65.1!**


BenjaminW3:
Testing shows that boost-1.65 fails to compile when using nvcc together with clang:
`/home/travis/build/boost/boost/smart_ptr/detail/local_counted_base.hpp(66): error: identifier "__builtin_assume" is undefined`

test/analysis/headerCheck/CMakeLists.txt
```diff
@@ -96,9 +99,12 @@ ALPAKA_ADD_EXECUTABLE(
 TARGET_COMPILE_OPTIONS(
     ${_TARGET_NAME}
     PRIVATE ${_DEFINITIONS_PRIVATE})
+TARGET_INCLUDE_DIRECTORIES(
+    ${_TARGET_NAME}
+    PRIVATE ${_INCLUDE_DIRECTORIES_PRIVATE})
 TARGET_LINK_LIBRARIES(
     ${_TARGET_NAME}
-    PUBLIC common Boost::unit_test_framework)
+    PRIVATE common ${_LINK_LIBRARIES_PRIVATE})
```
ax3l: maybe with `if(TARGET Boost::unit_test_framework)` ? This would still in most cases give the more stable target to link...

travis/boost-1-65-0-local_counted_base-builtin_assume-nvcc-clang.patch
```diff
@@ -0,0 +1,32 @@
...
+
+diff --git a/include/boost/smart_ptr/detail/local_counted_base.hpp b/include/boost/smart_ptr/detail/local_counted_base.hpp
+index 398b46d..9e3b253 100644
+--- a/include/boost/smart_ptr/detail/local_counted_base.hpp
++++ b/include/boost/smart_ptr/detail/local_counted_base.hpp
+@@ -60,6 +60,7 @@ public:
+ 
+     void add_ref() BOOST_SP_NOEXCEPT
+     {
++#if !defined(__NVCC__)
```
ax3l: @BenjaminW3 can you bring this patch to boost mainline?

BenjaminW3: I would like to have this in the next boost version. However, I am not sure, if this is the least invasive fix possible...
If you think that this is correct, I will create a PullRequest for `Boost.SmartPtr`.

test/analysis/headerCheck/CMakeLists.txt
BenjaminW3: I could, but this is only test code and I think less code is favourable.

travis/boost-1-65-0-local_counted_base-builtin_assume-nvcc-clang.patch
ax3l: maybe as a `BOOST_HAS_BUILTIN_ASSUME` switch?

ax3l:
Weird, I *can compile* the following snippet with `nvcc -std=c++11` (CUDA V8.0.44)
```C++
class local_counted_base
{
  private:

  local_counted_base & operator= ( local_counted_base const & );

    // not 'int' or 'unsigned' to avoid aliasing and enable optimizations
    enum count_type { min_ = 0, initial_ = 1, max_ = 2147483647 };

  count_type local_use_count_;

  public:
    constexpr local_counted_base() noexcept: local_use_count_( initial_ )
    {
    }

    constexpr local_counted_base( local_counted_base const & ) noexcept: local_use_count_( initial_ )
    {
}

    void add_ref() noexcept
    {
#if defined( __has_builtin )
# if __has_builtin( __builtin_assume )

        __builtin_assume( local_use_count_ >= 1 );

# endif
#endif

        local_use_count_ = static_cast<count_type>( local_use_count_ + 1 );
  }
};

int main()
{ return 0;}
```

travis/boost-1-65-0-local_counted_base-builtin_assume-nvcc-clang.patch
ax3l: yes, less invasive. see my snippet below: I try to compile a NVCC bug report in parallel but can not reproduce your nvcc+clang issue yet.

BenjaminW3:
The compilation error occured with clang-3.8.1, boost-1.65.0, CUDA 8.0, sm35 in Debug mode see [here](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/267456998)
minimally shortened command line:
`
/usr/local/cuda-8.0/bin/nvcc /alpaka/test/unit/acc/src//main.cpp -x=cu -c -o /alpaka/test/unit/acc/build/make/CMakeFiles/acc.dir/src/./acc_generated_main.cpp.o -ccbin /llvm/llvm-3.8.1/bin/clang++ -m64 -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_DEBUG=0 -DALPAKA_CI -Xcompiler ,\"-fno-omit-frame-pointer\",\"-fno-optimize-sibling-calls\",\"-fopenmp=libomp\",\"-g\" -Wno-deprecated-gpu-targets --expt-extended-lambda --generate-code arch=compute_35,code=sm_35 --generate-code arch=compute_35,code=compute_35 -std=c++11 -g -G --use_fast_math --ftz=false -DNVCC -I/usr/local/cuda-8.0/include -I/boost  -I/usr/include -I/alpaka/include
`

where main.cpp consists only of:
```
#define BOOST_TEST_MODULE acc
#include <boost/predef.h>
#include <boost/test/unit_test.hpp>
```

travis/boost-1-65-0-local_counted_base-builtin_assume-nvcc-clang.patch
BenjaminW3: I have done a search across all boost libraries and `Boost.SmartPtr` is the only library that uses `__builtin_assume`. So I think the `BOOST_HAS_BUILTIN_ASSUME` would be too complex as it requires changes to two repositories.

BenjaminW3:
@ax3l There is going to be a 1.65.1 release. Could you please make sure that your NVCC version fix that is already merged to boost.config master gets merged into this release? This might need some mailing list interaction...
Have you been able to verify the bug above with the given source/command line?

ax3l:
Oh, great to hear that! 

I am currently out of office for two weeks. 
Can you please propose on the list that
https://github.com/boostorg/config/pull/175
Is included? 

That is the fix you are referring to.

BenjaminW3:
@ax3l Both of our fixes, your Boost.Config change for CUDA 9.0 as well as my Boost.SmartPtr fix are now on Boost master as well as on the special boost-1.65.1 branch. I will update this PR once boost-1.65.1 is released. 1.65.0 will not be supported.

ax3l:
Thank you and +1 from Croatia.

BenjaminW3:
ready to be merged!

ax3l:
> boost-1.65.0 is not supported, only boost-1.65.1!

I think the missing support for 1.65.0 should be noted in the (install) docs or throw a warning/error in cmake, no?

BenjaminW3:
The problem I see with listing all the incompatible dependencies is that we only test the latest patch version of everything (`boost-1.65.1`, `clang-3.8.1`, `cmake-3.9.1`, etc...). I think it is highly likely that alpaka does not support many of those dependencies in their unpatched versions but we might never find out because testing everything is unfeasible. If we would start such a list it always feals like it should be comprehensive, which it would never be.

ax3l:
Sure, I was just wondering if we list a release that is not compatible in *any case*. But 1.65.0 is actually working with alpaka, as long as CUDA 9.0 (RC) is not used.

PR #394, opened 2017-09-05
add more tests for events

BenjaminW3:
* some renamings of event interna (wrong capilatization, some shortening)
* removes the useless reflexive specializations of `EventType` for the event types themselves
* removes the `event::create` method and the corresponding trait as it always only called the constructor
* adds code for a manually host side triggered event (`EventHostManualTrigger`) that streams can wait on (currently only for CPU streams). This is currently only test code and not part of alpaka itself.
* currently only tests CPU streams/event (CUDA stream testing will be added later)

### DIFF
--- a/include/alpaka/event/EventCpu.hpp
+++ b/include/alpaka/event/EventCpu.hpp
@@ -62,7 +62,7 @@ namespace alpaka
                         dev::DevCpu const & dev) :
                             m_uuid(boost::uuids::random_generator()()),
                             m_dev(dev),
-                            m_Mutex(),
+                            m_mutex(),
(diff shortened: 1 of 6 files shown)


BenjaminW3:
ready to merge!

BenjaminW3:
I had a look at cupla and it does not use the `event::create` method. So it should be save to remove it.

psychocoderHPC:
@BenjaminW3  Yes from the cupla side it should be save. Never the less I will try this PR with cupla tomorrow before merging.

BenjaminW3:
any progress?

BenjaminW3:
Thanks!

Issue #388, opened 2017-08-28
CPU event re queuing bug


psychocoderHPC:
If a event is re queued during a other async stream has a waiting operation to this event in the queue which is not executed than a deadlock of the alpaka system can be created.

- 1. kernel K1 is enqueued to async stream S1
- 2. kernel K2 is enqueued to asnyc stream S2
- 3. event E1 is enqueued to async stream S1
- 4. wait is executed that S2 is waiting on E1 
  - https://github.com/ComputationalRadiationPhysics/alpaka/blob/439c98b1a326e0fff73b2beeb77e084207a89647/include/alpaka/event/EventCpu.hpp#L514-L522
  - m_bIsWaitedFor is set to true to signalize the event that someone is waiting
  - a lambda is enqueued in S2 which starts a wait if E1 is not finished
- 5. K1 finish
- 6. now E1 finish, m_bIsWaitedFor is reseted, all waiting threads are notified but queued wait task from S2 is currently not registered because K2 is still running
- 7. E1 is re queued again but this time in S2
- 8. K2 finish
- 9. the waiting lambda from 4. is executed and because E1 is not finished this code https://github.com/ComputationalRadiationPhysics/alpaka/blob/439c98b1a326e0fff73b2beeb77e084207a89647/include/alpaka/event/EventCpu.hpp#L480-L483 is executed and S2 is waiting on E1, m_bIsWaitedFor is enabled in E1 again

E1 is queued behind the waiting lamda from 4. therefore S2 will never finish (**deadlock**).

I hope I pointed always to the correct specializations of wait. Currently I have no idea how we can solve this issue. 
I was not able to reproduce this issue in a minimal example.

maybe @BenjaminW3 has an idea how we can solve this bug.

BenjaminW3:
This really sounds like a bug.
Waits on events that are not valid anymore (have already been finished) should simply be skipped, so step 9 should never be executed. This could be prevented by adding a counter to the event that is increased each time the event is enqueued (`enqueueCount`). When a stream should wait for an event `enqueueCount` is saved and when the wait is reached it is compared to the current `enqueueCount` count. If they do not match, the event has already been finished and re-enqueued.
I will try to build some tests for the events and simulate this case as well.  This could take some days, but will be solved.

psychocoderHPC:
Thanks that is a good idea I also thought about a counter but  my workflow was to complicated, your idea is much easier. Will test it tomorrow.

[...]

PR #396, opened 2017-09-05
event tests for CUDA

BenjaminW3:
This enables the event tests also for CUDA (currently only available for CPU).
The existing tests are using a special event `EventHostManualTrigger` that streams can wait on and which can be singnaled to be ready by the host. This is used to create specific deterministic scenarios.
This PullRequest adds a version of this event for CUDA to the test code. It is implemented via the CUDA 8.0 driver API's `cuStreamWaitValue32` method which is not available in the runtime API.

@psychocoderHPC Could you please run the `test/unit/event` and find out if it does produce any errors when the CUDA backend is enabled? At the moment I do not have access to a linux machine with NVIDA hardware and CUDA 8. The CI only tells me that it builds correctly...
I do not know the PIConGPU event system and the execution dependencies in detail but such a host triggered event might also be a solution to the polling problem.


BenjaminW3:
@psychocoderHPC Could you please review the code and execute the test on a NVIDA GPU using CUDA 8 (see above)?

psychocoderHPC:
I will test it tomorrow.

BenjaminW3:
push

psychocoderHPC:
sry I was busy I try to test it today

psychocoderHPC:
if I compiling with CUDA enabled I get the following linker error

```
/opt/pkg/devel/cmake/3.7.2/bin/cmake -E cmake_link_script CMakeFiles/event.dir/link.txt --verbose=1
/opt/pkg/compiler/gnu/gcc/5.3.0/bin/g++   -g0 -O3 -m64 -fopenmp   CMakeFiles/event.dir/src/event_generated_EventTest.cpp.o CMakeFiles/event.dir/src/event_generated_main.cpp.o  -o event -Wl,-rpath,/opt/pkg/devel/boost/1.62.0/gnu/5.3.0/64/opt/lib /opt/pkg/devel/cuda/8.0/lib64/libcudart_static.a -lpthread -ldl /usr/lib/x86_64-linux-gnu/librt.so libcommon.a /opt/pkg/devel/boost/1.62.0/gnu/5.3.0/64/opt/lib/libboost_unit_test_framework.so libalpaka.a /opt/pkg/devel/cuda/8.0/lib64/libcudart.so -lpthread -lrt -fopenmp /opt/pkg/devel/cuda/8.0/lib64/libcuda.so 
/usr/lib/x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
collect2: error: ld returned 1 exit status
make[2]: *** [event] Error 1
```

Without the CUDA backend this PR builds without issues.

psychocoderHPC:
For CUDA `#define BOOST_TEST_DYN_LINK` must be added about `#define BOOST_TEST_MODULE event` to solve the linker issue.

Fix: 
```C++
#   define BOOST_TEST_DYN_LINK
#endif
```

psychocoderHPC:
After fixing the build the test case is crashing on a K20:

```
Running 20 test cases...
/bigdata/hplsim/scratch/widera/alpaka/test/unit/event/src//EventTest.cpp(140): fatal error: in "event/eventReEnqueueShouldBePossibleIfNobodyWaitsFor<St5tupleIJN6alpaka3dev9DevCudaRtENS0_6stream17StreamCudaRtAsyncEEE>": critical check false == alpaka::event::test(e1) has failed [false != true]
/bigdata/hplsim/scratch/widera/alpaka/test/unit/event/src//EventTest.cpp(197): fatal error: in "event/eventReEnqueueShouldNotBePossibleIfSomeoneWaitsFor<St5tupleIJN6alpaka3dev9DevCudaRtENS0_6stream17StreamCudaRtAsyncEEE>": critical check false == alpaka::event::test(e1) has failed [false != true]

*** 2 failures are detected in the test module "event"
```

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
```diff
@@ -343,3 +343,346 @@ namespace alpaka
...
+                    // The following are known issues related to Events and Metrics:
+                    // * In event or metric profiling, kernel launches are blocking. Thus kernels waiting
+                    //   on host updates may hang. This includes synchronization between the host and
+                    //   the device build upon value-based CUDA stream synchronization APIs such as
+                    //   cuStreamWaitValue32() and cuStreamWriteValue32().
+                    ALPAKA_CUDA_DRV_CHECK(
+                        cuStreamWaitValue32(
+                            (CUstream)stream.m_spStreamCudaRtSyncImpl->m_CudaStream,
+                            (CUdeviceptr)event.m_spEventImpl->m_devMem,
+                            0u,
```
psychocoderHPC: IMO this must be `16843009u`

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
```diff
@@ -343,3 +343,346 @@ namespace alpaka
...
+                    // The following are known issues related to Events and Metrics:
+                    // * In event or metric profiling, kernel launches are blocking. Thus kernels waiting
+                    //   on host updates may hang. This includes synchronization between the host and
+                    //   the device build upon value-based CUDA stream synchronization APIs such as
+                    //   cuStreamWaitValue32() and cuStreamWriteValue32().
+                    ALPAKA_CUDA_DRV_CHECK(
+                        cuStreamWaitValue32(
+                            (CUstream)stream.m_spStreamCudaRtAsyncImpl->m_CudaStream,
+                            (CUdeviceptr)event.m_spEventImpl->m_devMem,
+                            0u,
```
psychocoderHPC: Could it be that this muste be `16843009u` because `cuStreamWaitValue32` waits untill `(int32_t)(*addr - value) >= 0`. 
The event internal value (*addr) is set byte wise with cudaMemset to `0x01010101` if `trigger()` is called, before it is zero.

In the current implementation the event finished after queuing because the internal value is set to zero and `0 - 0 >= 0`.

psychocoderHPC:
I think there is a mistake within the test value and the initial value. Never the les the test is not running ( this means if I changed it to my suggestion it deadlocks). 
IMO the reason is the within `trigger()` the function `cudaMemset` is used to set the device value. This will block the device until all other work before is finsihed. The reson for that is that `cudaMemset` is not asynchronous.

psychocoderHPC:
With my little number if everything is working expect the test for requesting.

https://github.com/ComputationalRadiationPhysics/alpaka/blob/eb395e90187e860a36651fd08cde8f3b34de8d39/test/unit/event/src/EventTest.cpp#L213  is deadlocking because event `e1` is re-enqueued https://github.com/ComputationalRadiationPhysics/alpaka/blob/eb395e90187e860a36651fd08cde8f3b34de8d39/test/unit/event/src/EventTest.cpp#L204. 
The test only finsihed if I call trigger on `k2` before the `wait` is called.

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
psychocoderHPC: I tested the suggested fix, and it is working (expected the later showed re-enqueing test)

BenjaminW3:
@psychocoderHPC Thank you very much for testing and debugging this code!
Do you see a reason why this new functionality (`cuStreamWaitValue32`) is not available in the runtime API?
I will push a revised version over the weekend.

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
BenjaminW3: This is correct... I was wrongly thinking that memset sets a 32 bit value ;-)

BenjaminW3: `0x01010101u` should be more idiomatic

BenjaminW3:
@psychocoderHPC Do you say that the `eventReEnqueueShouldNotBePossibleIfSomeoneWaitsFor` test is the only one that fails when using CUDA?

psychocoderHPC:
Yes only this test fails after patching the wait condition and the dynamic linking.

The reason for the fail is that the event is re enqueued.

Am 23. September 2017 11:37:45 MESZ, schrieb Benjamin Worpitz <notifications@github.com>:
>@psychocoderHPC Do you say that the
>`eventReEnqueueShouldNotBePossibleIfSomeoneWaitsFor` test is the only
>one that fails when using CUDA?
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/ComputationalRadiationPhysics/alpaka/pull/396#issuecomment-331622979

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

BenjaminW3:
You are right, the behaviour of the CPU events is not identical to the CUDA events. I will have to change the test, to allow re-enqueuing even if someone waits for it. Furthermore, the CPU event implementation will have to be adapted.

psychocoderHPC:
but with reenqueing we can create deadlocks. We need to take care that this is not the case.

Am 23. September 2017 18:16:27 MESZ, schrieb Benjamin Worpitz <notifications@github.com>:
>You are right, the behaviour of the CPU events is not identical to the
>CUDA events. I will have to change the test, to allow re-enqueuing even
>if someone waits for it. Furthermore, the CPU event implementation will
>have to be adapted.
>
>-- 
>You are receiving this because you were mentioned.
>Reply to this email directly or view it on GitHub:
>https://github.com/ComputationalRadiationPhysics/alpaka/pull/396#issuecomment-331647421

-- 
Diese Nachricht wurde von meinem Android-Mobiltelefon mit K-9 Mail gesendet.

psychocoderHPC:
@BenjaminW3 Does the new changes have the risk to create dead looks while  re-enqueuing?

BenjaminW3:
Have you seen this while testing? Travis has found a synchronization issue that has to be fixed.

psychocoderHPC:
I have seen it before your changes. If I re-enqueue to the same stream it deadlock.  I will check the current changes tomorrow.

We need also check that we not reintroduce #388.

BenjaminW3:
There is a unit test that tests exactly the scenario from #388. So this can not happen again. However, in the current develop version, it was not possible to re-enqueue a CPU event that was still enqueued in a stream when someone was waiting for it.
I am currently fixing this within this Pull Request but there is still an error which I will investigate.

BenjaminW3:
I think I have found my bug.
@psychocoderHPC Could you please test the CUDA version again? The CPU event now can be re-enqueued even though someone is waiting on it. This is allowed by CUDA, but was not so on the CPU before. However, I am not sure what happens with such waits. I have implemented now, that they wait until the reenqueued event has been processed. I can not verify that this is equal to what CUDA does.

BenjaminW3:
push

psychocoderHPC:
I will test it soon, I am busy because of the big hill of work before me

psychocoderHPC:
I tested the latest commit, it is running, is it still WIP?

BenjaminW3:
If this runs successfully with CUDA 8+, then it is not WIP anymore and the CPU implementation should now match that of CUDA more closely. So this can be merged!

psychocoderHPC:
Could you please rebase this branch.

BenjaminW3:
Rebased and ready to be merged!

psychocoderHPC:
CI say's nooooooo ^^

BenjaminW3:
One job timed out because it was toooo slow... re-running ...

PR #397, opened 2017-09-07
refactor .travis.yml

BenjaminW3:
This prepares the usage of the preparation, install and test scripts outside of travis, e.g. to build or within a docker image.
* moved the content of the `.travis.yml` file into individual scripts (mainly a refactoring)
* use the `generic` travis image because we did not use any of the preinstalled compilers
* document that nvcc 8.0 supports older clang versions that we already tested
* all shell scripts are now checked with ShellCheck directly within the CI
  * this found an syntactic error that disabled the sanitizer builds. This has been fixed and some minor issues have been fixed and external issues (leaks within TBB and OpenMP) have been suppressed.


BenjaminW3:
ready to merge!

PR #398, opened 2017-09-08
AtomicStlLock: increase perfromance

psychocoderHPC:
This pull requests adds a table of mutex to AtomicStlLock to increase the concurrency. The size of teh table can be changed and is currently set to the magic number 16. 16 is a trait of between concurrency and size of used cache for the mutex array. 16 mutex each 40 bytes is 640 byte which will not eat the L1 but provide a good performance.

- use an array of mutex to increase concurrency
- add static hash method (non collision less)
- add documentation
- use AtomicStlLock with hash table size `16`

# Performance

I used PIConGPU with the KelvinHelmholtz example and the alpaka backend `AccCpuOmp2Threads`   on a 4 socket `AMD Opteron(TM) Processor 6276` (64 cores in total). I changed the atomic operation used on the block and threads level for the tests. The test will use many concurrent atomic operation between a group of 256 openMP threads.  I will show the total time used to performe 10 time steps.

| atomic | time [s] |
|:---:|:---:|
|AtomicOmpCritSec | 353 |
| AtomicStlLock(unchanged) | 303 |
| AtomicStlLock<4> | 239 |
| AtomicStlLock<8> | 225 |
| AtomicStlLock<16> | 194 |
| AtomicStlLock<32> | 197 |

# Why not using std::atomic<T>

To use std atomic the type must be defined as std::atomic. It is not allowed to cast a type e.g. `float` to `std::atomic<float>`. Therefore the usage for arbitrary data is not possible.

# Why is an own hash function used and not std::hash<Ptr*>

I checked the implementation of the std hash function and the implementation for a pointer is only a `reinterpret_cast`. This will produce strides between the hash indices and if we map than the hash to a element within the mutex array we will not use each element in the array which reduce the efficiency.
The hash function is not collision free for pointer where the type is not a power of two. This will be no problem for the usage but guarantee that we can always use all elements of the mutex array.
Du to the special type of hashing function I have not moved the implementation to the core of alpaka.

# Note

If have not removed the usage of `AtomicOmpCritSec` because I will add new openMP atomics without a critical section to give the openMP implementation the possibility to choose the best fitting atomic operation for the used architecture.

### DIFF
--- a/include/alpaka/atomic/AtomicStlLock.hpp
+++ b/include/alpaka/atomic/AtomicStlLock.hpp
@@ -24,6 +24,7 @@
 #include <alpaka/atomic/Traits.hpp>                 // AtomicOp
 
 #include <mutex>                                    // std::mutex, std::lock_guard
+#include <array>
 
 namespace alpaka
 {
@@ -34,7 +35,11 @@ namespace alpaka
         //
         //  Atomics can be used in the grids, blocks and threads hierarchy levels.
         //  Atomics are not guaranteed to be save between devices.
+        //
+        // \tparam THashTableSize size of the hash table to allow concurrency between
+        //                        atomics to different addresses
         //#############################################################################
+        template<size_t THashTableSize>
         class AtomicStlLock
         {
         public:
@@ -45,6 +50,34 @@ namespace alpaka
                 typename THierarchy,
                 typename TSfinae>
             friend struct atomic::traits::AtomicOp;
+
+            static constexpr size_t nextPowerOf2(size_t const value, size_t const bit = 0)
+            {
+                return value <= (1u << bit) ? (1u << bit) : nextPowerOf2(value, bit + 1u);
+            }
+
+            //-----------------------------------------------------------------------------
+            //! get a hash value of the pointer
+            //
+            // This is no perfect hash, there will be collisions if the size of pointer type
+            // is not a power of two.
+            //-----------------------------------------------------------------------------
+            template<typename TPtr>
+            static size_t hash(TPtr const * const ptr)
+            {
+                size_t const ptrAddr = reinterpret_cast< size_t >( ptr );
+                // using power of two for the next division will increase the performance
+                constexpr size_t typeSizePowerOf2 = nextPowerOf2(sizeof(TPtr));
+                // division removes the stride between indices
+                return (ptrAddr / typeSizePowerOf2);
+            }
+
+            //-----------------------------------------------------------------------------
+            //! get the size of the hash table
+            //
+            // The size is at least 1 or THashTableSize rounded up to the next power of 2
+            //-----------------------------------------------------------------------------
+            static constexpr size_t hashTableSize = THashTableSize == 0 ? 1u : nextPowerOf2(THashTableSize);
 
             //-----------------------------------------------------------------------------
             //! Default constructor.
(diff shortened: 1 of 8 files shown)


include/alpaka/atomic/AtomicStlLock.hpp
```diff
@@ -71,10 +104,14 @@ namespace alpaka
...
-            std::mutex & getMutex() const
+            template<typename TPtr>
+            std::mutex & getMutex(TPtr const * const ptr) const
             {
-                static std::mutex m_mtxAtomic; //!< The mutex protecting access for a atomic operation.
-                return m_mtxAtomic;
+                size_t const hashedAddr = hash(ptr) & (hashTableSize - 1u);
+                static std::array<
+                    std::mutex,
+                    hashTableSize> m_mtxAtomic; //!< The mutex protecting access for a atomic operation.
```
BenjaminW3: for a**n** atomic operation

include/alpaka/atomic/AtomicStlLock.hpp
```diff
@@ -46,6 +51,34 @@ namespace alpaka
                 typename TSfinae>
             friend struct atomic::traits::AtomicOp;
 
+            static constexpr size_t nextPowerOf2(size_t const value, size_t const bit = 0)
+            {
+                return value <= (1u << bit) ? (1u << bit) : nextPowerOf2(value, bit + 1u);
```
BenjaminW3: Visual Studio 2017 says: 
```
C:\projects\alpaka\include\alpaka/atomic/AtomicStlLock.hpp(56): warning C4334: '<<': result of 32-bit shift implicitly converted to 64 bits (was 64-bit shift intended?)
```
I think 1u is interpreted as 32 bit. Maybe a `static_cast<size_t>(1u)` might help?

include/alpaka/atomic/AtomicStlLock.hpp
```diff
@@ -46,6 +51,34 @@ namespace alpaka
...
+                // division removes the stride between indices
+                return (ptrAddr / typeSizePowerOf2);
+            }
+
+            //-----------------------------------------------------------------------------
+            //! get the size of the hash table
+            //
+            // The size is at least 1 or THashTableSize rounded up to the next power of 2
+            //-----------------------------------------------------------------------------
+            static constexpr size_t hashTableSize = THashTableSize == 0 ? 1u : nextPowerOf2(THashTableSize);
```
BenjaminW3: I think we should better simply `static_assert(THashTableSize > 0, "bla")`

BenjaminW3:
Thank you for implementing this optimization!
Am I right that with these changes PIConGPU runs 45% faster? (at least this example)

include/alpaka/atomic/AtomicStlLock.hpp
BenjaminW3: clang has a problem with this line 
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/atomic/AtomicStlLock.hpp:80:80: error: 
      use of default argument to function 'nextPowerOf2' that is declared later
      in class 'AtomicStlLock'
  ...constexpr size_t hashTableSize = THashTableSize == 0 ? 1u : nextPowerOf2...
                                                                 ^
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/atomic/AtomicStlLock.hpp:54:89: note: 
      default argument declared here
  ...constexpr size_t nextPowerOf2(size_t const value, size_t const bit = 0)
                                                                          ^
1 error generated.
```

psychocoderHPC:
Yes for this example we get an factor 2. But `AccCpuOmp2Threads` is not the typical used accelerator but perfect to test atomic operations.

psychocoderHPC:
Yes it is a factor two faster with `AccCpuOmp2Threads`.  But all in all `AccCpuOmp2Blocks` is a factor 40 faster because atomic in the hierarchy `Threads` are no atomic operations.

PR #402, opened 2017-09-13
fix OpenMP-4.0 test errors with clang

BenjaminW3:
This adds support for the OpenMP 4 backend for clang 3.9, 4 and 5.
Without the `thread_limit` annotation on the `omp teams` clause, clang did not provide enough threads to be available for the inner `omp parallel` clause. gcc did not require this but the code was already existing but commented out.
The data size of one test had to be reduced because the OpenMP 4 implementation of clang is extremely slow...


BenjaminW3:
ready to be merged!

PR #404, opened 2017-09-14
add more kernel unit tests

BenjaminW3:
* kernel with constructor and member
* additional parameter for kernel function objects operator()
* addititional template parameter for kernel function objects operator()
* templated kernel function object

and removed those cases that were previously part of the sharedMem integration test


BenjaminW3:
ready to be merged!

BenjaminW3:
push!

PR #406, opened 2017-09-14
remove uuid and clean up

BenjaminW3:
* remove uuid from CPU event and streams because it is unused. It was used in a very old version when the event implementations themselves had not been shared_ptr`s.
* replace comparison of underlying CUDA event/stream with comparison of impl shared_ptr (one less indirection and a raw pointer comparison instead of an unknown `operator==`) -> now identical to CPU
* rename all stream impl members to shorter and identical names for clearness. The same has already been done for the event impls

This is best reviewed commit by commit but everything belongs thematically together.


BenjaminW3:
@psychocoderHPC Could you please review this so that it can be merged when the builds were successfull?

PR #407, opened 2017-09-16
dockerize build

BenjaminW3:
Currently (without using docker) only the downloaded CUDA, clang and CMake versions are cached.

This PR creates a docker image which includes everything that is required to build a specific job (clang, CMake, CUDA, boost, boost libs build from those sources, everything installed via apt-get (git, gcc, libtsan, libtbb, libomp, shellcheck, sloccount, ...).

This is not yet a clean solution using a Dockerfile but instead uses a base ubuntu image, starts up a container, installs everything using the existing shell scripts and finally saves the containers image into the cache.

Saving the whole image for a job has multiple advantages:
* if it is already in the cache, the build is faster (6h instead of 7h accumulated build job times)
* less likely to fail due to network/apt-get/... install failures
* allows to test other distributions (e.g. ubuntu 16.04 which is required for CUDA 9.0)
* allows to export the docker image for usage outside of travis (future extension)


.travis.yml
```diff
@@ -74,62 +77,56 @@ env:
         # gcc 6 ASan is triggered within libtbb.so
         # gcc 7 ASan introduced 'stack-use-after-scope' which is triggered by GOMP_parallel
         # Analysis builds
-        - CXX=g++     CC=gcc   ALPAKA_GCC_VER=4.9     CMAKE_BUILD_TYPE=Debug   ALPAKA_CI_BOOST_BRANCH=boost-1.62.0 ALPAKA_CI_CMAKE_VER=3.9.1 ALPAKA_CI_SANITIZERS=                ALPAKA_CI_ANALYSIS=ON  ALPAKA_DEBUG=2 ALPAKA_ACC_GPU_CUDA_ENABLE=ON  ALPAKA_CUDA_VER=7.5 ALPAKA_CUDA_COMPILER=nvcc
-        - CXX=g++     CC=gcc   ALPAKA_GCC_VER=7       CMAKE_BUILD_TYPE=Debug   ALPAKA_CI_BOOST_BRANCH=boost-1.65.1 ALPAKA_CI_CMAKE_VER=3.7.2 ALPAKA_CI_SANITIZERS=                ALPAKA_CI_ANALYSIS=ON  ALPAKA_DEBUG=2
+        #- CXX=g++     CC=gcc   ALPAKA_GCC_VER=4.9     CMAKE_BUILD_TYPE=Debug   ALPAKA_CI_BOOST_BRANCH=boost-1.62.0 ALPAKA_CI_CMAKE_VER=3.9.1 ALPAKA_CI_SANITIZERS=                ALPAKA_CI_ANALYSIS=ON  ALPAKA_DEBUG=2 ALPAKA_ACC_GPU_CUDA_ENABLE=ON  ALPAKA_CUDA_VER=7.5 ALPAKA_CUDA_COMPILER=nvcc
```
psychocoderHPC: Why do you disabling all the tests?

psychocoderHPC: sry I missed that this is WIP

BenjaminW3:
ready to be merged!

PR #384, opened 2017-08-21
CUDA 9.0

BenjaminW3:
#376

* fixes many warnings of the form `warning : __host__ annotation on a defaulted function("~IdxBtZero") is ignored` by removing the annotation
* The CUDA 9.0 package that NVIDIA provides for Ubuntu 16.04 can successfully be installed on Ubuntu 14.04 ;-) So the previous Pull Request might not have been necessary...
* CUDA 9.0 supports GCC 6 but this combination fails to compile alpaka correctly. This may have to be investigated in the future when this combination is desired.

PR #410, opened 2017-10-01
use newer ubuntu images in CI

BenjaminW3:
This is a preparation for CUDA 9.0 support (#384) because it does not provide a package for ubuntu 14.04 anymore which we are currently using because it is the Travis default image.
For now Ubutnu 16.04 and 17.04 are used for some of the CPU only test jobs.


BenjaminW3:
ready to be merged!

PR #413, opened 2017-10-10
Remove typedef

ax3l:
Replace the only found two typedef with C++11 using.


BenjaminW3:
Thanks! I wonder why there is no compiler option to generate warnings when typedef is used. `-wPreferUsingOverTypedef`. There are some more things I would like to enforce by the compiler.

ax3l:
Agreed! 
But at least Clang-Tidy has an option for that.
Just experimented with it in https://github.com/ComputationalRadiationPhysics/picongpu/pull/2303

ax3l:
P.S.: the term "typedef" is still used at several places in the docs.

PR #415, opened 2017-10-11
test event: fix broken linker step

psychocoderHPC:
fix linker error if `ALPAKA_ACC_GPU_CUDA_ENABLE` is enabled.

This bug was introduced with #396 and mentioned in post https://github.com/ComputationalRadiationPhysics/alpaka/pull/396#issuecomment-331438871, but I missed that this change is needed do to my environment was not fully clean and this patch was always used in all of my tests in #396.

```
/usr/lib/x86_64-linux-gnu/crt1.o: In function `_start':
(.text+0x20): undefined reference to `main'
collect2: error: ld returned 1 exit status
make[2]: *** [event] Error 1
make[1]: *** [CMakeFiles/event.dir/all] Error 2
make: *** [all] Error 2
```

- [ ] do not merge this will maybe fixed with #417

### DIFF
--- a/test/unit/event/src/main.cpp
+++ b/test/unit/event/src/main.cpp
@@ -18,7 +18,9 @@
  * along with alpaka.
  * If not, see <http://www.gnu.org/licenses/>.
  */
-
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+#   define BOOST_TEST_DYN_LINK
+#endif
 #define BOOST_TEST_MODULE event
 
 #include <boost/predef.h>   // BOOST_COMP_CLANG


BenjaminW3:
The build of this Pull Request is failing with:
`CMakeFiles/event.dir/src/event_generated_main.cpp.o: In function `main':
/home/travis/boost/boost/test/unit_test.hpp:63: undefined reference to `boost::unit_test::unit_test_main(bool (*)(), int, char**)'`

I do not know all the details right now, but the CI currently builds boost as static libs and uses `Boost_USE_STATIC_LIBS` to build alpaka and its tests.
The CMake scripts should automatically add the `BOOST_TEST_DYN_LINK` option when `Boost_USE_STATIC_LIBS` is not given see [here](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/test/unit/event/CMakeLists.txt). However, this is not tested and may be defective. This should be fixed globally for all tests.

psychocoderHPC:
I found out that all other tests on our cluster (hypnos) with CUDA 8.0 have the some issue. I will modify this PR to cover all test cases.

psychocoderHPC:
@BenjaminW3 
Because of unknown reason the definition is not passed to the build process
```
# is not working
LIST(APPEND _DEFINITIONS_PRIVATE "-DBOOST_TEST_DYN_LINK")
```

I am currently checking the issue, where is the right place to add a define for all test cases?

BenjaminW3:
Currently not all unit tests depend on boost. So those which have the dependency, find boost and set the falgs themselves...

BenjaminW3:
What I wonder right now is, why it works when CUDA is not used?

psychocoderHPC:
I found out that

```
TARGET_COMPILE_OPTIONS(
    ${_TARGET_NAME}
    PRIVATE ${_DEFINITIONS_PRIVATE})
```
is not passed to nvcc. Thats the reason why [this cmake flag](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/test/unit/event/CMakeLists.txt#L56-L58) as no influence on nvcc.

I found the solution for the missing main function somewhere on stackoverflow, I think it was [this](https://stackoverflow.com/questions/33644914/boost-test-cannot-find-main) post.

psychocoderHPC:
@BenjaminW3 I opened an issue #417 for the definition which can not passed to `nvcc` we can evaluate there how we can fix it.

psychocoderHPC:
I canceled the travis CI test to allow other travis tests running.

psychocoderHPC:
is fixed with #421 for all tests

PR #416, opened 2017-10-11
host `constexpr` within device code

psychocoderHPC:
- add nvcc compiler flag to allow usage of host constexpr functions within device code
- add test case to validate host constexpr usage within the device

# Note
The clang cuda compiler support this feature by default (tested with clang 4.0)

- [ ] do not merge (I added a test commit to disable `-G`)

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -350,6 +350,10 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 IF(NOT CUDA_VERSION VERSION_LESS 7.5)
                     LIST(APPEND CUDA_NVCC_FLAGS "--expt-extended-lambda")
+                    LIST(APPEND CUDA_NVCC_FLAGS "--expt-relaxed-constexpr")
+                ELSE()
+                    # CUDA 7.0
+                    LIST(APPEND CUDA_NVCC_FLAGS "--relaxed-constexpr")
                 ENDIF()
 
                 FOREACH(_CUDA_ARCH_ELEM ${ALPAKA_CUDA_ARCH})
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -369,10 +369,10 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 IF(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
                     LIST(APPEND CUDA_NVCC_FLAGS "-g")
                     # https://github.com/ComputationalRadiationPhysics/alpaka/issues/428
-                    IF(CMAKE_COMPILER_IS_GNUCXX AND
-                       CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0 AND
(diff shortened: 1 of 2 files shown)


BenjaminW3:
On the CPU only builds you get: `KernelWithHostConstexpr.cpp:60:22: error: unused parameter 'acc' [-Werror=unused-parameter]` and even if this is solved you get `KernelWithHostConstexpr.cpp:59:24: error: but 'void kernel::KernelWithHostConstexpr::operator()(const TAcc&) const' does not throw; perhaps it should be declared 'noexcept' [-Werror=noexcept]`
I have used
```
// Do something useless on the accelerator.
alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Blocks>(acc);
```
within the other unit tests to solve both warnings at once.

BenjaminW3:
The CUDA only builds are ok (when the above issue is fixed).
However, the heterogeneous CPU/GPU builds fail with some hard to parse error...

psychocoderHPC:
I canceled the CI test by hand because the tests failed.

psychocoderHPC:
@BenjaminW3 Can you see why the test https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/290456792 is not passing. I cant find any error but the result of the test is `false`. I removed for this test the flag `-G`.

psychocoderHPC:
could it be that this is the reason why it fails?
```
gcc version 5.4.1 20160904 (Ubuntu 5.4.1-2ubuntu1~16.04)
: ${ALPAKA_ACC_GPU_CUDA_ONLY_MODE?"ALPAKA_ACC_GPU_CUDA_ONLY_MODE must be specified"}
: ${ALPAKA_CUDA_VER_MAJOR?"ALPAKA_CUDA_VER_MAJOR must be specified"}
: ${ALPAKA_CUDA_VER_MINOR?"ALPAKA_CUDA_VER_MINOR must be specified"}
# https://stackoverflow.com/questions/42218009/how-to-tell-if-any-command-in-bash-script-failed-non-zero-exit-status
err=0
trap 'err=1' ERR
```

BenjaminW3:
The `idx` test `MapIdx.cpp` fails to compile with
```
/home/travis/boost/boost/mpl/for_each.hpp(78): error: identifier "Typeinfo for ::std::integral_constant<unsigned long, (unsigned long)1ul> " is undefined in device code
```
This test is missing the
```
// \Hack: Boost.MPL defines BOOST_MPL_CFG_GPU_ENABLED to __host__ __device__ if nvcc is used.
// BOOST_AUTO_TEST_CASE_TEMPLATE and its internals are not GPU enabled but is using boost::mpl::for_each internally.
// For each template parameter this leads to:
// /home/travis/build/boost/boost/mpl/for_each.hpp(78): warning: calling a __host__ function from a __host__ __device__ function is not allowed
// because boost::mpl::for_each has the BOOST_MPL_CFG_GPU_ENABLED attribute but the test internals are pure host methods.
// Because we do not use MPL within GPU code here, we can disable the MPL GPU support.
#define BOOST_MPL_CFG_GPU_ENABLED
```
at the top equal to the other tests. I do not know if this is the reason, but it is my best guess for now.

PR #420, opened 2017-10-15
only allow Vec creation with correct types

BenjaminW3:
This had long been lying around on my hard drive.
Currently the vector constructor only checks if the first argument is of the specified type. This is inconsequent and is fixed to require all argument types to be identical.


psychocoderHPC:
I will check this and the other PR at the end of this week or the next. I am on an business trip.

BenjaminW3:
nvcc does not seem to support this.

Issue #428, opened 2017-10-20
CUDA 8 and gcc 4.9.2 not compatible


psychocoderHPC:
With gcc 4.9 and CUDA 8 the compiler crash if -DCMAKE_BUILD_TYPE=Debug is set (like in the CI tests).

Tested exmaple `bufferCopy`

psychocoderHPC:
The problem is the compiler flag -G for the cuda compiler, this will disable all optimizations. Maybe -G also increase the compile time of all tests.

psychocoderHPC:
I found the trigger for this issue in #427, but fixing the wrong used stream not solve this issue.

psychocoderHPC:
@BenjaminW3 I thinking about disabling the support for gcc 4.9.2 in combination with CUDA 8+.
A point against 4.9.2 is that full and correct C++11 is only supported since gcc 5.x.

BenjaminW3:
I am always for minimizing the supported compiler version. By removing support for gcc 4.9.2 we would also lose support for CUDA 7.0 and 7.5. How widespread is CUDA 8 on the relevant HPC and dev systems to   support such a change? Currently we have two work arounds in the code base specific to gcc 4.9.

psychocoderHPC:
But we can also only disable gcc 4.x if CUDA 8+ is used.

CC-ing: @ax3l

psychocoderHPC:
@BenjaminW3 Do you know why the buffer test passed the CI tests with CUDA8 and gcc 4.9? Do we check this combination?

BenjaminW3:
But then we are still limited to the C++11 that gcc 4.9.2 supports.

psychocoderHPC:
We can do a release within the next week and than remove the bad compiler^^

BenjaminW3:
CUDA 8 & gcc 4.9 should be checked in [job#4](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/.travis.yml#L92) but this is a Debug build and not a release build.

psychocoderHPC:
is this testing all the examples?

BenjaminW3:
The examples are build in all cases except when `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is set, which is not the case.
My current PR build: https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/290435436
Using gcc 4.9.4 and CUDA 8.0.
[...]

PR #431, opened 2017-10-20
Do not add -G on Debug when using gcc 4.9 and CUDA 8

ax3l:
For alpaka supported GCC versions less then 5.X, do not automatically add device debug symbols with `-G`.

This also warns the user verbosely when tried. See report in #428.

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -366,8 +366,17 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 SET(CUDA_HOST_COMPILER "${CMAKE_CXX_COMPILER}")
 
-                if(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
-                    LIST(APPEND CUDA_NVCC_FLAGS "-g" "-G")
+                IF(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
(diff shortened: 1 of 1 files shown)


alpakaConfig.cmake
```diff
@@ -366,8 +366,17 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
                 SET(CUDA_HOST_COMPILER "${CMAKE_CXX_COMPILER}")
 
-                if(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
-                    LIST(APPEND CUDA_NVCC_FLAGS "-g" "-G")
+                IF(CMAKE_BUILD_TYPE STREQUAL "Debug" OR CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo")
+                    LIST(APPEND CUDA_NVCC_FLAGS "-g")
+                    # https://github.com/ComputationalRadiationPhysics/alpaka/issues/428
+                    IF(CMAKE_COMPILER_IS_GNUCXX AND
+                       CMAKE_CXX_COMPILER_VERSION VERSION_LESS 5.0 AND
+                       CUDA_VERSION VERSION_EQUAL 8.0)
```
BenjaminW3: May we need to expand this to CUDA 9+?

ax3l: @psychocoderHPC did not see issues with CUDA 9.0 as he just told me

psychocoderHPC: I tested it with CUDA 9. CUDA 9 is not effected.

PR #433, opened 2017-10-21
reduce build time

BenjaminW3:
There have been fixed multiple issues and some additional changes:
* added build time stamps to the build log (moreutils ts)
* the `ALPAKA_CI` environment variable was not set. It is used to reduce the number of combinations/types tested/compiled.
* the CMake download was infunctional and always used the preinstalled version
* removed `UBsan` from the combined `ASan+UBSan` build
* removed the `ESan` build because it crashed sometimes and did not provide any useful information
* added `set -o pipefail` to correctly return errors from piped commands
* added `set -u` to error on usage of undefined bash variables to most scripts. This found many incorrect usages.

All of this should decrease thetest execution time and make future debugging easier.


BenjaminW3:
I have found some issues up to now:
* the CMake download was infunctional and always used the preinstalled version
* the `ALPAKA_CI` environment variable was not set. It is used to reduce the numberof combinations/types tested/compiled. This should solve our problem.

BenjaminW3:
Ready to be merged. Total sequential CI time reduced by half an hour from ~6.5 h to 6 h.

PR #434, opened 2017-10-24
add clang declval implementation

psychocoderHPC:
- add for clang CUDA compiler a `declval` implementation (can be used on host and device)
- fallback to `sdt::declval` for non clang compiler

Note: We need this function within PIConGPU but this is a so fundamental method that alpaka is the right place to hold the implementation.

std::declval can not be used on the device if CUDA code is compiled with clang
```
c++/5.3.0/type_traits:2202:5: note: candidate function not viable: call to __host__ 
function from __device__ function
    declval() noexcept
```

### DIFF
--- a/include/alpaka/core/Utility.hpp
+++ b/include/alpaka/core/Utility.hpp
@@ -1,0 +1,53 @@
+/**
+* \file
+* Copyright 2017 Rene Widera
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
+* it under the terms of the GNU Lesser General Public License as published by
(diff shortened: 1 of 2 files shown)


include/alpaka/core/Utility.hpp
```diff
@@ -0,0 +1,53 @@
...
+#   include <utility> // std::declval
+#endif
+
+
+namespace alpaka
+{
+    namespace core
+    {
+        //-----------------------------------------------------------------------------
+        //! convert any type to a reverence type
```
ax3l: typo: reference

PR #435, opened 2017-10-24
Do not add -G on Debug

psychocoderHPC:
This PR is a follow up of #431 and removes the
usage of `-G` for
  - GCC < 5.0 and CUDA < 9
  - Clang < 3.8 AND CUDA < 9

This change is needed to allow the usage of host side constexpr functions
within the device code #416.


psychocoderHPC:
@BenjaminW3 Could you please trigger the appveyor tests again, something went wrong or they have issues.

PR #438, opened 2017-10-26
fix: crash if no available CUDA device

psychocoderHPC:
If the CUDA backe-end is enabled and `getDevCount()` is called and there
is no CUDA device available than alpaka will exit with an error.
This is not necessary because if tehre is no device available than the number of
devices is zero.

- return zero if there is no device available

### DIFF
--- a/include/alpaka/pltf/PltfCudaRt.hpp
+++ b/include/alpaka/pltf/PltfCudaRt.hpp
@@ -90,7 +90,9 @@ namespace alpaka
                     ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
                     int iNumDevices(0);
-                    ALPAKA_CUDA_RT_CHECK(cudaGetDeviceCount(&iNumDevices));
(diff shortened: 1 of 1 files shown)


include/alpaka/pltf/PltfCudaRt.hpp
```diff
@@ -90,7 +90,9 @@ namespace alpaka
                     ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
                     int iNumDevices(0);
-                    ALPAKA_CUDA_RT_CHECK(cudaGetDeviceCount(&iNumDevices));
+                    cudaError_t error = cudaGetDeviceCount(&iNumDevices);
+                    if(error != cudaSuccess)
+                        iNumDevices = 0;
```
psychocoderHPC: Note: This line is needed because it is not clear if CUDA is changing the parameter of `cudaGetDeviceCount` in the case where the return value is an error.

Issue #419, opened 2017-10-14
style changes to prepare automatic reformatting


BenjaminW3:
I want to add automatic code formatting via clang-format to alpaka. It is not perfect, but it is easier and more consistent to rely on automatic formatting than to do it all by hand.
Before doing this, I want to apply some changes to the current style and would like to get your feedback:
* I want to reduce the funtion and class comment headers in size to get more real code per line without sacrificing the seperating function of the headers.
  I want to replace 
  ```c++
  //#############################################################################
  //! The CPU OpenMP 2.0 thread accelerator executor.
  //#############################################################################
  ```
  by
  ```c++
  //#############################################################################
  //! The CPU OpenMP 2.0 thread accelerator executor.
  ```
  and do the same for the function headers

* I want to fully remove the useless comment header for `Copy constructor`, `Copy assignment operator`, ...

* ~I want to remove the comments after the includes.~
  ```
  #include <alpaka/acc/Traits.hpp> // acc::traits::AccType
  ```
  ~They were helpful in the past but tend to get out-of-sync. Furthermore, clang-format does not support to align them in the way the are aligned now, which looks really strange.~

Edit: point 3 has been done

BenjaminW3:
any comments?

ax3l:
I like that and the points above look all very good to me ✨ 
We can also check-in the clang-format config file into https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md if it gets close enough :)

psychocoderHPC:
@BenjaminW3  I agree with all point expect 
```
//#############################################################################
//! The CPU OpenMP 2.0 thread accelerator executor.
```

I would like also to remove `//###########`

ax3l:
@BenjaminW3 did you succeed with a first `.clang-format` file draft?

BenjaminW3:
I already have a draft but I wanted to wait for clang 6 because it will introduce some more good formatting features.

ax3l:
Oh that's nice. Just for the courious, do you want to share the draft as a `[WIP]` in a PR here or in https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md ?

PR #439, opened 2017-10-31
remove comment seperators

BenjaminW3:
Removes the useless second comment seperator line as well as useless comments see #419.


psychocoderHPC:
Why is the first comment separator needed? What is the advantage of it?

BenjaminW3:
For now I prefer to have the seperating effect of the single lines as written in the corresponding ticket. Removing them later should be a very simple search and replace task.

PR #443, opened 2017-11-08
Improved bufferCopy example

theZiz:
I added support for non-CPU accelerators to the bufferCopy example.
While doing so I noticed that the example did not handle the possible pitch
of 2D or 3D memory correctly and added basic support for this.
The example can be steered for different dimensions and accelerators.

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -26,7 +26,18 @@
 #include <cstdint>
 #include <cassert>
 
+#define DIMENSION 3
+
 //#############################################################################
+
+template <size_t width>
+ALPAKA_FN_ACC size_t logIdToPitchedId(size_t const globalId,size_t const pitch)
+{
+    const size_t id_x = globalId % width;
+    const size_t id_y = globalId / width;
+    return id_x + id_y * pitch;
+}
+
 //! Prints all elements of the buffer.
 struct PrintBufferKernel
 {
@@ -38,7 +49,8 @@ struct PrintBufferKernel
     ALPAKA_FN_ACC auto operator()(
         TAcc const & acc,
         TData const * const buffer,
-        TExtent const & extents) const
+        TExtent const & extents,
+        size_t const pitch) const
     -> void
     {
         auto const globalThreadIdx = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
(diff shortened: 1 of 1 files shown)


BenjaminW3:
looks good!

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -51,7 +63,7 @@ struct PrintBufferKernel
         for(size_t i(linearizedGlobalThreadIdx[0]); i < extents.prod(); i += globalThreadExtent.prod())
         {
             // NOTE: hard-coded for unsigned int
-            printf("%u ", buffer[i]);
+            printf("%u:%u ", static_cast<int32_t>(i), buffer[logIdToPitchedId<2>(i,pitch)]);
```
BenjaminW3:
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/example/bufferCopy/src/bufferCopy.cpp:66:19: error: format '%u' expects argument of type 'unsigned int', but argument 2 has type 'int32_t {aka int}' [-Werror=format=]
             printf("%u:%u ", static_cast<int32_t>(i), buffer[logIdToPitchedId<2>(i,pitch)]);
```
You might need a `static_cast<int32_t>` for the second parameter as well.

theZiz: I thought about this, but didn't want to fix too much. Never touch a running system. :roll_eyes:  At least for little endian it works well this way. :stuck_out_tongue:
jk, I will change this, too.

BenjaminW3:
What does `logIdToPitchedId` stand for? Shouldn`t it be `linIdToPitchedId`?

BenjaminW3:
there is still one compiler error left. `bufferCopy.cpp(89): 'pitch': unreferenced formal parameter`. Because in debug builds the pitch is unused. You could simply add a `(void)pitch;` directly above the assert to remove this warning.

theZiz:
Well, both `globalId` and the resulting id are technically linear. So I thought, that "log" for "logical" may be a good naming. But I can also change it to `lin` if you think it fits better. I will fix the "unreferenced formal parameter" warning in the same commit.

BenjaminW3:
Personally I would prefer dimensionality suffixes. However, the code currently uses `linearizedGlobalThreadIdx` so I would use `lin` for onsistency. When looking at `log` I first had to think: "What is logarithmic in this test?"

BenjaminW3:
You are still missing the fix for `bufferCopy.cpp(89): 'pitch': unreferenced formal parameter`

theZiz:
Arg, sorry

theZiz:
Okay, I fixed it, using the same method you did for `data`.

PR #445, opened 2017-11-13
32 bit fixes

BenjaminW3:
This only fixes the compilation errors for 32 bit. This will not fix the compiler out of heap memory problems of some tests.


BenjaminW3:
Ready to be merged!

Issue #446, opened 2017-11-16
compilation problem in the integration test matMul


ivandrodri:
I have this compilation problem with the matMul example:

####################################################################

[ 80%] Building NVCC (Device) object matMul/CMakeFiles/matMul.dir/src/matMul_generated_main.cpp.o
/home/ivan/Desktop/Alpaka_git/alpaka/test/integ/matMul/src//main.cpp(297) (col. 15): internal error: assertion failed at: "/dvs/p4/build/sw/rel/gpu_drv/r361/r361_00/drivers/compiler/edg/EDG_4.10/src/il_to_str.c", line 7191

1 catastrophic error detected in the compilation of "/tmp/tmpxft_00003f69_00000000-5_main.cpp4.ii".
Compilation aborted.
Aborted (core dumped)
CMake Error at matMul_generated_main.cpp.o.cmake:282 (message):
  Error generating file
  /home/ivan/Desktop/Alpaka_git/alpaka/test/integ/build/matMul/CMakeFiles/matMul.dir/src/./matMul_generated_main.cpp.o

matMul/CMakeFiles/matMul.dir/build.make:63: recipe for target 'matMul/CMakeFiles/matMul.dir/src/matMul_generated_main.cpp.o' failed
make[2]: *** [matMul/CMakeFiles/matMul.dir/src/matMul_generated_main.cpp.o] Error 1
CMakeFiles/Makefile2:332: recipe for target 'matMul/CMakeFiles/matMul.dir/all' failed
make[1]: *** [matMul/CMakeFiles/matMul.dir/all] Error 2
Makefile:83: recipe for target 'all' failed
make: *** [all] Error 2

#####################################################################

thanks in advance for any help!

psychocoderHPC:
Could you please post which OS do you have used and which compiler, boost and CUDA (if used).
And all commands used up to the broken compile.

Am 16. November 2017 09:37:37 MEZ, schrieb ivandrodri <notifications@github.com>:
>I have this compilation problem with the matMul example:
>
>####################################################################
>
>[ 80%] Building NVCC (Device) object
>matMul/CMakeFiles/matMul.dir/src/matMul_generated_main.cpp.o
>/home/ivan/Desktop/Alpaka_git/alpaka/test/integ/matMul/src//main.cpp(297)
>(col. 15): internal error: assertion failed at:
[...]

PR #447, opened 2017-11-16
Fix nvcc 8.0 + gcc 5.4 + --expt-relaxed-constexpr

BenjaminW3:
reproduce #446


alpakaConfig.cmake
```diff
@@ -339,12 +339,23 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                    # CUDA 7.0
+                    LIST(APPEND CUDA_NVCC_FLAGS "--relaxed-constexpr")
+                ELSEIF(CUDA_VERSION VERSION_LESS 8.0)
+                    # CUDA 7.5
                     LIST(APPEND CUDA_NVCC_FLAGS "--expt-extended-lambda")
                     LIST(APPEND CUDA_NVCC_FLAGS "--expt-relaxed-constexpr")
+                ELSEIF(CUDA_VERSION VERSION_LESS 9.0)
+                    # CUDA 8.0
+                    LIST(APPEND CUDA_NVCC_FLAGS "--expt-extended-lambda")
+                    IF((NOT CMAKE_COMPILER_IS_GNUCXX) OR (CMAKE_COMPILER_IS_GNUCXX AND CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL 5))
```
psychocoderHPC: note: `--expt-relaxed-constexp` is available with gcc 5.X. 
Also it is not possible to bring this mainline else we lose that the feature that we can write the code once and run it on different platforms and compiler without `#ifdef` within the user code.

BenjaminW3: NVCC 8.0 in combination with gcc 5.4 triggers the following compiler error in all tests:
```C++
/usr/include/c++/5/tuple(484) (col. 17): error: calling a __host__ function("std::_Tuple_impl<(unsigned long)0ul,  ::std::_Bind< ::std::_Mem_fn<void (     ::alpaka::core::detail::ConcurrentExecPool<unsigned long,  ::std::thread,  ::std::promise, void,  ::std::mutex,  ::std::condition_variable, (bool)0> ::*)()>  ( ::alpaka::core::detail::ConcurrentExecPool<unsigned long,  ::std::thread,  ::std::promise, void,  ::std::mutex,  ::std::condition_variable, (bool)0>  *)>  > ::_Tuple_impl [subobject]") from a __device__ function("std::tuple< ::std::_Bind< ::std::_Mem_fn<void (     ::alpaka::core::detail::ConcurrentExecPool<unsigned long,  ::std::thread,  ::std::promise, void,  ::std::mutex,  ::std::condition_variable, (bool)0> ::*)()>  ( ::alpaka::core::detail::ConcurrentExecPool<unsigned long,  ::std::thread,  ::std::promise, void,  ::std::mutex,  ::std::condition_variable, (bool)0>  *)>  > ::tuple") is not allowed
```
So for this compiler combination the CUDA backend might be incompatible with the std::thread and fiber backends. I will try.

psychocoderHPC: It looks like https://github.com/ComputationalRadiationPhysics/alpaka/pull/416

BenjaminW3: I suppose that the error occurs because gcc 5.4 marked many more STL functions as `constepr`. However, this does not seem to be compatible with nvcc 8.0 and the functions it is replacing/overriding e.g.
`/usr/include/c++/5/cmath(362): error: calling a __host__ function("__builtin_logl") from a __device__ function("std::log") is not allowed`

I was able to replace the code that used `std::bind` with lambdas so that the CPU backends are functional again. The only two remaining errors are within the `rand` and `stream` unit tests. I will look for a workaround for the `stream` test failure but the CPU random number distributions based on the STL seem not be compatible with this compiler because of their usage of `std::log`.

PR #451, opened 2017-12-22
test CMake 3.10.1

BenjaminW3:
Now I use a more sophisticated semantic version detection because previously I used hard coded digit counts for major and minor version which did not work anymore with the minor version having two digits.


BenjaminW3:
Ready to merge!
Please ignore the appveyor build. I triggered 3 out-of-heap-space errors until now...

PR #452, opened 2017-12-23
test boost 1 66

BenjaminW3:
Currently fails on Windows because CMake-3.10 does not find the libs built by this boost version because the naming scheme has been changed.

- [x] wait for CMake 3.11.0 release which fixes this issue


BenjaminW3:
CMake 3.11 should fix this. It is not yet released or available on appveyor.

ax3l:
This is how I implement fallbacks for "too-old" CMake versions with newer boost versions. Be aware that you need to know all transitive, boost-internal dependencies and specify them explicitly:

https://github.com/ComputationalRadiationPhysics/picongpu/blob/263e4bb0e2a5979474e84014a283af328bf548de/include/picongpu/CMakeLists.txt#L152-L165

BenjaminW3:
>With such, Boost will compile with CMake versions that do not know the release yet and will only emit a very annoying warning.

I know that this is possible. The problem on Windows with the new boost version is, that the naming scheme of the libs has been changed to incorporate the architecture. So all non-current CMake versions fail to find boost libraries producing libs.

ax3l:
> I know that this is possible.

[...] and we should add it in Alpaka as we ;-)

> The problem on Windows with the new boost version [...]

How about simply not supporting unknown (too-new) versions of Boost on windows and throwing a clean CMake error? This problem will not change until boost has fully migrated to a sole-CMake based build suite I guess.

BenjaminW3:
ready to be merged!

ax3l:
@BenjaminW3 can you please instead of merging the `develop` branch into a PR just squash the changes and rebase? Make it slicker and cleaner in this case.

BenjaminW3:
The branch had a merge conflict and I tried out what this cool "resolve conflict" button in the GitHub UI did. It opened an online editor and created this merge ;-)

BenjaminW3:
I have rebased it again to remove the merge commit.

PR #455, opened 2018-01-17
Adopt to CUDA 9.1 which moved a header.

bertwesarg:


BenjaminW3:
Thanks for fixing this. I will try to find some time and add CUDA 9.1 to the compile tests.

PR #466, opened 2018-03-13
make alpaka a pure interface library

### LINKED ISSUE #460: OSX: ranlib: file libalpaka.a(Dymmy.cpp.o) has no symbols
Compiling PIConGPU with alpaka on OSX fails in the CMake step.

The main reason might be the "dummy" library that is created which we should replace with a more modern, CMake 3 `INTERFACE` target instead.

```
ranlib: file libalpaka.a(Dummy.cpp.o) has no symbols
```

Environment:
- macOS "High Sierra" (10.13.3)
- Apple Clang 9.0
- CMake 3.10.2
- Boost 1.65.1
- OpenMP 2.0 "blocks" backend

BenjaminW3:
Try to fix #460

* remove Dummy.cpp
* add IDE-only target for showing the sources

It would be good if someone could test this in a consuming project ;-)
* [x] merge after the 0.3.0 release

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -49,7 +49,6 @@ UNSET(_ALPAKA_COMMON_FILE)
 UNSET(_ALPAKA_ADD_EXECUTABLE_FILE)
 UNSET(_ALPAKA_ADD_LIBRRAY_FILE)
 UNSET(_ALPAKA_FILES_HEADER)
-UNSET(_ALPAKA_FILES_SOURCE)
 UNSET(_ALPAKA_FILES_OTHER)
 UNSET(_ALPAKA_VERSION_DEFINE)
 UNSET(_ALPAKA_VER_MAJOR)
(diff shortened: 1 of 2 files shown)


alpakaConfig.cmake
```diff
@@ -517,9 +515,12 @@ SET_SOURCE_FILES_PROPERTIES(${_ALPAKA_FILES_OTHER} PROPERTIES HEADER_FILE_ONLY T
 #-------------------------------------------------------------------------------
 # Target.
 IF(NOT TARGET "alpaka")
-    ADD_LIBRARY(
-        "alpaka"
-        ${_ALPAKA_FILES_HEADER} ${_ALPAKA_FILES_SOURCE} ${_ALPAKA_FILES_SCRIPT} ${_ALPAKA_FILES_CMAKE} ${_ALPAKA_FILES_DOC} ${_ALPAKA_FILES_OTHER})
+    ADD_LIBRARY("alpaka" INTERFACE)
+
+    # HACK: Workaround for the limitation that INTERFACE targets can not have sources.
```
ax3l: I would write "... can nit have public headers" since it's otherwise confusing

BenjaminW3: Hmmm... in CMake the command to add files to a target is called `target_sources`. Maybe I should directly use this wording.

ax3l:
But your problem is not that sources are missing, because their aren't any, but headers. CMake knows headers, see e.g. properties for installs of targets.

BenjaminW3:
For adding files (so that they can be seen in IDEs) to targets there is only [`target_sources`](https://cmake.org/cmake/help/v3.11/command/target_sources.html). This can be used for headers, sources and anything else. It is identical to mentioning the files directly within the `add_library` call.

ax3l:
Yes, and cmake still differentiates headers and sources. Just because a command to describe/add header files (of several) is called _sources (others are not) does not make our header files sources. Just change the comment please as suggested ;-)

BenjaminW3:
In this context, `target_sources()` is exactly what we would like to do, if it would work (even if we want to add C++ headers). So I will mention it in the comment. See the [comment in the CMake bug ticket](https://gitlab.kitware.com/cmake/cmake/issues/15234).
I do not currently see, where CMake knows about headers/sources? You can add everything to a target either by mentioning it in `add_library` or later via `target_sources`. As far as I know, CMake simply decides which compiler to call depending on the file ending . For headers, text documents, excel sheets, etc. this is simply none. However, it does not know anything about C++ headers vs. C++ sources.

ax3l:
What I was referring to above (sorry, from my phone and did not add links) was the syntax/options for
  install(targets ...
  https://cmake.org/cmake/help/v3.10/command/install.html

And properties for targets such as

  PRIVATE/PUBLIC_HEADER
  https://cmake.org/cmake/help/v3.10/prop_tgt/PUBLIC_HEADER.html#prop_tgt:PUBLIC_HEADER

on targets.

If you add an extra sentence about the target_sources command this is fine as well, I just want to avoid confusing readers on it by leaving out that information.
--
Sent from my smartphone.
Please excuse my brevity.

Issue #395, opened 2017-09-05
proposed renamings (after next release)


BenjaminW3:
CUDA uses the name `stream` to describe a sequential FIFO queue of work on the device. Other API's, e.g. OpenCL use the more comon term `CommandQueue` instead.
So my proposal is to rename all of our `stream`s to `queue`s for easier understanding. (`StreamCpuAsync`->`QueueCpuAsync`)
This would be a breaking change and would require changes to all current users. This would at least require a new version. This should best be combined with other major changes (e.g. dropping support of older CUDA versions, etc.). So I would delay this until such a major change is required.

What do you think?

ax3l:
Sure why not, that's a more general naming of it. 

Did you check how it's called in related domains, e.g. the Vulkan API has a similar concept for async executive pipelines.

BenjaminW3:
Vulkan also uses `Queue`.

ax3l:
+1

psychocoderHPC:
I am fine with such a change. But I would propose to  wait a little bit with this braking change.
The reason for this is that we are currently switched with PIConGPU to alpaka, this means alpaka is now used more frequently and we will find more parts we should maybe refactor.

We should collect up coming breaking changes within an issue.

ax3l:
Yes, please tag versions/releases before introducing breaking API changes.

BenjaminW3:
After seeing the ISAAC pull request I propose a second renaming: alpaka uses the name `Size` for types where `Idx` would be more appropriate. so the `size` namespace should be merged into the `idx` namespace and the only trait that the `size` namespace contains should be renamed as well.
Any objections?

PR #469, opened 2018-03-22
merge size namespace into idx namespace

BenjaminW3:
See #395

* size::Size -> idx::Idx
* size::SizeType -> idx::IdxType
* TSize -> TIdx


ax3l:
Thx!
Can you extend the PR description a little, please? E.g., why is `Idx` a good naming for an extend/size?

BenjaminW3:
Template types should be similar to units. So `Idx` is a better template name for all indices. Previously the accelerator was instantiated like `AccCpuSerial<TDim, TSize>`, now it is `AccCpuSerial<TDim, TIdx>`. The second template parameter was the return type (unit) of all methods returning indices/extents (where the extent is the maximum valid index) (getIdx<Block, Thread>, getWorkDiv<Grid, Block>...). I think this matches much better.
Alpaka users (cupla, ISAAC) already used `TIndex` or similar internally, so alpaka just adapts to it's users.

PR #471, opened 2018-03-28
add support for clang 6, remove 3.x and fix OpenMP 4.0

BenjaminW3:
* removed support for clang 3.5 up to 3.9
* added support for clang 6
* speed up the boost build by only building the required variant (debug/release)
* fix OpenMP 2.0 and 4.0 accelerator device descriptions to respect the `OMP_NUM_THREADS` in all cases
* fix OpenMP 4.0 execution
  If the OpenMP runtime thread count limit is lower than the block*thread count, the runtime may decide to use the requested number of threads for the blocks.
  Then the runtime is not able to use the correct number of threads for executing a block in parallel leading to wrong results.


BenjaminW3:
ready for review ;-)

ax3l:
Just to make issues/fixes more flexible: can you separate compiler-support and fixes such as the OpenMP fixes into independent PRs next time please? :-)

BenjaminW3:
I prefer it this way as well. However, the clang update revealed this bug because it's OpenMP runtime changed the behavior. So the new compiler was the only one that could prove that the fix works.

ax3l:
I see xD

Issue #462, opened 2018-03-09
appveyor builds crash with out of memory


BenjaminW3:
appveyor (Windows CI) often crashes with out of memory. This was the cases for a long time but it seems to have increased.
This does not block further development, as a successfull appveyor build is no requirement for merging.
This ticket is for documentation purposes because we can not add more then the available 3 GB RAM to the machines ;-) Builds have already been sequentialized so there is always ever only one single cpp file being build at a time.
Under this ticket the files leading to the crashes should be posted to get an overview. In the end we may disable the specific tests on windows or split them into multiple files.

BenjaminW3:
`C:\projects\alpaka\test\unit\mem\view\src\ViewSubViewTest.cpp(170): fatal error C1060: compiler is out of heap space`

PR #472, opened 2018-04-02
exclude test from appveyor CI build due to out-of-memory crashes

BenjaminW3:
Workaround for #462


test/unit/mem/view/CMakeLists.txt
```diff
@@ -59,6 +59,12 @@ SET(_SOURCE_DIR "src/")
 # Add all the source files in all recursive subdirectories and group them accordingly.
 append_recursive_files_add_to_src_group("${_SOURCE_DIR}" "${_SOURCE_DIR}" "cpp" _FILES_SOURCE)
 
+# FIXME: The appveyor CI builds crash with out of memory when building this file.
+IF(ALPAKA_CI)
```
ax3l: this will also exclude it in travis. Maybe just control the one file/test to be not executed on appveyor?

The general CMake way is to set `CTEST_CUSTOM_TESTS_IGNORE` in appveyor *externally* when calling cmake:
https://www.vtk.org/Wiki/CMake/Testing_With_CTest?ref=driverlayer.com/web

BenjaminW3: You are right, this also exludes the file from travis builds. I will fix this.
However, a simple test exclusion is not enough. This file fails to compile, so we have to exclude it from building at all.

BenjaminW3:
should be ready now!

ax3l:
@BenjaminW3 I restarted the one failed travis test and waited for it to finish, just in case you wonder why I did not merge it yet - it's still running ;)

BenjaminW3:
I know. Some of the tests are running too long and get killed. I will create a ticket for this. Especially the tests using ASan and UBSan are too slow. We may need to exclude some tests for those builds.

Issue #366, opened 2017-08-08
Organize Repo for Install


ax3l:
In order to easier install the library (with and without compiled examples or tests) it would be probably helpful if we re-organize some of the files in the repo close to how an installed install in a prefix would look like.

Assuming a user *could* decide to install into `/usr`, we should make sure that CMake files/modules are also [well-placed](https://cmake.org/cmake/help/v3.0/command/find_package.html) in either `lib/cmake` or `share/cmake` (Unix). See the link for more details. We can also have the files in the same structure inside the repo to reduce the complexity in the logic of install targets.

Likewise, files that are currently in [`_ALPAKA_FILES_OTHER`](https://github.com/ComputationalRadiationPhysics/alpaka/blob/655eb708ff0157b492d73d50727de151b60f6ee6/alpakaConfig.cmake#L483-L484) should either not be installed (`.gitignore`, CI yaml files, ...) or be moved to `share/doc/alpaka/` (README, COPYING, CHANGELOG, ...) ([example](https://packages.debian.org/stretch/amd64/grep/filelist)).

The logic in `alpakaConfig.cmake` should go in the root `CMakeLists.txt` and a very minimal `alpakaConfig.cmake` gets generated at install time. We [install this file](https://cmake.org/cmake/help/v3.0/command/find_package.html) to `lib/cmake/alpakaConfig.cmake` together with generated version files on unix and to `cmake/alpakaConfig.cmake` on windows.

In-project ("third-party source") usage of alpaka can use the CMake functionality `add_subdirectory` to talk to the root `CMakeLists.txt` which defines the same targets.

BenjaminW3:
I am not the right person to make those changes because I am not an expert regarding (different) unix install types and their requirements. Please create Pull Requests for the movement as well as for the CMake install rules after discussions about different portions has reached a consensus.

BenjaminW3:
The gitignore and CI files should not be installed. The ReadMe and the license (COPYING) should be installed but should not be moved within the repository because the current location is the standard.

ax3l:
Agreed to all of the above, but the README / COPYING need to be moved to a share folder, one can't place them in the install prefix (messy, un-associated & conflicts).

This does not mean they can't stay in the root of the repo, just not in the same structure of the installed package (but that's not too complicated to change in the logic).

j-stephan:
@ax3l Do you feel that this is still an issue? We have had installation support in our build system for quite a while now.

ax3l:
[...]

PR #478, opened 2018-04-09
[WIP] Installer

ax3l:
Prepares #366 

The goal is to "configure" - "build" (nothing to do) and "install" Alpaka just like any other C++ library via a standardized and well-defined CMake workflow.

### To Do

- [x] rebase against `develop`
- [ ] move logic from `alpakaConfig.cmake` to root `CMakeLists.txt`; generate `alpakaConfig.cmake`
- [ ] move and install helpers in `cmake/` as well
- [ ] remove `Findalpaka.cmake`: legacy script; only needed when a `*Config.cmake` package is not present
- [x] unset [`CMAKE_SIZEOF_VOID_P`](https://github.com/QuantStack/xtensor/blob/0.15.7/CMakeLists.txt#L167-L171) before creating `alpakaConfig.cmake`; [docs](https://cmake.org/cmake/help/v3.0/variable/CMAKE_SIZEOF_VOID_P.html)
- [ ] decide if `Alpaka` should be spelled uppercase in project name (and all CMake options/variables, config package, etc.)
- [ ] decide when to set `ALPAKA_ACC_...` vars: ideally allow *after* install as well

### References

- https://www.youtube.com/watch?v=bsXLMQ6WgIk
- https://cliutils.gitlab.io/modern-cmake/

Projects:
- https://github.com/openPMD/openPMD-api
- https://github.com/pngwriter/pngwriter
- https://github.com/QuantStack/xtensor

### DIFF
--- a/newVersion.sh
+++ b/newVersion.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+#
+# Copyright 2017-2018 Axel Huebl
+#
+# This file is part of alpaka.
+#
+# alpaka is free software: you can redistribute it and/or modify
+# it under the terms of the GNU Lesser General Public License as published by
+# the Free Software Foundation, either version 3 of the License, or
+# (at your option) any later version.
+#
+# alpaka is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+# GNU Lesser General Public License for more details.
+#
+# You should have received a copy of the GNU Lesser General Public License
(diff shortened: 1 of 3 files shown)


newVersion.sh
```diff
@@ -0,0 +1,124 @@
+#!/usr/bin/env bash
+#
+# Copyright 2017-2018 Axel Huebl
+#
+# This file is part of alpaka.
+#
+# alpaka is free software: you can redistribute it and/or modify
```
ax3l, line 7: With regards to #637, this new file's header will need an update :)

ax3l:
Superseded by @j-stephan 's #921 :rocket: :sparkles:

PR #479, opened 2018-04-10
Boost: Use CMake Targets

ax3l:
CMake targets are more reliable and the modern way to find dependencies. Dependencies as targets only need to be CMake-"linked", carrying all dependencies (includes, flags, requirements, etc.) as an object. Aaaannd... they are automatically `SYSTEM` libraries, which suppresses warnings from third-party headers with `-isystem` includes.

Especially CUDA linking is still cumbersome with the deprecated `FindCUDA.cmake` functionality, since `(target_)include_directories` "forget" that they are linked as `SYSTEM` - which throws tons of warnings e.g. from Boost in downstream projects.

Unfurtunately, the internal Boost dependencies are not well expressed yet. CMake decided to hard-code those in `FindBoost.cmake` for "known releases". Nevertheless, on the user-side it can happen to have a new Boost install with an older CMake install, which will not build properly if transitive dependencies are not properly requested. So we do that for our dependencies.

Added transitive dependency: Boost.Fiber depends on Boost.Thread which depends on Boost.Atomic. **Update:** required  for Boost.Thread `call_once` on POSIX platforms but generally seems to work without it.


alpakaConfig.cmake
```diff
@@ -569,6 +590,7 @@ IF(NOT TARGET "alpaka")
     IF(${_ALPAKA_INCLUDE_DIRECTORIES_PUBLIC_LENGTH} GREATER 0)
         TARGET_INCLUDE_DIRECTORIES(
             "alpaka"
+            SYSTEM
```
BenjaminW3: What will this do?

alpakaConfig.cmake
```diff
@@ -134,7 +134,7 @@ IF(${ALPAKA_DEBUG} GREATER 1)
     SET(Boost_DETAILED_FAILURE_MSG ON)
 ENDIF()
 IF(ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE)
-    FIND_PACKAGE(Boost ${_ALPAKA_BOOST_MIN_VER} QUIET COMPONENTS fiber context system thread chrono date_time)
+    FIND_PACKAGE(Boost ${_ALPAKA_BOOST_MIN_VER} QUIET COMPONENTS fiber context system thread atomic chrono date_time)
```
BenjaminW3: Are you really sure that atomic is necessary? See https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#814

alpakaConfig.cmake
```diff
@@ -213,8 +213,29 @@ IF(NOT Boost_FOUND)
...
 
 ELSE()
-    LIST(APPEND _ALPAKA_INCLUDE_DIRECTORIES_PUBLIC ${Boost_INCLUDE_DIRS})
-    LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC ${Boost_LIBRARIES})
+    IF(Boost_FIBER_FOUND)
+        # Boost fiber and default header-only libraries
+        IF(TARGET Boost::fiber)
+            LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC
+                 Boost::boost
+                 Boost::fiber Boost::context Boost::system Boost::thread
```
BenjaminW3: As far as I understad the FindBoost script, it already adds the fiber dependencies to the fiber target, so we do not need to mention them here again because they are transitively inherited.

alpakaConfig.cmake
ax3l: yes, transitive [via `Boost.Thread`](https://github.com/Kitware/CMake/blob/master/Modules/FindBoost.cmake#L823)

alpakaConfig.cmake
ax3l: Yes, transitive dependencies are pulled in automatically *as long as your CMake is new enough*.

I just checked the implementation of `FindBoost.cmake` again and it luckily changed recently:

- <= 3.9.0: Newer versions have no targets: https://github.com/Kitware/CMake/blob/v3.9.0/Modules/FindBoost.cmake#L752
- 3.10+: Finally! Assume newer releases *just have the same dependencies as before*: https://github.com/Kitware/CMake/blob/v3.10.0/Modules/FindBoost.cmake#L782-L783

The last thing is exactly what I implement here.

I would say, as long as we have our requirement not pushed to CMake 3.10+ let us keep it that explicit for transient dependencies *in case a user has a new Boost version that is not yet known by CMake*.

alpakaConfig.cmake
ax3l: Adds `-isystem` to anything inside `_ALPAKA_INCLUDE_DIRECTORIES_PUBLIC`: e.g. boost was in there.

If you want to add `alpaka` includes (`-I`) *and* boost includes (`-isystem`) do it in two commands to avoid warnings from third party headers.

ax3l: In other words: I will remove `_ALPAKA_INCLUDE_DIRECTORY` from `_ALPAKA_INCLUDE_DIRECTORIES_PUBLIC` and add it explicitly so we still get *our own warnings* :)

alpakaConfig.cmake
BenjaminW3: Then this is wrong in CMake see here: [1.53 doc](https://www.boost.org/doc/libs/1_53_0/doc/html/thread/build.html) or [1.66 doc](https://www.boost.org/doc/libs/1_66_0/doc/html/thread/build.html) and one of the last builds
```
Apr 06 09:09:36 Component configuration:
Apr 06 09:09:36 
Apr 06 09:09:36     - atomic                   : not building
Apr 06 09:09:36     - chrono                   : building
Apr 06 09:09:36     - container                : not building
Apr 06 09:09:36     - context                  : building
Apr 06 09:09:36     - coroutine                : not building
Apr 06 09:09:36     - coroutine2               : not building
Apr 06 09:09:36     - date_time                : building
Apr 06 09:09:36     - exception                : not building
Apr 06 09:09:36     - fiber                    : building
Apr 06 09:09:36     - filesystem               : not building
Apr 06 09:09:36     - graph                    : not building
Apr 06 09:09:36     - graph_parallel           : not building
Apr 06 09:09:36     - iostreams                : not building
Apr 06 09:09:36     - locale                   : not building
Apr 06 09:09:36     - log                      : not building
Apr 06 09:09:36     - math                     : not building
Apr 06 09:09:36     - metaparse                : not building
Apr 06 09:09:36     - mpi                      : not building
Apr 06 09:09:36     - program_options          : not building
Apr 06 09:09:36     - python                   : not building
Apr 06 09:09:36     - random                   : not building
Apr 06 09:09:36     - regex                    : not building
Apr 06 09:09:36     - serialization            : not building
Apr 06 09:09:36     - signals                  : not building
Apr 06 09:09:36     - system                   : building
Apr 06 09:09:36     - test                     : building
Apr 06 09:09:36     - thread                   : building
Apr 06 09:09:36     - timer                    : not building
Apr 06 09:09:36     - type_erasure             : not building
Apr 06 09:09:36     - wave                     : not building
```

ax3l: Interesting, but the linked 1.66 docs above state:

>  Boost.Thread uses by default Boost.Atomic in POSIX platforms to implement call_once..

BenjaminW3: No, this is a linux build. So at least the boost build system treats it as an optional dependency.

BenjaminW3: Currently, `Boost::atomic` is not built. The possibly incorrect transitive dependency of the `Boost::thread` target seems to require `Boost::atomic` on some builds but it can not be found and the build fails.
So we may need to build `Boost::atomic` when using the CMake targets.

ax3l: Yes, `Boost::thread` in CMake seems to just conservatively assume you might want to use all of it's functionality - which includes the `call_once` functionality we are not using. Anyway, we can just ad it. It does not add much to what it already pulls with `Boost.fiber`.

BenjaminW3:
I investigated the failing builds and found that they were due to the outdated cached docker images. They are not recreated when the environment defined in `.travis.yml` for a job has not changed. We are now requiring Boost::Atomic to be built, which was not available in the containers. Deleting the job cache was enough to get the builds green.

PR #486, opened 2018-05-13
Upgrade tested dependencies

BenjaminW3:
* gcc 8.0
* boost-1.67.0
* CMake 3.10.2->3.10.3 & 3.11.0->3.11.1


include/alpaka/block/shared/dyn/BlockSharedMemDynBoostAlignedAlloc.hpp
```diff
@@ -89,9 +93,16 @@ namespace alpaka
                             block::shared::dyn::BlockSharedMemDynBoostAlignedAlloc const & blockSharedMemDyn)
                         -> T *
                         {
+                            static_assert(
+                                core::vectorization::defaultAlignment >= alignof(T),
```
psychocoderHPC: Why do we have this requirement?

BenjaminW3: The block shared memory is allocated with this alignment (see above). Accessing wrongly aligned types is not possible. We could increase this alignment so that it is good for all types.

Issue #456, opened 2018-01-19
CUDA 9 requires Boost >=1.65


bertwesarg:
Boost <1.65 does not work with CUDA 9, because in CUDA 9 the macro `__CUDACC_VER__` is now a string constant, and these old boost versions still use this macro, instead of the new `__CUDACC_VER_[MAJOR,MINOR,BUILD]__` macros. Can we somehow detect this combination in CMake and fail gracefully, instead of:

```
In file included from /home/wesarg/opt/cuda-9.0.176/include/common_functions.h:50:0,
                 from /home/wesarg/opt/cuda-9.0.176/include/cuda_runtime.h:115,
                 from <command-line>:0:
/home/wesarg/opt/cuda-9.0.176/include/crt/common_functions.h:65:24: error: token ""__CUDACC_VER__"" is not valid in preprocessor expressions
 #define __CUDACC_VER__ "__CUDACC_VER__"
                        ^
/home/wesarg/opt/boost-1.63.0-gcc6/include/boost/config/compiler/nvcc.hpp:22:34: note: in expansion of macro ‘__CUDACC_VER__’
 #if !defined(__CUDACC_VER__) || (__CUDACC_VER__ < 70500)  
                                  ^~~~~~~~~~~~~~
/home/wesarg/opt/cuda-9.0.176/include/crt/common_functions.h:65:24: error: token ""__CUDACC_VER__"" is not valid in preprocessor expressions
 #define __CUDACC_VER__ "__CUDACC_VER__"
                        ^
/home/wesarg/opt/boost-1.63.0-gcc6/include/boost/config/compiler/nvcc.hpp:26:6: note: in expansion of macro ‘__CUDACC_VER__’
 #if (__CUDACC_VER__ > 80000) && (__CUDACC_VER__ < 80100)  
      ^~~~~~~~~~~~~~
```

BenjaminW3:
This could be added somewhere within the CUDA configuration in the [alpakaConfig.cmake](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/alpakaConfig.cmake#L338). Taking into account `Boost_MAJOR_VERSION, Boost_MINOR_VERSION, Boost_SUBMINOR_VERSION` should be possible there. Please note: we do not support 1.65.0 with CUDA 9. We only support 1.65.1.

ax3l:
1.65.1 and newer to be precise.

[...]

PR #487, opened 2018-05-13
move the CUDA 9 boost compatibility check into the CMake file

BenjaminW3:
#456

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -364,6 +364,12 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 ENDIF()
 
             ELSE()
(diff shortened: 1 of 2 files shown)


alpakaConfig.cmake
```diff
@@ -364,6 +364,12 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 ENDIF()
 
             ELSE()
+                # CUDA 9.0 removed the __CUDACC_VER__ macro. Boost versions lower than 1.65.1 still use this macro.
+                IF(CUDA_VERSION VERSION_GREATER 9.0 AND Boost_VERSION VERSION_LESS 1.65.1)
```
ax3l: Shouldn't this be >= 9.0? This will not catch CUDA 9.0

BenjaminW3: You are right, this was a mistake...

PR #495, opened 2018-05-20
Rename internal task creation traits for clarity

BenjaminW3:
This should not have any influence on external code (as there are no known external code bases defining new device or buffer types I know of).

### DIFF
--- a/include/alpaka/mem/buf/cpu/Copy.hpp
+++ b/include/alpaka/mem/buf/cpu/Copy.hpp
@@ -57,7 +57,7 @@ namespace alpaka
                         typename TViewDst,
                         typename TViewSrc,
                         typename TExtent>
-                    struct TaskCopyBase
+                    struct TaskCopyCpuBase
                     {
                         using ExtentSize = idx::Idx<TExtent>;
                         using DstSize = idx::Idx<TViewDst>;
@@ -86,7 +86,7 @@ namespace alpaka
                             "The source and the destination view are required to have the same element type!");
 
                         //-----------------------------------------------------------------------------
-                        TaskCopyBase(
+                        TaskCopyCpuBase(
                             TViewDst & viewDst,
                             TViewSrc const & viewSrc,
                             TExtent const & extent) :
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
@BenjaminW3  Why do you renamed this classes. The naming is unique because of the namespaces. With the new PR you have the platform type in the namespace and in the name of the object `XXX::TaskCopyXXXBase`.

BenjaminW3:
I was not sure about the naming before doing this either. However, it was beneficial to have project unique class names when looking for all places where a class is specialized and/or used.
Some python projects as well as the modern Windows API use this style as well, because it makes searching for classes in online documentation much easier.

psychocoderHPC:
@BenjaminW3 Could you please link example python projects. 

@ax3l, @tdd11235813  @theZiz @sbastrakov  What do you think?

I understand @BenjaminW3's  point that it is/could be better for searching the the documentation. Never the less IMO namespaces or class should be used to build groups instead of name prefixes but I am open for discussions.

sbastrakov:
I think both approaches are reasonable and either approach, ideally applied uniformly throughout the whole code, is fine.
Just to point out not the same, but quite similar situation. Some duplication of words in namespace and class names is currently present in other places as well e.g., `alpaka::dev::DevCpu`, `alpaka::queue::QueueCpuSync`.

psychocoderHPC:
@BenjaminW3  Why do you merged this PR without a peer review?

BenjaminW3:
I had the impression that reviews had been required before things can be merged. Therefore, I thought that  the completely green github status lines on the bottom of the page, which also include the reviews, meant that the PR can be merged.
I just had a look at the settings and was surprised that reviews were not required. I have enabled the required reviews now.

psychocoderHPC:
@BenjaminW3  Ahh OK ,I activated that one reviewer is not needed before it can be merged #498. This will remove the green button  as long as a review is missing.

theZiz:
> @ax3l, @tdd11235813 @theZiz @sbastrakov What do you think?

As Alpaka is already experienced as quite wordy I would prefer to keep the naming as short as possible and not duplicate namings. Imho it is the task of modern, C++ aware IDEs to help developers to differ between classes with the same name in different namespaces.

PR #499, opened 2018-05-25
add kernel::exec(...) as shortcut for queue::enqueue(q, exec::create(...))

BenjaminW3:
This further unifies the way in which tasks are handled.

There already have been sortcuts for creating and enqueueing memory copy and set tasks (`alpaka::mem::view::copy` / `alpaka::mem::view::set`). This introduces `alpaka::kernel::exec` that works in the same way. You still can store the execution task separately (as well as you can already store copy and set tasks separately).

Old:
```C++
auto const doSomethingKernelExecutorTask(	
    alpaka::exec::create<Acc>(	
        workDiv,	
        doSomethingKernel,	
        args...));
alpaka::queue::enqueue(queue, doSomethingKernelExecutorTask);
```
New:
```C++
auto const doSomethingKernelExecutorTask(	
    alpaka::kernel::createTaskExec<Acc>(	
        workDiv,	
        doSomethingKernel,	
        args...));
alpaka::queue::enqueue(queue, doSomethingKernelExecutorTask);
```
**New Shortcut**:
```C++
alpaka::kernel::exec<Acc>(
    queue,
    workDiv,	
    doSomethingKernel,	
    args...));
```


BenjaminW3:
@tdd11235813 You do not have to test those changes locally because they are already extensively checked with all accelerator types in multiple configurations in CI.

tdd11235813:
travis is not able to handle CUDA and the like, but I guess you refer to local CI instance at HZDR? 
I ask about the testing because today I have had errors with the CUDA path on alpaka develop branch, but maybe it is a mistake on my side, still figuring this out.

BenjaminW3:
At least compilation with CUDA is checked with Travis CI in multiple versions and configurations.
This will find most of the errors.

tdd11235813:
ok, good thing, that at least nvcc compile path is checked by travis. The errors like the host/device pointers mixing will lead to runtime crashes though. Nothing you really can do about it with alpaka at the moment, as native pointers are given to the kernel executor, right?

BenjaminW3:
Yes, such runtime errors can not be checked right now without a free open-source GPU cloud ;-)

example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -169,12 +167,11 @@ auto main()
...
     // as it is the case for the std::function example above.
-    auto const hiWorldBind (alpaka::exec::create<Acc> (
+    alpaka::kernel::exec<Acc> (
+        queue,
         workDiv,
-        std::bind( hiWorldFunction<Acc>, std::placeholders::_1, nExclamationMarks*3 )
-        ));
-
-    alpaka::queue::enqueue(queue, hiWorldBind);
+        std::bind( hiWorldFunction<Acc>, std::placeholders::_1, nExclamationMarks*2 )
```
tdd11235813: just a tiny detail, I guess `hiWorld...` could be replaced with `helloWorld` to be consistent with the example's title.

BenjaminW3: I would change this the next time I have to change something in this file.

BenjaminW3:
Any more reviews?

BenjaminW3:
As there do not seem to be more reviews incoming, I would like to merge this tomorrow.

include/alpaka/acc/AccCpuFibers.hpp
```diff
@@ -252,7 +252,7 @@ namespace alpaka
             };
         }
     }
-    namespace exec
```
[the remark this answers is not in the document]
BenjaminW3: Yes, this namespace is now removed.
Currently each non-patch release has breaking changes. All those changes will be documented in the corresponding release notes. I see the problem that users which are directly using the develop branch might need to check the commit history until there is an official release.

Issue #504, opened 2018-05-29
Memory bug? VectorAdd example fails


tdd11235813:
When using CUDA for vectorAdd example, memory glitches occur for non-small numElements.
Result is not computed for every 3rd row (as Elems = 3). For `numElements=123456` the first ~70k-100k rows are correct, but then the results are missing in every `Elems` row. The start index is random.

- when using printf inside the kernel or when using nvprof or cuda-memcheck, this does not happen
- CUDA volatile pointers or synchronizing the global writes by `__syncthreads` does not help
- the C buffer is initialized with 0s at the beginning

Possible output:
```
C[102624] == 0 != 2065228295
C[102627] == 0 != 1325975647
C[102630] == 0 != 2984720098
C[102633] == 0 != 3625720250
C[102636] == 0 != 3210258622
C[102639] == 0 != 4187900660
C[102642] == 0 != 3353799260
C[102645] == 0 != 1166409648
C[102648] == 0 != 1198068419
C[102651] == 0 != 2348462223
// ...
```
Have to dig further into this, it looks quite strange at the moment. [My current hip branch](https://github.com/tdd11235813/alpaka/tree/alpaka-hip).
Tested on Taurus/K80 with CUDA 8.0.61 and CUDA 9.1.85, gcc 5.3, boost 1.64.0. K80 GPUs had no ECC errors.

BenjaminW3:
Each third element does not sound like a memory error. If this is really the number of elements per thread, this might be a strong hint. Maybe there is some computation error in there or maybe some operation overflows when computing the first element for a given thread.

tdd11235813:
ok, it has nothing to do with element or index computation. It is probably caused by the different streams for memory and compute. When there is a `cudaDeviceSynchronize()` before `enqueue`ing the kernel and after the copies, it works. If such a sync is missing, then the kernel in the compute stream starts while memory is still copying.
What buffles me is, the cudaMemcpy is a device-wide sync and the host waits for its completion.

If you want to reproduce, take the current develop branch and change to CUDA backend in the vectorAdd example.
```C++
// in vectorAdd/src/main.cpp
// Define the accelerator                          
//using Acc = alpaka::acc::AccCpuSerial<Dim, Idx>; 
using Acc = alpaka::acc::AccGpuCudaRt<Dim, Idx>;   
using DevAcc = alpaka::dev::Dev<Acc>;              
[...]

PR #505, opened 2018-05-30
Alpaka HIP Backend

tdd11235813:
not intended for merging, just the current state of development.
Tested on mandelbrot and vectorAdd. cmake path is messy. Test integration at the beginning. Documentation not started yet.
(Further details next week.)


alpakaConfig.cmake
```diff
@@ -448,6 +463,147 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                    LIST(APPEND HIP_HIPCC_FLAGS "-Wno-deprecated-gpu-targets")
+                ENDIF()
+
+                IF(NOT CUDA_VERSION VERSION_LESS 7.5)
+                    LIST(APPEND HIP_HIPCC_FLAGS "--expt-extended-lambda")
+                ENDIF()
+
+                LIST(APPEND HIP_HIPCC_FLAGS "-arch=${ALPAKA_HIP_ARCH}")
+
+                IF(NOT MSVC)
```
BenjaminW3: Have you ever tried using it with MSVC? Else we may simply disable the backend here as well.

alpakaConfig.cmake
```diff
@@ -448,6 +463,147 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                    SET(ALPAKA_HIP_KEEP_FILES ON CACHE BOOL "activate keep files" FORCE)
+                ENDIF()
+                LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "general;${CUDA_CUDART_LIBRARY}")
+                LIST(APPEND _ALPAKA_INCLUDE_DIRECTORIES_PUBLIC ${HIP_INCLUDE_DIRS})
+
+                LIST(APPEND HIP_HIPCC_FLAGS
+                    -I${_ALPAKA_ROOT_DIR}/include/
+                    -I${_ALPAKA_ROOT_DIR}/test/common/include/
+                    #        "-Xcompiler -fopenmp"
+                    )
```
BenjaminW3: I am not sure what this should do. The comment should be removed and the rest can hopefully be removed or replaced.

alpakaConfig.cmake
```diff
@@ -448,6 +463,147 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+            ENDIF()
+        ENDIF()
+    ENDIF() # HIP
+
+#-------------------------------------------------------------------------------
+# Check options (just avoids CUDA+HIP conflict for now).
+#-------------------------------------------------------------------------------
+IF(ALPAKA_ACC_GPU_HIP_ENABLE AND ALPAKA_ACC_GPU_CUDA_ENABLE)
+    message(FATAL_ERROR "CUDA and HIP cannot be enabled both at the same time.")
+ENDIF()
```
BenjaminW3: Ah, here it is! Please move this to the other combination checks at the top.

alpakaConfig.cmake
```diff
@@ -463,13 +619,15 @@ ELSE()
...
 
+    if(NOT ALPAKA_ACC_GPU_HIP_ENABLE)
     # GNU
     IF(CMAKE_COMPILER_IS_GNUCXX)
         LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth-512")
     # Clang or AppleClang
     ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
         LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth=512")
     ENDIF()
+    ENDIF()
```
BenjaminW3: Please add a comment if this is not supported and correct the indentation.

alpakaConfig.cmake
```diff
@@ -512,18 +674,36 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
     MESSAGE(STATUS ALPAKA_ACC_GPU_CUDA_ENABLED)
 ENDIF()
 
+#------HIP enabled------
+IF(ALPAKA_ACC_GPU_HIP_ENABLE)
+    LIST(APPEND _ALPAKA_COMPILE_DEFINITIONS_PUBLIC "ALPAKA_ACC_GPU_HIP_ENABLED")
+    MESSAGE(STATUS ALPAKA_ACC_GPU_HIP_ENABLED)
+ENDIF()
+#-----------------------
```
BenjaminW3: Please remove those comment lines.

alpakaConfig.cmake
```diff
@@ -512,18 +674,36 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
 LIST(APPEND _ALPAKA_COMPILE_DEFINITIONS_PUBLIC "ALPAKA_DEBUG=${ALPAKA_DEBUG}")
 
 IF(ALPAKA_CI)
     LIST(APPEND _ALPAKA_COMPILE_DEFINITIONS_PUBLIC "ALPAKA_CI")
 ENDIF()
 
+#-----------------------
+# workaround to get the flags correctly to the hip compiler
+IF(ALPAKA_ACC_GPU_HIP_ENABLE)
+    SET(hip_defs ${_ALPAKA_COMPILE_DEFINITIONS_PUBLIC})
```
BenjaminW3: Please write all variables upper case and prefixed by `ALPAKA` and additionally by `_` for internal ones to be consistent. (e.g. `_ALPAKA_COMPILE_DEFINITIONS_HIP`)

alpakaConfig.cmake
```diff
@@ -512,18 +674,36 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+    list_add_prefix("-D" hip_defs)
+    LIST(APPEND HIP_HIPCC_FLAGS
```
BenjaminW3: The other thing with the includes from above should most probably also be done here.

alpakaConfig.cmake
```diff
@@ -512,18 +674,36 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+ENDIF()
+
 SET(_ALPAKA_INCLUDE_DIRECTORY "${_ALPAKA_ROOT_DIR}/include")
 SET(_ALPAKA_SUFFIXED_INCLUDE_DIR "${_ALPAKA_INCLUDE_DIRECTORY}/alpaka")
 
 SET(_ALPAKA_LINK_LIBRARY)
 LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "${_ALPAKA_LINK_LIBRARY}")
 
+#-------------------------------------------------------------------------------
+
```
BenjaminW3: Please remove those lines

cmake/addLibrary.cmake
```diff
@@ -130,6 +130,25 @@ FUNCTION(ALPAKA_ADD_LIBRARY libraryName)
                 ${optionArguments}
             )
         ENDIF()
+#-----Call HIP_ADD_LIBRARY-----------
```
BenjaminW3: Please remove the comment lines

example/vectorAdd/src/main.cpp
```diff
@@ -84,10 +84,18 @@ auto main()
...
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    using Acc = alpaka::acc::AccGpuCudaRt<Dim, Idx>;
+    using QueueAcc = alpaka::queue::QueueCudaRtSync;
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+    using Acc = alpaka::acc::AccGpuHipRt<Dim, Idx>;
+    using QueueAcc = alpaka::queue::QueueHipRtSync;
+#else
     using Acc = alpaka::acc::AccCpuSerial<Dim, Idx>;
+    using QueueAcc = alpaka::queue::QueueCpuSync;
+#endif
```
BenjaminW3: Please remove those things from the examples.

example/vectorAdd/src/main.cpp
```diff
@@ -124,13 +132,15 @@ auto main()
...
+    Data * const pBufHostB(alpaka::mem::view::getPtrNative(bufHostB));
     Data * const pBufHostC(alpaka::mem::view::getPtrNative(bufHostC));
+
     for (Idx i(0); i < numElements; ++i)
     {
-        pBufHostA[i] = static_cast<Data>(std::rand());
-        pBufHostB[i] = static_cast<Data>(std::rand());
+      pBufHostA[i] = static_cast<Data>(std::rand());
+      pBufHostB[i] = static_cast<Data>(std::rand());
+      pBufHostC[i] = 0;
```
BenjaminW3: Do we have to initialize and copy C? It should be overwritten by the kernel anyway.

include/alpaka/acc/AccGpuHipRt.hpp
```diff
@@ -0,0 +1,312 @@
...
+*
+* You should have received a copy of the GNU Lesser General Public License
+* along with alpaka.
+* If not, see <http://www.gnu.org/licenses/>.
+*/
+
+#pragma once
+
+#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+
```
BenjaminW3: Can we have some check here that the compiler really supports HIP? I had seen something some time ago on one of the branches.

include/alpaka/acc/AccGpuHipRt.hpp
```diff
@@ -0,0 +1,312 @@
...
+#include <alpaka/pltf/Traits.hpp>                   // pltf::traits::PltfType
+#include <alpaka/idx/Traits.hpp>
+
+// Implementation details.
+#include <alpaka/dev/DevHipRt.hpp>	// DevHipRt
+#include <alpaka/core/Hip.hpp>
+
+#include <boost/predef.h>                           // workarounds
+
+#include <typeinfo>                                 // typeid
```
BenjaminW3: You can remove those line-end comments for includes now because they have been remove everywhere.

include/alpaka/acc/AccGpuHipRt.hpp
```diff
@@ -0,0 +1,312 @@
...
+            typename... TArgs>
+        class ExecGpuHipRt;
+    }
+    namespace acc
+    {
+        //#############################################################################
+        //! The GPU HIP accelerator.
+        //!
+        //! This accelerator allows parallel kernel execution on devices supporting HIP or HCC
+        //#############################################################################
```
BenjaminW3: Please remove the second `//################` and `//--------------` lines everywhere because they have been removed in the rest of the code.

include/alpaka/alpaka.hpp
```diff
@@ -141,6 +155,8 @@
 #include <alpaka/mem/alloc/Traits.hpp>
 
 #include <alpaka/mem/buf/BufCudaRt.hpp>
+#include <alpaka/mem/buf/BufHipRt.hpp>
+
```
BenjaminW3: Could you please remove the empty line?

include/alpaka/alpaka.hpp
```diff
@@ -174,9 +190,11 @@
 #include <alpaka/pltf/PltfCpu.hpp>
 #include <alpaka/pltf/PltfCudaRt.hpp>
 #include <alpaka/pltf/Traits.hpp>
+#include <alpaka/pltf/PltfHipRt.hpp>
```
BenjaminW3: Could you please move this one line up?

psychocoderHPC:
> So we could see HIP as another CUDA compiler and use a macro within the source to switch from cuda to hip prefixes. The task of maintining all this duplicate source code is not neglectable.

@BenjaminW3 You are right but alpaka is designed to allow to use all backends in the some program context. This feature will be removed with a macro version. Never the less we should think about if we can reduce the amount of duplicated code.

BenjaminW3:
Even if it should be possible to use cuda and hip in parallel within the same compilation unit, we could still use macros in the following way:
AccGpuCudaHip.hpp
```C++
#pragma once

#define ALPAKA_CUDA_HIP_PREFIX cuda
#include "AccGpuCudaHipInternal.hpp"
#endif

#define ALPAKA_CUDA_HIP_PREFIX hip
#include "AccGpuCudaHipInternal.hpp"
#endif
```
The current `AccGpuCuda.hpp` would be renamed to `AccGpuCudaHipInternal.hpp` and the `cuda` prefix would be replace by the `ALPAKA_CUDA_HIP_PREFIX` macro. Furthermore, the `#pragma once` would have to be removed and those headers should never be included directly.

BenjaminW3:
I would really be most interested in a CI integration. Is this possible?
Having this, I could incorporate most of my comments myself.

tdd11235813:
thanks a lot for your review, I am going work on it next week.
Regarding CI I think it is possible for the CUDA path, as HIP is just calling nvcc. It is not implemented into the travis config yet.
For abstracting CUDA and HIP codes, these APIs may not be completely the same and there are functions missing in HIP, although I could not find any examples for it except the [Texture reference API](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#textures-and-cache-control).
So I guess the code could be kept tight for the most time, only few branches needed where HIP and CUDA take different paths.
At the moment CUDA and HIP cannot be compiled at the same time, when HIP is directed to use nvcc.
This configuration would be quite senseless, so I prefer to disallow this or to only enable the CUDA/HIP header for the first compilation phase, otherwise multiple definitions occur.

BenjaminW3:
I have now done some reading about what this HIP thing is all about and am unsure if it is really what we want to have within alpaka.
HIP is in the end only a translator for:
* HIP to CUDA
* HIP to HC

I do not really see an advantage of using this translator within alpaka.
* We already support CUDA natively, so we do not need the first feature.
* The HIP to HC feature would bring us support for the hcc compiler. But is it worth going through this translation process and bringing in these optional dependencies when we could directly target the HC backend ourselves?

theZiz:
There are two reasons, why we chose HIP. First of all, it is easy, as it is basically just a replacement `cuda` → `hip`.
Secondly, it can be tested without a working AMD gpu. The other way around, what would be the benefit of directly supporting `hcc`?

psychocoderHPC:
> The HIP to HC feature would bring us support for the hcc compiler. But is it worth going through this translation process and bringing in these optional dependencies when we could directly target the HC backend ourselves?

Yes, I absolutely agree with that but with this small step by including the HIP backend we can very fast and "easy" support AMD. Never the less we should remove HIP in later versions by an native HC backend.

tdd11235813:
in this case, do we still want to unify the HIP/CUDA world to another abstract layer? Or keep HIP as an own backend where it is right now? Or stop current development, if we could start on implementing the hcc backend?
IMHO, we should work on HIP as a separate backend to make alpaka available to AMD hardware, as HCC backend will not be there soon. 
HCC will be a different story and if we are lucky, we have a HCC backend in one year and can compare it to the HIP backend as well. Then we can drop HIP or HCC, whatever suits best.

psychocoderHPC:
HIP is something like alpaka and has two backends. Therefore it is not wrong to support HIP.
@ax3l provided a few month in the past the ling that there is an discussion to have a HIP front-end for  llvm (https://lists.llvm.org/pipermail/cfe-dev/2018-March/057426.html). In that case HIP makes more sense than HC.

tdd11235813:
ok, we talked about this and would proceed with the current way of integrating HIP into alpaka in this way.
Perhaps we can start with analysing SYCL and HCC more in depth soon.
So HIP would be the backend for the ROCm enabled GPUs at the moment.
For older AMD hardware (IIRC pre Fury, pre RX 470) SYCL could be the choice, which bases on OpenCL1.2. But for this discussion a separate issue will be openend then.

tdd11235813:
current dev goes here until another stage for reviewing is reached:
[compare this PR and new dev (dev-alpaka-hip)](https://github.com/tdd11235813/alpaka/compare/alpaka-hip...tdd11235813:dev-alpaka-hip)
dev-alpaka-hip:
- It turned out that cmake silently adds some compiler definitions besides of `CUDA_PROPAGATE_HOST_FLAGS`, so HIP part in `alpakaConfig.cmake` includes additional commands you will not find for the CUDA part
- error reported in issue #504 does not occur on HIP:nvcc path

tdd11235813:
compiles successfully with latest HIP 1.5.8292, CUDA8+CUDA9.1, gcc4.9.4+gcc5.5.
tests run successfully with latest HIP 1.5.8292, CUDA9.1, gcc4.9.4+gcc5.5.
[travis CI job report for HIP](https://travis-ci.org/tdd11235813/alpaka/jobs/410820204).

ToDo:
- update doc (where shall I place the readme about HIP installation and workarounds?)
- some cleanup
- more compiler combination tests

I hope, it is ok when I overwrite this PR with the new branch, when things will be finished for review.

BenjaminW3:
It is perfectly fine to update this PR.

tdd11235813:
current state
- [x] FindHIP.cmake updated and fixed
- [x] updated travis scripts ([tests before last rebase](https://travis-ci.org/tdd11235813/alpaka/builds/418277611))
- [x] few more compiler tests (HIP+clang did not work out of the box, so skipping)
- [x] HIP memcopies "workarounds" (just rely on UVA instead of using special commands for eg device to device copies)
  - one exception: there is no [cudaMemcpy3DAsync in HIP](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/CUDA_Runtime_API_functions_supported_by_HIP.md), so just have a synchronous cudaMemcpy3D at the moment
- [x] separate HIP doc for installing
- [x] HIP atomics API completed (equals with CUDA back-end)
- [x] fix Common.hpp after rebases ~~(memView and rand tests fail for now)~~
- [x] add [rocRAND](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#table-of-architecture-properties) for HIP random numbers
- [x] test and fix full debug mode
- [x] incorporate review ;)
- [x] check and update platform properties
- [x] add rocRAND to travis ~~(fails atm because `HIP_ROOT_DIR` is empty, although FindHIP.cmake run through, dont get it, whats wrong here, need to dig around)~~
- [x] fix missing flags for hipcc when in release build
  - there seems to be a [bug in FindHIP.cmake](https://github.com/ROCm-Developer-Tools/HIP/issues/631), however, `alpaka_add_executable` had to be a **macro** to not interrupt the scope
- [x] HIP P2P copy is there ~~but update depends on solution for P2P enabler functions (#615, #616) (at the moment memP2P test fails when P2P is not possible)~~
-  ~~place guards for atomics (see [HIP porting guide](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#table-of-architecture-properties))~~
    - macros for atomics support [only relevant](https://github.com/ROCm-Developer-Tools/HIP/blob/master/include/hip/nvcc_detail/hip_runtime.h) for Nvidia GPUs with CC <= 2.0, so they are useless and we keep  `BOOST_ARCH_PTX` on the Nvidia platform
- [x] checkout ROCm/AMD platform (building rocm stack takes way too long, but for travis/Ubuntu 16.04 there exists a [binary package](https://hub.docker.com/r/rocm/dev-ubuntu-16.04/))
  - got the the tools (boost, clang, hcc, hip, rocrand ..) together, now alpaka code needs to be modified for HCC
  - fix math/ header includes to be compatible with HCC
  - `~Foo() restrict(amp,cpu);` required for classes which are targeted to host and device
    - this differentiation is also needed for other host-device backends, otherwise destructor is undefined

tdd11235813:
PR is updated, but not final as you can see from the list above. However, you can have a look at the current code and if you see something, just give a sign.

example/vectorAdd/src/main.cpp
BenjaminW3: This is still valid. The examples are still hardcoded to the serial CPU accelerator.

include/alpaka/alpaka.hpp
BenjaminW3: still valid ;-)

include/alpaka/block/shared/st/Traits.hpp
```diff
@@ -152,6 +152,7 @@ namespace alpaka
                         >::type>
                     {
                         //-----------------------------------------------------------------------------
+                        ALPAKA_NO_HOST_ACC_WARNING
```
BenjaminW3: Is this really necessary here?

include/alpaka/core/BoostPredef.hpp
```diff
@@ -30,12 +30,50 @@
...
+#endif
+
 //-----------------------------------------------------------------------------
 // In boost since 1.68.0
 // CUDA language detection
 // - clang defines __CUDA__ and __CUDACC__ when compiling CUDA code ('-x cuda')
 // - nvcc defines __CUDACC__ when compiling CUDA code
-#if !defined(BOOST_LANG_CUDA)
+#if !defined(BOOST_LANG_CUDA) && \
+    !(BOOST_LANG_HIP && defined(__HIPCC__))
```
BenjaminW3: We would have to add this to boost as well or is there another way to fix this behaviour on our side? Maybe undef `BOOST_LANG_CUDA` afterwards when `BOOST_LANG_HIP` is detected? I wanted to remove this whole block as soon as we increase our required boost version.

include/alpaka/core/Common.hpp
```diff
@@ -31,13 +31,21 @@
...
+#endif
+
 //-----------------------------------------------------------------------------
 // clang CUDA compiler detection
 // Currently __CUDA__ is only defined by clang when compiling CUDA code.
-#if defined(__clang__) && defined(__CUDA__)
-    #define BOOST_COMP_CLANG_CUDA BOOST_COMP_CLANG
-#else
-    #define BOOST_COMP_CLANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
+#if !defined(BOOST_COMP_CLANG_CUDA)
```
BenjaminW3: How can this ever happen? This macro has not been ported to boost so there is no one who could define this.

BenjaminW3:
Thank you for this contribution! It looks very solid now. We have CI integration, we have CMake integration, we have all the code. However, I am still unsure about the massive amount of code duplication. I would still prefer the macro soultion I proposed above and only duplicate code where it is really necessary due to differences.

include/alpaka/core/Common.hpp
tdd11235813: well spotted. I guess this removal occurred with the changes due to `include/alpaka/core/BoostPredef.hpp`. Probably assumed it went there like the other definitions which can be found there.

tdd11235813:
at the moment I try to get it working with HIP(hcc), where other issues popped up like the mixed constructor/destructor host/device attributes. I prepared a PR last week, but it is not finished yet.
Regarding the code duplication, I look into it, where things will be different between CUDA and HIP code.
One difference is that HIP relies on Unified Memory, which makes some things easier. Not sure if we can assume the same for CUDA, maybe we would loose some windows support then.

include/alpaka/core/Common.hpp
tdd11235813: wait, the definition is there:
https://github.com/ComputationalRadiationPhysics/alpaka/blob/bebc38dc82b86c03a1edeba147a460c0e11f17c6/include/alpaka/core/Common.hpp#L41-L49

BenjaminW3: I do not understand what you mean. It is the place I added the comment on. It should be possible to simply revert your change here.

tdd11235813: ah I read it like the definition would be gone (diff shows it as removed), but you actually only mean the
 #if !defined(BOOST_COMP_CLANG_CUDA)`, that this is not required, right?

BenjaminW3: yes

include/alpaka/core/Common.hpp
```diff
@@ -31,13 +31,21 @@
     #include <intrin.h>
 #endif
 
+//-----------------------------------------------------------------------------
+// BOOST_PREDEF_MAKE_10_VVRRP(V)
+#if !defined(BOOST_PREDEF_MAKE_10_VVRRP)
+    #define BOOST_PREDEF_MAKE_10_VVRRP(V) BOOST_VERSION_NUMBER(((V)/1000)%100,((V)/10)%100,(V)%10)
+#endif
```
BenjaminW3: This should have been moved to BoostPredef.hpp and can be removed here.

include/alpaka/core/Common.hpp
tdd11235813: k, I'll remove that :)

tdd11235813:
sorry for slow responding. I have incorporated the review locally. 
However, things seem to become ugly due to the host/device signature restrictions by HCC. 
It produces compile errors (helloWorld, HIP and CPU-serial back-end enabled, compiles via HIP(HCC)):
```C++
/alpaka/meta/NdLoop.hpp:95:25: error: call from AMP-restricted function to CPU-restricted function                                                          
                        f(idx); 
// --> ndLoop is host+device, f is device (error also occurs when f is host)
// ... 
/alpaka/exec/ExecCpuSerial.hpp:146:23: note: in instantiation of function template specialization 'alpaka::meta::ndLoopIncIdx<alpaka::vec::Vec<std::integra\
l_constant<unsigned long, 3>, unsigned long>, (lambda at /home/test/vboxshare/include/alpaka/exec/ExecCpuSerial.hpp:148:21)>' requested here                                            
                meta::ndLoopIncIdx( // <-- within ExecCpuSerial, which is host
```
However, my current minimal examples which happily mix host device, just run fine (except linker errors for isolated device-only codes).

HCC shows at least the calling functions, however, system libraries are also included like for function bindings or assert. I'll try to find a shorter path through alpaka, to play around with this issue. 
Disabling all the host device attributes in alpaka does not work out of the box, there are many places with device specific things like intrinsics or indexing.

tdd11235813:
With the help of @psychocoderHPC more issues could be fixed today (not yet merged into this PR)
- unsupported dim3 in device -> now just use vec::Vec
- fixed some mismatching host-device signatures
- workarounds for syncthreads_[count|and|or]).

Now working on things like:
```
error:  '__assert_fail':  no overloaded function has restriction specifiers that are compatible with the ambient context 'assertValueUnsigned'
                     assert(arg >= 0);
```
and several 
```
call from AMP-restricted function to CPU-restricted function
```
errors, e.g. mem::alloc::* functions are host, but the traits function for alloc is host-device. The NdLoop and the functor signature issue mentioned above is still unsolved, but I'll check into it again tomorrow.

tdd11235813:
HIP(HCC) current WIP
- [x] workaround for assert 
  - to make it host-device
  - `hip_runtime.h` already provides this, but it will be overwritten later on by some external header, I guess
  - nevertheless I removed all unneeded `cassert` includes
- [x] missing device-dim3 on HCC -> vec::Vec
- [x] use vec-in-class
- [x] workarounds for unsupported syncthreads_[count|and|or]
- [x] minimal rocrand finder (also takes care of hiprand)
- [x] NdLoop compile-time error for host functor is "fixed", but see below
- [ ] mandelbrot, matMul, .. have empty results 
  - maybe memset/copy did not work, looks like uninitialized
  - maybe shared memory is also an issue
- [ ] adapt host to host-device signatures where needed
  - cannot set all host signatures to host-device ones, as there are host functions which would not compile under device
  - examples: `getBlockSharedMemDynSizeBytes`, host-side lambdas (for memset) which are called within host-device functions, create(), ..
- [ ] workaround for missing HIP pendant for `cuStreamWaitValue32` (for test 'event')
- [ ] test 'memView' causes memory access error and hangs
```
Memory access fault by GPU node-1 (Agent handle: 0x2769c30) on address 0x9db000. Reason: Page not present or supervisor privilege.
unknown location(0): fatal error: in "viewStaticAccMem/staticDeviceMemoryConstant<alpaka__acc__AccHipRt<std__integral_constant<unsigned_long,_2ul>,_unsigned_int>>": signal: SIGABRT (application abort requested)
/home/mwerner/cuda-workspace/alpaka/test/unit/mem/view/src//ViewStaticAccMem.cpp(123): last checkpoint
```
Current tests on AMD Vega64 system:
```
 1/25: bufferCopy .......   Passed    0.15 sec
 2/25: helloWorld .......   Passed    0.12 sec
 3/25: helloWorldLambda .   Passed    0.12 sec
 4/25: vectorAdd ........   Passed    0.12 sec
 5/25: headerCheck ......   Passed    0.31 sec
 6/25: axpy .............***Failed    0.26 sec # <-- results have precision errors of order 1e-2 !
 7/25: mandelbrot .......   Passed    0.18 sec # <-- runs fine, but image is empty!
 8/25: matMul ...........***Failed    0.17 sec # <-- results empty
 9/25: sharedMem ........***Failed    0.16 sec # <-- results empty
10/25: acc ..............   Passed    0.12 sec
11/25: atomic ...........***Failed    0.23 sec # <-- isValidAccDevProps assertions failed
12/25: blockShared ......***Failed    0.24 sec # <-- isValidAccDevProps assertions failed
13/25: blockSync ........***Failed    0.24 sec # <-- checks failed including isValidAccDevProps
14/25: core .............   Passed    0.12 sec
15/25: event ............***Not Run   0.00 sec # <-- cuStreamWaitValue32 not available in HIP
16/25: idx ..............   Passed    0.12 sec
17/25: kernel ...........***Failed    0.29 sec # <-- critical check failed, device code partially not available
18/25: memBuf ...........***Failed    0.70 sec # <-- critical check failed
19/25: memView ..........***Not Run   0.00 sec # <-- would cause memory access error with hangup
20/25: memP2P ...........   Passed    0.13 sec
21/25: meta .............   Passed    0.13 sec
22/25: queue ............***Failed    0.79 sec # <-- critical check failed
23/25: rand .............***Failed    0.25 sec # <-- isValidAccDevProps assertions failed
24/25: time .............***Failed    0.24 sec # <-- isValidAccDevProps assertions failed
25/25: vec ..............   Passed    0.12 sec
Total Test time (real) =   5.51 sec
```

ax3l:
@tdd11235813 please be aware of #644 and linked PR(s)

tdd11235813:
yup, thanks for ref, gonna check it out right on Monday.

include/alpaka/acc/AccHipRt.hpp
```diff
@@ -0,0 +1,313 @@
...
+                        // m_gridBlockCountMax
+                        std::numeric_limits<TIdx>::max(),
+                        // m_blockThreadExtentMax
+                        extent::getExtentVecEnd<TDim>(
+                            vec::Vec<dim::DimInt<3u>, TIdx>(
+                                static_cast<TIdx>(hipDevProp.maxThreadsDim[2]),
+                                static_cast<TIdx>(hipDevProp.maxThreadsDim[1]),
+                                static_cast<TIdx>(hipDevProp.maxThreadsDim[0]))),
+                        // m_blockThreadCountMax
+                        static_cast<TIdx>(hipDevProp.maxThreadsPerBlock),
```
BenjaminW3: You should apply the fixes from #646 here similar to the AccGpuCuda.

tdd11235813:
The *isValidAccDevProps* errors could be resolved with the latest update (adopted for HIP as advised by @BenjaminW3 ). The remaining failed tests are:
```
The following tests FAILED:
          7 - axpy (Failed)
          9 - matMul (Failed)
         10 - sharedMem (Failed)
         14 - blockSync (Failed)
         16 - event (Not Run)
         18 - kernel (Failed)
         19 - memBuf (Failed)
         20 - memView (Not Run)
         23 - queue (SEGFAULT)
```

include/alpaka/block/shared/st/Traits.hpp
tdd11235813: yes, and also in mem/view/Traits.hpp. Separated PRs are in the pipe, need to test other stuff before though.

tdd11235813:
After deeper investigations and few more workarounds I could fix the test issues, e.g. mandelbrot is rendered correctly with HIP(HCC) and HIP(NVCC). However some hard restrictions are there, mostly for memView test, see list below.
Code will be separated into two PRs, because some interfacing non-HIP classes are affected as well.
Afterwards I am going to update this PR. This is planned for the next week.

## Restrictions

- cosmetic reordering of few HIP includes.
- HCC: Adds missing HIPRT_CB definition.
- HCC: Constructors' attributes unified with destructors'.
  - host/device signature must match in HIP(HCC)
- HCC: Workaround for unsupported syncthreads_{count|and|or}.
  - uses temporary shared value and atomics
- HCC: Workaround for buggy hipStreamQuery, hipStreamSynchronize.
  - introduces own queue management
  - hipStreamQuery, hipStreamSynchronize did not work in multithread env.
- HCC: Workaround for missing cuStreamWaitValue32.
  - polls value each 10ms
- HCC: exclude hipMalloc3D, hipMallocPitch when size is zero
  - otherwise they throw an Unknown Error
- HCC: static device memory and constant memory are still unusable
- HCC: note, that printf in kernels is still not supported in HIP
- 3D is currently unsupported
  - missing hipMemcpy3DAsync is replaced with hipMemcpy3D though
  - TestAccs excludes 3D specialization of Hip back-end for now
- Finds HIP(HCC) rocrand library paths in cmake.
- HCC compatible header includes in alpaka/math/.
- sets proper OpenMP flags for HIP(nvcc).
- uses alpaka vector for device side dim3 types .
  - dim3 not available on device in HIP
- HIP+Clang is not working.
- HIP >= 1.5.1 required.


## Tests AMD Vega64 (all other back-ends disabled)

<details>

```bash
$ ./test/integ/mandelbrot/mandelbrot
Running 1 test case...
MandelbrotKernel( numRows:1024, numCols:1024, maxIterations:300, accelerator: AccGpuHipRt<2,j>, kernel: 16MandelbrotKernel, workDiv: {gridBlockExtent: (1024, 1), blockThreadExtent: (1, 1024), threadElemExtent: (1, 1)})
Execution time: 36 ms

*** No errors detected

# ---

$ make test
Running tests...
 Test project /home/mwerner/cuda-workspace/alpaka/build
       Start  1: bufferCopy
  1/26 Test  #1: bufferCopy .......................   Passed    0.14 sec
       Start  2: helloWorld
  2/26 Test  #2: helloWorld .......................   Passed    0.12 sec
       Start  3: helloWorldLambda
  3/26 Test  #3: helloWorldLambda .................   Passed    0.12 sec
       Start  4: reduce
  4/26 Test  #4: reduce ...........................   Passed    0.12 sec
       Start  5: vectorAdd
  5/26 Test  #5: vectorAdd ........................   Passed    0.12 sec
       Start  6: headerCheck
  6/26 Test  #6: headerCheck ......................   Passed    0.31 sec
       Start  7: axpy
  7/26 Test  #7: axpy .............................   Passed    0.16 sec
       Start  8: mandelbrot
  8/26 Test  #8: mandelbrot .......................   Passed    0.18 sec
       Start  9: matMul
  9/26 Test  #9: matMul ...........................   Passed    0.17 sec
       Start 10: sharedMem
 10/26 Test #10: sharedMem ........................   Passed    0.16 sec
       Start 11: acc
 11/26 Test #11: acc ..............................   Passed    0.12 sec
       Start 12: atomic
 12/26 Test #12: atomic ...........................   Passed    0.23 sec
       Start 13: blockShared
 13/26 Test #13: blockShared ......................   Passed    0.24 sec
       Start 14: blockSync
 14/26 Test #14: blockSync ........................   Passed    0.23 sec
       Start 15: core
 15/26 Test #15: core .............................   Passed    0.12 sec
       Start 16: event
 16/26 Test #16: event ............................   Passed    0.19 sec
       Start 17: idx
 17/26 Test #17: idx ..............................   Passed    0.12 sec
       Start 18: kernel
 18/26 Test #18: kernel ...........................   Passed    0.26 sec
       Start 19: memBuf
 19/26 Test #19: memBuf ...........................   Passed    0.25 sec
       Start 20: memView
 20/26 Test #20: memView ..........................   Passed    0.25 sec
       Start 21: memP2P
 21/26 Test #21: memP2P ...........................   Passed    0.11 sec
       Start 22: meta
 22/26 Test #22: meta .............................   Passed    0.14 sec
       Start 23: queue
 23/26 Test #23: queue ............................   Passed    1.19 sec
       Start 24: rand
 24/26 Test #24: rand .............................   Passed    0.27 sec
       Start 25: time
 25/26 Test #25: time .............................   Passed    0.23 sec
       Start 26: vec
 26/26 Test #26: vec ..............................   Passed    0.12 sec
 
 100% tests passed, 0 tests failed out of 26
 
 Total Test time (real) =   5.70 sec
```

</details>

## Tests Intel Nvidia GV100

<details>

```bash
$ ./test/integ/mandelbrot/mandelbrot
Running 5 test cases...
MandelbrotKernel( numRows:1024, numCols:1024, maxIterations:300, accelerator: AccCpuSerial<2,j>, kernel: 16MandelbrotKernel, workDiv: {gridBlockExtent: (1024, 1024), blockThreadExtent: (1, 1), threadElemExtent: (1, 1)})
Execution time: 2683 ms
MandelbrotKernel( numRows:1024, numCols:1024, maxIterations:300, accelerator: AccCpuThreads<2,j>, kernel: 16MandelbrotKernel, workDiv: {gridBlockExtent: (1024, 5), blockThreadExtent: (1, 224), threadElemExtent: (1, 1)})
Execution time: 63826 ms
MandelbrotKernel( numRows:1024, numCols:1024, maxIterations:300, accelerator: AccCpuOmp2Blocks<2,j>, kernel: 16MandelbrotKernel, workDiv: {gridBlockExtent: (1024, 1024), blockThreadExtent: (1, 1), threadElemExtent: (1, 1)})
Execution time: 204 ms
MandelbrotKernel( numRows:1024, numCols:1024, maxIterations:300, accelerator: AccCpuOmp2Threads<2,j>, kernel: 16MandelbrotKernel, workDiv: {gridBlockExtent: (1024, 37), blockThreadExtent: (1, 28), threadElemExtent: (1, 1)})
Execution time: 399 ms
MandelbrotKernel( numRows:1024, numCols:1024, maxIterations:300, accelerator: AccGpuHipRt<2,j>, kernel: 16MandelbrotKernel, workDiv: {gridBlockExtent: (1024, 1), blockThreadExtent: (1, 1024), threadElemExtent: (1, 1)})
Execution time: 0 ms

*** No errors detected

# ---

$ make test
Running tests...
 Test project /home/mwerner/cuda-workspace/alpaka/build
       Start  1: bufferCopy
  1/26 Test  #1: bufferCopy .......................   Passed    0.01 sec
       Start  2: helloWorld
  2/26 Test  #2: helloWorld .......................   Passed    0.01 sec
       Start  3: helloWorldLambda
  3/26 Test  #3: helloWorldLambda .................   Passed    0.00 sec
       Start  4: reduce
  4/26 Test  #4: reduce ...........................   Passed    4.09 sec
       Start  5: vectorAdd
  5/26 Test  #5: vectorAdd ........................   Passed    0.04 sec
       Start  6: headerCheck
  6/26 Test  #6: headerCheck ......................   Passed    0.00 sec
       Start  7: axpy
  7/26 Test  #7: axpy .............................   Passed    1.13 sec
       Start  8: mandelbrot
  8/26 Test  #8: mandelbrot .......................   Passed   69.78 sec
       Start  9: matMul
  9/26 Test  #9: matMul ...........................   Passed    1.47 sec
       Start 10: sharedMem
 10/26 Test #10: sharedMem ........................   Passed    4.61 sec
       Start 11: acc
 11/26 Test #11: acc ..............................   Passed    0.02 sec
       Start 12: atomic
 12/26 Test #12: atomic ...........................   Passed    3.80 sec
       Start 13: blockShared
 13/26 Test #13: blockShared ......................   Passed    2.73 sec
       Start 14: blockSync
 14/26 Test #14: blockSync ........................   Passed    2.88 sec
       Start 15: core
 15/26 Test #15: core .............................   Passed    0.01 sec
       Start 16: event
 16/26 Test #16: event ............................   Passed    1.14 sec
       Start 17: idx
 17/26 Test #17: idx ..............................   Passed    0.01 sec
       Start 18: kernel
 18/26 Test #18: kernel ...........................   Passed    4.52 sec
       Start 19: memBuf
 19/26 Test #19: memBuf ...........................   Passed    3.73 sec
       Start 20: memView
 20/26 Test #20: memView ..........................   Passed    3.98 sec
       Start 21: memP2P
 21/26 Test #21: memP2P ...........................   Passed    0.93 sec
       Start 22: meta
 22/26 Test #22: meta .............................   Passed    0.01 sec
       Start 23: queue
 23/26 Test #23: queue ............................   Passed    2.13 sec
       Start 24: rand
 24/26 Test #24: rand .............................   Passed    3.47 sec
       Start 25: time
 25/26 Test #25: time .............................   Passed    1.54 sec
       Start 26: vec
 26/26 Test #26: vec ..............................   Passed    0.01 sec
 
 100% tests passed, 0 tests failed out of 26
```

</details>

tdd11235813:
it is now running on AMD hardware, see comment before. However, with some restrictions.

tdd11235813:
but it is still true, that we should promote this as experimental back-end, also because of it is very likely that there are more funny glitches than the already uncovered.

tdd11235813:
I am going to remove conflicting `ALPAKA_FN_HOST` signatures instead of adding them to all methods of the corresponding class. Examples are AtomicStdLibLock, QueueCpuAsyncImpl, QueueCpuAsync, AtomicNoOp, ...

tdd11235813:
In `mem/alloc/Traits.hpp` there will be a compiler switch looking like this
```C++
#if defined( BOOST_COMP_HCC ) && BOOST_COMP_HCC
                    ALPAKA_FN_HOST             
#else                                          
                    ALPAKA_NO_HOST_ACC_WARNING 
                    ALPAKA_FN_HOST_ACC         
#endif                                         
                    static auto alloc(         
```
Same for `free()`.
Not sure, if alpaka currently supports device-sided memory allocations, but in principle it is possible, so I would not 'hostify' these two functions.

tdd11235813:
rebased upstream and fixed some subsequent issues. Tests succeeded on Vega64 and GV100.
Not sure, if travis runs through now, rocRand received an update which let travis fail in a previous version of this PR.

This requires and includes #668 .

ax3l:
as a side note: allocation-wise, we should make sure we can transparently use [mallocMC](https://github.com/ComputationalRadiationPhysics/mallocMC) as an abstraction for accelerator heap memory.

tdd11235813:
runtime tests will work after #687 is merged.
Tomorrow I could integrate the changes from #686 to HIP.
Then it should be ready for review and merge respectively.

(I put the mentioned mallocMC back-end development on our internal ToDo list, maybe we can start on that one within the next months)

alpakaConfig.cmake
```diff
@@ -574,6 +617,206 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+    IF(NOT DEFINED ALPAKA_HIP_VERSION)
+        SET(ALPAKA_HIP_VERSION 1.5.1)
+    ENDIF()
+
+    IF(ALPAKA_HIP_VERSION VERSION_LESS 1.5.1)
+        MESSAGE(WARNING "HIP < 1.5.1 is not supported!")
+        SET(_ALPAKA_FOUND FALSE)
+
+    ELSE()
+        #check if find package requires version for HIP
```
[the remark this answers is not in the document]
tdd11235813: I'll remove it.

alpakaConfig.cmake
```diff
@@ -574,6 +617,206 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                ENDIF()
+
+                IF(CUDA_VERSION VERSION_LESS 9.0)
+                    SET(ALPAKA_HIP_ARCH "20" CACHE STRING "GPU architecture")
+                ELSE()
+                    SET(ALPAKA_HIP_ARCH "30" CACHE STRING "GPU architecture")
+                ENDIF()
+
+                # CUDA 9.0 removed the __CUDACC_VER__ macro. Boost versions lower than 1.65.1 still use this macro.
+                IF(CUDA_VERSION VERSION_GREATER_EQUAL 9.0 AND Boost_VERSION VERSION_LESS 1.65.1)
```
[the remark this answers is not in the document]
tdd11235813: HIP uses nvcc back-end in this code part. So actually every check, that applies to pure CUDA, also has to be done for HIP(nvcc) part.

alpakaConfig.cmake
```diff
@@ -648,7 +900,34 @@ SET(_ALPAKA_INCLUDE_DIRECTORY "${_ALPAKA_ROOT_DIR}/include")
 SET(_ALPAKA_SUFFIXED_INCLUDE_DIR "${_ALPAKA_INCLUDE_DIRECTORY}/alpaka")
 
 SET(_ALPAKA_LINK_LIBRARY)
-LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "${_ALPAKA_LINK_LIBRARY}")
+LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "${_ALPAKA_LINK_LIBRARY}") # TODO: remove?
```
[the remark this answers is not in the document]
tdd11235813: yep, gonna remove it.

alpakaConfig.cmake
```diff
@@ -648,7 +900,34 @@ SET(_ALPAKA_INCLUDE_DIRECTORY "${_ALPAKA_ROOT_DIR}/include")
...
+    HIP_INCLUDE_DIRECTORIES(
+        # ${_ALPAKA_INCLUDE_DIRECTORY}
+        # ${_ALPAKA_INCLUDE_DIRECTORIES_PUBLIC}
+        ${HIP_INCLUDE_DIRS}
+        ${Boost_INCLUDE_DIRS}
+        ${_ALPAKA_ROOT_DIR}/test/common/include
+        )
+
+    IF(OPENMP_FOUND) # remove fopenmp link from nvcc, otherwise linker error will occur
+        ## LIST(REMOVE_ITEM _ALPAKA_COMPILE_OPTIONS_PUBLIC "${OpenMP_CXX_FLAGS}")
```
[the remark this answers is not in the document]
tdd11235813: yup, will be removed.

alpakaConfig.cmake
```diff
@@ -648,7 +900,34 @@ SET(_ALPAKA_INCLUDE_DIRECTORY "${_ALPAKA_ROOT_DIR}/include")
...
+        LIST(REMOVE_ITEM _ALPAKA_LINK_FLAGS_PUBLIC "${OpenMP_CXX_FLAGS}")
+        LIST(APPEND _ALPAKA_LINK_FLAGS_PUBLIC "-Xcompiler ${OpenMP_CXX_FLAGS}")
+        ## LIST(REMOVE_ITEM _ALPAKA_LINK_LIBRARIES_PUBLIC "${OpenMP_CXX_FLAGS}")
```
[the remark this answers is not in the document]
tdd11235813: yup, will be removed.

doc/markdown/user/implementation/mapping/HIP.md
```diff
@@ -0,0 +1,83 @@
...
+  - uses temporary shared value and atomics
+- Workaround for buggy hipStreamQuery, hipStreamSynchronize.
+  - introduces own queue management
+  - hipStreamQuery and hipStreamSynchronize did not work in multithread environment
+- Workaround for missing cuStreamWaitValue32.
+  - polls value each 10ms
+- static device memory and constant memory are still unusable
+- note, that printf in kernels is still not supported in HIP
+- 3D memory is currently disabled
+- dim3 structure is not available on device (use alpaka::vec instead)
```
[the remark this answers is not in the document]
tdd11235813: good point. Also going to update that.

tdd11235813:
updated git log (alpaka::vec::Vec), and HIP.md (mentioned HCC issues are now the same as in git log). removed TODO comments (or just the TODO word, e.g. comment on workaround for syncthreads_count ...).

include/alpaka/core/BoostPredef.hpp
BenjaminW3: Could you please implement the undef solution after for now? I would like to keep the implementation here exactly the one that is within boost. Furthermore, if `BOOST_LANG_CUDA` is defined by boost itself, we still have both languages defined.

include/alpaka/mem/buf/cpu/Set.hpp
```diff
@@ -151,9 +151,17 @@ namespace alpaka
                             {
                                 meta::ndLoopIncIdx(
                                     extentWithoutInnermost,
+
+                                    // workaround for HIP(HCC) to
+                                    // avoid forbidden host-call
+                                    // within host-device functions
+                                    #if defined(BOOST_COMP_HCC) && BOOST_COMP_HCC
+                                    ALPAKA_FN_HOST_ACC
```
BenjaminW3: Can we add this attribute unconditionally?

include/alpaka/core/BoostPredef.hpp
tdd11235813: ah ok, yes sure.

tdd11235813: I'll remove that block above, and insert an undef after `BOOST_LANG_HIP` is defined:
```C++
// ...
#define BOOST_LANG_HIP BOOST_VERSION_NUMBER_AVAILABLE
#if defined(BOOST_LANG_CUDA) && BOOST_LANG_CUDA      
    #undef BOOST_LANG_CUDA                           
#endif                                               
// ...
```

tdd11235813: or I also could set it to `BOOST_VERSION_NUMBER_NOT_AVAILABLE` ?

script/travis/run_tests.sh
```diff
@@ -25,12 +25,13 @@
...
 
 : ${ALPAKA_ACC_GPU_CUDA_ENABLE?"ALPAKA_ACC_GPU_CUDA_ENABLE must be specified"}
+: ${ALPAKA_ACC_GPU_HIP_ENABLE?"ALPAKA_ACC_GPU_HIP_ENABLE must be specified"}
 
 # https://stackoverflow.com/questions/42218009/how-to-tell-if-any-command-in-bash-script-failed-non-zero-exit-status
 err=0
 trap 'err=1' ERR
 
-if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "OFF" ]
+if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "OFF" ] && [ "${ALPAKA_ACC_GPU_HIP_ENABLE}" == "OFF" ];
```
BenjaminW3: Can we execute the tests when the platform is set to non-GPU?

include/alpaka/core/BoostPredef.hpp
BenjaminW3: I think `BOOST_VERSION_NUMBER_NOT_AVAILABLE ` would be correct.

BenjaminW3:
I still do not like all this unnecessary code duplication but it looks very good now!

script/travis/run_tests.sh
tdd11235813: HIP(nvcc) and HIP(hcc) requires Nvidia GPU and AMD GPU respectively. In alpaka HIP can only be enabled with the HIP-only mode. So no other back-ends are allowed, if you mean that?

include/alpaka/mem/buf/cpu/Set.hpp
tdd11235813: not directly, because CUDA cannot handle this annotation in front of a lambda function, while HCC requires it due to the more rigid host-device restrictions. There is a chain of functions, where at least one of them is annotated with `ALPAKA_FN_HOST_ACC`, and this is passed as a functor, thus required to be `ALPAKA_FN_HOST_ACC`.

tdd11235813:
Yeah the code duplication is ugly. It should be possible to unify it and to make separate codes only, where it is needed. It should be quite straightforward.

tdd11235813:
ok, I'll update this in a few minutes and then let's call it a day (Feierabend für heute).

tdd11235813:
There was a bug after rebasing in the code (BOOST_LANG_CUDA was undefined), which is fixed now and while waiting for travis here, the travis tests passed on [my side](https://travis-ci.org/tdd11235813/alpaka/builds/450930236), except for one test which run out of time (>50min).
I just restarted it (nvcc10 + Clang4 Debug), so let's see.
If all passed you can merge this,

BenjaminW3:
I do not know what happened but there is no new travis build.

tdd11235813:
ok, just recommited it, now a travis build is also started.

PR #512, opened 2018-06-06
Use inheriting constructor for ViewSubView

BenjaminW3:
This simplifies the implementation and adds a test case for the constructors without offset because this was missing.

### DIFF
--- a/test/unit/mem/view/src/ViewSubViewTest.cpp
+++ b/test/unit/mem/view/src/ViewSubViewTest.cpp
@@ -45,132 +45,208 @@
 #include <type_traits>
 #include <numeric>
 
-BOOST_AUTO_TEST_SUITE(memView)
-
-//#############################################################################
-//! 1D: sizeof(TIdx) * (5)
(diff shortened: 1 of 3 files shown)


test/unit/mem/view/src/ViewSubViewTest.cpp
```diff
@@ -45,132 +45,208 @@
...
+BOOST_AUTO_TEST_SUITE(memView)
+
+//-----------------------------------------------------------------------------
+BOOST_AUTO_TEST_CASE_TEMPLATE(
+    viewSubViewNoOffsetTest,
+    TAcc,
+    alpaka::test::acc::TestAccs)
+{
+    alpaka::test::mem::view::viewSubViewTestNoOffset<TAcc, float>();
+}
```
BenjaminW3: The diff looks like much has changed, but in the end only this single test case is new and the code from the other test case has been extracted into a function that can be reused.

BenjaminW3:
Ready for review ;-)

ax3l:
@sbastrakov @theZiz @tdd11235813 ping

PR #509, opened 2018-06-05
fix test analysis

psychocoderHPC:
Source Files where added two times into the source files list.
This results into a CMake error:

```
CMake Error: Files to be generated by multiple different commands: "/bigdata/hplsim/scratch/widera/alpaka/build/CMakeFiles/headerCheck.dir/src/acc/headerCheck_generated_AccCpuFibers.hpp.cpp.o.cmake"
```

- add explicit the test case main file to the list of source files
- add missing includes for CUDA backend files to pass the test
- change used CUDA version for the analysis test from CUDA 9.1 to CUDA 8.0 (CUDA 9.1 is not supported by clang 4.0.0)

fix #364

- [x] rebase against #517

PR #517, opened 2018-06-08
Tests: fix CMake empty `ALPAKA_CUDA_ARCH`

psychocoderHPC:
In `compileExec.sh` cmake variables where set during the execution of `cmake` e.g. `-DALPAKA_CUDA_ARCH="${ALPAKA_CUDA_ARCH}"`.
In the case where the environment variable `ALPAKA_CUDA_ARCH` is not defined the cuda architecture for
cmake will be set to empty. This overwrites our define for the minimum CUDA architecture which we derivein CMake out of the used CUDA Toolkit version. This results into an crash if we use clang as compiler.

Error:
```
clang-4.0: error: cannot find libdevice for sm_20. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.
```

### DIFF
--- a/script/travis/compileExec.sh
+++ b/script/travis/compileExec.sh
@@ -27,6 +27,36 @@
 set -eo pipefail
 
 #-------------------------------------------------------------------------------
+
+# create a cmake variable definition if an environment variable exists
+#
+# This function can not handle environment variables with spaces in its content.
+#
+# @param $1 environment variable name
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Bug found by @BenjaminW3 PR #509 https://github.com/ComputationalRadiationPhysics/alpaka/pull/509#issuecomment-394950722

PR #520, opened 2018-06-11
Release 0.3.1 to master

ax3l:
Merge the release branch `0.3.1` to `master`.

The release was already tagged in the last commit of `release-0.3.1` but that is fine as well.


ax3l:
Usually, it's fine to tag the last commit of a "dead-end" bugfix branch and publish the tag only.

In the current case, our `master` is still `0.3.0` so we could merge the fixes on top.

For the `0.4.0` release we just need to make them in sync again. There are two ways for that:
- open the PR for the `0.4.0` merge online, but merge on the terminal with `git merge -X theirs ...`

merge

- synchronize beforehand: after we merged #520 we can update `dev` with `git merge -X ours master`

ax3l:
If a `master` or as we use it `stable` branch is useful at all is a different question and I start to wonder nowadays as well.

The idea here was usually, that a non-git-fancy user that just does `git clone` gets something stable.

BenjaminW3:
My problem is not that we could merge 0.4.0 (which currently is develop) onto master (then 0.3.1).
My problem is that after merging this (after resolving some potential merge conflicts), develop and master will not share a common history anymore.

ax3l:
Yep, got that. Did you see the 2nd part of my message of branches with divergent history are synchronized?

BenjaminW3:
Yes, I know that we can merge back. However, then we still have those ugly merge commits and the history is still not the same (but the current code is the same). Maybe this is only an esthetical problem ;-)

ax3l:
Yes, nobody likes merge commit, I agree :-D

BenjaminW3:
If we really do not like the way it looks in the future, we could still force push to master or simply delete it ;-)

ax3l:
Noooooo. Never force push to mainline branches.

But I would be ok if you decide only to have a `develop` and "dead-end" release branches for alpaka. That's up to you, as long as it's transparent.

PR #522, opened 2018-06-11
Docs: Zenodo JSON

ax3l:
Add the authors of the [last releases](http://dx.doi.org/10.5281/zenodo.57867) and the meta data we add manually (references, keywords, grants, etc.).

This will autofill and automate the release process. Please add new software authors accordingly over time where applicable.


ax3l:
@BenjaminW3 sorry, fixed a typo and pushed again

ax3l:
Ah ok, was not sure.

That's just another option in the `branch` setting to dismiss reviews if something changes. We can enable or disable it as you like. Usually it makes sense but I find it rather inconvenient as well.

BenjaminW3:
> Usually it makes sense but I find it rather inconvenient as well.

+1

PR #527, opened 2018-06-12
Rename BOOST_ARCH_CUDA_DEVICE to BOOST_ARCH_PTX to match naming in boost

BenjaminW3:


include/alpaka/core/Common.hpp
```diff
@@ -55,9 +55,9 @@
 //-----------------------------------------------------------------------------
 // CUDA device architecture detection
 #if defined(__CUDA_ARCH__)
```
ax3l: @BenjaminW3 shouldn't we add a BOOST guard here? e.g. < 1.58

BenjaminW3: Yes, but I would like to keep those changes separate. We should do this when testing the 1.58 RC.

ax3l: but if we add the guard already alpaka will not break with boost > 1.57, no?

ax3l: you could also just a `#ifndef BOOST_ARCH_PTX` which is probably the best solution.

BenjaminW3: We can not be sure that it will really be part of 1.58 or that it works the way we intended to. Therefore, the check could also make it break as well.

BenjaminW3: The ifndef is the solution I had in mind when doing this renaming.

tdd11235813:
FYI, for HIP backend I would move to `BOOST_ARCH_HSA` ([HSA](https://gpuopen.com/rocm-do-you-speaka-my-language/)), which seems more appropriate anyway, instead of `BOOST_ARCH_HIP_DEVICE` now.

include/alpaka/core/Common.hpp
ax3l: @BenjaminW3 you pushed the PR again but the change with the `ifndef` is not in yet. Did you for get to `git add` the changed file first?

BenjaminW3: Oh, I did not want to do the ifndef thing within this PR. I just had to fix a merge conflict with the merged AtomicTest fixes.

tdd11235813:
side issue: can [this](https://github.com/BenjaminW3/alpaka/blob/788ebcff70e746b490d83087ed1a030bd0ce44ba/include/alpaka/core/Common.hpp#L140) 
```C++
#if BOOST_LANG_CUDA && !BOOST_COMP_CLANG_CUDA
```
be changed to
```C++
#if BOOST_COMP_NVCC
```
?
Asking this because the latter one would cover HIP(nvcc) automatically.

BenjaminW3:
I think this should be possible. I am thinking about removing `BOOST_COMP_CLANG_CUDA` completely.

Issue #525, opened 2018-06-12
CUDA 9.2 fails to compile


tdd11235813:
I want to test alpaka with CUDA 9, especially CUDA 9.2. Regarding CUDA 9.0 and 9.1 see issue #426 .
At the moment I use the modules on ZIH/Taurus cluster. Trying to compile `examples/bufferCopy` from current develop. At the moment on Taurus CUDA 8.0.61 is only working (with GCC 5.3.0 and Boost 1.64.0). CUDA 9.x requires at least Boost 1.65.1. I still need to build & try Boost with gcc 5.4 as it is missing on Taurus.

|CUDA 9.2.88|Boost >=1.65.1|
|---|---|
|gcc 4.9.3|:x:|
|gcc 5.3|:x:|
|gcc 5.4|:question:|
|gcc 5.5|:x:|
|gcc 6.x|:x:|
|gcc 7.x|:x:|

### GCC 7.1 + Boost 1.65.1 + CUDA 9.2.88

```
Currently Loaded Modules:
  1) modenv/classic (S)   2) cmake/3.10.1   3) git/2.15.1   4) gcc/7.1.0   5) bullxmpi/1.2.8.4   6) boost/1.65.1-gnu7.1   7) cuda/9.2.88
```
```bash
# alpaka/example/bufferCopy/build_cuda
cmake .. && make
```
Fails with:

```bash
[ 50%] Building NVCC (Device) object CMakeFiles/bufferCopy.dir/src/bufferCopy_generated_bufferCopy.cpp.o
[...]

PR #530, opened 2018-06-14
[WIP] Updates cmake compiler checks (CUDA9.2 +gcc|clang, CUDA+Fiber+?)

tdd11235813:
> NVCC 8.0 does not support clang 4+.  Please use clang 3.9!

vs.

> Clang versions < 4.0 are not supported!

is irritating, so I started to change the cmake error outputs including the following cases:
- CUDA 9.2 does not support either gcc or clang, see #525 (have not found a working combination yet)
- CUDA 9.0/9.1 only works with =<gcc4.9 due to the std::tuple error, see [here](https://github.com/ComputationalRadiationPhysics/picongpu/issues/2388) (an alpaka test case for std::tuple error would be nice)
- Fiber and CUDA backend should not be enabled at the same time (have not tested TBB, but probably the same?)


ax3l:
True, nvcc 8.0 supports clang 3.8 & 3.9 as host compiler: https://gist.github.com/ax3l/9489132

alpakaConfig.cmake
```diff
@@ -122,7 +122,13 @@ ENDIF()
 #-------------------------------------------------------------------------------
 # Check supported compilers.
 IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.0)
```
ax3l: I think this check is probably just too strict.

alpakaConfig.cmake
```diff
@@ -385,29 +391,29 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
                 ELSEIF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
                     IF(CUDA_VERSION VERSION_EQUAL 8.0)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 4.0)
-                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support clang 4+. Please use clang 3.9!")
+                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support clang 4+. Please use NVCC 9.0!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.0)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 4.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.0 does not support clang 4+. Please use clang 3.9!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.0 does not support clang 4+. Please use NVCC 9.0!")
```
tdd11235813: my bad, this should be `NVCC 9.1 or clang 3.8+`.

alpakaConfig.cmake
```diff
@@ -385,29 +391,29 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                         ENDIF()
                     ELSEIF((CUDA_VERSION VERSION_EQUAL 9.0) OR (CUDA_VERSION VERSION_EQUAL 9.1))
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 6.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.0-9.1 do not support GCC 7+ and fail compiling the std::tuple implementation in GCC 6+. Please use GCC 4.9 or 5!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.0-9.1 do not support GCC 7+ and fail compiling the std::tuple implementation in GCC 6+. Please use GCC 4.9!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.2)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 8.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.2 does not support GCC 8+. Please use GCC 4.9, 5, 6 or 7!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.2 does not support GCC!")
```
tdd11235813: the new error message sounds like as nvcc officially does not support the gcc, so maybe better to write `NVCC 9.2 and GCC 4.9 up to GCC 8 will fail to compile!`

tdd11235813:
at least clang 3.9.1 and nvcc 8 does not work.
```
nvcc fatal   : The version ('30901') of the host compiler ('clang') is not supported
```
Have no modules for clang 3.9 and 3.8 on Taurus, so no further testing at the moment.

alpakaConfig.cmake
ax3l: @BenjaminW3 before we change all of the above, do we even have problems with clang 3.9?

BenjaminW3: We have removed some special casing for clang 3.x 3 months ago. As far as I remember it was about OpenMP support, sanitizers and native CUDA compilation. It could be that we still can compile most user code with clang 3.9 or 3.8 but we simply do not have the possibility to test all those combinations in CI.

tdd11235813:
thanks for your help. I still fail to run all the tests successfully, but this probably will be an own issue.
The CUDA+Fibers conflict check is still not exhaustive, ie there are other backend combos we can disallow per cmake, if wanted. Or is CUDA + anything else generally bad? Instead of an error, maybe it is better to use a warning instead. I would not automatically disable one of the backends by cmake to make it run in the first place as the user might miss the missing backend.

alpakaConfig.cmake
ax3l: I am specifically asking about clang 3.9 which is still in wide use and if we test for clang 3.9 we could skip all of the corner-cases below.

BenjaminW3:
CUDA + Fibers should be the only combination that does not work together at the moment. However, this is not a general problem but a compiler and or library (fibers) problem which can possibly be solved.

alpakaConfig.cmake
BenjaminW3: Clang 3.9 as native CUDA compiler does not work without the workarounds we explicitly removed. If I remember correctly, it did not have support for cuRand. I do not want to re-add those workarounds and it would not get us anything to support this combination.
Clang 3.9 as a host compiler for NVCC might work but we already support the last 3 clang releases of the last 1,5 years. When the current 0.4.0 version gets released there will most probably already be clang 7.

ax3l: Ok, good with me. I just did not remember and the check here did look like a non-cuda accelerator exclusion. But in that case it makes sense.

alpakaConfig.cmake
```diff
@@ -389,25 +395,25 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 8.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.2 does not support GCC 8+. Please use GCC 4.9, 5, 6 or 7!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.2 in combination with GCC is not yet supported by alpaka!")
                         ENDIF()
                     ENDIF()
                 ELSEIF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")
                     IF(CUDA_VERSION VERSION_EQUAL 8.0)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 4.0)
-                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support clang 4+. Please use clang 3.9!")
+                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support clang 4+. Please use NVCC 9.0!")
```
ax3l: shouldn't you advise NVCC 9.1 then? (see below)

alpakaConfig.cmake
```diff
@@ -122,7 +122,13 @@ ENDIF()
...
 # Check supported compilers.
 IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.0)
-    MESSAGE(WARNING "Clang versions < 4.0 are not supported!")
+    MESSAGE(FATAL_ERROR "Clang versions < 4.0 are not supported!")
+    SET(_ALPAKA_FOUND FALSE)
+ENDIF()
+
+# Boost Fiber + CUDA will not work
+IF(ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE AND ALPAKA_ACC_GPU_CUDA_ENABLE)
+    MESSAGE(FATAL_ERROR "Fibers and CUDA back-end should not be enabled both at the same time.")
```
ax3l: should not -> must not

BenjaminW3: or "can not"

alpakaConfig.cmake
```diff
@@ -122,7 +122,13 @@ ENDIF()
 #-------------------------------------------------------------------------------
 # Check supported compilers.
 IF(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.0)
-    MESSAGE(WARNING "Clang versions < 4.0 are not supported!")
+    MESSAGE(FATAL_ERROR "Clang versions < 4.0 are not supported!")
+    SET(_ALPAKA_FOUND FALSE)
+ENDIF()
+
+# Boost Fiber + CUDA will not work
```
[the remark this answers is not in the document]
BenjaminW3: 👍 Remove the comment

alpakaConfig.cmake
```diff
@@ -389,25 +395,25 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.2)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 8.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.2 does not support GCC 8+. Please use GCC 4.9, 5, 6 or 7!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.2 in combination with GCC is not yet supported by alpaka!")
```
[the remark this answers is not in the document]
BenjaminW3: We could combine those two 9.2 messages for nvcc+gcc and nvcc+clang but they are failing for different reasons. Therefore I added separate messages.

alpakaConfig.cmake
BenjaminW3: 👍

alpakaConfig.cmake
```diff
@@ -389,25 +395,25 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                            MESSAGE(FATAL_ERROR "NVCC 9.0 does not support clang 4+. Please use NVCC 9.1!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.1)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.0)
                             MESSAGE(FATAL_ERROR "NVCC 9.1 does not support clang 5+. Please use clang 4!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.2)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 6.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.2 does not support clang 6+. Please use clang 4 or 5!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.2 combination with clang is not yet supported by alpaka!")
```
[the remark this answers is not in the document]
BenjaminW3: see other message

alpakaConfig.cmake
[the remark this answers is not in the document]
BenjaminW3: @tdd11235813 You can remove your changes for this line because it will be fixed with my open Pull Request. clang + CUDA 9.2 is still an open issue.

tdd11235813:
thanks again for your help and patience :) Please check again. 
PS: do you want me to squash these commits before merge? Also have not rebased with current upstream/develop yet.

alpakaConfig.cmake
```diff
@@ -395,19 +400,19 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                            MESSAGE(FATAL_ERROR "NVCC 9.0 does not support clang 4+. Please use NVCC 9.1!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.1)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.0)
                             MESSAGE(FATAL_ERROR "NVCC 9.1 does not support clang 5+. Please use clang 4!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 9.2)
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 6.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.2 does not support clang 6+. Please use clang 4 or 5!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.2 in combination with clang is not yet supported by alpaka!")
```
BenjaminW3: I just saw that we already support CUDA 9.2 when using clang 4.0. Clang 5.0 seems to be a problem. So this should be changed to `NVCC 9.2 does not support clang 6+ and fails compiling with clang 5. Please use clang 4!`

tdd11235813:
Incorporated your review. Commits squashed to single one for merging.

PR #533, opened 2018-06-17
Add support for CUDA 9.2 using gcc

BenjaminW3:
The error we have seen comes from the wrong placing of `ALPAKA_NO_HOST_ACC_WARNING` for one function. This seems to be supported by nvcc < 9.2.


BenjaminW3:
Clang does not seem to work for now... I will investigate this in a separate issue.

ax3l:
@BenjaminW3 can you rebase as well, so the other PR fixes are removed from the online diff?

BenjaminW3:
I have already rebased. So it seems I have missed those newline changes in the previous PR 😭

BenjaminW3:
I will move them into a separate PR.

ax3l:
Nevermind, I was actually referring to another change but I mismatched two PRs ;-)

BenjaminW3:
Ready to be merged! (I am not sure if the setting to remove approvals on each update is good for productivity)

include/alpaka/dev/cpu/SysInfo.hpp
```diff
@@ -59,7 +59,7 @@ namespace alpaka
             namespace detail
             {
 #if BOOST_ARCH_X86
-    #if BOOST_COMP_GNUC || BOOST_COMP_CLANG || (!BOOST_COMP_MSVC_EMULATED && __INTEL_COMPILER)
+    #if BOOST_COMP_GNUC || BOOST_COMP_CLANG || (!BOOST_COMP_MSVC_EMULATED && defined(__INTEL_COMPILER))
```
psychocoderHPC: @BenjaminW3 
- Is this `defined` fix only an issue with CUDA 9.2 or is it an general issue? 
- Do we need this `defined` to have INTEL support or is is only a stile change?

Issue #535, opened 2018-06-18
Separate CUDA back-ends nvcc & clang(ptxas) in README.md?


tdd11235813:
The README.md says clang 4/5/6 (native) supports CUDA 8-9.1, but clang 4/5 only supports CUDA 8 ([source](https://gist.github.com/ax3l/9489132#clang--x-cuda)). 
Maybe it is also better to add a separate row for "clang (CUDA)" as accelerator back-end to keep it more separated from the CUDA (nvcc) + clang selection? This would require a change to the matrix layout, also adding two additional columns, when CUDA 9.2 and clang 7 will be part of the matrix.

|Accelerator Back-end|gcc 4.9.2|gcc 5.4|gcc 6.3/7.2|clang 4| clang 5 |clang 6| clang 7 | MSVC 2017.7
|---|---|---|---|---|---|---|---|---|
|CUDA (nvcc)|:white_check_mark: (CUDA 8.0-9.1)|:white_check_mark: (CUDA 9.0-9.1)|:x:|:white_check_mark: (CUDA 9.1, 9.2?)|:white_check_mark: (CUDA 9.2?) | :x: | :x: |:x:|
|CUDA (clang) | :no_entry_sign: | :no_entry_sign: | :no_entry_sign: | :white_check_mark: (CUDA 8.0)| :white_check_mark: (CUDA 8.0)| :white_check_mark: (CUDA 8.0-9.0) | :white_check_mark: (CUDA 9.2 ?) | :no_entry_sign: |

- of course, clang calls `ptxas` from CUDA again, but instead of referring to `ptxas` or `nvcc`, `CUDA` should be sufficient, shouldn't be?

BenjaminW3:
I think this could also be solved by simply adding the supported versions to the table entries containing `native`. This is currently already done for nvcc e.g. `nvcc 8.0 -9.1` so this is simply missing for the entries containing `native`.
I am not sure if those separate comlumns with the  :no_entry_sign: for n/a provide more insigths than they confuse.
In both cases you will have to introduce new columns for the separate clang versions.
Could you do this directly within your current PR (and squash everything)?

BenjaminW3:
I have created a PR and done it myself ;-)

tdd11235813:
[...]

PR #536, opened 2018-06-18
Update readme for CUDA support

BenjaminW3:


ax3l:
I have to admit that "native CUDA" is not a term people will understand for `clang -x` / the clang CUDA frontend / clang as a device compiler

BenjaminW3:
You can open a new ticket if you have a better idea or want to collect ideas ;-)

ax3l:
Haha, that's my review comment since we update it right now ;-)

#535 suggests splitting it more visually which might not be wrong, since clang will always be a big lagging behind nvcc in supported CUDA range and the difference between host and device support for CUDA releases might be hard for users (clang will already have official device support before Nvidia even lists the version as supported host side, etc.)

PR #540, opened 2018-06-20
 Replaces C rand with C++11 random engine (test, example).

tdd11235813:
Replaces C random generator with C++11 random engine to create uniformly distributed numbers.

### DIFF
--- a/example/vectorAdd/src/main.cpp
+++ b/example/vectorAdd/src/main.cpp
@@ -21,7 +21,7 @@
 
 #include <alpaka/alpaka.hpp>
 
-#include <cstdlib>
+#include <random>
 #include <iostream>
 #include <typeinfo>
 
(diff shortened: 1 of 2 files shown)


BenjaminW3:
I had used `rand` because it is widely known and less to type than the c++11 random generators which might be better for examples. On the other hand I can understand that more modern features might be more consistent.

BenjaminW3:
You have to rebase and resolve the conflicts.

tdd11235813:
ok, rebased and pushed. 
Yeah `std::rand()` is just one line, but shall be deprecated some day as it has some problems (global state, quality, ..), so I thought it is better not to use and teach the old style. There are proposals like [this one](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0347r1.html) to simplify the use of `<random>`.

Issue #537, opened 2018-06-19
Failed tests with 'make test'


tdd11235813:
Can you run the runtime tests successfully?

I used this module and cmake settings:
```bash
module load cmake cuda/8.0.61 gcc/4.9.3
cmake -DBOOST_ROOT="$HOME/software/boost_1.65.1/taurus/" -DBOOST_LIBRARYDIR="$HOME/software/boost_1.65.1/taurus/lib" -DBoost_USE_STATIC_LIBS=ON -DBoost_USE_MULTITHREADED=ON -DBoost_USE_STATIC_RUNTIME=OFF -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF .. 
make -j 8
make test
```
### ~2:axpy~ (fixed)
* due to binary comparison of floats (should be `fabs(x-y)>eps` instead of `x!=y`)

### ~3:cudaOnly~ (fixed)
* empty test tree gives error.
It seems to be necessary, to build a second time with `ALPAKA_ACC_GPU_CUDA_ONLY_MODE`  set in cmake with other backends manually disabled, and there `make cudaOnly` will provide the test.
Have not investigated this, but isn't this possible to have this mode integrated in one build step?

### ~12:event~ (fixed)

```
12/21 Testing: event
 12/21 Test: event
 Command: "/home/matwerne/cuda-workspace/alpaka/build/test/unit/event/event"
 Directory: /home/matwerne/cuda-workspace/alpaka/build/test/unit/event
 "event" start time: Jun 19 15:27 CEST
 Output:
 ----------------------------------------------------------
 Running 20 test cases...
 unknown location(0): fatal error: in "event/eventTestShouldBeFalseWhileInQueueAndTrueAfterBeingProcessed<std__tuple<alpaka__dev__DevCudaRt,_alpaka__queue__QueueCudaRtAsync>>": std::runtime_error: /home/matwerne/cuda-workspace/alpaka/test/common/include/
 alpaka/test/event/EventHostManualTrigger.hpp(588) cuStreamWaitValue32( static_cast<CUstream>(queue.m_spQueueImpl->m_CudaQueue), reinterpret_cast<CUdeviceptr>(event.m_spEventImpl->m_devMem), 0x01010101u, CU_STREAM_WAIT_VALUE_GEQ) : 'unrecognized error co
 de': 'unrecognized error code'!
[...]

PR #544, opened 2018-06-23
Fix cuda memcpy and memset for zero sized buffers (division by zero)

BenjaminW3:
This fixes the failing memBuf test from #537

### DIFF
--- a/include/alpaka/mem/buf/cuda/Copy.hpp
+++ b/include/alpaka/mem/buf/cuda/Copy.hpp
@@ -563,7 +563,7 @@ namespace alpaka
                                 dstNativePtr,
                                 static_cast<std::size_t>(dstPitchBytesX),
                                 static_cast<std::size_t>(dstWidth),
-                                static_cast<std::size_t>(dstPitchBytesY/dstPitchBytesX));
+                                static_cast<std::size_t>(dstPitchBytesY / dstPitchBytesX));
                         cudaMemCpy3DParms.extent =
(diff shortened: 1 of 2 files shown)


ax3l:
Maybe @psychocoderHPC can take a look tomorrow and @tdd11235813 can verify it for his bugreport :)

psychocoderHPC:
looks good from my side but I will wait until @tdd11235813 responded before I give a :+1:

tdd11235813:
The memBuf test works now, thanks a bunch. 
I just would like to understand the parts and to know, which cases allow the use of 0s in extents. 
I guess the zero size test case is about *"don't fail at the basics even if there are extents with zeros"*. 
Now, the enqueue for copy and (mem)set would silently return before doing anything, but it makes sense here. The ZeroSizeTest also calls the backend specific alloc function with 0 bytes, which is not well defined (it's the question if alpaka should take care of this by returning NULL pointer if size is 0).
Subsequent functions in basicBufferOperationsTest and viewTestMutable include `copy` and `verifyViewsEqual`. `copy` and `set` should not do anything and the `verifyViewsEqualKernel`/`verifyBytesSetKernel` runs with 1 Thread, but the views should be empty, so nothing happens.

BenjaminW3:
Yes, I think you summarized everything correctly. 👍 
We skip the call to the underlying CUDA memory operation now. This should not be a problem because it would not do anything.

PR #547, opened 2018-06-24
Fix VS2017 CUDA build

BenjaminW3:
This fixes the build of Visual Studio 2017 combined with CUDA 9.2 (even though this is not a supported combination and needs a local hack).
I am not marking this as a supported combination in the ReadMe because we are not currently testing this in CI.


alpakaConfig.cmake
```diff
@@ -586,7 +584,7 @@ SET_SOURCE_FILES_PROPERTIES(${_ALPAKA_FILES_CMAKE} PROPERTIES HEADER_FILE_ONLY T
 append_recursive_files_add_to_src_group("${_ALPAKA_ROOT_DIR}/doc/markdown" "${_ALPAKA_ROOT_DIR}" "md" _ALPAKA_FILES_DOC)
 SET_SOURCE_FILES_PROPERTIES(${_ALPAKA_FILES_DOC} PROPERTIES HEADER_FILE_ONLY TRUE)
 
-SET(_ALPAKA_FILES_OTHER "${_ALPAKA_ROOT_DIR}/.gitignore" "${_ALPAKA_ROOT_DIR}/.travis.yml" "${_ALPAKA_ROOT_DIR}/appveyor.yml" "${_ALPAKA_ROOT_DIR}/COPYING" "${_ALPAKA_ROOT_DIR}/COPYING.LESSER" "${_ALPAKA_ROOT_DIR}/README.md")
+SET(_ALPAKA_FILES_OTHER "${_ALPAKA_ROOT_DIR}/.gitignore" "${_ALPAKA_ROOT_DIR}/.travis.yml" "${_ALPAKA_ROOT_DIR}/.zenodo.json" "${_ALPAKA_ROOT_DIR}/appveyor.yml" "${_ALPAKA_ROOT_DIR}/COPYING" "${_ALPAKA_ROOT_DIR}/COPYING.LESSER" "${_ALPAKA_ROOT_DIR}/README.md")
```
BenjaminW3: Unreleate but simply missing

example/bufferCopy/CMakeLists.txt
```diff
@@ -53,18 +53,25 @@ SET(_SOURCE_DIR "src/")
...
+
 IF(MSVC)
     LIST(APPEND _COMPILE_OPTIONS_PRIVATE "/wd4996")   # This function or variable may be unsafe. Consider using <safe_version> instead.
 ENDIF()
 
+# NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
+IF(ALPAKA_ACC_GPU_CUDA_ENABLE AND NOT ALPAKA_CUDA_COMPILER MATCHES "clang")
+    STRING(REPLACE ";" " " _COMPILE_OPTIONS_PRIVATE_STRING "${_COMPILE_OPTIONS_PRIVATE}")
+    SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_COMPILE_OPTIONS_PRIVATE_STRING}")
+ENDIF()
```
BenjaminW3: As was noted recently in one of the issues, `TARGET_COMPILE_OPTIONS` does not work when nvcc is used. Therefore, the `COMPILE_OPTIONS` have to be added to the `CMAKE_CXX_FLAGS` before calling `CUDA_ADD_EXECUTABLE`.
This is not a problem on linux because there the additional options only contain some enhanced warning options. However, on windows the `/bigobj` switch is required for the build to succeed.
With this change, the enhance warning options will also be used for builds with nvcc.

BenjaminW3:
Ready for review!

BenjaminW3:
Oh no, build is red again...

sbastrakov:
I have configuration MSVC 2017 + CUDA 9.2 and it builds fine for me with, e.g.
`cmake CMakeLists.txt -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DBUILD_TESTING=OFF -G "Visual Studio 15 2017 Win64"`
(disabled testing because of my weird Boost setup, that's not the point).
I explicitly disabled Fibers as it is not supported with CUDA (as the error message says), which seems fine. However, if I use
`cmake CMakeLists.txt -DALPAKA_ACC_GPU_CUDA_ONLY_MODE=ON -DBUILD_TESTING=OFF -G "Visual Studio 15 2017 Win64"`
the Fibers error still appears. Is it supposed to work like that? My understanding is that if one wants only GPU backends Fibers should not matter?

BenjaminW3:
`cmake CMakeLists.txt -DALPAKA_ACC_GPU_CUDA_ONLY_MODE=ON -DBUILD_TESTING=OFF -G "Visual Studio 15 2017 Win64"` is not supposed to work right now. We may want to fix this by disabling all accelerator backends except CUDA if this flag is given. Please create a new issue for this. Currently you have to disable all other backends by hand in this case.

alpakaConfig.cmake
ax3l: hm, would be better in a separate commit, ideally separate PR.

makes prs smaller, more portable and easier to review and handle regressions.

example/bufferCopy/CMakeLists.txt
ax3l: On windows you can set `/bigobj` generally, you need it all the time and it's just not set by default for some backwards compatibility reasons with an ancient windows/msvc version.

include/alpaka/vec/Vec.hpp
```diff
@@ -335,7 +337,7 @@ namespace alpaka
             {
                 core::assertValueUnsigned(iIdx);
                 auto const idx(static_cast<typename TDim::value_type>(iIdx));
-                assert(TDim::value > 0u && idx<TDim::value);
+                assert(idx<TDim::value);
```
ax3l: a bit easier to read as `assert(idx < TDim::value);`

include/alpaka/vec/Vec.hpp
```diff
@@ -353,7 +355,7 @@ namespace alpaka
...
```
ax3l: a bit easier to read with spaces around `<`

alpakaConfig.cmake
BenjaminW3: extracted

BenjaminW3:
Now really ready for review

BenjaminW3:
Push

test/unit/kernel/CMakeLists.txt
```diff
@@ -58,6 +58,13 @@ SET(_SOURCE_DIR "src/")
 # Add all the source files in all recursive subdirectories and group them accordingly.
 append_recursive_files_add_to_src_group("${_SOURCE_DIR}" "${_SOURCE_DIR}" "cpp" _FILES_SOURCE)
 
+# NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
+IF(ALPAKA_ACC_GPU_CUDA_ENABLE AND NOT ALPAKA_CUDA_COMPILER MATCHES "clang")
```
psychocoderHPC: why do we add this not to `ALPAKA_ADD_EXECUTABLE`?

psychocoderHPC: and maybe also to `ALPAKA_ADD_LIBRARY`?

BenjaminW3: The problem is that we would need to know the target dependencies within those functions to correctly add at least the `COMPILE_OPTIONS` at the moment of the `ALPAKA_ADD_EXECUTABLE` call. However, the dependencies are in general only added after creating the target which is too late for FindCUDA.
We could add a convenience function that wraps those five lines. However, I intend to remove all those occurences with a single one within a follow-up PR. Furthermore, this only works because we know that `common` does not have any transitive dependency which defines `COMPILE_OPTIONS`. To handle this correctly, we would need to know transitive dependencies of the given targets and get the properties for all of them. I do not know how to do this correctly.

ax3l: Maybe we could work with generator expressions for `ALPAKA_ADD_EXECUTABLE`/`_LIBRARY`?

BenjaminW3: FindCUDA does not support any such thing. It really only supports those `COMPILE_OPTIONS` which are part of `CMAKE_CXX_FLAGS` at the time `cuda_add_executable` is called. So nothing we could do.

ax3l: We really need to add Clang -x cuda to CMake so we can switch to the CUDACXX language feature and get rid of `FindCUDA` wrappers.

PR #546, opened 2018-06-24
Do not execute tests using EventHostManualTrigger if not supported

BenjaminW3:
Adds a trait and a test method `isEventHostManualTriggerSupported` to not use the `EventHostManualTrigger` when it is not supported. It was always supported with CUDA 8.0 but is only conditionally enabled in CUDA 9.
This fixes the failing tests due to this disabled feature in #537 by checking the trait before executing the test.

PR #548, opened 2018-06-25
Fixes CUDA driver API error handler.

tdd11235813:
- adds error handling to `cudaFree` (e.g. if pointer has been already deallocated, error will occur)
- the cast to `cudaError_t` in `cudaDrvCheck` gave wrong error codes, as runtime and driver API error codes are different, so driver API error handling is used now

Example (refers to #546):
```
/alpaka/test/common/include/alpaka/test/event/EventHostManualTrigger.hpp(650) cuStreamWaitValue32( static_cast<CUstream>(queue.m_spQueueImpl->m_CudaQueue), reinterpret_cast<CUdeviceptr>(event.m_spEventImpl->m_devMem), 0x01010101u, CU_STREAM_WAIT_VALUE_GEQ) : 
'unrecognized error code': 'unrecognized error code'!
```
now correctly reports:
```
'CUDA_ERROR_NOT_SUPPORTED': 'operation not supported'!
```

### DIFF
--- a/include/alpaka/core/Cuda.hpp
+++ b/include/alpaka/core/Cuda.hpp
@@ -177,15 +177,34 @@ namespace alpaka
                 int const & line)
             -> void
             {
-                if(error != CUDA_SUCCESS)
+                if(error == CUDA_SUCCESS)
+                    return;
+
+                char const * cu_err_name = nullptr;
+                char const * cu_err_string = nullptr;
+                CUresult cu_result_name = cuGetErrorName(error, &cu_err_name);
+                CUresult cu_result_string = cuGetErrorString(error, &cu_err_string);
+                std::string sError = std::string(file)
+                                   + "(" + std::to_string(line) + ") "
+                                   + std::string(desc) + " : '";
+                if( cu_result_name == CUDA_SUCCESS && cu_result_string == CUDA_SUCCESS )
                 {
-                    std::string const sError(std::string(file) + "(" + std::to_string(line) + ") " + std::string(desc) + " : '" + cudaGetErrorName(static_cast<cudaError_t>(error)) +  "': '" + std::string(cudaGetErrorString(static_cast<cudaError_t>(error))) + "'!");
-#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
-                    std::cerr << sError << std::endl;
-#endif
-                    ALPAKA_DEBUG_BREAK;
-                    throw std::runtime_error(sError);
+                    sError += std::string(cu_err_name) +  "': '"
+                            + std::string(cu_err_string) + "'!";
+                } else {
+                    // cuGetError*() failed, so append corresponding error message
+                    if( cu_result_name == CUDA_ERROR_INVALID_VALUE ) {
+                        sError += " cuGetErrorName: 'Invalid Value'!";
+                    }
+                    if( cu_result_string == CUDA_ERROR_INVALID_VALUE ) {
+                        sError += " cuGetErrorString: 'Invalid Value'!";
+                    }
                 }
+#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
+                std::cerr << sError << std::endl;
+#endif
+                ALPAKA_DEBUG_BREAK;
+                throw std::runtime_error(sError);
             }
         }
     }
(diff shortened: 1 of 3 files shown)


include/alpaka/core/Cuda.hpp
```diff
@@ -177,15 +177,31 @@ namespace alpaka
...
+                CUresult cu_result_string = cuGetErrorString(error, &cu_err_string);
+                std::string sError;
+                if( cu_result_name == CUDA_ERROR_INVALID_VALUE
+                    || cu_result_string == CUDA_ERROR_INVALID_VALUE ) {
+                    sError = std::string(__FILE__)
+                        + "(" + std::to_string(__LINE__) + ") "
+                        + "CUDA Driver API: Invalid Value.";
+                } else {
+                    sError = std::string(file)
+                        + "(" + std::to_string(line) + ") "
```
BenjaminW3: You could extract the duplicate `sError = std::string(file) + "(" + std::to_string(line) + ") "` diretly to the declaration of the string in front of the if.

include/alpaka/core/Cuda.hpp
```diff
@@ -177,15 +177,31 @@ namespace alpaka
...
+                if(error == CUDA_SUCCESS)
+                    return;
+
+                char const * cu_err_string = nullptr;
+                char const * cu_err_name = nullptr;
+                CUresult cu_result_name = cuGetErrorName(error, &cu_err_name);
+                CUresult cu_result_string = cuGetErrorString(error, &cu_err_string);
+                std::string sError;
+                if( cu_result_name == CUDA_ERROR_INVALID_VALUE
+                    || cu_result_string == CUDA_ERROR_INVALID_VALUE ) {
```
BenjaminW3: Could it be useful to add and check the two CUDA string separately? Then we could print something even when one of the two methods returns an error. At least I do not see a reason to make check them in combination.

include/alpaka/mem/buf/BufCudaRt.hpp
```diff
@@ -118,12 +118,13 @@ namespace alpaka
                 {
                     ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
-                    // Set the current device. \TODO: Is setting the current device before cudaFree required?
+                    // Set the current device.
```
BenjaminW3: Are you sure that `cudaSetDevice` is really necessary for `cudaFree`?

tdd11235813: `cudaMalloc` and `cudaFree` handle device dependent memory, so `cudaSetDevice` should be used in both cases. There is no documentation, that `cudaFree` is context-agnostic. It is true, that `cudaFree` is smart enough to find the correct device for the given memory address, if UVA is present (64-bit, TCC; however not tested the non-UVA case).

I have experienced in this [program](https://github.com/tdd11235813/read-nvml-clocks-pci), that `cudaFree` acts on the currently set device. The example uses a simple `cudaFree(0)` after `cudaSetDevice`, where last one does only lazy context creation/initialization but is a low-overhead call. `cudaFree` uses a given device context and if the device context is wrong, application clock settings do not apply (e.g. set by `nvidia-smi -ac`). If no `cudaFree(0)` is used, clock settings also do not apply at all. You then have to first run a warmup-kernel. It is not a real error, but it shows that cudaFree uses and writes context information.

tdd11235813:
Thanks for your review. Error output is now more clear. Squashed into 1 commit. Please have a look.

include/alpaka/core/Cuda.hpp
```diff
@@ -177,15 +177,44 @@ namespace alpaka
...
+                    sError = std::string(file)
+                        + "(" + std::to_string(line) + ") "
+                        + std::string(desc) + " : '"
+                        + std::string(cu_err_name) +  "': '"
+                        + std::string(cu_err_string) + "'!";
+                } else {
+                    // cuGetError*() failed as well, so print different error message
+                    sError =
+                        std::string(__FILE__)
+                        + "(" + std::to_string(__LINE__) + ") "
```
BenjaminW3: I do not get what those `__FILE__` and `__LINE__` macros will get us. They do not point to any useful line. I would only use the file and line from the original error.
You still only print the available information when both methods were successful. I would prefer to have both appended individually.

tdd11235813: ok, no problem, I am going to change that.

tdd11235813:
Example outputs:
```bash
# success
/EventHostManualTrigger.hpp(650) cuStreamWaitValue32(...) : 'CUDA_ERROR_NOT_SUPPORTED': 'operation not supported'!
# cuGetError* fail, e.g. both fails
/EventHostManualTrigger.hpp(650) cuStreamWaitValue32(...) : cuGetErrorName: 'Invalid Value'! cuGetErrorString: 'Invalid Value'!
```

PR #549, opened 2018-06-26
Move `runtime_error` out of the debug block

bertwesarg:
A user would get the runtime error only in a debug build, which differs to
other runtime errors, like the one in Omp2blocks which checks if
#threads/block == 1.

Converted to `BOOST_VERIFY_MSG` on the way.

### DIFF
--- a/include/alpaka/exec/ExecCpuOmp2Threads.hpp
+++ b/include/alpaka/exec/ExecCpuOmp2Threads.hpp
@@ -169,27 +169,28 @@ namespace alpaka
                         // Therefore we use 'omp parallel' with the specified number of threads in a block.
                         #pragma omp parallel num_threads(iBlockThreadCount)
                         {
-#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
-                            // GCC 5.1 fails with:
-                            // error: redeclaration of const int& iBlockThreadCount
-                            // if(numThreads != iBlockThreadCount
-                            //                ^
-                            // note: const int& iBlockThreadCount previously declared here
-                            // #pragma omp parallel num_threads(iBlockThreadCount)
-                            //         ^
-#if (!BOOST_COMP_GNUC) || (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0))
                             // The first thread does some checks in the first block executed.
-                            if((::omp_get_thread_num() == 0) && (acc.m_gridBlockIdx.sum() == 0u))
+                            if((::omp_get_thread_num() == 0) && (acc.m_gridBlockIdx.sum() == static_cast<TIdx>(0)))
                             {
                                 int const numThreads(::omp_get_num_threads());
+                                boost::ignore_unused(numThreads);
+#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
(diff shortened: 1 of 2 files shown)


include/alpaka/exec/ExecCpuOmp2Threads.hpp
```diff
@@ -169,27 +169,25 @@ namespace alpaka
...
 #endif
+                                // GCC 5.1 fails with:
+                                // error: redeclaration of const int& iBlockThreadCount
+                                // (numThreads == iBlockThreadCount)
+                                //                ^
+                                // note: const int& iBlockThreadCount previously declared here
+                                // #pragma omp parallel num_threads(iBlockThreadCount)
+                                //         ^
+#if (!BOOST_COMP_GNUC) || (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0)) || (BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(6, 0, 0))
+                                BOOST_VERIFY_MSG(numThreads == iBlockThreadCount,
```
BenjaminW3: Could we keep the `throw` for now instead of the `BOOST_VERIFY_MSG`?
We should do such a change consistently across all places after incestigating all options.
E.g. the doc says `BOOST_VERIFY_MSG(expr,msg) expands to ((void)(expr)) when NDEBUG is defined, to BOOST_ASSERT_MSG(expr,msg) when it's not`. This would be a change in behaviour because the original version throws when the `ALPAKA_DEBUG` level is high enough in both, release and debug builds.

bertwesarg: Done. That would have completely circumvent my intention anyway, thanks. Just to get it right, do you mean, that the `BOOST_VERIFY` in `ExecCpuOmp2Blocks.hpp` should be always runtime/compile time error?

bertwesarg: A new PR or just a second commit into this one?

BenjaminW3: 2nd commit should be good.

include/alpaka/exec/ExecCpuOmp2Threads.hpp
```diff
@@ -169,27 +169,27 @@ namespace alpaka
...
-                            // GCC 5.1 fails with:
-                            // error: redeclaration of const int& iBlockThreadCount
-                            // if(numThreads != iBlockThreadCount
-                            //                ^
-                            // note: const int& iBlockThreadCount previously declared here
-                            // #pragma omp parallel num_threads(iBlockThreadCount)
-                            //         ^
-#if (!BOOST_COMP_GNUC) || (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0))
                             // The first thread does some checks in the first block executed.
                             if((::omp_get_thread_num() == 0) && (acc.m_gridBlockIdx.sum() == 0u))
```
BenjaminW3: You will have to change the `acc.m_gridBlockIdx.sum() == 0u` to `acc.m_gridBlockIdx.sum() == static_cast<TIdx>(0)` to makethe compiler happy.

bertwesarg:
All checks have passed now, any further comments? Thanks.

PR #554, opened 2018-06-30
Add support for Vec::operator(<, >=, >)

BenjaminW3:


sbastrakov:
I totally understand the convenience of having it, having done similar stuff myself. Just to point out, from some perspective it is not logical. Having just element-wise >= and <= is fine, each defines a partial order on the set of vectors. However, with element-wise > present, now >= does not mean "greater or equal" in the sense of that element-wise greater and natural equality, and it actually can not be easily expressed as a combination of that two. Again, I still think the convenience overweights this little issue.

BenjaminW3:
Ready for review!

include/alpaka/vec/Vec.hpp
```diff
@@ -721,7 +764,7 @@ namespace alpaka
         }
 
         //-----------------------------------------------------------------------------
-        //! \return The element wise less than relation of two vectors.
+        //! \return The element wise less equal than relation of two vectors.
```
sbastrakov: I think 'element wise' present in many places is not a correct form, should be element-wise or elementwise https://en.wiktionary.org/wiki/elementwise

Also here 'less equal than' should probably be phrased as 'less than or equal'

BenjaminW3: Fixed

PR #559, opened 2018-07-03
Make the examples compilable and executable even when the CPU accelerator is not available

BenjaminW3:
When the CPU sequential accelerator is not available (e.g. CUDA only mode), thos examples should still compile and execute without an error. This is in preparation to execute the examples during the CI tests and for enabling `make test` .

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -26,8 +26,6 @@
 #include <cstdint>
 #include <cassert>
 
-#define DIMENSION 3
-
 //-----------------------------------------------------------------------------
 template <size_t width>
 ALPAKA_FN_ACC size_t linIdxToPitchedIdx(size_t const globalIdx, size_t const pitch)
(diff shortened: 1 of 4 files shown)


example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -138,8 +136,12 @@ struct FillBufferKernel
 auto main()
 -> int
 {
+// This example is hard-coded to use the sequential executor.
+// It can not be compiled if it is not available.
```
sbastrakov: Sorry for being picky about English again. I think this sentence could be improved: 'it' is used two times for different things, and also the point of this commit is so that the test actually compiles now but does nothing. Maybe the first sentence is enough and the second can be just removed?

BenjaminW3: I will remove the second line.

BenjaminW3: done

Issue #543, opened 2018-06-23
Remove the support to build each test on its own


BenjaminW3:
Currently each test can be build completely separate from all other tests. Each test is a new project which searches for `alpaka`, `boost` and the test `common` library and links them. They can be copied out of the alpaka repo and be built from anywhere as long as the `ALPAKA_ROOT` variable is correctly set.
This makes the CMakeLists unnecessarily complex.
The tests CMakeLists could be nearly half the size if we would only search for dependencies and apply test settings in the top-level `test` folders CMakeLists.txt. This is also the de-facto standard for CMake projects. I do not see a reason to extract single tests from the repo. The tests will always be built in combination with the alpaka library itself.
The examples currently have the same features. However, there I can see a reason why one would want to build them independently. Therefore I would keep everything as is for the examples.
What do you think?

ax3l:
yes, tests don't need a full project, `.cmake` snippets are enought

Ideally tests should just be build with the root `CMakeLists.txt` as long as the CMake commonly standardized variable `BUILD_TESTING` is `ON`. This is what #476 added as interface (also #478 keeps this).

Instead of adding a `add_subdirectory` for each test, just `include()` them as `.cmake` files that define `targets` that use dependencies from the root `CMakeLists.txt` (*or* a `find_package(... CONFIG)` of alpaka). #478 tries to unify this further.

PR #560, opened 2018-07-03
Add default google test arguments for execution with ctest / make test

BenjaminW3:
Add some enhanced default command line options to the tests executed via `ctest` or `make test` to enhance the output. Currently they are manually given as command line arguments to the executables in CI.


BenjaminW3:
Yes, we could add such a function. It is not that easy because the tests are currently completely independant. I would add such a thing once #543 is resolved, which I am currently working on.

BenjaminW3:
We could update the year but I do not see it as necessary. We may think about applying a script to the whole repo which updates the year for all files.

PR #566, opened 2018-07-09
NVCC 8.0: GCC <= 5.3

ax3l:
It turns out, GCC <= 5.3 does not yet have the problem with `std::tuple` in NVCC 8.0 and works.

Also applied in PIConGPU https://github.com/ComputationalRadiationPhysics/picongpu/pull/2636 and https://github.com/ComputationalRadiationPhysics/picongpu/pull/2639 .


ax3l:
Interestingly, we have a NVCC 9.0 + GCC 5.4 test in [.travis.yaml](https://github.com/ComputationalRadiationPhysics/alpaka/blob/e4abc9c5f2bfc98222eb29aaa9bc40952cbb675b/.travis.yml#L115)

BenjaminW3:
This is not 5.4, this is `GNU 5.5.0` see here: https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test/+packages?field.name_filter=&field.status_filter=published&field.series_filter=trusty
Ubuntu 14.04 is still at 5.4.1 so we could add yet another configuration...

ax3l:
Ye, NVCC 9.0 & GCC 5.5 seems to work: https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/401499859

alpakaConfig.cmake
```diff
@@ -381,12 +381,12 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
             ELSE()
                 IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
                     IF(CUDA_VERSION VERSION_EQUAL 8.0)
-                        IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.0)
```
BenjaminW3: Could you please retest this by adding a ubuntu 16.04 (5.5) and an ubuntu 14.04 (5.4) build to the travis.yml temporarily?

alpakaConfig.cmake
```diff
@@ -381,12 +381,12 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
-                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support GCC 6+ and fails compiling the std::tuple implementation in GCC 5+. Please use GCC 4.9!")
+                        IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.4)
+                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support GCC 6+ and fails compiling the std::tuple implementation in GCC 5.4+. Please use GCC 4.9 - 5.3!")
                         ENDIF()
                     ELSEIF((CUDA_VERSION VERSION_EQUAL 9.0) OR (CUDA_VERSION VERSION_EQUAL 9.1))
                         IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 6.0)
-                            MESSAGE(FATAL_ERROR "NVCC 9.0-9.1 do not support GCC 7+ and fail compiling the std::tuple implementation in GCC 6+. Please use GCC 4.9 or 5!")
+                            MESSAGE(FATAL_ERROR "NVCC 9.0 - 9.1 do not support GCC 7+ and fail compiling the std::tuple implementation in GCC 6+. Please use GCC 4.9 - 5.5!")
```
BenjaminW3: I would not mention 5.5 here explicitly because ther might be a 5.6 patch release and this will most probably work as well and we will not even find out that there was a new release. `4.9` also does not mention the patch release. gcc changed its versioning scheme between 4.9 and 5 so `4.8`, `4.9`, `5`, `6` should be the way to mention those releases. Only in the rare case that a patch release breaks something, I would explicitly mention it.

ax3l: My thinking was the following:
- 5.6 would still be allowed without a CMake error, but we know *at the time of release* the last tested version and should recommend that in warnings/errors to avoid further, cascading tries
- 5.6 would *not* be a patch release but a *minor*, which can potentially break things (see 5.3 vs 5.4 in NVCC 8.0)

alpakaConfig.cmake
BenjaminW3: Thanks for adding the builds, but you have to change/remove the check to see that it fails due to the tuple error.

ax3l:
the two travis tests failed as intended. will remove now again.

BenjaminW3:
They failed because you explicitly reject this combination. I wanted to let them fail due to the tuple error.

ax3l:
I thought you wanted to see the CMake guard works... Will re-submit the other test.

BenjaminW3:
Hmmm. they fail with
```
/usr/include/c++/5/cmath(362): error: calling a __host__ function("__builtin_logl") from a __device__ function("std::log") is not allowed
/usr/include/c++/5/cmath(633): error: calling a __host__ function("__builtin_isnan") from a __device__ function("std::isnan") is not allowed
```
This is not what we expected. There might be another incompatibility.
I am wondering if the std::tuple error should have been triggered before this error. Could you please try to comment out the add_subdirectory for the examples and the integration tests? I hope that the unit tests show the desired error.

ax3l:
Officially, on release NVCC 8.0 said it supports GCC <= 5.3
Although the header check is missing, I don't know if it is useful to investigate this further:
https://docs.nvidia.com/cuda/archive/8.0/cuda-installation-guide-linux/index.html

alpakaConfig.cmake
```diff
@@ -381,12 +381,12 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
             ELSE()
                 IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "GNU")
                     IF(CUDA_VERSION VERSION_EQUAL 8.0)
-                        IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.0)
-                            MESSAGE(FATAL_ERROR "NVCC 8.0 does not support GCC 6+ and fails compiling the std::tuple implementation in GCC 5+. Please use GCC 4.9!")
+                        IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 5.4)
+                            MESSAGE(WARNING "NVCC 8.0 does not support GCC 6+ and fails compiling the std::tuple implementation in GCC 5.4+. Please use GCC 4.9 - 5.3!")
```
ax3l: Should just read:
NVCC 8.0 does not support GCC 5.4+.

alpakaConfig.cmake
ax3l: What would be good is if we could find a way to build a GCC 5.3 with CUDA 8.0.
Nevertheless, we do locally on Hypnos.

PR #571, opened 2018-07-15
Enhance CMakeLists.txt

### LINKED ISSUE #550: Setting ALPAKA_ACC_GPU_CUDA_ONLY_MODE does not disable other backends
The issue was originally mentioned in [this discussion](https://github.com/ComputationalRadiationPhysics/alpaka/pull/547)

It seems setting this flag does not actually disable other backends, contradicting what the name suggests.

BenjaminW3:
* moved some CUDA compatibility checks to CMake (CUDA + fibers and clang CUDA + OMP are not supported)
* disabled such unsupported backend combinations by default if CUDA is enabled
* pass enabled accelerators to docker only if they are defined
* removed default definition of accelerator in travis environment (`ALPAKA_ACC_GPU_CUDA_ENABLE` and `ALPAKA_ACC_GPU_CUDA_ONLY_MODE`) will be removed later because they still need some work in the CI scripts
* Fixes #550 by disabling the CPU backends by default when `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is on. If the CPU backends are enabled on the command line, we correctly print an error.


script/travis/install.sh
```diff
@@ -40,7 +40,7 @@ if [ "${ALPAKA_CI_ANALYSIS}" == "ON" ] ;then ./script/travis/install_analysis.sh
 if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" ] ;then ./script/travis/install_cuda.sh ;fi
 if [ "${CXX}" == "g++" ] ;then ./script/travis/install_gcc.sh ;fi
 if [ "${CXX}" == "clang++" ] ;then source ./script/travis/install_clang.sh ;fi
-if [ "${ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE}" == "ON" ] ;then ./script/travis/install_tbb.sh ;fi
+if [[ ! -v ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE || "${ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE}" == "ON" ]] ;then ./script/travis/install_tbb.sh ;fi
```
psychocoderHPC: Why do you install `TBB` if the environment variable `ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE` is not set?

script/travis/install_boost.sh
```diff
@@ -65,7 +65,7 @@ then
 fi
 # Select the libraries required.
 ALPAKA_BOOST_B2+=" --with-test"
-if [ "${ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE}" == "ON" ]
+if [[ ! -v ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE || "${ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE}" == "ON" ]]
```
psychocoderHPC: Same as above, why is fiber installed if the variable `ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE` is not set?

script/travis/install.sh
BenjaminW3: Currently we install tbb only if the Flag is ON. However the CMakeLists.txt also uses the backend if the flag was not explicitly given but the backend is available. Now that I use this feature of the CMakeLists, I did not want to add the TBB enable flag to every CI job but have it available by default if this flag is not given to make tue handling equivalrnt to the CMakeLists.txt.
We ma want to differentiate between installation and enabling by using another variable, but this can be done later if it would be required to test such strange combinations.

script/travis/install_boost.sh
BenjaminW3: Same as above

script/travis/install.sh
psychocoderHPC: But is this if condition now not always true? If so we can remove the if fully

BenjaminW3:
Push 😉

script/travis/install.sh
psychocoderHPC: in that case please add a comment over the if condition that it is clear if someone else read the code or if I forget this pr.

ax3l:
Holidays ;-) Monday ;)
--
Sent from my smartphone.
Please excuse my brevity.

script/travis/install.sh
BenjaminW3: I have added a comment

BenjaminW3:
@ax3l It is monday ;-)

Issue #575, opened 2018-08-02
Rng: STL MT has huge state


ax3l:
The currently implemented `std::mt19937` state of the [`alpaka::rand::RandStl`](https://github.com/ComputationalRadiationPhysics/alpaka/blob/0d5657a474334d821b43616bd0e92b08643ddc8d/include/alpaka/rand/RandStl.hpp#L51-L70) is of 2KByte size.

That's way too large for most applications, as seen by @psychocoderHPC here:
https://github.com/ComputationalRadiationPhysics/picongpu/issues/2653

Maybe we also want to implement some sort of TinyMT, e.g.
http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/TINYMT/
https://github.com/MersenneTwister-Lab/TinyMT ([32-bit header](https://github.com/MersenneTwister-Lab/TinyMT/blob/master/tinymt/tinymt32.h), [usage example](https://stackoverflow.com/a/46186567/2719194))

Refs:
http://www.cplusplus.com/reference/random/mt19937/
https://en.cppreference.com/w/cpp/numeric/random/mersenne_twister_engine

ax3l:
Fun-fact, there is also *Mersenne Twister for Graphic Processors* (https://github.com/MersenneTwister-Lab/MTGP), might be the one in the CUDA SDK...

psychocoderHPC:
Also the std library give us a lot of other possibilities https://en.cppreference.com/w/cpp/numeric/random

ax3l:
The other stdlib rngs, only the lcg with small state, have horrible properties :)

sbastrakov:
@ax3l not all LCGs are evil, however should be used with caution ofc. In a way, Mersenne Twister is not that different - it's also linear and operates essentially by modulo (but in a different field).

ax3l:
@psychocoderHPC found your other engine, the device engine that feeds of the entropy pool ;-)
https://en.cppreference.com/w/cpp/numeric/random/random_device

sbastrakov:
Probably not much, as the comment in the example given by your link suggests.

On a serious note, in case @ax3l is allergic to LCGs, maybe it is worth to implement a xorwow on CPU, at least as a temporary solution (and so it is the same generator as in curand)? Should be rather simple.

ax3l:
> not all LCGs are evil, however should be used with caution

[...]

Issue #390, opened 2017-08-30
different random number generators for a accelerator


psychocoderHPC:
Currently the generation method for random numbers for a accelerator is fixed defined within alpaka.
To provide different generators per accelerator depending of the users needs we should think about an interface change.

# Why we need different generators:
- each generator provide different qualities of random numbers
- the size of the stored random number state is different and influence the memory usage and performance

e.g. PIConGPU provides [different methods](https://github.com/ComputationalRadiationPhysics/picongpu/tree/c3054c36ef5cc482233b4779fc9d8dbaed8d1967/src/libPMacc/include/random/methods) up to my [pull request](https://github.com/ComputationalRadiationPhysics/picongpu/pull/2226) to use the native alpaka generator which removes the possibility for the user to control the quality of the RNG generator.

BenjaminW3:
In alpaka the generators are already seperated from the distribution.
So in theory it would be possible to use different generators. However, there is some work to do:
* 1. the CUDA backend only provides the Xor generator. XorMin, MRG32k3a and MRG32k3aMin could easily be added to [RandCuRand.hpp](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/rand/RandCuRand.hpp). Please create corresponding Pull Requests.
* 2. the CPU backends only provide a mersenne twister generator. (Not even as a standalone generator but only direclty within `alpaka::rand::generator::createDefault`) This generator should be made its own class. There are some other generators provided by the C++ standarad library which could be added to [RandStl.hpp](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/rand/RandStl.hpp).
* 3. I have no Idea how to implement Xor, XorMin, ... generators similar to the CUDA ones.
* 4. due to those differences between the available generators, there is no way to use the same generator on all backends. `alpaka::rand::generator::createDefault` simply uses an unspecified generator.
* 5. there are no unit tests for the random functions

Points 1, 2 and 5 can be solved, but point 4 depends on point 3 which may be very hard.

psychocoderHPC:
Thanks for the summary. Point 3 is one reason why I opened this pull request. I think it is not possible or to hard to maintain that we have all Generators on all platforms. 
I will use this issue also for thinking about solution how we can handle that each platform maybe ships different algorithm and never the less give the user the opportunity to write code without #if to support the differences between the platforms.

[...]

PR #576, opened 2018-08-02
Rng: add std::random_device gen

ax3l:
Add the `std::random_device` generator as an alternative CPU rng engine.

From cppreference.com:
> note: demo only: the performance of many
> implementations of random_device degrades sharply
> once the entropy pool is exhausted. For practical use
> random_device is generally only used to seed
> a PRNG such as mt19937

Note: we should also add a trait & factory besides `createDefault` for RNGs. I used a hacked `createDefault` for testing. (This is outside of this PR, but needed for users to use this.)

Related to #575

### DIFF
--- a/include/alpaka/rand/RandStl.hpp
+++ b/include/alpaka/rand/RandStl.hpp
@@ -68,6 +68,35 @@ namespace alpaka
                 public:
                     std::mt19937 m_State;
                 };
+
+                //#############################################################################
+                //! The standard library's random device based on the local entropy pool.
+                //!
+                //! Warning: the entropy pool on many devices degrates quickly and performance
+                //!          will drop significantly when this point occures.
+                class RandomDevice
+                {
+                public:
+
+                    //-----------------------------------------------------------------------------
+                    RandomDevice() = default;
+                    RandomDevice(RandomDevice&&) :
+                        m_State{}
+                    {
+                    }
+
+                    //-----------------------------------------------------------------------------
+                    ALPAKA_FN_ACC_NO_CUDA RandomDevice(
+                        std::uint32_t const &,
+                        std::uint32_t const & = 0,
+                        std::uint32_t const & = 0) :
+                        m_State{}
+                    {
+                    }
+
+                public:
+                    std::random_device m_State;
+                };
             }
         }
 


include/alpaka/rand/RandStl.hpp
```diff
@@ -68,6 +68,35 @@ namespace alpaka
                 public:
                     std::mt19937 m_State;
                 };
+
+                //#############################################################################
+                //! The STL random device based on the local entropy pool.
```
sbastrakov: In view of #577 , maybe not call it STL in the new code?

ax3l: uff yes, copy paste from above xD

include/alpaka/rand/RandStl.hpp
```diff
@@ -68,6 +68,35 @@ namespace alpaka
...
+                //!
+                //! Warning: the entropy pool on many devices degrates quickly and performance
+                //!          will drop significantly when this point occures.
+                class RandomDevice
+                {
+                public:
+
+                    //-----------------------------------------------------------------------------
+                    RandomDevice() = default;
+                    RandomDevice(RandomDevice&&) :
```
sbastrakov: Is a move constructor required for generators in alpaka? I guess so, since `std::random_device` is not movable and you provided this implementation. But then I think explicitly adding a `= default` move constructor to `MersenneTwister` and, potentially, others would make the code more clear.

ax3l:
During init in the factory below it is needed. otherwise it won't compile since random_device is not movable

include/alpaka/rand/RandStl.hpp
ax3l: During init in the factory below it is needed. otherwise it won't compile since random_device is not movable

psychocoderHPC:
@ax3l I think the way how to use you implementation is: 

```C++
auto rngGenerator = alpaka::rand::createDefault( alapka::rand::RandomDevice{}, seed, subsequence ) ;
```

ax3l:
`TRand` (templ arg) is an accelerator? uhhh, fun xD

BenjaminW3:
Yes, this might work but it is not the intended usage ;-)

ax3l:
> Yes, this might work but it is not the intended usage ;-)

can we add an "offical" api? :)

BenjaminW3:
Yes, see @psychocoderHPC proposal in #390. We would need to categorize the RNGs by some criteria.

Issue #577, opened 2018-08-02
Rng: CPU Namings


ax3l:
The CPU rng should not be called `RandStl` but `RandStdlib`.

STL is the standard template library, which is part of the stdlib. `<random>` is of course not part of the STL.

sbastrakov:
Agree.

BenjaminW3:
👍 

I am still unsure how #579 plays into this. It adds something outside of the StdLib. It may need its own new file.
Maybe it would be better to seperate between the devices supported: `cuda/RandCuRand`, `cpu/RandStdLib`, `cpu/RandTinyMt`?

PR #579, opened 2018-08-03
CPU Rng: TinyMT

### LINKED ISSUE #575: Rng: STL MT has huge state

ax3l:
Implement the TinyMersenneTwister from Mutsuo Saito @MSaito (Hiroshima University) and Makoto Matsumoto @m-mat (The University of Tokyo) as standard lib conformant generator.

Reference: https://github.com/MersenneTwister-Lab/TinyMT
Usage under: modified 3-clause BSD license
(Previous odifications applied: 783b8f7013b1961a903fc52e4ed56ed7c743b6a6)
Only modification: now a single header, even the two init implementations: 16b17b72efb9eed82e56b313ea9855a61d21dc6d

We could also implement the `dispose` with the jump method, but I don't understand it's `lower_step`/`upper_step` params: [docs](http://www.math.sci.hiroshima-u.ac.jp/~m-mat/MT/TINYMT/JUMP/index.html) [src](https://github.com/MersenneTwister-Lab/TinyMT/blob/master/jump/jump32.h)

Use this as the new default on CPUs.
Test all CPU rngs in unit tests.

Does not yet rename, in order to make this patch easier back-portable #577

Close #575


include/alpaka/rand/RandStl.hpp
```diff
@@ -69,15 +90,87 @@ namespace alpaka
...
+                //! "Tiny" state mersenne twister implementation
+                //!
+                //! repository: github.com/MersenneTwister-Lab/TinyMT
+                //!
+                //! @license 3-clause BSD
+                //!
+                //! @author Mutsuo Saito (Hiroshima University)Tokio University.
+                //! @author Makoto Matsumoto (The University of Tokyo)
+                //!
+                //! size of state: 28 bytes (127 bits?!)
```
ax3l: I don't understand where [this](https://github.com/MersenneTwister-Lab/TinyMT/blob/master/tinymt/tinymt32.h#L33-L41) is 127 bits. At least for us its 28 bytes (228 bits) as we use it.

test/unit/rand/src/RandTest.cpp
```diff
@@ -75,6 +67,7 @@ class RandTestKernel
 #if !BOOST_ARCH_PTX
             BOOST_VERIFY(std::isfinite(r));
 #endif
+            boost::ignore_unused(r);
```
BenjaminW3: I do not think that this works here. `ignore_unused` was missing the `__device__` attribute. This leads to errors. This is the reason I added the compiler pragmas. Let's see if this works now.

include/alpaka/rand/TinyMT/LICENSE.txt
```diff
@@ -0,0 +1,30 @@
+Copyright (c) 2011, 2013 Mutsuo Saito, Makoto Matsumoto,
```
BenjaminW3: This license seems to be compatible with ours.
@ComputationalRadiationPhysics/alpaka-maintainers Does anyone see any problems.

include/alpaka/rand/RandStl.hpp
```diff
@@ -69,15 +90,91 @@ namespace alpaka
                     std::mt19937 m_State;
                 };
 
+                //#############################################################################
+                //! "Tiny" state mersenne twister implementation
+                //!
+                //! repository: github.com/MersenneTwister-Lab/TinyMT
+                //!
+                //! @license 3-clause BSD
```
BenjaminW3: clang complains: `RandStl.hpp:98:21: error: unknown command tag name [-Werror,-Wdocumentation-unknown-command]`
We have to be compatible with the strict subset of documentation that clang understands...

include/alpaka/rand/RandStl.hpp
```diff
@@ -29,25 +30,45 @@
 
 #include <random>
 #include <type_traits>
+#include <stdint.h>
```
BenjaminW3: Please use `<cstdint>` instead.

test/unit/rand/src/RandTest.cpp
BenjaminW3: I already thought about adding my own `ignore_unused` into alpaka. However with c++ 17 we get `maybe_unused` as variable attribute which is by far superior.

include/alpaka/rand/TinyMT/LICENSE.txt
ax3l: Yep, it's "new", "revised" or ["modified" 3-clause BSD](https://en.wikipedia.org/wiki/BSD_licenses#3-clause_license_(%22BSD_License_2.0%22,_%22Revised_BSD_License%22,_%22New_BSD_License%22,_or_%22Modified_BSD_License%22)) which is GPL compatible. We have to ship the license (added) and state it, maybe more centrally accessible, was well.

test/unit/rand/src/RandTest.cpp
ax3l: I also though it's just designed for params & templates, but it works for me, tested on gcc.

ax3l:
> Does anything speak against using this in CUDA as well?

It's not a matrix generator and afaik, CUDA uses their (tiny) state implementation already in cuRand.

include/alpaka/rand/RandStl.hpp
```diff
@@ -35,19 +56,38 @@ namespace alpaka
     namespace rand
     {
         //#############################################################################
-        //! The standard library rand implementation.
-        class RandStl
+        //! "Tiny" state mersenne twister implementation
+        class TinyMersenneTwister
```
BenjaminW3: I do not think that you need those intermediate classes. you most probably could directly use `rand::generator::cpu::RandomDevice{}` in the unit test. This would make it more prominent that they are limited to CPU.
The rest would need a general overhaul either way but I do not think that introducing those classes brings us anything at the moment.

ax3l: Nope, did not work since we check for the base class via an enable_if currently

include/alpaka/rand/TinyMT/tinymt32.h
```diff
@@ -0,0 +1,393 @@
...
+        ^ random->status[2];
+    x ^= (x << TINYMT32_SH0);
+    y ^= (y >> TINYMT32_SH0) ^ x;
+    random->status[0] = random->status[1];
+    random->status[1] = random->status[2];
+    random->status[2] = x ^ (y << TINYMT32_SH1);
+    random->status[3] = y;
+    //random->status[1] ^= -((int32_t)(y & 1)) & random->mat1;
+    //random->status[2] ^= -((int32_t)(y & 1)) & random->mat2;
+    int32_t const a = -((int32_t)(y & 1)) & (int32_t)random->mat1;
```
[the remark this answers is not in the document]
ax3l: Outsourced the C++ uniform engine wrapper already, added the diagnostics and if it compiles will remove the modifications from the original source

include/alpaka/rand/RandStl.hpp
BenjaminW3: We only check for `RandBase` with the `enable_if` and cast the object to `RandBase` if it is available. This is only the special case for a class that inherits from a random generator implementation (e.g. an accelerator). If there is no `RandBase`, we directly use the class itself without any casting. So at least conceptionally it should work because it is what I intended.

ax3l: Yep, I think it's broken. I tried this first and did not manage that way.

BenjaminW3: Ah, I see your point. You are typedefing `using RandStl = TinyMersenneTwister;`. Then you need it because `RandStl` is used by some accelerators.

BenjaminW3: But without this additional change of the default, it should work.

BenjaminW3: I do not think that you should specialize the `createDefault`method. We should add new methods like `createTinyMersenneTwister()`. You should not change the `RandStl` defiinition. You should change the implementation of the `createDefault` specialization for `RandStl` to not use the std mersenne twister but your one.

ax3l: Maybe I miss the point, but the previous `std::mersenne_twister` did not run on CUDA as well.

ax3l: And the implemented `cuRand` mersenne twister is actually the "tiny" MT that I am implementing right now. so we are closer to being "similar" than we were before ;D

ax3l: > The direct calls you did in the unit test are not what a user should be doing.

Exactly! In the unit tests I do:
- call the default for all ACCs
- call all other RNGs where supported, just for coverage

BenjaminW3: Yes, but this is no problem because `createDefault` simply returned something different given a CUDA accelerator.
Changing the implementation for `createDefault` on CPU is perfectly valid.

BenjaminW3: However, to do this, you do not need those intermediate classes and you should not replace the definition of `RandStl`.

BenjaminW3: What I want is to have 

```
auto genMersenneTwister = alpaka::rand::generator::createMersenneTwister(
            acc,
            12345u,
            6789u
        );
```

instead of 

```
auto genMersenneTwister = alpaka::rand::generator::createDefault(
            alpaka::rand::MersenneTwister{},
            12345u,
            6789u
        );
```

This `createMersenneTwister` should then be also implemented for CUDA (by simply using the cuRand version that is returned by `createDefault` or something better matching if available).

BenjaminW3: Or to have

```
enum class Generator{
Default,
MersenneTwister,
...
};
auto genMersenneTwister = alpaka::rand::generator::create<alpaka::rand::Generator::MersenneTwister>(
            acc,
            12345u,
            6789u
        );
```

ax3l: That's ok and something for #390 

I try not to mix those aspects, since I want to backport this "fix".
 
We can address this re-structuring from `createDefault(acc`, ...)` toward `create< Generator >(acc, ...)` in a follow-up I think.

test/unit/rand/src/RandTest.cpp
ax3l: ah year, we need a cuda-instrumented one for host - device qualifiers.

BenjaminW3: CI proved me correct:
```
alpaka/test/unit/rand/src/RandTest.cpp:71:20: error: reference to __host__ function 'ignore_unused<float>' in __host__ __device__ function
            boost::ignore_unused(r);
                   ^
```
You should simply revert those changes and keep the pragmas.

ax3l: added `alpaka::ignore_unused` in #582 :)

include/alpaka/rand/TinyMT/tinymt32.h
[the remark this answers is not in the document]
ax3l: updated and changed removed! :)

BenjaminW3:
gcc 5 seems to have a problem with your code:
```
In file included from /home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/rand/TinyMT/Engine.hpp:38:0,
                 from /home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/rand/RandStl.hpp:25,
                 from /home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/acc/AccCpuSerial.hpp:37,
                 from /home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/alpaka.hpp:33,
                 from /home/travis/build/ComputationalRadiationPhysics/alpaka/test/unit/rand/src/RandTest.cpp:30:
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/rand/TinyMT/tinymt32.h: In member function 'void rand_::RandTestKernel::operator()(const TAcc&) const [with TAcc = alpaka::acc::AccCpuOmp4<std::integral_constant<long unsigned int, 4ul>, short unsigned int>]':
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/rand/TinyMT/tinymt32.h:294:33: error: 'genDefault.TINYMT32_T::mat1' is used uninitialized in this function [-Werror=uninitialized]
     random->status[1] = random->mat1;
                                 ^
```
You may want to suppress this warning for this compiler and try if it still works.

ax3l:
creepy!

ax3l:
Seems there is indeed a state to set, will add.
https://github.com/jj1bdx/TinyMT/blob/master/jump/sample.c

ax3l:
@BenjaminW3 Houston, we finally have green light for liftoff! The sky is clear, rockets are fueled and thrusters are eager to burn!

include/alpaka/rand/RandStl.hpp
```diff
@@ -68,15 +89,47 @@ namespace alpaka
...
+                //!
+                //! @author Mutsuo Saito (Hiroshima University)Tokio University.
+                //! @author Makoto Matsumoto (The University of Tokyo)
+                //!
+                //! size of state: 28 bytes (127 bits?!)
+                class TinyMersenneTwister
+                {
+                public:
+                    //-----------------------------------------------------------------------------
+                    TinyMersenneTwister() = default;
```
ax3l: hm, I think those need to be `delete`d instead

include/alpaka/rand/TinyMT/Engine.hpp
```diff
@@ -0,0 +1,109 @@
...
+        {
+            // parameters from TinyMT/jump/sample.c
+            prng.mat1 = 0x8f7011ee;
+            prng.mat2 = 0xfc78ff1f;
+            prng.tmat = 0x3793fdff;
+
+            seed( seedValue );
+        }
+
+        TinyMTengine() = delete;
```
ax3l: uff, should not have deleted this one xD

ax3l: Fixed in #589

PR #580, opened 2018-08-05
Add support for separable compilation

### LINKED ISSUE #561: Add support for CUDA_SEPARABLE_COMPILATION
I was trying to make a function call to 'MathFunctions' from Alpaka. Here is the structure of my workspace:
/src
             -   sqrt.cpp  // where I modified from vectorAdd example to call ‘mysqrt’ function                
             -   mysqrt.cpp  
/MathFunction
	     -   MathFunction.h // mysqrt is defined here

The program simply pass the value 4 and 9 to calculate sqrt and add them up.
The approach seems to  work only for CPU, when I compile for GPU the 'cmake' process got through but the 'make' gave me this errors:
____________________________________________________
herbst66@hypnos4:~/HOME/alpaka/alpaka-develop/example/sqrt_GPU/build$ make
[ 33%] Building NVCC (Device) object CMakeFiles/sqrt.dir/src/sqrt_generated_sqrt.cpp.o
ptxas fatal  : Unresolved extern function '_Z6mysqrtd'
CMake Error at sqrt_generated_sqrt.cpp.o.cmake:282 (message):
  Error generating file
  /home/herbst66/HOME/alpaka/alpaka-develop/example/sqrt_GPU/build/CMakeFiles/sqrt.dir/src/./sqrt_generated_sqrt.cpp.o
make[2]: *** [CMakeFiles/sqrt.dir/src/sqrt_generated_sqrt.cpp.o] Error 1
make[1]: ***
[CMakeFiles/sqrt.dir/all] Error 2
make: *** [all] Error 2
____________________________________________________

Does anyone have any suggestion to fix it?
PS. I enclosed the workspace as a .zip file so you can have a look.

[sqrt.zip](https://github.com/ComputationalRadiationPhysics/alpaka/files/2163080/sqrt.zip)

BenjaminW3:
Fixes #561


BenjaminW3:
I have not yet been able to get it compile correctly. As has been described in #561 there seems to be a problem with `nvcc fatal : Unknown option '-generate-code arch'`.
Any help would be appreciated.

BenjaminW3:
Currently I am running into this error: https://gitlab.kitware.com/cmake/cmake/merge_requests/1008.
I do not know why because I do not use per-config generator expressions but it is fixed in CMake 3.12 and most probably starting with CMake 3.10.

BenjaminW3:
In Visual Studio it does work now with CMake 3.12 and the when leaving out the `-generate-code` options.
However, it only works when doing a clean rebuild. Doing partial builds fails with linker errors.

test/integ/separableCompilation/CMakeLists.txt
```diff
@@ -0,0 +1,40 @@
+#
+# Copyright 2018-2019 Benjamin Worpitz
+#
+# This file is part of alpaka.
+#
+# alpaka is free software: you can redistribute it and/or modify
```
ax3l: With regards to #637, this new file's header will need an update :)

test/integ/separableCompilation/include/mysqrt.hpp
```diff
@@ -0,0 +1,24 @@
+/**
+ * \file
+ * Copyright 2018 Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
```
ax3l: With regards to #637, this new file's header will need an update :)

test/integ/separableCompilation/src/main.cpp
```diff
@@ -0,0 +1,207 @@
+/**
+ * \file
+ * Copyright 2018 Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
```
ax3l: With regards to #637, this new file's header will need an update :)

test/integ/separableCompilation/src/mysqrt.cpp
```diff
@@ -0,0 +1,42 @@
+/**
+ * \file
+ * Copyright 2018-2019 Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
```
ax3l: With regards to #637, this new file's header will need an update :)

tdd11235813:
bump: we need to check this, if we can get this to work for at least a certain CUDA version on linux. Also a temporary workaround guide (e.g. modifying gencode options) would be ok.

BenjaminW3:
I do not think that I have the setup and time to find a solution which does not break the gencode options. I would be happy if someone else could work on this.

tdd11235813:
@BenjaminW3 this is absolutely ok, I already talked with @psychocoderHPC about this, so we try to test a few things and see how far we come

tdd11235813:
Ok, the issue was maybe just improper strings in cmake list `CUDA_NVCC_FLAGS`.
Recap:
```cmake
set(var a b c d e)
# is not the same as
set(var "a b c d e")
# and for lists like CUDA_NVCC_FLAGS use:
list(APPEND CUDA_NVCC_FLAGS -flag1 -flag2)

# if you write a single flag it might be ok to use "-flag" with quotes,
# but when a whitespace is detected between two flags the thing might become escaped
LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored")
# ... will escape the two flags with quotes depending how the list is used
# (curiously, quotes are not there in the nvcc compile steps)
# ... you could write a ';' instead of the whitespace to match the list character

# however this way without quotes is always fine:
LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored)
```
https://cmake.org/cmake/help/v3.12/command/list.html

Need to test around to ensure the fix, and I hope to have time tomorrow to provide a PR to your branch.

Edit: 
- Note: the nvcc linker has been confused about the quotes and when quotes are removed, linker is running fine
- test passed only on CUDA backend
  - CPU backend fails even when no CUDA is enabled, needs checking

tdd11235813:
@BenjaminW3 You can find the current changes [here](https://github.com/tdd11235813/alpaka/commit/acd3fcdc5a0cdd773bc1b8b182a7b58aecefe0e7). I have not opened a PR on your repo for now due to outdated branch. I also could open a new PR here on the alpaka repo based on your commit, if you wish.

BenjaminW3:
Thank you very much! I will try to update my PR tomorrow evening.

BenjaminW3:
I rebased the PR and applied your changes but it still seems to fail somehow.

alpakaConfig.cmake
```diff
@@ -587,7 +589,7 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 FOREACH(_CUDA_ARCH_ELEM ${ALPAKA_CUDA_ARCH})
                     # set flags to create device code for the given architecture
-                    SET(CUDA_NVCC_FLAGS ${CUDA_NVCC_FLAGS}
+                    LIST(APPEND CUDA_NVCC_FLAGS
                         "--generate-code arch=compute_${_CUDA_ARCH_ELEM},code=sm_${_CUDA_ARCH_ELEM} --generate-code arch=compute_${_CUDA_ARCH_ELEM},code=compute_${_CUDA_ARCH_ELEM}")
```
tdd11235813: remove the quotes here and put the two flags on two lines

alpakaConfig.cmake
```diff
@@ -628,7 +630,7 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 ENDIF()
                 # avoids warnings on host-device signatured, default constructors/destructors
                 IF(CUDA_VERSION GREATER_EQUAL 9.0)
-                    LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored")
+                    LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe" "--diag_suppress=esa_on_defaulted_function_ignored")
```
tdd11235813: should work because of the single flags, but maybe without quotes are less error-prone in the future (it is intriguing to just add another flag and then again we have the issue. This cannot happen when you do not use quotes for cmake list argument.

BenjaminW3:
Is separable compilation expected to work with clang as native CUDA compiler?
[This thread](http://lists.llvm.org/pipermail/llvm-dev/2017-August/116561.html) says that it was not supported in 2017 and if nobody added it it is not supported yet.

tdd11235813:
at least some CUDA compiler version mixes seem to work. The requirements are gcc+CUDA (like gcc7+CUDA9.2). Curiously, nvcc9.1+gcc4.9 fails with:

> nvcc fatal   : Don't know what to do with 'CUDA_cublas_device_LIBRARY-NOTFOUND'

Other than that it looks not too bad, but CPU side needs bugfixing.
We then could add some cmake checks for when separable compilation feature can be enabled.
A user is currently testing the feature on his application code as well.

BenjaminW3:
I fixed the following things now:
* the error in the test (missing wait)
* disabled the test for clang as CUDA compiler. For nvcc separable compilation is now enabled by default
* raised the CMake version from 3.11.0 to 3.11.4. This was the reason of the `nvcc fatal : Don't know what to do with 'CUDA_cublas_device_LIBRARY-NOTFOUND'` error.

nvcc+MSVC still does not work and fails with some boost mpl compilation errors. I will investigate this. It looks similar to #782. I will try to remove the dependency to `boost/math/special_functions/relative_difference.hpp` because it transitively includes mpl

BenjaminW3:
Do we want to enable `CUDA_SEPARABLE_COMPILATION` for nvcc by default or do we want to hide it behind an alpaka option `ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION`?

tdd11235813:
I would keep it as an option for now that can be enabled when needed. `ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION` is good. HIP(NVCC) does not seem to support this right away (hipify script from 2017 marks this as unsupported, docs are not refering to it at all).

psychocoderHPC:
I would also say we should have it currently as an optional flag.

BenjaminW3:
Ready for review/merge!

README.md
```diff
@@ -94,6 +94,7 @@ When *Boost.Fiber* is enabled and alpaka is build in C++17 mode with clang and l
 When an accelerator back-end using *CUDA* is enabled, version *8.0* of the *CUDA SDK* is the minimum requirement.
 *NOTE*: When using nvcc as *CUDA* compiler, the *CUDA accelerator back-end* can not be enabled together with the *Boost.Fiber accelerator back-end* due to bugs in the nvcc compiler.
```
ax3l: Note on the `*Note*:` lines ;-)

If you add *two spaces* at the end of a line in markdown, this will cause a line-break for each note but not a new paragraph :) This might make this more readable here.

BenjaminW3: Thanks for the tip. Looks much better now.

BenjaminW3:
I have disabled it by default for now. We can always enable it later.

ax3l:
> I have disabled it by default for now. We can always enable it later.

Imho fine to keep it on as well. Either way, you must enable it in your test case now, otherwise the case is just skipped in CI.

BenjaminW3:
You are right, I added it to some jobs to test that it works with and without the option.

README.md
```diff
@@ -91,9 +91,10 @@ However some of the accelerator back-end implementations require different boost
...
-*NOTE*: When using nvcc as *CUDA* compiler, the *CUDA accelerator back-end* can not be enabled together with the *Boost.Fiber accelerator back-end* due to bugs in the nvcc compiler.
-*NOTE*: When using clang as a native *CUDA* compiler, the *CUDA accelerator back-end* can not be enabled together with any *OpenMP accelerator back-end* because this combination is currently unsupported.
+When an accelerator back-end using *CUDA* is enabled, version *8.0* of the *CUDA SDK* is the minimum requirement.  
+*NOTE*: When using nvcc as *CUDA* compiler, the *CUDA accelerator back-end* can not be enabled together with the *Boost.Fiber accelerator back-end* due to bugs in the nvcc compiler.  
+*NOTE*: When using clang as a native *CUDA* compiler, the *CUDA accelerator back-end* can not be enabled together with any *OpenMP accelerator back-end* because this combination is currently unsupported.  
+*NOTE*: Separable compilation is only supported when using nvcc, not with clang as native *CUDA* compiler. It is disabled by default and can be enabled via the CMake flag `ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION`.
```
ax3l, line 97: Add LLVM mailing list link for reference here?

BenjaminW3: I do not think that this link is really a useful resource. It may not be the most up to date and complete information.

ax3l: Yep, but then it's transparent what we base our statement on :)
(Aka: it's not an Alpaka limitation)

BenjaminW3:
Now it should be ready to be merged.

PR #582, opened 2018-08-05
alpaka::ignore_unused

ax3l:
Add a device-side compatible `ignore_unused` function.
Trivial implementation as in `boost/core/ignore_unused.hpp`

0.3.3 backport note: only first commit for #579

### DIFF
--- a/include/alpaka/core/Assert.hpp
+++ b/include/alpaka/core/Assert.hpp
@@ -24,7 +24,7 @@
 #include <alpaka/core/Common.hpp>
 
 #if !BOOST_ARCH_PTX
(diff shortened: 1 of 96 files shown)


include/alpaka/core/Assert.hpp
```diff
@@ -88,7 +88,7 @@ namespace alpaka
                 -> void
                 {
 #if !BOOST_ARCH_PTX
```
BenjaminW3: It should be possible to remove those ifdefs now.

BenjaminW3: I would try this in a separate PR.

PR #583, opened 2018-08-06
Clean up alpaka::ignore_unused usages

BenjaminW3:


ax3l:
oh no, we pushed the same cleanup ;)
#584

ax3l:
We can merge the first and then you rebase this one? You catched a few more includes as it looks and this might work well.

BenjaminW3:
@ax3l I have rebased my changes.

test/unit/rand/src/RandTest.cpp
```diff
@@ -67,17 +66,19 @@ class RandTestKernel
             auto const r = dist(gen);
 #if !BOOST_ARCH_PTX
```
ax3l: probably `#if BOOST_ARCH_PTX #else` would be easier to read?

test/unit/rand/src/RandTest.cpp
```diff
@@ -67,17 +66,19 @@ class RandTestKernel
...
-#endif
+#else
             alpaka::ignore_unused(r);
+#endif
         }
 
         {
             auto dist(alpaka::rand::distribution::createNormalReal<double>(acc));
             auto const r = dist(gen);
 #if !BOOST_ARCH_PTX
```
ax3l: probably `#if BOOST_ARCH_PTX #else` would be easier to read?

PR #584, opened 2018-08-06
ignore_unused in Device

ax3l:
- Use `ignore_unused` in device functions.
- regroup `alpaka/core/` includes
- Add `alpaka/core/Unused.hpp` to `alpaka/alpaka.hpp` include.

Follow-up to #582

### DIFF
--- a/include/alpaka/idx/bt/IdxBtCudaBuiltIn.hpp
+++ b/include/alpaka/idx/bt/IdxBtCudaBuiltIn.hpp
@@ -34,8 +34,7 @@
 #include <alpaka/vec/Vec.hpp>
 #include <alpaka/core/Cuda.hpp>
 #include <alpaka/core/Positioning.hpp>
-
-//#include <alpaka/core/Unused.hpp>
+#include <alpaka/core/Unused.hpp>
 
 namespace alpaka
 {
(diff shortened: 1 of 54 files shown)


include/alpaka/idx/bt/IdxBtCudaBuiltIn.hpp
```diff
@@ -104,11 +103,11 @@ namespace alpaka
                 template<
                     typename TWorkDiv>
                 ALPAKA_FN_ACC_CUDA_ONLY static auto getIdx(
-                    idx::bt::IdxBtCudaBuiltIn<TDim, TIdx> const & /*idx*/,
+                    idx::bt::IdxBtCudaBuiltIn<TDim, TIdx> const & idx,
                     TWorkDiv const &)
                 -> vec::Vec<TDim, TIdx>
                 {
-                    //alpaka::ignore_unused(idx);
+                    alpaka::ignore_unused(idx);
```
sbastrakov: Why is this solved with `ignore_unused` and not the standard approach with an unnamed parameter (which is now `idx`)?

BenjaminW3: Because the name is information that would else be lost especially in a header inly library. Furthermore this requires ifdefs in some cases.

BenjaminW3: Furthermore, you need a name to add a doxygen description.

sbastrakov: @BenjaminW3 I see. However, to be fair it is not because the library is header-only, but only because the function is defined right at declaration. Otherwise, the declaration will have a parameter and implementation won't, which is logical for both. (Also I understand the value of ```ignore_unused``` when smth is conditionally used, which is not the case here).

ax3l: yes, you are both right.

ax3l:
@BenjaminW3 CI is finally done :)

Issue #572, opened 2018-07-25
CUDA 9.2-10.1 warning calling host function


psychocoderHPC:
During the compile of PIConGPU with [alpaka 0.3.2](https://github.com/ComputationalRadiationPhysics/alpaka/blob/0.3.2/include/alpaka/vec/Vec.hpp#L909-L928) and CUDA 9.2.88 on taurus (TU Dresden) I got the following warning.

```
alpaka/include/alpaka/vec/Vec.hpp(925): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu") is not allowed

alpaka/include/alpaka/vec/Vec.hpp(925): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu [subobject]") is not allowed
```
# modules loaded

```
Currently Loaded Modules:
  1) modenv/classic   (S)   4) cmake/3.10.1                 7) openmpi/2.1.1-gnu5.3-cuda
  2) gcc/5.3.0              5) hdf5/1.8.18-gcc-5.3.0-xmpi   8) cuda/9.2.88
  3) bullxmpi/1.2.8.4       6) zlib/1.2.8
```

BenjaminW3:
I remember having seen those warnings in the alpaka CI as well. They should not be a problem because I hope no one instantiates a CPU device within a CUDA kernel. They do not seem to result in an error even with Werror enabled.
They might be solved by fixing the attributes or ifdef-ing some code.

BenjaminW3:
I just had the look at the definition of the `alpaka::dev::DevCpu` constructor. It does not have any attributes so it should implicitly be `__host__` only. So I do not see a reason for this warning.

ax3l:
I am also seeing this with nvcc 9.2.88 and gcc 7.1

ax3l:
The default and move default contructors of `vec::Vec` lack the `ALPAKA_NO_HOST_ACC_WARNING` qualifier

ax3l:
Qualifier does not work. Using instead `ALPAKA_FN_HOST_ACC` on the defaulted members leads to warnings of the form:
```
 __device__ annotation is ignored on a function("Vec") that is explicitly defaulted on its first declaration

[...]

PR #474, opened 2018-04-06
remove support for CUDA 7.x

BenjaminW3:
This adds unconditional support for device lambdas (7.5) and heterogeneous lambdas (8.0).
Furthermore, this allows to move the `EventHostManualTrigger` (8.0) out of the test code to make it publicly available in a follow up Pull Request.

PR #588, opened 2018-08-07
Backport to 0.3.3: CPU RNGs, ignore_unused

ax3l:
Backports to the [0.3.3](https://github.com/ComputationalRadiationPhysics/alpaka/projects/3) release branch:
- RNG: add `std::random_device` #576
- first commit from `alpaka::ignore_unused` #582
- RNG: add TinyMT for CPU #579
- RNG: TinyMT Engine default seed #589
- remove leftover GCC+NVCC check if already in CMake #590

`version.hpp`: update to `0.3.3`

## Runtime tests

- [x] PIConGPU LWFA example issue https://github.com/ComputationalRadiationPhysics/picongpu/issues/2653
- [x] PIConGPU KHI example


test/unit/rand/src/RandTest.cpp
```diff
@@ -47,34 +48,27 @@ BOOST_AUTO_TEST_SUITE(rand_)
...
-        auto gen(alpaka::rand::generator::createDefault(acc, 12345u, 6789u));
-
-#if BOOST_COMP_CLANG
-    #pragma clang diagnostic push
-    #pragma clang diagnostic ignored "-Wunused-variable"
-#endif
 // gcc 5.4 in combination with nvcc 8.0 fails to compile the CPU STL distributions when --expt-relaxed-constexpr is enabled
 // /usr/include/c++/5/cmath(362): error: calling a __host__ function("__builtin_logl") from a __device__ function("std::log") is not allowed
-#if !((BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(5, 0, 0)) && (BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(8, 0, 0)) && (BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(9, 0, 0)))
+#if !((BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(5, 0, 0)) && (BOOST_COMP_NVCC == BOOST_VERSION_NUMBER(8, 0, 0)))
```
ax3l: @BenjaminW3 this was removed with #474

Since this branch still support CUDA 7.X, should we keep this? Or is this a legacy NVCC 9 change only?

BenjaminW3: I am not sure. As long as the CI is happy I do not have a problem to change a test on the old branch.
You made the check more strict so that it should not apply anymore because there is not exact nvcc 8.0.0. Previously all 8.*.* have been filtered out.

ax3l: hm, actually there are two CUDA 8 point releases. Not sure if even in `dev` a `8.0.0` would be catched...
https://gist.github.com/ax3l/9489132

ax3l: Will change it to the same as in #590.

ax3l: Ok, looks to me that the 0.3.X branch has the usual CMake checks from #566 already in place:
https://github.com/ComputationalRadiationPhysics/alpaka/blob/0.3.2/alpakaConfig.cmake#L382-L414

So we can apply #590 and remove the pre-compiler guards here as well

ax3l:
@BenjaminW3 if the CI passes, feel free to merge the `release-0.3.0` branch to `master` and tag it there as a new release.

I currently only have #572 but without a solution on the list. In case we can't fix it now, I move it to a 0.3.4 todo.

ax3l:
Ok, then let's merge this.
I then open the PR from `release-0.3.0` to `master` and we see if the CI is up to speed by then.

PR #590, opened 2018-08-07
NVCC 8.0 Exact Compare

ax3l:
Exclusions for NVCC 8.0 are too exact.
There is no zero-patch-level release, but two 8.0.X releases.

Update: checked in CMake now, removed.

This is the GCC 5.4+ "std::tuple" with CUDA 8.0 bug work-around. Since we check it in CMake, we could also remove it here.
https://gist.github.com/ax3l/9489132

Introduced in #474

### DIFF
--- a/test/unit/queue/src/QueueTest.cpp
+++ b/test/unit/queue/src/QueueTest.cpp
@@ -59,9 +59,6 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(
     BOOST_REQUIRE_EQUAL(true, alpaka::queue::empty(f.m_queue));
 }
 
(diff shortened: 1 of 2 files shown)


ax3l:
In the first place, I think this excluded the whole test for GCC 5.4 and CUDA 8.0 before CUDA 9.0 was released (which works with GCC 5.5 and GCC 6).

BenjaminW3:
Ah, we do not even allow to compile this at all. Then I would like to remove This completely.

ax3l:
We now check in CMake if the GCC-NVCC combination is known to work. Therefore we don't try this combination in CI and throw in configure state if a user would use this combination.

PR #594, opened 2018-08-08
Cleaner: AppVeyor Dot File

ax3l:
AppVeyor can also be set as dot file. Looks cleaner in the repo.

AppVeyor issue: [400]
> {"message":"Push request and project repositories do not match."}


ax3l:
@BenjaminW3 can you log into AppVeyor an re-enable the repo please? Looks like it probably has an outdated link, maybe because we renamed in the past or so...

BenjaminW3:
I can not select repos within the ComputationalRadiationPhysics organization in my appveyor account. Only my personal forks can be seen. Maybe I do not have the corresponding access rights anymore?

ax3l:
Weird, you should have... Will check if I can add it.

ax3l:
@BenjaminW3 ah no, GitHub access for apps changed. Granted access now to `ComputationalRadiationPhysics/`, so you should be able to see this in your AppVeyor projects now.

BenjaminW3:
Ah there it is again. I have triggered an initial build for develop. It could be that some things broke in the last weaks but a red Appveyor build does not prevent merging.

BenjaminW3:
You will have to push this branch again if you want to get a build on Appveyor.

PR #598, opened 2018-08-08
Fix Visual Studio build

BenjaminW3:
Should this be added to the 0.3.3 release?

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -623,6 +623,7 @@ LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "${_ALPAKA_LINK_LIBRARY}")
 
 # Add all the source and include files in all recursive subdirectories and group them accordingly.
 append_recursive_files_add_to_src_group("${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "hpp" _ALPAKA_FILES_HEADER)
+append_recursive_files_add_to_src_group("${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "h" _ALPAKA_FILES_HEADER)
 
 append_recursive_files_add_to_src_group("${_ALPAKA_ROOT_DIR}/script" "${_ALPAKA_ROOT_DIR}" "sh" _ALPAKA_FILES_SCRIPT)
 SET_SOURCE_FILES_PROPERTIES(${_ALPAKA_FILES_SCRIPT} PROPERTIES HEADER_FILE_ONLY TRUE)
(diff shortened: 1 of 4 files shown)


ax3l:
Yes, please add to the release branch for 0.3.3

ax3l:
@BenjaminW3 there seem to be some other AppVeyor issues. Are these relevant to get 0.3.3 released (aka: is something else broken for MSVC?)

include/alpaka/rand/TinyMT/Engine.hpp
```diff
@@ -28,21 +28,26 @@
 #if BOOST_COMP_CLANG
 #   pragma clang diagnostic push
 #   pragma clang diagnostic ignored "-Wold-style-cast"
-#   pragma clang diagnostic push
```
ax3l: that does not work. now it throws again in clang

ax3l:
@BenjaminW3 I will push to this branch to fix

ax3l:
@BenjaminW3 also had to rebase, sorry! Please `git reset --hard HEAD~` your commit locally before fetching my updates.

ax3l:
> still errors for clang, pop each pragma

I reverted your clang change, did not work as planned:

https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/413729129

BenjaminW3:
Thank you for fixing!

alpakaConfig.cmake
```diff
@@ -623,6 +623,7 @@ LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "${_ALPAKA_LINK_LIBRARY}")
 
 # Add all the source and include files in all recursive subdirectories and group them accordingly.
 append_recursive_files_add_to_src_group("${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "hpp" _ALPAKA_FILES_HEADER)
+append_recursive_files_add_to_src_group("${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "h" _ALPAKA_FILES_HEADER)
```
ax3l: if we do this, we probably have to move the warning suppression into the `.h` header itself...

ax3l: We should make Alpaka an installable library now, so we can compile this to an internal object and making the header internal only ;-)

BenjaminW3: I do not understand what adding this file to the IDE project has to do with the warnings and the installation. This line should not have any influence on compilation.

BenjaminW3: Ah, now I see. The analysis build with the self-contained header check fails. You can remove the line and I will see if I can exclude it from this check later.

ax3l:
> if we do this, we probably have to move the warning suppression into the .h header itself...

updates as described

alpakaConfig.cmake
ax3l: Seen to late. Do you think the state with moving the suppression into the impl is fine as well?

ax3l: Btw: maybe I misinterpreted the first fail on clang and that was the same reason

BenjaminW3: I liked the first version with unchanged header more. But both is good. You could try the clang suppression again.

ax3l: Me too, but more robust that way now.
Updated regarding clang now as you initially suggested.

PR #602, opened 2018-08-10
Enable peer memory copies

### LINKED ISSUE #596: CUDA Enable peer access for real device-to-device copy.
At the moment real P2P copy is not working, as the memP2P test (PR #586) shows.
It silently does copy over host. A `cudaDeviceEnablePeerAccess` is necessary to enable P2P.
This is for platforms with unified virtual addressing (UVA, requires 64bit and afaik Kepler+ while Fermi required TCC). I have not tested the non-UVA case.
If we could rely on UVA generally, no special peer copy branches in alpaka would be needed, as cudaMemcpy already does this by UVA.
For examples see [this post](https://stackoverflow.com/a/22699912).

Without `cudaDeviceEnablePeerAccess` (does `CUDA memcpy DtoH` + `CUDA memcpy HtoD`):

<details>

```
nvprof --print-gpu-trace test/unit/mem/p2p/memP2P
Running 30 test cases...
==19708== NVPROF is profiling process 19708, command: test/unit/mem/p2p/memP2P

*** No errors detected
==19708== Profiling application: test/unit/mem/p2p/memP2P
==19708== Profiling result:
   Start  Duration            Grid Size      Block Size     Regs*    SSMem*    DSMem*      Size  Throughput  SrcMemType  DstMemType           Device   Context    Stream  Name
590.45ms  3.7760us                    -               -         -         -         -      160B  40.410MB/s      Device           -    Tesla K80 (0)         1        26  [CUDA memset]
591.54ms  3.2960us                    -               -         -         -         -      160B  46.295MB/s      Device      Device    Tesla K80 (0)         1        26  [CUDA memcpy DtoH]
591.55ms  2.7840us                    -               -         -         -         -      160B  54.809MB/s      Device      Device    Tesla K80 (1)         2        19  [CUDA memcpy HtoD]
592.57ms  50.335us              (1 1 1)         (1 1 1)        32        0B        0B         -           -           -           -    Tesla K80 (1)         2        27  _ZN6alpaka4exec4cuda6detail10cudaKernelISt17integral_constantImLm1EElNS_4test3mem4view20VerifyBytesSetKernelEJNS8_6traits12IteratorViewIKNS_3mem3buf9BufCudaRtIjS5_lEEvEESH_hEEEvNS_3vec3VecIT_T0_EET1_DpT2_ [259]
<...>
```
</details>

<br/>

With `cudaDeviceEnablePeerAccess` (does a `CUDA memcpy PtoP`):

<details>

```
$ nvprof --print-gpu-trace test/unit/mem/p2p/memP2P 
Running 30 test cases...
==21411== NVPROF is profiling process 21411, command: test/unit/mem/p2p/memP2P

*** No errors detected
==21411== Profiling application: test/unit/mem/p2p/memP2P
==21411== Profiling result:
   Start  Duration            Grid Size      Block Size     Regs*    SSMem*    DSMem*      Size  Throughput  SrcMemType  DstMemType           Device   Context    Stream          Src Dev   Src Ctx          Dst Dev   Dst Ctx  Name
581.10ms  3.7760us                    -               -         -         -         -      160B  40.410MB/s      Device           -    Tesla K80 (0)         1        26                -         -                -         -  [CUDA memset]
581.79ms  3.0400us                    -               -         -         -         -      160B  50.193MB/s      Device      Device    Tesla K80 (0)         1        26    Tesla K80 (0)         1    Tesla K80 (1)         2  [CUDA memcpy PtoP]
582.81ms  50.303us              (1 1 1)         (1 1 1)        32        0B        0B         -           -           -           -    Tesla K80 (1)         2        27                -         -                -         -  _ZN6alpaka4exec4cuda6detail10cudaKernelISt17integral_constantImLm1EElNS_4test3mem4view20VerifyBytesSetKernelEJNS8_6traits12IteratorViewIKNS_3mem3buf9BufCudaRtIjS5_lEEvEESH_hEEEvNS_3vec3VecIT_T0_EET1_DpT2_ [255]
<...>
```

</details>

<br/>

My question is, where to enable peer access? It should be done once at program start, as some time is consumed by the required commands for each pair of GPUs

- `cudaDeviceCanAccessPeer`  ~30µs
- `cudaDeviceEnablePeerAccess` ~400µs

If peer access is already enabled earlier, error will occur. Thus, a variable has to track the state.
For testing I just added few lines to the corresponding createTaskCopy.
```C++
// include/alpaka/mem/buf/cuda/Copy.hpp:513
auto iDst = dev::getDev(viewDst).m_iDevice;                             
auto iSrc = dev::getDev(viewSrc).m_iDevice;                             
int peer_access=0;                                                      
ALPAKA_CUDA_RT_CHECK(cudaDeviceCanAccessPeer(&peer_access, iDst, iSrc));
if(peer_access && !peer0) {                                             
  ALPAKA_CUDA_RT_CHECK(cudaSetDevice(iSrc));                            
  ALPAKA_CUDA_RT_CHECK(cudaDeviceEnablePeerAccess(iDst, 0));            
  peer0=1; // is a global variable just for testing, but there can be more than one pair
}                                                                       
```

Btw, for HIP I want to use peer access anways, as HIP relies on Unified Memory, which allows me some easier workarounds for missing peer-copy-commands (can simply use memcpy commands, which fiddle out peer-to-peer or copy-over-host scenario by itself).

BenjaminW3:
This should fix #596.

It might not be the optimal solution but we can optimize this later if it is really required.


BenjaminW3:
The new version should be better.

BenjaminW3:
The documentation says `Returns cudaErrorPeerAccessAlreadyEnabled if direct access of peerDevice from the current device has already been enabled.` So we could also have called it unconditionally and ignored the `cudaErrorPeerAccessAlreadyEnabled` return code. So storing the already enabled ones is already an optimization.

BenjaminW3:
@tdd11235813 It would be good if you could test if this works because I do not have a system where I can do peer copies.

tdd11235813:
Thank you, @BenjaminW3. Yes, I was not thinking again. Of course, we could use the error value to handle this. 

Just a few changes:
- compiling fails, because the function is in another namespace.
`alpaka::mem::view::cuda::detail::enablePeerAccessIfNotAlreadyEnabled`.
- change to "`==`" in "`if(enabledPeerAccessDevies.find(devicePair) == enabledPeerAccessDevies.end())`"
- change `Devies` to `Devices`

Btw, when not using a lookup, but repeatedly calling `cudaDeviceEnablePeerAccess`, only the first call takes about 400µs. Subsequent calls take about 10µs (whole function 20µs), when P2P is already enabled. Your solution with lookup takes 1µs (whole function) after P2P was enabled, so we take this :)
(measured with std::chrono::high_resolution_clock, on two K80s, Haswell, pinned CPU).

BenjaminW3:
Thank you testing this! Sorry for not compiling the code before committing it!

BenjaminW3:
Ready for review!

PR #586, opened 2018-08-06
Unit test for peer-to-peer memory copies.

tdd11235813:
To test the HIP memory workarounds, I started writing a test for memory copies between two devices. If there is only one device, a boost test message is produced, but the test does not fail.

Scenario:
- allocates buffer on dev0 and dev1
- memsets buf0 on dev0 
- copies buf0 to buf1 (dev0 -> dev1)
- compares bytes set in buf1

At the moment, only `DefaultQueue` is tested, which is normally only async (doesn't exclude this some crucial test cases for alpaka?). A few data types are tested as well, which is not really needed.
I actually need to test sync + async implementation for 1D, 2D and 3D for every accelerator, where two devices are possible.
So, I think to implement another set of types for the test scenarios, just like `TestAccs` or `TestQueues`.

Would you like to have this test case in alpaka and if so, what would be the right style?

// I also checked with nvprof the transfers, see _Details_ for an example output with CUDA + Async.

<details>

```
$> CUDA_VISIBLE_DEVICES=1,2 nvprof test/unit/mem/p2p/memP2P
Running 30 test cases...
==8330== NVPROF is profiling process 8330, command: test/unit/mem/p2p/memP2P

*** No errors detected
==8330== Profiling application: test/unit/mem/p2p/memP2P
==8330== Profiling result:
            Type  Time(%)      Time     Calls       Avg       Min       Max  Name
 GPU activities:   45.88%  34.787ms         1  34.787ms  34.787ms  34.787ms  _ZN6alpaka4exec4cuda6detail10cudaKernelISt17integral_constantImLm3EElNS_4test3mem4view20VerifyBytesSetKernelEJNS8_6traits12IteratorViewIKNS_3mem3buf9BufCudaRtIjS5_lEEvEESH_hEEEvNS_3vec3VecIT_T0_EET1_DpT2_
                   44.54%  33.770ms         1  33.770ms  33.770ms  33.770ms  _ZN6alpaka4exec4cuda6detail10cudaKernelISt17integral_constantImLm3EEmNS_4test3mem4view20VerifyBytesSetKernelEJNS8_6traits12IteratorViewIKNS_3mem3buf9BufCudaRtIjS5_mEEvEESH_hEEEvNS_3vec3VecIT_T0_EET1_DpT2_
[...]

PR #604, opened 2018-08-10
remove static from templates in ViewTest

BenjaminW3:
I extracted the removal of the useless `static` keyword from #586 to test if they are the reason for the failing compilation. It does not seem so, but we can simply merge this.


BenjaminW3:
Reade for review ;-)

Issue #585, opened 2018-08-06
Less and not language specific ALPAKA_FN_* macros


tdd11235813:
Current `ALPAKA_FN_*` macros:
```C
#if BOOST_LANG_CUDA                                
    #define ALPAKA_FN_ACC_CUDA_ONLY __device__     
    #define ALPAKA_FN_ACC_NO_CUDA __host__         
    #if defined(ALPAKA_ACC_GPU_CUDA_ONLY_MODE)     
        #define ALPAKA_FN_ACC __device__           
    #else                                          
        #define ALPAKA_FN_ACC __device__ __host__  
    #endif                                         
    #define ALPAKA_FN_HOST_ACC __device__ __host__ 
    #define ALPAKA_FN_HOST __host__                
```
New version could be:
```C
#if BOOST_LANG_CUDA /* || <more languages> */
    #define ALPAKA_FN_DEVICE __device__
    #if defined(ALPAKA_ACC_GPU_CUDA_ONLY_MODE)
        #define ALPAKA_FN __device__
    #else
        #define ALPAKA_FN __device__ __host__ 
    #endif
    #define ALPAKA_FN_HOST_ACC __device__ __host__
    #define ALPAKA_FN_HOST __host__
```
This is minimal and also allows supporting other languages without involving new macros like `ALPAKA_FN_ACC_NO_HIP`.
What do you think?

sbastrakov:
I probably like 'acc' a little better than 'device', as generally it feels broader and used already in alpaka.

Anyways, I think in your proposal it would be more consistent to rename ```ALPAKA_FN_HOST_ACC``` to ```ALPAKA_FN_HOST_DEVICE```.

ax3l:
We use `acc`/accelerator already for the backend namings in Alpaka.
Device would be conistent with how we call "devices" in PIConGPU (as reference to GPUs or sockets).

tdd11235813:
yes, the inconsistency is a mistake.  

<details>
TL;DR: 

Here are the two options for the new layout, one with `DEVICE`:
```C
#if BOOST_LANG_CUDA /* || <more languages> */
    #define ALPAKA_FN_DEVICE __device__
    #if defined(ALPAKA_ACC_GPU_CUDA_ONLY_MODE)
        #define ALPAKA_FN __device__
    #else
        #define ALPAKA_FN __device__ __host__ 
    #endif
    #define ALPAKA_FN_HOST_DEVICE __device__ __host__
[...]

PR #605, opened 2018-08-11
Replace ALPAKA_FN_ACC_CUDA_ONLY by __device__

BenjaminW3:
As written in #585, `ALPAKA_FN_ACC_CUDA_ONLY` is only ever used in code that is guarded with `BOOST_LANG_CUDA`. It always resolves to `__device__`. To remove the confusion and reduce the number of macros (you know, macros are evil), the simple replacement should be enough.

### DIFF
--- a/doc/markdown/user/implementation/mapping/CUDA.md
+++ b/doc/markdown/user/implementation/mapping/CUDA.md
@@ -29,10 +29,11 @@ NOTE: You have to be careful when mixing alpaka and non alpaka CUDA code. The CU
 |CUDA|alpaka|
 |---|---|
 |\_\_host\_\_|ALPAKA_FN_HOST|
-|\_\_global\_\_|ALPAKA_FN_HOST_ACC|
-|\_\_device\_\_ \_\_host\_\_|ALPAKA_FN_HOST_ACC|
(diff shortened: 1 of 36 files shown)


BenjaminW3:
Ready for review!

doc/markdown/user/implementation/mapping/CUDA.md
```diff
@@ -29,10 +29,11 @@ NOTE: You have to be careful when mixing alpaka and non alpaka CUDA code. The CU
 |CUDA|alpaka|
 |---|---|
 |\_\_host\_\_|ALPAKA_FN_HOST|
-|\_\_global\_\_|ALPAKA_FN_HOST_ACC|
-|\_\_device\_\_ \_\_host\_\_|ALPAKA_FN_HOST_ACC|
-|\_\_device\_\_|ALPAKA_FN_ACC_CUDA_ONLY|
+|\_\_device\_\_|ALPAKA_FN_ACC*|
```
ax3l: I am confused, both lines are `ALPAKA_FN_ACC*` ...

ax3l: Ok, but let's write a comment below to describe that.
Here, we should also write which ones are "internal" for accelerator backends and which ones are "external" for users.

PR #607, opened 2018-08-13
Replace ALPAKA_FN_ACC_NO_CUDA with ALPAKA_FN_HOST

BenjaminW3:
`ALPAKA_FN_ACC_NO_CUDA` is not used consistently and always ever resolves to `__host__`. This has lead to confusion which can be reduced by replacing it directly with `ALPAKA_FN_HOST`.

This is only a first step. We could simply remove `ALPAKA_FN_HOST` in a seperate PR.


BenjaminW3:
Ready to be merged! (Appveyor seems to have an independant SSL problem but is not required for merging)

PR #616, opened 2018-08-21
Fix P2P test

### LINKED ISSUE #615: Do not throw an error when P2P access is not possible.
On systems, where GPUs are not directly connected via PCIe, P2P test fails. However, a warning message should be ok. I did not think about it when checking #602 .

Current Copy.hpp<cuda> code:
```C++
ALPAKA_CUDA_RT_CHECK(cudaDeviceCanAccessPeer(&canAccessPeer, devSrc, devDst));                                                
if(!canAccessPeer) {                                                                                                          
    std::string const sError("Peer access is not possible! src=" + std::to_string(devSrc) + " dst=" + std::to_string(devDst));
    throw std::runtime_error(sError);                                                                                         
}                                                                                                                             
```

BenjaminW3:
Fixes #615

### DIFF
--- a/include/alpaka/mem/buf/cuda/Copy.hpp
+++ b/include/alpaka/mem/buf/cuda/Copy.hpp
@@ -396,7 +396,9 @@ namespace alpaka
                     };
 
                     //-----------------------------------------------------------------------------
-                    inline auto enablePeerAccessIfNotAlreadyEnabled(
+                    //! Not being able to enable peer access does not prevent such device to device memory copies.
+                    //! However, those copies may be slower because the memory is copied via the CPU.
+                    inline auto enablePeerAccessIfPossible(
                         const int & devSrc,
                         const int & devDst)
                     -> void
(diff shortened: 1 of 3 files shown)


ax3l:
@tdd11235813 does this work for you?

tdd11235813:
Test case non P2P accessible GPUs
> CUDA_VISIBLE_DEVICES=1,2 test/unit/mem/p2p/memP2P --log_level=message
Running 30 test cases...
No two devices found to test peer-to-peer copy.  // <--- CPU back-end
Peer access between the two devices is not possible.
...
*** No errors detected

Output is ok, but now there are no memcopies in that case. The memcopies should be executed nonetheless, which would run internally via host memory. I know, the test is called memP2P, but it should be part of the test, that even in the nonP2P case, memory is copied from one device to another. Maybe the test should be called memD2D then ;)

Regarding the functions; An idea could be to refactor `enablePeerAccessIfNotAlreadyEnabled` to `enablePeerAccessIfPossible` and to not throw an error, if P2P is not possible or if it is already enabled. 
However, the reason should be returned then and therefore be buffered like the device index pair, so a failed P2P access can be detected and handled (within CUDA back-end code).

ROCm also provides a P2P API [ [1](https://rocm.github.io/ROCmMultiGPU.html), [2](https://rocm-documentation.readthedocs.io/en/latest/Remote_Device_Programming/Remote-Device-Programming.html), [3](https://github.com/RadeonOpenCompute/ROCm/issues/380) ]. Not sure, if it is needed to generalize the P2P API in alpaka. As long as P2P does not have side effects, it should be enabled where it is possible and that can be done "hidden" in the back-end. At the moment I do not see a realistic case, that would speak against the use of P2P.

BenjaminW3:
Ok, now I see your point. I will fix the `enablePeerAccessIfNotAlreadyEnabled` as you have described it.

BenjaminW3:
updated

include/alpaka/mem/buf/cuda/Copy.hpp
```diff
@@ -407,27 +409,33 @@ namespace alpaka
...
 #endif
                         auto const devicePair = std::make_pair(devSrc, devDst);
 
-                        if(enabledPeerAccessDevices.find(devicePair) == enabledPeerAccessDevices.end())
+                        if(alreadyCheckedPeerAccessDevices.find(devicePair) == alreadyCheckedPeerAccessDevices.end())
                         {
-                            int canAccessPeer=0;
+                            int canAccessPeer = 0;
                             ALPAKA_CUDA_RT_CHECK(cudaDeviceCanAccessPeer(&canAccessPeer, devSrc, devDst));
                             if(!canAccessPeer) {
```
tdd11235813: maybe the already checked pair, which does not have P2P access, also can be added to the `alreadyCheckedPeerAccessDevices`. 
A `cudaDeviceCanAccessPeer` takes about 20-30µs on a Kepler+Haswell system. On a Pascal/Volta+Broadwell system, this goes down to 1µs ... about the same time what a lookup takes.

BenjaminW3: Yes, this was my intention and therefore the renaming. I simply forgot to add it to the list.

tdd11235813:
ok, `BOOST_VERIFY` is [like an assert](https://www.boost.org/doc/libs/1_65_1/libs/assert/doc/html/assert.html#boost_verify). And it does nothing, when `NDEBUG` is defined? (in release mode some tests do not test?). So, if the `verify...` functions are done, a `BOOST_CHECK(true);` should be called. However, I do not see this issue in the memView test. 
For the memP2P code, we could move [this line](https://github.com/ComputationalRadiationPhysics/alpaka/blob/d48c3c0ecef8d0f9899c2615b8a6908cbac4e11a/test/unit/mem/p2p/src/P2P.cpp#L123) after the `#endif`, so it is executed in the end, if there were no errors before.

BenjaminW3:
Yes, `BOOST_VERIFY` is used within kernels because the boost test macros can not be used there. We should switch to or develop an alternative.

include/alpaka/mem/buf/cuda/Copy.hpp
BenjaminW3: fixed

BenjaminW3:
I have added a final `BOOST_CHECK(true)` for now. The rest should be solved in #617.

PR #619, opened 2018-08-27
Reduction example

kloppstock:
This adds an optimized reduction to the examples folder.


example/reduce/src/kernel.hpp
```diff
@@ -0,0 +1,170 @@
...
+        if (threadIndex < n)
+            sdata[threadIndex] = result;
+
+        alpaka::block::sync::syncBlockThreads(acc);
+
+        // --------
+        // Level 2: block + warp reduce, reading from shared memory
+        // --------
+
+#pragma unroll
```
BenjaminW3: `#pragma unroll` is not supported by all compilers (as the CI shows). Try to use `ALPAKA_UNROLL` from `alpaka/core/Unroll.hpp` instead. It might be a bit out of date but it hopefully works.

BenjaminW3:
Thank you very much for this example. I will do a real review after it is successfully compiling.
If you do not have the time to fix it, I can do it as well.

example/reduce/src/iterator.hpp
```diff
@@ -0,0 +1,422 @@
...
+    const uint64_t maximum;
+
+public:
+    //-----------------------------------------------------------------------------
+    //! Constructor.
+    //!
+    //! \param data A pointer to the data.
+    //! \param linearizedIndex The linearized index.
+    //! \param gridSize The grid size.
+    //! \param n The problem size.
```
sbastrakov: The comment does not fully match the function parameters (probably was not updated)

example/reduce/src/iterator.hpp
```diff
@@ -0,0 +1,422 @@
...
+                                                 uint64_t maximum)
+        : data(data), index(index), maximum(maximum)
+    {
+    }
+
+    //-----------------------------------------------------------------------------
+    //! Constructor.
+    //!
+    //! \param other The other iterator object.
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE Iterator(const Iterator &other)
```
sbastrakov: Wouldn't `= default` copy constructor express the same?

example/reduce/src/iterator.hpp
```diff
@@ -0,0 +1,422 @@
...
+    //!
+    //! \param other The other object.
+    //!
+    //! Returns true if objects are equal and false otherwise.
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE auto
+    operator==(const Iterator &other) const -> bool
+    {
+        return (this->data == other.data) && (this->index == other.index) &&
+               (this->maximum == other.maximum) &&
+               (this->gridSize == other.gridSize);
```
sbastrakov: Is `gridSize `actually a member (or maybe I'm just blind)?

example/reduce/src/reduce.cpp
```diff
@@ -0,0 +1,155 @@
...
+
+    DevAcc devAcc(alpaka::pltf::getDevByIdx<PltfAcc>(dev));
+
+    // reduce
+    bool result = reduce(devAcc, n);
+
+    // clean up
+    alpaka::dev::reset(devAcc);
+
+    return (result ? 0 : -1);
```
BenjaminW3: Please use `EXIT_SUCCESS` and `EXIT_FAILURE`.

example/reduce/src/reduce.cpp
```diff
@@ -0,0 +1,155 @@
...
+using MaxBlockSize = Accelerator::MaxBlockSize;
+
+//-----------------------------------------------------------------------------
+//! Reduces the numbers 1 to n.
+//!
+//! \param devAcc The accelerator object.
+//! \param n The problem size.
+//!
+//! Returns true if the reduction was correct and false otherwise.
+bool reduce(DevAcc devAcc, uint64_t n)
```
BenjaminW3: I would prefer if this was a real reduction function meaning:
* The source buffer is generated outside of this method and passed into it
* The reduction function is passed into this method
* The reuction specific result is check outside of this method
This should all happen within the main function.

example/reduce/src/reduce.cpp
ax3l: note: include `<cstdlib>` for that https://en.cppreference.com/w/cpp/utility/program/EXIT_status

example/reduce/src/reduce.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    // fill array with data
+    for (uint64_t i = 0; i < n; i++)
+        nativeHostMemory[i] = static_cast<T>(i + 1);
+
+    // define the reduction function
+    auto addFn = [] ALPAKA_FN_ACC(T a, T b) -> T { return a + b; };
+
+    // reduce
+    T result = reduce<T, decltype(addFn)>(devHost, devAcc, queue, n, hostMemory, addFn);
```
BenjaminW3: Do you really need to write the template parameters? You are providing all the parameters and the templates should be deducible.

example/reduce/src/reduce.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+    // clean up
+    alpaka::dev::reset(devAcc);
```
BenjaminW3: Do we need to reset explicitly? We should keep the examples minimal.

example/reduce/src/reduce.cpp
kloppstock: The ```decltype(addFn)``` seems to be deducible but it wont compile if the ```T``` in line 162 is left out es well.

PR #623, opened 2018-08-29
Replace BOOST_VERIFY by ALPAKA_CHECK and return success from all test kernels

### LINKED ISSUE #617: Replace BOOST_VERIFY in tests, as it does nothing in release mode
`BOOST_VERIFY` behaves [like an assert](https://www.boost.org/doc/libs/1_65_1/libs/assert/doc/html/assert.html#boost_verify). And it does nothing, when `NDEBUG` is defined, which means in release mode tests like `verifyBytesSet` and `verifyViewsEqual` do not test (it only evaluates the expression itself). 
It can be reproduced, when the `==` is switched to `!=` in the verify kernels. In release mode, the test mistakenly passes.

And just as an additional minor issue: `BOOST_VERIFY` does not report the positive state back to the test suite (as it is just an BOOST_ASSERT in non-release mode).
For example, if there would be only `BOOST_VERIFY` calls in the `memView` test, it would need something like a `BOOST_CHECK(true);` in the end, otherwise Boost does not know about the state of the test (it just passed without flagging it true or false). Of course, within the `memView` test there is a `BOOST_REQUIRE_EQUAL` (==`BOOST_TEST_REQUIRE`), which does report back. So here, `memView` has a defined state. For the memP2P test, this is missing, thus this test can have an undefined state, which yields a log message:
```
Test case memP2P/memP2PTest<alpaka__acc__AccGpuCudaRt<std__integral_constant<unsigned_long,_1ul>,_long>> did not check any assertions
```

Relates to #367 and #616 .

BenjaminW3:
Fixes #617

[edit by @psychocoder]
This PR also added 
  - an overview of the free memory of travis test instance
  - a new unit test to test generic lamdas

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -141,7 +141,7 @@ before_build:
     - cmd: if not "%ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE%"=="OFF" set ALPAKA_BOOST_B2=%ALPAKA_BOOST_B2% --with-fiber --with-context --with-thread --with-system --with-atomic --with-chrono --with-date_time
 
     - cmd: if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" set ALPAKA_BOOST_TOOLSET=msvc-14.1
-    - cmd: b2 -j2 --toolset=%ALPAKA_BOOST_TOOLSET% --layout=versioned %ALPAKA_BOOST_B2% architecture=x86 address-model=%ALPAKA_BOOST_ADDRESS_MODEL% variant=%ALPAKA_BOOST_VARIANT% link=static threading=multi runtime-link=shared define=_CRT_NONSTDC_NO_DEPRECATE define=_CRT_SECURE_NO_DEPRECATE define=_SCL_SECURE_NO_DEPRECAT define=BOOST_USE_WINFIBERS --stagedir="%ALPAKA_B2_STAGE_DIR%"
+    - cmd: b2 -j2 --toolset=%ALPAKA_BOOST_TOOLSET% --layout=versioned %ALPAKA_BOOST_B2% architecture=x86 address-model=%ALPAKA_BOOST_ADDRESS_MODEL% variant=%ALPAKA_BOOST_VARIANT% link=static threading=multi runtime-link=shared define=_CRT_NONSTDC_NO_DEPRECATE define=_CRT_SECURE_NO_DEPRECATE define=_SCL_SECURE_NO_DEPRECAT define=BOOST_USE_WINFIBERS define=_ENABLE_EXTENDED_ALIGNED_STORAGE --stagedir="%ALPAKA_B2_STAGE_DIR%"
 
     #-------------------------------------------------------------------------------
     # Install TBB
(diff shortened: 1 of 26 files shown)


.appveyor.yml
```diff
@@ -141,7 +141,7 @@ before_build:
     - cmd: if not "%ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE%"=="OFF" set ALPAKA_BOOST_B2=%ALPAKA_BOOST_B2% --with-fiber --with-context --with-thread --with-system --with-atomic --with-chrono --with-date_time
 
     - cmd: if "%APPVEYOR_BUILD_WORKER_IMAGE%"=="Visual Studio 2017" set ALPAKA_BOOST_TOOLSET=msvc-14.1
-    - cmd: b2 -j2 --toolset=%ALPAKA_BOOST_TOOLSET% --layout=versioned %ALPAKA_BOOST_B2% architecture=x86 address-model=%ALPAKA_BOOST_ADDRESS_MODEL% variant=%ALPAKA_BOOST_VARIANT% link=static threading=multi runtime-link=shared define=_CRT_NONSTDC_NO_DEPRECATE define=_CRT_SECURE_NO_DEPRECATE define=_SCL_SECURE_NO_DEPRECAT define=BOOST_USE_WINFIBERS --stagedir="%ALPAKA_B2_STAGE_DIR%"
+    - cmd: b2 -j2 --toolset=%ALPAKA_BOOST_TOOLSET% --layout=versioned %ALPAKA_BOOST_B2% architecture=x86 address-model=%ALPAKA_BOOST_ADDRESS_MODEL% variant=%ALPAKA_BOOST_VARIANT% link=static threading=multi runtime-link=shared define=_CRT_NONSTDC_NO_DEPRECATE define=_CRT_SECURE_NO_DEPRECATE define=_SCL_SECURE_NO_DEPRECAT define=BOOST_USE_WINFIBERS define=_ENABLE_EXTENDED_ALIGNED_STORAGE --stagedir="%ALPAKA_B2_STAGE_DIR%"
```
BenjaminW3: This is necessary for MSVC 15.8 support. It fixed a bug that changed the memory layout and you have to explicitly opt-in or opt-out. Else you will get an error when you use this feature.

.travis.yml
```diff
@@ -162,6 +162,10 @@ cache:
         - $ALPAKA_CI_DOCKER_CACHE_DIR
 
 script:
+    - sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install smem
+    - sudo smem
+    - sudo free -m -t
```
BenjaminW3: I added some memory statistics here and there and stopped some services that are not necessary because the new variables in each kernel increased the build times and I tried to enable paralle compilation but this increased the memory consumption too much. So we are still doing sequential compilation but the memory statistics are now here ;-)

test/common/include/alpaka/test/acc/Acc.hpp
```diff
@@ -253,17 +224,20 @@ namespace alpaka
                     // size_t is most probably identical to either std::uint64_t or std::uint32_t.
                     // This would lead to duplicate tests (especially test names) which is not allowed.
                     //std::size_t,
+#if !defined(ALPAKA_CI)
                     std::int64_t,
+#endif
                     std::uint64_t,
                     std::int32_t,
+#if !defined(ALPAKA_CI)
                     std::uint32_t,
```
BenjaminW3: I had to exclude the tests for `std::int64_t` and `std::uint32_t` as index type from the CI because the compilation times were much too high. While doing this I merged it with the special CUDA version from above.

BenjaminW3:
The appveyor build is now fixed.

PR #625, opened 2018-08-30
Fix ViewSubViewTest exclusion from Appveyor build

BenjaminW3:


BenjaminW3:
A different test fails as well. I will close this for now.

PR #627, opened 2018-09-03
Fix empty(QueueCpuAsync) returning true even though the last task is still executing

### LINKED ISSUE #621: QueueCpuAsync tells it is empty although a task is still running
I am having the following problem: I am writing a simple blur kernel example, in which the image is chunked in smaller parts, which shall be copied, processed and copied back in parallel in different queues to overlap copying and computing. The idea is, if one of these queue is empty, I know I can savely feed a new chunk to this queue.

However the check for emptyness of the stream does not work because of theese lines in https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/core/ConcurrentExecPool.hpp#L595 :
```C++
if(popTask(currentTaskPackage))
{
    currentTaskPackage->runTask();
}
```
There a task is popped from the queue and processed. However if I check for the emptyness of the queue **after** `popTask(currentTaskPackage)`, it will return `true` although `currentTaskPackage->runTask();` may still run. So in my example I am starting to copy data to the "free" chunk although the copying back of data is not finished (yet).

Maybe the same problems occurs for `alpaka::wait::wait( … )`, I did not check this.

BenjaminW3:
Fixes #621

By the way the test found that this did not work for `QueueCpuSync` and `QueueCudaRtSync` as well (when the last task was a callback).

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -174,7 +174,7 @@ namespace alpaka
                 TaskPkg(
                     TFnObj && func) :
                         m_Promise(),
-                        m_FnObj(std::forward<TFnObj>(func))
+                        m_FnObj(std::move(func))
                 {}
 
             private:
@@ -223,7 +223,7 @@ namespace alpaka
                 TaskPkg(
                     TFnObj && func) :
                         m_Promise(),
-                        m_FnObj(std::forward<TFnObj>(func))
+                        m_FnObj(std::move(func))
                 {}
 
             private:
@@ -253,6 +253,37 @@ namespace alpaka
                 // `std::remove_reference` enforces the function object to be copied.
                 typename std::remove_reference<TFnObj>::type m_FnObj;
             };
+
+            //-----------------------------------------------------------------------------
+            template<
+                typename TFnObj0,
+                typename TFnObj1,
+                typename = typename std::enable_if<!std::is_same<void, decltype(std::declval<TFnObj0>()())>::value>::type>
+            auto invokeBothReturnFirst(
+                    TFnObj0 && fn0,
+                    TFnObj1 && fn1)
+#ifdef BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION
+             -> decltype(std::declval<TFnObj0>()())
+#endif
+            {
+                auto ret = fn0();
+                fn1();
+                return std::move(ret);
+            }
+
+            //-----------------------------------------------------------------------------
+            template<
+                typename TFnObj0,
+                typename TFnObj1,
+                typename = typename std::enable_if<std::is_same<void, decltype(std::declval<TFnObj0>()())>::value>::type>
+            auto invokeBothReturnFirst(
+                    TFnObj0 && fn0,
+                    TFnObj1 && fn1)
+            -> void
+            {
+                fn0();
+                fn1();
+            }
 
             //#############################################################################
             //! ConcurrentExecPool using yield.
@@ -351,16 +382,29 @@ namespace alpaka
                     TFnObj && task,
                     TArgs && ... args)
 #ifdef BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION
+#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0))
+                // FIXME: gcc 4.9 does not support the syntax below. Restricting the return type to void works because we never use something else within alpaka.
+                -> decltype(std::declval<TPromise<void>>().get_future())
+#else
                 -> decltype(std::declval<TPromise<decltype(task(args...))>>().get_future())
 #endif
-                {
-#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0))
-                    auto boundTask(std::bind(task, args...));
-#else
-                    auto boundTask([=](){return task(args...);});
 #endif
-                    using TaskPackage = TaskPkg<TPromise, decltype(boundTask)>;
-                    auto pTaskPackage(new TaskPackage(std::move(boundTask)));
+                {
+                    auto boundTask([=](){return task(args...);});
+                    auto decrementNumActiveTasks = [this](){--m_numActiveTasks;};
+
+                    auto extendedTask(
+                        [boundTask, decrementNumActiveTasks]()
+                        {
+                            return
+                                invokeBothReturnFirst(
+                                    std::move(boundTask),
+                                    std::move(decrementNumActiveTasks)
+                                );
+                        });
+
+                    using TaskPackage = TaskPkg<TPromise, decltype(extendedTask)>;
+                    auto pTaskPackage(new TaskPackage(std::move(extendedTask)));
                     std::shared_ptr<ITaskPkg> upTaskPackage(pTaskPackage);
 
                     auto future(pTaskPackage->m_Promise.get_future());
@@ -399,8 +443,6 @@ namespace alpaka
                         if(popTask(currentTaskPackage))
                         {
                             currentTaskPackage->runTask();
-
-                            --m_numActiveTasks;
                         }
                         else
                         {
@@ -548,16 +590,29 @@ namespace alpaka
                     TFnObj && task,
                     TArgs && ... args)
 #ifdef BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION
+#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0))
+                // FIXME: gcc 4.9 does not support the syntax below. Restricting the return type to void works because we never use something else within alpaka.
+                -> decltype(std::declval<TPromise<void>>().get_future())
+#else
                 -> decltype(std::declval<TPromise<decltype(task(args...))>>().get_future())
 #endif
-                {
-#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0))
-                    auto boundTask(std::bind(task, args...));
-#else
-                    auto boundTask([=](){return task(args...);});
 #endif
-                    using TaskPackage = TaskPkg<TPromise, decltype(boundTask)>;
-                    auto pTaskPackage(new TaskPackage(std::move(boundTask)));
+                {
+                    auto boundTask([=](){return task(args...);});
+                    auto decrementNumActiveTasks = [this](){--m_numActiveTasks;};
+
+                    auto extendedTask(
+                        [boundTask, decrementNumActiveTasks]()
+                        {
+                            return
+                                invokeBothReturnFirst(
+                                    std::move(boundTask),
+                                    std::move(decrementNumActiveTasks)
+                                );
+                        });
+
+                    using TaskPackage = TaskPkg<TPromise, decltype(extendedTask)>;
+                    auto pTaskPackage(new TaskPackage(std::move(extendedTask)));
                     std::shared_ptr<ITaskPkg> upTaskPackage(pTaskPackage);
 
                     auto future(pTaskPackage->m_Promise.get_future());
@@ -601,8 +656,6 @@ namespace alpaka
                         if(popTask(currentTaskPackage))
                         {
                             currentTaskPackage->runTask();
-
-                            --m_numActiveTasks;
                         }
                         {
                             std::unique_lock<TMutex> lock(m_mtxWakeup);
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -277,6 +277,7 @@ namespace alpaka
                     TIdx concurrentExecutionCount) :
                     m_vConcurrentExecs(),
                     m_qTasks(),
+                    m_numActiveTasks(0u),
                     m_bShutdownFlag(false)
                 {
                     if(concurrentExecutionCount < 1)
@@ -327,7 +328,7 @@ namespace alpaka
                 //-----------------------------------------------------------------------------
                 //! Runs the given function on one of the pool in First In First Out (FIFO) order.
                 //!
-                //! \tparam TFnObj   The function type.
+                //! \tparam TFnObj  The function type.
                 //! \param task     Function object to be called on the pool.
                 //!                 Takes an arbitrary number of arguments and arbitrary return type.
                 //! \tparam TArgs   The argument types pack.
@@ -361,6 +362,7 @@ namespace alpaka
 
                     auto future(pTaskPackage->m_Promise.get_future());
 
+                    ++m_numActiveTasks;
                     m_qTasks.push(std::move(upTaskPackage));
 
                     return future;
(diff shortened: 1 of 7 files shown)


include/alpaka/queue/QueueCpuSync.hpp
```diff
@@ -186,8 +189,7 @@ namespace alpaka
                     queue::QueueCpuSync const & queue)
                 -> bool
                 {
-                    alpaka::ignore_unused(queue);
-                    return true;
+                    return !queue.m_spQueueImpl->m_bCurrentlyExecutingTask;
```
theZiz: I am wondering for which case this is useful? If I call enqueue, the task will not return until, so nobody except the task itself has a change to check to queue for emptyness. Of course a concurrent task could test the queue, but in that case `m_bCurrentlyExecutingTask `should be an atomic?

BenjaminW3: The task being executed can be a host callback which can check the state (this is exactly what the unit test is doing).

include/alpaka/queue/QueueCudaRtSync.hpp
```diff
@@ -247,19 +269,37 @@ namespace alpaka
...
+                                pCallbackSynchronizationData->state = CallbackState::finished;
                             }
-                        );
-                    }
+                            pCallbackSynchronizationData->m_event.notify_one();
+                        }
+                    );
 
-                    task();
+                    t.join();
```
BenjaminW3: The code here is not new but only copied from the `QueueCudaRtAsync`. This had been forgotten when a bug had been fixed there.

include/alpaka/queue/QueueCpuSync.hpp
theZiz: Ah, makes sense. I will just check the changes with my code. :wink:

BenjaminW3:
Hmmm... CI had one single build that was not happy with the `QueueCpuAsync` unit test.
I think I understand what the problem is but it is really really hard to solve.
The test is doing `wait(queue)` and after this expects that `empty(queue)` is true. To understand the failure you have to know how the CPU queue implements waiting. `wait(queue)` simply enqueues an event and waits for this event. The event itself is only a `condition_variable` which is notified once the event is "executed". Now there is a very very short time in between the event notification and the line where it is counted as finished. In the end it boils down to the event being a task that does essentially `m_conditionVariable.notify_all();` and the task queue doing `--m_numActiveTasks;` If the thread is interrupted in between, the event has been notified, the wait for the event/queue finishes and `empty(queue)` still returns false.

theZiz:
I understand the problem, but except for the test, I think the solution is "less wrong" than the old behaviour. Of course of you check for emptiness, you will get an unexpected result, but at least you can **assume** emptiness of the queue, as no user-given task is executed anymore.
Maybe this would be even a solution? Flagging tasks, whether they are "real" user tasks or internal tasks like for the event waiting. Not incrementing and decrementing the counter for the internal event task for `wait(queue)` may be a simple solution?

test/unit/queue/src/QueueTest.cpp
```diff
@@ -93,32 +93,45 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(
...
 
     // A synchronous queue will always stay empty because the task has been executed immediately.
     if(!alpaka::test::queue::IsSyncQueue<typename Fixture::Queue>::value)
     {
         alpaka::wait::wait(f.m_queue);
     }
 
-    BOOST_REQUIRE_EQUAL(true, alpaka::queue::empty(f.m_queue));
-    BOOST_REQUIRE_EQUAL(true, CallbackFinished);
+    BOOST_CHECK_EQUAL(true, alpaka::queue::empty(f.m_queue));
```
BenjaminW3: I would have to comment out this line to make the test successful again. I will think about substituting one bug by another one.

BenjaminW3:
I have now made changes to the events so that they are not signaled ready before they are removed from the queue. I had to try many ways to find the current solution but now it is cleaner than the original version.
I might have to fix a gcc-4.9 workaround I have removed at the moment but CI will tell me.

ax3l:
ping @theZiz

theZiz:
I was on small holidays the last 4 days, will review & check now.

theZiz:
Okay, as it seems, the change is now completely broken. The queues never get empty, so may test program runs into an endless loop while waiting for free resources. :disappointed: 
I will have a look into the changes, maybe I find the problem myself.

BenjaminW3:
That is bad. If you can reduce the calls to create the error or, even better, create a unit test that shows the faulty behavior, I can fix it.

theZiz:
Okay, I got a small example to show case the failure (no unit test, sorry):
```C++
#include <alpaka.hpp>
#include <cstdio>

#define CHECK_EMPTY 1
#define CHECK_WAIT 0

struct TestKernel
{
    template< typename T_Acc >
    ALPAKA_FN_ACC
    void operator()( T_Acc const &acc ) const
    {
        usleep(1000000); // 1 second
        printf("Kernel: Done\n");
    }
};

int main()
{
    using Dim = alpaka::dim::DimInt< 1 >;
    using Size = std::size_t;
    using Host = alpaka::acc::AccCpuSerial<Dim, Size>;
    using Acc = alpaka::acc::AccCpuOmp2Blocks<Dim, Size>;
    using Queue = alpaka::queue::QueueCpuAsync;
    using DevAcc = alpaka::dev::Dev<Acc>;
    using PltfAcc = alpaka::pltf::Pltf<DevAcc>;
    using WorkDiv = alpaka::workdiv::WorkDivMembers<Dim, Size>;

    DevAcc const devAcc( alpaka::pltf::getDevByIdx< PltfAcc >( 0u ) );
    Queue queue(devAcc);
    printf("Host: Emptiness of queue right after creation: %i\n",alpaka::queue::empty( queue ) );
    auto const workdiv = alpaka::workdiv::WorkDivMembers<
        Dim,
        Size
    > {
        alpaka::vec::Vec< Dim, Size >( static_cast< Size >( 1 ) ),
        alpaka::vec::Vec< Dim, Size >( static_cast< Size >( 1 ) ),
        alpaka::vec::Vec< Dim, Size >( static_cast< Size >( 1 ) )
    };

    TestKernel kernel;
    alpaka::kernel::exec<Acc>(
        queue,
        workdiv,
        kernel
    );
    printf("Host: Queued kernel\n");
#if CHECK_WAIT == 1
    printf("Host: Waiting for queue\n");
    alpaka::wait::wait( queue );
    printf("Host: Queue finished\n");
#endif
#if CHECK_EMPTY == 1
    int count = 0;
    while ( !alpaka::queue::empty( queue ) )
    {
        printf("Host: Queue not empty, waiting 0.1s (total: %i.%is)\n",count/10,count%10);
        count++;
        usleep(100000); // 0.1 seconds
    }
#endif
    return 0;
}
```
This program runs a kernel in background, which waits for 1 second and then prints that it is done. You can test for `alpaka::wait::wait` (which works) and busy waiting with `alpaka::queue::empty`(which does not work). So the result of this test program for the empty-check is:
```
Host: Emptiness of queue right after creation: 1
Host: Queued kernel
Host: Queue not empty, waiting 0.1s (total: 0.0s)
Host: Queue not empty, waiting 0.1s (total: 0.1s)
Host: Queue not empty, waiting 0.1s (total: 0.2s)
Host: Queue not empty, waiting 0.1s (total: 0.3s)
Host: Queue not empty, waiting 0.1s (total: 0.4s)
Host: Queue not empty, waiting 0.1s (total: 0.5s)
Host: Queue not empty, waiting 0.1s (total: 0.6s)
Host: Queue not empty, waiting 0.1s (total: 0.7s)
Host: Queue not empty, waiting 0.1s (total: 0.8s)
Host: Queue not empty, waiting 0.1s (total: 0.9s)
Kernel: Done
Host: Queue not empty, waiting 0.1s (total: 1.0s)
Host: Queue not empty, waiting 0.1s (total: 1.1s)
Host: Queue not empty, waiting 0.1s (total: 1.2s)
Host: Queue not empty, waiting 0.1s (total: 1.3s)
Host: Queue not empty, waiting 0.1s (total: 1.4s)
Host: Queue not empty, waiting 0.1s (total: 1.5s)
Host: Queue not empty, waiting 0.1s (total: 1.6s)
Host: Queue not empty, waiting 0.1s (total: 1.7s)
Host: Queue not empty, waiting 0.1s (total: 1.8s)
Host: Queue not empty, waiting 0.1s (total: 1.9s)
Host: Queue not empty, waiting 0.1s (total: 2.0s)
Host: Queue not empty, waiting 0.1s (total: 2.1s)
Host: Queue not empty, waiting 0.1s (total: 2.2s)
Host: Queue not empty, waiting 0.1s (total: 2.3s)
^C
```
I finished the program with Ctrl+C. You can also see, that the queue is empty right after creation.
I compiled the program with this simple Makefile:
```Makefile
all: cpu-test

cpu-test: cpu-test.cpp Makefile
	g++ -o cpu-test -O1 cpu-test.cpp -std=c++11 -fopenmp -I../alpaka/include/alpaka -I../alpaka/include -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_DEBUG=0 -DBOOST_ALL_NO_LIB
```
I hope this helps or you can create a unittest out of this. I will continue trying to find the bug in the changes myself, too.

BenjaminW3:
Thank you very much! This should be enough.
One thing I found while doing the implementation is that the behaviour of the unit tests change depending on the number of checks you are doing. Some of the constructs for getting the state of an event and testing a queue for being empty influence the state of the event. So it is a bit like quantum theory ;-)

theZiz:
The fun of checking concurrent code for correctness. :laughing:

BenjaminW3:
I am sorry to say, but it works perfectly on my machine with both defines. The only changes I made are to replace `usleep` with something standard compliant: `std::this_thread::sleep_for(std::chrono::seconds(1))`. I will push the hacked test to CI. Let`s see what Ubuntu says.

theZiz:
I'll will give it a try on Hypnos

theZiz:
Okay, on Hypnos it works with `usleep`. On my local machine it doesn't work neither with `std::this_thread::sleep_for(std::chrono::seconds(1))` nor `usleep`. But if I remove the sleep command at all and do busy waiting with full CPU load, it works. :confused:

I will try to figure out whether my system is broken or whether I am lucky to have the only configuration to trigger a real bug. :laughing:

include/alpaka/event/EventCpu.hpp
```diff
@@ -77,16 +77,25 @@ namespace alpaka
...
+                    auto wait(std::size_t const & enqueueCount) noexcept -> void
                     {
-                        return (m_LastReadyEnqueueCount >= enqueueCount);
+                        std::unique_lock<std::mutex> lk(m_mutex);
+                        assert(enqueueCount <= m_enqueueCount);
+
+                        while(enqueueCount > m_LastReadyEnqueueCount)
+                        {
+                            auto future = m_future;
+                            lk.unlock();
```
theZiz: Shoudln't that the other way around? locking, working, unlocking?

BenjaminW3: No, I think this is correct. At a specific point A in time we want to wait. In this case we lock the event mutex, check if we need to wait at all and copy the current future. Then we unlock the event and wait for the future to finish. This waiting can not be done in locked moder because it might deadlock for multiple reasons but most prominently because the execution of the event itself requires the lock to increase the `m_LastReadyEnqueueCount`.
We waited on a copy of the future because in the meantime the event could have been re-enqueued to a later point in time. This is no problem because after waiting we simply lock the event again and and the while check is false and we correctly waited for the event.

theZiz: You are right, sorry, I mixed up `lock` and `mutex` :man_facepalming:

theZiz:
Now for something completely different!
I found the "bug". I already had the topic branch with your first fix attempt and just merged the new changes to it. Will, figures out, this failed. So with removing the branch and repulling your latest changes, everything works fine, thx! :laughing: At least we got a new unit test... :roll_eyes:

BenjaminW3:
But the new unit test does not compile due to many warnings and it is not really new because both things are already tested. I have pushed some minor changes and revert the test.

BenjaminW3:
@theZiz Please approve if this works for you.

theZiz:
> @theZiz Please approve if this works for you.

Works! :+1:

BenjaminW3:
@theZiz You should be allowed to give a real review approval so that I can merge it.

theZiz:
> @theZiz You should be allowed to give a real review approval so that I can merge it.

I doubt so. @ax3l or @psychocoderHPC need to approve the PR.

psychocoderHPC:
I will check the PR in ~2h

BenjaminW3:
@psychocoderHPC Are you done?

psychocoderHPC:
no sry I start now. Was busy with other tasks

include/alpaka/queue/QueueCudaRtSync.hpp
```diff
@@ -247,19 +269,37 @@ namespace alpaka
...
 
-                    // If the callback has not yet been called, we wait for it.
-                    std::unique_lock<std::mutex> lock(pCallbackSynchronizationData->m_mutex);
-                    if(!pCallbackSynchronizationData->notified)
-                    {
-                        pCallbackSynchronizationData->m_event.wait(
-                            lock,
-                            [pCallbackSynchronizationData](){
-                                return pCallbackSynchronizationData->notified;
+                    // We start a new std::thread which stores the task to be executed.
```
psychocoderHPC: Do we always enqueue CUDA callbacks with alpaka?

psychocoderHPC: OK after reviewing the corresponding parts again I found that this is the code to create a callback introduced in #373. 
The code itself looks like it is the kernel enqueue stuff but kernel start in the name space alpaka::exec.

psychocoderHPC:
form my side it can be merged if expect @BenjaminW3  will fix the (style issue)[https://github.com/ComputationalRadiationPhysics/alpaka/pull/627#discussion_r218464880]

include/alpaka/queue/QueueCudaRtSync.hpp
BenjaminW3: I have not changed anything about the `cudaStreamAddCallback` here. The bug I found was that the callback was directly executed within the CUDA callback thread. In the CUDA callback thread you are not allowed to do CUDA calls, which I did in my test. This had already been fixed in the Async version. I simply copied the fix from over there.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -343,24 +382,34 @@ namespace alpaka
...
+                // FIXME: gcc 4.9 does not support the syntax below. Restricting the return type to void works because we never use something else within alpaka.
+                -> decltype(std::declval<TPromise<void>>().get_future())
 #else
-                    auto boundTask([=](){task(args...);});
+                -> decltype(std::declval<TPromise<decltype(task(args...))>>().get_future())
 #endif
+#endif
+                {
+                    auto boundTask([=](){return task(args...);});
+                    auto decrementNumActiveTasks = [this](){--m_numActiveTasks;};
```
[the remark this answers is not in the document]
BenjaminW3: Fixed

BenjaminW3:
Please re-approve

PR #630, opened 2018-09-04
Add a check for OpenMP version

### LINKED ISSUE #628: Selecting available accelerators does not check versions of OpenMP
As I understand, in case no specific accelerators are requested via cmake arguments, alpaka tries to use all accelerators available. However, with OpenMP it only checks that OpenMP is found and then both OMP2 and OMP4 accelerators are enabled (or rather not disabled). In case of building for Visual Studio only OpenMP 2 is supported though and so OMP4 accelerators are not suitable. I think it is better to add checks for versions of OpenMP and only enable the supported accelerators. In case that it reasonable, I can provide a PR for that (should be just a couple of lines).

sbastrakov:
Fixes #628. For my system (Win 10, VS 2017) with default options it disables the OpenMP 4 accelerator which has been a problem originally.
[Original cmake output](https://github.com/ComputationalRadiationPhysics/alpaka/files/2349459/output_original.txt), and here is [output after fix](https://github.com/ComputationalRadiationPhysics/alpaka/files/2349460/output_new.txt) ( `ALPAKA_ACC_CPU_BT_OMP4_ENABLED `is no longer set).

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -315,6 +315,12 @@ IF(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR A
         SET(ALPAKA_ACC_CPU_BT_OMP4_ENABLE OFF CACHE BOOL "Enable the OpenMP 4.0 CPU block and thread back-end" FORCE)
 
     ELSE()
+
+        # Check whether OpenMP 4 is supported
+        IF(OpenMP_CXX_VERSION VERSION_LESS 4.0)
+            SET(ALPAKA_ACC_CPU_BT_OMP4_ENABLE OFF CACHE BOOL "Enable the OpenMP 4.0 CPU block and thread back-end" FORCE)
+        ENDIF()
+
         # CUDA requires some special handling
         IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
             SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")


alpakaConfig.cmake
```diff
@@ -315,6 +315,12 @@ IF(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR A
         SET(ALPAKA_ACC_CPU_BT_OMP4_ENABLE OFF CACHE BOOL "Enable the OpenMP 4.0 CPU block and thread back-end" FORCE)
 
     ELSE()
+
+	    # Check whether only OpenMP 2 is supported
+	    IF(OpenMP_CXX_VERSION_MAJOR STREQUAL "2")
+            SET(ALPAKA_ACC_CPU_BT_OMP4_ENABLE OFF CACHE BOOL "Enable the OpenMP 4.0 CPU block and thread back-end" FORCE)
+		ENDIF()
```
BenjaminW3: Please replace the tabs by 4 spaces.

alpakaConfig.cmake
```diff
@@ -315,6 +315,12 @@ IF(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR A
         SET(ALPAKA_ACC_CPU_BT_OMP4_ENABLE OFF CACHE BOOL "Enable the OpenMP 4.0 CPU block and thread back-end" FORCE)
 
     ELSE()
+
+	    # Check whether only OpenMP 2 is supported
+	    IF(OpenMP_CXX_VERSION_MAJOR STREQUAL "2")
```
BenjaminW3: Could you please try if something like `IF(${OpenMP_CXX_VERSION_MAJOR}.${OpenMP_CXX_VERSION_MINOR} VERSION_LESS 4.0)` works? Mabe this needs some " other other tricks.

sbastrakov:
@BenjaminW3 thanks, fixed.

sbastrakov:
Did the check fail because of my changes, or there is some environment problem?

BenjaminW3:
I would hope that they do not exist ;-)

PR #635, opened 2018-09-08
Add compile-time error for atomic CUDA ops which are not available due to sm restrictions

### LINKED ISSUE #614: Add compiler messages when there is no specialization for atomics.
In AtomicCudaBuiltIn.hpp, there are some checks for PTX version:
```C
#if BOOST_ARCH_PTX >= BOOST_VERSION_NUMBER(3, 5, 0)
/// ... AtomicOp ...
#endif
```
There should be an else branch with a `#error` or `static_assert()`, otherwise user might have to search for a while until the reason for a missing type becomes clear. Maybe there are other places as well (this `BOOST_ARCH_PTX` thing only affects atomics code, will be the same for HIP).
I'll check into it later.

BenjaminW3:
Fixes basic case of #614


tdd11235813:
looks good to me, thanks!

PR #650, opened 2018-09-24
CUDA: Math & Public Headers

### LINKED ISSUE #647: CUDA 10 warnings about including an internal header
[ Followup to #455 ]

CUDA 10 now throws this warning:

```
In file included from .../include/alpaka/math/abs/AbsCudaBuiltIn.hpp:37:0,
                 from .../include/alpaka/math/MathCudaBuiltIn.hpp:32,
                 from .../include/alpaka/acc/AccGpuCudaRt.hpp:38,
                 from .../include/alpaka/alpaka.hpp:40,
                 from .../multigrid3d.cpp:11:
.../cuda-10.0.130/include/crt/math_functions.hpp:54:2: warning: #warning "crt/math_functions.hpp is an internal header file and must not be used directly.  Please use cuda_runtime_api.h or cuda_runtime.h instead." [-Wcpp]
 #warning "crt/math_functions.hpp is an internal header file and must not be used directly.  Please use cuda_runtime_api.h or cuda_runtime.h instead."
  ^~~~~~~
```

It looks like this include is not needed at all anymore in CUDA 10. Thus I propose to change it to:

```diff
 #include <type_traits>
-#if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(9, 0, 0)
-    #include <crt/math_functions.hpp>
-#else
+#if BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(9, 0, 0)
     #include <math_functions.hpp>
+#elif BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(10, 0, 0)
+    #include <crt/math_functions.hpp>
 #endif
 ```

Or is there some other preference to remove this warning?

ax3l:
Regression to previous fixes to CUDA headers #455.
The math functions of the CUDA API should be taken from the public "facade" header `<cuda_runtime.h>`

Fix #647

### DIFF
--- a/include/alpaka/math/abs/AbsCudaBuiltIn.hpp
+++ b/include/alpaka/math/abs/AbsCudaBuiltIn.hpp
@@ -32,12 +32,9 @@
 
 #include <alpaka/math/abs/Traits.hpp>
 
+#include <cuda_runtime.h>
 #include <type_traits>
-#if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(9, 0, 0)
-    #include <crt/math_functions.hpp>
-#else
-    #include <math_functions.hpp>
-#endif
+
 
 namespace alpaka
 {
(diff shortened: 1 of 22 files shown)


BenjaminW3:
Thanks! I will try to add CUDA 10 to the CI now that it is officially released.

PR #654, opened 2018-09-25
Add support for CUDA 10.0

BenjaminW3:


.travis.yml
```diff
@@ -154,6 +154,24 @@ matrix:
     - name: nvcc-9.2 + clang-4 Debug
       env: ALPAKA_CI_DOCKER_BASE_IMAGE_NAME=ubuntu:18.04 CXX=clang++ CC=clang ALPAKA_CI_CLANG_VER=4.0.0 CMAKE_BUILD_TYPE=Debug   ALPAKA_CI_BOOST_BRANCH=boost-1.65.1 ALPAKA_CI_CMAKE_VER=3.11.4 ALPAKA_CI_SANITIZERS=      ALPAKA_ACC_GPU_CUDA_ENABLE=ON  ALPAKA_CUDA_VERSION=9.2 ALPAKA_CUDA_COMPILER=nvcc ALPAKA_CUDA_ARCH="30;35"
 
+    ## CUDA 10.0
+    # nvcc + g++
+    - name: nvcc-10.0 + gcc-4.9 Debug
+      env: ALPAKA_CI_DOCKER_BASE_IMAGE_NAME=ubuntu:14.04 CXX=g++     CC=gcc   ALPAKA_CI_GCC_VER=4.9     CMAKE_BUILD_TYPE=Debug   ALPAKA_CI_BOOST_BRANCH=boost-1.67.0 ALPAKA_CI_CMAKE_VER=3.7.2  ALPAKA_CI_SANITIZERS=      ALPAKA_ACC_GPU_CUDA_ENABLE=ON  ALPAKA_CUDA_VERSION=10.0 ALPAKA_CUDA_COMPILER=nvcc ALPAKA_CUDA_ARCH="30;35"
```
psychocoderHPC: IMO it would be better to test the compute architectures `30;75`. Than we are sure that the min and max version compiles.

ax3l:
Just as a side note: initial CUDA 10.0 support just landed in clang trunk: https://github.com/llvm-mirror/clang/commit/252ecba16a1694e520ca44a2cb065e7899786fbf

From the devs notes, it will still be some work until new Turing features (SM 7.5 supports PTX 6.3) are fully supported: https://reviews.llvm.org/D52438 but they are on it.

That's what different `clang -x` versions support so far: https://gist.github.com/ax3l/9489132#clang--x-cuda (in case we want to handle this gracefully in CMake).

.travis.yml
BenjaminW3: done

BenjaminW3:
I will add clang-7.0 to CI after this PR is done. We are reaching a very large test matrix but we have to check every cuda+compiler combination.

PR #657, opened 2018-09-25
backport dev to 0.3.4 part 1

psychocoderHPC:
This PR backports the first fixes from the current dev to the next stable backport branch:

- partly backport Test boost 1.68.0 #606 
  - only backport creation and usage of `core/BoostPredef.hpp`
- backport Fix empty(QueueCpuAsync) returning true even though the last task is still executing #627  
  - this is a backported via copy and past because the name-ings of the file and variables differ between the branch `dev` and `0.3.X`

~~- [x] do not merge this is WIP~~


BenjaminW3:
This should really be the last backport release because the develop branch has largeley diverged from the 0.3.x release line.

psychocoderHPC:
Yes it makes more and more issues to backport fixes. We can plan to create a new major release in the near term.

psychocoderHPC:
You can now review this PR. As I know @BenjaminW3 said that the windows CI is not important. They is failing because they run out of memory.

theZiz:
I just checked the "empty(QueueCpuAsync)" fix with a back-ported asynchronous llama example and with 0.3.3 it showed the same error as it did with the develop branch, which is now fixed with 0.3.4. So the backport seems to work. :+1:

PR #660, opened 2018-09-26
Initialize result vector (vectorAdd). cout -> cerr.

tdd11235813:
- uses cerr instead of cout for results error in examples and tests
- ~~P2P test invokes BOOST_CHECK(true) as long as there was no error~~

### DIFF
--- a/example/reduce/src/reduce.cpp
+++ b/example/reduce/src/reduce.cpp
@@ -165,7 +165,7 @@ int main()
     T expectedResult = static_cast<T>(n / 2 * (n + 1));
     if (result != expectedResult)
     {
-        std::cout << "Results don't match: " << result << " != " << expectedResult
+        std::cerr << "Results don't match: " << result << " != " << expectedResult
                   << "\n";
         return EXIT_FAILURE;
     }
--- a/example/vectorAdd/src/main.cpp
+++ b/example/vectorAdd/src/main.cpp
@@ -127,8 +127,8 @@ auto main()
     BufHost bufHostC(alpaka::mem::buf::alloc<Data, Idx>(devHost, extent));
 
     // Initialize the host input vectors A and B
-    Data * const pBufHostA = alpaka::mem::view::getPtrNative(bufHostA);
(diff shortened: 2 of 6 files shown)


test/unit/mem/p2p/src/P2P.cpp
```diff
@@ -128,6 +127,7 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(
 
     testP2P<TAcc>( extent );
 #endif
+    BOOST_CHECK(true);
```
BenjaminW3: Why did you move this out of the if?

tdd11235813: yes I know and I think it is better style to generally ensure the test state in the end, if all things have passed. 
Either by checking some results, or by just saying: code reached that point -> test passed. And this should not be within a branch (it is kind of academic, I know).
A test case must have a state (failed or succeeded). Boost warns, if no state is set (from [this comment](https://github.com/ComputationalRadiationPhysics/alpaka/issues/617#issue-353343435)): 
```
# happens when there is no check within a test
Test case memP2P/memP2PTest<alpaka__acc__AccGpuCudaRt<std__integral_constant<unsigned_long,_1ul>,_long>> did not check any assertions
```
Of course, Boost just assumes then, that the test has passed successfully. 

If you do not want it, I remove that change. If you think it is completely wrong style, then I also would like to know :)

sbastrakov: I don't think `testP2P ` currently has a check in the normal case.
Also, being picky, why is the unsupported case made as ifdef auto-pass and not disabling the test as it is not applicable (dunno, but guess there is a way to conditionally enable).

tdd11235813: 1) removed that change.
2) First, I disabled the test via cmake, too. However, during the review it was recommended to have such a specific handling at one place in one source file, thus this ifdef.
3) memP2P test should be checked somewhere in the verifyBytesSet as it no more uses BOOST_VERIFY, but a variable (have not checked that code now)

PR #661, opened 2018-09-26
Change ALPAKA_ADD_EXECUTABLE|LIBRARY to CMake macros.

tdd11235813:
This is a requirement for future HIP back-end, as long as there is no solution to an HIP issue, see [here](https://github.com/ROCm-Developer-Tools/HIP/issues/631#issuecomment-416243382).

However, maybe it is a better solution to use a macro anyways? 
A macro does not open an extra scope.

### DIFF
--- a/cmake/addExecutable.cmake
+++ b/cmake/addExecutable.cmake
@@ -22,7 +22,9 @@ CMAKE_MINIMUM_REQUIRED(VERSION 3.7.0)
 
 #------------------------------------------------------------------------------
 # Calls CUDA_ADD_EXECUTABLE or ADD_EXECUTABLE depending on the enabled alpaka accelerators.
-FUNCTION(ALPAKA_ADD_EXECUTABLE In_Name)
+# Using a macro to stay in the scope (fixes lost assignment of linker command in FindHIP.cmake)
+# https://github.com/ROCm-Developer-Tools/HIP/issues/631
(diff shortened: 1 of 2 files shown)


tdd11235813:
yeah macro always sounds evil. Interestingly, cmake modules seem to mostly use macros for add_executable and add_library. Maybe because it prevents that any set command within goes local.

For alpaka+hip the problematic scheme looked like:
- FindHIP (sets a variable foo)
  - `ALPAKA_ADD_EXECUTABLE` (function)
    - `HIP_ADD_EXECUTABLE` (macro)
      - `SET(foo ...)` // `PARENT_SCOPE` makes cmake crazy, when `HIP_ADD_EXECUTABLE` is not embedded in a scope

psychocoderHPC:
I will merge it we should have an eye on Rocm and if they fix it than we should change it back to a function.

ax3l:
> we should have an eye on Rocm

@tdd11235813 can you please report his upstream and link the issue here? Otherwise why should it ever change?

tdd11235813:
it is already reported, I linked it in the first comment ( or [here](https://github.com/ROCm-Developer-Tools/HIP/issues/631) )

ax3l:
Huh, I overlooked that link. Sorry for the noise! :)

tdd11235813:
no problem, I admit the link was quite hidden ;)

PR #662, opened 2018-09-26
Replace assert with ALPAKA_ASSERT

tdd11235813:
- these changes came from HIP(HCC) testing branch
- there, an error occurred, when `__assert_fail` does not have the
right host-device signature, which is included with cassert
- in general, we rely on `core/Assert.hpp` anyways
- there, specific assert handling can be done safely for e.g. HIP,
and does not get interfered afterwards by a cassert

Edit after first review process:
**Replaces assert() with ALPAKA_ASSERT().**

- use ALPAKA_ASSERT() macro instead of assert()
- now, specific assert handling can be done in core/Assert.hpp
- it is disallowed to include cassert somewhere else,
because it can interfere the specified behavior
- at the moment this is only required by the future HIP(HCC) back-end

### DIFF
--- a/include/alpaka/mem/view/ViewSubView.hpp
+++ b/include/alpaka/mem/view/ViewSubView.hpp
@@ -30,10 +30,11 @@
 
 #include <alpaka/mem/view/ViewPlainPtr.hpp>
 #include <alpaka/vec/Vec.hpp>
+
+#include <alpaka/core/Assert.hpp>
 #include <alpaka/core/Common.hpp>
 
 #include <type_traits>
-#include <cassert>
 
 namespace alpaka
 {
@@ -101,7 +102,7 @@ namespace alpaka
                         std::is_same<TDim, dim::Dim<TOffsets>>::value,
                         "The dim type of TOffsets and the TDim template parameter have to be identical!");
 
-                    assert(((m_offsetsElements + m_extentElements) <= extent::getExtentVec(view)).foldrAll(std::logical_and<bool>()));
+                    ALPAKA_ASSERT(((m_offsetsElements + m_extentElements) <= extent::getExtentVec(view)).foldrAll(std::logical_and<bool>()));
                 }
                 //-----------------------------------------------------------------------------
                 //! Constructor.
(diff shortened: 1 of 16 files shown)


include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -31,7 +31,6 @@
 #include <alpaka/core/Assert.hpp>
 #include <alpaka/core/Common.hpp>
 
-#include <cassert>
```
BenjaminW3: assert is used in this file

include/alpaka/mem/view/ViewSubView.hpp
```diff
@@ -33,7 +33,6 @@
 #include <alpaka/core/Common.hpp>
 
 #include <type_traits>
-#include <cassert>
```
BenjaminW3: Here as well. We should not delete it here.

BenjaminW3:
We can only delete the include where it is not used.
Please introduce `ALPAKA_ASSERT()` into `Asssert.hpp` and include it correctly if you need a special version for HIP.

tdd11235813:
ok, although the assert include is provided by Assert.hpp, it is more explicit to have an `ALPAKA_ASSERT()` instead of `assert()` then, I agree. It would be required in all files listed above though.

tdd11235813:
ok, replaced assert with ALPAKA_ASSERT everywhere to be consistent. Not all former cassert include locations will be reached by HIP(HCC), but the assertion usage it should be consistent throughout alpaka. So, to be honest, I do not really like that 'requirement' to always use ALPAKA_ASSERT. What do you think?

tdd11235813:
great :D So, should this ALPAKA_ASSERT be documented somewhere?

BenjaminW3:
I think that it is an implementation detail for now so it does not need any special explanation.

tdd11235813:
k, so it can be merged, when the checks have succeeded.

tdd11235813:
Have not tested, but maybe things like this can be merged into that macro too:
```C++
#ifdef NDEBUG                                                                                                                                                                           
                    alpaka::ignore_unused(lhs);                                                                                                                                         
#else                                                                                                                                                                                   
                    ALPAKA_ASSERT(TLhs::value > lhs);                                                                                                                                   
#endif
```

PR #663, opened 2018-09-27
Adds missing ALPAKA_NO_HOST_ACC_WARNING to DevCpu

tdd11235813:
There were still several CUDA warnings like:
```
/alpaka/include/alpaka/block/shared/st/Traits.hpp(162): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu") is not allowed
```

This PR fixes DevCpu by adding `ALPAKA_NO_HOST_ACC_WARNING` to the used constructors to silent the annoying warnings completely.

Relates to #505, where I first tried to fix the mentioned files instead of DevCpu directly.

### DIFF
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -168,12 +168,16 @@ namespace alpaka
             {}
         public:
             //-----------------------------------------------------------------------------
+            ALPAKA_NO_HOST_ACC_WARNING
             DevCpu(DevCpu const &) = default;
             //-----------------------------------------------------------------------------
+            ALPAKA_NO_HOST_ACC_WARNING
             DevCpu(DevCpu &&) = default;
             //-----------------------------------------------------------------------------
+            ALPAKA_NO_HOST_ACC_WARNING
             auto operator=(DevCpu const &) -> DevCpu & = default;
             //-----------------------------------------------------------------------------
+            ALPAKA_NO_HOST_ACC_WARNING
             auto operator=(DevCpu &&) -> DevCpu & = default;
             //-----------------------------------------------------------------------------
             ALPAKA_FN_HOST auto operator==(DevCpu const &) const


BenjaminW3:
I do not understand why the compiler thinks that `alpaka::dev::DevCpu::DevCpu` is a `__host__ __device__` function.

tdd11235813:
k, I try to provide more information about the stack, where this is called...

tdd11235813:
that's what I get from bufferCopy as example:

> make bufferCopy                                                                                                                              
[  0%] Building NVCC (Device) object example/bufferCopy/CMakeFiles/bufferCopy.dir/src/bufferCopy_generated_bufferCopy.cpp.o
/alpaka/include/alpaka/block/shared/st/Traits.hpp(162): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu") is not allowed

> /alpaka/include/alpaka/block/shared/st/Traits.hpp(162): warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu [subobject]") is not allowed

> [100%] Linking CXX executable bufferCopy
[100%] Built target bufferCopy

sbastrakov:
Didn't @BenjaminW3 wrote [here](https://github.com/ComputationalRadiationPhysics/alpaka/issues/629#issuecomment-418977895) that CUDA treats a `= default `constructor of `DevCpu `as host device? I couldn't quickly confirm that, but if that is the case, then it explains the issue, since this class has a shared pointer member.

tdd11235813:
Maybe that is the point, have not found the conflicting host-device context yet, stopped to look into this.

psychocoderHPC:
Why do we not set all functions explicit to `ALPAKA_FN_HOST` instead of using the potential risky `ALPAKA_NO_HOST_ACC_WARNING` where the behavior is not equal for CUDA and CangCUDA.

reference: https://github.com/ComputationalRadiationPhysics/picongpu/pull/1707#issuecomment-267571760

psychocoderHPC:
It could be that HCC will not be happy about using `ALPAKA_FN_HOST`.

BenjaminW3:
It is nvcc since version 9.0 that complains about `warning : __host__ annotation on a defaulted function("~IdxBtZero") is ignored` see https://stackoverflow.com/questions/46469062/warning-host-annotation-on-a-defaulted-function-is-ignored-why . That is the reason why I removed all those annotation in https://github.com/ComputationalRadiationPhysics/alpaka/pull/384.

tdd11235813:
If feasible this warning can be suppressed, I wanted to propose a PR including this anyways:
```cmake
LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored")
```
As @psychocoderHPC mentioned it, HCC wants matching host/device signatures. If somewhere in the function chain is a 'host' signature, all the functions seem to require it.

ax3l:
Could we alternatively suggest the same (imho useful) behavior for defaulted member functions in CUDA 9+ to HCC?

tdd11235813:
what do you mean exactly? HCC throws errors, which cannot be suppressed.
Wanted to test clang, but compile error occurred, see #671

ax3l:
Just saw your analysis [here](https://github.com/ComputationalRadiationPhysics/alpaka/issues/629#issuecomment-419111418), I think the suggested approach makes sense. See discussion in the issue.

tdd11235813:
ok, in [#668](https://github.com/ComputationalRadiationPhysics/alpaka/pull/668/files) this suppressor is already included for the NVCC part in the cmake configuration. clang should not need it, because it only uses ptxas. Not sure, if clang itself has an issue with it. clang with cuda(clang) does not work for me most of the times, but the [travis tests have passed](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/436170539).

tdd11235813:
the xcudafe flag is for suppressing the warning of "over"-annotated defaulted functions, example:

> warning: __host__ annotation is ignored on a function("~IdxGbRef") that is explicitly defaulted on its first declaration

This PR fixes a different thing, ie., calling a host function within a host-device function triggers that nvcc warning.

> warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu") is not allowed

But, I had another idea or question: Why the pragma behind `ALPAKA_NO_HOST_ACC_WARNING` is not directly integrated into `ALPAKA_FN_HOST_ACC`? I wonder if it works at all, because there are:
```C++
ALPAKA_NO_HOST_ACC_WARNING
template<...>
ALPAKA_FN_HOST_ACC void foo()
```
If it would work, then above changes to DevCpu are probably not required, as the pragma should affect invoked functions as well.

Btw, this PR is not required for HIP to work, it is just about suppressing the warnings from nvcc.

BenjaminW3:
To answer your last question: Yes, the macros have to be separate due to the syntax that is required for templates.

I am still not convinced that the PR is correct. `DevCpu` should be a host class that only calls host functions. There should not be a reason to suppress host<->device calls here.

tdd11235813:
ok thanks. Regarding this PR, that's totally ok, I see the point of your question. I just do not have the time now to dig into that code just for a warning message, so I guess we close this for now.

Issue #399, opened 2017-09-08
definition of atomic operations


psychocoderHPC:
What is the definition of an atomic operation in alpaka. Is a atomic operation only atomic between the same operation type (means `+` and `-` to the same address are not save) or are all atomic operations to an address atomic independent to the operation (access and op on same address are guaranteed thread save).

Case 1: only the same kind of operation is thread save
Case 2: only the same kind and type used within the operation is thread save
Case 3: the access to a memory address is thread save independent of the operation
Case 4: the access to a memory address of the **same type** is thread save independent of the operation

psychocoderHPC:
From the view of an programmer the case 4 would be the best one because sometimes it is useful to mic and atomicAdd and atomicExch.

In the current alpaka implementation `Case 4` is fulfilled expect `AtomicOmpCritSec` where compare and swap is not atomic to all other operations.

psychocoderHPC:
I am currently implementing fast atomic operations for `AtomicOmpCritSec` and found that `atomicMin`, `atomicMax` and `compare and swap can not be implemented to fulfill case 4 :-( without a for loop and than only for fundamental types.
To allow the user to specialize the atomic operation for own type e.g by using `omp critical section` than it can only fulfill case 2.

[Documentation](http://scc.ustc.edu.cn/zlsc/sugon/intel/compiler_c/main_cls/cref_cls/common/cppref_pragma_omp_atomic.htm)

BenjaminW3:
What is the behaviour in CUDA? Can this be tested?

psychocoderHPC:
IMO case 4 is fulfilled for CUDA, I can't found any source which says something other. A test is not so easy because to test if case 4 is fulfilled all positives must be interpreted as false positives (due to the nature of race conditions), only a negative test will is exact.

psychocoderHPC:
If I implement `AtomicOmpCritSec` with `omp atomic capture` is will be faster than the new AtomicStlLock #398 but only supports fundamental types.

btw: CUDA also supports a hand full types.

BenjaminW3:
I would expect CUDA to implement Case 3 but we might never find out.

psychocoderHPC:
Not sure case 3 means that you can mix 64 and 32 bit types because only the address counts.

psychocoderHPC:
Is there anything against the case 4 definition from your side? If not I will implement the new omp atomics as case 4.

BenjaminW3:
[...]

PR #664, opened 2018-09-28
 OpenMP builtin atomics

psychocoderHPC:
Use native OpenMP atomics discussed in #399

- introduce a new class `AtomicOmpBuiltIn` (followed the CUDA naming scheme we already used)
- remove old implementation with pure critical sections
- use `AtomicOmpBuiltIn` within all OpenMP accelerators

A good reference for native supported OpenMP operations is: https://software.intel.com/en-us/node/524509


include/alpaka/atomic/AtomicOmpBuildIn.hpp
```diff
@@ -0,0 +1,326 @@
...
+namespace alpaka
+{
+    namespace atomic
+    {
+        //#############################################################################
+        //! The OpenMP accelerator atomic ops.
+        //
+        //  Atomics can be used in the blocks and threads hierarchy levels.
+        //  Atomics are not guaranteed to be save between devices or grids.
+        class AtomicOmpBuildIn
```
BenjaminW3: Could you please fix the spelling and use `AtomicOmpBuiltIn`?

psychocoderHPC: thx, I fixed it and forced push the changes

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
```diff
@@ -0,0 +1,326 @@
...
+            //-----------------------------------------------------------------------------
+            ALPAKA_FN_HOST AtomicOmpBuiltIn(AtomicOmpBuiltIn const &) = delete;
+            //-----------------------------------------------------------------------------
+            ALPAKA_FN_HOST AtomicOmpBuiltIn(AtomicOmpBuiltIn &&) = delete;
+            //-----------------------------------------------------------------------------
+            ALPAKA_FN_HOST auto operator=(AtomicOmpBuiltIn const &) -> AtomicOmpBuiltIn & = delete;
+            //-----------------------------------------------------------------------------
+            ALPAKA_FN_HOST auto operator=(AtomicOmpBuiltIn &&) -> AtomicOmpBuiltIn & = delete;
+            //-----------------------------------------------------------------------------
+            /*virtual*/ ~AtomicOmpBuiltIn() = default;
```
sbastrakov: Not related to this PR, but to educate myself. What is the point of `/*virtual*/`?

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
```diff
@@ -0,0 +1,326 @@
...
+                ALPAKA_FN_ACC_CUDA_ONLY static auto atomicOp(
+                    atomic::AtomicOmpBuiltIn const &,
+                    T * const addr,
+                    T const & value)
+                -> T
+                {
+                    T old;
+                    auto & ref(*addr);
+                    // atomically update ref, but capture original value in old
+                    #pragma omp atomic capture
```
sbastrakov: `capture` is not available in OpenMP 2.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
```diff
@@ -0,0 +1,326 @@
...
+                    { old = ref; ref -= value; }
```
sbastrakov: Isn't such formatting odd? I mean, i see the logic, but find it not worth it (subjective ofc).

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
```diff
@@ -0,0 +1,326 @@
...
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_ACC_CUDA_ONLY static auto atomicOp(
+                    atomic::AtomicOmpBuiltIn const &,
+                    T * const addr,
+                    T const & value)
+                -> T
+                {
+                    T old;
+                    auto & ref(*addr);
+                    // atomically update ref, but capture original value in old
```
sbastrakov: Not important, but should be *the* original value. Present multiple times.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
psychocoderHPC: offline discussed wit @sbastrakov :+1: 
 - To avoid that the user is calling the destructor of this class if this class is a chield of the accelerator we can make the destructor `protected`
- An other cleaner way suggested by @sbastrakov is let the accelerator not `public` inherit all childs, instead provide a cast operator. With that way the methods of the childs will not be visible in the accelerator.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
psychocoderHPC: I will guard all Spezialisation for OpenMP2. I need to check first since which version capture is supported.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
sbastrakov: Upon thinking a bit more about it, making the destructor `protected` should be just strictly (a little) better than the current way without any need for other changes. But  for such a base class with basically no real 'interface' inherited by accelerators, it is just not practical to do further changes.

Issue #651, opened 2018-09-24
Reuse threads in multi-threaded context? Threadsafety?


tdd11235813:
The idea is to avoid spawning OpenMP `parallel` regions within an already parallelized region, wherein alpaka is executed. 
One current solution on the user side could be to let the threads sleep and to use nested parallelism (do not have the details at the moment). However, the DLR team experienced heavy performance impacts, so this would not be a viable solution.
Another idea would be to reuse the parallel region and to avoid another `parallel` region. 
alpaka first has to check, whether it is in a parallel region by `omp_in_parallel()`. 
This is just a proof-of-concept, coming from the DLR dev team. 

```patch
--- a/thirdparty/alpaka-0.3.0/include/alpaka/exec/ExecCpuOmp2Blocks.hpp
+++ b/thirdparty/alpaka-0.3.0/include/alpaka/exec/ExecCpuOmp2Blocks.hpp
@@ -154,8 +154,28 @@ namespace alpaka

                 // Execute the blocks in parallel.
                 // NOTE: Setting num_threads(number_of_cores) instead of the default thread number does not improve performance.
-                #pragma omp parallel
+                int const inParallel(omp_in_parallel());
+                if (inParallel)
                 {
+                    parallel(boundKernelFnObj, blockSharedMemDynSizeBytes, numBlocksInGrid, gridBlockExtent);
+                }
+                else
+                {
+                    #pragma omp parallel
+                    parallel(boundKernelFnObj, blockSharedMemDynSizeBytes, numBlocksInGrid, gridBlockExtent);
+                }
+
+                // Reset the dynamic thread number setting.
+                ::omp_set_dynamic(ompIsDynamic);
+            }
+
+            TKernelFnObj m_kernelFnObj;
+            std::tuple<TArgs...> m_args;
+private:
+            template<typename FnT, typename BsT, typename GBT>
+            ALPAKA_FN_HOST auto parallel(const FnT& boundKernelFnObj, const BsT& blockSharedMemDynSizeBytes, const TSize& numBlocksInGrid, const GBT& gridBlockExtent) const
+            -> void
+            {
 #if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
                     // The first thread does some debug logging.
                     if(::omp_get_thread_num() == 0)
@@ -194,14 +214,7 @@ namespace alpaka
                         // After a block has been processed, the shared memory has to be deleted.
                         block::shared::st::freeMem(acc);
[...]

PR #666, opened 2018-09-29
Allow to run OpenMP 2 block backend within parallel region

BenjaminW3:
Proposal for #651

This does not change anything for the standard case (except one extra function call per thread).

I do not yet have a full understanding of the requirements. How is the kernel enqued? Is the enqueue already in a parallel region (are there multiple enqueues)? Is it in a single-threaded region? Should we add some single-threaded regions for the setup before the parallel part?

### DIFF
--- a/include/alpaka/exec/ExecCpuOmp2Blocks.hpp
+++ b/include/alpaka/exec/ExecCpuOmp2Blocks.hpp
@@ -148,56 +148,72 @@ namespace alpaka
                     throw std::runtime_error("Only one thread per block allowed in the OpenMP 2.0 block accelerator!");
                 }
 
-                // Force the environment to use the given number of threads.
-                int const ompIsDynamic(::omp_get_dynamic());
-                ::omp_set_dynamic(0);
-
-                // Execute the blocks in parallel.
-                // NOTE: Setting num_threads(number_of_cores) instead of the default thread number does not improve performance.
-                #pragma omp parallel
+                if(::omp_in_parallel() != 0)
                 {
-#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
-                    // The first thread does some debug logging.
-                    if(::omp_get_thread_num() == 0)
-                    {
-                        int const numThreads(::omp_get_num_threads());
-                        std::cout << BOOST_CURRENT_FUNCTION << " omp_get_num_threads: " << numThreads << std::endl;
-                    }
-#endif
-                    acc::AccCpuOmp2Blocks<TDim, TIdx> acc(
-                        *static_cast<workdiv::WorkDivMembers<TDim, TIdx> const *>(this),
-                        blockSharedMemDynSizeBytes);
-
-                    // NOTE: schedule(static) does not improve performance.
-#if _OPENMP < 200805    // For OpenMP < 3.0 you have to declare the loop index (a signed integer) outside of the loop header.
-                    std::intmax_t iNumBlocksInGrid(static_cast<std::intmax_t>(numBlocksInGrid));
-                    std::intmax_t i;
-                    #pragma omp for nowait schedule(guided)
-                    for(i = 0; i < iNumBlocksInGrid; ++i)
-#else
-                    #pragma omp for nowait schedule(guided)
-                    for(TIdx i = 0; i < numBlocksInGrid; ++i)
-#endif
-                    {
-                        acc.m_gridBlockIdx =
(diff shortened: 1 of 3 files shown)


psychocoderHPC:
>  Is the enqueue already in a parallel region (are there multiple enqueues)?

The alpaka is used within an other library which already spawned a parallel region.

> How is the kernel enqued?

In the current case we discuss one of the threads (mostly the omp master) is calling alpaka kernel.

>  Is it in a single-threaded region?

Not in the DLR use case (as I know)

> Should we add some single-threaded regions for the setup before the parallel part?

What do you mean with that? Could you please explain it, maybe on an example.

include/alpaka/exec/ExecCpuOmp2Blocks.hpp
```diff
@@ -148,56 +148,72 @@ namespace alpaka
...
                 }
 
-                // Force the environment to use the given number of threads.
-                int const ompIsDynamic(::omp_get_dynamic());
-                ::omp_set_dynamic(0);
-
-                // Execute the blocks in parallel.
-                // NOTE: Setting num_threads(number_of_cores) instead of the default thread number does not improve performance.
-                #pragma omp parallel
+                if (omp_in_parallel() != 0)
```
theZiz: Until now you always out `::` in front of global namespace function from OpenMP. I think this should be kept.

include/alpaka/exec/ExecCpuOmp2Threads.hpp
```diff
@@ -148,6 +148,12 @@ namespace alpaka
                 TIdx const blockThreadCount(blockThreadExtent.prod());
                 int const iBlockThreadCount(static_cast<int>(blockThreadCount));
                 alpaka::ignore_unused(iBlockThreadCount);
+
+                if (omp_in_parallel() != 0)
```
theZiz: Same with `::`

include/alpaka/exec/ExecCpuOmp4.hpp
```diff
@@ -150,6 +150,11 @@ namespace alpaka
                 auto const maxTeamCount(maxOmpThreadCount/static_cast<int>(blockThreadCount));
                 auto const teamCount(std::min(maxTeamCount, static_cast<int>(gridBlockCount)));
 
+                if (omp_in_parallel() != 0)
```
theZiz: Same with `::`

include/alpaka/exec/ExecCpuOmp2Blocks.hpp
```diff
@@ -148,56 +148,72 @@ namespace alpaka
...
+                        blockSharedMemDynSizeBytes,
+                        numBlocksInGrid,
+                        gridBlockExtent);
+                }
+            }
+
+            template<
+                typename FnObj>
+            ALPAKA_FN_HOST auto parallelFn(
+                const FnObj& boundKernelFnObj,
```
theZiz: Shouldn't be the `const on the right side of the type as for the fourth parameter?

BenjaminW3: Yes, I will fix this. Simply copied it from the proposal.

include/alpaka/exec/ExecCpuOmp2Blocks.hpp
BenjaminW3: Will do so

BenjaminW3:
@theZiz I have incorporated your comments.

BenjaminW3:
@psychocoderHPC 
>>Should we add some single-threaded regions for the setup before the parallel part?

>What do you mean with that? Could you please explain it, maybe on an example.

[The first part of the function executing the kernel](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/exec/ExecCpuOmp2Blocks.hpp#L99-L150), where the shared memory size is computed and other things are done should only be executed by a single thread. It may be working now because those functions o not have side-effects but it does not feel correct to me. We may want to add this to a single-threaded region.

psychocoderHPC:
> The first part of the function executing the kernel, where the shared ...

NO this is not allowed to be in the run in one thread. If more than one thread in a parallel region calling a kernel than each of them is Independent and can be called with different sizes for the parameter and therefore the shared memory amount can be different for each kernel call instance.
If you mean we should move the shared memory calculation into a single-thread region to avoid race conditions than I would say there is not need for that.

Issue #655, opened 2018-09-25
ViewPlainPtr doesn't work as promised


krzikalla:
ViewPlainPtr has a 

`auto operator=(ViewPlainPtr &&) -> ViewPlainPtr & = default;`

However, since there are some const members in ViewPlainPtr , this operator will never work. 
Also, why is 

`auto operator=(ViewPlainPtr const &) -> ViewPlainPtr & = delete;`

Compared with the other line I don't see a reason for the deletion here.

BenjaminW3:
Let's first think about the constructors.
* copy constructor: `ViewPlainPtr(ViewPlainPtr const &) = delete` is wrong. You should be able to create a copy of the View. I will fix this.
* move constructor: There is nothing to move in this class, the members are all simple types which have to be copied anyway. If we want to keep the const members, we could fix this constructor by implementing it by hand doing copies on the members.

I am still undecided about the assignment operators. I clearly see the benefit of const members but I also see why you would want to assign the views.

BenjaminW3:
Both constructors (copy and move) now work as expected. As written above I will have a look at the assignment.

PR #667, opened 2018-09-30
Add test for ViewPlainPtr

BenjaminW3:
This is a preparation for #655 and does not yet fix or change anything.

### DIFF
--- a/test/unit/mem/view/src/ViewPlainPtrTest.cpp
+++ b/test/unit/mem/view/src/ViewPlainPtrTest.cpp
@@ -1,0 +1,228 @@
+/**
+ * \file
+ * Copyright 2018 Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Lesser General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * alpaka is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Lesser General Public License for more details.
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with alpaka.
+ * If not, see <http://www.gnu.org/licenses/>.
+ */
+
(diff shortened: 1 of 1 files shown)


BenjaminW3:
push!

test/unit/mem/view/src/ViewPlainPtrTest.cpp
```diff
@@ -0,0 +1,228 @@
...
+namespace alpaka
+{
+namespace test
+{
+namespace mem
+{
+namespace view
+{
+    //#############################################################################
+    //! 1D: sizeof(TIdx) * (5)
```
ax3l: outdated comments?

test/unit/mem/view/src/ViewPlainPtrTest.cpp
```diff
@@ -0,0 +1,228 @@
...
+    //! 2D: sizeof(TIdx) * (5, 4)
+    //! 3D: sizeof(TIdx) * (5, 4, 3)
+    //! 4D: sizeof(TIdx) * (5, 4, 3, 2)
+    // We have to be careful with the extents used.
+    // When Idx is a 8 bit signed integer and Dim is 4, the extent is extremely limited.
```
ax3l: it's `size_t` now (and the name is `Tidx` not `Idx`?)

test/unit/mem/view/src/ViewPlainPtrTest.cpp
```diff
@@ -0,0 +1,228 @@
...
+    //#############################################################################
+    template<
+        std::size_t Tidx>
```
ax3l: Shouldn't this be spelled `TIdx`?

test/unit/mem/view/src/ViewPlainPtrTest.cpp
BenjaminW3: You are right, this is outdated but not only here but in all such places. I will create a follow up PR to fix this everywhere.

test/unit/mem/view/src/ViewPlainPtrTest.cpp
BenjaminW3: Here as well...

test/unit/mem/view/src/ViewPlainPtrTest.cpp
BenjaminW3: I will move this whole helper struct into a common place for all tests.

PR #668, opened 2018-10-02
Removes or changes certain host(-device) annotations

tdd11235813:
- suppresses warnings about ignored annotation on NVCC, e.g.
`warning : __host__ annotation on a defaulted function("~IdxBtZero") is ignored`
- removes host annotations from certain constructors/destructors
- adds certain missing host-device annotations
- these changes came from HIP(HCC) testing branch, where HCC throws an error, if signatures do not match
- only targets code, which is reachable by the future HIP(HCC) back-end (with all other back-ends being disabled)

Notes: 
- for the sake of better consistency I tried to remove all `ALPAKA_FN_HOST`, but it breaks HIP(HCC) again due to signature issues
- so it is likely, that there are methods, where `ALPAKA_FN_HOST` can be removed without breaking anything
- it is also likely, that there are methods, where `ALPAKA_FN_HOST` or `ALPAKA_FN_ACC` or `ALPAKA_FN_HOST_ACC` are currently missing, because these methods did not show up in the tests
  - the changes in this PR are coming from the iterative process making the HC compiler happy and by watching on some obvious functions. If one knows all possible function chains in alpaka, then he can apply the HCC rules on signature matching.


include/alpaka/mem/view/ViewPlainPtr.hpp
```diff
@@ -90,7 +89,8 @@ namespace alpaka
                 //-----------------------------------------------------------------------------
                 auto operator=(ViewPlainPtr &&) -> ViewPlainPtr & = default;
                 //-----------------------------------------------------------------------------
-                ~ViewPlainPtr() = default;
+                ALPAKA_NO_HOST_ACC_WARNING
+                ALPAKA_FN_HOST_ACC ~ViewPlainPtr() = default;
```
BenjaminW3: This is somehow inconsistent now. removed on the constructor, adde on the destructor.

include/alpaka/mem/view/ViewPlainPtr.hpp
```diff
@@ -313,10 +313,11 @@ namespace alpaka
                     dev::DevCudaRt>
                 {
                     //-----------------------------------------------------------------------------
+                    ALPAKA_NO_HOST_ACC_WARNING
```
BenjaminW3: Shouldn't it be necessary on the method in line 288 as well?

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -97,7 +97,7 @@ namespace alpaka
                 return *this;
             }
             //-----------------------------------------------------------------------------
-            /*virtual*/ ~WorkDivMembers() = default;
+            /*virtual*/ ALPAKA_FN_HOST_ACC ~WorkDivMembers() = default;
```
BenjaminW3: Missing on the constructor?

include/alpaka/mem/view/ViewPlainPtr.hpp
tdd11235813: weird, it should be on the ctor as well, why did I remove it? :/

include/alpaka/mem/view/ViewPlainPtr.hpp
tdd11235813: compiler did not complain, but from the code it should be there on 288 as you said.

include/alpaka/vec/Vec.hpp
```diff
@@ -256,7 +256,7 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             auto operator=(Vec &&) -> Vec & = default;
             //-----------------------------------------------------------------------------
-            ~Vec() = default;
+            ALPAKA_FN_HOST_ACC ~Vec() = default;
```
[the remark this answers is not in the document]
tdd11235813: do we have a test that uses one of these unlabeled copy/... ctors? The Vec class constructor and destructor have the correct annotation, where the latter one must unify all the used annotations of the ctors. If copy constructor does not have any signature, I think HCC is able to derive it. It would be however a problem, when a copy ctor is flagged as host-only, while it is used within a host-device function, or when the class constructor is device-only. 
I can make it looking more consistent, but it is still not consistent throughout the whole alpaka code. Because in almost all cases I only touched those lines, HCC was complaining about. And I tested the changes back on the HIP branch on Intel and AMD, so I wonder why the ViewPlainPtr passed. 
Well, I will do another iteration on this next week. Would you like to see the consistency on only these classes above, or also on other classes as well, which is not required for HIP(HCC) to work?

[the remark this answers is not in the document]
BenjaminW3: For now it is enough if those classes you touched are consistent.

tdd11235813:
rebased and fixed some stuff. There were some flags missing which have been extracted from the HIP branch now.

include/alpaka/mem/view/ViewPlainPtr.hpp
```diff
@@ -82,14 +82,24 @@ namespace alpaka
...
+                ALPAKA_NO_HOST_ACC_WARNING
+                ALPAKA_FN_HOST_ACC
                 auto operator=(ViewPlainPtr const &) -> ViewPlainPtr & = delete;
                 //-----------------------------------------------------------------------------
+                ALPAKA_NO_HOST_ACC_WARNING
+                ALPAKA_FN_HOST_ACC
                 auto operator=(ViewPlainPtr &&) -> ViewPlainPtr & = default;
                 //-----------------------------------------------------------------------------
+                ALPAKA_NO_HOST_ACC_WARNING
+                ALPAKA_FN_HOST_ACC
```
BenjaminW3: Who is using the `ViewPlainPtr` in device code and where? Until now my impression was that the views/bufs are a host only thing (for now) and we pass raw pointers+size into the kernels. You could change this while resolving the merge conflicts here.

tdd11235813: ok, I do not know, if this is needed on device at all, but I would follow your thoughts and change the annotation to host only. Tested it already and everything passed.

alpakaConfig.cmake
```diff
@@ -548,6 +548,8 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 IF(ALPAKA_CUDA_SHOW_REGISTER)
                     LIST(APPEND CUDA_NVCC_FLAGS "-Xptxas=-v")
                 ENDIF()
+                # avoids warnings on host-device signatured, default constructors/destructors
+                LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored")
```
BenjaminW3: You have to move this into `IF(CUDA_VERSION VERSION_GREATER_EQUAL 9.0)`.

PR #679, opened 2018-10-10
Allow to copy and move construct ViewPlainPtr

BenjaminW3:
This is a part of #655. I will think about the assignment operators separately.

[edit from psychocoder]
PR adds a test for copy and move constructor

### DIFF
--- a/include/alpaka/mem/view/ViewPlainPtr.hpp
+++ b/include/alpaka/mem/view/ViewPlainPtr.hpp
@@ -82,13 +82,19 @@ namespace alpaka
                 {}
 
                 //-----------------------------------------------------------------------------
-                ViewPlainPtr(ViewPlainPtr const &) = delete;
+                ViewPlainPtr(ViewPlainPtr const &) = default;
                 //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


sbastrakov:
I'm probably missing smth, but why the body of the move constructor is required in this case? It looks to me as it just repeats what the default copy constructor would do, and so acts exactly the same as the deleted move constructor would.

BenjaminW3:
Yes, it does exactly the same as the copy-constructor. Moving of this class simply means to copy the pointer and the other members. When the move-constructor is deleted, we would not be able to move the class at all but there is no reason to disallow it, even though it does not have any advantages. The compiler is not able to generate this default move-constructor due to the const members.

sbastrakov:
I do not think it is correct. std::move is essentially a glorified typecast. So deleted move constructor results only in copy constructor being called instead of move, which is the desired behavior here. [Example, bottom of page, class C](https://en.cppreference.com/w/cpp/language/move_constructor)
My point is, there is nothing wrong with the code currently, but the same effect could have been achieved easier and potentially better with respect to future changes to this class.

PR #682, opened 2018-10-10
Added logo

theZiz:
See discussion at https://github.com/ComputationalRadiationPhysics/alpaka/issues/656


theZiz:
> The `doc/images/alpaka_inkscape.svg` looks wrong when being shown by github in c
> chrome on windows.

Yes, it looks wrong as thumbnail in Dolphin, too. First of all it uses a font, if you don't have the font, it look wrong. Secondly Inkscape add some non standard annotations. This is the reason why there is also an `alpaka.svg` which transformed the text to a path. :wink:

psychocoderHPC:
I think about backporting this PR to 0.3.4

ax3l:
nah, no need really ;-)

Issue #673, opened 2018-10-03
Clang still fails


tdd11235813:
In #659 helloWorld was not compiling, leading to the fix in #665 from @BenjaminW3 .
After fixing #671 there are still lots of errors for all other tests. I wonder why clang7 is working in travis environment.

- error: exception specification of overriding function is more lax than base version
  - 'example/bufferCopy/CMakeFiles/__bufferCopy__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'example/reduce/CMakeFiles/__reduce__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'example/vectorAdd/CMakeFiles/__vectorAdd__/all' failed
- error: linker command failed with exit code 1 (use -v to see invocation)
  - 'test/analysis/headerCheck/CMakeFiles/__headerCheck__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'test/integ/axpy/CMakeFiles/__axpy__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'test/integ/mandelbrot/CMakeFiles/__mandelbrot__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'test/integ/matMul/CMakeFiles/__matMul__/all' failed
- error: exception specification of overriding function is more lax than base version
- error: declaration requires an exit-time destructor [-Werror,-Wexit-time-destructors]
  - 'test/integ/sharedMem/CMakeFiles/__sharedMem__/all' failed
- error: linker command failed with exit code 1 (use -v to see invocation)
  - 'test/unit/acc/CMakeFiles/__acc__/all' failed
- error: exception specification of overriding function is more lax than base version
- error: declaration requires an exit-time destructor [-Werror,-Wexit-time-destructors]
- error: exception specification of overriding function is more lax than base version
  - 'test/unit/atomic/CMakeFiles/__atomic__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'test/unit/block/shared/CMakeFiles/__blockShared__/all' failed
- error: exception specification of overriding function is more lax than base version
  - 'test/unit/block/sync/CMakeFiles/__blockSync__/all' failed
- error: linker command failed with exit code 1 (use -v to see invocation)
  - 'test/unit/core/CMakeFiles/__core__/all' failed
- error: linker command failed with exit code 1 (use -v to see invocation)
  - 'test/unit/event/CMakeFiles/__event__/all' failed
[...]

PR #683, opened 2018-10-11
Suppress warning in clang about exit time destructor

BenjaminW3:
Fix exit-time-destructor warning mentioned in #673.

### DIFF
--- a/include/alpaka/atomic/AtomicStdLibLock.hpp
+++ b/include/alpaka/atomic/AtomicStdLibLock.hpp
@@ -23,6 +23,8 @@
 
 #include <alpaka/atomic/Traits.hpp>
 
+#include <alpaka/core/BoostPredef.hpp>
+
 #include <mutex>
 #include <array>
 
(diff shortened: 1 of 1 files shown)


ax3l:
hm... but the warning is actually true. That's not a fix, but a disabling of the diagnostics.

BenjaminW3:
Yes, the warning is true but we explicitly wanted this to happen in that way. I remember the PR and the discussion with @psychocoderHPC where he introduced the change.

include/alpaka/atomic/AtomicStdLibLock.hpp
```diff
@@ -23,6 +23,8 @@
 
 #include <alpaka/atomic/Traits.hpp>
 
+#include <alpaka/core/BoostPredef.h>
```
tdd11235813: suffix to hpp: `<alpaka/core/BoostPredef.hpp>`

BenjaminW3: done

PR #684, opened 2018-10-16
0.3.4: Support for CUDA 10 and bug fixes

BenjaminW3:
Compatibility Changes:
* added support for boost-1.68.0
* added support for CUDA 10
* support for glibc < 2.18 (fix missing macros)
* added checks for available OpenMP versions

Bug Fixes:
* fixed `empty(StreamCpuAsync)` returning true even though the last task is still in progress
* fixed integer overflows in case of `int16_t` being used as accelerator index type
* made some throwing destructors not throwing to support clang 7
* fixed broken `alpaka::math::min` for non-integral types

New Features:
* added `prepareForAsyncCopy` which can be called to enable async copies for a specific buffer (if it is supported)
* allowed to run alpaka OpenMP 2 block accelerated kernels within existing parallel region
* added `alpaka::ignore_unused` which can be used in kernels


ax3l:
during release, `include/version.hpp` was not updated to `0.3.4`

PR #688, opened 2018-10-27
Travis update

BenjaminW3:
* Remove sudo:required because this is the default now and the non-sudo environment will be removed soon: https://changelog.travis-ci.com/linux-builds-run-on-vms-by-default-77106
* Update CMake version


BenjaminW3:
push

PR #695, opened 2018-11-11
Check that the kernel return type is void

### LINKED ISSUE #692: Make sure that the kernel function returns void with `ALPAKA_ACC_GPU_CUDA_ONLY_MODE`
We should check at compile time that the `TKernelFnObj` passed into the executor returns void.
There is no way to handle functions returning a value because each thread produces one.

BenjaminW3:
Addresses #692 partially

### DIFF
--- a/include/alpaka/kernel/Traits.hpp
+++ b/include/alpaka/kernel/Traits.hpp
@@ -184,6 +184,12 @@ namespace alpaka
                 args...))
 #endif
         {
+                // FIXME: For unknown reasons std::result_of fails with __device__ lambdas.
+#if !defined(ALPAKA_ACC_GPU_CUDA_ONLY_MODE)
+            static_assert(
+                std::is_same<typename std::result_of<TKernelFnObj(TAcc const &, TArgs const & ...)>::type, void>::value,
+                "The TKernelFnObj is required to return void!");
+#endif
             static_assert(
                 dim::Dim<typename std::decay<TWorkDiv>::type>::value == dim::Dim<TAcc>::value,
                 "The dimensions of TAcc and TWorkDiv have to be identical!");
--- a/doc/markdown/user/implementation/library/Usage.md
+++ b/doc/markdown/user/implementation/library/Usage.md
@@ -37,7 +37,7 @@ struct MyKernel
         typename TAcc>    // Templated on the accelerator type.
     ALPAKA_FN_ACC       // Macro marking the function to be executable on all accelerators.
   auto operator()(    // The function / kernel to execute.
-        TAcc & acc/*,     // The specific accelerator implementation.
+        TAcc const & acc/*,     // The specific accelerator implementation.
         ...*/) const      // Must be 'const'.
     -> void
     {
--- a/example/helloWorldLambda/src/helloWorldLambda.cpp
+++ b/example/helloWorldLambda/src/helloWorldLambda.cpp
@@ -35,7 +35,7 @@
 template<
     typename TAcc>
 void ALPAKA_FN_ACC hiWorldFunction(
-    TAcc& acc,
+    TAcc const & acc,
     size_t const nExclamationMarks)
 {
     using Dim = alpaka::dim::Dim<TAcc>;
@@ -115,7 +115,7 @@ auto main()
     alpaka::kernel::exec<Acc>(
         queue,
         workDiv,
-        [] ALPAKA_FN_ACC (Acc & acc, size_t const nExclamationMarksAsArg) -> void {
+        [] ALPAKA_FN_ACC (Acc const & acc, size_t const nExclamationMarksAsArg) -> void {
             auto globalThreadIdx    = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc);
             auto globalThreadExtent = alpaka::workdiv::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
             auto linearizedGlobalThreadIdx = alpaka::idx::mapIdx<1u>(globalThreadIdx, globalThreadExtent);
@@ -153,7 +153,7 @@ auto main()
     alpaka::kernel::exec<Acc> (
         queue,
         workDiv,
-        std::function<void(Acc&, size_t)>( hiWorldFunction<Acc> ),
+        std::function<void(Acc const &, size_t)>( hiWorldFunction<Acc> ),
         nExclamationMarks);
 
     


include/alpaka/kernel/Traits.hpp
```diff
@@ -184,6 +184,12 @@ namespace alpaka
                 args...))
 #endif
         {
+            static_assert(
+                std::is_same<
+                    typename std::result_of<
+                        TKernelFnObj(const TAcc &, const TArgs & ...)>::type,
+                    void>::value,
```
psychocoderHPC: The indention looks somehow strange

BenjaminW3:
I got some strange errors for this branch in CUDA_ONLY_MODE. If someone could have a look at the code/CI logs and has an idea, I would be thankful.

sbastrakov:
I've spend some time looking into that build issue, but currently am also puzzled.

BenjaminW3:
Thanks for looking at it! Maybe someone else has an idea.
It looks likt the CUDA compiler uses some strange type for `__device__` only lambdas which can not be correctly used in this way.

psychocoderHPC:
We need to check if we can reproduce this behavior in a mini app without alpaka. Than we can open an bug on the NVIDIA website. Maybe it is not supported and I missed it in the docu or it is a real bug.

Am 18. November 2018 07:49:24 MEZ schrieb Benjamin Worpitz <notifications@github.com>:

BenjaminW3:
After thinking about it, the error may not be wrong at all. In the code I am testing what the return type of the given function would be when called with the accelerator and the parameters. However, it is not possible to call a `__device__` function in this place because the current function is a `__host__` only function.
This is only a wild guess and we might have to test this and I do not think that static_asserts should take into account the function attributes of the current function but this may be related to the underlying problem.

BenjaminW3:
I have given up for now. It works in all build jobs except when `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is enabled. I would like to get this merged because it will still find incorrect uses in almost all cases.
The issue should not be closed and I will update it after this has been merged.

ax3l:
@BenjaminW3 can you reproduce the device side lambda issue with a minimal example?
We can submit it to the nvcc bug tracker if you like.

BenjaminW3:
I would hope that we can reproduce it with something like this but I have not yet tried:
```C++
#include <cstdio>
#include <type_traits>

template<typename T>
void doSomething(const T& t)
{
    static_assert(std::is_same<typename std::result_of<T()>::type, void>::value, "Failed");
    // t();
}

int main()
{
    const auto lambda0 = [](){printf("%s", __func__);};
    doSomething(lambda0);
#ifdef __CUDACC__
    const auto lambda1 = [] __host__ () {printf("%s", __func__);};
    doSomething(lambda1);
    const auto lambda2 = [] __host__ __device__ () {printf("%s", __func__);};
    doSomething(lambda2);
    const auto lambda3 = [] __device__ () {printf("%s", __func__);};
    doSomething(lambda3);
#endif
}
```

ax3l:
@BenjaminW3 getting this:
```bash
$ nvcc -std=c++11 -arch=sm_35 main.cu  # with and without --expt-extended-lambda
main.cu(16): error: The operator() function for a lambda cannot be explicitly annotated with execution space annotations (__host__/__device__/__global__), the annotations are derived from its closure class

main.cu(18): warning: a __host__ function("lambda []()->auto::operator()") redeclared with __host__ __device__, hence treated as a __host__ __device__ function

main.cu(18): error: The operator() function for a lambda cannot be explicitly annotated with execution space annotations (__host__/__device__/__global__), the annotations are derived from its closure class

main.cu(20): warning: a __host__ function("lambda []()->auto::operator()") redeclared with __device__, hence treated as a __host__ __device__ function

main.cu(20): error: The operator() function for a lambda cannot be explicitly annotated with execution space annotations (__host__/__device__/__global__), the annotations are derived from its closure class

3 errors detected in the compilation of "/tmp/tmpxft_0000152b_00000000-9_main.cpp1.ii".

$ nvcc --version
Copyright (c) 2005-2016 NVIDIA Corporation
Built on Sun_Sep__4_22:14:01_CDT_2016
Cuda compilation tools, release 8.0, V8.0.44
```

BenjaminW3:
I have never tried to compile this code. Please move the attributes in between the `[]` and `()`. This should hopefully be supported by nvcc.

BenjaminW3:
We may still get errors when trying to execute the given lambda. So you might have to remove the `t();` line.

ax3l:
Yep, here we go:
```bash
$ nvcc -std=c++11 -arch=sm_35 --expt-extended-lambda main.cu
main.cu: In instantiation of ‘void doSomething(const T&) [with T = __nv_dl_wrapper_t<__nv_dl_tag<int (*)(), main, 2u> >]’:
main.cu:21:20:   required from here
main.cu:7:1: error: static assertion failed: Failed
     static_assert(std::is_same<typename std::result_of<T()>::type, void>::value, "Failed");
 ^
```

From a similar case in two-year old [clang](https://bugs.llvm.org/show_bug.cgi?id=30567) that the Kokkos team reported, it looks like the scope of definition is the one determining the implicit attributes? (`main()` is host): https://reviews.llvm.org/D25105 (`cfe/trunk/test/SemaCUDA/implicit-device-lambda.cu`) There they place the attributes the same (edited your example to reflect).

Back to the problem: it's in the `__device__`-only lambda:
```C++
    doSomething(lambda3);
```

And I see it in both NVCC 8.0.44 and 10.0.130

Reported as Nvidia Bug no. 2460363.

ax3l:
Feedback:
> This is a documented limitation of extended `__device__` lambdas. The signature of the `operator()` function for an extended `__device__` lambdas may not have the expected form during host compilation, because the device code inside its body is removed when the program is compiled by the host compiler, and the lambda is replaced with a placeholder type instance.
From https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#extended-lambda-restrictions, see 12:
>
> As described above, the CUDA compiler replaces a `__device__` extended lambda defined in a host function with a placeholder type defined in namespace scope. This placeholder type does not define a `operator()` function equivalent to the original lambda declaration. An attempt to determine the return type or parameter types of the `operator()` function may therefore work incorrectly in host code, as the code processed by the host compiler will be semantically different than the input code processed by the CUDA compiler. However, it is ok to introspect the return type or parameter types of the `operator()` function within device code. Note that this restriction does not apply to `__host__ __device__` extended lambdas.
>
> Example:
...This restriction does not apply to extended `__host__ __device__` lambda. Please consider using `__host__ __device__` lambda as a workaround.

But they could still implement e.g. a `result_of` trait for it, imho :)

More feedback:
> It’s hard to resolve this issue because need to remove the body of the __device__ lambda in the code sent to the host compiler, because the host compiler may error out on GPU specific idioms e.g. inline asm. But this also means that return type of the device lambda changes to ‘void’ (which is why introspecting on the return type e.g. with std::result_of is not supported).

PR #698, opened 2018-11-14
backport to 0.3.5 chunk 1

ax3l:
Contains:

- OpenMP: if 3.1+ is available, use builtin atomics #664
- Travis update #688

Investigating if we can simply fix this severe performance regression in https://github.com/ComputationalRadiationPhysics/picongpu/issues/2802 - Works!! :sparkling_heart:


BenjaminW3:
I can see that this helps to improve the performance but why should this have been a regression?

BenjaminW3:
I will merge it even though there is a red build. It is the same builds that already failed when releasing 0.3.4. I tried a rerun but it still took too long and was killed.

ax3l:
It kind of renders production run in PIConGPU not usable, that's why I marked it a PIConGPU (performance) regression. Luckily, it's easy to fix with @psychocoderHPC recent upgrades.

I do not have plans to urgently backport more, so if you like we can make this a release already.

BenjaminW3:
I will release it.
I would not intend to do any more bugfix releases on this branch. Instead I have opened a project to create a 0.4.0 release.

ax3l:
Just updated my comment: yes we did in parallel start to upgrade to Alpaka `dev`: in PIConGPU and cupla: https://github.com/ComputationalRadiationPhysics/picongpu/pull/2807

I still keep defects documented in a 0.3.6 project, so we have a known list of warnings/performance regressions/defects for what we currently use in production.

ax3l:
during release, `include/version.hpp` was not updated to `0.3.5` (still at `0.3.3`)

PR #700, opened 2018-11-20
Use libc++ instead of libstdc++ for clang builds

### LINKED ISSUE #696: Use libc++ instead of libstdc++ when building with clang.
By using the gcc standard library, we can not use features which are only available within clang (e.g. coroutines). This may also be the reason why we have seen issues e.g. in #673 which have not been found in CI.

BenjaminW3:
Fixes #696 by using libstdc++ for all clang builds instead of libstdc++.

I think we should still support building with libstdc++. Adding all builds twice would enlarge the build matrix very much. Therefore I have randomly selected the stdlib for the clang jobs as is already done for boost versions, debug/release builds and many other things.

Ubuntu 14.04 does not seem to be compatible with libc++ combined with clang. I have added an error into the CI scripts for this.

We allow the user to select the stdlib. This means that CMake does not know the stdlib that will be used and therefore we can not warn or error on invalid combinations. The only solution for this problem would be https://stackoverflow.com/questions/42593022/cmake-detect-which-library-libc-or-libstdc-is-configured-to-be-used-against but I do not like it. This should be handled separately from this PR.


ax3l:
What you could do in CMake is the following: add an alpaka switch for the stdlib and use a [find script](https://github.com/meta-toolkit/meta-cmake/blob/master/FindLIBCXX.cmake).

include/alpaka/exec/ExecGpuCudaRt.hpp
```diff
@@ -145,7 +145,8 @@ namespace alpaka
         public:
 // gcc-4.9 libstdc++ does not support std::is_trivially_copyable.
 // MSVC std::is_trivially_copyable seems to be buggy (last tested at 15.7).
-#if (!BOOST_COMP_MSVC) && !(defined(__GLIBCXX__) && (__GLIBCXX__))
+// libc++ in combination with CUDA does not seem to work.
+#if (!BOOST_COMP_MSVC) && !(defined(__GLIBCXX__) && (__GLIBCXX__)) && !(defined(_LIBCPP_VERSION) && BOOST_LANG_CUDA)
```
tdd11235813, line 149: just a note: `BOOST_LANG_CUDA` is always true in this line, otherwise ExecGpuCudaRt.hpp would have thrown a compile error in the beginning (see line 28).

I guess this check would be required in HIP too, when it runs on NVCC platform .. just check against `__HIP_PLATFORM_NVCC__` in the line above, [also see HIP porting guide](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary).
Travis CI tests compiling HIP+NVCC, i.e., HIP(nvcc8.0) + gcc-4.9 Debug.

BenjaminW3: I will have a look at this in #703.

BenjaminW3:
Ready to be merged!

psychocoderHPC (approved):

Issue #701, opened 2018-11-23
Alpaka Does not Detect Power8 Linux


ax3l:
Compiling PIConGPU 0.4.2 for an OpenPOWER8 cluster in https://github.com/ComputationalRadiationPhysics/picongpu/pull/2821 throws that the OS cannot be detected.

Throws `#error` pragma in [dev/cpu/SysInfo.hpp](https://github.com/ComputationalRadiationPhysics/alpaka/blob/0.3.5/include/alpaka/dev/cpu/SysInfo.hpp#L233)

### Software Versions

Alpaka 0.3.5
Boost 1.68.0: can **NOT** verify an underlying error in https://github.com/boostorg/predef/issues/90
PIConGPU 0.4.2
CUDA 9.2.88
CMake 3.11.3
GCC 6.4.0

### System Info

[D.A.V.I.D.E.](https://wiki.u-gov.it/confluence/display/SCAIUS/UG3.2%3A+D.A.V.I.D.E.+UserGuide) compiled on one of its head node.

```
$ lsb_release -a
LSB Version:	:core-4.1-noarch:core-4.1-ppc64le
Distributor ID:	CentOS
Description:	CentOS Linux release 7.5.1804 (AltArch) 
Release:	7.5.1804
Codename:	AltArch

$ uname -a (on head node)
Linux davidefe02 3.10.0-862.3.2.el7.ppc64le #1 SMP Tue May 22 03:45:03 BST 2018 ppc64le ppc64le ppc64le GNU/Linux
```

### Proposal

First of all: weird, my [minimal example](https://github.com/boostorg/predef/issues/90) works and detects `BOOST_OS_LINUX`. Pasting our [hacked predef](https://github.com/ComputationalRadiationPhysics/alpaka/blob/0.3.5/include/alpaka/core/BoostPredef.hpp) over it still sets the macro.

**It seems `nvcc` does not forward the `linux`/``__linux`` macro defines on Power!** -> reported as Nvidia Bug ID 2448610 -> fixed in CUDA 10.1+

Let's apply a CMake based work-around for now, setting `linux` on device if on linux.

Also, we should try to query this information via a POSIX interface in order to support Linux/OSX and the like at one swipe as long as a POSIX layer exists. If there is no POSIX interface for it, we could take some of the established queries from KWSys from Kitware: [link to class](https://gitlab.kitware.com/utils/kwsys/blob/master/SystemInformation.cxx#L1456-1461)

PR #702, opened 2018-11-23
Power Linux: Fix NVCC Macro

ax3l:
NVCC forgets, only on OpenPOWER systems, to forward the `linux`/`__linux` macros. This causes alpaka to fail to compile.

Proposed fix for CUDA/NVCC bug found in #701.

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -541,6 +541,14 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 SET(CUDA_PROPAGATE_HOST_FLAGS ON)
 
+                # nvcc sets no linux/__linux macros on OpenPOWER linux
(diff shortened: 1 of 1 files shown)


BenjaminW3 (approved):

psychocoderHPC (approved):

alpakaConfig.cmake
```diff
@@ -541,6 +541,14 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 SET(CUDA_PROPAGATE_HOST_FLAGS ON)
 
+                # nvcc sets no linux/__linux macros on OpenPOWER linux
+                # nvidia bug id: 2448610
```
ax3l, line 545: Note: fixed in CUDA 10.1+

Issue #637, opened 2018-09-11
Relicensing vom LGPL3+ to MPL2


ax3l:
I am currently talking to @SylvainCorlay and they are interested in using Alpaka for GPU support. Since they have (American) business customers, they can't include copyleft licenses in their stack.

Since Alpaka is header-only and LGPL is ~~not very well defined~~ *very bloated, yet in v3+ okay* for such kind of libraries, see e.g. Eigen discussions in the past [[1]](https://eigen.tuxfamily.org/index.php?title=Licensing_FAQ&oldid=1117#How_is_Eigen_licensed.3F) [[2]](https://eigen.tuxfamily.org/index.php?title=Main_Page#License) [[3]](https://eigen.tuxfamily.org/index.php?title=News:Relicensing_to_MPL2!), I don't really see how we can benefit from it. Also, Alpaka is designed in a way that it is fully user-level extensible via traits, making the definition of "improvements"/"changes" on Alpaka "core" very hard to define.

The discussion I would therefore start is if we want to move Alpaka (and cupla) to a license without copyleft. E.g. ISC, MIT license, new BSD, MPL2 or Apache2. If I remember correctly, @juckel told me that the recommended licenses at HZDR are currently GPL(3+), LGPL(3+) and Apache2. Imho, ISC-like licenses are fine too since Europe is not too big in software patents, but Apache2 also keeps some options open for us for potential trademarks which is useful.

Update: Nevertheless, an important issue with [Apache2](https://en.wikipedia.org/wiki/Apache_License) is, that it is incompatible with GPL2. In the projects we develop, we always use the latest GPL (v3) (and allow "or newer"), but we should not exclude GPL2. Therefore, using as Eigen a [MPL2](https://en.wikipedia.org/wiki/Mozilla_Public_License) license would have the [same benefits](https://choosealicense.com/licenses/) without the GPL2 incompatibility. (We want to be in the Linux kernel at some point, don't we? ;-) )

Besides the discussion about this, we would then need to collect the agreement from all contributors, e.g. in a GitHub issue. But let us discuss first in the next days/weeks.

ax3l:
Just for comparison, spack is currently migrating to a dual license MIT/Apache2 and is basically also shipping as-source since it's a python lib. See rationale here: https://github.com/spack/spack/issues/9144

BenjaminW3:
[...]

PR #708, opened 2018-12-01
Enable check for TriviallyCopyable for CUDA and HIP backends

### LINKED ISSUE #703: Investigate usage of std::is_trivially_copyable
It is meant to check that the kernel can be copied to the GPU with a simple memcpy. However the check fails on all platforms so something might be wrong.

BenjaminW3:
Fixes #703

Checks that all parameters and all kernel function objects are trivially copyable.
We do not yet enable it on CPU.

### DIFF
--- a/include/alpaka/meta/IsTriviallyCopyable.hpp
+++ b/include/alpaka/meta/IsTriviallyCopyable.hpp
@@ -1,0 +1,51 @@
+/* Copyright 2019 Benjamin Worpitz
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/BoostPredef.hpp>
+#include <alpaka/core/Common.hpp>
+
+#include <type_traits>
+
+#if BOOST_LANG_CUDA
+#include <nvfunctional>
+#endif
+
+namespace alpaka
+{
+    namespace meta
+    {
+        template<
+            class T>
+        struct IsTriviallyCopyable
+            : std::integral_constant<bool,
+#if BOOST_LIB_STD_GNU
+                __has_trivial_copy(T)
+#else
+                std::is_trivially_copyable<T>::value
+#endif
+#if BOOST_LANG_CUDA && !BOOST_COMP_CLANG_CUDA
+                || __nv_is_extended_device_lambda_closure_type(T)
+                || __nv_is_extended_host_device_lambda_closure_type(T)
+#endif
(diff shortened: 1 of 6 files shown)


example/helloWorldLambda/src/helloWorldLambda.cpp
```diff
@@ -150,13 +153,23 @@ auto main()
...
     // std::function and provide it to the alpaka 
     // library.
+#if !BOOST_LANG_CUDA && !BOOST_LANG_HIP
     alpaka::kernel::exec<Acc> (
         queue,
         workDiv,
         std::function<void(Acc&, size_t)>( hiWorldFunction<Acc> ),
         nExclamationMarks);
+#endif
+#if BOOST_LANG_CUDA
```
psychocoderHPC: We should not start to use #ifdef within the example because the goal of alpaka is to avoid that. (if this is only for testing all is fine)
We can ship a meta::function with alpaka which is doing `ifdef`'s.

include/alpaka/meta/IsTriviallyCopyable.hpp
```diff
@@ -0,0 +1,62 @@
+/**
+* \file
+* Copyright 2018 Benjamin Worpitz
+*
+* This file is part of alpaka.
+*
+* alpaka is free software: you can redistribute it and/or modify
```
ax3l: With regards to #637, this new file's header will need an update :)

test/unit/meta/src/IsTriviallyCopyableTest.cpp
```diff
@@ -0,0 +1,107 @@
+/**
+ * \file
+ * Copyright 2018 Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * alpaka is free software: you can redistribute it and/or modify
```
ax3l: With regards to #637, this new file's header will need an update :)

ax3l (changes requested):

test/unit/meta/src/IsTriviallyCopyableTest.cpp
```diff
@@ -0,0 +1,95 @@
...
+        IsTriviallyCopyableReference == IsTriviallyCopyableResult,
+        "alpaka::meta::IsTriviallyCopyable failed!");
+}
+
+struct C {
+    C(){}
+    C(C const&) {}
+};
+
+/*struct D {
```
psychocoderHPC: Why is this commented out? Could you add a docstring with the reason, please.

psychocoderHPC (changes requested):

test/unit/meta/src/IsTriviallyCopyableTest.cpp
BenjaminW3: Not commented out anymore

ax3l (dismissed):

PR #709, opened 2018-12-09
Use boost.test header-only

BenjaminW3:
This is available in all boost version we support. This was explicitly documented starting with the upcoming boost-1.69.0 release which somehow broke our build by failing to find the prebuild boost.test library.


alpakaConfig.cmake
```diff
@@ -282,7 +282,6 @@ IF(${ALPAKA_DEBUG} GREATER 1)
     MESSAGE(STATUS "Boost_MINOR_VERSION: ${Boost_MINOR_VERSION}")
     MESSAGE(STATUS "Boost_SUBMINOR_VERSION: ${Boost_SUBMINOR_VERSION}")
     MESSAGE(STATUS "Boost_LIB_DIAGNOSTIC_DEFINITIONS: ${Boost_LIB_DIAGNOSTIC_DEFINITIONS}")
-    MESSAGE(STATUS "Boost_LIBRARIES: ${Boost_LIBRARIES}")
```
ax3l, line 285: why did you remove that status? Is it truly empty now or was it already before?

BenjaminW3: It is a duplicate. It is already printed in another line.

ax3l:
Just a note that this will, obviously, significantly increase compile time for tests and CI. [develop](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/464908903), [PR](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/466185815)
alpaka_boost_test_compile

BenjaminW3:
@ax3l Those build time comparisons have to be taken with care. There is a huge difference (~6-10 min) between builds where the cached docker image with all the dependencies is used and runs where this image is created before the real test.

However, in this case the build times really got worse by approx. 3 min per job. We are now building boost.test inline within each test project so 3 minutes for more then 20  targets is around 9 sec per target which is reasonable for building boost.test.
Nearly all other testing frameworks are header-only or at least by default header-only (catch, gtest, ...) so this would be similar with other frameworks.

I am still undecided if we really want to have this build time increase.

ax3l:
If you like, I refactored the whole test suite to stand-alone work with catch2. Will open a PR later on your updates.

I use catch2 in a way where most implementation parts are only compiled once in a `.cpp` file for the `main()` and then linked against the test cases that pull only part of the header. Quite nifty and we could later on speed it up even more with PCH for the tests via cotire.

BenjaminW3:
@ax3l Even though we may end up with a new test framework I would like to get this merged first.
Your PR builds on top of this one and this one already includes some clean-ups so I would prefer to get this merged even though it is only an intermediate step.

ax3l:
@BenjaminW3 sounds good to me.

Is it possible you accidentally added co-routine executors to this PR yesterday? Somehow the change went up a lot.

BenjaminW3:
ah.... this slipped in. I have removed it.

ax3l:
Yes, also somehow new tests were added in this "fix" PR. Can we keep this please minimal? Or does this only look like new ones were added? Can you rebase against `develop` to make sure?

BenjaminW3:
The tests are not new. I have split the integration tests into the test and the main.cpp similar to the unit tests for easier migration and maybe better compilation time (at least when building in parallel).
I am not sure why github is so bad in showing the changes. It detects on file as changed (removed nearly everything) but it should also detect the second file as a copy of the original where only some lines are missing.

ax3l:
Ok, then I have to solve some conflicts now but I will manage. I build the runner only once and link it to all tests, which saves building the same thing in a `src.cpp` all the time.

ax3l (approved):
Thx! Please do not refactor further in this PR.

PR #711, opened 2018-12-11
check for kernel void return type

psychocoderHPC:
follow up of #695

- reenable the kernel return type evaluation if `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is used
- add on device return type evaluation for the accelerator `AccGpuCudaRt` and `AccGpuHipRt`
- disable return type check on device side if clang is the device compiler

### DIFF
--- a/include/alpaka/exec/ExecGpuCudaRt.hpp
+++ b/include/alpaka/exec/ExecGpuCudaRt.hpp
@@ -87,6 +87,11 @@ namespace alpaka
 #if BOOST_ARCH_PTX && (BOOST_ARCH_PTX < BOOST_VERSION_NUMBER(2, 0, 0))
     #error "Cuda device capability >= 2.0 is required!"
 #endif
+                    static_assert(
+                        std::is_same<typename std::result_of<
+                            TKernelFnObj(acc::AccGpuCudaRt<TDim, TIdx> const &, TArgs const & ...)>::type, void>::value,
+                        "The TKernelFnObj is required to return void!");
+
                     acc::AccGpuCudaRt<TDim, TIdx> acc(threadElemExtent);
 
                     kernelFnObj(
--- a/include/alpaka/exec/ExecGpuCudaRt.hpp
+++ b/include/alpaka/exec/ExecGpuCudaRt.hpp
@@ -87,11 +87,14 @@ namespace alpaka
 #if BOOST_ARCH_PTX && (BOOST_ARCH_PTX < BOOST_VERSION_NUMBER(2, 0, 0))
     #error "Cuda device capability >= 2.0 is required!"
 #endif
+
+// with clang it is not possible to query std::result_of for a pure device lambda created on the host side
+#if !(BOOST_COMP_CLANG_CUDA && BOOST_COMP_CLANG)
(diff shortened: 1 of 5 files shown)


BenjaminW3 (dismissed):

ax3l (dismissed):
Ha, of course! :D

psychocoderHPC:
`clang-5 + CUDA-8.0 Release ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is currently not happy with my changes but is doing is well.

https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/466453024#L909-L913

psychocoderHPC:
I disabled the return type evaluation if clang is the device compiler. I do not remove it only if `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is activated because the validation will fail also if the user provides a `__device__` only functor even if `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is false.

include/alpaka/exec/ExecGpuCudaRt.hpp
```diff
@@ -87,6 +87,14 @@ namespace alpaka
 #if BOOST_ARCH_PTX && (BOOST_ARCH_PTX < BOOST_VERSION_NUMBER(2, 0, 0))
     #error "Cuda device capability >= 2.0 is required!"
 #endif
+
+// with clang it is not possible to query std::result_of for a pure device lambda created on the host side
+#ifndef BOOST_COMP_CLANG_CUDA BOOST_COMP_CLANG
```
BenjaminW3: This should most probably be `#if !(BOOST_COMP_CLANG_CUDA && BOOST_COMP_CLANG)`

BenjaminW3:
> I do not remove it only if ALPAKA_ACC_GPU_CUDA_ONLY_MODE is activated because the validation will fail also if the user provides a __device__ only functor even if ALPAKA_ACC_GPU_CUDA_ONLY_MODE is false.

But if the user provides a `__device__` only function object, it would not be valid without `ALPAKA_ACC_GPU_CUDA_ONLY_MODE ` und would most probably not compile at all.

psychocoderHPC:
> > I do not remove it only if ALPAKA_ACC_GPU_CUDA_ONLY_MODE is activated because the validation will fail also if the user provides a **device** only functor even if ALPAKA_ACC_GPU_CUDA_ONLY_MODE is false.
> 
> But if the user provides a `__device__` only function object, it would not be valid without `ALPAKA_ACC_GPU_CUDA_ONLY_MODE ` und would most probably not compile at all.

This depends strongly on the used functions within the kernel. What you can do e.g. is to use a kernel to reset only one value on the gpu. IN that case, no alpaka index is queried and a pure `__device__` kernel would work even if `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` is false.

include/alpaka/exec/ExecGpuCudaRt.hpp
psychocoderHPC: thx for the hint, I updated the PR

BenjaminW3 (approved):

Issue #481, opened 2018-04-25
Remove Core Boost Dependency


ax3l:
Boost is a heavy and unstable dependency, which is notoriously breaking with CUDA and HPC compilers such as XL. It's internal dependencies are not cleanly exposed and can change for each release. Proper CMake support is still in the making. It adds to the complexity to get started with alpaka and dependencies of header-only libs like alpaka are generally very nasty.

Besides the optional fiber backend - which is totally fine if it pulls boost - I would propose to lighten the burden of dependencies in Alpaka. The core alpaka library should therefore get rid of it.

Boost.Tests can be replaced with the stand-alone and modern Catch2 library - we use it in two projects already and it works wonderful and switching is easy:
  https://github.com/openPMD/openPMD-api/pull/110

Compiler detection and macros such as ignore unused can be easily re-implemented.

Instead of mp11 #459 we could use [Brigand](https://github.com/edouarda/brigand) which is also used in large projects such as vtk-m:
  https://gitlab.kitware.com/vtk/vtk-m/commits/master/vtkm/internal/brigand.hpp

or [other libs](https://github.com/ComputationalRadiationPhysics/picongpu/issues/1997) that are active, standalone, with good test and compiler coverage and well-maintained.

theZiz:
I beg to differ. First of all, it is not the fault of boost, that it breaks notoriously with some compilers, but most of the time the fault of these compilers. As different libraries often use the same approach we are also very likely to run into the same problems with each of them. Instead of the chance to use a preinstalled boost version which every hpc system has plenty of, we damn the users to install just another template only library nearly no hpc system has available at default (like Catch2 or Brigand).

If you want to replace `boost::mpl` I would suggest to implement a very basic metaprogramming library ourself loosely leant to `boost::mp11` if you are afraid of staying with boost at all.

I don't see the alpaka test cases as a crucial part of the library. It is important for our CI, but for users the alpaka examples are way more important. So I would use for testing whatever suits best and **somebody has experience in**, so we can easily add more tests to make alpaka more bug prove.

ax3l:
> most of the time the fault of these compilers

chicken-egg question but boost lacks coverage, backports and has a long turn-around for fixes (half a year+). Do to the lack of coverage, previously applied work-arounds for compilers have often be broken in new releases again.

> Instead of the chance to use a preinstalled boost version which every hpc system has plenty of

[...]

PR #712, opened 2018-12-11
Tests: Migrate to Catch 2.4.0+

ax3l:
Migrate from Boost.Test to [Catch2](https://github.com/catchorg/Catch2#how-to-use-it) for:

- compile time
- stability
- better native compiler support
- automated, modern [code coverage and tooling](https://github.com/catchorg/Catch2/blob/v2.5.0/README.md) on mainline repo
- no C++98 work-around legacy with a modern code base
- no long-term bugs (e.g. [on OSX](https://github.com/openPMD/openPMD-api/pull/101)) and ominous valgrind issues
- triggers no warnings (as long as we didn't migrate to CMake's `CUDA` language feature)
- no dependencies (fully self-contained: just needs C++11 & a stdlib, easily ship-able when vendored)

Uses catch2 in a way where most implementation parts are only compiled once [in a `.cpp` file](https://github.com/catchorg/Catch2/blob/v2.5.0/docs/slow-compiles.md#practical-example) for the `main()` and then linked against the test cases that pull only part of the header. Quite nifty and we could later on speed it up even more with [PCHs](https://github.com/catchorg/Catch2/blob/v2.5.0/docs/ci-and-misc.md#precompiled-headers-pchs) for the tests, e.g. via [cotire](https://github.com/sakra/cotire).

Related to #481.

- [x] If you like, I can [vendor a fixed single-header release](https://github.com/catchorg/Catch2/blob/v2.5.0/docs/tutorial.md#where-to-put-it) into an alpaka third-party dir so one does not even need to pre-install Catch2 at all. I did this before in other libs and would add a switch that allows switching to an external dependency as well: #718
- [ ] If you like, we can enable some [Catch2 fast-compile](https://github.com/catchorg/Catch2/blob/master/docs/configuration.md#other-toggles) options on CI only, making runtime crashes a bit less detailed but one would likely debug local anyway.
- [x] Compile times on AppVeyor look significantly improved ([`develop` before](https://ci.appveyor.com/project/BenjaminW3/alpaka-t7xmf/builds/20923270)) and Travis-CI ([`develop` before](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/464908903)) seems very well too, even including the initial docker image rebuild.

## Debugger Options

In case you want to break into the debugger on failures, run an individual test executable with
```
gdb --args ./testName -a -b
```

as `-h` will show: `-a` aborts on the first error and `-b` breaks into the debugger.

## Developers Install of Catch2

If we don't vendor a single-header version directly in Alpaka yet, which I would say we can do, install catch 2.4.0+ for yourself with user-level package managers. For example:

```bash
# (Linux/OSX)
spack install catch
spack load catch
```

or via
```bash
# (Win/Linux/OSX)
conda install -c schrodinger catch2
```

or via conan from [bincrafters](https://bintray.com/bincrafters/public-conan/catch2%3Abincrafters) (Win/Linux/OSX)

or via [vcpkg](https://github.com/Microsoft/vcpkg/tree/master/ports/catch2) (Windows)

or via [aur](https://www.archlinux.org/packages/community/any/catch2/) (Arch Linux)

or just quickly [build it form source](https://github.com/catchorg/Catch2/blob/master/docs/cmake-integration.md#installing-catch2-from-git-repository) and extend your `CMAKE_PREFIX_PATH`


script/travis/docker_install.sh
```diff
@@ -28,11 +28,11 @@ ls "${ALPAKA_CI_DOCKER_CACHE_DIR}"
 
 ALPAKA_DOCKER_BUILD_REQUIRED=1
 
-if [ -f "${ALPAKA_CI_DOCKER_CACHE_IMAGE_FILE_PATH}" ]
-then
-    # NOTE: The image being available is not the only precondition. If anything within any of the scripts has changed in comparison to the ones that created the docker image, we might have to rebuild the image.
-    ALPAKA_DOCKER_BUILD_REQUIRED=0
-fi
+#if [ -f "${ALPAKA_CI_DOCKER_CACHE_IMAGE_FILE_PATH}" ]
```
ax3l: This is confusing. How is the workflow to re-build the image? Just remove the cache on Travis-CI?

ax3l:
Switched to Catch 2.4.0+ (instead of 2.3.0+) now since I saw in tests and changelog that runtime improved further in this release.

script/travis/docker_install.sh
BenjaminW3: By commenting out this piece of code, you enforce rebuilding of the docker image on each build. Deleting the travis cache would be the other option.

alpakaConfig.cmake
```diff
@@ -829,6 +828,9 @@ ENDIF() # HIP
 IF(MSVC)
     # Empty append to define it if it does not already exist.
     LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC)
+
+    # Disable autolink feature of boost. We explicitly link the required libs.
+    LIST(APPEND _ALPAKA_COMPILE_DEFINITIONS_PUBLIC "BOOST_ALL_NO_LIB")
```
BenjaminW3: This is not necessary in your branch.

ax3l:
@BenjaminW3 I am still getting kicked in a few clang-cuda examples by boost (ha, it struggles me ^^)
```
In file included from alpaka/test/integ/axpy/src/main.cpp:23:
In file included from boost/math/special_functions/relative_difference.hpp:9:
In file included from boost/math/special_functions/fpclassify.hpp:16:
In file included from boost/limits.hpp:14:
In file included from /boost/config.hpp:61:
boost/config/detail/suffix.hpp:510:26: error: __float128 is not supported on this target
   __extension__ typedef __float128 float128_type;
```

in the `axpy` test. Some work-around we usually apply there? Adding `-DBOOST_MATH_DISABLE_FLOAT128` did not cut it.

script/travis/docker_install.sh
BenjaminW3: Do you want to reenable this part so that the average build that does not change the travis.yml files is faster?

test/integ/axpy/src/axpy.cpp
```diff
@@ -239,14 +220,23 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(
...
+};
 
-BOOST_AUTO_TEST_SUITE_END()
+TEST_CASE( "axpy", "[axpy]")
+{
+    using TestAccs = alpaka::test::acc::EnabledAccs<
+        alpaka::dim::DimInt<1u>,
+        std::size_t>;
+
+    alpaka::meta::forEachType< TestAccs >( TestTemplate() );
```
BenjaminW3, line 241: This is a cool workaround but I will be missing the boost.test version where we had one test per type and could easily see if the test is failing for a specific type only.

test/unit/event/src/EventTest.cpp
```diff
@@ -195,63 +173,64 @@ BOOST_AUTO_TEST_CASE_TEMPLATE(
...
 }
+};
 
 //-----------------------------------------------------------------------------
 // github issue #388
-BOOST_AUTO_TEST_CASE_TEMPLATE(
-    waitForEventThatAlreadyFinishedShouldBeSkipped,
-    TDevQueue,
-    TestQueues)
+struct TestTemplateFinishedShouldBeSkipped
```
BenjaminW3, line 230: Could you please fix the names of those structs so that it matches the test name?

test/integ/axpy/src/axpy.cpp
ax3l: We could add a [section](https://github.com/catchorg/Catch2/blob/master/docs/tutorial.md#test-cases-and-sections) or some other kind of annotation.

alpakaConfig.cmake
BenjaminW3: You can now remove the line that adds the `BOOST_ALL_NO_LIB`.

BenjaminW3:
@ax3l Did we already have the `__float128` problem before?

ax3l:
It was likely masqueraded: It's interesting that we pass `-std=c++11` *and* `-std=gnu++11` to the `axpy` build with clang. This could be a stdlib mismatch.

Looks a bit like the comments [on this SA](https://stackoverflow.com/questions/43316533/float128-is-not-supported-on-this-target) (not the answer) to me.

BenjaminW3:
Could you please try if we could remove the whole block from the `before_install.sh` mentioning `# FIXME: BOOST_AUTO_TEST_CASE_TEMPLATE is not compilable with nvcc in Release mode.` ?

ax3l:
The problem is likely caused by the way that we still manually add our compiler flags for C++ instead of using target properties. I "linked"/included the external library now more old school to avoid populating the std flags twice.

ax3l:
Is it ok if we switch to a newer CMake, e.g. 3.11+ ? Improves handling of imported targets a lot and also (3.8+) adds better control over target properties/features.

script/travis/run_build.sh
```diff
@@ -55,6 +55,10 @@ echo "KMP_ALL_THREADS=${KMP_ALL_THREADS}"
 echo "KMP_TEAMS_THREAD_LIMIT=${KMP_TEAMS_THREAD_LIMIT}"
 echo "OMP_THREAD_LIMIT=${OMP_THREAD_LIMIT}"
 echo "OMP_NUM_THREADS=${OMP_NUM_THREADS}"
+echo "ALPAKA_CI_CATCH2_ROOT_DIR=${ALPAKA_CI_CATCH2_ROOT_DIR}"
+ls ${ALPAKA_CI_CATCH2_ROOT_DIR}
+ls -R ${ALPAKA_CI_CATCH2_ROOT_DIR}
```
BenjaminW3: Can we remove those traces?

ax3l: Yes, will clean up as soon as the CI passes.

test/unit/event/src/EventTest.cpp
ax3l: Didn't I? I tried to give them consequent suffixes. But in this test case they were already especially longish :)

BenjaminW3:
After thinking about it some time I would prefer to include the testing framework within the project. I do not see any advantage of requiring the user to download the framework. Do you see any?
The user should still be able to set a different path to the test framework but the default should be the one included in the repo.
Should we use a submodule for this?

ax3l:
Sure thing! Catch2 comes as a single header, either we check it in as such or download it on the fly with cmake. I am not a big fan of submodules for their usebility outside of pure meta repos.

ax3l:
@BenjaminW3 can we please merge this before the other CI changes come in? I would like to avoid rebasing this a lot.
I can add a vendored catch2 header later on in a follow-up (I have the PR ready).

ax3l:
Uh, Catch 2.5.0 added the `TEMPLATE_TEST_CASE` and `TEMPLATE_TEST_CASE_METHOD` macros: https://github.com/catchorg/Catch2/pull/1437

test/integ/axpy/src/axpy.cpp
ax3l: Update: Catch 2.5.0 added the `TEMPLATE_TEST_CASE` and `TEMPLATE_TEST_CASE_METHOD` macros: https://github.com/catchorg/Catch2/pull/1437

BenjaminW3: Will this really help us? It looks like they support only expanded type lists. We are generating the type list dynamically so there still seems to be one step missing.

ax3l: Yes, it looks like it's just close to what I build as well.

BenjaminW3 (approved):

ax3l:
Final CI compile-time comparison (similar to [here](https://github.com/ComputationalRadiationPhysics/alpaka/pull/709#issuecomment-446159571)): recent [develop](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/464908903) vs. [PR](https://travis-ci.org/ComputationalRadiationPhysics/alpaka/builds/474307597) with caches up
(3:40h vs. 3:03h).

[Catch2 fast compile](https://github.com/catchorg/Catch2/blob/master/docs/configuration.md#other-toggles) *not* yet activated and not yet vendored into our source.

catch2

ax3l:
@BenjaminW3 hurray, CI finished :)

ax3l:
@BenjaminW3 Catch 2.6.0 now contains a [`TEMPLATE_PRODUCT_TEST_CASE`](https://github.com/catchorg/Catch2/blob/v2.6.0/docs/test-cases-and-sections.md#type-parametrised-test-cases) macro as we had it before :)

PR #713, opened 2018-12-18
CMake: 3.11.0+

ax3l:
Newer CMake Version add several improvements that we want to use.

3.8.0+ adds support for extended `target_compile_features` that can be used to fine-graine control C++ standard minimum requirements (contrary to manual flags or exact settings with `CXX_STANDARD`).

3.11.0+ improves handling for `IMPORTED` targets so that they can be modified in properties. (Useful for #712)


BenjaminW3:
I always like to drop old dependencies. It is more or less a question of what is installed on the systems our users are using. Does the average HPC cluster use CMake 3.11.0 (release March 28, 2018)?

ax3l:
Yes, 3.11+ is quite wide spread (Taurus, Hemera, Davide, Summit, Piz Daint, ~~Hypnos~~ (*to be integrated in Hemera soon*)). The nice thing about CMake dependencies is, that CMake itself as a build tool is "dependency free". That means requesting a new CMake version is very easy for us when it is missing, because it triggers no long dependency chains that need a rebuild.

(Contrary to a new boost, MPI, or compiler version.)

As developers, the easiest way to get new CMake is to either download the binary releases or using some user-level package manager such as `conda` or `spack` to overwrite it in one's user scope.

ax3l:
@BenjaminW3 should I delete the travis cache for this dependency update?

BenjaminW3:
You do not have to delete the cache because it is automatically done when an environment variable is changed. This happens when you change the travis. yml file

ax3l:
@BenjaminW3  @psychocoderHPC merge-able :)

BenjaminW3 (approved):

PR #714, opened 2018-12-22
Unify bash set options for all scripts and create travis entry point script

BenjaminW3:


BenjaminW3:
Ready to be merged!

ax3l (approved):
Very nice and great you add `set ...` options as well!

Issue #685, opened 2018-10-18
Test Travis CI windows support


BenjaminW3:
Travis CI has added support for Windows see https://blog.travis-ci.com/2018-10-11-windows-early-release
We should try this out to unify our build scripts and reduce the number of tools in use.

BenjaminW3:
I successfully build alpaka on a Travis CI Windows host! https://travis-ci.org/BenjaminW3/alpaka/builds/471273699
I will try to integrate the changes when I have time.
I would prefer to drop appveyor when we have reached the state of the current appveyor build.

PR #715, opened 2018-12-24
Move windows CI builds from appveyor to Travis CI.

BenjaminW3:
Implements #685

Notes:
* I had to set the `CC` and `CXX` environment variables within the `before_install.sh` (executed in the git bash) again because those are set to g++/gcc even though they are already overwritten in the .travis.yml file. This may be a problem with the current Travis CI windows environment and may be fixed later.
* We do not yet explicitly set the 64 bit compiler build so the compilation may be done by the 32 bit compiler (both can produce 32 and 64 bit results)

TODO:
* [x] merge #714 and rebase
* [x] delete the appveyor project/account (after this has been merged)


.travis.yml
```diff
@@ -190,23 +209,24 @@ matrix:
...
 cache:
     directories:
         - $ALPAKA_CI_DOCKER_CACHE_DIR
 
 script:
     - set -eovx pipefail
-    - sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install smem
-    - sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install moreutils
-    - ./script/travis/script.sh | ts
+    - if [ "$TRAVIS_OS_NAME" = "linux" ] ;then sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install smem ;fi
```
ax3l, line 226: you can make this in one line: `... install smem moreutils`

BenjaminW3: I know. They have been added at different times and I still want to move the installation of `smem` to the place where it is really needed.

BenjaminW3:
Ready to be merged!

psychocoderHPC (approved):

ax3l (approved):

ax3l:
Will merge, the failing AppVeyor makes sense since it is removed now.

PR #718, opened 2019-01-02
Vendor Catch2: v2.5.0

ax3l:
Add a third-party header for a vendored catch2 for our tests.

Source: https://github.com/catchorg/Catch2/releases
Version: 0.2.5

## New CMake Option

By default, we will build the tests with the internally vendored Catch2 in order to make testing easy. If you want to force using an external Catch2 install (expand your `CMAKE_PREFIX_PATH` with its location), then set: `-DALPAKA_USE_INTERNAL_CATCH2=OFF`


script/travis/before_install.sh
```diff
@@ -89,7 +89,6 @@ if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" ] || [ "${ALPAKA_ACC_GPU_HIP_ENABLE
 then
     if [ "${ALPAKA_CUDA_COMPILER}" == "nvcc" ]
     then
```
BenjaminW3, line 94: Please try to remove the whole `CUDA` block.

BenjaminW3:
The changes hit a known bug in gcc-5.5:
* https://gcc.gnu.org/bugzilla/show_bug.cgi?id=76731
* https://github.com/tensorflow/tensorflow/issues/10220 .

It works with gcc-5.4, gcc-4.9 or gcc-6 and also when building with nvcc 10.0. It previously worked because we had been building debug builds only but not release.

Could you please revert the changes in `before_install.sh` so that this can be merged and create a bug ticket to resolve the nvcc-9+gcc-5.5+release issue?

test/CMakeLists.txt
```diff
@@ -26,9 +26,15 @@ FIND_PACKAGE(alpaka REQUIRED)
 
 ADD_SUBDIRECTORY("common/")
 
-find_package(Catch2 2.4.0 CONFIG REQUIRED)
-set_target_properties(Catch2::Catch2 PROPERTIES IMPORTED_GLOBAL TRUE)
-message(STATUS "Catch2: Found version ${Catch2_VERSION}")
+OPTION(ALPAKA_USE_INTERNAL_CATCH "Use internally shipped Catch2" ON)
+
+IF(ALPAKA_USE_INTERNAL_CATCH)
```
BenjaminW3: Could we rename it to `ALPAKA_USE_INTERNAL_CATCH2`?

ax3l: sure!

BenjaminW3 (approved):

PR #724, opened 2019-01-04
Execute tests on windows

BenjaminW3:
It works!
I had to disable the OpenMP 2.0 threads backend for the windows jobs because the tests are failing. I will investigate this separately.


ax3l (approved):
Nice, congrats!

PR #727, opened 2019-01-05
Clean up usage of catch2

BenjaminW3:
* Move `CatchMain` project into the `test` folder (for IDEs supporting this)
* Add `CATCH_CONFIG_FAST_COMPILE` define in the hope that it makes the compilation faster
* Add `CATCH_CONFIG_WINDOWS_CRTDBG` define for memory leak checking under windows
* remove unnecessary/duplicate defines

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -39,13 +39,16 @@ ENDIF()
 add_library(CatchMain CatchMain.cpp)
 # target_compile_features(CatchMain PUBLIC cxx_std_11)  # min C++11
 set_target_properties(CatchMain PROPERTIES
+    FOLDER "test"
     CXX_STANDARD 11  # exactly C++11
     CXX_EXTENSIONS OFF
     CXX_STANDARD_REQUIRED ON
     POSITION_INDEPENDENT_CODE ON
     WINDOWS_EXPORT_ALL_SYMBOLS ON
 )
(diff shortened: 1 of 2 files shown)


test/CMakeLists.txt
```diff
@@ -39,13 +39,16 @@ ENDIF()
...
 set_target_properties(CatchMain PROPERTIES
+    FOLDER "test"
     CXX_STANDARD 11  # exactly C++11
     CXX_EXTENSIONS OFF
     CXX_STANDARD_REQUIRED ON
     POSITION_INDEPENDENT_CODE ON
     WINDOWS_EXPORT_ALL_SYMBOLS ON
 )
-IF("${CMAKE_CXX_COMPILER_ID}" STREQUAL "MSVC")
+target_compile_definitions(CatchMain PUBLIC "CATCH_CONFIG_FAST_COMPILE")
```
ax3l, line 49: I think either as `-DCATCH_CONFIG_FAST_COMPILE` or just without ""

ax3l (approved):

test/CMakeLists.txt
BenjaminW3: `-DCATCH_CONFIG_FAST_COMPILE` won't work here because the `-D` is automatically prepended.
`"CATCH_CONFIG_FAST_COMPILE"` and `CATCH_CONFIG_FAST_COMPILE` would be identical here but in other places we pass the defines as "strings" as well.

PR #730, opened 2019-01-06
Fix CI not building HIP, replace ALPAKA_HIP_ARCH with ALPAKA_CUDA_ARCH, clean-up HIP CI handling

### LINKED ISSUE #728: Fix HIP backend not being tested in CI at all
Due to `ALPAKA_ACC_GPU_HIP_ONLY_MODE` not being passed from the scripts to the build, the build simply says `HIP back-end must be used together with ALPAKA_ACC_GPU_HIP_ONLY_MODE` and disables the HIP backend. The build is then a standard g++ build.

There are some other things wrong in the HIP build that I will fix as well.

BenjaminW3:
Fixes #728

* Due to ALPAKA_ACC_GPU_HIP_ONLY_MODE not being passed from the scripts to
the build, the build simply says `HIP back-end must be used together with
ALPAKA_ACC_GPU_HIP_ONLY_MODE` and disables the HIP backend. The build is
then a standard g++ build. This is fixed.
* Fixed the HIP build that was failing due to untested broken changes when adding the kernel void return checks 
* There is no difference between ALPAKA_HIP_ARCH and ALPAKA_CUDA_ARCH in the only HIP platform we currently support (ALPAKA_HIP_PLATFORM=nvcc). You already have to set
ALPAKA_CUDA_VERSION so it is much more logical to use ALPAKA_CUDA_ARCH as
well.
* Do not set ALPAKA_HIP_PLATFORM unconditionally for all jobs in CI
* Do not set ALPAKA_ACC_GPU_CUDA_ONLY_MODE unconditionally for all jobs in CI


ax3l (approved):
Cool, LGTM! :)

tdd11235813:
Thanks for the fixes :) 
I tested the AMD platform with the HCC backend for HIP, thus `ALPAKA_HIP_ARCH` is also for AMD targets when `ALPAKA_HIP_PLATFORM=hcc`. Although we do not test it at the moment on travis, the scripts still should be open for AMD platforms (because we internally have CI nodes with AMD GPUs, so we will use them sooner or later).

Issue #726, opened 2019-01-05
Visual Studio 2017 15.9 (regression): OpenMP not working in variadic templates


BenjaminW3:

BenjaminW3:
Our internal sanity check is triggered saying: `The OpenMP 2.0 runtime did not use the number of threads that had been required!`

sbastrakov:
This is just a guess, ofc the reason may be completely unrelated.

Strictly speaking, [the failing check](https://github.com/ComputationalRadiationPhysics/alpaka/blob/27a7e11bf4d7096b8ed9eea7f3612fcc78b3cc4e/include/alpaka/exec/ExecCpuOmp2Threads.hpp#L194) actually does not *always* have to hold. According to [the spec (page 9 in the internal numbering, paragraph from line 28)](https://www.openmp.org/wp-content/uploads/cspec20.pdf), in case the requested number in the `#pragma omp parallel num_threads(...)` exceeds the number available on the system, the behaviour is implementation-defined. So it is possible the CI system (or due to our configuration) does not provide enough resources and this check fails.

BenjaminW3:
Yes, I know that this is possible. Therefore we added the check because the result would be invalid in case the runtime used a different number of threads.
I was able to reproduce this locally and added a check for `omp_in_parallel` which always returns false for both OpenMP 2.0 backends. So there seems to be something wrong with the Windows OpenMP build.

sbastrakov:
@BenjaminW3 sure, I see the point of that. But what if due to some unfortunate configuration we request more threads than the system supports (dunno, have a hard time imagining that myself). 

Btw maybe add info about the requested and actual number of threads to the exception message?

ax3l:
Yes, the situation in travis: more physical cores (N>2) "visible" than accessible (2). Could confuse OpenMP startup. Maybe we just export `OMP_NUM_THREADS` to e.g. 2?

sbastrakov:
Don't we do it already? iirc that's the standard travis recommendation for OpenMP-enabled codes.

Edit: I meant not that I'm sure we do it already (currently I don't know that), but I'm sure we should do it.

BenjaminW3:
@sbastrakov found the reason and reported it to microsoft: https://developercommunity.visualstudio.com/content/problem/426580/c-variadic-template-function-incorrectly-interacts.html

ax3l:
Can you please update the PR description with the known failing/working VS versions as long as we still know them? :)

[...]

PR #733, opened 2019-01-13
Add additional checks for OpenMP executors

BenjaminW3:
This adds a check that `omp_in_parallel` is true when within a parallel region.
The OpenMP runtime does not necessarily create a parallel region when only one thread is required.

This shows that OpenMP on windows does not work at all right now.

### DIFF
--- a/include/alpaka/exec/ExecCpuOmp2Threads.hpp
+++ b/include/alpaka/exec/ExecCpuOmp2Threads.hpp
@@ -173,14 +173,15 @@ namespace alpaka
                         // Therefore we use 'omp parallel' with the specified number of threads in a block.
                         #pragma omp parallel num_threads(iBlockThreadCount)
                         {
-                            // The first thread does some checks in the first block executed.
-                            if((::omp_get_thread_num() == 0) && (acc.m_gridBlockIdx.sum() == static_cast<TIdx>(0)))
+                            #pragma omp single nowait
                             {
-                                int const numThreads(::omp_get_num_threads());
-                                alpaka::ignore_unused(numThreads);
-#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
-                                std::cout << BOOST_CURRENT_FUNCTION << " omp_get_num_threads: " << numThreads << std::endl;
-#endif
+                                // The OpenMP runtime does not create a parallel region when only one thread is required in the num_threads clause.
+                                // In all other cases we expect to be in a parallel region now.
+                                if((iBlockThreadCount > 1) && (::omp_in_parallel() == 0))
+                                {
+                                    throw std::runtime_error("The OpenMP 2.0 runtime did not create a parallel region!");
+                                }
+
                                 // GCC 5.1 fails with:
                                 // error: redeclaration of const int& iBlockThreadCount
                                 // if(numThreads != iBlockThreadCount)
@@ -189,6 +190,7 @@ namespace alpaka
                                 // #pragma omp parallel num_threads(iBlockThreadCount)
                                 //         ^
 #if (!BOOST_COMP_GNUC) || (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0)) || (BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(6, 0, 0))
+                                int const numThreads(::omp_get_num_threads());
                                 if(numThreads != iBlockThreadCount)
                                 {
                                     throw std::runtime_error("The OpenMP 2.0 runtime did not use the number of threads that had been required!");
@@ -199,7 +201,7 @@ namespace alpaka
                                 acc);
 
                             // Wait for all threads to finish before deleting the shared memory.
-                            // This is done by default if the omp 'nowait' clause is missing
+                            // This is done by default if the omp 'nowait' clause is missing on the omp parallel directive
                             //block::sync::syncBlockThreads(acc);
                         }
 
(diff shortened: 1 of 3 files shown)


sbastrakov (dismissed):
I believe this is *not* a runtime problem, but I guess either wrong settings or a bug in Visual Studio. The issue is (at least on my machine), the parallel section in question works like it is compiled with disabled OpenMP. To check I first added a separate parallel section to body of `alpaka::kernel::exec()` and there it also behaves as if OpenMP was disabled. At the same time, from the `main()` it all works reasonably (so generally OpenMP is enabled). Then I reproduced it as a stand-alone code here:

```
#include <iostream>
#include "omp.h"

void normalFunction()
{
    #pragma omp parallel num_threads(2)
    {
        #pragma omp single
        std::cout << "normalFunction(): omp_in_parallel() = " << omp_in_parallel()
            << ", omp_get_num_threads() = " << omp_get_num_threads() << "\n";
    }
}

template<typename... TArgs>
void problematicFunction(TArgs const & ... args)
{
    #pragma omp parallel num_threads(2)
    {
        #pragma omp single
        std::cout << "problematicFunction(): omp_in_parallel() = " << omp_in_parallel()
            << ", omp_get_num_threads() = " << omp_get_num_threads() << "\n";
    }
}

int main()
{
    normalFunction();
    problematicFunction();
    return 0;
}

```

Output from my Visual Studio:
```
normalFunction(): omp_in_parallel() = 1, omp_get_num_threads() = 2
problematicFunction(): omp_in_parallel() = 0, omp_get_num_threads() = 1
```

Not sure what causes the issue with the parameter-packed function (template function without parameter pack works fine). I've randomly tried adding `/Zc:twoPhase-` from [here](https://docs.microsoft.com/en-us/cpp/build/reference/zc-twophase?view=vs-2017) (also there is a warning about it in VS), but to no avail.

Btw the same probably causes #726 .

BenjaminW3:
I will further investigate the Windows issues.
I still want to get this sanity checking merged in separately.

sbastrakov:
@BenjaminW3 I think in case I've correctly identified the issue, the proposed changes are probably not the best way to check for it.

sbastrakov:
I mean, it seems to me that the situation the parallel section is not created is maybe not actually possible with a proper configuration. So checking for that inside the kernel execution logic is not nice. If this is just a VS bug, we could check for it separately, as it has nothing to do with our code.

BenjaminW3:
I exactly want to check for proper runtime behaviour/configuration with my checks. We already have such checks for the correct thread count. It should never fail.

include/alpaka/exec/ExecCpuOmp2Threads.hpp
```diff
@@ -173,14 +173,15 @@ namespace alpaka
...
-                            // The first thread does some checks in the first block executed.
-                            if((::omp_get_thread_num() == 0) && (acc.m_gridBlockIdx.sum() == static_cast<TIdx>(0)))
+                            #pragma omp single nowait
                             {
-                                int const numThreads(::omp_get_num_threads());
-                                alpaka::ignore_unused(numThreads);
-#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
-                                std::cout << BOOST_CURRENT_FUNCTION << " omp_get_num_threads: " << numThreads << std::endl;
-#endif
+                                // The OpenMP runtime does not create a parallel region when only one thread is required in the num_threads clause.
```
sbastrakov, line 178: Since now it seems this is not what causes the issue, maybe rewrite this comment?

BenjaminW3: No, this is not the Windows issue.
The linux CI showed that the OpenMP runtime does not necessarily create a parallel region when only one thread is requested.

sbastrakov:
@BenjaminW3 I see, thanks for clarifying (while I still believe it is far better to check for it outside, since it has absolutely nothing to do with the context). ~~Not sure how I could remove my 'requested changes', but I guess the motivation comment still has to be rewritten, as it turned out to be not true.~~

include/alpaka/exec/ExecCpuOmp2Threads.hpp
sbastrakov: Ah. I got really confused by your statement 'This shows that OpenMP on windows does not work at all right now.'

sbastrakov:
Btw reported the VS specific issue to Microsoft.

BenjaminW3:
Could you share the link?

sbastrakov:
Ofc: https://developercommunity.visualstudio.com/content/problem/426580/c-variadic-template-function-incorrectly-interacts.html
I was not sure how long does it take to get a link, so wrote immediately so that we do not report the same thing twice.

BenjaminW3:
Thanks for investigating this! I am pretty sure that it worked with one of the previous VS versions. It could also have been VS 2015 where I last saw it working.

BenjaminW3:
Please review/approve ;-)

psychocoderHPC (approved):

bertwesarg:
Dear all,

sorry for being late here, but I don't understand what we try to achieve here in the `Omp2Blocks` case. I consider it wrong to require that one uses OpenMP threads as long he also has more than one block to process. It may be worth a warning, but not a runtime error.

bertwesarg:
Just to be clear, it is currently not possible to run with `OMP_NUM_THREADS=1`.

psychocoderHPC:
@bertwesarg could you please open an issue for that so we can better track the state.

PR #736, opened 2019-01-25
HIP: `cpu::Copy`: fix missing suffix

psychocoderHPC:
The helper funtion `meta::ndLoopIncIdx` is defined with the suffix `ALPAKA_FN_HOST_ACC`.
This requires that all functors used with the meta function have the same suffix if the HIP backend (on AMD Rocm) is used else the code will not compile.
For cuda we workround it with a proprietary pragma `__pragma(hd_warning_disable)`

This PR set the correct sffix and removes the non backend agnostic function `std::memcpy`.

cc-ing: @tdd11235813

### DIFF
--- a/include/alpaka/mem/buf/cpu/Copy.hpp
+++ b/include/alpaka/mem/buf/cpu/Copy.hpp
@@ -181,12 +181,13 @@ namespace alpaka
                             {
                                 meta::ndLoopIncIdx(
                                     extentWithoutInnermost,
-                                    [&](vec::Vec<DimMin1, ExtentSize> const & idx)
+                                    [&] ALPAKA_FN_HOST_ACC (vec::Vec<DimMin1, ExtentSize> const & idx)
                                     {
-                                        std::memcpy(
-                                            reinterpret_cast<void *>(this->m_dstMemNative + (vec::cast<DstSize>(idx) * dstPitchBytesWithoutOutmost).foldrAll(std::plus<DstSize>())),
-                                            reinterpret_cast<void const *>(this->m_srcMemNative + (vec::cast<SrcSize>(idx) * srcPitchBytesWithoutOutmost).foldrAll(std::plus<SrcSize>())),
-                                            static_cast<std::size_t>(this->m_extentWidthBytes));
+                                        auto dst = reinterpret_cast<uint8_t *>(this->m_dstMemNative + (vec::cast<DstSize>(idx) * dstPitchBytesWithoutOutmost).foldrAll(std::plus<DstSize>()));
+                                        auto const src = reinterpret_cast<uint8_t const *>(this->m_srcMemNative + (vec::cast<SrcSize>(idx) * srcPitchBytesWithoutOutmost).foldrAll(std::plus<SrcSize>()));
+                                        size_t const numBytes = static_cast<std::size_t>(this->m_extentWidthBytes);
+                                        for(size_t b = 0u; b < numBytes; ++b)
+                                            dst[b] = src[b];
                                     });
                             }
                         }


include/alpaka/mem/buf/cpu/Copy.hpp
```diff
@@ -181,12 +181,13 @@ namespace alpaka
...
                                 meta::ndLoopIncIdx(
                                     extentWithoutInnermost,
-                                    [&](vec::Vec<DimMin1, ExtentSize> const & idx)
+                                    [&] ALPAKA_FN_HOST_ACC (vec::Vec<DimMin1, ExtentSize> const & idx)
                                     {
-                                        std::memcpy(
-                                            reinterpret_cast<void *>(this->m_dstMemNative + (vec::cast<DstSize>(idx) * dstPitchBytesWithoutOutmost).foldrAll(std::plus<DstSize>())),
-                                            reinterpret_cast<void const *>(this->m_srcMemNative + (vec::cast<SrcSize>(idx) * srcPitchBytesWithoutOutmost).foldrAll(std::plus<SrcSize>())),
-                                            static_cast<std::size_t>(this->m_extentWidthBytes));
+                                        auto dst = reinterpret_cast<uint8_t *>(this->m_dstMemNative + (vec::cast<DstSize>(idx) * dstPitchBytesWithoutOutmost).foldrAll(std::plus<DstSize>()));
```
BenjaminW3, line 186: I would think that we could declare this variable as `auto const` because the const should apply to the pointer and not to the value but I am not 100% sure.

include/alpaka/mem/buf/cpu/Copy.hpp
```diff
@@ -181,12 +181,13 @@ namespace alpaka
...
+                                        auto const src = reinterpret_cast<uint8_t const *>(this->m_srcMemNative + (vec::cast<SrcSize>(idx) * srcPitchBytesWithoutOutmost).foldrAll(std::plus<SrcSize>()));
+                                        size_t const numBytes = static_cast<std::size_t>(this->m_extentWidthBytes);
+                                        for(size_t b = 0u; b < numBytes; ++b)
+                                            dst[b] = src[b];
```
BenjaminW3, line 190: We do not have any measurements, but I would expect this code to be much slower.

BenjaminW3: Could you please add `{ }` around the loop body?

include/alpaka/mem/buf/cpu/Copy.hpp
psychocoderHPC: I can check if it works

include/alpaka/mem/buf/cpu/Copy.hpp
psychocoderHPC: > We do not have any measurements, but I would expect this code to be much slower.

I played in the past with https://github.com/raas/mbw where different methods were tested to copy memory. As I remember correctly it is mostly independent.

Another solution would be to declare the meta function as pure HOST function but then we can not use it on the device side.

Never the less I will think again about it if we can handle it somehow better.

include/alpaka/mem/buf/cpu/Copy.hpp
sbastrakov: Indeed, `auto const` would make `dst` have type `uint8_t * const`, as opposed to `uint8_t *` in the original code.

For `src` it is done fine: both the pointer and the data are `const` as they should be.

include/alpaka/mem/buf/cpu/Copy.hpp
psychocoderHPC: @tdd11235813 already opened an issue for the suffix qualifier https://github.com/ROCm-Developer-Tools/HIP/issues/659
I will wait if someone from the HIP dev team is answering soon before I will invest more time into this.

tdd11235813:
~~some good news, finally had some time to test host-device traits for HCC and it seems to work, so we could resolve it by adding things like this:~~
host-device attribute traits still not solved, but the wrapper can help to adjust lambda's host-device behavior:

**HCC helper for host-only functors**
```c++
template<typename TFunc>                          
struct HipUseHostOnly {                           
    TFunc func;                                   
    HipUseHostOnly(TFunc _func) : func(_func) {   
    }                                             
    template<typename... TArgs>                   
    ALPAKA_FN_HOST                                
    auto operator()(TArgs... args) const -> void {
        func(args...);                            
    }                                             
    template<typename... TArgs>                   
    ALPAKA_FN_ACC // actually needs an ALPAKA_FN_DEVICE
    auto operator()(TArgs... args) const -> void {
        // empty                                  
    }                                             
};                                                
```
**Wrap functors in the host-device backend functions**
```c++
// in alpaka/meta/NdLoop.hpp
ALPAKA_NO_HOST_ACC_WARNING                                     
template<                                                      
    typename TExtentVec,                                       
    typename TFnObj>                                           
ALPAKA_FN_HOST_ACC auto ndLoopIncIdx(                          
    TExtentVec const & extent,                                 
    TFnObj const & f)                                          
-> void                                                        
{                                                              
    ndLoop(                                                    
        meta::MakeIndexSequence<dim::Dim<TExtentVec>::value>(),
        extent,                                                
        HipUseHostOnly<TFnObj>(f));                            
}                                                              
```

- The helper needs to be adapted to the corresponding backend/compiler.
- The HIP/HCC always runs in only-mode, so `ALPAKA_FN_ACC` works, but `ALPAKA_FN_DEVICE` (was removed on my request^^) or `__device__` should be the better choice here

**Test**
- RX Vega64, ROCm 2.2.31 dev (HCC 1.3 - Clang 9.0) // dev system host=fwk388
- runtime test with CUPLA and cudaMemcpy2D
```c++
int a1[] = {4};                                                         
int a2[] = {1};                                                         
std::cout << a2[0] << " A\n";                                           
// prints 1
checkCudaErrors(cudaMemcpy2D(a2, 4, a1, 4, 4, 1, cudaMemcpyHostToHost));
std::cout << a2[0] << " B\n";                               
// prints 4            
```

Edit: changed optimistic intro to more pessimistic one ;)

tdd11235813:
dammit, guess I was too hasty again. The actual host-device SFINAE concept is not there yet. I am not able to construct the meta-type `is_host_only</*TFnObj::operator()...*/>` for SFINAE-ing the correct wrapper depending on the _\_\_attribute___ `[[hc]]` or `[[cpu]]` or `[[hc]] [[cpu]]`. 
SFINAE-ing C++ function attributes with HCC/clang remains an open question.
Actually, the compiler is too strict without a reason, because in the end the device-stage will never be called at all, so why HCC is bothering it ?!

So, you could wrap the lambda where it is defined and where the host-device attribute is known, e.g. in the TaskCopyCpu in Copy.hpp:
```c++
ALPAKA_FN_HOST auto operator()() const
-> void                               
{                                     
// ...
  auto func =                                         
    [&](vec::Vec<DimMin1, ExtentSize> const & idx) { std::memcpy(/*...*/); };
// ...
  meta::ndLoopIncIdx(                       
    extentWithoutInnermost,               
    HipUseHostOnly<decltype(func)>(func)); //<-- Functor that only calls lambda when on host
```
This seems to be equivalent to call separate functions like `ndLoopIncIdxHostOnly`, because when you know, you are host-only, then you can call a host-only subroutine. 
However, the wrapper is more flexible, because all subsequent calls in the chain cannot fail. 
For the example above: if HCC requires the device function somewhere in the chain, then it is fine because it is just empty defined in the wrapper. 

So, how to proceed?! Wrapping all host-only/device-only lambdas?

psychocoderHPC:
I discussed this problem with @tdd11235813  offline. He will provide a function like `std::bind` e.g. `bindScope<PLACEMENT>()` to build the wrapper.

psychocoderHPC:
@tdd11235813 We need to check if a simple `inline` is solving the issue too. https://llvm.org/docs/CompileCudaWithLLVM.html#id13 The clang docu for CUDA-Clang says that if a `__host__ __device__` meta function is not inlined than the compiler will try to build both version (device and host) but if the meta function is inlined it is not plreplem to call only a `__host__` function within a `__host__ __device__` function if the function is only called from the host side.

Since HIP based on clang the behavior could be equal.

psychocoderHPC:
update: `inline` alone not helped but using `__attribute__((always_inline)) __device__ __host` for HIP solved many issues.
Maybe `__forceinline__` defined [here](https://github.com/ROCm-Developer-Tools/HIP/blob/673617f1cd5ff01dd9d9918a9273c33f03a5babe/include/hip/hcc_detail/host_defines.h#L55) will also work.

tdd11235813:
none of these options worked:
- `__forceinline__` to all `ALPAKA_FN_*` flags for host/device attributes
- `__attribute__((always_inline))` to all `ALPAKA_FN_*` flags for host/device attributes
- two separate functions instead of one function with combined `__host__ __device__` (NdLoop*) (would have been surprising ..)

sbastrakov:
Discussed offline with @psychocoderHPC about [his earlier message](https://github.com/ComputationalRadiationPhysics/alpaka/pull/736#issuecomment-494854111), a wrapper like `std::function` is probably preferable over smth like `std::bind`.

tdd11235813:
k, looking into this. Btw, I also had a look on `nvfunctional`, but it seems, it still does not help to remove the flags for kernels/lambdas in Alpaka and it has many restrictions and drawbacks anyways (refs: [CUDA doc polymorphic-function-wrappers](https://docs.nvidia.com/cuda/cuda-c-programming-guide/#polymorphic-function-wrappers), [kokkos lambdas wiki2017](https://github.com/kokkos/kokkos/wiki/Lambda-Dispatch#lambdas-in-kokkos)).
So, Kokkos also uses annotations like `KOKKOS_LAMBDA` for `[=] __device__`  and so on ([ref](https://github.com/kokkos/kokkos/blob/develop/core/src/Kokkos_Macros.hpp)).

psychocoderHPC:
fix will be provided by #869

PR #743, opened 2019-02-04
Use platform independant way of checking variable existance

### LINKED ISSUE #741: Fix bash variable existance checks on OSX
All of the `[ -v Variable ]` tests fail because it is not supported on osx. Those errors seem to be ignored but this leads to wrong arguments to cmake (none of the variables are passed to cmake) and therefore all defaults are used.

BenjaminW3:
Fixes #741


ax3l (approved):

bertwesarg:
`${ALPAKA_CI_GCC_VER+x}` evaluates also to `x` if `ALPAKA_CI_GCC_VER=` is set but the value is the empty string. Is this expected?

Also, I used the following to test for variable existence for ages:

```sh
test "${VAR:+set}" = set
```

Which has the minor benefit, that you can 'read' what is tested for, replace 'set' with whatever verb you prefer.

BenjaminW3:
> ${ALPAKA_CI_GCC_VER+x} evaluates also to x if ALPAKA_CI_GCC_VER= is set but the value is the empty string. Is this expected?

The originally used `-v` tests if the variable exists at all. It does not matter if it is empty. This is exactly what we want to test in those cases.

I simply used the solution that is mentioned in the [Stackoverflow issue](https://stackoverflow.com/questions/3601515/how-to-check-if-a-variable-is-set-in-bash) which @ax3l posted in the original issue.

psychocoderHPC (approved):

ax3l:
Yes, @bertwesarg 's answer is what I usually took before as well until we saw this SO proposed abbreviation. We use Bert's syntax quite often in scripts though, to init default values. I hope the current way is still clear enough to readers, since it's used all the time now :D

PR #748, opened 2019-02-11
Rename ExecXxx to TaskKernelXxx

### LINKED ISSUE #693: Rename executor to something different.
We may want to rename our executor classes/concept to something different due to the upcoming conflict with the C++2x feature with the identical name. Renaming would prevent confusion. 
We have to compare the two versions and see if they are really similar (what I do not believe). We have to define the responsibilities of our implementation and search for a better name.

Currently the executors combine two responsibilities:
* bind the work division parameters and the arguments to the kernel function object
* execute the kernel (backend specific)

I am currently looking at if we could separate both concerns.

BenjaminW3:
Fixes #693 

This makes it much more clear what those classes are doing and will prevent
confusion with the C++2x executors.

Renames:
* executor -> execution task
* `namespace exec` -> `namespace kernel` (now more consistent because the trait to create those kernel execution tasks was already in there)
* `ExecXxx` -> `TaskKernelXxx`


BenjaminW3:
Ready for approvals ;-)

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,7 +26,7 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
 Kernel Definition
 -----------------
 
-There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that the kernel entry point is required to be a function object which has to have the accelerator as first argument.
+There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that it has to return void and the kernel entry point is required to be a function object which has to have the accelerator as first argument.
 This means, the kernel is an object that has implemented the `operator()` member function and can be called like any other function.
 The following code snippet shows a basic example kernel function object.
```
sbastrakov, line 31: should be "...a basic example *of a* kernel..." (not related to this PR)

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,7 +26,7 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
 Kernel Definition
 -----------------
 
-There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that the kernel entry point is required to be a function object which has to have the accelerator as first argument.
+There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that it has to return void and the kernel entry point is required to be a function object which has to have the accelerator as first argument.
```
sbastrakov, line 29: Surely may fix this some time later, but I find this paragraph a little confusing. It starts with basically "A function has to be a function object which returns void", which is weird with both "function is function object" and the concept of "function object return type". The second part could be simplified as well. I suggest the following formulation:
A kernel entry point is [or is required to be, if we prefer this wording] a function object. Its function call operator is an alpaka accelerator function taking an accelerator as its first argument and returning void.

sbastrakov (approved):

doc/markdown/user/implementation/library/Usage.md
BenjaminW3: I will change the documentation issues in a follow-up PR.

PR #749, opened 2019-02-12
Enhance documentation

BenjaminW3:


doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,19 +26,24 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
...
 -----------------
 
-There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that it has to return void and the kernel entry point is required to be a function object which has to have the accelerator as first argument.
-This means, the kernel is an object that has implemented the `operator()` member function and can be called like any other function.
-The following code snippet shows a basic example kernel function object.
+A kernel is a special function object which has to conform to the following requirements:
+* it has to fulfill the std::is_trivially_copyable trait (has to be copyable via memcpy)
+* the `operator()` is the kernel entry point
+  * it has to be an accelerator executable function
+  * it has to return void.
```
ax3l:
```suggestion
  * it has to return `void`.
```

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,19 +26,24 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
 Kernel Definition
 -----------------
 
-There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that it has to return void and the kernel entry point is required to be a function object which has to have the accelerator as first argument.
-This means, the kernel is an object that has implemented the `operator()` member function and can be called like any other function.
-The following code snippet shows a basic example kernel function object.
+A kernel is a special function object which has to conform to the following requirements:
+* it has to fulfill the std::is_trivially_copyable trait (has to be copyable via memcpy)
```
ax3l:
```suggestion
* it has to fulfill the `std::is_trivially_copyable` trait (has to be copyable via memcpy)
```

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,19 +26,24 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
...
+* the `operator()` is the kernel entry point
+  * it has to be an accelerator executable function
+  * it has to return void.
+  * it's first argument has to be the accelerator (templated for arbitrary accelerator backends).
```
ax3l:
```suggestion
  * its first argument has to be the accelerator (templated for arbitrary accelerator backends).
```

ax3l (commented):
Love it, thanks! Minor inline comments

BenjaminW3:
I really love the new github `apply suggestions` feature and how it allows to apply all changes as batch in a single commit!

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -8,15 +8,15 @@ Accelerator Executable Functions
 
 Functions that should be executable on an accelerator have to be annotated with the execution domain (one of `ALPAKA_FN_HOST`, `ALPAKA_FN_ACC` and `ALPAKA_FN_HOST_ACC`).
 They most probably also require access to the accelerator data and methods, such as indices and extents as well as functions to allocate shared memory and to synchronize all threads within a block. 
-Therefore the accelerator has to be passed in as a constant template reference variable as can be seen in the following code snippet.
+Therefore the accelerator has to be passed in as a templated constant reference parameter as can be seen in the following code snippet.
```
sbastrakov, line 11: I'm not against it, just saying that I've probably never seen "templated" used in such constructs (e.g. also on line 34). On the other hand, it is a short way to express that.

sbastrakov: Ah, ofc I've never seen it, as "template" is only a noun, not adjective or verb, so "templated" makes no grammatical sense. Again, we might still keep it as it is not likely to cause confusion.

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,19 +26,24 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
 Kernel Definition
 -----------------
 
-There is no difference between the kernel entry point function and any other accelerator function in *alpaka* except that it has to return void and the kernel entry point is required to be a function object which has to have the accelerator as first argument.
-This means, the kernel is an object that has implemented the `operator()` member function and can be called like any other function.
-The following code snippet shows a basic example kernel function object.
+A kernel is a special function object which has to conform to the following requirements:
+* it has to fulfill the `std::is_trivially_copyable` trait (has to be copyable via memcpy)
```
sbastrakov, line 30: `memcpy` -> `std::memcpy` (since `std::` is already used in this sentence)?

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -26,19 +26,24 @@ ALPAKA_FN_ACC auto doSomethingOnAccelerator(
...
+* the `operator()` is the kernel entry point
+  * it has to be an accelerator executable function
+  * it has to return `void`.
+  * its first argument has to be the accelerator (templated for arbitrary accelerator backends).
```
sbastrakov, line 34: Think here it is relatively important to be precise and say "an accelerator", not "the accelerator".
Edit: actually, given the explanation in the parenthesis it is probably okay as is.

doc/markdown/user/implementation/library/Usage.md
```diff
@@ -8,15 +8,15 @@ Accelerator Executable Functions
 
 Functions that should be executable on an accelerator have to be annotated with the execution domain (one of `ALPAKA_FN_HOST`, `ALPAKA_FN_ACC` and `ALPAKA_FN_HOST_ACC`).
```
sbastrakov, line 9: Not sure if important, but maybe worth clarifying that here and later by "function" we actually mean both functions and function templates and usually actually the latter with at least an accelerator type as a template parameter.

doc/markdown/user/implementation/library/Usage.md
ax3l: c memcopy from `<cstring>` as well ;-)

doc/markdown/user/implementation/library/Usage.md
ax3l: Ok, then we can leave it as is

doc/markdown/user/implementation/library/Usage.md
ax3l: Actually we only mean `functors`, all the time. Maybe we just want to link the wikipedia article to `function objects`/`functors` once in the beginning: https://en.wikipedia.org/wiki/Function_object#In_C_and_C++

Maybe we can write something like:

While those annotations are used both with free standing functions as well as member functions, we recommend to write in the concept of [C++ functor classes](https://en.wikipedia.org/wiki/Function_object#In_C_and_C++) on which the `operator()` needs such an annotation (as well as further called member functions).

sbastrakov: ~~Do we? I thought only a kernel has to be a functor and internally it could also call normal functions, just properly annotated.~~ Edit: Sorry, originally missed the last paragraph of your message.

Isn't it better to link https://en.cppreference.com/w/cpp/utility/functional instead of the Wikipedia? Edit: maybe it is a bit too technical and Wiki is indeed better.

BenjaminW3: I explicitly wanted to avoid the usage of the word functor because it has different meanings in different contexts.
This paragraph should be the abstract explanation for functions, function templates and function objects containing one of those.

doc/markdown/user/implementation/library/Usage.md
BenjaminW3: Yes, really also C-memcpy.

doc/markdown/user/implementation/library/Usage.md
BenjaminW3: It is not uncommon but I am also no language lawyer ;-)
https://stackoverflow.com/questions/3575901/can-lambda-functions-be-templated
https://www.linguee.de/englisch-deutsch/uebersetzung/to+be+templated.html

doc/markdown/user/implementation/library/Usage.md
ax3l: Sounds legit, okay with me :)

doc/markdown/user/implementation/library/Usage.md
ax3l: I've read "templated" as adjective many times before, it's understandable slang and ok if you ask me: https://www.google.com/search?q="templated"+C%2B%2B

ax3l (approved):

PR #755, opened 2019-02-26
Adds sincos function to math

tdd11235813:
- currently only for doubles
- adds math/sincos directory
- HIP seems to support sincos, sincosf, ...
- (license header already uses MPL2)

WIP:
- not yet tested on all platforms (CPU, AMD)
- no unit test for now (create `test/unit/math/sincos/` ?)
- sincosf() ? sincospi()? sincospif()?

Using `sincos()` gives performance boost. Tested on a small DFT kernel in CUPLA and pure CUDA with both having same speedup from 48ms to 27ms when using sincos instead of sin and cos. 
Thanks to @psychocoderHPC for the hint to that function :)

Usage example (see API in [CUDA sincos](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html#group__CUDA__MATH__DOUBLE_1gbe0e6a063a8f38850b0323933cf3320b)):
```c++
double vsin, vcos;
sincos(arg, &vsin, &vcos);
```
The C-pointer interface is not good, just hacked it into alpaka to get our CUPLA code running. 
~~STL~~ [C++ Standard](https://stackoverflow.com/questions/24328173/c-libstd-compute-sin-and-cos-simultaneously) does not provide sincos.

**Edit**: the CPU backend in CUPLA does not call the kernel. The sincos is compiled anyway probably due to an include. Need to check these issues. CUPLA also needs a define for that sincos then.

**Edit-2**: after offline discussion with @psychocoderHPC CUPLA will not get a math layer, so one just has to call `alpaka::math::sincos(...)`.


BenjaminW3:
Thanks for this addition!
What is the behaviour of the CUDA builtin version if you pass in a nullptr for one of the two outputs?

'currently only for doubles' does not seem to be true. You also allow float.

Can you add tests for double and float?

tdd11235813:
- adds ALPAKA_ASSERT to check nullptr
- fixes result signature (reference => pointer)
- adds unit test for sincos

current state:
- CUDA/HIP still uses double-precision interface (going to to add float)
- going to test it on AMD

sbastrakov:
I understand it is super-late for this, so do not suggest to actually do the following change. Just maybe to discuss / keep in mind for future. 

Why do we like the C style with pointers and then have to check if they are zero, and not return a pair of two elements? Then it can be called like 
```cpp
double sinValue, cosValue;
std::tie(sinValue, cosValue) = sincos(1.2);
```

tdd11235813:
I also would prefer this. It actually was a quick hack and I wanted to have the same interface as CUDA does, which is where the C interface.came from. But it is not required for alpaka, so we could go for a C++ approach as you suggested.
Edit: I'll add a separate traits for that, so both interfaces are possible and the pointer check only happens in the C styled interface.

psychocoderHPC:
I aggree with @sbastrakov and we should use a C++ interface instead of a C interface. @tdd11235813 Could you please change it and use `tie`.

tdd11235813:
hrrr, get unspecified launch errors on CUDA side, even when I only move std::tuple around where no sincos is involved at all. I also tried invoking make_tuple on a higher level, but still fails. Debugger/cuda-memcheck refers to illegal address, so I guess something is lost on the return path or I missed something. Need a break, will look again into it today or tomorrow.

sbastrakov:
Ah, ofc when suggesting I failed to think of nvcc.

psychocoderHPC:
@tdd11235813 Ok so please ~~keep it as pointer~~ reference for now. I missed that `std::tie` is only available on the host side. The `__device__` is missing on all methods of the class.

include/alpaka/math/sincos/SinCosCudaBuiltIn.hpp
```diff
@@ -0,0 +1,77 @@
...
+            };
+
+            //! sincos trait specialization.
+            template<>
+            struct SinCos<SinCosCudaBuiltIn, float>
+            {
+                __device__ static auto sincos(
+                    SinCosCudaBuiltIn const & sincos,
+                    float const & arg,
+                    float * result_sin,
```
psychocoderHPC: Could you please change the signature to references instead of pointer. This had the advantage that we do not need to check if the pointer is valid.

psychocoderHPC (dismissed):
can we please use references instead of pointer.

tdd11235813:
ok, no prob. Using refs now. Tested on AMD as well. Let's see if travis CI likes it, too. If you are happy, feel free to merge.

BenjaminW3:
What is the current state of the PR? There still some compilation issues.

psychocoderHPC:
@tdd11235813  and me have vocation. we will work on it next week.

tdd11235813:
although being on vacation I was curious, so for the record for next week: clang failed due to conversion to other precision:
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/test/unit/math/sincos/src/sincos.cpp:50:24: error: implicit conversion loses floating-point precision: 'double' to 'float' [-Werror,-Wconversion]
        FP check_sin = sin(arg);
           ~~~~~~~~~   ^~~~~~~~
```
and nvcc-8+gcc4.9 failed in debug case (?).
```
/alpaka/math/sincos/SinCosCudaBuiltIn.hpp(56): error: expression preceding parentheses of apparent call must have (pointer-to-) function type
```

sbastrakov:
I think the problem of conversion appears due to a little inaccuracy of calling `sin()`, which depending on implementation may result in calling the C-style `sin()` which only works on and returns doubles (as this is permitted to be in the global namespace). Specifying `std::sin()` or somehow else explicitly using the `std` version should solve that.

With the second issue there is a collision between function argument name `sincos` and the function `::sincos()` called inside. Not sure if it is a technically valid code, but just as a guess. The specialization for `float` just below does not have this collision.

tdd11235813:
- `std::sin()` cannot be used in real device codes (HIP also complains about it with error), so the alpaka math implementation for sin and for cos is required ~~(or I just implement a ifdef switch in the test for now)~~
- need to fix doxygen to match name changes (clang complains)
- sincos name conflict resolved (left a code comment, where renaming was mandatory, but also renamed other parts for the sake of consistency)

tdd11235813:
- now sin() and cos() come up with the same issue (no one tested these alpaka math functions before)
  - going to extend the renaming to other alpaka math functions as well (e.g. `sin`->`sin_builtin`)  
- other tests failed due to a travis issue (cuda repo could not be downloaded / stalled)

Edit:
- math fixes will go into a separate PR (=> PR #784)
- unit tests for math will be a separate issue

sbastrakov:
I think this is already resolved anyways, but just for a reference: probably [this](https://llvm.org/docs/CompileCudaWithLLVM.html#id7) explains why my initial suggestion with calling `std::sin()` does not work with CUDA.

tdd11235813:
travis looks good so far, so when all tests eventually passed, this can be merged.

BenjaminW3 (approved):

PR #760, opened 2019-03-12
Add test that parameters can be passed to a kernel by value and by const&

BenjaminW3:
Passing values by non-const reference is not allowed.

This does not change anything but it adds explicit tests describing what should be working.

### DIFF
--- a/test/unit/kernel/src/KernelWithAdditionalParam.cpp
+++ b/test/unit/kernel/src/KernelWithAdditionalParam.cpp
@@ -15,9 +15,8 @@
 
 #include <catch2/catch.hpp>
 
-
 //#############################################################################
-class KernelWithAdditionalParam
+class KernelWithAdditionalParamByValue
 {
 public:
     //-----------------------------------------------------------------------------
@@ -27,7 +26,7 @@ public:
     ALPAKA_FN_ACC auto operator()(
         TAcc const & acc,
         bool * success,
-        std::int32_t const & val) const
+        std::int32_t val) const
     -> void
     {
         alpaka::ignore_unused(acc);
@@ -37,7 +36,7 @@ public:
 };
 
 //-----------------------------------------------------------------------------
-struct TestTemplate
+struct TestTemplateByValue
 {
 template< typename TAcc >
 void operator()()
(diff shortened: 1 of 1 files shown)


test/unit/kernel/src/KernelWithAdditionalParam.cpp
```diff
@@ -48,13 +47,98 @@ void operator()()
...
+
+        REQUIRE(fixture(kernel, 42));
+    }
+};
+
+TEST_CASE("KernelWithAdditionalParamByConstRef", "[kernel]")
+{
+    alpaka::meta::forEachType<alpaka::test::acc::TestAccs>(
+        TestTemplateByConstRef());
+}
```
sbastrakov: Maybe add a newline at the end?

test/unit/kernel/src/KernelWithAdditionalParam.cpp
```diff
@@ -48,13 +47,98 @@ void operator()()
...
 
-TEST_CASE( "KernelWithAdditionalParam", "[kernel]")
+TEST_CASE("KernelWithAdditionalParamByValue", "[kernel]")
 {
-    alpaka::meta::forEachType< alpaka::test::acc::TestAccs >( TestTemplate() );
+    alpaka::meta::forEachType<alpaka::test::acc::TestAccs>(TestTemplateByValue());
 }
+
+/*
+Passing a parameter by non-const reference is not allowed.
```
sbastrakov: Maybe change `non-const reference` to the standard formulation `reference to non-const`?

test/unit/kernel/src/KernelWithAdditionalParam.cpp
BenjaminW3: fixed

test/unit/kernel/src/KernelWithAdditionalParam.cpp
BenjaminW3: done

BenjaminW3:
Ready for approvals ;-)

sbastrakov (approved):

Issue #601, opened 2018-08-09
Standalone/Auxiliary Headers


ax3l:
I think during the "install" of Alpaka, we should provide auxiliary headers such as "alpaka/backend/omp2b.hpp" that pre-define all CMake-set defines for users.

With such, user using interpreters such as `cling` or tools such as #600 can just do:
```C++
#include <alpaka/backend/omp2b.hpp>
#include <alpaka/alpaka.hpp>

#include <iostream>

// ...
```

and start learning/playing/debugging alpaka right away.

(Note: assuming the user starts e.g. a `cling -std=c++11 -x cuda` or `cling -std=c++11-x -fopenmp` powered kernel.)

The "auxiliary" headers could be very simplified regarding alpaka features. E.g., one could limit to single backends.

Maybe we also could/need to work some compiler-workaround activations into such headers, instead of cmake-setting them.

Do I miss something important here? I think that would be awesome!

ax3l:
cc @SimeonEhrig

sbastrakov:
I think it's a great idea! Also maybe helpful even for some "normal" (clone repo and build locally) alpaka users. I would probably use sometimes it to change quicker in my MSVS instead of re-doing cmake / changing preprocessor macro in project settings.

ax3l:
I think for general projects, CMake targets and even install is definitely the way to go.
Nevertheless, I want people to learn and experiment with Alpaka quickly with modern tools (binder, compiler explorer, etc.) and REPL, for which a simple include would be ideal (also linking is no issue there, no need to stay header-only).

SimeonEhrig:
I talked to Benjamin about the feature. We have the idea that we offer two solutions. One is Axel’s solution. Providing headers with defines for each backend. It should work without "cmake install".

The second solution is to create a custom header with cmake during the install time. The header should contain the defines of all backends which were available during install time. Additionally, the custom header should be automatically include by the `alpaka.hpp` header, so we only need to include the `alpaka.hpp` and all available backends are enabled.

Maybe the second feature should be disabled by default and activated via cmake flag.

BenjaminW3:
I will close this issue for now. The basic standalone headers have been implemented and we should create new tickets for extending this if necessary.

PR #761, opened 2019-03-12
Add Headers for standalone usage without CMake

BenjaminW3:
Implements standalone headers for #601

Please vote for your favorite:

Folder:
 * `backend`
 * ...

File names:
 * `BackendCpuThreads` (would be consistent with all other classes as they use the folder name as the prefix)
 * `EnableCpuThreads` (might be more natural language)
 * `CpuThreads` 

The defines themselves are already somehow inconsistemt as the include `ACC` but they do not only enable the compilation of an accelerator but a complete backend consisting of accelerator+kernel-execution-task+all other deps.

### DIFF
--- a/include/alpaka/standalone/CpuFibers.hpp
+++ b/include/alpaka/standalone/CpuFibers.hpp
@@ -1,0 +1,14 @@
+/* Copyright 2019 Benjamin Worpitz
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#ifndef ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLED
+    #define ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLED
+#endif
--- a/include/alpaka/standalone/CpuOmp2Blocks.hpp
+++ b/include/alpaka/standalone/CpuOmp2Blocks.hpp
@@ -1,0 +1,14 @@
+/* Copyright 2019 Benjamin Worpitz
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#ifndef ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED
+    #define ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED
+#endif
--- a/include/alpaka/standalone/CpuOmp2Threads.hpp
+++ b/include/alpaka/standalone/CpuOmp2Threads.hpp
@@ -1,0 +1,14 @@
+/* Copyright 2019 Benjamin Worpitz
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#ifndef ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED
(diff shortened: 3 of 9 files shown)


bertwesarg:
To use alpaka without Cmake, one also has to `#define ALPAKA_DEBUG`. At least I can't find a `#ifndef ALPAKA_DEBUG/#define ALPAKA_DEBUG`.

bertwesarg:
Would it make sense to put some basic error checking in these headers? I.e., the OpenMP backends could check for `_OPENMP`, plus versoin checks. `GpuCuda` could check `__CUDA__` or `__CUDACC__`. etc. p.p.

BenjaminW3:
If `#define ALPAKA_DEBUG` is really necessary then this is a bug and I will try to fix this.

BenjaminW3:
The checks for `_OPENMP` and the like are really useful and they are already part of alpaka. All the files relying on compiler support are guarded like this:

```c++
#ifdef ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED

#if _OPENMP < 200203
    #error If ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED is set, the compiler has to support OpenMP 2.0 or higher!
#endif

...

#endif
```

So it is not necessary to duplicate this within those headers because the errors will be created either way.

BenjaminW3:
@bertwesarg `core/Debug.hpp` contains
```
#ifndef ALPAKA_DEBUG
    //! Set the minimum log level if it is not defined.
    #define ALPAKA_DEBUG ALPAKA_DEBUG_DISABLED
#endif
```

It should be included from all headers using `ALPAKA_DEBUG` because else the self-contained-header check together with the error for warning on usage of undeclared defines would lead to a build error.

BenjaminW3:
You still can vote for the folder/file names (see top)!

ax3l:
cc @SimeonEhrig : it's coming, this should help you with convenient cling-ing of Alpaka!

ax3l:
I think `backend/` or `config/` are good folder names (we might need more helper headers besides backend selection, e.g. Debug, Verbosity, some kind of synchronization enforcement instead of streams, etc.).

I would say if it's in `backend/` then do not prefix the filename again `backend/CpuThreads.hpp`, otherwise in `config/BackendCpuThreads.hpp`.

We could also do `alpaka/enable/CpuThreads.hpp`, `alpaka/enable/Debug.hpp`, `alpaka/disable/CUDAstreams.hpp`, ... That's a bit like Apache configs and quite natural :D

bertwesarg:
> @bertwesarg `core/Debug.hpp` contains
> 
> ```
> #ifndef ALPAKA_DEBUG
>     //-----------------------------------------------------------------------------
>     //! Set the minimum log level if it is not defined.
>     #define ALPAKA_DEBUG ALPAKA_DEBUG_DISABLED
> #endif
> ```
> 
> It should be included from all headers using `ALPAKA_DEBUG` because else the self-contained-header check together with the error for warning on usage of undeclared defines would lead to a build error.

Sorry, I missed that than.

BenjaminW3:
We (@BenjaminW3 @bertwesarg @psychocoderHPC @sbastrakov @tdd11235813 @SimeonEhrig) voted offline for `standalone/CpuOmp2Threads.hpp`.

BenjaminW3:
I would like to get this first step merged independently of future enhancements (e.g. boost version compatibility detection etc.).
Please approve!

bertwesarg:
I'm fine with it.

sbastrakov (dismissed):

BenjaminW3:
Please approve!

ax3l:
Okay with me :)

ax3l (approved):

psychocoderHPC:
There is a discussion in an [cupla issue](https://github.com/ComputationalRadiationPhysics/cupla/pull/112#issuecomment-499020489) That the naming `standalone` is maybe not good. @ComputationalRadiationPhysics/alpaka-maintainers please join the discussion. I like to have the same naming in alpaka and cupla to not confuse the user.
So we need to collect arguments for different namings to evaluate if we should rename `alpaka/standalone` or keep the naming.

ax3l:
I think to avoid confusion with single-file, amalgated, dependency free "standalone header" projects [[1](https://github.com/mpark/variant/releases)] [[2](https://github.com/nlohmann/json)] [[3](https://github.com/catchorg/Catch2)] we should call those cmake-free headers "facade headers", which is the feature we are providing with them in common nomenclature [[4](https://github.com/include-what-you-use/include-what-you-use/blob/master/docs/IWYUPragmas.md#which-pragma-should-i-use)].

PR #762, opened 2019-03-13
Replace BOOST_CURRENT_FUNCTION with C++11 __func__

BenjaminW3:
Removes one more dependency to boost.

### DIFF
--- a/include/alpaka/core/Debug.hpp
+++ b/include/alpaka/core/Debug.hpp
@@ -12,8 +12,6 @@
 
 #include <alpaka/core/BoostPredef.hpp>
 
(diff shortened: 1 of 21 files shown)


sbastrakov:
Isn't `__func__` C99 and not C++11 ? Not sure if that should be a stopper, as now even MSVS seems to support it, just saying.

BenjaminW3:
You are right, [C99](https://en.cppreference.com/w/c/language/function_definition). Should not matter for this PR as long as all compilers implement it.

BenjaminW3:
Please approve!

ax3l (approved):

Issue #756, opened 2019-03-04
CMake Option: Alpaka CXX Standard


ivandrodri:
I have a problem to compile a module in c++14 with ALPAKA. 

I have an ALPAKA class that is compiled into a library, "AlpakaLibrary".

Now I would like to have my executable, "MAIN", compiled in c++14 and link AlpakaLibrary and another c++14 host-library, "c++14HostLibrary", to it. Something like this :

SET(_TARGET_NAME "MAIN")

ALPAKA_ADD_EXECUTABLE(
    ${_TARGET_NAME}
    ${_FILES_SOURCE})
TARGET_LINK_LIBRARIES(
    ${_TARGET_NAME}
    PUBLIC "AlpakaLibrary" "c++14HostLibrary")

I have two questions:

1) When I replace ALPAKA_ADD_EXECUTABLE --> ADD_EXECUTABLE I have many ALPAKA compilation errors and I don't understand why. Should I use always ALPAKA_ADD_EXECUTABLE ? 

2) I want to compile MAIN in c++14 but If I set the CMAKE flags manually, i.e.:

set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++14)

I have plenty of compilation errors for ALPAKA. 

So what is that right way to link an ALPAKA library with a c++14 module?

ax3l:
Hi,

> Should I use always ALPAKA_ADD_EXECUTABLE

yes, that's the current wrapper we deploy since CMake's `CUDA` language does not yet support `clang -x cuda` as a CUDA compiler: https://gitlab.kitware.com/cmake/cmake/issues/16586

For that reason, we manually add `-std=c++11` instead of using a "C++11 or newer" [`target_compile_feature(s)`](https://github.com/openPMD/openPMD-api/blob/0.7.1-alpha/CMakeLists.txt#L322-L324).

> compile MAIN in c++14

We probably have to expose a CMake control variable to allow setting another C++ standard for the time being, which can be set from apps.

These are the locations you could change already to C++14:
- https://github.com/ComputationalRadiationPhysics/alpaka/blob/0.3.5/alpakaConfig.cmake#L456
- https://github.com/ComputationalRadiationPhysics/alpaka/blob/0.3.5/alpakaConfig.cmake#L516

ivandrodri:
Thanks a lot!

ax3l:
[...]

PR #766, opened 2019-03-15
Allow to select the required C++ version via ALPAKA_CXX_STANDARD

BenjaminW3:
Implements #756

Necessary changes:
* Build issues in CI for builds using clang & libstdc++-5 & C++17 because the `libstdc++-5` that is used seems to be too old for some c++14 features: ```In file included from /home/travis/build/BenjaminW3/alpaka/test/integ/axpy/src/axpy.cpp:13:
/home/travis/build/BenjaminW3/alpaka/thirdParty/catch2/include/catch2/catch.hpp:631:53: error: no template named 'bool_constant' in namespace 'std'; did you mean '__bool_constant'?
        inline constexpr auto is_unique<T, Rest...> = std::bool_constant<
                                                      ~~~~~^```
=>use libstdc++-7 for c++17 builds
* Build issues in CI for builds using clang & libstdc++-7 & C++17 because boost.coroutine2 (used by boost.fibers) is incompatible with the stdlib (multiple apply methods issue)
=> disable boost.fibers for this combination
* Build issues in CI for builds using clang-CUDA & libstdc++: Due to the minimum requirement to compile with C++11 and because extensions are enabled by default, CMake adds `-std=gnu++11` instead of `-std=c++11`. Because alpaka is a `INTERFACE` library (header-only) we are [not allowed to set `CXX_EXTENSIONS` to `OFF`](https://gitlab.kitware.com/cmake/cmake/issues/17183) and transitively disable extensions for inherited targets. ```/usr/lib/gcc/x86_64-linux-gnu/5.5.0/../../../../include/c++/5.5.0/type_traits:311:39: error: __float128 is not supported on this target
    struct __is_floating_point_helper<__float128>```
=> define `__float128=void` for clang CUDA builds see [bug report](https://bugs.llvm.org/show_bug.cgi?id=13530#c6)


ax3l (commented):
Just to check, are we sure we need the `-std=c++XY` flag manually? Would we get around with only setting the target property `cxx_std_XY` or is this still an issue with clang cuda?
Because in that case, the gnu/std extensions could also be controlled easily with a target property and we could get rid of the CXX_FLAGS hacking.

ax3l:
I think the info in https://gitlab.kitware.com/cmake/cmake/issues/17183 for interface libraries is outdated. Recently, [CMake 3.11+ IIRC](https://cmake.org/cmake/help/v3.11/release/3.11.html#commands), imported interface libraries are allowed to be manipulated to some degree...

BenjaminW3:
As long as we are using `FindCUDA` for controlling nvcc we have to add `-std=c++XY` to `CUDA_NVCC_FLAGS`. For non-CUDA builds we do not set it anywhere. I am not sure about `HIP_HIPCC_FLAGS` but I would expect it to be the same.

BenjaminW3:
Yes, `INTERFACE_COMPILE_FEATURES` works and we are already using it but `INTERFACE_CXX_EXTENSIONS` does not work.

ax3l:
Should we make the `c++/gnu++` part in `-std=c++` just another option for now? Something like
```cmake
SET(ALPAKA_CXX_EXTENSION "c++" CACHE STRING "C++ standard extension (for nvcc and hip)")
SET_PROPERTY(CACHE ALPAKA_CXX_EXTENSION PROPERTY STRINGS "c++;gnu++")
```

~~We could probably also infer its default from the CMake defaults for a compiler (maybe via `get_target_property(VAR target CXX_EXTENSIONS)` or something else)~~

BenjaminW3:
The problem with the old solution was that alpaka defined `--std=c++11` instead of letting the user do this. In the new version I am using the CMake preferred version of setting requirements. With this we have no influence on the `-std` flag that is used. It is the user who sets the c++11 or gnu11 flags. We could use an intermediate solution and define `-std=c++${ALPAKA_CXX_STANDARD}` within alpaka, but this will not be future proof.
What I really want to achieve is to be able to remove `ALPAKA_CXX_STANDARD` completely.
In CI the user of alpaka which defines the c++-level is the unit test, which relies on CMake to set the flags depending on the maximum requirement of all transitive dependencies. CMake uses gnu if the user did not explicitly disable extensions on the target (not available for interface targets, only for real libs).
A real user could also manually set the std flag.

psychocoderHPC (approved):
Looks good from my side but please wait for @ax3l's go.

ax3l:
> With this we have no influence on the -std flag that is used. 

Yep, and that's a good thing since it's set by the end user or package manager.

> What I really want to achieve is to be able to remove ALPAKA_CXX_STANDARD completely.

Exactly my point.
As we still have to set it manually for nvcc/hip, the current PR still hard-codes `-std=c++XY` without gnu extensions in flags for CUDA (`HIP_HIPCC_FLAGS` and `CUDA_NVCC_FLAGS`). I thought we might want this to be configured as well for the meantime.

BenjaminW3:
You are right, we could possibly add an option to select the c++ mode for the nvcc builds. However, this is another feature request.

script/travis/before_install.sh
```diff
@@ -59,6 +59,30 @@ then
...
         fi
     fi
+
+    if [ "${ALPAKA_CI_STDLIB}" == "libstdc++" ]
+    then
+        if [ "${CXX}" == "clang++" ]
+        then
+            if [ ! -z ${ALPAKA_CXX_STANDARD+x} ]
+            then
+                if (( "${ALPAKA_CXX_STANDARD}" >= 17 ))
```
ax3l, line 69: I just realized: even if `ALPAKA_CXX_STANDARD` is set to `11`, with a new compiler this might select C++14 or even C++17 or C++2a with `target_compile_features`

PR #769, opened 2019-03-27
Add support for clang 8.0

BenjaminW3:
Clang now warns if a `= default` method is deleted by the compiler due to other restrictions (e.g. member variable that can not be copied).
Furthermore it warns on empty expressions due to useless `;`. This was the case for some of our usages of test macros.


bertwesarg:
> Furthermore it warns on empty expressions due to useless `;`. This was the case for some of our usages of test macros.

The trick here is to change the macros not the callsite:

```C++
#define ALPAKA_CHECK(success, expression) \
    do \
    { \
        if(!(expression)) \
        { \
            printf("ALPAKA_CHECK failed because '!(%s)'\n", #expression); \
            success = false; \
        } \
    } while ( 0 )
```

Now it requires a ';' and it could be converted to an no-op:
 
```C++
#define ALPAKA_CHECK(success, expression) do { } while ( 0 )
```

BenjaminW3:
Yes, this would also work, but I do not yet see an advantage of introducing more syntax elements just to silence this warning ;-)

bertwesarg:
> Yes, this would also work, but I do not yet see an advantage of introducing more syntax elements just to silence this warning ;-)

But my proposal would result in a much smaller patch, and the misusage of the macro would be cought by all compilers. It also is more pleasing to the programmer eyes, as _function calls_ without a terminating ';' look odd.

psychocoderHPC:
> > Furthermore it warns on empty expressions due to useless `;`. This was the case for some of our usages of test macros.
> 
> The trick here is to change the macros not the callsite:
> 
> ```c++
> #define ALPAKA_CHECK(success, expression) \
>     do \
>     { \
>         if(!(expression)) \
>         { \
>             printf("ALPAKA_CHECK failed because '!(%s)'\n", #expression); \
>             success = false; \
>         } \
>     } while ( 0 )
> ```
> 
> Now it requires a ';' and it could be converted to an no-op:
> 
> ```c++
> #define ALPAKA_CHECK(success, expression) do { } while ( 0 )
> ```

I do not like the C way to add a while loop to enforce a semicolon. I prefer the shorter version

```C++
#define FOO() \
{           \
// code     \
}           \
( (void) 0 )
```

psychocoderHPC:
@BenjaminW3 Like @bertwesarg  I would also prefere to keep the semicolon after a macro, else it looks not natureal and you need alsways to know where you must/can add a semicolon.

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -37,9 +37,9 @@ ALPAKA_FN_ACC auto testAtomicAdd(
                 acc,
                 &operand,
                 value);
-    ALPAKA_CHECK(*success, operandOrig == ret);
+    ALPAKA_CHECK(*success, operandOrig == ret)
```
psychocoderHPC: please reintroduce the semicolon and enforce it with `do{}while(true)` or `((void) 0)`

psychocoderHPC (changes requested):

bertwesarg:
> ```c++
> #define FOO() \
> {           \
> // code     \
> }           \
> ( (void) 0 )
> ```

Duh, that are **two** statements. Even though your coding style requires always **oc-tokens** around _blocks_, it can't be enforced. Only GCC warns about macros which have multiple statements and are used as _blocks_. I.e.,

```c++
if ( foo )
    FOO();
```

psychocoderHPC:
@bertwesarg Yes you are right, I missed that important point.

bertwesarg:
lgtm now

bertwesarg (approved):

psychocoderHPC (approved):

psychocoderHPC:
thx for the update!

PR #775, opened 2019-04-06
Use newer boost (1.67+) for clang-7 & c++17 & libstdc++

BenjaminW3:
This combination was disabled when the C++17 support was added. A newer boost version resolves the issue.


ax3l (approved):
Ok but should we document this somewhere and check the versions match in our CMake file?
Because Alpaka still supports Boost 1.62.0+.

BenjaminW3:
I have not yet found a way to check the standard library and the version that is used in CMake because this is completely a decision of the user via additional compile flags.

ax3l:
Ah ok, then maybe let us at least document the requirement on dependencies in `README.md`?

BenjaminW3:
I will add it to the readme.

PR #781, opened 2019-04-22
Improve clang-format support

BenjaminW3:
~This PR adds multiple things:~
* ~a `.clang-format` style that matches the current alpaka style as good as I was able to. We can easily change it later.~
* ~a CMake target `clang-format` for applying the `.clang-format` settings on the whole codebase (files in `include`, `test` and `example` folders with `hpp|h|cpp|c` extensions;  it does not format the `thirdParty` or other folders in the project.)~
* ~a CMake target `clang-format-check` which checks if the codebase is correctly formatted~
* ~a `CONTRIBUTING.md` file stating that clang-format is required for formatting the code. We can extend the file with other topics later.~

~Furthermore, the following things have been done:~
* ~the whole codebase was formatted~
* ~the CI will check that the code is correctly formatted, else the build will fail~

~On Windows the files to be formatted are not collected at configure-time because the number of files leads to command line length problems on Windows. Therefore the directories are iterated at runtime of the target. On linux files are collected at configure time.
CMake prints out a warning when the clang-format version is not 10.0 because this is the version enforced by CI.~

Improvements to the existing clang-format usage.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -99,7 +99,7 @@ append_recursive_files_add_to_src_group("${_ALPAKA_ROOT_DIR}/docs/markdown" "${_
 set_source_files_properties(${_ALPAKA_FILES_DOC} PROPERTIES HEADER_FILE_ONLY TRUE)
 
 append_recursive_files_add_to_src_group("${_ALPAKA_ROOT_DIR}/.github" "${_ALPAKA_ROOT_DIR}" "yml" _ALPAKA_FILES_OTHER)
-list(APPEND _ALPAKA_FILES_OTHER "${_ALPAKA_ROOT_DIR}/.gitignore" "${_ALPAKA_ROOT_DIR}/.zenodo.json" "${_ALPAKA_ROOT_DIR}/LICENSE" "${_ALPAKA_ROOT_DIR}/README.md")
+list(APPEND _ALPAKA_FILES_OTHER "${_ALPAKA_ROOT_DIR}/.clang-format" "${_ALPAKA_ROOT_DIR}/.gitignore" "${_ALPAKA_ROOT_DIR}/.zenodo.json" "${_ALPAKA_ROOT_DIR}/LICENSE" "${_ALPAKA_ROOT_DIR}/README.md")
 set_source_files_properties(${_ALPAKA_FILES_OTHER} PROPERTIES HEADER_FILE_ONLY TRUE)
 
 if(TARGET alpaka)
--- a/README.md
+++ b/README.md
@@ -203,6 +203,10 @@ consider citing us accordingly in your derived work and publications:
 }
 ```
 
+Contributing
+------------
+
+Rules for contributions can be found in [CONTRIBUTING.md](CONTRIBUTING.md)
 
 Authors
 -------
--- a/thirdParty/.clang-format
+++ b/thirdParty/.clang-format
@@ -1,0 +1,4 @@
+---
+DisableFormat: true
+SortIncludes: false
+...


psychocoderHPC:
Note: I will review it as soon as possible but I am currently ill and next week I have vocation.

sbastrakov:
I love this change, but feel not qualified in this matter to review it.

script/clang-format-check.sh
```diff
@@ -0,0 +1,23 @@
...
+    if [ ! -z "$comp_res" ] ; then
+        echo "$v failed the clang-format conformance check." >&2
+        res=1
+    fi
+done
+
+if [ $res -eq 0 ]; then
+    echo "Succeeded checking conformance to clang-format."
+else
+    echo "Failed checking conformance to clang-format." >&2
```
psychocoderHPC: Can we add the command which the user must run to autoformat the code to fix the error.

script/clang-format-check.sh
```diff
@@ -0,0 +1,23 @@
...
+fi
+
+exit $res
```
psychocoderHPC: EOF missing

script/clang-format-check.bat
```diff
@@ -0,0 +1,26 @@
...
+    )
+)
+
+if %res% EQU 0 (
+    echo Succeeded checking conformance to clang-format.
+) else (
+    echo Failed checking conformance to clang-format. >&2
+)
+
+exit %res%
```
psychocoderHPC: EOF missing

psychocoderHPC (changes requested):

script/clang-format-check.sh
BenjaminW3: This PR also adds a `clang-format` target which will do it for you.

BenjaminW3:
Could someone try if this really works on linux?

script/clang-format-check.sh
psychocoderHPC: But you need clang 8 for that, right? If so we should mention it in the error message.

psychocoderHPC:
I try to test it next week.

script/clang-format-check.sh
```diff
@@ -0,0 +1,23 @@
...
+res=0
+
+clang_format_binary="$1"
+
+shift
+
+for v in $@; do
+    comp_res=$("$clang_format_binary" -style=file "$v" | diff "$v" -)
+    if [ ! -z "$comp_res" ] ; then
+        echo "$v failed the clang-format conformance check." >&2
```
psychocoderHPC: please output `$comp_res` else it is not clear what is wrong

CMakeLists.txt
```diff
@@ -27,6 +27,73 @@ if(POLICY CMP0074)
...
+    /usr/local/bin
+    /usr/bin
+    "C:/Program Files/LLVM/bin"
+    "$ENV{CONDA_PREFIX}/Library/bin"
+    "${HOMEBREW_PREFIX}/bin"
+)
+
+find_program(
+    CLANG_FORMAT
+    NAMES "clang-format"
```
psychocoderHPC: you should search also for names like `clang-format-8` and maybe rate this `clang-format-8` higher than `clang-format`
I installed clang format 8 for my ubuntu 16.04 and do not have a binary `clang-format`

```
find_program(
    CLANG_FORMAT
    NAMES 
        "clang-format-9"
        "clang-format-8"
        "clang-format"
    PATHS ${CLANG_TOOLS_SEARCH_PATHS}
    DOC "Path to clang-format"
)
```

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+  AfterExternBlock: true
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces: false
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+BreakBeforeBinaryOperators: All
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: BeforeComma
```
psychocoderHPC: Should be `AfterComma` else you will have first a comma and than the line break

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+BreakInheritanceList: BeforeComma
```
psychocoderHPC: some shoudl be `AfterComma`

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+AlignTrailingComments: false
+AllowAllParametersOfDeclarationOnNextLine: false
+AllowShortBlocksOnASingleLine: false
+AllowShortCaseLabelsOnASingleLine: false
+AllowShortFunctionsOnASingleLine: Empty
+AllowShortIfStatementsOnASingleLine: false
+AllowShortLoopsOnASingleLine: false
+AlwaysBreakAfterReturnType: None
+AlwaysBreakBeforeMultilineStrings: true
+AlwaysBreakTemplateDeclarations: Yes
```
psychocoderHPC: maybe we need to add `PenaltyBreakTemplateDeclaration: 0` not sure if it is the default

script/clang-format-check.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/bash
```
psychocoderHPC: please make this file executable. `chmod 755` and than commit it

psychocoderHPC (changes requested):
note: clang format is also reordering the include order

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+BreakInheritanceList: BeforeComma
+BreakStringLiterals: true
+ColumnLimit: 0
+CompactNamespaces: false
+ConstructorInitializerAllOnOneLineOrOnePerLine: true
+ConstructorInitializerIndentWidth: 8
+ContinuationIndentWidth: 4
+Cpp11BracedListStyle: true
+DerivePointerAlignment: false
+FixNamespaceComments: false
```
psychocoderHPC: can we make this `FixNamespaceComments: true` I have very often hard times to see in which namespace a function is located if I scroll down.

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+IndentCaseLabels: false
+IndentPPDirectives: None
```
psychocoderHPC: I suggest `IndentPPDirectives: AfterHash` or `BeforeHash` since we have a lot of preprocessor switches in the code it is sometimes hard to see where they start and end

```C++
//---------------------------------------HIP-----------------------------------
// __HIPCC__ is defined by hipcc (if either __HCC__ or __CUDACC__ is defined)
#if !defined(BOOST_LANG_HIP)
#    if defined(__HIPCC__) && (defined(__CUDACC__) || defined(__HCC__))
#        include <hip/hip_runtime.h>
//HIP defines "abort()" as "{asm("trap;");}", which breaks some kernels
#        undef abort
// there is no HIP_VERSION macro
#        define BOOST_LANG_HIP BOOST_VERSION_NUMBER_AVAILABLE
#        if defined(BOOST_LANG_CUDA) && BOOST_LANG_CUDA
#            undef BOOST_LANG_CUDA
#            define BOOST_LANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
#        endif
#    else
#        define BOOST_LANG_HIP BOOST_VERSION_NUMBER_NOT_AVAILABLE
#    endif
#endif
```

psychocoderHPC:
`[&](TArgs const & ... args)` is changed by clang format into `[&](TArgs const &... args)` but can IMO not be avoided.

script/clang-format-check.sh
BenjaminW3: I will extend the error message.

script/clang-format-check.sh
BenjaminW3: What do you get from looking at the return code when you already know that it was nonzero?

CMakeLists.txt
BenjaminW3: Will add `clang-format-8`

.clang-format
BenjaminW3: I saw some really strange behaviour when indenting the preprocessor directives. Maybe it works now. However, this PR is not about the specific `.clang-format` file but about the tools. We can always enable this later.

BenjaminW3:
@psychocoderHPC Did the `make clang-format` target work for you?

script/clang-format-check.sh
psychocoderHPC: This variable contains the diff and show you all the wrong formated lines.

psychocoderHPC:
Yes make clang-format works if I change the find method like like I wrote above

script/clang-format-check.sh
BenjaminW3: I added printing the variable content

.clang-format
psychocoderHPC: sry I added the wrong keyword here: It must be `AfterColon` the suggested `AfterComma` is not existing.

.clang-format
psychocoderHPC: sry I added the wrong keyword here: It must be `AfterColon` the suggested `AfterComma` is not existing.

.clang-format
```diff
@@ -0,0 +1,77 @@
...
+  AfterExternBlock: true
+  BeforeCatch: true
+  BeforeElse: true
+  IndentBraces: false
+  SplitEmptyFunction: false
+  SplitEmptyRecord: false
+  SplitEmptyNamespace: false
+BreakBeforeBinaryOperators: All
+BreakBeforeTernaryOperators: true
+BreakConstructorInitializers: AfterComma
```
psychocoderHPC: This and the next line must be `AfterColon` I added the wrong keyword in my previous suggestion.

psychocoderHPC (dismissed):

ax3l:
You might want to add to your `CONTRIBUTING.md` how to add this check as a (client side) [git pre-commit hook](https://git-scm.com/book/pl/v2/Customizing-Git-Git-Hooks).

Alternatively, you can also rely on projects such as https://github.com/pre-commit/pre-commit with a little `.yaml` file and a lightweight python dependency.

Just an idea to make this more comfortable.

psychocoderHPC:
@BenjaminW3  How would we go on with this PR. as @ax3l mentioned we need at least information in `CONTRIBUTION.md`  about the tools and requirements.

BenjaminW3:
There is not yet a requirement to use formatting nor is any code fotmatted yet. I would only add documentation when it is finally enabled which is not part of this PR.

psychocoderHPC:
I added WIP to the PR name

BenjaminW3:
My original intention was to get this merged as is without adding formatting requirements. I will think about if we should format the source code together with the tools and the CI enforcement getting merged.

BenjaminW3:
I updated the PR and the descritption on the top.

BenjaminW3:
Ready to be approved/merged!

ax3l:
@BenjaminW3 little conflict

sbastrakov:
@BenjaminW3 @psychocoderHPC what is the status of this PR?

BenjaminW3:
It is still my goal to integrate this. I will try to update it when I find some time.

BenjaminW3:
I split off a refactoring I did in this PR and created #976.

BenjaminW3:
This PR originally wanted to add a test that checks whether all files are correctly formatted.
Now that we have support for github actions I am not sure if we should still enforce this for all PRs.
As a first step I will definitely add the `.clang-format` file and format the code base once.
After doing so, we have the following options:
1) trust in developers to format the code before merging and do not care if some lines are not correctly formatted
2) Add a check into PR build which ensures that the code is correctly formatted before it can be merged
3) Use an action that automatically runs a script after merging which formats all the files and adds a commit on top if there are any unformatted files

We could probably extend option 2 (but I do not yet have any experience doing so):
2b) Extend 2) by adding a bot command to PRs like `!format!` which you can write as a PR comment. This starts an action that automatically adds a commit with correctly formatted code to the PR.

@alpaka-group/alpaka-maintainers What do you think?
I would like to start with 1) so that we get a baseline.

SimeonEhrig:
I would suggest option two. The reason is really simple. It's quite likely that you get fake diffs using clang-format. If a developer runs clang-format over an entire file and parts of it are not formatted correctly before, diffs are created on code lines that the developer has never touched.

psychocoderHPC:
> * trust in developers to format the code before merging and do not care if some lines are not correctly formatted

I do not prefer this option. Same reason @SimeonEhrig said (code changes in not touched code parts). I like the option two more, were we check the format in a CI job and provide the user with the command he needs execute to format all code.


> We could probably extend option 2 (but I do not yet have any experience doing so):
2b) Extend 2) by adding a bot command to PRs like !format! which you can write as a PR comment. This starts an action that automatically adds a commit with correctly formatted code to the PR.

A bot would be the second step of the CI job where the code is only checked. This would allow us as maintainer to fix format issues  without user interaction.

bernhardmgruber:
I would like to revive this PR and at least get the .clang-format file merged. Are there still any objections against the state of the .clang-format checks?

Should we create a separate PR and just add the .clang-format file for now?

BenjaminW3:
I would also like to get the two `.clang-format` files merged and the repository formatted once even without checks in CI.
A new PR should be good for this

psychocoderHPC:
> I would also like to get the two `.clang-format` files merged and the repository formatted once even without checks in CI.
> A new PR should be good for this

We should not forget to say which clang-format version we require. There are small differences.
I would like to see a helper script we can run to be sure all folders with cpp and h files will always be formated.

bernhardmgruber:
> We should not forget to say which clang-format version we require. There are small differences.
> I would like to see a helper script we can run to be sure all folders with cpp and h files will always be formated.

While there are indeed small differences, they should be negligible. It should be rare that we run clang-format on the entire codebase. I recently found out that you can run `git clang-format ...` to just format your current changes, your changes on a branch etc. That keeps formatting local. So essentially, a user can run `git clang-format develop` on their branch and have all their changes, since they branched off develop, formatted.

For formatting everything, there is already a guide here: https://github.com/ComputationalRadiationPhysics/mallocMC/blob/dev/CONTRIBUTING.md

psychocoderHPC:
@BenjaminW3 What is missing to finish this PR? I would vote for bringing clang format in because we have more and more developers who working on various projects and formatting all correct is therefore impossible.

bernhardmgruber:
If finishing this PR is too hard, I would be super happy, if we could just add a new PR with just the `.clang-format` file. The CI checks can be postponed for later IMO.

psychocoderHPC:
Info: the final clang format file we will be using for all our projects will be voted https://github.com/ComputationalRadiationPhysics/contributing/pull/39

BenjaminW3:
I removed the parts of this PR which now have been added by other PRs. There are still some improvements left.
Please approve ;-)

bernhardmgruber (approved):

Issue #783, opened 2019-05-07
How do I handle multithreaded queues? 


krzikalla:
The following code shall synchronize dest with the content of src. It shall work with all types of queues (asynchronous and synchronous) and it shall work multithreaded.

```
enum { OUT_OF_SYNC, SYNCING, IN_SYNC };
std::atomic<int> state = OUT_OF_SYNC;

template<class T>
void StartSynchronize(T& queue)
{
  int expected = OUT_OF_SYNC;
  if (state.compare_exchange_strong(expected, SYNCING))
  {
    alpaka::mem::view::copy(queue, dest, src, size);
  }
}

template<class T>
void WaitSynchronize(T& queue)
{
  alpaka::wait::wait(queue);
  int expected = SYNCING;
  state.compare_exchange_strong(expected, IN_SYNC);
}  
```

The problem is, that QueueCpuSync::waitForThread is a noop and hence doesn't wait for the copy, which might be running in another thread. 
Is there a way to do the above with Alpaka means and independent of the particular queue type?

BenjaminW3:
If I understand this ticket and #768 correctly they are the same issue. Could you please answer my question I already gave you in the other ticket?

krzikalla:
OK, so we have to distinguish between asynchronous and synchronous queues:

```
template<class T>
void StartSynchronize(T& queue)
{
  int expected = OUT_OF_SYNC;
  if (state.compare_exchange_strong(expected, SYNCING))
  {
    alpaka::mem::view::copy(queue, dest, src, size);
    if (queue is synchron)
    { 
       state = IN_SYNC;
    }      
  }
}

template<class T>
void WaitSynchronize(T& queue)
{
  if (queue is synchron)
  { 
     while (state != IN_SYNC) {}
  }
  else
  {
    if (state != IN_SYNC)
    {
      alpaka::wait::wait(queue);
      state = IN_SYNC;
    }
  }
}  
```

How do I tell, if a queue is synchron? And am I allowed to call alpaka::wait::wait from multiple threads?

BenjaminW3:
> OK, so we have to distinguish between asynchronous and synchronous queues

This is not the goal of alpaka and if this is currently necessary e have to fix this.

> am I allowed to call alpaka::wait::wait from multiple threads?

I can not tell this for sure but by looking at the code this should work.

psychocoderHPC:
> > OK, so we have to distinguish between asynchronous and synchronous queues
> 
[...]

PR #786, opened 2019-05-11
Fix multithreaded usage of QueueCpuSync

### LINKED ISSUE #768: Thread sanitizer remark
Starting at alpaka-0.3.5/include/alpaka/stream/StreamCpuSync.hpp:177 we have:

  stream.m_spStreamImpl->m_bCurrentlyExecutingTask = true;
  task();
  stream.m_spStreamImpl->m_bCurrentlyExecutingTask = false;

If the enqueue() is called by multiple threads for the same stream, then the reads and writes to m_bCurrentlyExecutingTask are racing, since m_spStreamImpl is shared. I don't know what this variable is for beside the empty()-call below, but whatever it is, it won't work.

BenjaminW3:
Fixes #768 and #783

Currently the QueueCpuSync executes multiple tasks in parallel when called
from different threads. This violates the in-order-execution property of
the current queues.

### DIFF
--- a/include/alpaka/queue/QueueCpuSync.hpp
+++ b/include/alpaka/queue/QueueCpuSync.hpp
@@ -18,6 +18,9 @@
 #include <alpaka/queue/Traits.hpp>
 #include <alpaka/wait/Traits.hpp>
 
+#include <atomic>
+#include <mutex>
+
 namespace alpaka
 {
     namespace event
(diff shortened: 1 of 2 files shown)


BenjaminW3:
Ready to be merged!

psychocoderHPC (approved):

PR #792, opened 2019-05-21
fix cmake for tests

psychocoderHPC:
- fix path to internal `catch2`
- remove keyword to `general` (it is optional and the default, was also used with the wrong syntax)

@BenjaminW3 Why do we linked against `general`. I do not find any information on what kind of library it is.


BenjaminW3:
general is no lib but a [CMake keyword](https://cmake.org/cmake/help/v3.0/command/target_link_libraries.html). It is purely optional so it can be removed.

test/common/CMakeLists.txt
```diff
@@ -53,7 +53,7 @@ TARGET_COMPILE_OPTIONS(
     PUBLIC ${_COMMON_COMPILE_OPTIONS_PUBLIC})
 TARGET_LINK_LIBRARIES(
     ${_COMMON_TARGET_NAME}
-    PUBLIC alpaka ${_COMMON_LINK_LIBRARIES_PUBLIC})
+    PUBLIC "alpaka;${_COMMON_LINK_LIBRARIES_PUBLIC}")
```
psychocoderHPC: I looked again into the [cmake documentation](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) and it looks like this must be separated by spaces.
I will change it.

psychocoderHPC:
> general is no lib but a [CMake keyword](https://cmake.org/cmake/help/v3.0/command/target_link_libraries.html). It is purely optional so it can be removed.

Ahh the syntax was wrong. If you like to use `general` you must have a space before the library name `    LIST(APPEND _COMMON_LINK_LIBRARIES_PUBLIC "general ${CUDA_CUDA_LIBRARY}")`

test/common/CMakeLists.txt
psychocoderHPC: CMake is allowing a lists and space separated entries. Never the less I will change it to separate entries.

BenjaminW3 (approved):

PR #795, opened 2019-05-26
CI dependency version updates

BenjaminW3:
* Update CUDA 10.1 to Update 1 and add clang 8 build
* Remove ESan build option as it has not been used and removed from clang
* Update CMake 3.14.0 to 3.14.4


alpakaConfig.cmake
```diff
@@ -534,8 +534,8 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                             MESSAGE(FATAL_ERROR "NVCC 10.0 does not support clang 7+. Please use clang 4, 5 or 6!")
                         ENDIF()
                     ELSEIF(CUDA_VERSION VERSION_EQUAL 10.1)
-                        IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 8.0)
-                            MESSAGE(FATAL_ERROR "NVCC 10.1 does not support clang 8+. Please use clang 4, 5, 6 or 7!")
+                        IF(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 9.0)
+                            MESSAGE(FATAL_ERROR "NVCC 10.1 does not support clang 9+. Please use clang 4, 5, 6, 7 or 8!")
```
ax3l, line 538: That support has actually been added in CUDA 10.1 Update 1 - aka CUDA 10.1.168, not in the original CUDA 10.1 release https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

ax3l (approved):

alpakaConfig.cmake
BenjaminW3: Yes, I know. I do not know a way to differentiate those two versions within CMake.

ax3l: Does `CUDA_VERSION` not include a patch level?

PR #796, opened 2019-05-27
fix cuda warning

psychocoderHPC:
Fix CUDA warning that a host function is called from a host device function.

- remove in `alpaka/mem` all prefix usage `ALPAKA_FN_HOST_ACC` and the corresponding `ALPAKA_NO_HOST_ACC_WARNING`.

I am not sure why all host functions were prefixed with `ALPAKA_FN_HOST_ACC`. My tests with HIP and CUDA 10.0 and 9.2 shows they are not required.


BenjaminW3:
`alloc` was a HD function because we once wanted to add support for in-kernel memory allocation. We never did so, so we can remove it for now.

BenjaminW3 (approved):

PR #802, opened 2019-07-15
[WIP] Adds bindScope to control host-device context. Adds ALPAKA_LAMBDA.

tdd11235813:
This PR offers a solution to #736. HCC host-device restrictions tend to ripple through the call hierarchy like the "restrict" keyword from C++AMP. A heterogeneous call hierarchy, like we have, becomes host & device at some point, so callables need proper annotation and overloading of `operator()` to support both HCC and NVCC restrictions. While HCC lambdas must provide both host and device overloads within a host-device call chain, NVCC lambdas must come with host-device annotations directly. Thus, I have added `bindScope()` and `ALPAKA_LAMBDA`.

Note, that this behavior might change in the HIP-clang compiler, but I would not expect this at the moment (compiling HIP-clang still involves HCC, so I guess they will put this behavior into HIP-clang again).

- BindScope wraps a callable object and executes it only in the given scope.
  - bindScope() is no-op for non-HIP/non-CUDA backends
  - (BindScope objects are host-device though)
- ALPAKA_LAMBDA defines compiler-dependent host-device annotation + capture
  - is actually only required for CUDA to support context-embedded lambdas (see [cuda doc](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#extended-lambda))
  - adds compile-define ALPAKA_CXX_STANDARD for ALPAKA_CLASS_LAMBDA for C++17+
    - (ALPAKA_CLASS_LAMBDA currently not used, just referring to kokkos)
- extends lambda in CPU copy with bindScope() and ALPAKA_LAMBDA
- adds viewCopy to memView test
  - tests copying buffers in all directions (H2D, D2D, D2H, H2H)
  - involves bindScope() and ALPAKA_LAMBDA in CPU copy
- adds test for bindScope() and ALPAKA_LAMBDA

Example:
```c++
auto alpaka_callable = 
  alpaka::core::bindScope< alpaka::core::Scope::HostAndDevice > (
  ALPAKA_LAMBDA                           
  (TAcc const & acc,                      
    bool * success,                     
    std::uint32_t const & arg1)         
  -> void                                 
  {                                       
    alpaka::ignore_unused(acc); 
    ALPAKA_CHECK(*success, 42u == arg1);
  });
```

**Edit:** might help #708 (std::function / nvstd::function for existing functions). 
An existing function might be called from a bindScoped Alpaka lambda.
**Edit-2:** todo from travisCI: 
- [ ] fix unused parameters

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -977,6 +977,8 @@ IF(ALPAKA_ACC_GPU_HIP_ENABLE)
         LIST(APPEND _ALPAKA_LINK_FLAGS_PUBLIC "-Xcompiler ${OpenMP_CXX_FLAGS}")
     ENDIF()
     IF(ALPAKA_HIP_PLATFORM MATCHES "hcc")
+        # https://llvm.org/docs/AMDGPUUsage.html
+        # TODO: provide cmake interface for AMD GPU targets
         # GFX600, GFX601, GFX700, GFX701, GFX702, GFX703, GFX704, GFX801, GFX802, GFX803, GFX810, GFX900, GFX902
         SET(_ALPAKA_LINK_LIBRARIES_PUBLIC "${_ALPAKA_LINK_LIBRARIES_PUBLIC}" "--amdgpu-target=gfx803 --amdgpu-target=gfx900 --amdgpu-target=gfx906")
     ENDIF()
--- a/include/alpaka/core/BindScope.hpp
+++ b/include/alpaka/core/BindScope.hpp
@@ -1,0 +1,235 @@
+/* Copyright 2019 Benjamin Worpitz, Matthias Werner
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+
+#pragma once
+
+#include <alpaka/core/Common.hpp>
+#include <alpaka/core/Assert.hpp>
+
+#include <type_traits>
+
+
+namespace alpaka
+{
+    namespace core
+    {
+        /**
+         * Scopes are required for BindScope functor and the bindScope function.
+         */
+        enum class Scope {
+            HostOnly,
+            DeviceOnly,
+            HostDevice,
+            Default
+        };
+
+        /**
+         * BindScope wraps a callable object and executes it only in the given scope.
+         *
+         * A BindScope object can always be created both on host and on device.
+         * The callable is only called when the provided scope matches.
+         * The behaviour is undefined, when a method is called outside the specified scope.
+         *
+         * HIP(HC): BindScope adds missing host or device functions, otherwise HC will complain.
+         * Note, that in HIP/HC the host-device annotations belong to the function signature (attributes are applied to the function type).
+         *
+         * NVCC: BindScope invokes the callable only in the specified scope and the compiler stage,
+         * and calls ALPAKA_ASSERT(false) otherwise.
+         * A lambda still needs host or device annotation, use ALPAKA_LAMBDA.
+         */
+        template<typename TFunc, Scope TScope>
+        struct BindScope;
+
+        /**
+         * Returns BindScope wrapper callable for HIP and CUDA backends.
+         *
+         * For the other backends it just returns the callable itself.
+         */
+        template<Scope TScope, typename TFunc>
+        ALPAKA_FN_INLINE
+#if BOOST_LANG_CUDA || BOOST_LANG_HIP
+        auto bindScope( const TFunc callable ) -> BindScope<TFunc, TScope> {
+            return BindScope<TFunc, TScope>( callable );
+        }
+#else
+        auto bindScope( const TFunc callable ) -> TFunc {
+            // no host-device semantics, so no wrapper needed, just return the callable
+            return callable;
+        }
+#endif
+
+
+
+// --- HIP(NVCC) ---
+#if BOOST_COMP_NVCC || BOOST_COMP_CLANG_CUDA
+
+        // Default NVCC: enable host + device for callable.
+        template<typename TFunc, Scope TScope>
+        struct BindScope {
+            const TFunc m_callable;
+
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC
+            explicit BindScope(const TFunc callable) : m_callable{callable} {}
+
+            ALPAKA_NO_HOST_ACC_WARNING
+            template<typename... TArgs>
+            ALPAKA_FN_HOST_ACC
+            auto operator()(TArgs&&... args) const -> decltype(m_callable(std::forward<TArgs>(args)...)) {
+                return m_callable(std::forward<TArgs>(args)...);
+            }
+        };
+
+        // HostOnly NVCC:
+        template<typename TFunc>
+        struct BindScope<TFunc, Scope::HostOnly> {
+            const TFunc m_callable;
+
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC
+            explicit BindScope(const TFunc callable) : m_callable{callable} {}
+
+#if BOOST_ARCH_PTX
+            ALPAKA_NO_HOST_ACC_WARNING
+            template<typename... TArgs>
+            ALPAKA_FN_HOST_ACC
+            auto operator()(TArgs&&... args) const -> void { // void, otherwise callable() must be invoked for decltype()
+                ALPAKA_ASSERT(false);
+                alpaka::ignore_unused(args...);
+            }
+#else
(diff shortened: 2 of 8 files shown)


alpakaConfig.cmake
```diff
@@ -1005,6 +1007,8 @@ IF(NOT TARGET "alpaka")
         INTERFACE cxx_std_${ALPAKA_CXX_STANDARD}
     )
 
+    target_compile_definitions("alpaka" INTERFACE ALPAKA_CXX_STANDARD=${ALPAKA_CXX_STANDARD})
```
BenjaminW3: We should not add this here generally. `target_compile_features` (that we are already using) should be the way forward.
If this is necessary, the inheriting target probably did something wrong.

BenjaminW3:
👍 you found a solution, 👎 it is not your fault, but it is ugly...
I would want to release version 0.4.0 without this change before.

include/alpaka/core/BindScope.hpp
```diff
@@ -0,0 +1,244 @@
...
+#else
+  #define ALPAKA_LAMBDA [=]__host__ __device__
+  #if ALPAKA_CXX_STANDARD>14
+    #define ALPAKA_CLASS_LAMBDA [=,*this] __host__ __device__
+  #endif
+#endif
+
+#else // HIP(HCC) and CPU backends
+  // lambdas do not require a host-device attributes
+  #define ALPAKA_LAMBDA [=]
```
BenjaminW3: Could we remove the lambda capture from the macro and create a `ALPAKA_FN_LAMBDA` equivalent to `ALPAKA_FN_HOST_ACC` which only contains the attributes?

tdd11235813:
I only see either ugly or dirty solutions to this. It might be possible to hide the bindScope behavior into the macro. Then we would need `ALPAKA_FN_ACC_LAMBDA`, `ALPAKA_FN_HOST_LAMBDA` and maybe a generic `ALPAKA_FN_LAMBDA`. I am fine with excluding this from 0.4.0, also because testing HIP-clang is todo. A solution will be necessary for our applications though.

include/alpaka/core/BindScope.hpp
tdd11235813: for heterogeneous lambda kernels copy-by-reference is not really feasible, so always doing "[=]" + relying on compilers ability to only copy used variables was probably the idea, when kokkos introduced this macro. At least our lambdas are not only kernels and Alpaka is highly explicit anyways. So yes, we could leave it to the user in this current version.

psychocoderHPC: After thinking about it I would also prefer `ALPAKA_FN_LAMBDA` or 

```
#define ALPAKA_LAMBDA_CAPTURE( ... ) [=, __VA_ARGS__ ] __host__ __device__
// or ALPAKA_CAPTURE
```

tdd11235813:
See [changes here](https://github.com/tdd11235813/alpaka/commit/ec5881d11fb776b8fe4138371c5eaffa894fb5d5)
- macro shortcuts for bindscope

Edit: have to take @psychocoderHPC code cuz flags have to come after the capture, update in progress
Edit2: removed code snippet, updated below

sbastrakov:
@tdd11235813 that last code piece looks nice

tdd11235813:
- added [documentation](https://github.com/tdd11235813/alpaka/blob/pr-bindscope/doc/markdown/user/implementation/library/Usage.md#portable-lambdas-and-scope)
- today's [changes](https://github.com/tdd11235813/alpaka/compare/539a8dbf1e93a71f980858a2f3637f9027109aaf...bindscope)
- rebased with recent alpaka dev

From Usage.md doc:

<details>

```c++                                                                
// create a host-scoped callable from a lambda                        
auto alpaka_callable =                                                
  // function wrapper, returns a host-device aware callable           
  alpaka::core::bindScope< alpaka::core::Scope::Host > (              
    // annotation wrapper for lambda (default capture: [=])           
    ALPAKA_FN_LAMBDA (/* args */) /* -> return type */ { /*code*/ } );
                                                                      
// more readable with shortcuts for bindScope                         
ALPAKA_FN_SCOPE_HOST(                                                 
  ALPAKA_FN_LAMBDA (/* args */) { /*code*/ } );                       
                                                                      
// explicitly set captures (uncaptured variables will use [=])        
ALPAKA_FN_SCOPE_HOST(                                                 
  ALPAKA_FN_LAMBDA_CAPTURE(&my_capture) (/* args */) { /*code*/ } );  
```                                                                   
</details>

edit: doc snippet here is no more up-to-date

include/alpaka/core/Common.hpp
```diff
@@ -127,3 +127,28 @@
 #else
     #define ALPAKA_STATIC_ACC_MEM_CONSTANT
 #endif
+
+
+// ALPAKA_LAMBDA defines compiler dependent host-device annotation and capture
+// - from: https://github.com/kokkos/kokkos/wiki/Lambda-Dispatch
+#if BOOST_COMP_NVCC || BOOST_COMP_CLANG_CUDA
+
+#if BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(8, 0, 0)
```
BenjaminW3: We do not support nvcc versions lower than 8.0 anyways so this can probably be removed.

include/alpaka/core/Common.hpp
```diff
@@ -127,3 +127,28 @@
...
+  #define ALPAKA_FN_LAMBDA [=] __device__
+  #define ALPAKA_FN_LAMBDA_CAPTURE(...) [=,__VA_ARGS__] __device__
+#else
+  #define ALPAKA_FN_LAMBDA [=] __host__ __device__
+  #define ALPAKA_FN_LAMBDA_CAPTURE(...) [=,__VA_ARGS__] __host__ __device__
+  #if ALPAKA_CXX_STANDARD>=17
```
BenjaminW3: `ALPAKA_CXX_STANDARD` is not (and probably should not be) added as a preprocessor definition.
You have to rely on a different mechanism to find out if this is allowed.

include/alpaka/mem/buf/cpu/Copy.hpp
```diff
@@ -170,13 +171,13 @@ namespace alpaka
                             {
                                 meta::ndLoopIncIdx(
                                     extentWithoutInnermost,
-                                    [&](vec::Vec<DimMin1, ExtentSize> const & idx)
```
BenjaminW3, line 172: At least fo code where I know that it will always be executed on the host only, this probably makes the code slower. Currently we are capturing by reference, with `ALPAKA_FN_LAMBDA` a copy will be enforced.
I really would love to separate the capture list from the attributes.

include/alpaka/core/BindScope.hpp
```diff
@@ -0,0 +1,231 @@
...
+        };
+#endif // HCC
+
+    } // core
+} // alpaka
+
+// macro definitions to provide shortcuts
+#define ALPAKA_FN_SCOPE_HOST alpaka::core::bindScope< alpaka::core::Scope::HostOnly >
+#define ALPAKA_FN_SCOPE_DEVICE alpaka::core::bindScope< alpaka::core::Scope::DeviceOnly >
+#define ALPAKA_FN_SCOPE_HOST_AND_DEVICE alpaka::core::bindScope< alpaka::core::Scope::HostAndDevice >
```
BenjaminW3: Do we need the explicit `AND` in this macros name?

include/alpaka/core/Common.hpp
tdd11235813: is removed.

include/alpaka/core/Common.hpp
tdd11235813: no more using it. Removed class-specific C++17 capture, too. Not required at the moment.

include/alpaka/mem/buf/cpu/Copy.hpp
tdd11235813: host-lambdas may have ref-capture, but not host-device. nvcc will complain about that and `ALPAKA_FN_LAMBDA` is defined as host-device.
Offered solution: 
- host-lambdas are plain-style including capture, only need host-scope binding for HIP/HCC
- `ALPAKA_FN_LAMBDA` is for the non-host and includes `[=]`, cuz everything else would not compile with nvcc
- documentation on this updated

include/alpaka/core/BindScope.hpp
tdd11235813: no

tdd11235813:
marked as WIP, since HCC is deprecated and the new hip-clang might behave differently (more like cuda-clang), so do not merge. bindScope part might become useless in the near future. I've just got the proper build instructions for in-development hip-clang, but did not have time to test it yet.

BenjaminW3:
@tdd11235813 Is this still relevant or can this be closed?

tdd11235813:
closed as we are going to drop the deprecated HCC(AMD) for HIP, replacing it with the upcoming hip-clang(AMD). 
Edit: hip-clang is more friendly and does not require the proposed workarounds

PR #800, opened 2019-06-25
Fixes HIP issues

tdd11235813:
- fixes #794 (HIP(HCC) queue and time failures)
- fixes #735 (replaces deprecated hipLaunchKernel())
- fixes warnings
  - disables compiler warnings only for library includes
- fixes host-device attribute errors (adds missing ALPAKA_FN_HOST_ACC, ..)
- updates minimum version requirements for HIP
- tested on HIP(ROCm 2.5.0, HIP 1.5.19211)
- non-HIP change: see test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
  - removed move assignment operator
- HIP travis changes
  - use gcc5 instead of gcc4.9, and cuda 9.2 instead of cuda 8
    - otherwise issues with math lib/header
  - updates paths for HIP requirements (HIP_PLATFORM, ...)
  - installs rocRAND if not present (moved to run.sh because of the path variables)
  - fixes HIP version to roc-2.5.0 (was just "master" before)
- fixes cmake (HIP_PLATFORM, ALPAKA_HIP_PLATFORM)

=> tests run successfully on HIP(HCC) on AMD Vega64
=> tests run successfully on HIP(NVCC) on V100

Planned:
- separate PR for HIP/HCC lambda and host-device attribute issues
- separate PR for selecting HIP/AMD device architecture

Edit: updated lists and plan

PR #803, opened 2019-07-15
Atomics test reduced to 1D size_t accelerator set

tdd11235813:
This is actually more a question. Can the accelerator set be reduced to `<Dim=1D, Idx=size_t>` for atomics tests? It would save a lot of compile time. If so, PR would add this fix and it removes the restrictions for CI (in the hope compile-time is now feasible).

### DIFF
--- a/include/alpaka/atomic/Op.hpp
+++ b/include/alpaka/atomic/Op.hpp
@@ -243,7 +243,17 @@ namespace alpaka
                 {
                     auto const old(*addr);
                     auto & ref(*addr);
+
+// gcc-7.4.0 assumes for an optimization that a signed overflow does not occur here.
+// That's fine, so ignore that warning.
+#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC == BOOST_VERSION_NUMBER(7, 4, 0))
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wstrict-overflow"
+#endif
                     ref = ((old == compare) ? value : old);
+#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC == BOOST_VERSION_NUMBER(7, 4, 0))
+#pragma GCC diagnostic pop
+#endif
                     return old;
                 }
             };
--- a/test/unit/atomic/src/AtomicTest.cpp
+++ b/test/unit/atomic/src/AtomicTest.cpp
@@ -950,21 +950,19 @@ struct TestTemplate
     template< typename TAcc >
     void operator()()
     {
-        // This test exceeds the maximum compilation time on MSVC.
-#if !defined(ALPAKA_CI)
         TestAtomicOperations<TAcc, unsigned char>::testAtomicOperations();
         TestAtomicOperations<TAcc, char>::testAtomicOperations();
         TestAtomicOperations<TAcc, unsigned short>::testAtomicOperations();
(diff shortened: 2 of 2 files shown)


BenjaminW3 (dismissed):
Looks good.

tdd11235813:
hipRAND is not found in travis CI, probably some paths wrong, already looking into this. Wondering, how #800 passed then.
Edit: answer is travis cache^^ ... hip include issue will be fixed in #804

tdd11235813:
Do you see how to resolve this? Only happens with the gcc-7.4.0, see travis report.
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/test/unit/atomic/src/AtomicTest.cpp: In function 'testAtomicCas<alpaka::acc::AccCpuSerial<std::integral_constant<unsigned long, 1ul>, unsigned long>, long long>(alpaka::acc::AccCpuSerial<std::integral_constant<unsigned long, 1ul>, unsigned long> const&, bool*, long long)void':
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/atomic/Op.hpp:246:45: error: assuming signed overflow does not occur when simplifying conditional to constant [-Werror=strict-overflow]
                     ref = ((old == compare) ? value : old);
```
Code is:
```c++
ALPAKA_FN_HOST_ACC auto operator()(        
    T * addr,                              
    T const & compare,                     
    T const & value) const                 
-> T                                       
{                                          
    auto const old(*addr);                 
    auto & ref(*addr);                     
    ref = ((old == compare) ? value : old);
    return old;                            
}                                          
```

tdd11235813:
I am not convinced that this warning is an issue, the code looks clean to me and it only happens for gcc7.4.0, thus I ask for:
```c++
// only gcc-7.4.0 issues a strict-overflow warning                                                                                                                                                                                                            
#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC == BOOST_VERSION_NUMBER(7, 4, 0))                                                                                                                                                                                     
#pragma GCC diagnostic push                                                                                                                                                                                                                                   
#pragma GCC diagnostic ignored "-Wstrict-overflow"            
#endif                                                                                                                                                                                                
                    ref = ((old == compare) ? value : old);                                                                                                                                                                                                   
#if BOOST_COMP_GNUC && (BOOST_COMP_GNUC == BOOST_VERSION_NUMBER(7, 4, 0))                                                                                                                                                                                     
#pragma GCC diagnostic pop                                                                                                                                                                                                                                    
#endif
```

Edit: Updated code (was wrong before)

BenjaminW3:
Looks like the now reduced template hierarchy leads gcc-7.4 to do some more optimizations. One of the optimizations assumes that `signed overflow does not occur` and prints this warning.
It seems legitimate to supress it like you proposed.

tdd11235813:
ah ok, thanks. Branch is rebased with develop and updated including the comment.
```c++
// gcc-7.4.0 assumes for an optimization that a signed overflow does not occur here.
// That's fine, so ignore that warning.         
...                                    
```
Also tested gcc7.4.1, which did not have the warning. 
Edit: oh of course, now also gcc8 shows this warning in travis :/
Edit-2: no, it was a misread (just an unused-variable, cuz I missed if/endif in the code, thus code was only visible to gcc7.4.0)

BenjaminW3 (approved):

PR #804, opened 2019-07-16
Fixes hiprand issue in travis CI.

tdd11235813:
- fixed search path for hiprand include
- moved hiprand install routine back to `install_hip.sh` (it should still work when we will have more than one platforms to consider)

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -771,21 +771,23 @@ IF(ALPAKA_ACC_GPU_HIP_ENABLE)
                 ENDIF()
 
                 # random numbers library ( HIP(NVCC) ) /hiprand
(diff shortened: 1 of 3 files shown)


BenjaminW3 (approved):

tdd11235813:
I think it is still using cache (log shows cached docker file), and it does not even invoke the hip installer.
May I delete the travis cache on develop branch to rerun the HIP tests? (From reading [travis cache doc](https://docs.travis-ci.com/user/caching/) it seems to derive caches from the default branch develop).

BenjaminW3:
No problem, simply delete the cache

tdd11235813:
ok, after clearing the cache the remaining HIP tests passed. I did not rerun the non-hip backends, because the changes only affected the HIP part. So this can be merged.

PR #807, opened 2019-07-20
Add unit tests for std::function and std::bind kernels and remove from examples

BenjaminW3:
They are not supported on non CPU backends.


BenjaminW3:
Ready to be merged!

psychocoderHPC (approved):

PR #809, opened 2019-07-29
Template Depth: C++11 Default (1024)

ax3l:
Technically, we only have to set this for GCC<4.8 and Clang<4.0 (AppleClang<9.0), where this value was pretty low, e.g. 256.

Either way, we should not necessarily set it smaller than the minimum value required for compilers in C++11, which is 1024.

Refs:

- https://pybind11.readthedocs.io/en/stable/faq.html#recursive-template-instantiation-exceeded-maximum-depth-of-256
- https://bugs.llvm.org/show_bug.cgi?id=18417
- https://github.com/llvm/llvm-project/commit/e55b4737c026ea2e0b44829e4115d208577a67b2

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -883,12 +883,14 @@ ELSE()
         LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "general;rt")
     ENDIF()
 
-    # GNU
-    IF(CMAKE_COMPILER_IS_GNUCXX)
-        LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth-512")
-    # Clang or AppleClang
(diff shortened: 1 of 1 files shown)


alpakaConfig.cmake
```diff
@@ -885,10 +885,10 @@ ELSE()
 
     # GNU
     IF(CMAKE_COMPILER_IS_GNUCXX)
-        LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth-512")
+        LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth-1024")
```
ax3l: We can (and should) remove this for GNU if we are not supporting GCC<4.8 (which we probably do not).

BenjaminW3 (approved):

psychocoderHPC (approved):

PR #810, opened 2019-07-31
examples: Fix Acc choice

jkelling:


example/vectorAdd/src/main.cpp
```diff
@@ -73,21 +73,41 @@ class VectorAddKernel
     }
 };
 
+template<class A, class B>
+#if defined(ALPAKA_ACC_CPU_BT_OMP4_ENABLED)
+using DefaulAcc = alpaka::acc::AccCpuOmp4<A,B>;
```
sbastrakov: I guess there is a typo and `DefaulAcc` should be `DefaultAcc` everywhere?

example/vectorAdd/src/main.cpp
```diff
@@ -73,21 +73,41 @@ class VectorAddKernel
...
+using DefaulAcc = alpaka::acc::AccCpuOmp2Threads<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+using DefaulAcc = alpaka::acc::AccCpuSerial<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED)
+using DefaulAcc = alpaka::acc::AccCpuThreads<A,B>;
+#elif defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+using DefaulAcc = alpaka::acc::AccGpuCudaRt<A,B>;
+#else
+#define NOP
+#warning "No supported backend selected."
```
sbastrakov: It looks like this branch would not compile, as there is nothing after `template<class A, class B>`

example/vectorAdd/src/main.cpp
jkelling: Yes. Fixed.

example/vectorAdd/src/main.cpp
jkelling: Fixed.

example/vectorAdd/src/main.cpp
```diff
@@ -73,21 +73,42 @@ class VectorAddKernel
...
+using DefaultAcc = alpaka::acc::AccCpuFibers<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED)
+using DefaultAcc = alpaka::acc::AccCpuOmp2Threads<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+using DefaultAcc = alpaka::acc::AccCpuSerial<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED)
+using DefaultAcc = alpaka::acc::AccCpuThreads<A,B>;
+#elif defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+using DefaultAcc = alpaka::acc::AccGpuCudaRt<A,B>;
+#else
```
BenjaminW3: Is there any specific rule behind this "default" ordering? Should there be any? Currently `AccCpuSerial` was always used, now it will most probably use `AccCpuOmp4` when not building with GPU support enabled.

example/vectorAdd/src/main.cpp
```diff
@@ -73,21 +73,42 @@ class VectorAddKernel
...
+#define NOP
+#warning "No supported backend selected."
+#endif
+
 auto main()
 -> int
 {
-// This example is hard-coded to use the sequential backend.
-#if defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+#ifndef NOP
```
BenjaminW3: Currently the examples are hard-coded to use only one specific accelerator. This was done because my initial version which compiled the examples for all executors was seen as too complex and too verbose for beginners.
Now you want to use a middle way between "one hard-coded accelerator" and "all executors" which is "some default accelerator".
This will:
* make it a bit harder to understand the examples
* make it a bit easier to get the example to execute independent of your config

I would want to get this applied to all examples. The `DefaultAcc` selection code could be shared within a single header. One question would be where to store this header? Currently the examples are freestanding. It should be possible to copy them everywhere and they should still compile. However, such a `DefaultAcc` is nothing that should be part of the main `alpaka` library. It should probably also be renamed to `ExampleDefaultAcc` to not make it sound like something from the lib itself.

example/vectorAdd/src/main.cpp
```diff
@@ -73,21 +73,42 @@ class VectorAddKernel
...
     using Dim = alpaka::dim::DimInt<1u>;
     using Idx = std::size_t;
 
     // Define the accelerator
-    using Acc = alpaka::acc::AccCpuSerial<Dim, Idx>;
+    using Acc = DefaultAcc<Dim, Idx>;
     using DevAcc = alpaka::dev::Dev<Acc>;
     using PltfAcc = alpaka::pltf::Pltf<DevAcc>;
     using QueueAcc = alpaka::queue::QueueCpuBlocking;
+    std::cout << "Using alpaka device: " << typeid(Acc).name() << std::endl;
```
BenjaminW3: Could you please fix the text and write `accelerator` instead of `device`?

example/vectorAdd/src/main.cpp
jkelling: The ordering is the same as used in cmake, specifically the one you see when you open ccmake. So you need to have all backends in front of the one you want turned off.

example/vectorAdd/src/main.cpp
jkelling: I had the shared header in mind, when I wrote this. But I think it could be in alpaka to reduce the learning curve and similar backend selection code may be used helpful to reduce boilerplate in small alpaka application, which would have to replicate it otherwise.

example/vectorAdd/src/main.cpp
jkelling: Fixed.

Btw.: This line prints out the mangled name. Do you know a way to do some c++filt on it in-line? Strangle, the c++filt on my laptop does not demangle the name I get.

example/vectorAdd/src/main.cpp
BenjaminW3: The order in ccmake seems to be alphabetically as it is not the order they are [mentioned in CMake](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/alpakaConfig.cmake#L82-L88). You are also missing `ALPAKA_ACC_GPU_HIP_ENABLE`. Furthermore, the order in CMake still has no meaning and is mainly the order they were added.
We should come up with a good default order. @ComputationalRadiationPhysics/alpaka-maintainers Any wishes?

example/vectorAdd/src/main.cpp
BenjaminW3: `alpaka::acc::getAccName<Acc>()` could give you what you want.

BenjaminW3: Besides from this, there is no easy platform independent way to get the demangled name.

jkelling: Added the getAccName()

example/vectorAdd/src/main.cpp
sbastrakov: Having a good default order is a very good point.

I would put all GPU-based accelerators first, and then CPU-based. In each group would order by decreasing popularity. So start with GPU CUDA, then in the CPU group start with OpenMP from newest to oldest version, then other stuff.

example/vectorAdd/src/main.cpp
sbastrakov: I do not think having a shared header necessarily prevents the examples from being freestanding. In case we put this header somewhere inside `alpaka/example/...`, the additional include path could be set in CMakeLists.txt for examples after `FIND_PACKAGE("alpaka" REQUIRED)`?

On whether such a header should be part of alpaka I lean towards @BenjaminW3 's view. However, in case there are lots of code like that, maybe worth considering making a new "alpakautil" library for such stuff.

tdd11235813:
- would like it when the examples match the cmake Alpaka backend configuration (the single fixed CPU-backend was always a bit irritating to me w.r.t. Alpaka's goals, but I understand Benjamin's reasoning)
- could live with a shared header solution, I prefer freestanding examples though
- note, that `reduce` is an example which already uses cmake's accelerator configurations, in a different way to what is proposed here (imho the `reduce` example code is at the moment too "uncomfortable" anyways, so it is probably out of our discussion here)

jkelling:
I just noticed, that this would not compile for backends, which do not use the Cpu queue, as this is hard coded:
https://github.com/ComputationalRadiationPhysics/alpaka/blob/dd4070ccc55fe9dfc6c6eac9ba96f39504ea322c/example/vectorAdd/src/main.cpp#L109
Is there a trait, which returns the type of a default Queue one can use with a given platform or device? Maybe two of them: one for blocking and one for non-blocking?

psychocoderHPC:
> Is there a trait, which returns the type of a default Queue one can use with a given platform or device? Maybe two of them: one for blocking and one for non-blocking?

No currently wo do not have it. We discussed it in the last hackathon and I like to add it but did not find time for it :-(

psychocoderHPC:
@jkelling Could you please add the queues, depending on the accelerator,  to your accelerator selection part of this PR so that we can use this code with all accelerators. We need your PR soon because we need alpaka examples which are doing some calculations very soon.

jkelling:
There should be a trait implemented by the backends which returns an adequate queue for the choosen accelerator.

Specifically there should be either two (fir blocking and async) or an int template param which allows the caller describe what kind of queue they want using flags. Which version us better depends in whether there will be different variations on the blocking and async kinds in the future.

In any case, this feature needs to go into the alpake core before this PR can be fixed and go forward.

BenjaminW3:
@jkelling @psychocoderHPC The tests are using a [`DefaultQueue<TDev>`](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/test/common/include/alpaka/test/queue/Queue.hpp#L24-L89) trait which gives a default queue type given a device type.
This should work perfectly for the examples as well. However, for this to work we would have to move this into alpaka. I can also think of implementing some advanced traits to select a queue with the given requirements but this would take time.
The fastest way should be to move the marked lines into an `alpaka/default/DefaultQueue.hpp` header which is not included by `alpaka.hpp`. This way we could also place the `DefaultAcc.hpp` in this folder. What do you think?

psychocoderHPC:
@BenjaminW3: We should implement a trait where you can give in properties like `blocking`, `nonblocking` and an accelerator and you get the queue back. This will than also remove the `ifdef`'s in the user code.
This is also what we discussed at the last hackathon but I have currently no time to add this feature.

The `DefaultQueue.hpp` would be possible but I think a general trade is much more flexible and very useful for all user codes.

psychocoderHPC:
please have a look to #894, I implemented a queue selection function so changing the accelerator is a one-line change.

sbastrakov:
@jkelling what is the status of this PR? Now it should be easier to implement this idea due to simplifications for devices and queues.

jkelling:
I could pick this back up now. Apparently the things that need going are:
1. Put the definition of `alpaka::ExampleDefauiltAcc` into a header, maybe `~alpaka/example/include/alpaka/ExampleDefaultAcc.hpp` to keep it out of alpaka? Colud also be `~alpaka/example/include/alpaka/examples/ExampleDefaultAcc.hpp` to fit into the alpaka include tree if it were to be installed, but this seems to be a bit silly to me, because then we also just put it into alpaka.
2. Decide on an order. Going with what @sbastrakov suggested, I would propose GPU first, then parallel CPU and sequential CPU last (some test requires CpuSeq to be always enabled, maybe to create a host buffer, so it would always override anything put behind  it.).
3. Include this in all examples.

I would like some agreement on where to put the header to start.

bernhardmgruber:
Regarding a common header in the example folder: IIUC examples are currently standalone. I can just copy e,g, the vectorAdd folder (containing vectorAdd.cpp and CMakeLists.txt) and then build it somewhere else (given that alpaka is installed). Adding a header outside the vectorAdd folder now breaks this behavior. However, I have no strong feelings on this. Do as you like.

sbastrakov:
Good point @bernhardmgruber . So it seems the options are a common file for examples which makes them two-file, or putting this into alpaka itself. In the latter case the examples are still stand-alone, but the main library has this extra include file. What are your thoughts @BenjaminW3 ?

BenjaminW3:
I would be ok with an additional header within alpaka. `example/DefaultExecutor.hpp`?  (with the matching name space and typename)

jkelling:
Above `ExampleDefaultAcc` was suggested for the accelerator alias. I think `DefaultExecutor` would be misleading.

jkelling:
I added this to all examples, except reduce, which uses an additional abstraction layer for backends so that inserting the default accelerator does not work. Changes:
* callled the alias `ExampleDefaultAcc`
* left a commented line in each example which selects AccCpuSerial as an example for manual selection

include/alpaka/example/ExampleDefaultAcc.hpp
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+namespace alpaka
+{
+    namespace example
+    {
+        //! Alias for the default accelerator used by examples. From a list of
+        //! all accelerators the first one which is enabled is chosen.
+        //! AccCpuSerial is selected last.
+        template<class A, class B>
```
BenjaminW3: Could you please use the TDim and TIdx spelling that is used in the implementations?

example/vectorAdd/src/vectorAdd.cpp
```diff
@@ -72,13 +73,15 @@ class VectorAddKernel
     }
 };
 
+
```
BenjaminW3: Any reason for the new newline?

include/alpaka/example/ExampleDefaultAcc.hpp
```diff
@@ -0,0 +1,53 @@
...
+        using ExampleDefaultAcc = alpaka::acc::AccCpuOmp2Threads<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED)
+        using ExampleDefaultAcc = alpaka::acc::AccCpuThreads<A,B>;
+#elif defined(ALPAKA_ACC_CPU_BT_OMP4_ENABLED)
+        using ExampleDefaultAcc = alpaka::acc::AccCpuOmp4<A,B>;
+#elif defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+        using ExampleDefaultAcc = alpaka::acc::AccCpuSerial<A,B>;
+#else
+        class ExampleDefaultAcc;
+        #warning "No supported backend selected."
```
BenjaminW3, line 50: What is the advantage of warning over error here?

jkelling: The idea is that this header does not break compilation if included and none of the accerlators is enabled: maybe the user want to use custom one in the example. The warning should only help with understanding the errors which ensue if `ExampleDefaultAcc` is used with no accelerator being selected.

BenjaminW3: ok, thanks for lr the explanation.

BenjaminW3 (approved):

sbastrakov (approved):

PR #819, opened 2019-08-12
Use CMake 3.15 in CI

BenjaminW3:


BenjaminW3:
There seems to be a problem with clang used as CUDA compiler and CMake-3.15.2 missing to add the `-std=c++XX` compiler argument.
@ComputationalRadiationPhysics/alpaka-maintainers Is anyone capable of and interested in trying this out locally?

psychocoderHPC:
Not sure what we should test exactly: 

modules: `1) cmake/3.15.2   2) gcc/7.3.0   3) boost/1.68.0   4) clang/7.0.0   5) cuda/9.2`

# test compile

```
export CXX=clang++
# compiling cuda only
cmake ../alpaka/ -DBUILD_TESTING=ON
make VERBOSE=1
```

output:
```
...
clang/7.0.0/bin/clang++  -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_DEBUG=0 -D__float128=void -I/bigdata/hplsim/scratch/widera/alpaka/include -isystem /trinity/shared/pkg/devel/cuda/9.2/include  
-Werror -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-disabled-macro-expansion -Wno-global-constructors 
-Wno-padded --cuda-gpu-arch=sm_60 --cuda-path=/trinity/shared/pkg/devel/cuda/9.2 
-Qunused-arguments -Wno-unused-local-typedef -ffast-math -ffp-contract=fast 
-x cuda -o CMakeFiles/bufferCopy.dir/src/bufferCopy.cpp.o 
-c /bigdata/hplsim/scratch/widera/alpaka/example/bufferCopy/src/bufferCopy.cpp
...
```

As you said `std=c++11` is not set to the compiler and therefore it is not compiling.

BenjaminW3:
Now the question is: Do we do something wrong or is it CMake which broke our use-case?

ax3l:
Does we even need an explicit `-std` flag in that case? [Clang 6+](https://gist.github.com/ax3l/53db9fa8a4f4c21ecc5c4100c0d93c94) defaults to C++14 and if we request via CMake to have C++11 or newer than no flags needs to be set.

BenjaminW3:
@ax3l This seems not to be true when `-x cuda` is used. Possibly it uses C++98 as default when building CUDA? Maybe CMake expects the default to be c++11 now?

ax3l:
You are right, for some reason `clang -x cuda` defaults to C++98 (`199711L`) o.0
https://cuda.godbolt.org/z/Ngti3t

ax3l:
Submitted a bug report https://bugs.llvm.org/show_bug.cgi?id=43067

BenjaminW3:
Even though clang might have inconsistent defaults, the real bug is that CMake is not providing the correct std argument. We should probably tell them because there might not be many users and they will probably not find it out themselves. The hard part would be to find out if CMake needs to add a std argument or not depending on the other command line parameters.

ax3l:
Yes and IIRC, we are not using the CMake CUDA target for `clang -x cuda`, since CMake does not support another [CUDA compiler](https://gitlab.kitware.com/cmake/cmake/issues/16586) other than `nvcc` yet. So it should be in our CMake logic and not upstream, no?

BenjaminW3:
I will try to create a workaround on our sie first. On our side we know the CMake version, the Compiler and if we are compiling CUDA.

BenjaminW3:
I added the workaround for the CMake bug. Please approve.

sbastrakov (approved):

psychocoderHPC (approved):

psychocoderHPC:
It works!! I used [this example](https://github.com/ComputationalRadiationPhysics/alpaka/pull/819#issuecomment-522535413)

```
clang++  -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_DEBUG=0 
-D__float128=void -I/bigdata/hplsim/scratch/widera/alpaka/include -isystem /trinity/shared/pkg/devel/cuda/9.2/include  -Werror 
-Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic 
-Wno-disabled-macro-expansion -Wno-global-constructors -Wno-padded
--cuda-gpu-arch=sm_60 --cuda-path=/trinity/shared/pkg/devel/cuda/9.2 
-Qunused-arguments -Wno-unused-local-typedef -ffast-math 
-ffp-contract=fast -std=c++11 
-x cuda -o CMakeFiles/bufferCopy.dir/src/bufferCopy.cpp.o -c /bigdata/hplsim/scratch/widera/alpaka/example/bufferCopy/src/bufferCopy.cpp
```

PR #821, opened 2019-08-15
Clean up includes

BenjaminW3:
* reorder to conform to style guide most specialized to most generic
* only include what is required (In many places we only needed BoostPredef.hpp instead of Common.hpp)
* Remove unnecessary inconsistent use of ALPAKA_NO_HOST_ACC_WARNING for AcosCudaBuiltIn
* Add missing header files to reduce example IDE project
* remove double empty newlines after license header


sbastrakov (approved):
Not sure that cleanup of includes is necessarily connected to removing extra empty lines after license headers. Guess the logic is that both are part of the "file header".

psychocoderHPC (approved):

Issue #820, opened 2019-08-13
collective OpenMP queue


psychocoderHPC:
In #783 a workflow is described to use user spawned parallel regions to share the workload between different threads.

example how it can look like
```C++
    // Define the index domain
    using Dim = alpaka::dim::DimInt<1>;
    using Idx = size_t;

    // Define the accelerator
    using Acc = alpaka::acc::AccCpuOmp2Blocks<Dim, Idx>;
    using Dev = alpaka::dev::Dev<Acc>;

    using Queue = alpaka::queue::QueueCpuOmp2Collective;
    using Pltf = alpaka::pltf::Pltf<Dev>;

    auto dev = alpaka::pltf::getDevByIdx<Pltf>(0u);
    Queue queue(dev);

    std::vector<int> results(4, -1);

    using Vec = alpaka::vec::Vec<Dim, Idx>;
    Vec const elementsPerThread(Vec::all(static_cast<Idx>(1)));
    Vec const threadsPerBlock(Vec::all(static_cast<Idx>(1)));
    Vec const blocksPerGrid(results.size());

    using WorkDiv = alpaka::workdiv::WorkDivMembers<Dim, Idx>;
    WorkDiv const workDiv(
        blocksPerGrid,
        threadsPerBlock,
        elementsPerThread);

    // user is spawning a parallel region to controll the number of threads and avoid thread creation during the runtime from alpaka
    #pragma omp parallel num_threads(static_cast<int>(results.size()))
    {
        auto kernel =
        [&] ALPAKA_FN_ACC (
            Acc const & acc) noexcept
        -> void
        {
            size_t threadId = alpaka::idx::getIdx<alpaka::Grid, alpaka::Blocks>(acc)[0];
            // avoid that one thread is doing all the work
            std::this_thread::sleep_for(std::chrono::milliseconds(200u * threadId));
            results[threadId] = threadId;
        };

        // the kernel will be executed collectively 
        alpaka::kernel::exec<Acc>(
               queue,
               workDiv,
               kernel);

        alpaka::wait::wait(queue);
    }

    for(size_t i = 0; i < results.size(); ++i)
    {
        REQUIRE(static_cast<int>(i) == results.at(i));
    }
```

psychocoderHPC:
I have implemented it already in alpaka https://github.com/psychocoderHPC/alpaka/tree/topic-collectiveOmpQueue2 but I am not happy with the highly specific queue. 
There are many open questions e.g. 
[...]

PR #822, opened 2019-08-15
support external queue implementations

psychocoderHPC:
solves #820

By providing a queue interface class it is possible to handle user defines queues within alpaka.

- add queue interface `ICpuQueue`
- remove usage of hardcoded register function for each queue type within the device

The second commit provides a OpenMP collective queue as test case for an externally defined queue which is designed with the requirements of #820 in mind.

### DIFF
--- a/include/alpaka/queue/QueueCpuBlocking.hpp
+++ b/include/alpaka/queue/QueueCpuBlocking.hpp
@@ -12,6 +12,7 @@
 
 #include <alpaka/core/Unused.hpp>
 #include <alpaka/dev/DevCpu.hpp>
+#include <alpaka/queue/cpu/ICpuQueue.hpp>
 
 #include <alpaka/dev/Traits.hpp>
 #include <alpaka/event/Traits.hpp>
@@ -39,7 +40,7 @@ namespace alpaka
             {
                 //#############################################################################
                 //! The CPU device queue implementation.
-                class QueueCpuBlockingImpl final
+                class QueueCpuBlockingImpl final : public cpu::ICpuQueue
                 {
                 public:
                     //-----------------------------------------------------------------------------
@@ -59,6 +60,20 @@ namespace alpaka
                     //-----------------------------------------------------------------------------
                     ~QueueCpuBlockingImpl() = default;
 
+                    //-----------------------------------------------------------------------------
+                    void enqueue(std::shared_ptr<cpu::ICpuQueue> & iQueue, event::EventCpu & ev) final
+                    {
+                        auto spQueueImpl = std::static_pointer_cast<QueueCpuBlockingImpl>(iQueue);
+                        queue::enqueue(spQueueImpl, ev);
+                    }
+
+                    //-----------------------------------------------------------------------------
+                    void wait(std::shared_ptr<cpu::ICpuQueue> & iQueue, event::EventCpu const & ev) final
+                    {
+                        auto spQueueImpl = std::static_pointer_cast<QueueCpuBlockingImpl>(iQueue);
+                        wait::wait(spQueueImpl, ev);
+                    }
+
                 public:
                     dev::DevCpu const m_dev;            //!< The device this queue is bound to.
                     std::mutex mutable m_mutex;
@@ -77,7 +92,7 @@ namespace alpaka
                 dev::DevCpu const & dev) :
                     m_spQueueImpl(std::make_shared<cpu::detail::QueueCpuBlockingImpl>(dev))
             {
-                dev.m_spDevCpuImpl->RegisterBlockingQueue(m_spQueueImpl);
+                dev.m_spDevCpuImpl->RegisterQueue(m_spQueueImpl);
             }
             //-----------------------------------------------------------------------------
             QueueCpuBlocking(QueueCpuBlocking const &) = default;
@@ -217,3 +232,5 @@ namespace alpaka
         }
     }
 }
+
+#include <alpaka/event/EventCpu.hpp>
--- a/include/alpaka/queue/QueueCpuBlocking.hpp
+++ b/include/alpaka/queue/QueueCpuBlocking.hpp
@@ -61,17 +61,15 @@ namespace alpaka
                     ~QueueCpuBlockingImpl() = default;
 
                     //-----------------------------------------------------------------------------
-                    void enqueue(std::shared_ptr<cpu::ICpuQueue> & iQueue, event::EventCpu & ev) final
+                    void enqueue(event::EventCpu & ev) final
                     {
-                        auto spQueueImpl = std::static_pointer_cast<QueueCpuBlockingImpl>(iQueue);
-                        queue::enqueue(spQueueImpl, ev);
+                        queue::enqueue(*this, ev);
                     }
 
                     //-----------------------------------------------------------------------------
-                    void wait(std::shared_ptr<cpu::ICpuQueue> & iQueue, event::EventCpu const & ev) final
+                    void wait(event::EventCpu const & ev) final
                     {
-                        auto spQueueImpl = std::static_pointer_cast<QueueCpuBlockingImpl>(iQueue);
-                        wait::wait(spQueueImpl, ev);
+                        wait::wait(*this, ev);
                     }
 
                 public:
--- a/include/alpaka/queue/QueueCpuBlocking.hpp
+++ b/include/alpaka/queue/QueueCpuBlocking.hpp
@@ -38,9 +38,18 @@ namespace alpaka
         {
             namespace detail
             {
+#if BOOST_COMP_CLANG
+    // avoid diagnostic warning: "has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Werror,-Wweak-vtables]"
+    // https://stackoverflow.com/a/29288300
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wweak-vtables"
+#endif
                 //#############################################################################
                 //! The CPU device queue implementation.
                 class QueueCpuBlockingImpl final : public cpu::ICpuQueue
+#if BOOST_COMP_CLANG
+    #pragma clang diagnostic pop
+#endif
                 {
                 public:
                     //-----------------------------------------------------------------------------
@@ -57,8 +66,6 @@ namespace alpaka
                     auto operator=(QueueCpuBlockingImpl const &) -> QueueCpuBlockingImpl & = delete;
                     //-----------------------------------------------------------------------------
                     auto operator=(QueueCpuBlockingImpl &&) -> QueueCpuBlockingImpl & = delete;
-                    //-----------------------------------------------------------------------------
-                    ~QueueCpuBlockingImpl() = default;
 
                     //-----------------------------------------------------------------------------
                     void enqueue(event::EventCpu & ev) final
--- a/include/alpaka/queue/QueueCpuNonBlocking.hpp
+++ b/include/alpaka/queue/QueueCpuNonBlocking.hpp
@@ -11,6 +11,7 @@
 #pragma once
 
 #include <alpaka/dev/DevCpu.hpp>
+#include <alpaka/queue/cpu/ICpuQueue.hpp>
 
 #include <alpaka/dev/Traits.hpp>
 #include <alpaka/event/Traits.hpp>
@@ -42,7 +43,7 @@ namespace alpaka
             {
                 //#############################################################################
                 //! The CPU device queue implementation.
-                class QueueCpuNonBlockingImpl final
+                class QueueCpuNonBlockingImpl final : public cpu::ICpuQueue
                 {
                 private:
                     //#############################################################################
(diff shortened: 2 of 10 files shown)


psychocoderHPC:
CC-ing: @krzikalla

psychocoderHPC:
As I wrote in #820 I do not think that the highly specific collective queue implementation fitts into the generic concepts for alpaka. So IMO this is the way how to support the user requirements.

test/unit/queue/src/QueueTest.cpp
```diff
@@ -159,7 +161,27 @@ void operator()()
...
+        ,
+        std::tuple<alpaka::dev::DevCudaRt, alpaka::queue::QueueCudaRtBlocking>,
+        std::tuple<alpaka::dev::DevCudaRt, alpaka::queue::QueueCudaRtNonBlocking>
+#endif
+#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+        ,
+        std::tuple<alpaka::dev::DevHipRt, alpaka::queue::QueueHipRtBlocking>,
+        std::tuple<alpaka::dev::DevHipRt, alpaka::queue::QueueHipRtNonBlocking>
+#endif
+    >;
```
BenjaminW3, line 171: Why did you remove it completely from `test/queue/Queue.hpp`? I would prefer to still have the `TestQueues` there and only appending the `QueueCpuOmp2Collective` here.

BenjaminW3: Or simply add it to `TestQueues` there.

test/unit/event/src/EventTest.cpp
```diff
@@ -285,7 +286,28 @@ void operator()()
...
+        ,
+        std::tuple<alpaka::dev::DevCudaRt, alpaka::queue::QueueCudaRtBlocking>,
+        std::tuple<alpaka::dev::DevCudaRt, alpaka::queue::QueueCudaRtNonBlocking>
+#endif
+#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+        ,
+        std::tuple<alpaka::dev::DevHipRt, alpaka::queue::QueueHipRtBlocking>,
+        std::tuple<alpaka::dev::DevHipRt, alpaka::queue::QueueHipRtNonBlocking>
+#endif
+    >;
```
BenjaminW3, line 295: Same as in `QueueTest.cpp`

include/alpaka/queue/cpu/ICpuQueue.hpp
```diff
@@ -0,0 +1,46 @@
...
+            //#############################################################################
+            //! The CPU queue interface
+            class ICpuQueue
+            {
+            public:
+                //-----------------------------------------------------------------------------
+                ICpuQueue() = default;
+                //! enqueue the event into the given queue ------------------------------------
+                virtual void enqueue(std::shared_ptr<ICpuQueue> &, event::EventCpu &) = 0;
+                //! the given queue is waiting for the event-----------------------------------
```
BenjaminW3: Could you please use the current style with a separate line for the comment. This would make it easier to mass-replace them later.

include/alpaka/queue/cpu/ICpuQueue.hpp
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+
+#pragma once
```
BenjaminW3, line 10: Could you please remove the second new-line?

test/unit/queue/src/QueueTest.cpp
psychocoderHPC: I moved it to the tests because in this file defines traits like `IsBlockingQueue`. This make it impossible to have all trait definitions outside of the file `Queue.hpp` but I do not like to define the traits within `Queue.hpp` because I like to have clear borders between the "user code` (the collective queue) and alpaka.

include/alpaka/queue/cpu/ICpuQueue.hpp
psychocoderHPC: ^^ I do not find examples for the style but think I got what you mean

include/alpaka/queue/QueueCpuNonBlocking.hpp
```diff
@@ -73,6 +74,20 @@ namespace alpaka
                     //-----------------------------------------------------------------------------
                     ~QueueCpuNonBlockingImpl() = default;
 
+                    //-----------------------------------------------------------------------------
+                    void enqueue(std::shared_ptr<cpu::ICpuQueue> & iQueue, event::EventCpu & ev) final
+                    {
+                        auto spQueueImpl = std::static_pointer_cast<QueueCpuNonBlockingImpl>(iQueue);
```
BenjaminW3: Do we really need the separate `iQueue` parameter? We should have a `enqueue` and `wait` overload for the non-shared_ptr types. Then we only would have to pass `*this`.

test/unit/queue/src/QueueTest.cpp
BenjaminW3: But then please leave the alpaka-queues in there and append the new one.

BenjaminW3:
There is another option to the interface inheritance. We could probably give the `DevCpu` a template parameter for additional queues it has to support. It then would hold a vector of variants of the built-in queues and the new queues. However, I do not think that this would be better.

include/alpaka/queue/QueueCpuBlocking.hpp
```diff
@@ -217,3 +232,5 @@ namespace alpaka
         }
     }
 }
+
+#include <alpaka/event/EventCpu.hpp>
```
BenjaminW3, line 241: Why is this strange include necessary?

test/unit/queue/src/CollectiveQueue.cpp
```diff
@@ -0,0 +1,174 @@
...
+    // only one thread is allowed to flip the value
+    REQUIRE(numFlippedValues == 1u);
+    // only one value is allowed to differ from the initial value
+    REQUIRE(numNonIntitialValues == 1u);
+}
+};
+
+TEST_CASE( "queueCollective", "[queue]")
+{
+    TestCollective{}();
```
BenjaminW3: Do we need the separate class at all? It is only used in the templatized tests as workaround. We should be able to move the code directly in here.

test/common/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -0,0 +1,454 @@
+/* Copyright 2019 Axel Huebl, Benjamin Worpitz, Matthias Werner
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+
```
BenjaminW3: Please remove the second newline here as well.

test/common/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -0,0 +1,454 @@
...
+#pragma once
+
+#ifdef ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED
+
```
BenjaminW3, line 13: Do we need the check here similar to all other OMP dependant files?
```
#if _OPENMP < 200203
    #error If ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED is set, the compiler has to support OpenMP 2.0 or higher!
#endif
```

test/common/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -0,0 +1,454 @@
...
+        //-----------------------------------------------------------------------------
+        //! The test queue specifics.
+        namespace queue
+        {
+            namespace traits
+            {
+                //#############################################################################
+                //! The blocking queue trait specialization for a OpenMP2 collective CPU queue.
+                template<>
+                struct IsBlockingQueue<
```
BenjaminW3, line 450: Could we move the `IsBlockingQueue` trait (maybe as `IsBlocking`) for this and all the other queues to the main `alpaka` library?
Could we please rename this queue to `QueueCpuBlockingOmp2Collective` so hat it matches the style of all others?

include/alpaka/queue/QueueCpuBlocking.hpp
psychocoderHPC: This is needed to pass the header checks where we check that a header is including all dependencies e.g. https://github.com/ComputationalRadiationPhysics/alpaka/pull/822/files#diff-5b4348d85ae1c4f4c1c432f2aa3b97bdR74 will call wait with an event but the specialization for the meaning of wait is within `EventCpu.hpp`. But since event depends on queue and here we depend of knowing the event we need to pull in the specializations.

test/common/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
psychocoderHPC: > Could we please rename this queue to `QueueCpuBlockingOmp2Collective` so that it matches the style of all others?

I also thought at this name before but it is not blocking within a parallel region so I avoided to use blocking within the name

psychocoderHPC: > Could we move the `IsBlockingQueue` trait (maybe as `IsBlocking`) for this and all the other queues to the main `alpaka` library?

We can do it but outside of this PR.

BenjaminW3: But why does `IsBlockingQueue` then return true? A queue that is blocking is defined by the fact that a call to `queue::enqueue(queue, task);` (which is also hidden within `kernel::exec`) does not return before the task is executed. Is this true or false for this queue?

include/alpaka/queue/QueueCpuNonBlocking.hpp
psychocoderHPC: I will try it again. I started with that before but somehow I was unhappy with that. I do not know anymore why so I will try it again.

test/common/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
psychocoderHPC: because it depends where you use it (in a parallel region or not) outside it is blocking and to not move over tests those require a nonblocking queue it is set to blocking.

include/alpaka/queue/cpu/ICpuQueue.hpp
```diff
@@ -0,0 +1,45 @@
...
+        namespace cpu
+        {
+
+            //#############################################################################
+            //! The CPU queue interface
+            class ICpuQueue
+            {
+            public:
+                //-----------------------------------------------------------------------------
+                ICpuQueue() = default;
```
BenjaminW3: Do we really need the explicitly defaulted default construcor? I think this is not necessary for an abstract interface class.

include/alpaka/queue/QueueCpuNonBlocking.hpp
```diff
@@ -73,6 +74,18 @@ namespace alpaka
                     //-----------------------------------------------------------------------------
                     ~QueueCpuNonBlockingImpl() = default;
```
BenjaminW3: We should now remove this destructor as it is missing the virtual keyword and it now has a base class.

include/alpaka/queue/QueueCpuBlocking.hpp
```diff
@@ -59,6 +60,18 @@ namespace alpaka
                     //-----------------------------------------------------------------------------
                     ~QueueCpuBlockingImpl() = default;
```
BenjaminW3: We should now remove this destructor as it is missing the virtual keyword and it now has a base class.

test/common/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -0,0 +1,455 @@
...
+                    //-----------------------------------------------------------------------------
+                    QueueCpuOmp2CollectiveImpl(QueueCpuOmp2CollectiveImpl const &) = delete;
+                    //-----------------------------------------------------------------------------
+                    QueueCpuOmp2CollectiveImpl(QueueCpuOmp2CollectiveImpl &&) = delete;
+                    //-----------------------------------------------------------------------------
+                    auto operator=(QueueCpuOmp2CollectiveImpl const &) -> QueueCpuOmp2CollectiveImpl & = delete;
+                    //-----------------------------------------------------------------------------
+                    auto operator=(QueueCpuOmp2CollectiveImpl &&) -> QueueCpuOmp2CollectiveImpl & = delete;
+                    //-----------------------------------------------------------------------------
+                    ~QueueCpuOmp2CollectiveImpl() = default;
```
BenjaminW3: Here as well

BenjaminW3:
CI complains: `CollectiveQueue.cpp(57): error: __host__ or __device__ annotation on lambda requires --expt-extended-lambda nvcc flag`

BenjaminW3 (approved):

PR #828, opened 2019-08-23
Added tests for existing math operators

jkrude:
Tested StdLib-backend fully.
Tested Cuda-backend compiling.
Not tested Hip-backend.
Edit: Refractored Code according to https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md.


test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
+//
```
psychocoderHPC: please use the alpaka header for all files. example: https://github.com/ComputationalRadiationPhysics/alpaka/blob/fe28f1b20b863179623267326e28c2cb66b9283e/test/analysis/headerCheck/src/main.cpp#L1-L8

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+ * 3. Specify the alpaka implementation.
+ * 4. Define the range in which the operator should be testes against.
+ */
+
+/*
+ * if you need to add a new range you have to add it to the switch case
+ *  - in the kernel class
+ *  - in the TestTemplate
+ */
+
```
psychocoderHPC, line 44: please use only one empty line

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+#include <alpaka/alpaka.hpp>
+#include <alpaka/test/MeasureKernelRunTime.hpp>
+#include <alpaka/test/acc/Acc.hpp>
+#include <alpaka/test/queue/Queue.hpp>
+
+#include <iostream>
+#include <typeinfo>
+#include <cmath>
+
+/*
```
psychocoderHPC: could you please use doxygen `@file`

```
/** @file
```
http://www.doxygen.nl/manual/commands.html#cmdfile

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+ * atan2     | Y | R^2\{(0,0)}
+ * fmod      | Y | R^2\{(x,0)|x in R}
+ * max       | Y | R^2
+ * min       | Y | R^2
+ * remainder | Y | R^2\{(x,0)|x in R}
+ * sincos    | Y | R^2 -> return 2 R
+*/
+
+/*
```
psychocoderHPC: could you please merge all separate comments into the first

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+            TData * results
+    ) const
+    -> void
+    {
+        auto const gridThreadIdx(alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u]);
+        TData argX;
+        TData argY;
+        if(gridThreadIdx < numOps)
+        {
+            // sizeRes = numOps * sizeArgs
```
psychocoderHPC: is this dead code or should it be a comment?

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+            switch (structs[gridThreadIdx].range)
+            {
+                case Range::POSITIVE_ONLY:
+                    for(TIdx i(0); i < sizeArgs/2 -1; ++i)
```
psychocoderHPC: should be `sizeArgs / 2 - 1` (spaces around binary operators)

psychocoderHPC: please check also the later code to not spam with comments I will not mark each missing space

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+        using QueueAcc = alpaka::test::queue::DefaultQueue<DevAcc>;
+        using PltfHost = alpaka::pltf::PltfCpu;
+
+
+        std::cout << "\nTesting next AccType \n\n";
+        // the functions that will be tested
+
+
+        TestStruct<TAcc, Data> arr[] =
+                {/*     normal callback,    const callback, alpaka callback,                           definition range*/
```
psychocoderHPC: please use only 4 spaces for indentions.

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+        typename T>
+using std_func_ptr_const = T const & (*) (T const & , T const &);
+
+
+template<
+        typename TAcc,
+        typename T>
+struct TestStruct
+{
+    std_func_ptr<T> stdOp ;
```
psychocoderHPC: can you add some documentation e.g. that `stdOp` can be a `nullptr` and the behavior in that case?

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+                            argX = argsX[i];
+                        argY = argsY[i];
+                        results[i + gridThreadIdx* sizeArgs] = structs[gridThreadIdx].alpakaOp(acc, argX, argY);
+                    }
+                    break;
+
+                case Range::Y_NOT_ZERO:
+                    for(TIdx i(0); i < sizeArgs/2 -1; ++i)
+                    {
+                        if(i == sizeArgs/2 -1 || i == sizeArgs/2)
```
psychocoderHPC: `i == sizeArgs / 2 - 1 || i == sizeArgs / 2` will always be `false` because the loop condition to not allow this values for `i`

psychocoderHPC: same for `X_NOT_ZERO`

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+        {
+            for (Idx j(0); j < sizeArgs; ++j) {
+                pBufHostRes[j + i * sizeArgs] = -1;
+            }
+        }
+
+        // C++11 random generator for uniformly distributed numbers in [-100, 100].
+        std::random_device rd{};
+        std::default_random_engine eng{rd()};
+        std::uniform_real_distribution<Data> dist(0, 100);
```
psychocoderHPC: could you please define the `100` as `constexpr` in this file (maybe at the top) and use the variable name later in the code.

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+
+        // This is just for a better understanding which results are unchanged.
+        for (Idx i(0); i < numOps; ++i)
+        {
+            for (Idx j(0); j < sizeArgs; ++j) {
+                pBufHostRes[j + i * sizeArgs] = -1;
+            }
+        }
+
+        // C++11 random generator for uniformly distributed numbers in [-100, 100].
```
psychocoderHPC: - please move the comment to the corresponding line.
- please update the range to `[0; 100)`  see  line 247

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+        }
+        pBufHostArgsX[sizeArgs / 2 - 1] = 0.0;
+        pBufHostArgsY[sizeArgs / 2 - 1] = 0.0;
+
+        pBufHostArgsX[sizeArgs / 2] = -0.0;
+        pBufHostArgsY[sizeArgs / 2] = -0.0;
+
+        for (Idx i(sizeArgs / 2 + 1); i < sizeArgs; ++i)
+        {
+            pBufHostArgsX[i] = dist(eng) - 100;
```
psychocoderHPC: please add a comment: change range to `[-100;0)` for the second half of the container

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+        using PltfAcc = alpaka::pltf::Pltf<DevAcc>;
+        using QueueAcc = alpaka::test::queue::DefaultQueue<DevAcc>;
+        using PltfHost = alpaka::pltf::PltfCpu;
+
+
+        std::cout << "\nTesting next AccType \n\n";
+        // the functions that will be tested
+
+
+        TestStruct<TAcc, Data> arr[] =
```
psychocoderHPC: could you please use `std:Array` than we can later in the code use `array.size()` instead `Idx const numOps = sizeof(arr) / sizeof(TestStruct<TAcc, Data>);`

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+        auto memBufHostRes(alpaka::mem::buf::alloc<Data, Idx>(devHost, sizeRes));
+        auto memBufHostStructs(alpaka::mem::buf::alloc<TestStruct<TAcc, Data>, Idx>(devHost, extent));
+
+        Data *const pBufHostArgsX = alpaka::mem::view::getPtrNative(memBufHostArgsX);
+        Data *const pBufHostArgsY = alpaka::mem::view::getPtrNative(memBufHostArgsY);
+        Data *const pBufHostRes = alpaka::mem::view::getPtrNative(memBufHostRes);
+        TestStruct<TAcc, Data> *const pBufHostStructs = alpaka::mem::view::getPtrNative(memBufHostStructs);
+
+        // This is just for a better understanding which results are unchanged.
+        for (Idx i(0); i < numOps; ++i)
```
psychocoderHPC: please remove the space after `for`

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+};
+
+TEST_CASE("binaryOps", "[binaryOps]")
+{
+    using TestAccs = alpaka::test::acc::EnabledAccs<
+            alpaka::dim::DimInt<1u>,
+            std::size_t>;
+
+    alpaka::meta::forEachType< TestAccs >( TestTemplate() );
+}
```
psychocoderHPC: missing EOF

test/unit/math/unaryOps/src/unaryOps.cpp
```diff
@@ -0,0 +1,398 @@
+//
```
psychocoderHPC: please use the alkpaka license header

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,432 @@
...
+#include <typeinfo>
+#include <cmath>
+
+/*
+ * atan2     | Y | R^2\{(0,0)}
+ * fmod      | Y | R^2\{(x,0)|x in R}
+ * max       | Y | R^2
+ * min       | Y | R^2
+ * remainder | Y | R^2\{(x,0)|x in R}
+ * sincos    | Y | R^2 -> return 2 R
```
psychocoderHPC: since you are checking `sincos` now in this math unit tests we should remove the `sincos` unit test which we already have.

test/unit/math/unaryOps/src/unaryOps.cpp
```diff
@@ -0,0 +1,398 @@
...
+    {
+        using Dim = alpaka::dim::Dim<TAcc>;
+        using Idx = alpaka::idx::Idx<TAcc>;
+        using Data = double;
+        using DevAcc = alpaka::dev::Dev<TAcc>;
+        using PltfAcc = alpaka::pltf::Pltf<DevAcc>;
+        using QueueAcc = alpaka::test::queue::DefaultQueue<DevAcc>;
+        using PltfHost = alpaka::pltf::PltfCpu;
+
+
```
psychocoderHPC: please remove one line

test/unit/math/unaryOps/src/unaryOps.cpp
```diff
@@ -0,0 +1,398 @@
...
+                { &std::exp,    &alpaka::math::exp<TAcc,Data>,      Range::UNRESTRICTED      },
+                { &std::floor,  &alpaka::math::floor<TAcc,Data>,    Range::UNRESTRICTED      },
+                { &std::log,    &alpaka::math::log<TAcc,Data>,      Range::POSITIVE_ONLY     },
+                { &std::round,  &alpaka::math::round<TAcc,Data>,    Range::UNRESTRICTED      },
+                { &rsqrt<Data>, &alpaka::math::rsqrt<TAcc,Data>,    Range::POSITIVE_ONLY     },
+                { &std::sin,    &alpaka::math::sin<TAcc,Data>,      Range::UNRESTRICTED      },
+                { &std::sqrt,   &alpaka::math::sqrt<TAcc,Data>,     Range::POSITIVE_AND_ZERO },
+                { &std::tan,    &alpaka::math::tan<TAcc,Data>,      Range::UNRESTRICTED      },
+                { &std::trunc,  &alpaka::math::trunc<TAcc,Data>,    Range::UNRESTRICTED      }
+
```
psychocoderHPC: please remove

test/unit/math/unaryOps/src/unaryOps.cpp
```diff
@@ -0,0 +1,398 @@
...
+            for(Idx j(0);j < sizeArgs;++j)
+            {
+                pBufHostRes[j+i*sizeArgs] = -1;
+            }
+        }
+
+        // C++11 random generator for uniformly distributed numbers in {-100,..,100}
+        std::random_device rd{};
+        std::default_random_engine eng{ rd() };
+        std::uniform_real_distribution<Data> dist(0, 100);
```
psychocoderHPC: can we move the data generation to a separate file and share it between both new math unit tests

psychocoderHPC (changes requested):
Thanks for your first PR. Only a few small changes are required

test/unit/CMakeLists.txt
```diff
@@ -28,8 +28,10 @@ ADD_SUBDIRECTORY("core/")
 ADD_SUBDIRECTORY("event/")
 ADD_SUBDIRECTORY("idx/")
 ADD_SUBDIRECTORY("kernel/")
-ADD_SUBDIRECTORY("math/sincos/")
 ADD_SUBDIRECTORY("mem/buf/")
+ADD_SUBDIRECTORY("math/sincos/")
```
sbastrakov, line 31: It is probably a little more logical to put these 3 lines before the `"mem/..."` includes

test/unit/math/binaryOps/src/binaryOps.cpp
jkrude: Yes but than we have to specialize the size in the initialization, because there is no easy conversion from `T arr [] = {...}` to `std::array<T, ?> arr = {...}`

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,627 @@
...
+};
+
+
+struct TestTemplate
+{
+    template<
+        typename TAcc>
+    void operator()( )
+    {
+        using Data = double; // this can be changed to float without problems
```
tdd11235813: single + double precision have to be tested, so you could provide test cases with `TestTemplate<float>` and `TestTemplate<double>`
Edit: `<>` were accidentally parsed to HTML

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+/** List of all operators
+ * operator  | in std | definition |  range
+ * atan2     | Y | R^2\{(0,0)}
+ * fmod      | Y | R^2\{(x,0)|x in R}
+ * max       | Y | R^2
+ * min       | Y | R^2
+ * remainder | Y | R^2\{(x,0)|x in R}
+ *
+ * sincos is testes separately,
```
tdd11235813: - ... is "tested" ... 
- if sincos has a unique signature {(arg, &result1, &result2)}, then it is probably not needed to add another file like ternaryOps.cpp if there is only one candidate.

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,627 @@
...
+ * 2. If the std function has a const reference signature use this
+ * 3. Specify the alpaka implementation.
+ * 4. Define the range in which the operator should be testes against.
+ *
+ * If you need to add a new range you have to add it to the switch case
+ *  - in the kernel class
+ *  - in the TestTemplate
+ *  
+ *  If the wrong range is used,
+ *  the bahaviour depends on the individual implementation.
```
tdd11235813, line 42: behaviour

test/unit/math/binaryOps/src/binaryOps.cpp
```diff
@@ -0,0 +1,627 @@
...
+
+// used for all operators that need const references (like min())
+template<
+        typename T>
+using std_func_ptr_const = T const & (*) (T const & , T const &);
+
+/*!
+ * @struct TestStruct
+ * @tparam TAcc The type of the used accelerator, important for alpaka callback.
+ * @tparam T The data type (float || double).
```
tdd11235813, line 74: ok for now, but half precision could be an option too

test/unit/math/dataGen.hpp
```diff
@@ -0,0 +1,53 @@
...
+    template< typename Data >
+    auto fillWithRndArgs(
+        Data * buffer,
+        size_t size,
+        size_t range
+    ) -> void
+    {
+        std::random_device rd {};
+        std::default_random_engine eng { rd( ) };
+        std::uniform_real_distribution< Data > dist(
```
tdd11235813, line 37: this will generate unpredictable numbers each run and {the distribution} is [not platform portable](https://stackoverflow.com/questions/14840901/is-the-random-library-in-c11-portable).
I would like to know the numbers chosen for the test, so either we use hard-coded numbers or we might want to use seedable and platform-portable random numbers.

tdd11235813 (changes requested):

test/unit/math/binaryOps/src/binaryOps.cpp
sbastrakov: Unfortunately so. Only this [experimental one](https://en.cppreference.com/w/cpp/experimental/make_array) (I do not suggest to use it here).

test/unit/math/dataGen.hpp
BenjaminW3: +1 for a seed that is printed before the test is executed

BenjaminW3:
All the build jobs complain about some compilation issues. Are you working on it?

tdd11235813:
Actually I do not really like the code because of the functions pointers (it is also not working).
When talked about a solution in the first time, we had arrays in mind to describe the set of functions.
The function pointers' host-device interoperability actually would require more dirty code ([example](https://leimao.github.io/blog/Pass-Function-Pointers-to-Kernels-CUDA/)) and it becomes more and more ugly and it is better to remake it in a good and functional C++ style now :)
(Edit: this is not the fault of @jkrude, I just did not have enough time to think about good design)

I discussed with @psychocoderHPC again, and that I like the simple row-wise-like enumeration of the related math functions+range.
Functors would be more easy to handle for host-device compiling, we just need a way to compact code, so one does **not** have to write dozens of functor boilerplate codes.

```c++
// sketching a possible functor layout for math test
struct TestOpFoo {
  static constexpr Range range = Range::Unrestricted;
  static constexpr OpType optype = OpType::Unary;

  ALPAKA_FN_ACC
  template<typename T>
  auto operator() ( T arg1 ) -> T {
     return alpaka::math::foo( arg1 );
  }

  ALPAKA_FN_HOST
  template<typename T>
  auto expect( T arg1 ) -> T {
     return std::foo(arg1);
  }
};

// creating a tuple with all the Ops, passing it to single kernel
// OR using a compile-time foreach, invoking a kernel for each functor 
...
```

Then for each math function a functor has to be written and registered for the test kernel. Just need a clean way to compact functors and their processing ...

Btw, code-style should be discussed, too ( eg. curly braces for initialization, except for elemental types, where assignment is ok, like in for-loops ).

Edits: typo + edit after 1st paragraph

tdd11235813:
re-written and pushed in #847 separately, so closing this.

PR #829, opened 2019-08-26
Fix #750 by adding a dummy bBusyWaiting parameter to the EventCpu

### LINKED ISSUE #750: Unify constructors of EventCpu, EventCudaRt and EventHipRt
At the moment, the [EventCpu constructor][1] only takes the device as an argument, while the [EventCudaRt constructor][2] and the [EventHipRt constructor][3] both allow for an optional second parameter `bBusyWaiting` that allows to pass the flag `cudaEventBlockingSync` to the underlying native event.
As this flag might be needed when porting cuda applications to alpaka (e.g. using cupla), it is desirable to have the flag available for cupla (see [this pull request][4]), but currently implementing this requires an ifdef for the accelerator. This should be handled by alpaka instead.

[1]: https://github.com/ComputationalRadiationPhysics/alpaka/blob/3b0e68d292c49da34276803c3bddf490fdcfd2b8/include/alpaka/event/EventCpu.hpp#L112
[2]: https://github.com/ComputationalRadiationPhysics/alpaka/blob/3b0e68d292c49da34276803c3bddf490fdcfd2b8/include/alpaka/event/EventCudaRt.hpp#L118
[3]: https://github.com/ComputationalRadiationPhysics/alpaka/blob/3b0e68d292c49da34276803c3bddf490fdcfd2b8/include/alpaka/event/EventHipRt.hpp#L125
[4]: https://github.com/ComputationalRadiationPhysics/cupla/pull/98

DerWaldschrat:


BenjaminW3:
You have to use `alpaka::ignore_unused(bBusyWait);` within the constructor body to make the compilers happy.

BenjaminW3:
Could you please squash all the commits into a single one?

BenjaminW3 (approved):

PR #837, opened 2019-09-11
alpaka/math Overloaded float specialization

jkrude:
Fixed all falsly copy-pasted "standard libary".
Regarding all, but the min and max functions:
Added a template specialization for floats, to explicitly use the float backends.

- [ ]  depends on #828 (math will be tested since then)

Edit: added dependency

### DIFF
--- a/include/alpaka/math/abs/AbsCudaBuiltIn.hpp
+++ b/include/alpaka/math/abs/AbsCudaBuiltIn.hpp
@@ -58,6 +58,21 @@ namespace alpaka
                     return ::abs(arg);
                 }
             };
+
+            template<>
+            struct Abs<
+                AbsHipBuiltIn,
+                float>
+            {
+                __device__ static auto abs(
+                    AbsHipBuiltIn const & abs_ctx,
+                    float const & arg)
+                -> float
+                {
+                    alpaka::ignore_unused(abs_ctx);
+                    return ::fabsf(arg);
+                }
+            };
         }
     }
 }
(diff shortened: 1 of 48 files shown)


tdd11235813:
only HIP complains:
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/math/pow/PowHipBuiltIn.hpp(76): error: identifier "PowCudaBuiltIn" is undefined

/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/math/pow/PowHipBuiltIn.hpp(81): error: identifier "PowCudaBuiltIn" is undefined
```
probably just a typo: Cuda -> Hip

BenjaminW3:
In how far does this depend on #828?
I know that it would be good to add this together with the corresponding tests. However, the tests are not yet ready and there are no code changes that depend on the other PR so it would be ok for me to merge this before the tests.

psychocoderHPC:
For me it looks like more this PR is independent of #828 and #828 is testing the new functionality you provided here.

jkrude:
Totally correct. This PR is not directly depending on #828, but if #828 is finished the new functionality could fully be tested. So not only the compiler-errors are checked but the functionality too. That is why I thought it would make sense waiting for #828.

tdd11235813:
the math tester in #828 actually enables testing of the alpaka math functions already at compile-time. Before, not all math functions have been used by the tests and thus some bugs have been undetected. By the math tester @jkrude could find the bugs. Thus this relies on #828, so we can detect possible bugs introduced by this PR :)

BenjaminW3 (approved):

BenjaminW3:
Lets finalize #828 so that this is properly tested!

PR #845, opened 2019-09-20
fix kernel parameter constness

### LINKED ISSUE #843: On 0.4.0 arrays can't be passed to kernels
Something like this:

```
double data[4] = { };
alpaka::queue::enqueue(stream, alpaka::kernel::createTaskKernel<Accelerator>(workDiv, kernel, data));

```

doesn't work in 0.4.0 anymore, since data is converted to a const pointer (actually std::ref creates a std::reference_wrapper<const double[4]>).

psychocoderHPC:
fix #843

Fix that kernel parameter require constness.

# Require

- [x] runtime test
- [x] creating a unit test

### DIFF
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -61,10 +61,10 @@ namespace alpaka
             ALPAKA_FN_HOST TaskKernelCpuOmp2Blocks(
                 TWorkDiv && workDiv,
                 TKernelFnObj const & kernelFnObj,
-                TArgs const & ... args) :
+                TArgs && ... args) :
                     workdiv::WorkDivMembers<TDim, TIdx>(std::forward<TWorkDiv>(workDiv)),
                     m_kernelFnObj(kernelFnObj),
-                    m_args(args...)
+                    m_args(std::forward<TArgs>(args)...)
             {
 
                 static_assert(
@@ -99,7 +99,7 @@ namespace alpaka
                 // Get the size of the block shared dynamic memory.
                 auto const blockSharedMemDynSizeBytes(
                     meta::apply(
-                        [&](TArgs const & ... args)
+                        [&](typename std::decay<TArgs>::type const & ... args)
                         {
                             return
                                 kernel::getBlockSharedMemDynSizeBytes<
@@ -119,7 +119,7 @@ namespace alpaka
                 // TODO: With C++14 we could create a perfectly argument forwarding function object within the constructor.
                 auto const boundKernelFnObj(
                     meta::apply(
-                        [this](TArgs const & ... args)
+                        [this](typename std::decay<TArgs>::type const & ... args)
                         {
                             return
                                 std::bind(
@@ -218,7 +218,7 @@ namespace alpaka
             }
 
             TKernelFnObj m_kernelFnObj;
-            std::tuple<TArgs...> m_args;
+            std::tuple<typename std::decay<TArgs>::type...> m_args;
         };
     }
 
--- a/include/alpaka/kernel/TaskKernelGpuCudaRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuCudaRt.hpp
@@ -159,10 +159,10 @@ namespace alpaka
             ALPAKA_FN_HOST TaskKernelGpuCudaRt(
                 TWorkDiv && workDiv,
                 TKernelFnObj const & kernelFnObj,
-                TArgs const & ... args) :
+                TArgs && ... args) :
                     workdiv::WorkDivMembers<TDim, TIdx>(std::forward<TWorkDiv>(workDiv)),
                     m_kernelFnObj(kernelFnObj),
-                    m_args(args...)
+                    m_args(std::forward<TArgs>(args)...)
             {
                 static_assert(
                     dim::Dim<typename std::decay<TWorkDiv>::type>::value == TDim::value,
@@ -180,7 +180,7 @@ namespace alpaka
             ~TaskKernelGpuCudaRt() = default;
 
             TKernelFnObj m_kernelFnObj;
-            std::tuple<TArgs...> m_args;
+            std::tuple<typename std::decay<TArgs>::type...> m_args;
         };
     }
 
@@ -335,7 +335,7 @@ namespace alpaka
                     // Get the size of the block shared dynamic memory.
                     auto const blockSharedMemDynSizeBytes(
                         meta::apply(
-                            [&](TArgs const & ... args)
+                            [&](typename std::decay<TArgs>::type const & ... args)
                             {
                                 return
                                     kernel::getBlockSharedMemDynSizeBytes<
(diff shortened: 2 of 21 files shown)


psychocoderHPC:
~~grrrrrr this fix is not working for other examples. @sbastrakov and me currently working on a fix. Any fix before we solved the issue is welcome.~~

psychocoderHPC:
We hopefully fixed it now.

example/helloWorld/src/helloWorld.cpp
```diff
@@ -27,9 +27,10 @@ struct HelloWorldKernel
 {
     //-----------------------------------------------------------------------------
     template<
-        typename TAcc>
+        typename TAcc,
+        typename T>
     ALPAKA_FN_ACC auto operator()(
-        TAcc const & acc) const
```
psychocoderHPC: these example changes must be removed

test/common/include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -53,7 +53,7 @@ namespace alpaka
                 typename... TArgs>
             auto operator()(
                 TKernelFnObj const & kernelFnObj,
-                TArgs const & ... args)
+                TArgs && ... args)
```
BenjaminW3, line 58: You probably want to use `std::forward` at the place where `args` is used within this method.

include/alpaka/kernel/Traits.hpp
```diff
@@ -121,7 +121,7 @@ namespace alpaka
             TKernelFnObj const & kernelFnObj,
             vec::Vec<TDim, idx::Idx<TAcc>> const & blockThreadExtent,
             vec::Vec<TDim, idx::Idx<TAcc>> const & threadElemExtent,
-            TArgs const & ... args)
+            TArgs && ... args)
```
BenjaminW3: Usage without `std::forward` is somehow useless.

BenjaminW3:
Thanks for working on this fix!
What I see is:
* The `CreateTaskKernel::createTaskKernel` speciaizations now use perfect frowarding instead of capturing the arguments  by `const &`.
* The `TaskKernel*` constructors also now use perfect frowarding instead of capturing the arguments  by `const &`.
* The `TaskKernel*`s still store copies of the arguments (`std::decay` was necessary to remove the possible references)

All of those 3 points together enables passing "arbitrary" objects into the tasks which was not posible with `const &`.

I am not sure about the mix of `typename std::decay<TArgs>::type const &` and `TArgs &&` in the `TaskKernel*` implementations.

include/alpaka/kernel/Traits.hpp
psychocoderHPC: This were the places I overlooked on Friday short before I leave the office, thx

psychocoderHPC:
> I am not sure about the mix of `typename std::decay<TArgs>::type const &` and `TArgs &&` in the `TaskKernel*` implementations.

This should be ok, but @sbastrakov and I will have again a look at it. If each thread in the worker is again coping depends on the signature of the user kernel.

include/alpaka/kernel/Traits.hpp
```diff
@@ -245,7 +245,7 @@ namespace alpaka
             TQueue & queue,
             TWorkDiv const & workDiv,
             TKernelFnObj const & kernelFnObj,
-            TArgs const & ... args)
+            TArgs && ... args)
```
BenjaminW3, line 248: `std::forward` is missing for the usage

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -99,7 +99,7 @@ namespace alpaka
                 // Get the size of the block shared dynamic memory.
                 auto const blockSharedMemDynSizeBytes(
                     meta::apply(
-                        [&](TArgs const & ... args)
+                        [&](typename std::decay<TArgs>::type const & ... args)
```
BenjaminW3, line 102: May we be able to have a `using TArgsDecayed = typename std::decay<TArgs>::type`?

psychocoderHPC: `std::decay_t<>` is available with C++14. I am currently thinking of adding it in alpaka and switch with c++14 to the stl version.

psychocoderHPC: should I use `#if __cplusplus >= 201402L` to detect or is there a boost define available (I do not find any)

What I found on stackoverflow https://stackoverflow.com/questions/26089319/is-there-a-standard-definition-for-cplusplus-in-c14

BenjaminW3: Having it as a typedef would still be less code in mot cases. Even with `decay_t` we would have to write it multiple time.
I am not sure if it is worth all the work when we switch to C++14 for the next alpaka release.

include/alpaka/kernel/Traits.hpp
```diff
@@ -217,7 +217,7 @@ namespace alpaka
                     TArgs...>::createTaskKernel(
                         workDiv,
                         kernelFnObj,
-                        args...);
+                        std::forward<TArgs>(args)...);
```
BenjaminW3, line 220: The method return type is still mising the `std::forward`. (C++14 will allow us to remove it completely)

psychocoderHPC:
looks like I get it managed. I am currently testing HIP with hcc backend. On hemera the tests fully compile now.

include/alpaka/kernel/TaskKernelGpuCudaRt.hpp
```diff
@@ -497,9 +497,9 @@ namespace alpaka
                     // This forces the type of a float argument given with std::forward to this function to be of type float instead of e.g. "float const & __ptr64" (MSVC).
                     // If not given by value, the kernel launch code does not copy the value but the pointer to the value location.
                     meta::apply(
-                        [&](TArgs ... args)
+                        [&](typename std::decay<TArgs>::type const & ... args)
```
BenjaminW3, line 497: This conflicts with the comment now. For HIP you have removed the comment. Do you want to remove it here a well?

psychocoderHPC: Yes must this comment should be removed too

psychocoderHPC:
offline discussed: @sbastrakov will help to provide the test cases so that we can get this PR faster in

alpakaConfig.cmake
```diff
@@ -264,7 +264,7 @@ IF(ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE)
     ENDIF()
 
 ELSE()
-    FIND_PACKAGE(Boost ${_ALPAKA_BOOST_MIN_VER} QUIET)
```
sbastrakov: How is this change related to the PR topic?

psychocoderHPC: well spotted, it is not related and must be re-added

sbastrakov:
I made some unit tests [here](https://github.com/psychocoderHPC/alpaka/pull/1).

BenjaminW3 (approved):

BenjaminW3:
@psychocoderHPC @sbastrakov Can this be merged now?

psychocoderHPC:
@BenjaminW3 Yes it can now be merged. I added the last missing mark for the CI tests. 
@sbastrakov thx for the CI tests

PR #846, opened 2019-09-22
Use cudaDeviceGetAttribute instead of slow cudaGetDeviceProperties

BenjaminW3:
See https://devblogs.nvidia.com/cuda-pro-tip-the-fast-way-to-query-device-properties/

### DIFF
--- a/include/alpaka/dev/DevCudaRt.hpp
+++ b/include/alpaka/dev/DevCudaRt.hpp
@@ -93,6 +93,7 @@ namespace alpaka
                     dev::DevCudaRt const & dev)
                 -> std::string
                 {
+                    // There is cudaDeviceGetAttribute as faster alternative to cudaGetDeviceProperties to get a single device property but it has no option to get the name
                     cudaDeviceProp cudaDevProp;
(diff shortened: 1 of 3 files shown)


psychocoderHPC (approved):

include/alpaka/dev/DevCudaRt.hpp
```diff
@@ -93,6 +93,7 @@ namespace alpaka
                     dev::DevCudaRt const & dev)
                 -> std::string
                 {
+                    // There is cudaDeviceGetAttribute as faster alternative to cudaGetDeviceProperties to get a single device property but it has no option to get the name
```
ax3l, line 96: Maybe just use `cuDeviceGetName`?

psychocoderHPC: `cuDeviceGetName` is driver API. Currently, we are using only in one of the alpaka tests the driver API.
Since I do not think that someone will call this function within time-critical code we should not use `cuDeviceGetName` even if it looks like an alternative.

Note: if you use the driver API you must also take care because the device order is not required equal to the runtime api.

PR #866, opened 2019-10-24
[WIP] add `pthread` as alpaka dependency

psychocoderHPC:
`std::future` needs `-pthread` at compile time to work correctly

see cupla issue: https://github.com/ComputationalRadiationPhysics/cupla/pull/128#issuecomment-545078917

We already passed `pthread` to the linker but we need also activate `pthread` usage for the compiler.
The implementation expresses the dependancy pthread by using `FIND_PACKAGE(Threads REQUIRED)`

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -220,8 +220,6 @@ IF(MSVC)
     ENDIF()
 ELSE()
     # Add linker options.
-    # lipthread:
-    LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "general;pthread")
     IF(NOT APPLE)
         # librt: undefined reference to `clock_gettime'
         LIST(APPEND _ALPAKA_LINK_LIBRARIES_PUBLIC "general;rt")
@@ -236,6 +234,17 @@ ELSE()
         CMAKE_CXX_COMPILER_VERSION VERSION_LESS 4.0))
         LIST(APPEND _ALPAKA_COMPILE_OPTIONS_PUBLIC "-ftemplate-depth=1024")
     ENDIF()
+ENDIF()
+
+#-------------------------------------------------------------------------------
+# Find Threads.
+IF(NOT MSVC)
+    # for std::future we need to pass the pthreads flags for the compiler and linker
+    # https://github.com/ComputationalRadiationPhysics/cupla/pull/128#issuecomment-545078917
+    # The dependency pthread will be searched here and after defining the target alpaka
+    # we apply the dependency
+    SET(THREADS_PREFER_PTHREAD_FLAG TRUE)
+    FIND_PACKAGE(Threads REQUIRED)
 ENDIF()
 
 #-------------------------------------------------------------------------------
@@ -1087,6 +1096,11 @@ ENDIF()
 IF((ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE) AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
     STRING(REPLACE ";" " " _ALPAKA_COMPILE_OPTIONS_STRING "${_ALPAKA_COMPILE_OPTIONS_PUBLIC}")
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_ALPAKA_COMPILE_OPTIONS_STRING}")
+ENDIF()
+
+IF(NOT MSVC)
+    # dependency for std::future
+    TARGET_LINK_LIBRARIES("alpaka" INTERFACE Threads::Threads)
 ENDIF()
 
 #-------------------------------------------------------------------------------


alpakaConfig.cmake
```diff
@@ -238,6 +236,17 @@ ELSE()
     ENDIF()
 ENDIF()
 
+#-------------------------------------------------------------------------------
+# Find Threads.
+IF(NOT MSVC)
+    # for std::future we need to pass the pthread flags for the compiler and linker
```
sbastrakov: `pthread` should be `pthreads` I guess.

alpakaConfig.cmake
```diff
@@ -1089,6 +1098,11 @@ IF((ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE) AND ALPAKA_CUDA_COM
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_ALPAKA_COMPILE_OPTIONS_STRING}")
 ENDIF()
 
+IF(NOT MSVC)
+    # dependency for std::future
+    TARGET_LINK_LIBRARIES("alpaka" INTERFACE  Threads::Threads)
```
sbastrakov: Started discussing offline: I am not sure this should be `INTERFACE`, think `PRIVATE` instead since we only use `std::future` internally and this is not part of alpaka API.

psychocoderHPC: CMake is saying:
```
  INTERFACE library can only be used with the INTERFACE keyword of
  target_link_libraries

```
Since the target alpaka is an `INTERFACE`  so we need to use `INTERFACE` here.

psychocoderHPC: but there is one space to much after `INTERFACE`, this I will fix after but like to wait for other reviews and suggestions first.

alpakaConfig.cmake
```diff
@@ -238,6 +236,17 @@ ELSE()
     ENDIF()
 ENDIF()
 
+#-------------------------------------------------------------------------------
+# Find Threads.
+IF(NOT MSVC)
```
BenjaminW3, line 241: Are we sure that this is required on Mac as well?

psychocoderHPC: No I am not sure. I expect it since MAC is also using Pthreads for the system threads.
IMO it is not a big deal if we require on MAC pthreads too.

psychocoderHPC: before we added pthreads for MAC to the linker: https://github.com/ComputationalRadiationPhysics/alpaka/pull/866/files#diff-cd45f124e237a4bd8df18ffab83a5664L224
So this PR will only additionally add it to the compiler too.

alpakaConfig.cmake
sbastrakov: I see, thanks for explanations!

BenjaminW3:
HIP does not seem to like the `-pthread` option.

BenjaminW3:
I am still wondering, why this is not a problem for the alpaka CI and other projects using alpaka?

psychocoderHPC:
> HIP does not seem to like the `-pthread` option.

Yes `nvcc` also do not know this flag because it must be passed to the host compiler

psychocoderHPC:
> I am still wondering, why this is not a problem for the alpaka CI and other projects using alpaka?

I am not sure why but clearly need it on different systems. I think it is a side effect because alpaka is enabling all backends and this flag is automatically added with `OpenMP`

alpakaConfig.cmake
```diff
@@ -238,6 +236,17 @@ ELSE()
...
+    # for std::future we need to pass the pthreads flags for the compiler and linker
+    # https://github.com/ComputationalRadiationPhysics/cupla/pull/128#issuecomment-545078917
+    # The dependency pthread will be searched here and after defining the target alpaka
+    # we apply the dependency
+    SET(THREADS_PREFER_PTHREAD_FLAG TRUE)
```
ax3l, line 246: maybe even:
```cmake
if(NOT DEFINED THREADS_PREFER_PTHREAD_FLAG)
    SET(THREADS_PREFER_PTHREAD_FLAG TRUE)
endif()
```
so users can overwrite it from the command-line.

BenjaminW3:
Maybe source of related information: https://github.com/google/googletest/pull/2483

ax3l:
That diff looks more like a bunch of pre-CMake 3.1 work-arounds to me, otherwise identical :)

ienorand:
As far as I've understood whilst digging into this for googletest, nvcc needs `-Xcompiler -pthread` instead, see https://gitlab.kitware.com/cmake/cmake/issues/18008 .

(quoting myself) https://github.com/google/googletest/pull/2483#issuecomment-538761499
> So reading up on CUDA a bit it seems like -pthreads should be passed using -Xcompiler for nvcc as a special case, which seems to have been implemented only relatively recently in CMake according to https://gitlab.kitware.com/cmake/cmake/merge_requests/2512. Which seems to have ended up in Cmake 3.13.0.

Since you are using Cmake 3.11.4, that might be why you are seeing the nvcc issues still?

ax3l:
Thank you for the feedback!
Alright, so if we bump our build dependencies to [CMake 3.13.0+](https://gitlab.kitware.com/cmake/cmake/commit/d81c4db489d773e4a1ce6be49eb9c41582e68ae3) ([CMake 3.14.0+ branch](https://gitlab.kitware.com/cmake/cmake/commit/69275d3b6a9787268f7bd80bb24d8e4cbb18710e)) or the potentially the backported [CMake 3.12.4](https://gitlab.kitware.com/cmake/cmake/commit/d17755f9cf583387fab83ad5744b93c746aa5994) we should be able to solve this as well.

psychocoderHPC:
> Thank you for the feedback!
> Alright, so if we bump our build dependencies to [CMake 3.13.0+](https://gitlab.kitware.com/cmake/cmake/commit/d81c4db489d773e4a1ce6be49eb9c41582e68ae3) ([CMake 3.14.0+ branch](https://gitlab.kitware.com/cmake/cmake/commit/69275d3b6a9787268f7bd80bb24d8e4cbb18710e)) or the potentially the backported [CMake 3.12.4](https://gitlab.kitware.com/cmake/cmake/commit/d17755f9cf583387fab83ad5744b93c746aa5994) we should be able to solve this as well.

I will check if CMake 3.12.4  solved the issue, since we using still the old why to use nvcc (not the CMake native compiler) I am not sure if your linked backport solves it.

ax3l:
Yes, I was wondering that as well - we are not using the CUDA language feature for targets yet (since we also support `clang -x cuda` as compiler).

psychocoderHPC:
From our CI it looks like only hip is affected by this problem. I try to solve this issue but have the problem that I have currently no access to a working nvidia system with hip support :-(

ax3l:
Doesn't Summit have official HIP modules to run on Nvidia?

j-stephan:
I'm not entirely sure why but this will also be fixed once #921 is merged. The CI from #921 reports the following line for HIP + nvcc + `Threads::Threads` (correctly choosing `-lpthread` instead of `-pthread`:

```
Feb 24 13:18:44 [  2%] Linking HIP executable bufferCopy

Feb 24 13:18:44 cd /home/travis/build/ComputationalRadiationPhysics/alpaka/build/example/bufferCopy && /home/travis/CMake/bin/cmake -E cmake_link_script CMakeFiles/bufferCopy.dir/link.txt --verbose=1

Feb 24 13:18:44 /home/travis/hip/bin/hipcc_cmake_linker_helper /opt/rocm/hcc    CMakeFiles/bufferCopy.dir/src/bufferCopy_generated_bufferCopy.cpp.o -o bufferCopy  -lpthread /usr/lib/x86_64-linux-gnu/librt.so /home/travis/hip/hiprand/lib/libhiprand-d.so -lcudart
```

psychocoderHPC:
> I'm not entirely sure why but this will also be fixed once #921 is merged. The CI from #921 reports the following line for HIP + nvcc + `Threads::Threads` (correctly choosing `-lpthread` instead of `-pthread`:
> 
> ```
> Feb 24 13:18:44 [  2%] Linking HIP executable bufferCopy
> 
> Feb 24 13:18:44 cd /home/travis/build/ComputationalRadiationPhysics/alpaka/build/example/bufferCopy && /home/travis/CMake/bin/cmake -E cmake_link_script CMakeFiles/bufferCopy.dir/link.txt --verbose=1
> 
> Feb 24 13:18:44 /home/travis/hip/bin/hipcc_cmake_linker_helper /opt/rocm/hcc    CMakeFiles/bufferCopy.dir/src/bufferCopy_generated_bufferCopy.cpp.o -o bufferCopy  -lpthread /usr/lib/x86_64-linux-gnu/librt.so /home/travis/hip/hiprand/lib/libhiprand-d.so -lcudart
> ```

offline discussed: `-lpthread` and `-pthread` are different things and `-pthread` should be a compiler flag. We will look into this issue after the cmake refactoring is merged.

BenjaminW3:
@psychocoderHPC Is this now resolved?

j-stephan:
Nope, this will need further work.

psychocoderHPC:
Yes this must still be solved. I will work with @j-stephan on that part but it will take some time.

alpakaConfig.cmake
```diff
@@ -1089,6 +1098,11 @@ IF((ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE) AND ALPAKA_CUDA_COM
     SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_ALPAKA_COMPILE_OPTIONS_STRING}")
 ENDIF()
 
+IF(NOT MSVC)
+    # dependency for std::future
+    TARGET_LINK_LIBRARIES("alpaka" INTERFACE Threads::Threads)
```
ax3l, line 1103: just generally: the target name `alpaka` does not need to be placed in quotes `""`.
Maybe remove everywhere in a follow-up.

BenjaminW3:
2 of the 3 new lines from this PR are now already part of [`alpakaCommon.cmake`](https://github.com/alpaka-group/alpaka/blob/develop/cmake/alpakaCommon.cmake#L141-L142)
* `FIND_PACKAGE(Threads REQUIRED)`
* `TARGET_LINK_LIBRARIES("alpaka" INTERFACE Threads::Threads)`

Only `SET(THREADS_PREFER_PTHREAD_FLAG TRUE)` is not there yet. Is this required?
Can this PR be closed as it is already done?

psychocoderHPC:
> Only `SET(THREADS_PREFER_PTHREAD_FLAG TRUE)` is not there yet. Is this required?
> Can this PR be closed as it is already done?

The is the most important line of this PR because this will set `-pthread` for compiling which is activating some defines which are required.
Please keep this PR open. 

Note: I am not available next week.

j-stephan:
What is the state here?

psychocoderHPC:
This PR is strongly required but I never found time to work on it.

alpakaConfig.cmake
bernhardmgruber, line 241: IIRC you can use `find_package` and `target_link_library` with `Threads` just fine with MSVC on Windows. No special handling is required. Please remove the `IF(NOT MSVC)` here and down below.

bernhardmgruber (changes requested):

j-stephan:
Closed by #1420.

PR #868, opened 2019-11-02
Replace XxxBase typedefs with inherited concept tags

BenjaminW3:
The XxxBase typedefs are replaced by corresponding `ConceptXxx` tags.
Implementors of a concept do not need to define the `XxxBase` typedef to
the current class anymore, but inherit from
`concepts::Implements<ConceptXxx, XxxImpl>`.

Furthermore, this removes the trait specializations for types with XxxBase typedefs.

This deletes more than 1600 lines of code!


include/alpaka/core/Concepts.hpp
```diff
@@ -0,0 +1,53 @@
...
+#pragma once
+
+#include <type_traits>
+
+namespace alpaka
+{
+    namespace concepts
+    {
+        //#############################################################################
+        //! Tag used in class inheritance hierarchies that describes that a specific conecpt (TConcept)
```
ax3l:
```suggestion
        //! Tag used in class inheritance hierarchies that describes that a specific concept (TConcept)
```

ax3l (commented):
Really cool, thank you!

include/alpaka/core/Concepts.hpp
```diff
@@ -0,0 +1,53 @@
...
+{
+    namespace concepts
+    {
+        //#############################################################################
+        //! Tag used in class inheritance hierarchies that describes that a specific concept (TConcept)
+        //! is implemented by the given base class (TBase).
+        template<
+            typename TConcept,
+            typename TBase>
+        struct ImplementsConcept
```
sbastrakov: Just as a thought. Perhaps `ImplementsConcept` is a little wordy, given it is always used like `concepts::ImplementsConcept`. Maybe worth considering to shorten to just `Implements`? Some other languages have `implements` as a keyword for interfaces which in some way are inbetween of C++ interfaces and concepts.

psychocoderHPC: I like the idea to shorten the name

BenjaminW3: Done ;-) (I had it this way first)

include/alpaka/core/Concepts.hpp
```diff
@@ -0,0 +1,53 @@
...
+namespace alpaka
+{
+    namespace concepts
+    {
+        //#############################################################################
+        //! Tag used in class inheritance hierarchies that describes that a specific concept (TConcept)
+        //! is implemented by the given base class (TBase).
+        template<
+            typename TConcept,
+            typename TBase>
```
sbastrakov, line 23: I am a little confused, why base class here? It looks like the type actually provided here is a child class that implements the concept and inherits from `Implements`. Edit: I understand that in `ImplementationBaseType::base()` the second argument is a base class, but I think it makes sense only there, but not generally.

include/alpaka/core/Concepts.hpp
```diff
@@ -0,0 +1,53 @@
...
+                static auto base(Implements<TConcept, TBase>*) -> TBase;
+
+                using type = decltype(base(std::declval<TDerived*>()));
+
+                static_assert(std::is_base_of<type, TDerived>::value, "The type implementing the concept has to be a publicly accessible base class!");
+            };
+        }
+
+        //#############################################################################
+        //! Given a derieved class (TDerived) it finds the base class (TBase) which implements the given concept (TConcept).
```
sbastrakov: Typo: `derieved ` should be `derived`.

include/alpaka/core/Concepts.hpp
BenjaminW3: Before this change we had inheritance hierarchies like `AccYyy->XxxImpl` where `XxxImpl` had a typedef that tagged it `using XxxBase = XxxImpl;` as implementing the Xxx concept.
This PR transformed it into `AccYyy->XxxImpl->Implements<ConceptXxx, XxxImpl>`.
So the `XxxBase` typedef has been transformed into the `TBase` template parameter. Therefore the `Base` name ;-)
The inheritance instead of the tag will allow us in the future to check at compile time that a specific class implements a concept only once which is not possible with typedefs.
Currently the implementation classes tag themselves as `XxxImpl->Implements<ConceptXxx, XxxImpl>`. This was the same for the typedef and in this case it does not really look like a `Base` class that is given as a parameter. Looking at it from a language standpoint is is not wrong either because `std::is_base_of<T, T>::value` is always true ;-)
However, I am thinking about moving the tagging onto the most derived class at least for some cases which would then look like this: `AccYyyl->Implements<ConceptXxx, XxxImpl>`. In this case the parameter really is a strict base-class. My use case would be the `AtomicHierarchy`. The atomic implementations can not know for which atomic hierarchy they are used (grid, block, thread) so they can not tag themselves as implementing a specific concept. The `AtomicHierarchy` on the other hand can tag itself and say that Impl1 implements the grid and block atomics and Impl2 implements the thread atomics.

include/alpaka/core/Concepts.hpp
BenjaminW3: fixed

psychocoderHPC (approved):

PR #869, opened 2019-11-07
fix hip-hcc compile

psychocoderHPC:
- fix wrong kernel call method for blocking queues
- fix compiler error (see #736): pure host function is interpreted for device 
  - this fix disable the method content for the device path

This PR replaces #736


psychocoderHPC:
The kernel start method was already fixed in  #800 for non blocking queues but somehow missed for blocking queues.

sbastrakov (dismissed):

sbastrakov (approved):

BenjaminW3 (approved):

PR #870, opened 2019-11-10
Simplify AtomicHierarchy

BenjaminW3:
This ads some metaprogramming constructs:
* `Unique`: removes duplicates from a type list
* `InheritFromList`: object that inherits from a list of types

Given those two generic constructs it is possible to replace the complicated inheritance logic as well as the AtomicBase trait.


BenjaminW3:
Ready for review/merge!

sbastrakov (approved):

PR #871, opened 2019-11-10
Do not specialze traits for shared_ptr<Xxx> but only for Xxx

BenjaminW3:
This removes some std::shared_ptr copies and makes the code easier to understand.


BenjaminW3:
Ready for review/merge!

sbastrakov (approved):

PR #872, opened 2019-11-10
Test concepts and extend with base case for types without inheritance

BenjaminW3:


include/alpaka/core/Concepts.hpp
```diff
@@ -27,27 +27,73 @@ namespace alpaka
...
+            //! Returns the type that implements the given concept in the inheritance hierarchy.
+            template<
+                typename TConcept,
+                typename TDerived,
+                typename Sfinae = void
+            >
+            struct ImplementationBaseType;
+
+            //#############################################################################
+            //! Base case for types that do not inherit from "Implements<TConcept, ...>" i the type itself.
```
psychocoderHPC: there is `i` in the doc string

include/alpaka/core/Concepts.hpp
```diff
@@ -27,27 +27,73 @@ namespace alpaka
...
         }
 
         //#############################################################################
-        //! Given a derived class (TDerived) it finds the base class (TBase) which implements the given concept (TConcept).
+        //! Returns the type that implements the given concept in the inheritance hierarchy.
         template<
             typename TConcept,
-            typename TDerived>
+            typename TDerived
+        >
```
psychocoderHPC: I think `>` must be on the line above (alpaka style). The same style issue exists a few times in the file.

psychocoderHPC (dismissed):

include/alpaka/core/Concepts.hpp
BenjaminW3: fixed

include/alpaka/core/Concepts.hpp
BenjaminW3: fixed

BenjaminW3:
Ready for review/merge!

include/alpaka/core/Concepts.hpp
```diff
@@ -27,24 +27,62 @@ namespace alpaka
...
+            template<
+                typename TConcept,
+                typename TDerived>
+            struct ImplementsConcept {
+                template<
+                    typename TBase>
+                static auto implements(Implements<TConcept, TBase>&) -> std::true_type;
+                static auto implements(...) -> std::false_type;
+
+                using type = decltype(implements(std::declval<TDerived&>()));
```
sbastrakov: Shouldn't this part be `private`? Also, not sure if `type` as class name is consistent with the style (it may be, I'm just not sure)?

BenjaminW3: Yes, I will make this private.

BenjaminW3: I made all things that should not be accessible private. the `type` typedef ha been remove.

sbastrakov (dismissed):

include/alpaka/core/Concepts.hpp
BenjaminW3: After I made them private all compilers build correctly except gcc who complains:
```
Concepts.hpp:33:20: error: all member functions in class 'alpaka::concepts::detail::ImplementsConcept<TConcept, TDerived>' are private [-Werror=ctor-dtor-privacy]
```
The compiler is somehow right but the diagnostic is useless. I will remove the access-specifier because those classes are traits and within the `detail` namespace so there are two reasons not to instantiate them at all.

BenjaminW3: Can you please approve this PR again?

sbastrakov (approved):

PR #874, opened 2019-11-14
HIP-clang support

psychocoderHPC:
Add support for the compiler HIP-clang.

- extent CMake `ALPAKA_HIP_PLATFORM` with the option `clang`
- add define `BOOST_COMP_HIP` if hip-clang is used

I know that we currently not cover hip-clang in the CI. The reason is that hip-clang is currently not final released and the installation is not as easy. Never the less we need these changes to work more closely together with AMD and Cray. At all HIP-clang is nearly equal to HIP-hcc but is more relaxed with the function attributes `__host__` and `__device__` so that #802 is not needed.
If someone breaks HIP-clang support somehow with a follow up PR it is not dramatic I will keep an eye on it and fix it if needed.

### DIFF
--- a/include/alpaka/core/Utility.hpp
+++ b/include/alpaka/core/Utility.hpp
@@ -27,7 +27,7 @@ namespace alpaka
         // within an alpaka accelerator kernel too.
         // This function can be used only within std::decltype().
         //-----------------------------------------------------------------------------
-#if BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA
+#if BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA || BOOST_COMP_HIP
         template< class T >
         ALPAKA_FN_HOST_ACC
         typename std::add_rvalue_reference<T>::type
--- a/include/alpaka/kernel/TaskKernelGpuHipRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuHipRt.hpp
@@ -41,6 +41,7 @@
 
 #include <alpaka/core/BoostPredef.hpp>
 #include <alpaka/core/Hip.hpp>
+#include <alpaka/core/Utility.hpp>
 #include <alpaka/meta/ApplyTuple.hpp>
 #include <alpaka/meta/Metafunctions.hpp>
 
@@ -76,11 +77,15 @@ namespace alpaka
 #if BOOST_ARCH_PTX && (BOOST_ARCH_PTX < BOOST_VERSION_NUMBER(2, 0, 0))
     #error "Cuda device capability >= 2.0 is required!"
 #endif
+
 #pragma clang diagnostic push
 #pragma clang diagnostic ignored "-Wignored-attributes"
                     static_assert(
-                        std::is_same<typename std::result_of<
-                            TKernelFnObj(acc::AccGpuHipRt<TDim, TIdx> const &, TArgs const & ...)>::type, void>::value,
+                        std::is_same<
+                            decltype(kernelFnObj(
+                                alpaka::core::declval<acc::AccGpuHipRt<TDim, TIdx> const>(),
+                                args...)),
+                        void>::value,
                         "The TKernelFnObj is required to return void!");
 #pragma clang diagnostic pop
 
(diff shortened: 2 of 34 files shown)


include/alpaka/kernel/TaskKernelGpuHipRt.hpp
```diff
@@ -76,14 +76,16 @@ namespace alpaka
...
+#   pragma clang diagnostic push
+#   pragma clang diagnostic ignored "-Wignored-attributes"
                     static_assert(
                         std::is_same<typename std::result_of<
                             TKernelFnObj(acc::AccGpuHipRt<TDim, TIdx> const &, TArgs const & ...)>::type, void>::value,
                         "The TKernelFnObj is required to return void!");
-#pragma clang diagnostic pop
-
+#   pragma clang diagnostic pop
+#endif
```
BenjaminW3: Is the `pragma` unsupported or the `static_assert`? I would like to keep the `static_assert` or at least add a comment why this does not work.

include/alpaka/core/Utility.hpp
```diff
@@ -27,7 +27,7 @@ namespace alpaka
         // within an alpaka accelerator kernel too.
         // This function can be used only within std::decltype().
         //-----------------------------------------------------------------------------
-#if BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA
+#if BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA || BOOST_COMP_HIP
```
sbastrakov, line 30: If the (lack of) brackets what was intended here?

include/alpaka/queue/QueueHipRtBlocking.hpp
```diff
@@ -106,7 +106,7 @@ namespace alpaka
                     dev::DevHipRt const m_dev;   //!< The device this queue is bound to.
                     hipStream_t m_HipQueue;
 
-#if BOOST_COMP_HCC  // NOTE: workaround for unwanted nonblocking hip streams for HCC (NVCC streams are blocking)
+#if BOOST_COMP_HCC || BOOST_COMP_HIP  // NOTE: workaround for unwanted nonblocking hip streams for HCC (NVCC streams are blocking)
```
sbastrakov: In the comment should probably also be `HCC or HIP`?

include/alpaka/core/Utility.hpp
psychocoderHPC: yes they are not needed

include/alpaka/queue/QueueHipRtBlocking.hpp
sbastrakov: Same for other occurrences of this comment (next files).

test/unit/core/src/BoostPredefTest.cpp
```diff
@@ -31,9 +31,12 @@ TEST_CASE("printDefines", "[core]")
 #if BOOST_COMP_NVCC
     std::cout << "BOOST_COMP_NVCC:" << BOOST_COMP_NVCC << std::endl;
 #endif
-#if BOOST_COMP_HCC
+#if BOOST_COMP_HCC || BOOST_COMP_HIP
```
sbastrakov: Is it intended that for HIP now both lines with `BOOST_COMP_HCC:` and `BOOST_COMP_HIP:` will be printed?

include/alpaka/kernel/TaskKernelGpuHipRt.hpp
psychocoderHPC: The evaluation of the static assert expression failed but I now got it solved.

```
                    static_assert(
                        std::is_same<
                            decltype(kernelFnObj(
                                alpaka::core::declval<acc::AccGpuHipRt<TDim, TIdx> const>(),
                                args...)),
                        void>::value,
                        "The TKernelFnObj is required to return void!");
```

include/alpaka/queue/QueueHipRtBlocking.hpp
psychocoderHPC: I will remove this for HIP-clang and will add a `static_assert` that we currently not support callbacks with `HIP-clang`.
This line is only a workaround because the hip callbacks are not synchronous. The current workaround adds polling but this is destroying all performance e.g. for PIConGPU.

psychocoderHPC:
ready

test/unit/core/src/BoostPredefTest.cpp
psychocoderHPC: thx must be removed

sbastrakov (approved):
Looks good to me, thanks!

BenjaminW3 (approved):

PR #877, opened 2019-11-16
Make memcpy and memset operations for QueueCudaRtBlocking blocking async

### LINKED ISSUE #851: cudaMemset is not blocking
Related to issue #850 the [cuda documentation for cudaMemset](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior__memcpy-sync) is saying 

```
Memset
The synchronous memset functions are asynchronous with respect 
to the host except when the target is pinned host memory or a 
Unified Memory region, ...
```

This means calling `cudaMemset` to NON pinned or managed memory and directly after call a kernel which works with the date within another stream (unrelated if we created the stream with `cudaStreamNonBlocking`) we will have the possibility of a data race.

[update] I added NON into my sentence because as @sbastrakov pointed out it is safe to use cudaMemSet for pinned memory but not for NON pinned.

BenjaminW3:
Hopefully fixes: #851, #850 and #504

### DIFF
--- a/include/alpaka/mem/buf/cuda/Copy.hpp
+++ b/include/alpaka/mem/buf/cuda/Copy.hpp
@@ -25,8 +25,6 @@
 #include <alpaka/dim/DimIntegralConst.hpp>
 #include <alpaka/extent/Traits.hpp>
 #include <alpaka/mem/view/Traits.hpp>
-#include <alpaka/queue/QueueCudaRtNonBlocking.hpp>
-#include <alpaka/queue/QueueCudaRtBlocking.hpp>
 
 #include <alpaka/core/Assert.hpp>
 #include <alpaka/core/Cuda.hpp>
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
@BenjaminW3 thx for taken over and add the fixes. We need to do the same for hip(hcc/clang/nvidia) but we should do it in a separate PR.
I will test this PR against #850.

psychocoderHPC:
We need to apply these changes to HIP too because HIP-nvcc will definitively show the same behavior.
I need to check if HIP-clang is also effected.

psychocoderHPC:
I was on the way to add a test case, but this would require a few refactorings within our test helpers. We need to be able to select a `Blocking` queue for an accelerator. The current workflow within tests is to get the default queue this is a non blocking queue if debug is disabled else the blocking queue.
A test based on the example from #850 would not guarantee that we trigger the data race, therefore, it is  maybe also not the best test case.

psychocoderHPC (approved):

psychocoderHPC:
My manual tests showed that the bug is fixed.

BenjaminW3:
I will try to apply this to HIP as soon as I find time.

PR #879, opened 2019-11-18
fix warning: implicit deleted constructor

psychocoderHPC:
fix warning for`RandHipRand` and `RandCuRand`

- the default constructor can be called from the device or host side.

```
In file included from alpaka/test/unit/kernel/src//KernelLambda.cpp:15:
In file included from alpaka/test/common/include/alpaka/test/acc/TestAccs.hpp:12:
In file included from alpaka/include/alpaka/alpaka.hpp:29:
In file included from alpaka/include/alpaka/acc/AccGpuHipRt.hpp:30:
alpaka/include/alpaka/rand/RandHipRand.hpp:61:21: warning: explicitly defaulted default constructor is implicitly deleted [-Wdefaulted-function-deleted]
                    Xor() = default;
                    ^
alpaka/include/alpaka/rand/RandHipRand.hpp:78:42: note: default constructor of 'Xor' is implicitly deleted because field 'm_State' has no default constructor
                    hiprandStateXORWOW_t m_State;
```

### DIFF
--- a/include/alpaka/rand/RandCuRand.hpp
+++ b/include/alpaka/rand/RandCuRand.hpp
@@ -48,9 +48,12 @@ namespace alpaka
                 public:
 
                     //-----------------------------------------------------------------------------
-                    //! After calling this constructor the instance is not valid initialized and
-                    //! need to be overwritten with a valid object
-                    Xor() = default;
+                    // After calling this constructor the instance is not valid initialized and
+                    // need to be overwritten with a valid object
+                    //-----------------------------------------------------------------------------
+                    ALPAKA_FN_HOST_ACC Xor() : m_State(curandStateXORWOW_t{})
+                    {
+                    }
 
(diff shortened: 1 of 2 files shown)


BenjaminW3:
Would it be useful to do the same for the CUDA version? It is most probably also not initialize correctly when the default contructor is used. This is different for the CPU versions which have a reaonable default constructor.

psychocoderHPC:
yes maybe it is better to remove it everywhere. Let me check it.

psychocoderHPC:
Removing the standard constructor is not a good idea from my side. We added the constructor in the past #255 to have the same behavior from then std random number generators.

I will think about a fix which is not breaking the example linked in #255

psychocoderHPC:
I updated the PR. The default constructor will initialize the object as before.

BenjaminW3 (approved):

PR #881, opened 2019-11-20
Fix QueueHipRtBlocking not being blocking

BenjaminW3:
Follow-up to #877

I had to update ROC from 2.7.0 to 2.8.0 because this is the first version
which supports hipMemset3DAsync.


psychocoderHPC:
> I had to update ROC from 2.7.0 to 2.8.0 because this is the first version

np I was only on the way to do it. Since the AMD ecosystem is currently changing fast I see no problem to enforce newer versions and skip older versions.

psychocoderHPC (approved):

include/alpaka/mem/buf/hip/Copy.hpp
```diff
@@ -993,7 +1000,11 @@ namespace alpaka
                     // Initiate the memory copy.
                     ALPAKA_HIP_RT_CHECK(
                         hipMemcpy3D(
```
psychocoderHPC: must be `hipMemcpy3DAsync`

psychocoderHPC (dismissed):

include/alpaka/mem/buf/hip/Copy.hpp
BenjaminW3: fixed

BenjaminW3:
Ready to be approved/merged!

psychocoderHPC (approved):

Issue #838, opened 2019-09-11
Compiler errors and warnings 


krzikalla:
CUDA V10.1.105
gcc 7.3.0
alpaka-0.3.5

Compiled with --std=c++14. 
I got lot of errors despite your table states, that the gcc/nvcc combination is running.
In alpaka-0.3.5/include/alpaka/core/ConcurrentExecPool.hpp there is the line

`auto currentTaskPackage(std::shared_ptr<ITaskPkg>{nullptr});`

at various places. Apparently at least one compiler deduces currentTaskPackage as a constexpr.
I could fix it by replacing these lines with:

`std::shared_ptr<ITaskPkg> currentTaskPackage{nullptr};`

Also I got the following warning quite often:

`warning: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu") is not allowed
`

~And an~

~`internal compiler error: in maybe_undo_parenthesized_ref, at cp/semantics.c:1705`~

~Edit. the last one is a duplicate of #765.~

sbastrakov:
@krzikalla thanks for your report!
I am slightly confused by the configuration since alpaka 0.3.5 did not claim support for CUDA 10, and gcc 7.3 was not at the table for that release. This is not to say these defects are definitely not present at the current `develop`.

Edit: now I see at #765 that this configuration was actually working before. Thanks for referencing that issue here!

sbastrakov:
@krzikalla so far I did not manage to reproduce the issues when building the alpaka examples. Can you please specify which alpaka backends were enabled in your failing build?

krzikalla:
ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED 
ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED 
ALPAKA_ACC_GPU_CUDA_ENABLED 

are defined.

krzikalla:
The offending lines are still there, however slightly changed:

auto currentTaskPackage = std::shared_ptr<ITaskPkg>{nullptr};

currentTaskPackage was deduced as a constexpr before that change.

krzikalla:
The following code is a minimal example;

```
#include <memory>
#include <iostream>

int main(int argc, char *argv[])
{
  auto x (std::shared_ptr<int>{nullptr});
  //auto x = std::shared_ptr<int>{nullptr};
[...]

PR #884, opened 2019-11-21
fix warning for `std::__shared_counter`

psychocoderHPC:
supress warning for CUDA 10.1 - 10.2:

```
alpaka/vec/Vec.hpp(1040): warning ComputationalRadiationPhysics#2912-D: calling a __host__ function("std::__shared_count<( ::__gnu_cxx::_Lock_policy)2> ::__shared_count") from a __host__ __device__ function("alpaka::dev::DevCpu::DevCpu") is not allowed
```

Note that the suppression number is CUDA version-specific and must be fixed for each version independently. 

issue and solution mentioned in #838

### DIFF
--- a/alpakaConfig.cmake
+++ b/alpakaConfig.cmake
@@ -675,6 +675,21 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                     LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored)
                 ENDIF()
 
+                # avoids warnings on host-device signature of 'std::__shared_count<>'
+                IF(CUDA_VERSION EQUAL 10.0)
+                    LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=2905)
+                ENDIF()
+
+                # avoids warnings on host-device signature of 'std::__shared_count<>'
+                IF(CUDA_VERSION EQUAL 10.1)
+                    LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=2912)
+                ENDIF()
+
+                # avoids warnings on host-device signature of 'std::__shared_count<>'
+                IF(CUDA_VERSION EQUAL 10.2)
+                    LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=2976)
+                ENDIF()
+
                 IF(ALPAKA_CUDA_KEEP_FILES)
                     MAKE_DIRECTORY("${PROJECT_BINARY_DIR}/nvcc_tmp")
                     LIST(APPEND CUDA_NVCC_FLAGS "--keep" "--keep-dir" "${PROJECT_BINARY_DIR}/nvcc_tmp")


alpakaConfig.cmake
```diff
@@ -675,6 +675,21 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                     LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored)
                 ENDIF()
 
+                # avoids warnings on host-device signatured for 'std::__shared_count<>'
```
sbastrakov: `signatured for ` should probably be `signature of `

sbastrakov: Same in other 2 places.

sbastrakov (dismissed):

BenjaminW3 (dismissed):

alpakaConfig.cmake
```diff
@@ -675,6 +675,21 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                ENDIF()
+
+                # avoids warnings on host-device signature of 'std::__shared_count<>'
+                IF(CUDA_VERSION EQUAL 10.1)
+                    LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe --diag_suppress=2912")
+                ENDIF()
+
+                # avoids warnings on host-device signature of 'std::__shared_count<>'
+                IF(CUDA_VERSION EQUAL 10.2)
+                    LIST(APPEND CUDA_NVCC_FLAGS "-Xcudafe --diag_suppress=2976")
```
BenjaminW3: Does this do what it is expected to do? The other `-Xcudafe` uses do not have `"` around so that the `-Xcudafe` and the `--diag_suppress` are added as separate elements to the list.

psychocoderHPC: Yes it is doing what is should. `"-Xcudafe --diag_suppress=2976"` should be one parameter because this is then also avoiding that the order is changed. `--diag_suppress=2976` is the argument of `-Xcudafe`.

psychocoderHPC: somehow two cases crashed: https://travis-ci.org/ComputationalRadiationPhysics/alpaka/jobs/615075195#L5087 I will try it again without the quotes.

BenjaminW3 (approved):

krzikalla (commented):
I had success with these switches:
'-Xcudafe', '--diag_suppress=esa_on_defaulted_function_ignored'

It seems more compatible than specifying particular numbers.
However I think, that a proper fix would be to either 
- introduce special annotation macros for defaulted functions, which then resolve only to the actual annotation depending on the compiler pass 

or

- resolve the ALPAKA_FN_ACC macro in general only to the actual annotation depending on the compiler pass

psychocoderHPC:
> I had success with these switches:
> '-Xcudafe', '--diag_suppress=esa_on_defaulted_function_ignored'
> 
> It seems more compatible than specifying particular numbers.
> However I think, that a proper fix would be to either
> 
> * introduce special annotation macros for defaulted functions, which then resolve only to the actual annotation depending on the compiler pass
> 
> or
> 
> * resolve the ALPAKA_FN_ACC macro in general only to the actual annotation depending on the compiler pass

I have not seen anymore the warning since we added this suppression flags. You must take care that those flags depend on the compiler version.

Do you still see this warnings?

krzikalla:
We don't use cmake and hence your fix won't help us. 
We supressed a warning number in our make framework befroehand. 
Now the warning reappered and I found out about "esa_on_defaulted_function_ignored". Maybe it helps you too.

psychocoderHPC:
Ok, we set it for all CUDA versions too: https://github.com/alpaka-group/alpaka/blob/1efc4b9985891fdff2d90fbb1e8cfe23ec3c9998/cmake/alpakaCommon.cmake#L530

Issue #885, opened 2019-11-22
Add support for CUDA 10.2


BenjaminW3:
We should add CUDA 10.2 to CI.

Notable things from the Release Notes:
* `"nvcc now supports a new flag -Werror=all-warnings to turn all generated warnings into errors instead."` I will try this out.
* `"Add option -forward-unknown-to-host-compiler to nvcc that allows forwarding of options not recognized by nvcc to the host compiler."` May this be useful for us?
* `"CUDA 10.2 (Toolkit and NVIDIA driver) is the last release to support macOS for developing and running CUDA applications. Support for macOS will not be available starting with the next release of CUDA."` We never officially supported this but now it is declared dead.
* `"Support for Kepler sm_30 architecture based products will be dropped starting with the next release of CUDA."` I did not find any code in alpaka that is dependent on this version. When `sm_35` is removed we can remove some workarounds.

psychocoderHPC:
> We should add CUDA 10.2 to CI.
> 
> Notable things from the Release Notes:
> 
> * `"nvcc now supports a new flag -Werror=all-warnings to turn all generated warnings into errors instead."` I will try this out.
> * `"Add option -forward-unknown-to-host-compiler to nvcc that allows forwarding of options not recognized by nvcc to the host compiler."` May this be useful for us?

I think it is a nice feature but since we still need to handle the old cuda versions I would keep explicit option handling.

> * `"CUDA 10.2 (Toolkit and NVIDIA driver) is the last release to support macOS for developing and running CUDA applications. Support for macOS will not be available starting with the next release of CUDA."` We never officially supported this but now it is declared dead.

+1

> * `"Support for Kepler sm_30 architecture based products will be dropped starting with the next release of CUDA."` I did not find any code in alpaka that is dependent on this version. When `sm_35` is removed we can remove some workarounds.

As I remember we only set for testing explicit sm_35 or so. Never the less we need to update our cmake because we activate sm_30 if CUDA 9+ is used https://github.com/ComputationalRadiationPhysics/alpaka/blob/066a2e9c30edd09d2831ef9ef8f69cce8fbcb120/alpakaConfig.cmake#L432-L436
If we fix this now and add the lines that for CUDA 11+ the default is sm_60 than we have a high change that the next alpaka release will run on CUDA 11 without other changes.

PR #886, opened 2019-11-23
Add support for CUDA 10.2

BenjaminW3:
Implements #885

Enables `-Werror=all-warnings` so that warnings from the cuda compiler frontend also make the build fail.


BenjaminW3:
The `nvcc 10.2 + clang 8.0` build currently fails with:
```
Nov 23 11:58:56 /usr/local/cuda-10.2/bin/nvcc /home/travis/build/BenjaminW3/alpaka/example/bufferCopy/src//bufferCopy.cpp -x=cu -c -o /home/travis/build/BenjaminW3/alpaka/build/example/bufferCopy/CMakeFiles/bufferCopy.dir/src/./bufferCopy_generated_bufferCopy.cpp.o -ccbin /home/travis/llvm/bin/clang++ -m64 -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_DEBUG=0 -DALPAKA_CI -Xcompiler ,\"-stdlib=libc++\",\"-fopenmp=libomp\",\"-O3\",\"-DNDEBUG\" --expt-extended-lambda --expt-relaxed-constexpr --generate-code arch=compute_70,code=sm_70 --generate-code arch=compute_70,code=compute_70 -std=c++11 --use_fast_math --ftz=false -Xcudafe --display_error_number -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored -Xcudafe --diag_suppress=2976 -Werror=all-warnings -DNVCC -I/usr/local/cuda-10.2/include -I/usr/include -I/home/travis/build/BenjaminW3/alpaka/include -I/home/travis/boost
/usr/local/cuda-10.2/include/crt/math_functions.h(8983): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.h(8996): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.h(9004): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.h(9018): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.hpp(375): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.hpp(381): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.hpp(383): error #311: cannot overload functions distinguished by return type alone
/usr/local/cuda-10.2/include/crt/math_functions.hpp(389): error #311: cannot overload functions distinguished by return type alone
8 errors detected in the compilation of "/tmp/tmpxft_00000136_00000000-6_bufferCopy.cpp1.ii".
Nov 23 11:58:59 -- Removing /home/travis/build/BenjaminW3/alpaka/build/example/bufferCopy/CMakeFiles/bufferCopy.dir/src/./bufferCopy_generated_bufferCopy.cpp.o
Nov 23 11:58:59 /home/travis/CMake/bin/cmake -E remove /home/travis/build/BenjaminW3/alpaka/build/example/bufferCopy/CMakeFiles/bufferCopy.dir/src/./bufferCopy_generated_bufferCopy.cpp.o
CMake Error at bufferCopy_generated_bufferCopy.cpp.o.Release.cmake:279 (message):
  Error generating file
  /home/travis/build/BenjaminW3/alpaka/build/example/bufferCopy/CMakeFiles/bufferCopy.dir/src/./bufferCopy_generated_bufferCopy.cpp.o
Nov 23 11:58:59 example/bufferCopy/CMakeFiles/bufferCopy.dir/build.make:63: recipe for target 'example/bufferCopy/CMakeFiles/bufferCopy.dir/src/bufferCopy_generated_bufferCopy.cpp.o' failed
Nov 23 11:58:59 make[2]: Leaving directory '/home/travis/build/BenjaminW3/alpaka/build'
make[2]: *** [example/bufferCopy/CMakeFiles/bufferCopy.dir/src/bufferCopy_generated_bufferCopy.cpp.o] Error 1
make[1]: *** [example/bufferCopy/CMakeFiles/bufferCopy.dir/all] Error 2
Nov 23 11:58:59 CMakeFiles/Makefile2:1017: recipe for target 'example/bufferCopy/CMakeFiles/bufferCopy.dir/all' failed
Nov 23 11:58:59 make[1]: Leaving directory '/home/travis/build/BenjaminW3/alpaka/build'
make: *** [all] Error 2
Nov 23 11:58:59 Makefile:94: recipe for target 'all' failed
```
Any ideas what we can do about this?

sbastrakov:
@BenjaminW3 I don't. It might be an incompatibility between these compiler and CUDA versions, not the [first](https://github.com/torch/cutorch/issues/834) time it [happens](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=864840). But I don't know for sure, might still be smth on our side.

alpakaConfig.cmake
```diff
@@ -462,9 +462,9 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                     IF(CUDA_VERSION GREATER_EQUAL 10.0)
                         MESSAGE(FATAL_ERROR "Clang versions lower than 8 do not support CUDA 10.0 or greater!")
                     ENDIF()
-                ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 9.0)
+                ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 10.0)
                     IF(CUDA_VERSION GREATER_EQUAL 10.1)
-                        MESSAGE(FATAL_ERROR "Clang versions lower than 9 do not support CUDA 10.1 or greater!")
+                        MESSAGE(FATAL_ERROR "Clang versions lower than 10 do not support CUDA 10.1 or greater!")
```
ax3l: that's not quite right: clang -x cuda 9.0 supports CUDA 10.1:
https://gist.github.com/ax3l/9489132#clang--x-cuda

alpakaConfig.cmake
```diff
@@ -678,16 +686,14 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+                ELSEIF(CUDA_VERSION EQUAL 10.1)
                     LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=2912)
+                ELSEIF(CUDA_VERSION EQUAL 10.2)
+                    LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=2976)
                 ENDIF()
 
-                # avoids warnings on host-device signature of 'std::__shared_count<>'
                 IF(CUDA_VERSION EQUAL 10.2)
-                    LIST(APPEND CUDA_NVCC_FLAGS -Xcudafe --diag_suppress=2976)
+                    LIST(APPEND CUDA_NVCC_FLAGS -Werror=all-warnings)
```
ax3l: `-Werror` of any kind should never be shipped to users. Activate only in CI, but not in CMake default flags. We cannot assume downstream usage and will break a lot when setting this.

BenjaminW3: You are right, I will add this to the tests only.

alpakaConfig.cmake
BenjaminW3: I will try this out.

alpakaConfig.cmake
```diff
@@ -462,9 +462,9 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                     IF(CUDA_VERSION GREATER_EQUAL 10.0)
                         MESSAGE(FATAL_ERROR "Clang versions lower than 8 do not support CUDA 10.0 or greater!")
                     ENDIF()
-                ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 9.0)
-                    IF(CUDA_VERSION GREATER_EQUAL 10.1)
-                        MESSAGE(FATAL_ERROR "Clang versions lower than 9 do not support CUDA 10.1 or greater!")
+                ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 10.0)
```
ax3l, line 469: This is still a bit off. You need two blocks here and [LLVM trunk](https://github.com/llvm-mirror/clang/blob/master/include/clang/Basic/Cuda.h) (upcoming Clang 10.0) does not yet add CUDA 10.2 support:
```cmake
ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 9.0)
    IF(CUDA_VERSION GREATER_EQUAL 10.1)
        MESSAGE(FATAL_ERROR "Clang versions lower than 9 do not support CUDA 10.1 or greater!")
    ENDIF()
ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 10.0)
    IF(CUDA_VERSION GREATER_EQUAL 10.2)
        MESSAGE(FATAL_ERROR "Clang versions lower than 10 do not support CUDA 10.2 or greater!")
    ELSE()
        MESSAGE(WARNING "At the time of release it was UNKNOWN if Clang 10 supports CUDA 10.2!")
    ENDIF()
ENDIF()
```
https://gist.github.com/ax3l/9489132#clang--x-cuda

BenjaminW3:
The problems of nvcc 10.2 with clang 8.0 within `math_function.h` and `math_function.hpp` are some declarations of `isnan` and `isinf`. I do not see anything we could do about this so I will disable this combination for now.

alpakaConfig.cmake
BenjaminW3: We already have this compiler combination in our CI build. It was not refelcted by the cmake checks.

alpakaConfig.cmake
BenjaminW3: I know, was only a temporary test ;-)

ax3l:
Interesting, Clang 8.0 host-compiler support is not even new but part of 10.1.168+ already... Does compiling in `Release` mode help?

BenjaminW3:
I tried both, release and debug. Does not help. With nvcc 10.1 it works in CI. They probably broke something.

ax3l:
Urgh, thanks for testing!

ax3l:
Feedback from my Nvidia contact: Clang 8.0 works, but we should not pass `-Xcompiler "-stdlib=libc++"` which breaks compatibility in the build matrix. Can you re-enable accordingly, please?

BenjaminW3:
You are right, with libstdc++it seems to work.

BenjaminW3:
Ready for approval/merging!

test/unit/queue/src/CollectiveQueue.cpp
```diff
@@ -25,7 +25,7 @@
 struct QueueCollectiveTestKernel
 {
     template<typename TAcc>
-    ALPAKA_FN_ACC auto operator()(
+    auto operator()(
```
sbastrakov, line 28: Is this change related to the CUDA 10.2 support?

BenjaminW3: Yes. This fixes a host-device warning that gets an error with the 10.2 werror setting.

alpakaConfig.cmake
```diff
@@ -466,6 +466,10 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
                     IF(CUDA_VERSION GREATER_EQUAL 10.1)
                         MESSAGE(FATAL_ERROR "Clang versions lower than 9 do not support CUDA 10.1 or greater!")
                     ENDIF()
+                ELSEIF(CMAKE_CXX_COMPILER_VERSION LESS 10.0)
+                    IF(CUDA_VERSION GREATER_EQUAL 10.2)
+                        MESSAGE(FATAL_ERROR "Clang versions lower than 10 do not support CUDA 10.2 or greater!")
```
sbastrakov, line 471: In the updated readme the table states that cuda 10.1 - 10.2 is supported for clang 4 - 8. Seems to contradict this error message. (Perhaps the error message is right, but the  table in the readme is wrong).

sbastrakov (dismissed):

alpakaConfig.cmake
sbastrakov: Same slightly upper with clang 9 and cuda 10.1

BenjaminW3: ~Yes, the ReadMe is wrong. nvcc, 10.2 + clang 9 is not supported.~

BenjaminW3: After looking at it again, everything seems to be correct. Those checks here are for the clang native CUDA support, not about the clang versions supported by nvcc.
What we check in the previous condition is that clang 1 to 8 do not support CUDA 10.1+ which is true.
In the marked line we say that clang 1 to 9 do not support CUDA 10.2+ which is also true.
We did not change anything in the readme for clangs supported CUDA versions.

sbastrakov: Ah, my bad, thanks for clarifying.

sbastrakov (approved):

alpakaConfig.cmake
ax3l: yeah, but it is unknown if Clang 10 will support CUDA 10.2, that's why I suggested a different message there ;) Trunk is not yet updated for CUDA 10.2

test/CMakeLists.txt
```diff
@@ -57,14 +57,19 @@ SET_TARGET_PROPERTIES(
...
+IF(ALPAKA_ACC_GPU_CUDA_ENABLE AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
+    # NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
     GET_TARGET_PROPERTY(_COMMON_COMPILE_OPTIONS common COMPILE_OPTIONS)
     # If the property does not exist, the variable is set to NOTFOUND.
     IF(_COMMON_COMPILE_OPTIONS)
         STRING(REPLACE ";" " " _COMMON_COMPILE_OPTIONS_STRING "${_COMMON_COMPILE_OPTIONS}")
         SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_COMMON_COMPILE_OPTIONS_STRING}")
     ENDIF()
+    # nvcc supports werror starting with 10.2
+    IF(CUDA_VERSION GREATER_EQUAL 10.2)
```
ax3l, line 69: Adding this here is not ideal either, since for example package managers in unanticipated environments might run your tests and a simple warning will mark a package as broken.

Please just enable `-Werror=all-warnings` in CI, e.g. via a travis environment variable or additional CMake option of Alpaka (that is by default not on). Shipping auto-enabled `-Werror` flags of any kind in CMake scripts is a bug :)

BenjaminW3: We already did this for years with the normal compilers Werror flag as most other open source projects are also doing. So this is at least not regression.
Furthermore it is the exact goal of setting the Werror flag to let the tests fail if they do not compile cleanly. Why should we treat warnings differently to compile errors? Having no watnings is our definition of our tests passing. So if a package manager for any reason (I do not know why it should do this at all) runs the tests and they produce a warning on this platform, then our tests fail and this is exactly what we want. That's the whole goal behind Werror.

ax3l: I am just telling you from a package management point of view, that will use software with newer environments than what were available and testable at release, that this is not an ideal practice.

Package managers such as spack, homebrew, et al. run tests to see if a package is working in a larger environment (integration tests). One usually just runs what comes with the package, to avoid duplication and maintenance overhead.

As developers, we can recommend best practices but it is not our place to anticipate new additions to `-Wall`, `-Wextra` or whatever a compiler decides to warn on. There are more compilers and workflows out there than what we do daily, e.g. people that build their own compiler frontends, optimizer passes, tooling, etc. on projects such as ours. Maybe a new NVCC just warns on something totally minor. Let's not patronize them. Warnings are called warnings for a reason.

BenjaminW3: I see your point. However a newer, self built or unknown compiler is in no way different to newer CUDA versions, newer or self built CMake versions, newer OSs or any other dependency. The Compiler is also only a dependency of alpaka. All of those dependencies can make the build or the tests fail at compile or runtime.
I do not see a reason why we should have lower requirements than is possible on the compilation while we expect everything to generate, run and calculate perfectly.
If we lower the requirements on one dependency, we should also reduce the requirements on numerical stability, quality of random numbers and other things that may behave unexpectedly on untested platforms.
Finding issues on untested platforms is the goal of tests and for alpaka a clean compilation is what we defined as a requirement for a successful test execution.

ax3l: > If we lower the requirements on one dependency, we should also reduce the requirements on numerical stability, quality of random numbers and other things that may behave unexpectedly on untested platforms.

I think it's not that black and white. A compiler warning in a newer nvcc might just be fine, no reason to break the tests outside of CI for this. A test that passes and throws a warning is still a tests that passes its primary *coded* purpose. A CMake flag that also enables `-Werror` intentionally can then also be used on top of this.

This is just compartmentalization of goals into smaller scopes of interest. Our goal in mainline is to have the code base error and warning free. That's why we require `-Werror` for all code to come in. A user's primary downstream goal is to have the tests technically compute the right results as a baseline for whatever they want to do. They can decide if they want to treat warnings as errors as well. I tell you from experience that many use cases do not want this to be the case, because warnings are not errors, some warnings are false positives in new environments, etc. Let's just optionalize this and enable this for us.

PR #889, opened 2019-11-27
remove deprecated NVIDIA architectures

psychocoderHPC:
~~With the next CUDA release (10.3 or 11.0) the support for the compute architectures <60 will be removed.~~ With CUDA 10.2+, SM 3.5-5.0 are marked as deprecated. SM 3.0 will be dropped in the next release (10.3 or 11.0).

This PR avoids that alpaka will fail because of using a not supported compute architectures by default.


sbastrakov (dismissed):

sbastrakov (dismissed):

BenjaminW3 (approved):

alpakaConfig.cmake
```diff
@@ -431,8 +431,10 @@ IF(ALPAKA_ACC_GPU_CUDA_ENABLE)
             SET(ALPAKA_CUDA_VERSION "${CUDA_VERSION}")
             IF(CUDA_VERSION VERSION_LESS 9.0)
                 SET(ALPAKA_CUDA_ARCH "20" CACHE STRING "GPU architecture")
-            ELSE()
+            ELSEIF(CUDA_VERSION VERSION_LESS 10.3)
                 SET(ALPAKA_CUDA_ARCH "30" CACHE STRING "GPU architecture")
+            ELSE()
+                SET(ALPAKA_CUDA_ARCH "60" CACHE STRING "GPU architecture")
```
BenjaminW3: I am not sure if this is the most defensive thing we can do. The release notes only say that sm_30 will be removed. The others are only deprecated. We could use `sm_35` as new default.

ax3l: I agree with Benjamin, do not assume removal yet for all but the announced SM 3.0 drop. Also, the deprecation period could go on for more than one release.
```suggestion
                SET(ALPAKA_CUDA_ARCH "35" CACHE STRING "GPU architecture")
```

ax3l:
```suggestion
                SET(ALPAKA_CUDA_ARCH "35" CACHE STRING "GPU architecture")
```

BenjaminW3 (dismissed):

alpakaConfig.cmake
sbastrakov: Yes, also discussed offline with @psychocoderHPC that setting to `sm_35` is better.

psychocoderHPC:
I updated the PR

BenjaminW3 (approved):

sbastrakov (approved):

PR #892, opened 2019-12-04
Structure fix

j-stephan:
This PR updates the library structure to the renamed concepts. Note that this is a quick and dirty fix for the SVGs since they don't seem to play nice with Inkscape.

This PR also adds an entry to .gitignore in order to ignore local .vimrc files.


BenjaminW3 (dismissed):
Thank you very much! I do not even know which program I used to create those initially.

I think we can update even more things. All the things called `*Stl` are now called `*StdLib` (`Rand*`, `Math*`, `Atomic*`). Furthermore I think we renamed the `Size` concept to `Idx`.

BenjaminW3:
May it be possible to squash all your commits into a single one?

BenjaminW3:
@j-stephan Do you have time to incorporate my suggested improvements?

j-stephan:
Yeah, sorry for the delay. I'm currently focused on my thesis defence which is due on Tuesday. I should be able to get back to this afterwards, or maybe even this weekend if time permits.

j-stephan:
Okay, I renamed everything. Shouldn't the commits be squashed automatically via GitHub's "squash and merge"?

psychocoderHPC:
> Okay, I renamed everything. Shouldn't the commits be squashed automatically via GitHub's "squash and merge"?

It would be nice if you can squash it by hand. We do not use this option because mostly the PRs are well split into commits. So a later review or diff is easier.

psychocoderHPC:
OK since the tests are mostly finished I will use this time the squash button.

psychocoderHPC (approved):

PR #894, opened 2019-12-09
implement queue properties

### LINKED ISSUE #175: trait to get stream type
Currently the user needs background knowledge to get the right stream type depending on the selected device.
We need something that gets the stream out of the device or maybe out of the platform if #167 is merged.

Current usage:

``` C++
using AccDev = ::alpaka::dev::DevCpu;
//async stream
using AccStream = ::alpaka::stream::StreamCpuAsync;
//sync stream
using AccStream = ::alpaka::stream::StreamCpuAsync;

//or for GPU
using AccDev = ::alpaka::dev::DevCudaRt;
using AccStream = ::alpaka::stream::StreamCudaRtAsync;
using AccStream = ::alpaka::stream::StreamCudaRtSync;
```

My preferred way:

``` C++
using AccDev = ::alpaka::dev::DevCpu;
//async stream
using AccStream = ::alpaka::stream::getType<AccDev, ::alpaka::stream::Async>;
//sync stream
//using AccStream = ::alpaka::stream::getType<AccDev, ::alpaka::stream::Sync>;

//or for GPU
using AccDev = ::alpaka::dev::DevCudaRt;
using AccStream = ::alpaka::stream::getType<AccDev, ::alpaka::stream::Async>
using AccStream = ::alpaka::stream::getType<AccDev, ::alpaka::stream::Sync>
```

This needs no knowledge about the naming and which devices needs which stream type.

psychocoderHPC:
close #175

Implement a trait `queue::Queue` to derive the queue from the
accelerator/device or the platform based on a property.

- add concepts for
  - alpaka platforms
  - alpaka accelertors
- add queue trait `Queue`
- add queue properies e.g. `Blocking`, `NonBlocking`
- move declaration of `ImplementsConcept` from `detail` into the `concepts` namespace

Removes the precompiler guard that examples can only be useful executed
with the serial accelerator.

- add missing synchronization in the examples to support nonblocking
queues

The PR is split into two commits, it is easier to check them seperate.

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -130,21 +130,50 @@ struct FillBufferKernel
 auto main()
 -> int
 {
-// This example is hard-coded to use the sequential backend.
-#if defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
-
+// Fallback for the CI with disabled sequential backend
+#if defined(ALPAKA_CI) && !defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+    return EXIT_SUCCESS;
+#else
     // Define the index domain
     using Dim = alpaka::dim::DimInt<3u>;
     using Idx = std::size_t;
 
-    // Define the accelerator
+    // Define the device accelerator
+    //
+    // It is possible to choose from a set of accelerators
+    // that are defined in the alpaka::acc namespace e.g.:
+    // - AccGpuCudaRt
+    // - AccCpuThreads
+    // - AccCpuFibers
+    // - AccCpuOmp2Threads
+    // - AccCpuOmp2Blocks
+    // - AccCpuOmp4
+    // - AccCpuSerial
     using Acc = alpaka::acc::AccCpuSerial<Dim, Idx>;
-    using DevQueue = alpaka::queue::QueueCpuBlocking;
+    // Defines the synchronization behavior of a queue
+    //
+    // choose between Blocking and NonBlocking
+    using AccQueueProperty = alpaka::queue::Blocking;
+    using DevQueue = alpaka::queue::Queue<Acc, AccQueueProperty>;
     using DevAcc = alpaka::dev::Dev<Acc>;
     using PltfAcc = alpaka::pltf::Pltf<DevAcc>;
 
+    // Define the device accelerator
+    //
+    // It is possible to choose from a set of accelerators
+    // that are defined in the alpaka::acc namespace e.g.:
+    // - AccCpuThreads
+    // - AccCpuFibers
+    // - AccCpuOmp2Threads
+    // - AccCpuOmp2Blocks
+    // - AccCpuOmp4
+    // - AccCpuSerial
     using Host = alpaka::acc::AccCpuSerial<Dim, Idx>;
-    using HostQueue = alpaka::queue::QueueCpuBlocking;
+    // Defines the synchronization behavior of a queue
+    //
+    // choose between Blocking and NonBlocking
+    using HostQueueProperty = alpaka::queue::Blocking;
+    using HostQueue = alpaka::queue::Queue<Host, HostQueueProperty>;
     using DevHost = alpaka::dev::Dev<Host>;
     using PltfHost = alpaka::pltf::Pltf<DevHost>;
 
@@ -302,6 +331,7 @@ auto main()
         pDeviceBuffer1,                                 // 1st kernel argument
         extents,                                        // 2nd kernel argument
         deviceBuffer1Pitch);                            // 3rd kernel argument
+    alpaka::wait::wait(devQueue);
     std::cout << std::endl;
 
     alpaka::kernel::exec<Acc>(
@@ -311,6 +341,7 @@ auto main()
         pDeviceBuffer2,                                 // 1st kernel argument
         extents,                                        // 2nd kernel argument
         deviceBuffer2Pitch);                            // 3rd kernel argument
+    alpaka::wait::wait(devQueue);
     std::cout << std::endl;
 
     alpaka::kernel::exec<Host>(
@@ -320,6 +351,7 @@ auto main()
         pHostBuffer,                                    // 1st kernel argument
         extents,                                        // 2nd kernel argument
         hostBuffer1Pitch);                              // 3rd kernel argument
+    alpaka::wait::wait(hostQueue);
     std::cout << std::endl;
 
     alpaka::kernel::exec<Host>(
@@ -329,11 +361,9 @@ auto main()
         pHostViewPlainPtr,                              // 1st kernel argument
         extents,                                        // 2nd kernel argument
         hostViewPlainPtrPitch);                         // 3rd kernel argument
+    alpaka::wait::wait(hostQueue);
     std::cout << std::endl;
 
-    return EXIT_SUCCESS;
-
-#else
     return EXIT_SUCCESS;
 #endif
 }
--- a/include/alpaka/acc/AccGpuHipRt.hpp
+++ b/include/alpaka/acc/AccGpuHipRt.hpp
@@ -39,6 +39,7 @@
 
 // Implementation details.
 #include <alpaka/core/ClipCast.hpp>
+#include <alpaka/core/Concepts.hpp>
 #include <alpaka/dev/DevHipRt.hpp>
 #include <alpaka/core/Hip.hpp>
 
@@ -78,7 +79,8 @@ namespace alpaka
             public block::shared::st::BlockSharedMemStHipBuiltIn,
             public block::sync::BlockSyncHipBuiltIn,
             public rand::RandHipRand,
-            public time::TimeHipBuiltIn
+            public time::TimeHipBuiltIn,
+            public concepts::Implements<ConceptAcc, AccGpuHipRt<TDim, TIdx>>
         {
         public:
             //-----------------------------------------------------------------------------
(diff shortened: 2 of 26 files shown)


psychocoderHPC:
This PR solves the problem from #175. It will not give us the possibility to have a list of accelerators and run the example on all accelerators which are available (selected in CMake).
This is IMO important but no strong requirement for the 0.4.0 release.

include/alpaka/queue/Traits.hpp
```diff
@@ -78,5 +86,12 @@ namespace alpaka
                 ::empty(
                     queue);
         }
+
+        //-----------------------------------------------------------------------------
+        //! Queue based on the environment and a property
```
sbastrakov, line 91: Maybe clarify here that out-of-the-box the environment type can be accelerator, device or platform type.

psychocoderHPC:
suggestion: we can add ONLY to the CMake of the examples an option `ALPAKA_DEFAULT_ACC` where one of the activated accelerators can be selected and use this one in the code. This would give us the possibility to select the accelerator from CMake instead of have `AccCpuSerial` hard coded.

include/alpaka/core/Concepts.hpp
```diff
@@ -25,20 +25,20 @@ namespace alpaka
         {
         };
 
-        namespace detail
-        {
+        template<
+            typename TConcept,
+            typename TDerived>
+        struct ImplementsConcept {
```
BenjaminW3, line 33: Can you add some documentation to this trait now that it is public? Something like `Checks whether the concept is implemented by the given class`.

include/alpaka/pltf/PltfCpu.hpp
```diff
@@ -21,7 +21,8 @@ namespace alpaka
     {
         //#############################################################################
         //! The CPU device platform.
-        class PltfCpu
+        class PltfCpu :
+            public concepts::Implements<ConceptPltf, void>
```
BenjaminW3: Can you please use the type of the current class `PltfCpu` instead of void? Please change this here and in all other uses. This allows to use the rest of the concept base type logic later. I would also like to add a `static_assert` into `Implements` that this parameter is not `void`.

include/alpaka/pltf/PltfHipRt.hpp
```diff
@@ -41,6 +42,12 @@ namespace alpaka
         };
     }
 
+    namespace queue
+    {
+        class QueueHipRtBlocking;
+        class QueueHipRtNonBlocking;
+    }
+
```
BenjaminW3: Are those forward declarations necessary? You did not add them to the CPU and CUDA versions.

include/alpaka/dev/DevHipRt.hpp
```diff
@@ -12,6 +12,7 @@
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
 
 #include <alpaka/core/BoostPredef.hpp>
+#include <alpaka/core/Concepts.hpp>
```
BenjaminW3: You do not use concepts here.

include/alpaka/dev/DevCudaRt.hpp
```diff
@@ -12,6 +12,7 @@
 #ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
 
 #include <alpaka/core/BoostPredef.hpp>
+#include <alpaka/core/Concepts.hpp>
```
BenjaminW3: You do not use concepts here.

include/alpaka/dev/DevCpu.hpp
```diff
@@ -16,8 +16,12 @@
 
 #include <alpaka/queue/cpu/ICpuQueue.hpp>
 #include <alpaka/core/Unused.hpp>
+#include <alpaka/core/Concepts.hpp>
```
BenjaminW3: You do not use concepts here.

include/alpaka/queue/Properties.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/queue/Traits.hpp>
```
BenjaminW3: We probably do not need this include.

include/alpaka/queue/Traits.hpp
```diff
@@ -40,6 +40,14 @@ namespace alpaka
...
                 typename TSfinae = void>
             struct Empty;
+
+            //#############################################################################
+            //! Queue for an accelerator
+            template<
+                typename TAcc,
+                typename TProperty,
+                typename TSfinae = void>
+            struct Queue;
```
BenjaminW3: To be consistent with all the other traits this would have to be called `QueueType`. Only the public alias templates are called `Queue` (without `Type`).

include/alpaka/pltf/PltfCpu.hpp
psychocoderHPC: OK will use the class itself e.g. `public concepts::Implements<ConceptPltf, PltfCpu>`

include/alpaka/pltf/PltfHipRt.hpp
psychocoderHPC: no not needed. copy past issue

BenjaminW3:
Thanks for implementing this!

include/alpaka/acc/Traits.hpp
```diff
@@ -94,4 +101,27 @@ namespace alpaka
...
+        namespace traits
+        {
+            template<
+                typename TAcc,
+                typename TProperty>
+            struct Queue<
+                TAcc,
+                TProperty,
+                typename std::enable_if<
+                    concepts::ImplementsConcept<acc::ConceptAcc, TAcc>::value
```
sbastrakov, line 116: As a thought. Perhaps there is a sort of name collision between `concepts::ImplementsConcept` which checks if a concept is implemented (and this PR makes it public) and `concepts::Implements` which declares a concept is implemented. Maybe `concepts::ImplementsConcept` can be renamed to smth like `concepts::IsImplemented`?

include/alpaka/acc/AccGpuHipRt.hpp
```diff
@@ -78,7 +79,8 @@ namespace alpaka
             public block::shared::st::BlockSharedMemStHipBuiltIn,
             public block::sync::BlockSyncHipBuiltIn,
             public rand::RandHipRand,
-            public time::TimeHipBuiltIn
+            public time::TimeHipBuiltIn,
+            public concepts::Implements<ConceptAcc, void>
```
BenjaminW3: Can you please replace the `void`s with their implementation here and in the other accelerator classes as well? So `AccGpuHipRt` here?

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -130,21 +130,50 @@ struct FillBufferKernel
 auto main()
 -> int
 {
-// This example is hard-coded to use the sequential backend.
-#if defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
-
+// This example is working if CUDA/HIP is exclusively used
```
sbastrakov: I am confused. Is the example working or not working for exclusive CUDA / HIP? From the code looks like not.

psychocoderHPC: If you use the alpaka exclusive cmake options e.g. `ALPAKA_ACC_GPU_CUDA_ONLY_MODE` or ` ALPAKA_ACC_GPU_CUDA_ONLY_MODE` you will have only a device accelerator but some examples require a host and a device accelerator.

BenjaminW3 (dismissed):

BenjaminW3:
You still have problems with the Windows CUDA builds because they do not have the serial accelerator backend enabled. CI does this, becaue compile times on Windows were bad. They are much better now, so we could enable it again. However, the examples should build in all combinations of enabled accelerators.

psychocoderHPC:
> They are much better now, so we could enable it again. However, the examples should build in all combinations of enabled accelerators.

This is a task I like to do after the release. It requires more changes.

psychocoderHPC:
Yes after fighting over one week with the CI the PR passed.

include/alpaka/acc/Traits.hpp
sbastrakov: @psychocoderHPC do you plan to change it in this PR? It's just a name after all, not critical.

include/alpaka/queue/Traits.hpp
```diff
@@ -78,5 +86,14 @@ namespace alpaka
                 ::empty(
                     queue);
         }
+
+        //-----------------------------------------------------------------------------
+        //! Queue based on the environment and a property
+        //
+        // Environment type can be a accelerator, device or a platform.
```
sbastrakov: Perhaps add `\tparam` comments, since this class is supposed to be used in a client app (and the updated examples do so)?
Also, `a accelerator` should be `an accelerator`.

include/alpaka/acc/Traits.hpp
psychocoderHPC: I will not change the naming inside of this PR because it is out of the scope.

sbastrakov (approved):

include/alpaka/queue/Traits.hpp
psychocoderHPC: fixed

sbastrakov:
@psychocoderHPC @BenjaminW3 is there a reason TBB is not in the list of accelerators for the examples? I mean [such lists](https://github.com/ComputationalRadiationPhysics/alpaka/blob/432331fcc737445ac43ebcd54f82d68d3a82d571/example/vectorAdd/src/vectorAdd.cpp#L90).

psychocoderHPC:
I think we missed to add TBB in the list

PR #895, opened 2019-12-10
Change of supported gcc version in cuda8 to gcc 5.4

SimeonEhrig:
The official CUDA 8 Docker container from nvidia uses gcc 5.4.


psychocoderHPC:
Gcc 5.4 was disabled by https://github.com/ComputationalRadiationPhysics/alpaka/pull/566 because of issues with `std::tuple`. We need to check if this problem still exists with the current development branch.

SimeonEhrig:
I tested Alpaka with `ctest` and the official Nvidia Docker container (GCC 5.4.0, CUDA 8.0.61) with activated and without activated CUDA backend. All tests passed. I'm not sure if the error occurs with GCC 5.4 or 5.5. I think it was also not clear in #566 and Travis only offers GCC 5.5, so the result was that the version was limited to the officially supported version 5.3.

BenjaminW3:
We do not have any nvcc 8.0 + gcc 5.* build in our CI tests. We should probably add this to see if it really works.

ax3l:
Interesting, but maybe this official docker container uses a patched gcc and/or stdlib.
As Rene said, although this should in principle work, nvcc breaks if including anything related to tuples as we found.

Just in case you want to check, some refs.:
- https://gist.github.com/ax3l/9489132
- https://gitlab.com/nvidia/container-images/cuda

PrometheusPi (dismissed):

SimeonEhrig:
I added the CUDA 8 GCC 5.5 test and it crashed. The bug is known and has been discussed in this issue: https://github.com/ComputationalRadiationPhysics/picongpu/issues/2388

I think nvidia uses a modified version in the Docker container. So, how can we solve this problem? My idea is to undo the change in the GCC version check and add a CMake flag that can disable the version check. Of course, the default setting is that version checking is enabled.

ax3l:
I think if vanilla GCC does not work reliably then we should not support it. We don't have time to discuss/support such corner cases with users when they face it and as far as I remember there is no urgent use case that would need this?

psychocoderHPC:
I will close this PR because #900 will remove the support for CUDA 8. 
If this feature is strongly required we should push it as bugfix to alpaka 0.4.1

SimeonEhrig:
It's okay because I only need CUDA 8 for Cling and Cling uses the standalone header instead of cmake.

PR #896, opened 2019-12-10
fix memory alignment

### LINKED ISSUE #873: memory pinning
Currently, we are only able to pin memory if we are using CUDA. We have an open issue to pin memory maybe with hwloc #634.
I am currently working on adding memory pinning for AMD HIP-clang and found the following documentation for [hipHostRegister](https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674)

```
...
Developers are strongly encouraged to register memory blocks which are aligned to the host cache-line size. (typically 64-bytes but can be obtains from the CPUID instruction).
...
```

The problem is that we in alpaka first allocate memory and then pin this memory. This will lead in data races when the driver is syncing two allocations within the same cache-line.
We are fighting since a few months in PIConGPU with an [issue 2996](https://github.com/ComputationalRadiationPhysics/picongpu/issues/2996)  which also points to this behavior on a Power9 processor.
For CUDA this behavior is not described in the official documentation.

## The problem:

Since our allocation and the pinning is independent we have no way to guarantee that we fulfill the allocation conditions (one allocation per cache-line for pinned memory).

## Possible Solutions

- extent the allocation interface and add the possibility to define that memory must be aligned and consume at least a full cache-line
  - this also requires that we can create pinned cpu buffer
- align each allocation to the cache-line size and consume at least a full cache-line
  - this can be negative for the performance if the user needs many small allocations

Please provide ideas since we need to solve this issue: @ComputationalRadiationPhysics/alpaka-developers

psychocoderHPC:
fix #873

Align host memory for alapka host buffer always to 4kib to avoid data
races when a GPU updates host memory.

I do not set the memory alignment for HIP to the minimum value (documentation says that memory must be aligned to a cache line). I can currently not clearly verify the needed alignment because the AMD tools are not working on the system where I have access. So I decided to stay conservative and added an alignment that works for HIP(cuda) and fulfills the requirements from the documentation.

### DIFF
--- a/include/alpaka/mem/alloc/AllocCpuBoostAligned.hpp
+++ b/include/alpaka/mem/alloc/AllocCpuBoostAligned.hpp
@@ -16,6 +16,8 @@
 
 #include <boost/align.hpp>
 
+#include <algorithm>
+
 namespace alpaka
 {
     namespace mem
@@ -51,10 +53,24 @@ namespace alpaka
                         std::size_t const & sizeElems)
                     -> T *
                     {
+#if (defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_LANG_CUDA) || (defined(ALPAKA_ACC_GPU_HIP_ENABLED) && BOOST_LANG_HIP)
+                        // For CUDA host memory must be aligned to 4 kib to pin it with `cudaHostRegister`,
+                        // this was described in older programming guides but was removed later.
+                        // From testing with PIConGPU and cuda-memcheck we found out that the alignment is still required.
+                        //
+                        // For HIP the required alignment is the size of a cache line.
+                        // https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674
+                        // To avoid issues with HIP(cuda) the alignment will be set also for HIP(clang/hcc)
+                        // to 4kib.
+                        // @todo evaluate requirements when the HIP ecosystem is more stable
+                        constexpr size_t minAlignement = 4096;
+#else
+                        constexpr size_t minAlignement = TAlignment::value;
+#endif
                         alpaka::ignore_unused(alloc);
                         return
                             reinterpret_cast<T *>(
-                                boost::alignment::aligned_alloc(TAlignment::value, sizeElems * sizeof(T)));
+                                boost::alignment::aligned_alloc(std::max(TAlignment::value, minAlignement), sizeElems * sizeof(T)));
                     }
                 };
 


include/alpaka/mem/alloc/AllocCpuBoostAligned.hpp
```diff
@@ -51,10 +53,24 @@ namespace alpaka
                         std::size_t const & sizeElems)
                     -> T *
                     {
+#if (defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_LANG_CUDA) || (defined(ALPAKA_ACC_GPU_HIP_ENABLED) && BOOST_LANG_HIP)
+                        // For CUDA host memory must be aligned to 4 kib to pin is with `cudaHostRegister`,
```
BenjaminW3:
```suggestion
                        // For CUDA host memory must be aligned to 4 kib to pin it with `cudaHostRegister`,
```

include/alpaka/mem/alloc/AllocCpuBoostAligned.hpp
```diff
@@ -51,10 +53,24 @@ namespace alpaka
...
+                        // this was described in older programming guides but was removed later.
+                        // From testing with PIConGPU and cuda-memcheck we found out that the alignment is still required.
+                        //
+                        // For HIP the required alignment is the size of a cache line.
+                        // https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674
+                        // To avoid issues with HIP(cuda) the alignment will be set also for HIP(clang/hcc)
+                        // to 4kib.
+                        // @todo evaluate requirements when the HIP ecosystem is more stable
+                        constexpr size_t minAlignement = 4096;
```
BenjaminW3, line 66: Could we change this to `std::max(TAlignment::value, 4096u)` and remove the std::max below so that we do not do `std::max(TAlignment::value, TAlignment::value)` for the other case? Compilers might probably optimize this but I think this would be clearer.

psychocoderHPC: I would like to keep it because this has already the structure to handle HIP and CUDA different and it is easy to read.
With C++14 `std::max` is constexpr and before it is one comparison before a very expensive malloc call.

sbastrakov: I like a separate `minAlignement` variable, since it has a clear meaning on its own and the limitations are defined in terms of this value. I would prefer additionally to make another variable, like `[constexpr in case it's allowed] size_t alignment = std::max(TAlignment::value, minAlignement);` and use it later for the `aligned_alloc` call.

BenjaminW3 (approved):

include/alpaka/mem/alloc/AllocCpuBoostAligned.hpp
ax3l: You can keep it as is and still make it more readable and compile-time in C++11 by writing:
```c++
constexpr size_t minAlignment = ...;

constexpr size_t alignment = (TAlignment::value > minAlignment) ? TAlignment::value : minAlignment;

boost::alignment::aligned_alloc(alignment, ...);
```

Typo in all vars: Alignment

PR #900, opened 2019-12-24
Raise minimum to -std=c++14 and remove support for CUDA 8.0 and gcc 4.9

BenjaminW3:
C++14 allows us to remove boilerplate code and lets us use new language
features (auto return type deduction, std::integer_sequence, generic
lambdas, ...).
GCC 4.9 as well as CUDA 8.0 do not support C++14 fully and are therefore
not supported anymore.

This removes more than 700 lines of boilerplate code! 🎉


BenjaminW3:
@ComputationalRadiationPhysics/alpaka-developers I would like to get your feedback:
When C++14 is enabled (`-std=c++14`) then many of the standard library functions are now `constexpr`.
This reults in many errors of the kind: `error #2891: calling a constexpr __host__ function("operator()") from a __device__ function("foldr") is not allowed. The experimental flag '--expt-relaxed-constexpr' can be used to allow this.` because `std::plus`, `std::multiplies`, `std::min`, `std::max`, `std::epsilon` and friends are now `constexpr`. So they can not be used in C++14 within kernels and large parts of alpaka when '--expt-relaxed-constexpr' is not defined.
This problem is independent of this PR and already exists independentently of making C++14 the default. It is not visible in CI because we do not test this combination.
My ask is now: Can we make the definition of `--expt-relaxed-constexpr` mandatory?

ax3l:
Good call, has my support. (Many CUDA/C++ developers I know argue this should be default in NVCC anyway.)

test/unit/kernel/src/KernelWithHostConstexpr.cpp
```diff
@@ -7,9 +7,6 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
-// NVCC needs --expt-relaxed-constexpr
-#if !defined(__NVCC__) || (defined(__NVCC__) && defined(__CUDACC_RELAXED_CONSTEXPR__))
```
psychocoderHPC, line 11: Not sure if we can remove this. Relaxed constexpr can still be disabled in cmake. This flag allows accessing `__host__ ` constexpr function on the device.

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#constexpr-functions

psychocoderHPC (changes requested):

psychocoderHPC:
> This problem is independent of this PR and already exists independentently of making C++14 the default. It is not visible in CI because we do not test this combination.
> My ask is now: Can we make the definition of `--expt-relaxed-constexpr` mandatory?

I would like it. The only problem with this flag is that it is pulling all host code signatures into the device compile path. Compiling will fail if you include a host library header e.g. jsoncpp in the user code where [bitfields](https://en.cppreference.com/w/cpp/language/bit_field) are used for host structures because nvcc is not supporting bitfields on the device.

ax3l:
Just use nlohmann/json instead ;)
https://github.com/nlohmann/json

BenjaminW3:
@psychocoderHPC 
> The only problem with this flag is that it is pulling all host code signatures into the device compile path.

Shouldn't it only bring the `constexpr __host__` functions into the device compile path? Does it also bring all `structs` into the device compile path even when they do not have a `contexpr` constructor?

psychocoderHPC:
I tried to reproduce the bitfield issue in a mini-app but without success.

The original error was with the lib `jsonCpp` -> `json.h(647): error: "Bitfields and field types containing bitfields are not supported in packed structures and unions for device compilation!"` 
Since the used app is closed source it is not possible to provide the original code :-(

@BenjaminW3 I think the future development of alpaka without host side constexpr functions will become very hard, so you have green light from my side to make `--expt-relaxed-constexpr` mandatory.

test/unit/kernel/src/KernelWithHostConstexpr.cpp
BenjaminW3: As discussed below we now enforce this. I will add a compile-time check.

psychocoderHPC (dismissed):

psychocoderHPC:
@BenjaminW3 could you please rebase to resolve the merge issue

BenjaminW3:
I resolved the conflicts.

BenjaminW3:
I finally got a green build. Please approve and merge! I already have 3 follow-up PRs in the queue ;-)

include/alpaka/meta/Filter.hpp
```diff
@@ -49,7 +49,7 @@ namespace alpaka
             {
                 using type =
                     typename std::conditional<
-                        TPred<T>::value,    // TODO: Remove '::value' when C++14 variable templates are supported.
+                        TPred<T>::value,
```
sbastrakov, line 52: Question: the `::value` is not removed as the deleted comment suggests, is it the intention?

README.md
```diff
@@ -91,7 +91,7 @@ However some of the accelerator back-end implementations require different boost
 When an accelerator back-end using *Boost.Fiber* is enabled, `boost-fiber` and all of its dependencies are required to be built in C++11 mode `./b2 cxxflags="-std=c++11"`.
```
sbastrakov, line 91: Is C++11 fine here, or should it also become C++14?

include/alpaka/meta/Filter.hpp
BenjaminW3: Yes, this was intended. The comment did not apply anymore.

README.md
BenjaminW3: C++11 should be enough. C++14 also works.

sbastrakov (approved):

psychocoderHPC (approved):

PR #862, opened 2019-10-18
math/abs: Added trait specialisation for double.

jkrude:
In include/alpaka/math/abs:
Int is also supported in cuda/hip, so fabs is used for double type.
abs for int
fabs for double <-- was missing
fabsf for float

PR #784, opened 2019-05-07
Fixes name conflicts in alpaka math functions.

tdd11235813:
Compilers like gcc4.9+nvcc8 failed to distinguish between variable and function names in the math traits:
```c++
// alpaka::math::min(...) in min/Traits.hpp:
        ALPAKA_FN_HOST_ACC auto min(
            T const & min, // <-- changed to min_ctx now
```
```c++
// implementation in min/MinCudaBuiltIn.hpp
__device__ static auto min(
                    MinCudaBuiltIn const & min, // <-- changed to min_ctx now
```

- changed CUDA and HIP, but also Std parts for the sake of consistency
- compiled and run all the tests on HIP+AMD system, although not all math functions are currently tested
  - unit time test only fails to build when the too old AMD GPU target '--amdgpu-target=gfx701' is used in alpakaConfig.cmake (is ToDo to provide a more convenient AMD GPU target selector via cmake)

The error you previously received:
> /home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/math/sin/SinCudaBuiltIn.hpp(58): error: expression preceding parentheses of apparent call must have (pointer-to-) function type

Edit: updated the doxygen parts, rerunning travis
Edit2: did not have time to update yet ...

PR #902, opened 2020-01-02
backport to 0.3.6

psychocoderHPC:
Backport last missing PR's

- fix: cuda exceptions #844
- math/abs: Added trait specialisation for double. #862
- alpaka/math Overloaded float specialization #837
- Fixes name conflicts in alpaka math functions. #784

# manual tests

- [x] use these changes together with PIConGPU 0.4.3

### DIFF
--- a/include/alpaka/core/Cuda.hpp
+++ b/include/alpaka/core/Cuda.hpp
@@ -85,6 +85,8 @@ namespace alpaka
                     std::cerr << sError << std::endl;
 #endif
                     ALPAKA_DEBUG_BREAK;
+                    // reset the last error to allow user side error handling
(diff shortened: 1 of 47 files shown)


BenjaminW3 (approved):

psychocoderHPC:
@BenjaminW3 I restarted one of the atomic tests but could be the test is running too long and therefore fails. If so we should ignore the test and merge it so that I can at least next week push the 0.3.6 bug fix release.

BenjaminW3:
As far as I remember, the test was already failing in the last few patch releases. It can be merged as is.

sbastrakov (approved):

PR #907, opened 2020-01-07
Replace alpaka::meta::IndexSequence with C++14 std::index_sequence

BenjaminW3:
This replaces one occurence. Others will follow later.

### DIFF
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
@@ -171,14 +171,14 @@ namespace alpaka
                                     throw std::runtime_error("The OpenMP 2.0 runtime did not create a parallel region!");
                                 }
 
-                                // GCC 5.1 fails with:
+                                // GCC fails with:
                                 // error: redeclaration of const int& iBlockThreadCount
                                 // if(numThreads != iBlockThreadCount)
                                 //                  ^
                                 // note: const int& iBlockThreadCount previously declared here
                                 // #pragma omp parallel num_threads(iBlockThreadCount)
                                 //         ^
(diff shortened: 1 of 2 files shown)


include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
```diff
@@ -171,14 +171,14 @@ namespace alpaka
...
 
-                                // GCC 5.1 fails with:
+                                // GCC fails with:
                                 // error: redeclaration of const int& iBlockThreadCount
                                 // if(numThreads != iBlockThreadCount)
                                 //                  ^
                                 // note: const int& iBlockThreadCount previously declared here
                                 // #pragma omp parallel num_threads(iBlockThreadCount)
                                 //         ^
-#if (!BOOST_COMP_GNUC) || (BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(5, 0, 0)) || (BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(6, 0, 0))
```
BenjaminW3, line 181: We do not support version lower than 5.0 at all so this is useless.

include/alpaka/meta/ApplyTuple.hpp
```diff
@@ -12,8 +12,6 @@
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Unused.hpp>
 
-#include <alpaka/meta/IntegerSequence.hpp>
```
ax3l, line 15: Can't this file itself be removed then as well? :)

BenjaminW3: It will get removed in the future but there are other usages that have to be replaced as well. I am already working on it but there seem to be some compiler issues.

sbastrakov (approved):

psychocoderHPC (approved):

PR #913, opened 2020-01-15
Add forgotten TBB accelerator to the list in comments of the examples

sbastrakov:
As pointed out [here](https://github.com/ComputationalRadiationPhysics/alpaka/pull/894#issuecomment-574127255). Thanks @kloppstock for discovering this

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -148,6 +148,7 @@ auto main()
     // - AccCpuOmp2Threads
     // - AccCpuOmp2Blocks
     // - AccCpuOmp4
+    // - AccCpuTbbBlocks
     // - AccCpuSerial
     using Acc = alpaka::acc::AccCpuSerial<Dim, Idx>;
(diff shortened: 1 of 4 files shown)


example/reduce/src/reduce.cpp
```diff
@@ -32,6 +32,7 @@
 // - CpuThreads
 // - CpuOmp2Blocks
 // - CpuOmp4
+// - CpuTBBBlocks
```
sbastrakov: Note: here I assume TBB is okay. Also this list does not have Fibers, and the names lack `Acc`.

psychocoderHPC: checked the file `alpakaConfig.hpp` of this example and found that there is no config for TBB available

sbastrakov: Right, and there is a comment it's not supported. Removed my change for this file, thanks for pointing out.

BenjaminW3 (approved):

ax3l (approved):

PR #915, opened 2020-01-17
Use C++14 helper types for traits

sbastrakov:
Replace the old `typename std::[trait]<...>::type` with C++14 `std::[trait]_t<...>`.
Add occasionally missing `#include <type_traits>`.

There is a similar helper for values, but this is C++17, so not changed.

TODO: test properly, since  I could have introduced typos.

### DIFF
--- a/test/common/include/alpaka/test/mem/view/Iterator.hpp
+++ b/test/common/include/alpaka/test/mem/view/Iterator.hpp
@@ -11,6 +11,8 @@
 
 #include <alpaka/alpaka.hpp>
 
+#include <type_traits>
+
 namespace alpaka
 {
     //-----------------------------------------------------------------------------
@@ -35,10 +37,10 @@ namespace alpaka
                     template<
                         typename T,
                         typename TSource>
-                    using MimicConst = typename std::conditional<
+                    using MimicConst = std::conditional_t<
                         std::is_const<TSource>::value,
-                        typename std::add_const<T>::type,
-                        typename std::remove_const<T>::type>;
+                        std::add_const_t<T>,
+                        std::remove_const_t<T>>;
 
 #if BOOST_COMP_GNUC
     #pragma GCC diagnostic push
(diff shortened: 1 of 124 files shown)


test/common/include/alpaka/test/mem/view/Iterator.hpp
```diff
@@ -35,10 +37,10 @@ namespace alpaka
                     template<
                         typename T,
                         typename TSource>
-                    using MimicConst = typename std::conditional<
+                    using MimicConst = std::conditional_t<
                         std::is_const<TSource>::value,
-                        typename std::add_const<T>::type,
-                        typename std::remove_const<T>::type>;
+                        std::add_const_t<T>,
+                        std::remove_const_t<T>>;
```
BenjaminW3, line 43: This is an alias template. You probably have to adjust the call to `typename MimicConst<T, TSource>::type` as well now to make the compiler happy.

BenjaminW3:
Thanks for modernizing the code!

test/common/include/alpaka/test/mem/view/Iterator.hpp
```diff
@@ -50,7 +52,7 @@ namespace alpaka
                         typename TSfinae = void>
                     class IteratorView
                     {
-                        using TViewDecayed = typename std::decay<TView>::type;
+                        using TViewDecayed = std::decay_t<TView>;
                         using Dim = alpaka::dim::Dim<TViewDecayed>;
                         using Idx = alpaka::idx::Idx<TViewDecayed>;
                         using Elem = typename MimicConst<alpaka::elem::Elem<TViewDecayed>, TView>::type;
```
BenjaminW3: here ;-)
and further down in this file.

sbastrakov: Thanks! Fixed and rebased.

BenjaminW3 (approved):

PR #918, opened 2020-01-21
Remove commented out code

BenjaminW3:
Those code pieces were never functional so nobody will miss them.

### DIFF
--- a/include/alpaka/kernel/TaskKernelGpuCudaRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuCudaRt.hpp
@@ -139,20 +139,6 @@ namespace alpaka
             public workdiv::WorkDivMembers<TDim, TIdx>
         {
         public:
-// gcc-4.9 libstdc++ does not support std::is_trivially_copyable.
-// MSVC std::is_trivially_copyable seems to be buggy (last tested at 15.7).
-// libc++ in combination with CUDA does not seem to work.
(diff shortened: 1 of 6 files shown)


include/alpaka/kernel/TaskKernelGpuCudaRt.hpp
```diff
@@ -139,20 +139,6 @@ namespace alpaka
             public workdiv::WorkDivMembers<TDim, TIdx>
         {
         public:
-// gcc-4.9 libstdc++ does not support std::is_trivially_copyable.
-// MSVC std::is_trivially_copyable seems to be buggy (last tested at 15.7).
-// libc++ in combination with CUDA does not seem to work.
-#if (!BOOST_COMP_MSVC) && !(defined(__GLIBCXX__) && (__GLIBCXX__)) && !(defined(_LIBCPP_VERSION) && BOOST_LANG_CUDA)
-            static_assert(
```
sbastrakov, line 146: Is this and the other `static_assert` in the `#if` guards also never invoked?

BenjaminW3: Yes, they are never invoked as they do not work at the moment. 
There is another PR #708 where I wanted to fix them but this might not be possible.

sbastrakov (approved):

PR #920, opened 2020-01-30
Add comment to pitch function at the bufferCopy example

SimeonEhrig:
The comment explained why it could be problematic to calculate the pitch itself. Feel free to improve it ;-)

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -283,6 +283,10 @@ auto main()
     alpaka::mem::view::copy(devQueue, deviceBuffer1, hostViewPlainPtr, extents);
     alpaka::mem::view::copy(devQueue, deviceBuffer2, hostBuffer, extents);
 
+    // Depending on the accelerator, the allocation function may introduce
+    // padding between rows/planes of multidimensional memory allocations.
+    // Therefore the pitch (distance between consecutive rows/planes) may be
+    // greater than the space required for the data.
     Idx const deviceBuffer1Pitch(alpaka::mem::view::getPitchBytes<2u>(deviceBuffer1) / sizeof(Data));
     Idx const deviceBuffer2Pitch(alpaka::mem::view::getPitchBytes<2u>(deviceBuffer2) / sizeof(Data));
     Idx const hostBuffer1Pitch(alpaka::mem::view::getPitchBytes<2u>(hostBuffer) / sizeof(Data));


example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -283,6 +283,10 @@ auto main()
     alpaka::mem::view::copy(devQueue, deviceBuffer1, hostViewPlainPtr, extents);
     alpaka::mem::view::copy(devQueue, deviceBuffer2, hostBuffer, extents);
 
+    // Depending on the accelerator, the allocation function may use memory
+    // alignment.
+    // This makes it possible for the pitch to be greater than the memory
+    // which is required for the data.
```
sbastrakov: On the language level, I think it may be formulated more directly, smth like "It allows the pitch to exceed the data size". However, on the logical level I do not get how is this a consequence of the first sentence.

BenjaminW3: Maybe better:
```
Depending on the accelerator, the allocation function may introduce padding between rows/planes of multidimensional memory allocations.
Therefore the pitch (distance between consecutive rows/planes) may be greater than the space required for the data.
```

SimeonEhrig: This is a good explanation. I was not sure, if pitch is a common term.

sbastrakov (approved):

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -283,6 +283,10 @@ auto main()
     alpaka::mem::view::copy(devQueue, deviceBuffer1, hostViewPlainPtr, extents);
     alpaka::mem::view::copy(devQueue, deviceBuffer2, hostBuffer, extents);
 
+    // Depending on the accelerator, the allocation function may introduce
+    // padding between rows/planes of multidimensional memory allocations.
+    // Therefore the pitch (distance between consecutive rows/planes) may be
```
ax3l, line 288: This might be slightly confusing for readers now, since the pitch is always greater or equal the data extent. Also, both padding and pitch are introduced as terms, but not brought into correlation to each other. Also, the "why" is not yet clear for new users.

A better wording might be:

Depending on the accelerator, the Alpaka allocation function for multidimensional buffers may introduce padding between rows/planes to avoid bank conflicts when accessing this data later on.
The data size plus this padding is called "pitch" and is equivalent to the distance between consecutive rows/planes.
Consequently, pitched data with non-zero padding occupies (slightly) more memory than non-pitched data.

PR #923, opened 2020-02-08
Fix exception in TaskKernelCpuOmp2Blocks when OMP_NUM_THREADS==1

### LINKED ISSUE #922: Omp2Blocks does throw runtime error with `OMP_NUM_THREADS=1`
Since #733, it is not possible anymore to run the `Omp2Blocks` backend with `OMP_NUM_THREADS=1`.

BenjaminW3:
Fixes #922

When `OMP_NUM_THREADS` is set to 1, the OpenMP runtime might not create a parallel region so we can not expect one.

I will cherry-pick this to the 0.4 release branch after this is merged.


bertwesarg:
Fine with me.

.travis.yml
```diff
@@ -167,7 +167,7 @@ matrix:
     - name: gcc-6 Debug c++17
       env: ALPAKA_CI_DOCKER_BASE_IMAGE_NAME=ubuntu:18.04 CXX=g++     CC=gcc   ALPAKA_CI_GCC_VER=6       CMAKE_BUILD_TYPE=Debug   ALPAKA_CI_STDLIB=libstdc++ ALPAKA_CI_BOOST_BRANCH=boost-1.70.0 ALPAKA_CI_CMAKE_VER=3.14.7 OMP_NUM_THREADS=2 ALPAKA_CXX_STANDARD=17
     - name: gcc-7 Release
-      env: ALPAKA_CI_DOCKER_BASE_IMAGE_NAME=ubuntu:16.04 CXX=g++     CC=gcc   ALPAKA_CI_GCC_VER=7       CMAKE_BUILD_TYPE=Release ALPAKA_CI_STDLIB=libstdc++ ALPAKA_CI_BOOST_BRANCH=boost-1.65.1 ALPAKA_CI_CMAKE_VER=3.13.5 OMP_NUM_THREADS=3
+      env: ALPAKA_CI_DOCKER_BASE_IMAGE_NAME=ubuntu:16.04 CXX=g++     CC=gcc   ALPAKA_CI_GCC_VER=7       CMAKE_BUILD_TYPE=Release ALPAKA_CI_STDLIB=libstdc++ ALPAKA_CI_BOOST_BRANCH=boost-1.65.1 ALPAKA_CI_CMAKE_VER=3.13.5 OMP_NUM_THREADS=1
```
sbastrakov, line 170: Looks to me as if before we only did it only for > 1 threads and now only for exactly 1. Shouldn't both cases be used?

psychocoderHPC (approved):

.travis.yml
BenjaminW3: I randomly selected a few jobs that use `OMP_NUM_THREADS=1` now. Before we only had jobs with 2+. We still have those so now we test 1 to 4 depending on the job.

PR #921, opened 2020-02-05
Installation via CMake

j-stephan:
This PR is a WIP to make alpaka installable to the filesystem. It installs the headers to `CMAKE_INSTALL_PREFIX/include/alpaka` and the required CMake files for `find_package(alpaka)` to `CMAKE_INSTALL_PREFIX/share/alpaka`. I didn't test what happens on Windows yet.

As this required some refactoring I also included the following changes:

* Bump minimum CMake version to 3.15
* All CMake commands are now lowercase
* Replaced some deprecated CMake calls with their non-deprecated counterparts
* Updated the FindTBB module. This enables linking to the `tbb::tbb` target.
* OpenMP is now linked as a target (`OpenMP::OpenMP_CXX`) instead of specifying headers, libraries and flags on our own.
* Removed a lot of string operations in favour of `target_link_libraries(alpaka INTERFACE xxx)`. If the compile flags need to be queried this can be done with CMake's `get_property()` command.

This successfully compiles for a non-modified vectorAdd example with the following CMakeLists.txt:

```cmake
cmake_minimum_required(VERSION 3.15)

project(bla)

find_package(alpaka)

add_executable(vectorAdd vectorAdd.cpp)
target_link_libraries(vectorAdd alpaka::alpaka)
```

There is still some cleanup and testing required. We also have some checks for clang < 4.0 which we should probably remove since we don't support it anyway.

PR #925, opened 2020-02-11
Update CI to more recent CMake versions

j-stephan:
This is a preparatory PR for #921. Separate PR because #921 is quite large already. This PR removes all CMake versions < 3.15 from `.travis.yml` and updates CMake 3.15 and 3.16 to their most recent patch levels (3.15.7 and 3.16.4, respectively).


BenjaminW3 (dismissed):

BenjaminW3:
You could also increase the `cmake_minimum_required` version in this PR but there is no need to do so.

ax3l:
In order to avoid inconsistencies:
Ideally increase all `cmake_minimum_required` to the lowest tested number, and also update the docs in `README.md` with the new minimal version (3.15.7, I guess?).

Later on, the [spack recipe here](https://github.com/ComputationalRadiationPhysics/spack-repo/blob/gh-pages/packages/alpaka/package.py) also needs its minimal versions updated :)

j-stephan:
Alright, all CMake files have been updated to a CMake 3.15.7+ requirement. Adapted the README.md as well.

BenjaminW3 (dismissed):

psychocoderHPC (dismissed):

ax3l (dismissed):

ax3l:
Do we have to clean the travis cache to get rid of some older `3.15.5` CMake versions?

BenjaminW3:
This should not be necessary.
The cache is automatically reset for a specific job when something within the job description changed (e.g. env variables) . Problematic is only when we do changes to the build/install scripts which are not detected.

BenjaminW3:
The current build issue is that we do not install CMake for Mac at all. On Mac we use the version that is already installed which seems to be 3.15.5.
The easiest fix would be to set this version as minimum in the CMake scripts.

BenjaminW3 (dismissed):

j-stephan:
Since there is another Mac node complaining about the CMake version (it wants 3.15.3 now) I'm considering to just set `cmake_minimum_required` to 3.15 and be done with it. The patch level shouldn't matter much anyway.

BenjaminW3:
Sounds reasonable

PR #928, opened 2020-02-12
Backend unification of the gpu backend

jkrude:
Merged all *`CudaBuiltIn` and *`HipBuiltIn` etc. into one `UniformedCudaHipBuiltIn` file.
A `AccGpuCudaRt` and an `AccGpuHipRt` still exist. They inherit from `AccGpuUniformedCudaHipRt` and are there for backwards compatibility as well as future changes.

PR #940, opened 2020-03-02
unify CUDA/HIP function calls/types

psychocoderHPC:
This PR is a follow up of #928 to remove most code duplication between CUDA/HIP.

- add macro `ALPAKA_API_PREFIX()` to refix functions and types
- add helper `ALPAKA_PP_CONCAT()` to construct `make_cudaPitched()`
- reduce duplicated runtime error checks for CUDA/HIP

I discussed with @sbastrakov if it is better to use a predefined list of function pointer/function objects or a macro to prefix the functions and types. Using a macro was at the end prepared by us to avoid that we need to maintain a global list of functions and types we are using.
It is also not clear about runtime overhead or compatibility when we use function objects. Function objects avoid inlining and the side effect to separable compile for CUDA and HIP is not clear.

- [x] merge after #941, the first commit contains #941

### DIFF
--- a/include/alpaka/core/Cuda.hpp
+++ b/include/alpaka/core/Cuda.hpp
@@ -34,10 +34,8 @@
 #include <cuda_runtime.h>
 #include <cuda.h>
 
-#include <array>
 #include <type_traits>
 #include <utility>
-#include <iostream>
 #include <string>
 #include <stdexcept>
 #include <cstddef>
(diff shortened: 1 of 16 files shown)


include/alpaka/core/Cuda.hpp
```diff
@@ -50,151 +50,10 @@
...
-
-//-----------------------------------------------------------------------------
-//! CUDA driver error checking with log and exception.
-#define ALPAKA_CUDA_DRV_CHECK(cmd)\
-    ::alpaka::cuda::detail::cudaDrvCheck(cmd, #cmd, __FILE__, __LINE__)
-
+#define ALPAKA_PP_CONCAT_DO(X,Y) X##Y
+#define ALPAKA_PP_CONCAT(X,Y) ALPAKA_PP_CONCAT_DO(X,Y)
+//! prefix a name with `hip`
+#define ALPAKA_API_PREFIX(name) ALPAKA_PP_CONCAT_DO(cuda,name)
```
sbastrakov: Comment seems wrong, should be `cuda`

include/alpaka/core/Cuda.hpp
```diff
@@ -50,151 +50,10 @@
...
-    }
-}
-
-//-----------------------------------------------------------------------------
-//! CUDA driver error checking with log and exception.
-#define ALPAKA_CUDA_DRV_CHECK(cmd)\
-    ::alpaka::cuda::detail::cudaDrvCheck(cmd, #cmd, __FILE__, __LINE__)
-
+#define ALPAKA_PP_CONCAT_DO(X,Y) X##Y
+#define ALPAKA_PP_CONCAT(X,Y) ALPAKA_PP_CONCAT_DO(X,Y)
```
sbastrakov, line 52: These two are general, maybe move to some alpaka preprocessor file (not sure if it exists), also to avoid duplication with HIP.

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -0,0 +1,133 @@
...
+        ::alpaka::uniform_cuda_hip::detail::rtCheckIgnore(cmd, #cmd, __FILE__, __LINE__, __VA_ARGS__)
+#else
+    #if BOOST_COMP_CLANG
+        #pragma clang diagnostic push
+        #pragma clang diagnostic ignored "-Wgnu-zero-variadic-macro-arguments"
+    #endif
+    //-----------------------------------------------------------------------------
+    //! CUDA runtime error checking with log and exception, ignoring specific error values
+    #define ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE(cmd, ...)\
+        ::alpaka::uniform_cuda_hip::detail::rtCheckLastError("'" #cmd "' A previous CUDA call (not this one) set the error ", __FILE__, __LINE__);\
```
sbastrakov: `A previous CUDA call` should probably be `A previous HIP call` ?

cmake/alpakaCommon.cmake
```diff
@@ -495,7 +495,7 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
 
                 set(CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
 
-                if((${CMAKE_BUILD_TYPE} STREQUAL "Debug") OR (${CMAKE_BUILD_TYPE} STREQUAL "RelWithDebInfo"))
+                if((CMAKE_BUILD_TYPE STREQUAL "Debug") OR (CMAKE_BUILD_TYPE STREQUAL "RelWithDebInfo"))
```
psychocoderHPC: This is part of #941 and is only for compatibility in this PR.

include/alpaka/mem/view/ViewPlainPtr.hpp
```diff
@@ -314,26 +314,21 @@ namespace alpaka
...
-                            cudaGetSymbolAddress(
-                                reinterpret_cast<void **>(&pMemAcc),
-                                *pMem));
-#else
-    #ifdef __HIP_PLATFORM_NVCC__
-                        ALPAKA_HIP_RT_CHECK(hipCUDAErrorTohipError(
-                            cudaGetSymbolAddress(
+
+#if BOOST_LANG_CUDA
+                        ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(hipCUDAErrorTohipError(
```
sbastrakov, line 325: Should `hipCUDAErrorTohipError` be here?

include/alpaka/pltf/PltfUniformCudaHipRt.hpp
```diff
@@ -155,77 +149,48 @@ namespace alpaka
...
 #if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
                 //-----------------------------------------------------------------------------
                 //! Prints all the device properties to std::cout.
                 ALPAKA_FN_HOST static auto printDeviceProperties(
 #if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
-                    cudaDeviceProp const & devProp)
+                    cudaDeviceProp const & devProp
 #else
-                    hipDeviceProp_t const & devProp)
+                    hipDeviceProp_t const & devProp
```
sbastrakov, line 191: So this type is actually called differently, not just a prefix change?

sbastrakov (changes requested):

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -0,0 +1,133 @@
...
+            using Error_t = hipError_t;
+#endif
+            //-----------------------------------------------------------------------------
+            //! CUDA runtime API error checking with log and exception, ignoring specific error values
+            ALPAKA_FN_HOST inline auto rtCheck(
+                Error_t const & error,
+                char const * desc,
+                char const * file,
+                int const & line)
+            -> void
```
sbastrakov, line 57: (Not important), can't we skip `-> void` now?

include/alpaka/core/Cuda.hpp
psychocoderHPC: I also thought about it but it is only needed here. @BenjaminW3 do you have a suggestion if we should stay with the two line duplication ore move it into an own file.

include/alpaka/pltf/PltfUniformCudaHipRt.hpp
psychocoderHPC: yes different type in CUDA and HIP

include/alpaka/core/Cuda.hpp
BenjaminW3: If it is only used here, we can keep it here for now.

include/alpaka/core/UniformCudaHip.hpp
BenjaminW3: yes we can, but no need to do now.

include/alpaka/core/UniformCudaHip.hpp
BenjaminW3: `A previous API call`?

sbastrakov: Ah, yes, it is unified, I somehow thought this piece was HIP-only.

sbastrakov (dismissed):
Great!

psychocoderHPC:
ready to merge

sbastrakov (approved):

sbastrakov:
@BenjaminW3 should we wait for your review of this PR? To me looks good to be merged.

BenjaminW3 (approved):

PR #942, opened 2020-03-03
[WIP] Add INSTALL_ONLY mode

j-stephan:
* Enable easier installation of header files
  * Change compatibility settings to `SameMinorVersion`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -38,6 +38,7 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 # Options and Variants
 
 option(alpaka_BUILD_EXAMPLES "Build the examples" ON)
+option(alpaka_INSTALL_ONLY "Install the headers without running CMake configuration for binary targets" OFF)
 
 include(CTest)
 # automatically defines: BUILD_TESTING, default is ON
@@ -72,7 +73,9 @@ set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${_ALPAKA_ROOT_DIR}/cmake/modules/")
 set(_ALPAKA_INCLUDE_DIRECTORY "${_ALPAKA_ROOT_DIR}/include")
 set(_ALPAKA_SUFFIXED_INCLUDE_DIR "${_ALPAKA_INCLUDE_DIRECTORY}/alpaka")
 
-include(${_ALPAKA_ROOT_DIR}/cmake/alpakaCommon.cmake)
+if(NOT alpaka_INSTALL_ONLY)
+    include(${_ALPAKA_ROOT_DIR}/cmake/alpakaCommon.cmake)
+endif()
 
 # Add all the source and include files in all recursive subdirectories and group them accordingly.
 append_recursive_files_add_to_src_group("${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "${_ALPAKA_SUFFIXED_INCLUDE_DIR}" "hpp" _ALPAKA_FILES_HEADER)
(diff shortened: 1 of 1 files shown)


j-stephan:
I mainly need this for spack. In the current version the CMake script checks for system CUDA (and throws an error if I forgot to set GCC to something older than 9) even if I only want to install the header files and the CMake modules.

This PR introduces the new variable `alpaka_INSTALL_ONLY` which is set to `OFF` by default. If enabled it will skip all checks required to build a target with alpaka and just enables the installation part instead.

psychocoderHPC:
I think the common workflow in other projects is the other way around. You install normally your project by default. If you like to install and build test you need to enable it.
I suggest to change the logic so that by default alpaka is installed only and optional you can activate building with tests.

psychocoderHPC:
I checked the installation of CMake it self and `BUILD_TESTING=ON` is the default.

Never the less you can disable tesing via `-DBUILD_TESTING=OFF` instead if introducing a new option.

psychocoderHPC:
offline discussed with @j-stephan: The installation of alpaka must be able without any non default CMake flags. There was a confusion between the CMakeLists.txt and the CMake file generated from `alpakaConfig.cmake.in` during the installation or usage of `add_subdirectory`

`cmake -DBUILD_TESTING=OFF alpaka_source_code` should install alpaka without analyzing alpaka dependencies.
`cmake alpaka_source_code` installs alpaka with and analyze the backend depedancy (only for the test cases)

psychocoderHPC:
After a second discussion we realized that our alpaka workflows not based on the usage instead they reflect our CI/testing behavior.
Alpaka is by default activating all back-ends those are available based on the installed dependencies. This is useful for the tests but the common user workflow and also how our examples looks like is different.
An example is explicit saying use accelerator XY and not give me a list of all accelerators.

So my suggestion is to disable all backend by default and let the user choose the right backend. Build a target `alpaka::auto` where alpaka is choosing all available possible backends e.g. for our testing/CI workflows.

We need also to take care that we can build typical HPC workflows: e.g. alpaka is installed with spack
`spack install alpaka +openmp` in that case a CUDA backend should not be available from the installed alpaka code. The reason is that spack is not able to create module dependencies if we express the dependency during the installation.
`spack install alpaka +cuda +openmp` should give the user the possibility to select on or both accelerators.

Short: my point is that we should try to mimic default workflows how to use software and that alpaka can be used in a HPC environment where e.g. the admin is installing alpaka and is running the test and is responsible to provide all dependancies needed for the current installed flavor of alpaka.

CC-ing: @ax3l  @BenjaminW3  @sbastrakov

BenjaminW3:
Can this be closed now after #1016?

psychocoderHPC:
I will close this PR and pulled missing part into the issue #1019

ax3l:
@j-stephan I think my pr in https://github.com/spack/spack/pull/16869 might have fixed this problem, too?

> I mainly need this for spack. In the current version the CMake script checks for system CUDA (and throws an error if I forgot to set GCC to something older than 9) even if I only want to install the header files and the CMake modules.

Otherwise please elaborate a bit more, cannot follow.

j-stephan:
> I think my pr in spack/spack#16869 might have fixed this problem, too?

I think you introduced a new environment variable for the host compiler to spack, right? That is good news, but not what I was aiming for here.

I was interested in disabling the backend checks altogether (by default) since they are not needed for an alpaka installation. I still haven't achieved this completely since other work got in the way. Parts of this PR have been superseded by #1016, the missing parts should be in #1019. I will have a look at it now that we are slowly transitioning to CMake 3.18 and first-class CUDA support.

PR #945, opened 2020-03-06
drop HIP-hcc support

psychocoderHPC:
HIP-hcc is depricated and we are handling a few workarounds which
increases the amount of maintenaince work.

This pull request is removing any HIP-hcc support.

PR #864, opened 2019-10-22
[WIP] Find hiprand/rocrand by ROCm provided config files.

tdd11235813:
WIP because: nvcc backend: linker complains about .so file that has version suffix (`/opt/rocm/rocrand/hiprand/lib/libhiprand.so.1.8.2`, gcc5.5.0, nvcc 10.1.168) (it works w/o suffix)
- there is also an .so link, either the finder could prefer this or there is a linker flag to deal with this?
- in the previous version the custom finder preferred the .so suffix, thus the error did not occur
- hip(hcc/clang) works (so should hip-clang), 

[...]

PR #948, opened 2020-03-06
fix: missing hipRand and rocRand library

psychocoderHPC:
replacement PR for #864

- search for rocrand and hiprand

This bug was not found because `CPLUS_INCLUDE_PATH` was used in my environment.
Please only review the last commit, the first commit is #945

- [x] rebase against #945 
- [x] rebase against #965

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -679,7 +679,29 @@ if(ALPAKA_ACC_GPU_HIP_ENABLE)
                 endif()
                 target_include_directories(alpaka INTERFACE ${HIP_RAND_INC})
                 target_link_libraries(alpaka INTERFACE ${HIP_RAND_LIBRARY})
-            endif() # nvcc
+            elseif(ALPAKA_HIP_PLATFORM MATCHES "clang")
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
in all older versions of alpaka I never linked in `libhiprand` and `librocrand`. This PR is doing it and I got issues at runtime and somehow always the libs from `/opt/rocm/hiprand/` will be used instead of my self compiled version in my home directory.

psychocoderHPC:
@BenjaminW3  could you please have a look to https://github.com/ComputationalRadiationPhysics/alpaka/pull/948/checks?check_run_id=533853918#step:3:3277
I am wondering why it is crashing and worked before. It looks like `-G` is not supported by `nvcc` in windows

BenjaminW3:
We have exactly the same error in one build of #959. We did not change anything so either CUDA has changed or something is flaky.

psychocoderHPC:
> We have exactly the same error in one build of #959. We did not change anything so either CUDA has changed or something is flaky.

I think we need to disable `-G` for windows for `nvcc`. ~~I will open an PR.~~

[update]: I realized that this is only a warning and not the reason for the error in the CI.

BenjaminW3 (approved):

PR #949, opened 2020-03-07
Add support for Visual Studio 2019

BenjaminW3:
This makes use of Github Actions for building on Windows with Visual Studio 2019.
We already use GitHub Actions for updating the doxygen documentation and it looks like the next reasonable step. Especially because it allows for 20 parallel builds for open source projects whereas Travis only has 5.


BenjaminW3:
Ready for review/merge!

BenjaminW3:
ready for review/merge ;-)

psychocoderHPC:
Is this a replacement for travis tests? or an addition for the windows tests?

Only a note: we are working on getting the HZDR CI into alpaka to support scalable tests on multiple architectures (Linux, maybe MACOSX). Windows will stay on travis.

psychocoderHPC (approved):

BenjaminW3:
Currently this is both. VS 2019 tests are new but now directly within github. My goal is to move all the tests from travis to github. This would remove the necessity for yet another account.

PR #954, opened 2020-03-12
allow instances of concept definitions

psychocoderHPC:
Define all concepts as empty implementation to allow instance creation.

This will allow to implement more user friendly ways (optional) to handle concepts.

```
auto conceptImpl = getConcept( ConceptIdxBt{} );
// instead of
auto conceptImpl = getConcept< ConceptIdxBt >();
```

I like to reduce the explicit template usage in `cupla` as much as possible.

### DIFF
--- a/include/alpaka/acc/Traits.hpp
+++ b/include/alpaka/acc/Traits.hpp
@@ -29,9 +29,9 @@ namespace alpaka
     //! The accelerator specifics.
     namespace acc
     {
-        struct ConceptUniformCudaHip;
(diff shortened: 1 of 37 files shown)


sbastrakov (approved):

sbastrakov:
Discussed offline with @psychocoderHPC : I also like the syntax with template parameters derived from arguments more than the explicitly specializing the template types. I felt a little weird that in this case, it concerns concepts, as it is not clear what does it mean to instantiate a concept. However, still think it makes sense to add.

Issue #135, opened 2016-01-13
out of memory access


psychocoderHPC:
The method [pitchedOffsetBytesDim](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/view/ViewSubView.hpp#L398) of `ViewSubView` results in a out of memory access if `Tidx` is equal than the `Tdim-1` of the view.

The issue is triggered by `Tidx + 1` in `mem::view::getPitchBytes<Tidx + 1u>(view);`

psychocoderHPC:
Maybe we can remove this function because is is not used.

@BenjaminW3 Could you shortly explain what this method should do?

BenjaminW3:
First of all [this line](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/view/ViewSubView.hpp#L353) is wrong, there is one `view` argument too much.

Why should there be a out of memory access? `mem::view::getPitchBytes<TDim + 1u>(view);` should always return `1` for all types.

psychocoderHPC:
If someonecalled [pitchedOffsetBytesDim](https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/mem/view/ViewSubView.hpp#L392) with `Tidx = 1` on a 3 dimensional view than `Tidx + 1` = `2` is used to call `mem::view::getPitchBytes<2>()`. 
I think this is currently not the case bug it can be used wrong.

BenjaminW3:
I was able to reproduce a memory access error. I will provide a fix and a unit test veryfying it within the next days.

psychocoderHPC:
Big thanks. @erikzenker and me will look to the other issues.

BenjaminW3:
The only open issue is #138 because #135, #136,#137 and #141 are all solved by my test.

BenjaminW3:
`pitchedOffsetBytesDim` is a private class method which is called with the correct indices since #143.

PR #959, opened 2020-03-20
Turn QueueCPU* into template QueueGeneric*<TDev>

jkelling:
This patch turns QueueCpuBlocking and QueueCpuNonBlocking into template with the Device type as parameter.

### Motivation
This will allow reusing the blocking and non-blocking CPU queue implementations for new backend, which do not require their own queue implementation but which do need a specific device class to abstract target devices. At the moment candidates are the [OpenMP 5 target-offload backend](https://github.com/jkelling/alpaka) and OpenACC.

### Main Changes
* add templates QueueGenericBlocking, QueueGenericNonBlocking, EventGeneric, IGenericQueue (using the code of their CPU predecessors, except with TDev instead of DevCpu)
* add aliases QueueCpuBlocking, QueueCpuNonBlocking, EventCpu, ICpuQueue, maintaning the old file structure, to limit the impact on other part of the code using the old classes
* replace forward declarations of EventCpu in some tests with includes as they would collide with the alias
* add new traits RegisterQueue and GetAllQueues for DevCpu


include/alpaka/queue/IGenericQueue.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/BoostPredef.hpp>
+
+// #include <alpaka/event/EventGeneric.hpp>
+// #include <alpaka/dev/DevCpu.hpp>
```
BenjaminW3: Remove when not needed?

include/alpaka/queue/IGenericQueue.hpp
```diff
@@ -0,0 +1,57 @@
...
+    // avoid diagnostic warning: "has no out-of-line virtual method definitions; its vtable will be emitted in every translation unit [-Werror,-Wweak-vtables]"
+    // https://stackoverflow.com/a/29288300
+    #pragma clang diagnostic push
+    #pragma clang diagnostic ignored "-Wweak-vtables"
+#endif
+
+        //#############################################################################
+        //! The CPU queue interface
+        template<
+            typename TDev>
```
BenjaminW3: Could we make `IGenericQueue` depend on `TEvent` instead of `TDev`? The queue does not need to know its device at all but only the Event. `event::EventGeneric<TDev>` can be provided at the place the template is instantiated.

include/alpaka/queue/IGenericQueue.hpp
```diff
@@ -0,0 +1,57 @@
...
+        class IGenericQueue
```
BenjaminW3: I really do not like the `Generic` names for the queues and the event here. They are not that generic at all. The `Cpu` part of the name previously had two meanings:
* the queue holds `EventCpu`
* the queue is executed on the CPU

You made the first point generic but what it still differentiates from a generic queue or the CUDA queues is that it can only be run on the host CPU.
I would prefer if this would be called `IGenericCpuQueue` and still be located in the `cpu` subfolder. The same is true for the queue impls and the events.

include/alpaka/queue/QueueGenericNonBlocking.hpp
```diff
@@ -0,0 +1,244 @@
...
+                -> bool
+                {
+                    return queue.m_spQueueImpl->m_workerThread.isIdle();
+                }
+            };
+        }
+    }
+}
+
+#include <alpaka/event/EventGeneric.hpp>
```
BenjaminW3: This looks strange

include/alpaka/event/EventGeneric.hpp
```diff
@@ -0,0 +1,508 @@
...
+                                                                            //!< If m_enqueueCount == m_LastReadyEnqueueCount, the event is currently not enqueued
+                };
+            }
+        }
+
+        //#############################################################################
+        //! The CPU device event.
+        template<
+            typename TDev>
+        class EventGeneric final : public concepts::Implements<wait::ConceptCurrentThreadWaitFor, EventGeneric<TDev>>
```
BenjaminW3: As written in the other comment, I would really like to call this `EventCpuGeneric` and move it into a `cpu` subfolder.

include/alpaka/dev/Traits.hpp
```diff
@@ -145,5 +159,40 @@ namespace alpaka
...
+        //-----------------------------------------------------------------------------
+        //! \return all queues associated with device
+        //! What this method does is dependent on the accelerator.
+        template<
+            typename TDev>
+        ALPAKA_FN_HOST auto registerQueue(
+            TDev const & dev)
+#ifdef BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION
+        -> decltype(traits::GetAllQueues<TDev>::getAllQueues(dev))
+#endif
```
BenjaminW3: Please remove this ifdef as we are c++14 only now.

include/alpaka/dev/DevCpu.hpp
```diff
@@ -244,6 +244,43 @@ namespace alpaka
...
+            template<>
+            struct GetAllQueues<
+                dev::DevCpu>
+            {
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_HOST static auto getAllQueues(
+                    dev::DevCpu const & dev)
+#ifdef BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION
+                -> decltype(dev.m_spDevCpuImpl->GetAllQueues())
+#endif
```
BenjaminW3: please remove the ifdef

include/alpaka/dev/Traits.hpp
```diff
@@ -64,6 +64,20 @@ namespace alpaka
...
+                typename T,
+                typename TSfinae = void>
+            struct RegisterQueue;
+
+            //#############################################################################
+            //! The device get all queues trait.
+            template<
+                typename T,
+                typename TSfinae = void>
+            struct GetAllQueues;
```
BenjaminW3: Do we need those public traits at all?
The `ICpuQueue` interface with the `registerQueue` and `getAllQueues` methods is an implementation detail of the CPU queues (and therefore lives in the `cpu` subfolder) and should not be publicly visible even after your changes. The CUDA queues do not implement this and do not need this.
It should be possible to live without those traits because the `QueueCpuGenericXxx` gets its specific device implementation as template parameter and the device simply has to implement a `RegisterQueue` method without the need for any explicit public interface.

BenjaminW3 (changes requested):
Thank your for providing the first steps to add the new OpenMP 4 support!

include/alpaka/event/EventGeneric.hpp
psychocoderHPC: ~~The only problem I see with the name `EventCpuGeneric` is that the upcoming OpenMP5 and OpenAcc backends are not only for CPU.~~
[updated] The reason for the naming is [here](https://github.com/ComputationalRadiationPhysics/alpaka/pull/959/files#r395969995) better described

include/alpaka/event/EventGeneric.hpp
```diff
@@ -0,0 +1,508 @@
...
+            template<
+                typename TDev>
+            struct Enqueue<
+                queue::generic::detail::QueueGenericNonBlockingImpl<TDev>,
+                event::EventGeneric<TDev>>
+            {
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_HOST static auto enqueue(
+#if !(BOOST_COMP_CLANG_CUDA && BOOST_ARCH_PTX)
+                    queue::generic::detail::QueueGenericNonBlockingImpl<TDev> & queueImpl,
```
psychocoderHPC: please use `alpaka::ignore_unused(queueImpl)` in the function body.

```
#if BOOST_COMP_CLANG_CUDA && BOOST_ARCH_PTX
    alpaka::ignore_unused(queueImpl);
#endif
```

psychocoderHPC: Could be that this code is derived from the original implementation but we should change it directly in this PR.

include/alpaka/queue/QueueCpuBlocking.hpp
```diff
@@ -11,7 +11,7 @@
 
 #include <alpaka/core/Unused.hpp>
 #include <alpaka/dev/DevCpu.hpp>
-#include <alpaka/queue/cpu/ICpuQueue.hpp>
+#include <alpaka/queue/QueueGenericBlocking.hpp>
```
psychocoderHPC: all expect this include should be removed

include/alpaka/queue/QueueCpuNonBlocking.hpp
```diff
@@ -10,7 +10,7 @@
 #pragma once
 
 #include <alpaka/dev/DevCpu.hpp>
-#include <alpaka/queue/cpu/ICpuQueue.hpp>
+#include <alpaka/queue/QueueGenericNonBlocking.hpp>
```
psychocoderHPC: all expect this include should be removed.

include/alpaka/queue/QueueGenericNonBlocking.hpp
```diff
@@ -0,0 +1,244 @@
...
+                typename TDev,
+                typename TTask>
+            struct Enqueue<
+                queue::QueueGenericNonBlocking<TDev>,
+                TTask>
+            {
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_HOST static auto enqueue(
+#if !(BOOST_COMP_CLANG_CUDA && BOOST_ARCH_PTX)
+                    queue::QueueGenericNonBlocking<TDev> & queue,
```
psychocoderHPC: Please use `alpaka::ignore_unused()`

psychocoderHPC (changes requested):

include/alpaka/queue/IGenericQueue.hpp
jkelling: I did it this way to ensure the template argument will always be derived from EventGeneric and exclude arbitrary types. If we do not want to be this strict, we could use `template<typename TEvent>` here. What shall it be?

include/alpaka/queue/IGenericQueue.hpp
jkelling: OK, I thought the name was only related to the Dev / Event type.

Howver, the name `CPU` then is still too general in that it only says that it lives on the host. There are multiple ways to implement a device-wise generic queue. This on used C++11 threads, on different option would be OpenMP tasks. With this in mind, we could call them, e.g. `IGenericThreadsQueue`, `QueueThreadsBlocking`. The interface would still go into the cpu subfolder.

include/alpaka/queue/QueueCpuBlocking.hpp
jkelling: I cannot follow, can you please clarify?

This file defines an alias for `QueueGenericBlocking<DevCpu>` and thus needs the definition of that template.

Do you mean this file should be removed and thus the alias? This would entail lots of changes elsewhere, i.e. `QueueCpuBlocking` -> `QueueGenericBlocking<DevCpu>`

include/alpaka/event/EventGeneric.hpp
jkelling: indeed

include/alpaka/queue/IGenericQueue.hpp
BenjaminW3: I do not see a reason to be so strict. Maybe there will be another queue that uses a totally different event.

include/alpaka/queue/IGenericQueue.hpp
BenjaminW3: I like your proposed names

include/alpaka/queue/IGenericQueue.hpp
jkelling: But `IGenericThreadsQueue` is a polymorphic interface for `QueueThreadsGeneric*` which uses `EventGeneric<TDev>`.

Allowing `IGenericThreadsQueue` to be used with other queues than instances of `QueueThreadsGeneric*` would extent the original scope of `ICpuQueue`. If we extent the scope this far, `IGenericThreadsQueue` could also be used for something like the Cuda queue, and thus would be restricted to living on the host anymore.

jkelling: At least this is what I understood the intent behind ICpuQueue to be: a polymorphic interface for queues using EventCpu.

include/alpaka/queue/IGenericQueue.hpp
jkelling: I meant to write `QueueGenericThreadsBlocking` instead of `QueueThreadsBlocking`, Is that ok, too?

include/alpaka/queue/QueueGenericNonBlocking.hpp
jkelling: This is just copied from the implementation of QueueCpuNonBlocking. I guess someone wanted to have the queue header also provide the event definition. The Event header cannot included at the top though, because it requires the definition of the queue for the trait specializations. I did not try removing it, assuming it may break other code.

include/alpaka/dev/Traits.hpp
jkelling: So, it would be better to require TDev to implement some concept `QueueRegister`. I agree with skipping the traits.

jkelling: Just one issue: DevCpu does not have `GetAllQueues()` and `RegisterQueue()`, they are only present in DevCpuImpl. In introduced the traits because it looked a it strange when the queue would just reach into the internal implementation of DevCpu. So how should this be solved?
1. Just access `dev.m_spDevOmp4Impl->RegisterQueue(queue.m_spQueueImpl)`.
2. Move the member functions to DevCpu, as they are part of its interface.

BenjaminW3: both sound reasonable so do what you think is easier. For the first solution you might get access issues and the members have to have the same name so the second might be easier.

include/alpaka/queue/QueueGenericNonBlocking.hpp
BenjaminW3: I am fine with this

include/alpaka/queue/IGenericQueue.hpp
BenjaminW3: Also fine

include/alpaka/dev/Traits.hpp
jkelling: I will do option 2 then, because it is cleaner, though  more work than 1.

include/alpaka/dev/DevCpu.hpp
```diff
@@ -68,16 +68,30 @@ namespace alpaka
                 {
                 private:
 
+                public:
```
BenjaminW3, line 69: Could you please delete the `private:` section above?

include/alpaka/queue/IGenericQueue.hpp
BenjaminW3: I am also ok with this. My point was that there might be another type of event that can be used with the `IGenericThreadsQueue` that is still compatible with the current CPU queue and device. but is not necessarily a specialization of `EventGeneric`.

include/alpaka/queue/QueueCpuBlocking.hpp
BenjaminW3: I think @psychocoderHPC  meant that you can remove nearly all other includes in this file because you only create a typedef for a speialization of `QueueGenericBlocking`. So you only need its header and the header of the template parameter.
This is also true for the other typedef-only headers.

include/alpaka/queue/IGenericQueue.hpp
jkelling: There can be no event which is not a specialization of `EventGenericThreads`, that can also be used with `QueueGenericThreats*` and by extension `QueueCpu*`, because the queue explictly uses `EventGenericThreads<TDev>` and also overrides `enqueue()` and `wait()` using this argument type. So if `IGenericThreadsQueue` were not restricted one of the following would need to change:
* formally allow `IGenericThreadsQueue` to be a base for different queue types
* add another template parameter `TEvent` to `QueueGeneric`, one could still enforce that `TEvent` and `TDev` are connected by saying
  ```c++
  template < typename TDev, template<typename TTDev> TEvent = EventGenericThreads >
  class QueueGenericThreats...
  ```
At the moment I would prefer to leave it strict. There is the option of relaxing it should the need arise, as `IGenericThreadsQueue` is an implementation detail from the perspective of user code.

include/alpaka/queue/QueueCpuBlocking.hpp
jkelling: That does make a lot of sense, thanks. Done.

BenjaminW3 (dismissed):

BenjaminW3:
```
/home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/event/EventCpu.hpp(18): error #135: namespace "alpaka::dev" has no member "DevCpu"
```

jkelling:
> ```
> /home/travis/build/ComputationalRadiationPhysics/alpaka/include/alpaka/event/EventCpu.hpp(18): error #135: namespace "alpaka::dev" has no member "DevCpu"
> ```

So this is one of the two failing builds? I can never retrieve those build logs.
What is the cmake configuration in this case? What are the final compile flags? Is this running with `VERBOSE=1`?

jkelling:
I, see, I deleted one include too many and did not see the compiler error afterwards (or it build because of cached files somehow).

BenjaminW3:
`ALPAKA_DEBUG=2`

jkelling:
There is still one windows check failing and I cannot download the log. Any hints?

BenjaminW3:
```
CMake Error at C:/Program Files/CMake/share/cmake-3.16/Modules/FindCUDA.cmake:1602 (configure_file):
configure_file Problem configuring file
```
I am pretty sure that this is not your fault but a temporary issue in CI.

psychocoderHPC:
@jkelling  could you please rebase this PR against the last development branch, this will fix the failing tests.

jkelling:
@psychocoderHPC  or @BenjaminW3  can you pleas review the latest commit I added?
https://github.com/ComputationalRadiationPhysics/alpaka/pull/959/commits/449269e2ba95928bf61a1cef954053483524156c

I found this change when rebasing my omp4 branch onto this one. Thematically belongs to this PR.
This is a change which might need to be discussed:
The template function`wait::traits::detail:currentThreadWaitForDevice` is made less general in this commit and moved into EventGenericThreads.hpp and is now `wait::traits::generic::currentThreadWaitForDevice`
1. The reason for making it less generic, is that this way the speicalizations of `CurrentThreadWaitFor<Dev>` are shorter (one line)
2. It was not used outside of CurrentThreadWaitFor<DevCpu>
3. Having something in a namespace detail generic enough to be used in any context is contradictory.
4. point 2 and 3 are in line the policy discussed here: https://github.com/ComputationalRadiationPhysics/alpaka/pull/959#discussion_r395971248

include/alpaka/event/EventGenericThreads.hpp
```diff
@@ -316,6 +317,33 @@ namespace alpaka
     {
         namespace traits
         {
+            namespace generic
+            {
+                template<typename TDev>
+                ALPAKA_FN_HOST auto currentThreadWaitForDevice(
+                    TDev const & dev
+                )
+                ->void
```
psychocoderHPC, line 326: IMO for void we should not use trailing return values. Since we depend on C++14 we can skip the type if we use auto but in a void method/function it is a little bit risky if someone is adding a return.

```C++
                template<typename TDev>
￼                ALPAKA_FN_HOST void currentThreadWaitForDevice(
￼                    TDev const & dev
￼                )
```
@BenjaminW3 What do you think?

psychocoderHPC:
> @psychocoderHPC or @BenjaminW3 can you pleas review the latest commit I added?
> [449269e](https://github.com/ComputationalRadiationPhysics/alpaka/commit/449269e2ba95928bf61a1cef954053483524156c)
> 
> I found this change when rebasing my omp4 branch onto this one. Thematically belongs to this PR.
> This is a change which might need to be discussed:
> The template function`wait::traits::detail:currentThreadWaitForDevice` is made less general in this commit and moved into EventGenericThreads.hpp and is now `wait::traits::generic::currentThreadWaitForDevice`
> 
> 1. The reason for making it less generic, is that this way the speicalizations of `CurrentThreadWaitFor<Dev>` are shorter (one line)
> 2. It was not used outside of CurrentThreadWaitFor
> 3. Having something in a namespace detail generic enough to be used in any context is contradictory.
> 4. point 2 and 3 are in line the policy discussed here: [#959 (comment)](https://github.com/ComputationalRadiationPhysics/alpaka/pull/959#discussion_r395971248)

IMO this change is ok, the method is not exposed to the user.

include/alpaka/event/EventGenericThreads.hpp
jkelling: I agree, that the trailing void looks a bit silly, but it was like this in the original code.

Would be happy to change it, but this is not the only place with a trialing void, e.g. https://github.com/jkelling/alpaka/blob/queuegeneric/include/alpaka/event/EventGenericThreads.hpp#L72 , https://github.com/jkelling/alpaka/blob/queuegeneric/include/alpaka/event/EventGenericThreads.hpp#L267 , https://github.com/jkelling/alpaka/blob/queuegeneric/include/alpaka/event/EventGenericThreads.hpp#L307 , https://github.com/jkelling/alpaka/blob/queuegeneric/include/alpaka/event/EventGenericThreads.hpp#L326 ...

Should all be changed like this?

jkelling: There are also trailing -> bool . What about those?

BenjaminW3: I would keep it. We use trailing return types for everything. I prefer consistency.

BenjaminW3 (approved):

include/alpaka/event/EventGenericThreads.hpp
jkelling: ... code cleanup might be better as a separate PR, IMHO.

jkelling: My page was not updated to show @BenjaminW3's comment when I wrote mine. Marking this as resolved for this PR.

psychocoderHPC (approved):

PR #969, opened 2020-04-03
update github links to alpaka and cupla

psychocoderHPC:
Set links to the nex alpaka location.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -31,7 +31,7 @@ set(PACKAGE_VERSION "${ALPAKA_VERSION_MAJOR}.${ALPAKA_VERSION_MINOR}.${ALPAKA_VE
 
 project(alpaka VERSION      ${ALPAKA_VERSION_MAJOR}.${ALPAKA_VERSION_MINOR}.${ALPAKA_VERSION_PATCH}
                DESCRIPTION  "The alpaka library is a header-only C++14 abstraction library for accelerator development."
-               HOMEPAGE_URL "https://github.com/ComputationalRadiationPhysics/alpaka"
+               HOMEPAGE_URL "https://github.com/alpaka-group/alpaka"
                LANGUAGES    CXX)
 
 set_property(GLOBAL PROPERTY USE_FOLDERS ON)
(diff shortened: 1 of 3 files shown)


BenjaminW3 (dismissed):

psychocoderHPC:
@BenjaminW3  do you know why `continuous-integration/travis-ci ` is in the test list and not provide the Details button?

BenjaminW3:
I do not know where this is coming from. It looks like it is the old Travis integration point that was replaced by the new version some time ago.

psychocoderHPC:
Could be my fault I installed the travis app to the organisation. I removed it now

BenjaminW3:
I would prefer the travis app over the old thing. At the old place I think we had a link to the travis results and it looked like the app.

psychocoderHPC:
Yes, reactivated the travis app again. I have not configured the old way but could be I need to remove the webhook fully.

BenjaminW3:
Maybe it automatically re-enabled itself when the Repo was moved.

psychocoderHPC:
OK I need remove the check for the branch

psychocoderHPC:
gitissue

Somehow I miss the option `Travis CI - Pull Request` to say that this test must pass befor ethe PR can be merged. I have it in cupla but somehow not in alpaka.

psychocoderHPC:
OK I pushed this PR again and now it was also able to activate this option

psychocoderHPC:
note: I activated also all other CI test as required too.

psychocoderHPC:
@BenjaminW3 Do you ever restarted the actions because one of the tests failed?
I realized that it is only possible to restart all actions, not only the broken pipe. After restarting all tests will fail (see current output). Maybe you know better whats going wrong after restarting the tests.

I saw this error in the past also before we transferred the repo, so this will not be the root of the issue.

BenjaminW3:
Restarting individual jobs is not possible yet. Restarting all always failed for me. That was the reason why I did not require the github builds yet.

BenjaminW3 (approved):

BenjaminW3:
I disabled the github actions build as merge barrier for now. Maybe our build scripts are doing something wrong that makes the rebuilds fail.

psychocoderHPC:
> Restarting individual jobs is not possible yet. Restarting all always failed for me. That was the reason why I did not require the github builds yet.

Ahh ok, than we should remove the require for now.

PR #970, opened 2020-04-06
Remove unnecessary includes of boost/config.hpp

BenjaminW3:
We use the defines from `boost.predef` and not the ones from `boost.config`.

We do not need the variadic template workarounds for a long time now since we removed the dependency to boost.mpl long ago.

### DIFF
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -18,23 +18,6 @@
 #endif
 
 //-----------------------------------------------------------------------------
-// Boost does not yet correctly identify clang when compiling CUDA code.
-// After explicitly including <boost/config.hpp> we can safely undefine some of the wrong settings.
-#if BOOST_COMP_CLANG_CUDA
-    #include <boost/config.hpp>
-    #undef BOOST_NO_CXX11_VARIADIC_TEMPLATES
-#endif
-
-//-----------------------------------------------------------------------------
-// Boost disables variadic templates for nvcc (in some cases because it was buggy).
-// However, we rely on it being enabled.
-// After explicitly including <boost/config.hpp> we can safely undefine the wrong setting.
-#if BOOST_COMP_NVCC
-    #include <boost/config.hpp>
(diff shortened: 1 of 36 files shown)


include/alpaka/core/Common.hpp
```diff
@@ -17,23 +17,6 @@
...
-    #include <boost/config.hpp>
-    #undef BOOST_NO_CXX11_VARIADIC_TEMPLATES
-#endif
-
-//-----------------------------------------------------------------------------
-// Boost disables variadic templates for nvcc (in some cases because it was buggy).
-// However, we rely on it being enabled.
-// After explicitly including <boost/config.hpp> we can safely undefine the wrong setting.
-#if BOOST_COMP_NVCC
-    #include <boost/config.hpp>
```
ax3l, line 33: but we are using `BOOST_LANG_CUDA` and `BOOST_LANG_HIP` in line 44 (27), which is defined by this include.

ax3l: update: ah, I remember. we introduced `alpaka/core/BoostPredef.hpp` that contains those work-arounds and the include, right?

BenjaminW3: Yes, we use the defines from boost.predef and not the ones from boost.config

BenjaminW3: We do not need the variadic template workarounds for a long time now since we removed the dependency to boost.mpl long ago.

ax3l: hm, no that one does not include `boost/config.hpp`, but set the defines we need here via `boost/predef.h`.

But where do we set `BOOST_NO_CXX11_VARIADIC_TEMPLATES` now? Cannot find other occurrences in the code base.

update: see above, collided mid-air.
(PR description? ;-) )

ax3l: I added your details to the PR description, because that saves reviewers time.

ax3l (approved):

PR #973, opened 2020-04-08
Always use out-of-class version of vec::createVecFromIndexedFn

BenjaminW3:
We once needed a workaround but it seems to work correctly now.
I am not sure why exactly, but it could be related to the c++14
std::integer_sequence usage.


sbastrakov (approved):

psychocoderHPC (approved):

psychocoderHPC:
Tested with HIP-clang on fwk388 and it is working well.

PR #977, opened 2020-04-14
Simplify vector operator construction

BenjaminW3:
This was lying around on my hard drive for a long time.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -381,10 +381,11 @@ namespace alpaka
         namespace detail
         {
             //#############################################################################
-            //! A function object that returns the sum of the two input vectors elements.
+            //! This is used to create a Vec by applying a binary operation onto the corresponding elements of two input vectors.
             template<
+                template<typename> class TFnObj,
                 std::size_t Tidx>
-            struct CreateAdd
+            struct CreateVecByApplyingBinaryFnToTwoIndexedVecs
             {
                 //-----------------------------------------------------------------------------
                 ALPAKA_NO_HOST_ACC_WARNING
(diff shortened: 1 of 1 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -582,32 +526,16 @@ namespace alpaka
...
-                    typename TVal>
-                ALPAKA_FN_HOST_ACC static auto create(
-                    Vec<TDim, TVal> const & p,
-                    Vec<TDim, TVal> const & q)
-                -> bool
-                {
-                    return p[Tidx] >= q[Tidx];
-                }
-            };
+            using CreateVecFromTwoIndexedVecsGreaterEqual = CreateVecByApplyingBinaryFnToTwoIndexedVecs<std::greater_equal, Tidx>;
```
psychocoderHPC, line 538: I would say that `std::*` functors will not work here because they can not be used on all devices e.g. CUDA/HIP.

psychocoderHPC (changes requested):

include/alpaka/vec/Vec.hpp
BenjaminW3: They should be constexpr with c++14 which would allow us to use them in device code.

psychocoderHPC: But the operation (functor) evaluation is runtime not compile time.

BenjaminW3: As long as the functor is constexpr (which it is), there should be no problem to evaluate it at runtime.

psychocoderHPC: I am still confused. `constexpr` only means that the function can be evaluated at compile time with compile time values (input). This is here not the case.
Do we have a test where we create somewhere on the device a vector where the generator with a functor from `std::*` is used?

psychocoderHPC: Mhh each `+` operation on the device should use this generator. I am still confused how the compiler can generate device code e.g. for CUDA out of an host functor.

BenjaminW3: What I am relying on here is `--expt-relaxed-constexpr` which we have enabled by default.

psychocoderHPC: Normally the compiler perform the `constexpr` operation during the compile time but maybe it works with runtime values too.
I will check your changes with HIP-clang too, to avoid that we remove required code.

psychocoderHPC: HIP-clang works!

psychocoderHPC (approved):

PR #978, opened 2020-04-15
Example heat-equation

jkrude:
Alpaka implementation of solving the heat-equation based on a CPU version from @sbastrakov.


BenjaminW3:
clang as a CUDA compier complains about:
```
heatEquation.cpp:203:28: warning: using integer absolute value function 'abs' when argument is of floating point type [-Wabsolute-value]
        auto const error = abs( pNextHost[ i ] - exactSolution( i * dx, tMax ) );
```

example/heatEquation/src/heatEquation.cpp
```diff
@@ -0,0 +1,210 @@
...
+
+    // Copy device -> host
+    alpaka::mem::view::copy(queue, uNextBufHost, uNextBufAcc, extent);
+    alpaka::wait::wait(queue);
+
+    // Calculate error
+    double maxError = 0.0;
+    for( uint32_t i = 0; i < numNodesX; i++ )
+    {
+        auto const error = abs( pNextHost[ i ] - exactSolution( i * dx, tMax ) );
```
psychocoderHPC: you must use `std::abs` because `abs()` is integer only.
This will than also solve the current CI issue that `error` is an integer.

example/heatEquation/src/heatEquation.cpp
```diff
@@ -0,0 +1,210 @@
...
+//! u(x, 0) = sin(pi * x)
+//!
+//! \param x value of x
+//! \param t value of t
+double exactSolution(
+    double const x,
+    double const t)
+{
+    constexpr double pi = 3.14159265358979323846;
+    return exp( -pi * pi * t ) * sin( pi * x );
```
psychocoderHPC: please use `std::exp` and std::sin`

example/heatEquation/src/heatEquation.cpp
```diff
@@ -0,0 +1,210 @@
...
+    for( uint32_t i = 0; i < numNodesX; i++ )
+    {
+        auto const error = abs( pNextHost[ i ] - exactSolution( i * dx, tMax ) );
+        maxError = std::max( maxError, error );
+    }
+    std::cout << "Max error to the exact solution at t = tMax: " << maxError << "\n";
+
+    return EXIT_SUCCESS;
+#endif
+}
```
psychocoderHPC: empty line missing

example/heatEquation/src/heatEquation.cpp
```diff
@@ -0,0 +1,210 @@
...
+    // Calculate error
+    double maxError = 0.0;
+    for( uint32_t i = 0; i < numNodesX; i++ )
+    {
+        auto const error = abs( pNextHost[ i ] - exactSolution( i * dx, tMax ) );
+        maxError = std::max( maxError, error );
+    }
+    std::cout << "Max error to the exact solution at t = tMax: " << maxError << "\n";
+
+    return EXIT_SUCCESS;
```
psychocoderHPC, line 303: Could we return `EXIT_FAILURE` if the error is to large. Not sure `to large` for this example is but we should set it based on a test run on different hardware.

psychocoderHPC (changes requested):
thx for this new example

psychocoderHPC:
Please also rebase against #979

example/heatEquation/src/heatEquation.cpp
sbastrakov: Agree, we should add a check and report whether the test passed or failed. I will find the appropriate tolerance for this check (this is not immediately obvious). Output of error itself was useful during the development, but not good for testing.

example/heatEquation/src/heatEquation.cpp
sbastrakov: This is true, my sloppy coding for the original version.

psychocoderHPC:
@jkrude  The curent issue is a heap buffer overflow.
https://travis-ci.com/github/alpaka-group/alpaka/jobs/321991946#L2496-L2515

example/heatEquation/src/heatEquation.cpp
```diff
@@ -0,0 +1,210 @@
...
+        double const & extent,
+        double const & dx,
+        double const & dt
+
+    )const -> void
+    {
+        // Each kernel executes one element
+        double const r = dt / ( dx * dx );
+        int idx = alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        if(idx > 0 && idx < extent){
```
psychocoderHPC: IMO this is the BUG: it must be

```
if(idx > 0 && idx < extent - 1u)
{ ...
```

because you access later `uCurrBuf[idx + 1]` and this will be outside of the buffer for `idx=extent -1`

psychocoderHPC (changes requested):
IMO the bug is here

example/heatEquation/src/heatEquation.cpp
psychocoderHPC: @sbastrakov You can keep the output, I saw it also in many CUDA examples. 
Never the less could you please provide the PR with a approximately error range checking.

psychocoderHPC (approved):

psychocoderHPC:
I will merge this PR as it is the error margin validation can be pushed as separate PR.

PR #981, opened 2020-04-21
Reduce code duplication in CUDA memcpy and memset implementations

BenjaminW3:
Since they are now using the same underlying CUDA API calls, they can reuse
the implementation and add the final wait depending on the queue type.

Result: -270 lines of code

### DIFF
--- a/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp
+++ b/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp
@@ -53,6 +53,52 @@ namespace alpaka
             {
                 namespace detail
                 {
+                    //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
@BenjaminW3 Please set always the milestone. This will help to create the changelog or later use the PIConGPU tool to auto generate it.

BenjaminW3:
@psychocoderHPC Could you please approve the changes? The Mac build is flaky.

psychocoderHPC:
Yes, I am still review it.

BenjaminW3:
The changeset looks huge but I only moved the code to common implementation base classes (they were already there for copy but not for set).

psychocoderHPC (approved):

PR #986, opened 2020-04-25
Refactor math operations unit tests

BenjaminW3:
* rename `mathOps` to `math` unit test to match folder name like other tests
* merge `sincos` test target into `math` test
* move headers into src folder to remove relative includes


BenjaminW3:
Again one of the random Mac failures. Can be ignored.

psychocoderHPC (approved):

PR #989, opened 2020-04-27
fix CMake compiler option handling

psychocoderHPC:
Currenlty we split very often the compiler option and the argument.
This creates issues when the compiler parses the given arguments.
CMake or we in alpaka remove duplicated options and creates therefore wrong argument lines:

example:
`-Xcudafe --display_error_number -Xcudafe --diag_suppress=esa_on_defaulted_function_ignored`
was changed to
`-Xcudafe --display_error_number --diag_suppress=esa_on_defaulted_function_ignored`
and results in an error:
`nvcc fatal   : Unknown option '-diag_suppress'` 

The fix is that the option and argument is explicit grouped.

This bug is included in 0.4.0 too, a backport must be manually applied because we changed our CMake structure in the develop branch.

I found the bug during [tests with docker for PIConGPU](https://github.com/ComputationalRadiationPhysics/picongpu/pull/3196#issuecomment-618902562).  
If the bug becomes visible depends on the order of arguments. In the worst case an option could be ignored by the compiler without throwing an error.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -488,8 +488,7 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 foreach(_CUDA_ARCH_ELEM ${ALPAKA_CUDA_ARCH})
                     # set flags to create device code for the given architecture
                     list(APPEND CUDA_NVCC_FLAGS
-                        --generate-code arch=compute_${_CUDA_ARCH_ELEM},code=sm_${_CUDA_ARCH_ELEM}
-                        --generate-code arch=compute_${_CUDA_ARCH_ELEM},code=compute_${_CUDA_ARCH_ELEM}
+                        "--generate-code arch=compute_${_CUDA_ARCH_ELEM},code=sm_${_CUDA_ARCH_ELEM} --generate-code arch=compute_${_CUDA_ARCH_ELEM},code=compute_${_CUDA_ARCH_ELEM}"
                     )
                 endforeach()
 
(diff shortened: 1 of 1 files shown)


BenjaminW3:
On windows it complains: `nvcc fatal   : Unknown option '-generate-code arch'`

psychocoderHPC:
> On windows it complains: `nvcc fatal : Unknown option '-generate-code arch'`

Where do you get this error from? I do not see any broken windows CI test. Is it on your machine?

BenjaminW3:
No, it is the 7th build from the bottom (github action)

psychocoderHPC:
I will check it  maybe it is not allowed to embed the ptx code when seperateable compile is selected.


Am 27. April 2020 19:28:39 MESZ schrieb Benjamin Worpitz <notifications@github.com>:

psychocoderHPC:
The problem is when separate able compile is activated CMake is keeping the quotes `"` around some command during the linking. 

```
cd /home/rwidera/workspace/buildAlpaka/example/bufferCopy && /opt/spack-modules/linux-ubuntu16.04-broadwell/gcc-5.4.0/cuda-10.2.89-2gpveayvfoo7rfq2gbj7s2ykggvfxplz/bin/nvcc --expt-extended-lambda --expt-relaxed-constexpr "--generate-code arch=compute_70,code=sm_70" "--generate-code arch=compute_70,code=compute_70" -std=c++14 --use_fast_math --ftz=false "-Xcudafe --display_error_number" "-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored" "-Xcudafe --diag_suppress=2976" -m64 -ccbin /usr/bin/c++ -dlink /home/rwidera/workspace/buildAlpaka/example/bufferCopy/CMakeFiles/bufferCopy.dir/src/./bufferCopy_generated_bufferCopy.cpp.o -o /home/rwidera/workspace/buildAlpaka/example/bufferCopy/CMakeFiles/bufferCopy.dir/./bufferCopy_intermediate_link.
```

psychocoderHPC:
> The problem is when separate able compile is activated CMake is keeping the quotes `"` around some command during the linking.
> 
> ```
> cd /home/rwidera/workspace/buildAlpaka/example/bufferCopy && /opt/spack-modules/linux-ubuntu16.04-broadwell/gcc-5.4.0/cuda-10.2.89-2gpveayvfoo7rfq2gbj7s2ykggvfxplz/bin/nvcc --expt-extended-lambda --expt-relaxed-constexpr "--generate-code arch=compute_70,code=sm_70" "--generate-code arch=compute_70,code=compute_70" -std=c++14 --use_fast_math --ftz=false "-Xcudafe --display_error_number" "-Xcudafe --diag_suppress=esa_on_defaulted_function_ignored" "-Xcudafe --diag_suppress=2976" -m64 -ccbin /usr/bin/c++ -dlink /home/rwidera/workspace/buildAlpaka/example/bufferCopy/CMakeFiles/bufferCopy.dir/src/./bufferCopy_generated_bufferCopy.cpp.o -o /home/rwidera/workspace/buildAlpaka/example/bufferCopy/CMakeFiles/bufferCopy.dir/./bufferCopy_intermediate_link.
> ```

When `CUDA_SEPARABLE_COMPILATION` `nvcc` will be used for linking insted of the normal compiler. For the linker the `CUDA_NVCC_FLAGS` which are defined as `STRING` will somehow not remove the quotes.
I am currently looking into the CMake find module, it looks like a bug.

psychocoderHPC:
I will close this issue. The problem was triggered by our third party project in PIConGPU.

The problem why `-Xcudafe` was removed when it was multiple times used as option is an issue in ISAAC. ISAAC removes duplicated options in `CUDA_NVCC_FLAGS` [see here](https://github.com/ComputationalRadiationPhysics/isaac/blob/07a8bac06ac2090c6c7269bc5a009e33d68c76e4/lib/ISAACConfig.cmake#L149)

Normally removing duplicated options should not break the build system. The problem is that we add options normally wrong. I tried to solve it in this PR but when enabling `CUDA_SPERATABLE_COMPILE`  the commands will be passt including the quotes to nvcc.
I will not waist more time into it, FindCUDA is not activly developed (as I know) since CMake has native CUDA support.

PR #991, opened 2020-04-29
fix CMake compiler option handling

psychocoderHPC:
This is the replacement PR for #989

Currenlty we split very often the compiler option and the argument.
This creates issues when the compiler parses the given arguments.

The fix is that the compiler option and argument will be passed with `=` instead
of a space in between.

The bug is triggered if a cmake script removes duplicated flags with `list(REMOVE_DUPLICATES CUDA_NVCC_FLAGS)` [ref](https://github.com/ComputationalRadiationPhysics/isaac/blob/07a8bac06ac2090c6c7269bc5a009e33d68c76e4/lib/ISAACConfig.cmake#L149)

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -488,8 +488,8 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 foreach(_CUDA_ARCH_ELEM ${ALPAKA_CUDA_ARCH})
                     # set flags to create device code for the given architecture
                     list(APPEND CUDA_NVCC_FLAGS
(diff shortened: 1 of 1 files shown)


BenjaminW3 (approved):

BenjaminW3:
From my point of view this can be merged. Do you want to try this with any other external repositories (PIConGPU)?

psychocoderHPC:
> From my point of view this can be merged. Do you want to try this with any other external repositories (PIConGPU)?

It can be merged. I tested it with the upcoming release branch of PIConGPU. This PR is compatible with the separable compile issue and allow to use remove duplication on the flag list.

Issue #15, opened 2014-11-26
add cuda kernel predicate voting operations


BenjaminW3:
- [x] __all/__all_sync(predicate)
  Evaluate predicate for all active threads of the warp and return non-zero if and only if predicate evaluates to non-zero for all of them. Supported by devices of compute capability 1.2 and higher.
- [x] __any/__any_sync(predicate)
  Evaluate predicate for all active threads of the warp and return non-zero if and only if predicate evaluates to non-zero for any of them. Supported by devices of compute capability 1.2 and higher.
- [x] __ballot/__ballot_sync(predicate)
  Evaluate predicate for all active threads of the warp and return an integer whose Nth bit is set if and only if predicate evaluates to non-zero for the Nth thread of the warp and the Nth thread is active. Supported by devices of compute capability 2.0 and higher.

For each of these warp vote operations, the result excludes threads that are inactive (e.g., due to warp divergence). Inactive threads are represented by 0 bits in the value returned by __ballot() and are not considered in the reductions performed by __all() and __any().

felicepantaleo:
just checking if there is any plan on this

psychocoderHPC:
@felicepantaleo Thanks for triggering this issue gain. @alpaka-group/alpaka-maintainers we need to increase the priority to implement these intrinsics.

sbastrakov:
As a note, CUDA has these operations on the warp level. Alpaka currently does not have any abstraction for it. So in case we want to make this in Alpaka, I think we should first define what is warp abstraction, or do it on the block level, or somehow else resolve this.

psychocoderHPC:
> As a note, CUDA has these operations on the warp level. Alpaka currently does not have any abstraction for it. So in case we want to make this in Alpaka, I think we should first define what is warp abstraction, or do it on the block level, or somehow else resolve this.

If we implement it on block level all threads must be call this functions collectively so it is not possible to use this function when you have branches in the code.

sbastrakov:
Ah, true. But does it essentially mean that to define such operations in Alpaka, it needs to define the warp abstraction first?

BenjaminW3:
What we should do depends on the use case. Does the user really want Warp level operations or block level operations?

When the user wants Warp-level operations it is really easy to implement a basic version because on CPU devices our warp size could simply be 1 and so the operations would nearly get no-ops while we would have to expose the warp size and the real operations on the CUDA backend.
[...]

PR #1003, opened 2020-05-18
Add warp voting functions

sbastrakov:
This PR adds concepts and implementation for simple warp voting functions: `all()`, `any()`, `ballot()`, as well as accessing warp size on host and device sides.

There is a discrepancy between modern CUDA vs. HIP and CUDA < 9.0. The [modern CUDA versions](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#warp-vote-functions) have `_sync` and accept a custom thread mask. However, in HIP there seems to be no interface to get this mask similar to `__activemask()` in CUDA, and instead all of these functions are [automatically applied for such threads](https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_kernel_language.md#warp-vote-and-ballot-functions). So I went for a more simple HIP-like interface and for CUDA backend use the modern function inside.

For CPUs the fake warp size is 1 and all operations behave accordingly. This is what was discussed in #15.

This is currently a prototype, not tested yet. TODO:
- [x] test with OpenMP, CUDA, HIP
- [x] add unit tests
- [x] add documentation for users


sbastrakov:
cc @bernhardmgruber

include/alpaka/warp/Traits.hpp
```diff
@@ -0,0 +1,162 @@
...
+namespace alpaka
+{
+    //-----------------------------------------------------------------------------
+    //! The thread warp specifics
+    namespace warp
+    {
+        struct ConceptWarp{};
+
+        //-----------------------------------------------------------------------------
+        //! The block synchronization traits.
```
sbastrakov: Note: copy-paste mistake, comment needs to be updated.

BenjaminW3:
I really like this!

test/common/include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -50,6 +51,14 @@ namespace alpaka
                             alpaka::workdiv::GridBlockExtentSubDivRestrictions::Unrestricted))
             {}
             //-----------------------------------------------------------------------------
+            KernelExecutionFixture(
+                WorkDiv const & workDiv) :
```
BenjaminW3, line 55: Could we use inheriting constructors here and let the other one above inherit from this one?

BenjaminW3 (dismissed):
I really like this!

test/common/include/alpaka/test/KernelExecutionFixture.hpp
sbastrakov: I originally tried that. But it turned out somewhat awkward to express in this case: the constructor above would need to not initialize members (as initialization in both constructors is not allowed), but it needs this values for the extent computation call.

sbastrakov:
@BenjaminW3 let me quickly rebase to tidy up the history. I also previously got some weird CI error with OpenMP 4 on two configurations, that does not reproduce on Hemera. And another issue with PTX version with CUDA on one configuration, trying to reproduce this one.

sbastrakov:
So for clang 10 CUDA 9.2 configuration I am getting error
`ptxas /tmp/Activemask-0ce246.s, line 236; error   : Feature 'activemask' requires PTX ISA .version 6.2 or later`. Not really sure what causes it: CUDA 9.2 [should have PTX ISA 6.2](https://docs.nvidia.com/cuda/archive/9.2/parallel-thread-execution/index.html). Googling suggests the issue may be with the driver.

BenjaminW3:
I do not know the reason but you could try to play around by adding `ALPAKA_CUDA_ARCH: "72"` or similar to the jobs env.

BenjaminW3:
The only relevant difference to other clang 10+CUDA build jobs is that this one is using `libc++` instead of `libstdc++`.

sbastrakov:
Setting `ALPAKA_CUDA_ARCH: "72"` for that configuration did not help.

sbastrakov:
@psychocoderHPC perhaps you have an idea what fails in the CI?

include/alpaka/warp/Traits.hpp
```diff
@@ -0,0 +1,193 @@
...
+                ConceptWarp,
+                TWarp>;
+            return traits::GetSize<
+                ImplementationBase>
+            ::getSize(
+                warp);
+        }
+
+        //-----------------------------------------------------------------------------
+        //! Evaluates predicate for all active threads of the warp and returns
```
psychocoderHPC, line 117: We need to check if this is correct doxygen. IMO `//!` is for the brief 

https://www.star.bnl.gov/public/comp/sofi/doxygen/docblocks.html

psychocoderHPC (changes requested):

include/alpaka/warp/Traits.hpp
BenjaminW3: I always use `//! ` and Doxygen does not differentiate at all.

sbastrakov: I think I copied this style from some Alpaka file. But do not remember, perhaps the original was indeed a brief, but my comments should certainly not be one.

psychocoderHPC:
for documentation: an overview which ISA is used by which driver/CUDA library: https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#release-notes

include/alpaka/warp/Traits.hpp
sbastrakov: Okay, after reading the document linked I think this style is actually fine, and is not a brief. (Covered by point 3 of the doc).

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,152 @@
...
+            struct All<
+                WarpUniformCudaHipBuiltIn>
+            {
+                //-----------------------------------------------------------------------------
+                __device__ static auto all(
+                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
+                    std::int32_t predicate)
+                -> std::int32_t
+                {
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
```
psychocoderHPC: The test linux_clang-10_cuda-9.2_release us with this fix working, this mean that something is wrong with our CI. IMO the CI test is using the wrong CUDA version. Instead of CUDA 9.2 maybe an older installed version is used.

psychocoderHPC: https://github.com/alpaka-group/alpaka/pull/1003/checks?check_run_id=794879618#step:4:5738 is showing that we are using the second branch for CUDA < 9.2

sbastrakov: Turned out, I made a mistake in the check and should have checked the clang cuda version, not nvcc version. This is why we went to another branch. Investigating further now.

psychocoderHPC:
The usage of activemask was introduced in clang here: https://github.com/llvm-mirror/clang/commit/4fbb9360b74b83cec392f205eb7df60838482304

In clang 9.0 ist was not using the ptx command `activemask`

We need to find out why clang is creating PTX ISA <6.2

psychocoderHPC:
The issue is reproducable on our test system: fwk394

```
module load llvm-10.0.0-gcc-8.3.0-pooiqx6
spack load cuda@9.2.88
module load cmake-3.17.1-gcc-8.3.0-ezyvmq4
module load boost-1.68.0-gcc-5.4.0-3et7t3g

# fail to compile
cmake -G 'Unix Makefiles' -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON '-DCMAKE_CXX_FLAGS= -stdlib=libc++' '-DCMAKE_EXE_LINKER_FLAGS= -lc++ -lc++abi' -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DALPAKA_CUDA_VERSION=9.2 -DALPAKA_CUDA_ARCH=70 -DALPAKA_CUDA_COMPILER=clang -DALPAKA_ACC_GPU_HIP_ENABLE=OFF ../alpaka/

# compiles because the sm version is changed 60
cmake -G 'Unix Makefiles' -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON '-DCMAKE_CXX_FLAGS= -stdlib=libc++' '-DCMAKE_EXE_LINKER_FLAGS= -lc++ -lc++abi' -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DALPAKA_CUDA_VERSION=9.2 -DALPAKA_CUDA_ARCH=60 -DALPAKA_CUDA_COMPILER=clang -DALPAKA_ACC_GPU_HIP_ENABLE=OFF ../alpaka/

```

To get temporary files is added with ccmake to  `CMAKE_CXX_FLAGS` `-save-temps -Wno-extra-semi-stmt -Wno-comma -Wno-error`

After looking into the generated ptx code for sm_70 I can see that 6.1 is the used ISA.

```
.version 6.1
.target sm_70
.address_size 64
```

psychocoderHPC:
@SimeonEhrig  was able to compile all with clang 10.0.1 in a docker container. I am currently compile clang 10.0.1 to reproduce it. It could be that clang 10.0.0 is selecting always the wrong ISA.

psychocoderHPC:
my tests with `llvm-10.0.1-rc1` still failing.

I suggest the following patch to solve the issue with 10.0.0

```diff
diff --git a/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp b/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
index 68398a6fc..74b885999 100644
--- a/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
@@ -71,13 +71,16 @@ namespace alpaka
             {
                 //-----------------------------------------------------------------------------
                 __device__ static auto all(
-                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
+                    warp::WarpUniformCudaHipBuiltIn const & warp,
                     std::int32_t predicate)
                 -> std::int32_t
                 {
-#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    auto const mask = static_cast<uint32_t>(warp::activemask(warp));
+
                     return __all_sync(
-                        __activemask(),
+                        mask,
                         predicate);
 #else
                     return __all(predicate);
@@ -92,13 +95,16 @@ namespace alpaka
             {
                 //-----------------------------------------------------------------------------
                 __device__ static auto any(
-                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
+                    warp::WarpUniformCudaHipBuiltIn const & warp,
                     std::int32_t predicate)
                 -> std::int32_t
                 {
-#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    auto const mask = static_cast<uint32_t>(warp::activemask(warp));
+
                     return __any_sync(
-                        __activemask(),
+                        mask,
                         predicate);
 #else
                     return __any(predicate);
@@ -113,13 +119,16 @@ namespace alpaka
             {
                 //-----------------------------------------------------------------------------
                 __device__ static auto ballot(
-                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
+                    warp::WarpUniformCudaHipBuiltIn const & warp,
                     std::int32_t predicate)
                 -> std::uint64_t
                 {
-#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    auto const mask = static_cast<uint32_t>(warp::activemask(warp));
+
                     return __ballot_sync(
-                        __activemask(),
+                        mask,
                         predicate);
 #else
                     return static_cast<std::uint64_t>(__ballot(predicate));
@@ -137,8 +146,13 @@ namespace alpaka
                     warp::WarpUniformCudaHipBuiltIn const & /*warp*/)
                 -> std::uint64_t
                 {
-#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+#if BOOST_COMP_CLANG_CUDA == BOOST_VERSION_NUMBER(10, 0, 0)
+                    return static_cast<std::uint64_t>(__ballot_sync(0xffffffff,1));
+#else
                     return static_cast<std::uint64_t>(__activemask());
+#endif
 #else
                     // No HIP intrinsic for it, emulate via ballot
                     return static_cast<std::uint64_t>(__ballot(1));

```

SimeonEhrig:
> my tests with `llvm-10.0.1-rc1` still failing.
> 
> I suggest the following patch to solve the issue with 10.0.0
> 
> ```diff
> diff --git a/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp b/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
> index 68398a6fc..74b885999 100644
> --- a/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
> +++ b/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
> @@ -71,13 +71,16 @@ namespace alpaka
>              {
>                  //-----------------------------------------------------------------------------
>                  __device__ static auto all(
> -                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
> +                    warp::WarpUniformCudaHipBuiltIn const & warp,
>                      std::int32_t predicate)
>                  -> std::int32_t
>                  {
> -#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
> +
> +#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
> +                    auto const mask = static_cast<uint32_t>(warp::activemask(warp));
> +
>                      return __all_sync(
> -                        __activemask(),
> +                        mask,
>                          predicate);
>  #else
>                      return __all(predicate);
> @@ -92,13 +95,16 @@ namespace alpaka
>              {
>                  //-----------------------------------------------------------------------------
>                  __device__ static auto any(
> -                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
> +                    warp::WarpUniformCudaHipBuiltIn const & warp,
>                      std::int32_t predicate)
>                  -> std::int32_t
>                  {
> -#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
> +
> +#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
> +                    auto const mask = static_cast<uint32_t>(warp::activemask(warp));
> +
>                      return __any_sync(
> -                        __activemask(),
> +                        mask,
>                          predicate);
>  #else
>                      return __any(predicate);
> @@ -113,13 +119,16 @@ namespace alpaka
>              {
>                  //-----------------------------------------------------------------------------
>                  __device__ static auto ballot(
> -                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/,
> +                    warp::WarpUniformCudaHipBuiltIn const & warp,
>                      std::int32_t predicate)
>                  -> std::uint64_t
>                  {
> -#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
> +
> +#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
> +                    auto const mask = static_cast<uint32_t>(warp::activemask(warp));
> +
>                      return __ballot_sync(
> -                        __activemask(),
> +                        mask,
>                          predicate);
>  #else
>                      return static_cast<std::uint64_t>(__ballot(predicate));
> @@ -137,8 +146,13 @@ namespace alpaka
>                      warp::WarpUniformCudaHipBuiltIn const & /*warp*/)
>                  -> std::uint64_t
>                  {
> -#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 2, 0)
> +
> +#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
> +#if BOOST_COMP_CLANG_CUDA == BOOST_VERSION_NUMBER(10, 0, 0)
> +                    return static_cast<std::uint64_t>(__ballot_sync(0xffffffff,1));
> +#else
>                      return static_cast<std::uint64_t>(__activemask());
> +#endif
>  #else
>                      // No HIP intrinsic for it, emulate via ballot
>                      return static_cast<std::uint64_t>(__ballot(1));
> ```

@psychocoderHPC 
Can you compile another example and check the PTX ISA Level?

psychocoderHPC:
> @psychocoderHPC
> Can you compile another example and check the PTX ISA Level?

Yes I will do it tomorrow,

psychocoderHPC:
@SimeonEhrig 

main.cpp
```
#include <cstdio>
#include <cuda_runtime.h>

__global__ void foo()
{
        printf("foo me\n");
}

int main()
{
        foo<<<1,1>>>();
        return 0;
}
```

on fwk394
```
spack load llvm@10.0.1-rc1
spack load cuda@9.2.88
module load cmake-3.17.1-gcc-8.3.0-ezyvmq4
module load boost-1.68.0-gcc-5.4.0-3et7t3g

clang++ -x cuda -std=c++14 main.cpp -save-tem^C -Wno-extra-semi-stmt -Wno-comma -Wno-error --cuda-gpu-arch=sm_70 --cuda-path=$CUDA_HOME -stdlib=libc++ -L$CUDA_HOME/lib64 -lcudart_static -ldl -lrt -pthread
```

results in ISA 6.1:
```
.version 6.1
.target sm_70
.address_size 64
```

llvm was compiled with spack on fwk394: (not I update the spack recipe to have access to the rc1 of clang 10.0.1

```
sudo spack install llvm@10.0.1-rc1%gcc@8.3.0 ~all_targets+clang+compiler-rt~cuda+gold+internal_unwind+libcxx+lld+lldb~omp_debug~omp_tsan+polly~python~shared_libs~split_dwarf build_type=Release
```

psychocoderHPC:
I checked with the mini example clang 10.0.0 again. The wrong ISA is only created with CUDA 9.2.
This behavior is equal to clang 10.0.1-rc1.
If I choose cuda 10.0+ the mini app is producing the correct ISA.

SimeonEhrig:
> I checked with the mini example clang 10.0.0 again. The wrong ISA is only created with CUDA 9.2.
> This behavior is equal to clang 10.0.1-rc1.
> If I choose cuda 10.0+ the mini app is producing the correct ISA.

I test it with different CUDA and Clang versions.
(same arguments as @psychocoderHPC , except `-stdlib=libc++` -> I compiled with gnu lib) 

CUDA 9.1 + Clang 10.0.1 = ISA 6.1
CUDA 9.1 + Clang 9.0.1 = ISA 6.1

CUDA 9.2 + Clang 10.0.1 = ISA 6.1
CUDA 9.2 + Clang 9.0.1 = ISA 6.1

CUDA 10.0 + Clang 10.0.1 = ISA 6.3
CUDA 10.0 + Clang 9.0.1 = ISA 6.3

psychocoderHPC:
I found the issue: https://github.com/llvm/llvm-project/blob/a9627b7ea7e2b47488188cb1d80a23b24a8066b2/clang/lib/Driver/ToolChains/Cuda.cpp#L661 It must be `+ptx62`

@SimeonEhrig  Where can I open a llvm issue. They do not handle it over github. Or could you provide a bugfix to there main repo. But we need to test the fix first.

update: https://github.com/llvm/llvm-project/blob/a9627b7ea7e2b47488188cb1d80a23b24a8066b2/clang/lib/Basic/Targets/NVPTX.cpp#L46-L59 must be updated too.

And there are some more places in if I compare the diff where they added a new CUDA version: https://github.com/llvm/llvm-project/commit/a9627b7ea7e2b47488188cb1d80a23b24a8066b2#diff-a0dc310bd5611992c6c4a84e934ec4ff

It looks like they skipped PTX ISA 6.2 fully.

psychocoderHPC:
@sbastrakov In this PR we need to apply the workaround for active mask for all clang + CUDA 9.2.

sbastrakov:
@psychocoderHPC will do. Thanks @psychocoderHPC and @SimeonEhrig for investigating and figuring out the issue! I was totally lost at that point.

SimeonEhrig:
> I found the issue: https://github.com/llvm/llvm-project/blob/a9627b7ea7e2b47488188cb1d80a23b24a8066b2/clang/lib/Driver/ToolChains/Cuda.cpp#L661 It must be `+ptx62`
> 
> @SimeonEhrig Where can I open a llvm issue. They do not handle it over github. Or could you provide a bugfix to there main repo. But we need to test the fix first.
> 
> update: https://github.com/llvm/llvm-project/blob/a9627b7ea7e2b47488188cb1d80a23b24a8066b2/clang/lib/Basic/Targets/NVPTX.cpp#L46-L59 must be updated too.
> 
> And there are some more places in if I compare the diff where they added a new CUDA version: [llvm/llvm-project@a9627b7#diff-a0dc310bd5611992c6c4a84e934ec4ff](https://github.com/llvm/llvm-project/commit/a9627b7ea7e2b47488188cb1d80a23b24a8066b2#diff-a0dc310bd5611992c6c4a84e934ec4ff)
> 
> It looks like they skipped PTX ISA 6.2 fully.

For LLVM bugs you have to use her own bug tracker: https://bugs.llvm.org/

include/alpaka/warp/Traits.hpp
```diff
@@ -0,0 +1,193 @@
...
+        //! thread of the warp is active.
+        //!
+        //! \tparam TWarp The warp implementation type.
+        //! \param warp The warp implementation.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<
+            typename TWarp>
+        ALPAKA_FN_ACC auto activemask(
+            TWarp const & warp)
+        -> std::uint64_t
```
psychocoderHPC: we should not enforce a 64bit activemask. The retuen value should be implementaion dependant. 32bit for CUDA and 64bit for HIP
The comment should have a doxygen \return 64bit or 32bit type depending on the accelerator.

psychocoderHPC (changes requested):

include/alpaka/warp/Traits.hpp
sbastrakov: I do not agree. I think it's much easier for a user to work with a fixed type, and not have conditional types depending on the backend.

sbastrakov: And with your suggestion, a user is forced to write those ugly `static_cast`s that we have inside already.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,168 @@
...
+            template<>
+            struct GetSize<
+                WarpUniformCudaHipBuiltIn>
+            {
+                //-----------------------------------------------------------------------------
+                __device__ static auto getSize(
+                    warp::WarpUniformCudaHipBuiltIn const & /*warp*/)
+                -> std::int32_t
+                {
+                    return warpSize;
```
psychocoderHPC, line 63: where is this warpSize  coming from?

sbastrakov: It is automatic variable in CUDA and HIP device side, like `threadIdx`

include/alpaka/warp/Traits.hpp
psychocoderHPC: A user can use `auto mask = alpaka::activeMask()`. The point is we waist here register and performance.
A typically use case for active mask is `uint32_t threads = alpaka::popc(mask)`. 
Do you have a use case in mind where a static cast is required?

sbastrakov: Discussed offline, I will implement this suggestion.

include/alpaka/dev/Traits.hpp
```diff
@@ -136,6 +143,21 @@ namespace alpaka
                     dev);
         }
 
+        //-----------------------------------------------------------------------------
+        //! \return The warp size on the device in number of threads.
+        template<
+            typename TDev>
+        ALPAKA_FN_HOST auto getWarpSize(
+            TDev const & dev)
+        -> std::size_t
```
psychocoderHPC, line 152: Suggestion: we should not enforce a 64bit type here and let the compiler select the return type by looking to the implementation of the trait..

include/alpaka/warp/Traits.hpp
```diff
@@ -0,0 +1,194 @@
...
+        //! Returns warp size.
+        //!
+        //! \tparam TWarp The warp implementation type.
+        //! \param warp The warp implementation.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<
+            typename TWarp>
+        ALPAKA_FN_ACC auto getSize(
+            TWarp const & warp)
+        -> std::int32_t
```
psychocoderHPC, line 76: Suggestion: we should remove the return value here, this will give the compiler the chance to use 8Bit or 16Bit depending of what is returned in the implementation.

psychocoderHPC (approved):

psychocoderHPC:
issue: https://github.com/alpaka-group/alpaka/pull/1003/checks?check_run_id=834700330#step:4:5093

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,162 @@
...
+            struct All<
+                WarpUniformCudaHipBuiltIn>
+            {
+                //-----------------------------------------------------------------------------
+                __device__ static auto all(
+                    warp::WarpUniformCudaHipBuiltIn const & warp,
+                    std::int32_t predicate)
+                {
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    auto const mask = static_cast<std::uint32_t>(activemask(warp));
```
psychocoderHPC: BUG: do NOT cast the active mask to 32Bit, this will fail on AMD hardware where the active mask is 64bit.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,162 @@
...
+            struct Any<
+                WarpUniformCudaHipBuiltIn>
+            {
+                //-----------------------------------------------------------------------------
+                __device__ static auto any(
+                    warp::WarpUniformCudaHipBuiltIn const & warp,
+                    std::int32_t predicate)
+                {
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    auto const mask = static_cast<std::uint32_t>(activemask(warp));
```
psychocoderHPC: do not cast to 32bit

test/unit/warp/src/Activemask.cpp
```diff
@@ -0,0 +1,126 @@
...
+        typename TAcc>
+    ALPAKA_FN_ACC auto operator()(
+        TAcc const & acc,
+        bool * success) const
+    -> void
+    {
+        std::int32_t const warpExtent = alpaka::warp::getSize(acc);
+        ALPAKA_CHECK(*success, warpExtent == 1);
+
+        ALPAKA_CHECK(*success, alpaka::warp::activemask(acc) != 0u);
```
psychocoderHPC: Since we know that only one thread is active we should check for `ALPAKA_CHECK(*success, alpaka::warp::activemask(acc) == 1u);`
Because results where other than the first bit are set will be wrong.

psychocoderHPC (changes requested):

include/alpaka/dev/Traits.hpp
sbastrakov: This is not kernel-side, so does it matter?

include/alpaka/warp/Traits.hpp
sbastrakov: I don't really like the suggestion (again, a user has to do type casts, as it's not even clear whether the result is signed or unsigned). But I guess this is logical given a similar change was already done for `activemask()` and `ballot()`. Will do.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
sbastrakov: This branch is for CUDA only tho?

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
sbastrakov: And here CUDA-only as well

test/unit/warp/src/Activemask.cpp
sbastrakov: I agree

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
psychocoderHPC: ohh I missed it, you are right!

include/alpaka/dev/Traits.hpp
psychocoderHPC: I think for a common use case not, therefore it is only a suggestion.

sbastrakov:
@psychocoderHPC @BenjaminW3 now all changes are implemented, rebased, and CI is finally happy. Could you re-review and merge if everything is okay?

psychocoderHPC (approved):

PR #1004, opened 2020-05-18
Add popcount intrinsic

sbastrakov:
This PR implements an alpakafied `popcount()` to be used in kernels. It falls back to CUDA/HIP built-ins on GPUs, tries to use compiler-specific intrinsic on CPUs, and has a fallback standard library implementation.

The lonely `popcount()` as of now may seem not worthy to create another directory and concept. I made it assuming the list may be extended in the future, and this sets an example for such extensions. The choice of `popcount` is since it is required for [this work](https://github.com/ComputationalRadiationPhysics/mallocMC/pull/173) by @bernhardmgruber , and I think generally useful. However, we may think of putting it in some other way.

I added a unit test. Works on my machine with Visual Studio, I assume the tests would catch if something is wrong for other platforms.

Not sure where to put it in the docs.

Note: I choose name `popcount` as it would be named so in standard C++20.


sbastrakov:
@psychocoderHPC @BenjaminW3 do you know if there is some issue with macos CI? It shows to have failed, but no details given. (Unlike the actual issues shown in my other PR).

BenjaminW3:
yes, github seems to have some issues. Sometimes some builds do not start on macos. You can ignore them and merge either way.

sbastrakov:
Okay, now waiting for travis so far, but good to know, thanks for info.

include/alpaka/intrinsic/Traits.hpp
```diff
@@ -0,0 +1,81 @@
...
+        //-----------------------------------------------------------------------------
+        //! The intrinsics traits.
+        namespace traits
+        {
+            //#############################################################################
+            //! The popcount trait.
+            template<
+                typename TWarp,
+                typename TSfinae = void>
+            struct Popcount;
```
psychocoderHPC, line 34: As we did for all other traits e.g. math traits we should add the argument type to the `class`. This will than remove also the requirement of two type overloaded free entry function `popcount`.
Having the type in the trait class will give us also the possibility to specialize the function for 16bit or 8bit, if in the future available, without a need of a cast.

psychocoderHPC (changes requested):

include/alpaka/intrinsic/Traits.hpp
sbastrakov: I thought about it. But I decided not to do it, as now popcount explicitly works for two types only. This is not like `cos` for example that one could call for any type, including user-defined ones.

sbastrakov: So to reiterate, I could do a template parameter there. But i think this is misleading, as the type is not supposed to be generic, it is strictly one of the two supported types. And so I think it's better to be explicit. In terms of code line number both are about the same, as it's two template specialization with one function vs. non-template with two function overloads.

psychocoderHPC: but there is also a `__popcnt16` available. Maybe only for the MS compiler.

psychocoderHPC: Another point for having the type in the template is that we are working on getting vector types into alpaka, these will benefit from have it in the template signature. It can also be refactored later.

@BenjaminW3 Do you have a preference here?

sbastrakov: The common subset is 32-bit and 64-bit only, all of CUDA, HIP and gcc only support these two versions. With vector types you have a point, but the implementation would need to get changed anyways it seems, as now the code relies on specific programming model or compiler intrinsics (as is the point of intrinsics), and would need to distinguish vectors and use the appropriate library like Vc for that.

sbastrakov: But generally I'm not against your proposal, think it's also fine, but just slightly worse than my version.

psychocoderHPC: let us wait for a third opinion. I am not fixed to my suggestion.

BenjaminW3: In general I would prefer the template parameter version but in this special case I am also fine with the current solution. We can still change it later when required.

BenjaminW3 (dismissed):

include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,78 @@
...
+            struct Popcount<
+                IntrinsicUniformCudaHipBuiltIn>
+            {
+                //-----------------------------------------------------------------------------
+                __device__ static auto popcount(
+                    intrinsic::IntrinsicUniformCudaHipBuiltIn const & /*intrinsic*/,
+                    unsigned int value)
+                -> int
+                {
+                    return __popc(value);
```
sbastrakov, line 65: Oh no, clang + CUDA for some reason has the input as `int`, not `unsigned` like every other implementation.

sbastrakov: This causes tests to fail. Will fix

test/unit/intrinsic/src/IntrinsicTest.cpp
```diff
@@ -0,0 +1,77 @@
+/* Copyright 2020 Sergei Bastrakov
+ *
```
sbastrakov, line 2: @BenjaminW3 @psychocoderHPC what is the preferred convention of calling such test source files: with or without `Test` at the end? I believe alpaka uses both ways.

BenjaminW3: I think you can keep it like it is.

sbastrakov:
Now the failing CI for HIP gives

> CMake Error at cmake/VerifyCompiler.cmake:24 (find_package):
>   Could not find a package configuration file provided by "hip" with any of
>   the following names:
>     hipConfig.cmake
>     hip-config.cmake

I don't think this is the problem on this PR's side.

sbastrakov:
Oh, I meant to re-run one job and it restarted all of them

psychocoderHPC:
> Now the failing CI for HIP gives
> 
> > CMake Error at cmake/VerifyCompiler.cmake:24 (find_package):
> > Could not find a package configuration file provided by "hip" with any of
> > the following names:
> > hipConfig.cmake
> > hip-config.cmake
> 
> I don't think this is the problem on this PR's side.

I restarted the job

psychocoderHPC (dismissed):

sbastrakov:
I believe the same HIP travis job failed again with the same message @psychocoderHPC .

psychocoderHPC:
@sbastrakov I updated our HIP requirements in #1006, I hope this will solve the issue.

BenjaminW3:
@sbastrakov Please rebase to get the build green again.

psychocoderHPC (approved):

PR #1006, opened 2020-05-20
update required HIP version to 3.3.0

psychocoderHPC:
Update to the latest HIP release. Even with this release alpaka is not 100% usable with HIP because there are still bugs in HIP those will be fixed with the next HIP release [ref](https://github.com/ROCm-Developer-Tools/HIP/issues/1982)

- fix wrong used rocrand branch
- expose HIP to CMake via environment variable `CMAKE_PREFIX_PATH`
- define `HIP_PLATFORM` and `HIP_RUNTIME` before install HIP

I hope this update is also solving the HIP CI issues in #1004

PR #1007, opened 2020-05-21
Test Ubuntu 20.04

BenjaminW3:
This builds some of the CI jobs that are supported on Ubuntu 20.04.

Basic C++ builds work fine with Ubuntu 20.04 (some older compiler versions
are not available from external repositories).
However, there are many incompatibilities with the CUDA builds.

Furhermore, some clang versions require `*_finite` functions from glibc which
are not available anymore so they can not be used when -ffast-math is
enabled like in some of our tests.
see: https://bugs.llvm.org/show_bug.cgi?id=45034

Besides some installation changes nothing had to be done on the code itself.


BenjaminW3:
When this is merged, I will work on moving all builds to github actions. This will also allow to selectively require specific build targets as merge barrier which would allow us to handle the HIP issue much easier.

sbastrakov (dismissed):

psychocoderHPC:
I stopped the CI (travis not actions) for this job by hand and will restart it after #1006

BenjaminW3:
I rebased onto the fixed develop. Please approve!

psychocoderHPC (approved):

Issue #944, opened 2020-03-05
simplification of the user interface


psychocoderHPC:
# Motivation
Alpaka gives users high flexibility and freedom for their implementations.
Alpaka is explicit everywhere and can therefore be controlled in fine granularity.
For IMO 90% of the users, simple usage - at least to get started with a library - is very important. 

# Proposal 

Alpaka is from the view of a user very hard to use, therefore I like to do the first step and propose an example of our `vector add` written against a pseudo API `simple`.

## Example

```C++
using namespace alpaka;

// Define the accelerator
// - GpuCuda
// - CpuThreads
// - CpuFibers
// - CpuOmp2Threads
// - CpuOmp2Blocks
// - CpuOmp4
// - CpuTbbBlocks
// - CpuSerial
// - GpuHip
constexpr size_t dim = 1;
auto numDevices = simple:acc::count(simple::acc::GpuCuda);
// an accelerator instance is bound to an device
auto acc = simple::Acc<dim>(simple::acc::CpuSerial, numDevices % 1);
auto queue = simple::Queue(acc, simple::queue::NonBlocking{});

// Define the work division
size_t const elementsPerThread(3u);
// create array for extent indices based on the accelerator dimension
auto const extent = simple::mem::Array(acc, 123456u);

// Define the buffer element type
using Data = std::uint32_t;

// Get the host device for allocating memory on the host.
auto const accHost = simple::Acc<dim>(simple::acc::CpuSerial, numDevices % 1);

// Allocate 3 host memory buffers
auto bufHostA = simple::mem::Vector(accHost, extent);
auto bufHostB = simple::mem::Vector(accHost, extent);
auto bufHostC = simple::mem::Vector(accHost, extent);

// Random generator for uniformly distributed numbers in {1,..,42}
auto engine = simple::rng::Engine(accHost, simple::rng::engine::MersenneTwister{}, 42u);
auto rng = simple::rng::distribution::Uniform<Data>(1, 42);

// Initialize the host input vectors A and B
for (Idx i(0); i < numElements; ++i)
{
    bufHostA[i] = rng(engine);
    bufHostB[i] = rng(engine);
    bufHostC[i] = 0;
}

// Allocate 3 buffers on the accelerator
auto bufAccA = simple::mem::Vector(acc, extent);
auto bufAccB = simple::mem::Vector(acc, extent);
auto bufAccC = simple::mem::Vector(acc, extent);

// Copy Host -> Acc
[...]

PR #1017, opened 2020-05-29
simplify alpaka usage

psychocoderHPC:
# Motivation 

Currently alpaka is hard to use because the user needs to know how all concepts e.g platform, accelerator, device, queue,... and how they are connected to each other and which concept must be used where.
I started a discussion in  #944 to simplify the interface for alpaka.
This PR is introducing a full change of the usage, instead it provides the required trait specifications to allow the user a accelerator based programming. This means instead that the user needs to handle device and platform types he can use the accelerator to allocate buffers, query device information. 
Before this PR the user got the device type from an accelerator and used the device type to get the platform. This explicit traversing over types with a 1:1 relation is removed by forwarding traits to the next hierarchy level when needed. 

# General Simple Workflow

1. define an accelerator type `Acc`
2. get the device instance based on `Acc`
3. use `Acc` to create buffers

# Changes 

- accelerator
  - specialize platform traits to query device information (redirect to
platform trait)
  - specialize buffer for accelerators
- device
  - add concept `ConceptDev`
  - specialize `DevType` for a device
  - specialize platform traits to query device information (redirect to
platform trait)
- mem
   - Buffer and views not using the template parameter `TDev` directly.
     Instead `TDev` is handled as concept and the device type is first
evaluated.
- Use the simplified accelerator centric view for all examples.

### DIFF
--- a/include/alpaka/mem/buf/Traits.hpp
+++ b/include/alpaka/mem/buf/Traits.hpp
@@ -99,7 +99,8 @@ namespace alpaka
                 typename TElem,
                 typename TDim,
                 typename TIdx>
-            using Buf = typename traits::BufType<TDev, TElem, TDim, TIdx>::type;
+            using Buf = typename traits::BufType<
+                alpaka::dev::Dev<TDev>, TElem, TDim, TIdx>::type;
 
             //-----------------------------------------------------------------------------
             //! Allocates memory on the given device.
--- a/include/alpaka/pltf/Traits.hpp
+++ b/include/alpaka/pltf/Traits.hpp
@@ -10,7 +10,6 @@
 #pragma once
 
 #include <alpaka/core/Common.hpp>
-#include <alpaka/dev/Traits.hpp>
 
 #include <alpaka/core/Concepts.hpp>
 #include <alpaka/queue/Traits.hpp>
@@ -38,6 +37,16 @@ namespace alpaka
                 typename TSfinae = void>
             struct PltfType;
 
+            template<
+                typename TPltf>
+            struct PltfType<
+                TPltf,
+                typename std::enable_if<concepts::ImplementsConcept<pltf::ConceptPltf, TPltf>::value>::type
+            >
+            {
+                using type = typename concepts::ImplementationBase<dev::ConceptDev, TPltf>;
+            };
+
             //#############################################################################
             //! The device count get trait.
             template<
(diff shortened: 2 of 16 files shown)


bertwesarg:
Hi, nice simplification. One question though, why not moving `alpaka::pltf::getDevByIdx` into `alpaka::getDevByIdx` or some other name space? This way, the term `pltf' should be completely avoidable.

psychocoderHPC:
> Hi, nice simplification. One question though, why not moving `alpaka::pltf::getDevByIdx` into `alpaka::getDevByIdx` or some other name space? This way, the term `pltf' should be completely avoidable.

I like that idea. This will than also mirror the fact that `getDevByIdx`, `getDevCount` can be used with different concepts.
We can make another PR if this gets excepted.

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -99,7 +99,8 @@ namespace alpaka
                 typename TElem,
                 typename TDim,
                 typename TIdx>
-            using Buf = typename traits::BufType<TDev, TElem, TDim, TIdx>::type;
+            using Buf = typename traits::BufType<
+                alpaka::dev::Dev<TDev>, TElem, TDim, TIdx>::type;
```
BenjaminW3, line 103: I like this version. You are passing in something (`TDev`) that can give you the real device type via `alpaka::dev::Dev<>`. This extends the valid types easily.
If it is already a device, it uses the reflexive `alpaka::dev::Dev<>` specialization of the device type itself.
If it is not a device, it simply has to have a specialization of `alpaka::dev::Dev<>`.

include/alpaka/pltf/Traits.hpp
```diff
@@ -51,6 +51,50 @@ namespace alpaka
...
+                ALPAKA_FN_HOST static auto getDevCount()
+                -> std::size_t
+                {
+                    ALPAKA_DEBUG_FULL_LOG_SCOPE;
+
+                    return traits::GetDevCount<
+                        typename pltf::traits::PltfType<TDev>::type>
+                            ::getDevCount();
+                }
+            };
```
BenjaminW3: I would prefer if we could get rid of those specializations here.
We should do the same that is done for the Buffer (if possible).
May we change the implementation of the trait accessor methods below?
Instead of calling `traits::GetDevCount<TPltf>::getDevCount()` they could call `traits::GetDevCount<alpaka::pltf::Pltf<TPltf>>::getDevCount()`.
The platforms would need a reflexive specialization of the `alpaka::pltf::Pltf<>` trait and the device already has a `alpaka::pltf::Pltf<>` specialization.

BenjaminW3:
@psychocoderHPC I have pushed a simplification of your simplification onto your branch. So nothing to do for you anymore ;-)

psychocoderHPC:
@BenjaminW3 From my side the changes looks good. We should check if it is useful to add self reflection also to other traits e.g. QueueType in a separate PR.

BenjaminW3 (approved):

PR #1018, opened 2020-05-29
Add ffs intrinsic

sbastrakov:
This is extension of #1004 , adding `ffs()` integer intrinsic.


BenjaminW3:
May we want to add a MSVC version?
https://gitlab.freedesktop.org/cairo/cairo/commit/f5167dc2e1a13d8c4e5d66d7178a24b9b5e7ac7a

BenjaminW3 (dismissed):

include/alpaka/intrinsic/Traits.hpp
```diff
@@ -77,5 +84,51 @@ namespace alpaka
...
+            typename TIntrinsic>
+        ALPAKA_FN_ACC auto ffs(
+            TIntrinsic const & intrinsic,
+            long long value)
+        -> int
+        {
+            using ImplementationBase = concepts::ImplementationBase<ConceptIntrinsic, TIntrinsic>;
+            return traits::Ffs<
+                ImplementationBase>
+            ::ffs(
```
psychocoderHPC, line 130: should this not be `ffsll()` and require the incude `<ctsring>` in this file see https://linux.die.net/man/3/ffsll

psychocoderHPC:
general question: Under windows `int64_t` is `long int` under linux `long long int` but we have only a specialization for `int` and `long long int`. Do we ne need to add also `long int`? IMO `int64_t` should lead into an compile error (not tried I have no windows^^).

test/unit/intrinsic/src/Ffs.cpp
```diff
@@ -0,0 +1,77 @@
...
+
+    alpaka::test::KernelExecutionFixture<Acc> fixture(
+        alpaka::vec::Vec<Dim, Idx>::ones());
+
+    FfsTestKernel<int> kernel32bit;
+    REQUIRE(
+        fixture(
+            kernel32bit));
+
+    FfsTestKernel<long long> kernel64bit;
```
psychocoderHPC: IMO we should check `int64_t` and `int32_t`, Those types are typically used by the user and maps to `long long` or `long` but this is OS dependent.

psychocoderHPC (dismissed):

sbastrakov:
~~@BenjaminW3 MS version is signed, not unsigned, unlike GCC and CUDA/HIP. After spending some time on it, I do not see a super easy way to express it via pure intrinsic calls.~~

Edit: my comment above is wrong, it was the other way around with signed/unsigned and I see how they are expressing in the code you linked. I will add it.

test/unit/intrinsic/src/Ffs.cpp
sbastrakov: That is a good point. The problem is that how GCC and CUDA define it is for these types, of not fixed width. I can make alpaka interface more modern C++ and `static_assert` internally that the type matches. Are you fine with it?

include/alpaka/intrinsic/Traits.hpp
sbastrakov: I followed the same logic as popcount: alpaka provides a unified interface (be it with templates or overloading like we discussed during review of #1004), without this ugly C postfixes
Edit: and this is call to the method, not that Linux function (which I do not use at all).

test/unit/intrinsic/src/Ffs.cpp
sbastrakov: @psychocoderHPC I guess my last answer also relates to your [earlier comment](https://github.com/alpaka-group/alpaka/pull/1018#issuecomment-635957171), not just the one above.

sbastrakov:
@BenjaminW3 @psychocoderHPC thanks for your comments. I added intrinsic-based implementation for Visual Studio, and tried to be more modern C++ with 32-bit vs. 64-bit types, and rebased.

include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
```diff
@@ -79,6 +79,30 @@ namespace alpaka
...
+            //#############################################################################
+            template<>
+            struct Ffs<
+                IntrinsicUniformCudaHipBuiltIn>
+            {
+                //-----------------------------------------------------------------------------
+                __device__ static auto ffs(
+                    intrinsic::IntrinsicUniformCudaHipBuiltIn const & /*intrinsic*/,
+                    int value)
+                -> int
```
psychocoderHPC: why do you not return a `std::int32_t` here too? To have the same return type you have on the CPU?

psychocoderHPC:
@sbastrakov Please have a look to this CUDA CI issue: https://github.com/alpaka-group/alpaka/pull/1018/checks?check_run_id=730928384#step:4:6184

```
/home/runner/work/alpaka/alpaka/include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp:89:40: note: candidate function
                __device__ static auto ffs(
                                       ^
/home/runner/work/alpaka/alpaka/include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp:98:40: note: candidate function
                __device__ static auto ffs(
                                       ^
```

include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
sbastrakov: Yes, I should, thanks for pointing out.

BenjaminW3 (approved):

sbastrakov:
Yes, comments of @psychocoderHPC were addressed. He is on vacation however.

BenjaminW3:
@sbastrakov If your are done you can merge it.

PR #1022, opened 2020-06-03
Examples: Enable subdirectories and use alpaka_ROOT

j-stephan:
This enables easier out-of-source builds for the examples. Instead of pointing CMake to `/path/to/installed/alpaka/lib/cmake/you/get/the/idea` it is now sufficient to point to `/path/to/installed/alpaka`.

This should be merged into the new release branch as we need this for the CERN workshop.

### DIFF
--- a/example/bufferCopy/CMakeLists.txt
+++ b/example/bufferCopy/CMakeLists.txt
@@ -34,9 +34,15 @@ project(${_TARGET_NAME})
 # Find alpaka.
 
 if(NOT TARGET alpaka::alpaka)
-    set(ALPAKA_ROOT "${CMAKE_CURRENT_LIST_DIR}/../../" CACHE STRING "The location of alpakaConfig.cmake")
-    list(APPEND CMAKE_MODULE_PATH "${ALPAKA_ROOT}")
-    find_package(alpaka REQUIRED)
+    option(USE_ALPAKA_SOURCE_TREE "Use alpaka's source tree instead of an alpaka installation" OFF)
+
+    if(USE_ALPAKA_SOURCE_TREE)
+        # Don't build the examples recursively
+        set(alpaka_BUILD_EXAMPLES OFF)
+        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../.." "${CMAKE_BINARY_DIR}/alpaka")
+    else()
+        find_package(alpaka REQUIRED)
+    endif()
 endif()
 
 #-------------------------------------------------------------------------------
--- a/example/heatEquation/CMakeLists.txt
+++ b/example/heatEquation/CMakeLists.txt
@@ -35,9 +35,15 @@ project(${_TARGET_NAME})
 # Find alpaka.
 
 if(NOT TARGET alpaka::alpaka)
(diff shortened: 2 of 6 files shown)


psychocoderHPC:
IMO the right CMake way to do it is by using `add_subirectory` as I did in cupla https://github.com/alpaka-group/cupla/blob/7e7509463b48d393edc7596d36b20de2b0fef2f4/cuplaConfig.cmake#L119

CMake automatically checks environment variables e.g. `<projectname>_ROOT` means `alpaka_ROOT`
The other default way in cmake is to set the CMake variable `<projectname>_DIR` e.g. `cmake -Dalpaka_DIR=PATHTOALAPKA`

In that case I would say we should use `add_subdirectory` because the examples are user code examples so they should be written as they are external user code.

psychocoderHPC:
My suggestion is:

```
if(NOT TARGET alpaka::alpaka)
    set(alpaka_PROVIDER "intern" CACHE STRING "Select which alpaka is used")
    set_property(CACHE alpaka_PROVIDER PROPERTY STRINGS "intern;extern")

    if(${alpaka_PROVIDER} STREQUAL "intern")
       # do not build examples recursively
        set(alpaka_BUILD_EXAMPLES OFF)
        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../../" ${CMAKE_BINARY_DIR}/alpaka)
    else()
        find_package(alpaka)
    endif()
endif()
```

but maybe `alpaka_PROVIDER` should be upper case letters because most of the alpaka cmake options are upper case.

The provider option will only be visible if you compile only an example `cmake <path_to_alpaka> -D-Dalpaka_PROVIDER=extern`
CMake will in this case check the environment variable `alpaka_DIR` and `CMAKE_PREFIX_PATH`.
If both will not be provided the error looks like this

```
CMake Warning at CMakeLists.txt:44 (find_package):
  By not providing "Findalpaka.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "alpaka", but
  CMake did not find one.

  Could not find a package configuration file provided by "alpaka" with any
  of the following names:

    alpakaConfig.cmake
    alpaka-config.cmake

  Add the installation prefix of "alpaka" to CMAKE_PREFIX_PATH or set
  "alpaka_DIR" to a directory containing one of the above files.  If "alpaka"
  provides a separate development package or SDK, be sure it has been
  installed.


CMake Error at CMakeLists.txt:51 (alpaka_add_executable):
  Unknown CMake command "alpaka_add_executable".
```

update:
```
        set(alpaka_BUILD_EXAMPLES OFF)
```
This is required because when you install alpaka and build examples and tests at the same time those variable will be enabled and therefore you will add all targets again.

j-stephan:
Yeah, I just noticed that we could discard `ALPAKA_ROOT` in favour of `alpaka_ROOT`. I'm not sure I follow the reasoning with your `add_subdirectory` approach. It makes sense for PIConGPU but I don't see the need for alpaka being a subdirectory of our own examples.

j-stephan:
@psychocoderHPC Okay, I cleaned up the CMakeLists.txt for the examples. Please review :-)

psychocoderHPC (approved):

PR #1024, opened 2020-06-03
Tests for BufferSlicing

jkrude:
### Intention
- Create a **test-interface** which hides most of the alpaka-interface for a **clean easy** to understand test (`TestContainer`)
- Create a simple easy to understand way of **looping** over a **multi-dimensional** set of test-**parameters** (`ForEachType`)
- Test the copy / slicing functionality of alpaka

I think it would be useful to have reusable and reliable implementations for the first two points. I tried to create a first draft, but even after some time there is room for improvements, therefore I would appreciate some help.

### Problem 1 : Nested `forEachType`
The test should be run with a variety of combinations of parameters, specific
- Data-Types (int, float, double)
- Dimensions (1D, 2D, 3D)
- Accelerators (all enabled accelerators)

To realize this I am using the `alpaka::meta::forEachType` function, but I couldn't figure out a good way to use it nested with the different parameter-types.
A desired look would be something similar to this:
```cpp
ForEachType<DataTypes>:run(
ForEachType<DimTypes>::run(
ForEachType<TestAccs>::run(
TestSlicing) ) );
```
Where each loop is populating the test with a template-parameter.
#### Idea One
Using macros would allow a nice look but seems unsafe and preprocessor loops not like a good idea.
```cpp
#define ForEach(Types, Expression)
...
ForEach(DataTypes,
ForEach(Dimensions,
// test code (no functor needed)
 ) )
```
### Idea two
Using the forEachType wraped in a struct:
> Here the second parameter is the input for the secon loop, but I couldnt find a way to inject the TestSlicing for the third loop.

`ForEach<DimTypes>(ForEach<Dimensions>(), ForEach<TestAcc>())`

### Problem 2: Usage of `alpaka::vec::Vec`with variable dimensions
Because different extents for each dimension are desired I can't use `alpaka::vec::Vec::all()`, but my solution looks quite verbose and seems quite prone to errors. Therefore, suggestions are more than welcome.

EDIT: rephrasing and restructuring 

- [x] rebase after #1064 is merged


BenjaminW3:
> The test should be run with a variety of combinations of parameters

There should not be a need to do this with nested `forEachType` function calls. What I get from your description is exactly what most of the unit tests are already doing. However, this is all done purely with types.
Alpaka provides:
* `alpaka::test::dim::TestDims` a type list of dimensions (1D, 2D, 3D and for CPU even 4D)
* `alpaka::test::idx::TestIdxs` a type list of all possible index types (uint32_t, int16_t, ..., ..., ...)
* `alpaka::test::acc::EnabledAccs` a template type list of all enabed accelerators
* `alpaka::test::acc::TestAccs` a type list of `EnabledAccs` instantiated with the `CartesianProduct` of the `TestDims` and `TestIdxs`.

If you want more, you could simply build the `CartesianProduct` of the `TestAccs` with your own `DataTypes` list or something similar. Then you only need one single `forEachType` to iterate over the cartesion product.
I added the type parametrized testing support to catch2 so that we do not even need to write `forEachType` anymore in our unit tests but you can still see how it looked before in `unit\kernel\src\KernelLambda.cpp`.

After looking at `unit\math\src\math.cpp` we should probably also apply this there and get rid of the `ForEachFunctor`. It should be possible to calculate the `CartesianProduct` of `TestAccs` * {`float`, `double`} * {`UnaryFunctors`, `BinaryFunctors`} outside of the test body and let catch2 create a test case per entry.

BenjaminW3:
What is the goal of this PR?
The slicing (sub-view with offset) and support is already tested in `ViewSubViewTest.cpp`. It may be hidden but the real test that is executed is in `Viewtest.hpp` `testViewImmutable` and `testViewMutable`.
The logic for the buffer/view tests reduces the required amount of test code immensely. We have multiple types that comply with the View concept: {`ViewSubView`, `ViewPlainPtr`, `BufCpu`, ...}. The tests for those types only differ in how they create the buffer or view that is created but they all simply execute the shared test code in `testViewImmutable` and `testViewMutable` (depending on if they are const or not). `testViewImmutable` and `testViewMutable` simply test all operations which have to be valid on a view (copy, set, getExtentVec, getPitchBytesVec, ...). So we do not write tests for a specific method (copy) but for compliance with a concept (view, buffer, ...) and apply this to all implementations of this concept (and implementations of inheriting concepts).
All the buffers/views are already tested in all dimensionalities and the `ViewSubView` also does this with and without offsets (slicing).

sbastrakov:
@BenjaminW3 we discussed it with @psychocoderHPC and @jkrude . I think the issue is that multi-dimensional slicing and copying is prone to errors, as e.g. previously in the HIP implementation before @psychocoderHPC encountered it and pointed out to them. So the goal was to have it as a test and not have this situation repeated.

BenjaminW3:
@sbastrakov But then we should extend the already existing `ViewSubViewTest.cpp` / `testViewMutable` if it really is missing something.

BenjaminW3:
So the next step would be to find out if the test would find the issue we had with HIP and if it does not do so, only then we would have to adapt the test so that it fails.

sbastrakov:
I am not sure if that test is relevant for this case. Upon a brief look, it seems the tests you are talking about concern only slicing itself. However, implementation of copying itself also needs to be tested, as it is non-trivial. Iirc (was work by @psychocoderHPC) that HIP implementation in some cases assumed a continuous piece of memory and only worked for that case, not a general one. So in this PR slicing is only used to create a complicated "general case" situation.

BenjaminW3:
As far as I know the existing tests copy from sliced to non-sliced memory and back. We will have to double-check that.

BenjaminW3:
Here we are doing the copy operations in the tests: https://github.com/alpaka-group/alpaka/blob/develop/test/common/include/alpaka/test/mem/view/ViewTest.hpp#L356-L373

psychocoderHPC:
> What is the goal of this PR?
> The slicing (sub-view with offset) and support is already tested in `ViewSubViewTest.cpp`. It may be hidden but the real test that is executed is in `Viewtest.hpp` `testViewImmutable` and `testViewMutable`.
> The logic for the buffer/view tests reduces the required amount of test code immensely. We have multiple types that comply with the View concept: {`ViewSubView`, `ViewPlainPtr`, `BufCpu`, ...}. The tests for those types only differ in how they create the buffer or view that is created but they all simply execute the shared test code in `testViewImmutable` and `testViewMutable` (depending on if they are const or not). `testViewImmutable` and `testViewMutable` simply test all operations which have to be valid on a view (copy, set, getExtentVec, getPitchBytesVec, ...). So we do not write tests for a specific method (copy) but for compliance with a concept (view, buffer, ...) and apply this to all implementations of this concept (and implementations of inheriting concepts).
> All the buffers/views are already tested in all dimensionalities and the `ViewSubView` also does this with and without offsets (slicing).

This PR provides explicit tests. I had hard times to debug HIP issues those are covered by this tests. The hidden tests are very hard to understand and it is not obvious what is tested at all. Thats the reason why @jkrude implemented this tests.

psychocoderHPC:
Note: This tests based on my tests I wrote for HIP: https://github.com/ROCm-Developer-Tools/HIP/issues/1791
I used the nerly the same code also to ahow that hipMemset3D was broken.

psychocoderHPC:
> Here we are doing the copy operations in the tests: https://github.com/alpaka-group/alpaka/blob/develop/test/common/include/alpaka/test/mem/view/ViewTest.hpp#L356-L373

I would say this test is testing the copy for a full buffer only. Slicing by using an offset and new extent is not tested.

btw: This PR can later be extendet ro copy buffer with different pitches. This was yhe root of all issues in HIP.

BenjaminW3:
> I would say this test is testing the copy for a full buffer only. Slicing by using an offset and new extent is not tested.

I am pretty sure that the existing test tests copying from a sliced to a continuous buffer and back. We may want to add other variations/parameters.

psychocoderHPC:
@BenjaminW3 Are you in general against this PR? As I wrote before this tests based on the code I used to debug the HIP issues. 
With the current tests we had it pointed to issues in HIP but the error was a side effect deep inside of the test logic. We need to take care that the unit tests are not to abstract and generalized because this makes it hard to find the issue.

BenjaminW3:
I am not against this PR but I see unnecessary duplication.
We should at least get rid of all the nested type iteration stuff and use the existing meta programming and type parameterized testing support.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,543 @@
...
+    typename Data
+>
+struct TestSlicing
+{
+    template< typename Acc >
+    auto operator()( ) -> void
+    {
+        constexpr Idx extentsDimOne = 13;
+        constexpr Idx extentsDimTwo = 7;
+        constexpr Idx extentsDimThree = 5;
```
BenjaminW3: You could take a look at what the original of this test is doing for generating the `Vec`s at compile time:
`auto const extentBuf(alpaka::vec::createVecFromIndexedFn<Dim, alpaka::test::CreateExtentBufVal>(Idx()));`
`CreateExtentBufVal` is a template which generates a compile time value depending on the dimension and `createVecFromIndexedFn` calls this for each entry of the `Vec` it should create.
`CreateExtentBufVal` is using an explicit way to generate the return value but you could also use specialization to generate arbitrary numbers.
This allows you to get correctly sized `Vec`s for a specific dimensionality.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,543 @@
...
+
+        constexpr Idx extentsSubViewDimOne = 5;
+        constexpr Idx extentsSubViewDimTwo = 3;
+        constexpr Idx extentsSubViewDimThree = 3;
```
BenjaminW3: The original of this test also has a second explicit meta function which generates the sizes of the corresponding SubView called `CreateExtentViewVal`.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,543 @@
...
+
+        constexpr Idx offsetDimOne = 3;
+        constexpr Idx offsetDimTwo = 2;
+        constexpr Idx offsetDimThree = 1;
```
BenjaminW3: The original of this test is using `auto const offsetView(alpaka::vec::Vec<Dim, Idx>::all(static_cast<Idx>(1)))` but we could also improve this to get some more irregularity like it is done here.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,543 @@
...
+    /*
+     * FIXME
+     * Current call-chain works this way but does'nt look like it.
+     * Desired (or similar):
+     * ForEachDataType::call(
+     *     ForEachDimension::call(
+     *         ForEachAccelerator::call(
+     *             TestSlicing() ) ) )
+     */
+    ForEachDataType< size_t >( );
```
BenjaminW3: By only testing `size_t` and not all the other possible index types you lose a lot of test coverage in comparison to the original test.

jkrude:
Thanks for the input. I was not aware of `forEachType` with cross products of tuples, because I misunderstood `TestAccs` as something similar to EnabledAccs. (Maybe something like `AllTestAccCombinations` would be a better description?)
Plans:
- simplify the cmpCorrectlySlicedBuffer methods 
- implement cpu - 4Dim

I used the new CreateExtentBufVal from #1029, should there be different extents tested or is this enough?

BenjaminW3:
Testing those extents should be enough.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,459 @@
...
+>
+{
+    using Vec = alpaka::vec::Vec<
+        alpaka::dim::DimInt< 3u >,
+        Idx
+    >;
+
+
+    template< typename Data >
+    static auto cmpCorrectlySlicedBuffer(
```
BenjaminW3: This method does not compare anything. It only writes the data into a 1D buffer without padding/offsets.
Please use a different name for this.

Furthermore, the original test uses similar code to access the values (it does some more things). It is written in a dimension independent way:
https://github.com/alpaka-group/alpaka/blob/develop/test/common/include/alpaka/test/mem/view/Iterator.hpp#L135-L153
Would be nice if we could apply a similar logic here.

jkrude: I am not quite sure if my [implementation](https://github.com/alpaka-group/alpaka/blob/5726e8ae9ac25fcc7a8f8761ae8f7866061041a2/test/unit/mem/copy/src/BufSlicing.cpp#L308-L321) fulfills your request.

It seems that the slicing doesn't work properly with 4 dimensions.
The problem seems independent of `Idx` / `Acc` or `Datatype` and occurs in the slicing process with `slicedBuffer`.

(_cmp was short for compute, but I agree that's misleading._=

BenjaminW3:
Clang complains about:
```
/home/runner/work/alpaka/alpaka/test/unit/mem/copy/src/BufSlicing.cpp:207:30: error: comparing floating point with == or != is unsafe [-Werror,-Wfloat-equal]
            REQUIRE( ptrA[i] == ptrB[i] );
```

I think you can do something like `REQUIRE(ptrA[i] == Approx(ptrB[i]));`

test/unit/mem/copy/src/BufSlicing.cpp
BenjaminW3: I did not take a deep look but it is much shorter and it seems to work so I am fine ;-)

We can ignore 4D for now.

psychocoderHPC:
There is an error in the tests: https://github.com/alpaka-group/alpaka/pull/1024/checks?check_run_id=803639998#step:3:2642

psychocoderHPC:
> There is an error in the tests: https://github.com/alpaka-group/alpaka/pull/1024/checks?check_run_id=803639998#step:3:2642

It looks like a conversion issue: with HIPclang I got the following error:

```
/home/rwidera/workspace/alpaka/test/catch_main/../../thirdParty/catch2/include/catch2/catch.hpp:1186:9: note: expanded from macro 'INTERNAL_CATCH_TEMPLATE_LIST_TEST_CASE'
        INTERNAL_CATCH_TEMPLATE_LIST_TEST_CASE_2( INTERNAL_CATCH_UNIQUE_NAME( ____C_A_T_C_H____T_E_M_P_L_A_T_E____T_E_S_T____ ), INTERNAL_CATCH_UNIQUE_NAME( ____C_A_T_C_H____T_E_M_P_L_A_T_E____T_E_S_T____F_U_N_C____ ), Name, Tags, TmplList )
        ^
/home/rwidera/workspace/alpaka/test/catch_main/../../thirdParty/catch2/include/catch2/catch.hpp:1177:19: note: expanded from macro 'INTERNAL_CATCH_TEMPLATE_LIST_TEST_CASE_2'
                t.reg_tests();                                        \
                  ^
/home/rwidera/workspace/alpaka/include/alpaka/dev/DevUniformCudaHipRt.hpp:223:19: note: candidate constructor (the implicit copy constructor) not viable: no known conversion from 'alpaka::mem::buf::BufCpu<int, std::integral_constant<unsigned long, 2>, unsigned int>' to 'const alpaka::mem::buf::BufUniformCudaHipRt<int, std::integral_constant<unsigned long, 2>, unsigned int>' for 1st argument
            class BufUniformCudaHipRt;
                  ^
/home/rwidera/workspace/alpaka/include/alpaka/dev/DevUniformCudaHipRt.hpp:223:19: note: candidate constructor (the implicit move constructor) not viable: no known conversion from 'alpaka::mem::buf::BufCpu<int, std::integral_constant<unsigned long, 2>, unsigned int>' to 'alpaka::mem::buf::BufUniformCudaHipRt<int, std::integral_constant<unsigned long, 2>, unsigned int>' for 1st argument
/home/rwidera/workspace/alpaka/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp:87:32: note: candidate constructor template not viable: requires 4 arguments, but 1 was provided
                ALPAKA_FN_HOST BufUniformCudaHipRt(
                               ^
fatal error: too many errors emitted, stopping now [-ferror-limit=]
20 errors generated when compiling for gfx906.
CMake Error at bufSlicing_generated_BufSlicing.cpp.o.cmake:196 (message):
  Error generating file
  /home/rwidera/workspace/buildAlpaka/test/unit/mem/copy/CMakeFiles/bufSlicing.dir/src/./bufSlicing_generated_BufSlicing.cpp.o
```

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,308 @@
...
+
+
+    auto createDeviceBuffer(Vec extents) -> BufDevice
+    {
+        BufDevice bufDevice(
+            alpaka::mem::buf::alloc<
+                TData,
+                TIdx
+            >(
+                devHost,
```
psychocoderHPC: BUG: this must be `devHost`

psychocoderHPC (changes requested):
This is not solving the compile issue but is wrong.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -0,0 +1,308 @@
+/* Copyright 2019 Axel Huebl, Benjamin Worpitz, Jakob Krude
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <alpaka/acc/AccCpuSerial.hpp>
```
psychocoderHPC: please remove this include

psychocoderHPC (changes requested):

psychocoderHPC:
@jkrude after #1064 is merged you should rebase. There is a bug which prevent passing the CI for you example.
If you integrated my comments above + #1064 you PR will pass the CI.

BenjaminW3:
I merged #1064

test/unit/mem/copy/src/BufSlicing.cpp
jkrude: Do you mean `devAcc`?

psychocoderHPC: ohh yes I repeated the error ^^, it must be `devAcc`

jkrude:
- have run it on amd and nvidia gpu without an error
- squashed commits

psychocoderHPC (dismissed):

psychocoderHPC:
error: https://github.com/alpaka-group/alpaka/pull/1024/checks?check_run_id=849591489#step:3:2649

@sbastrakov said we should disable the warning for windows

sbastrakov:
Yes @jkrude , we already disable some warning occasionally. The warning code is part of the output for the failed CI jobs.

jkrude:
What would be the alpaka way of suppressing a warning?
My suggestion would be:
```cpp
#if defined(_MSC_VER)
    #pragma warning(push)
    #pragma warning(disable: 4127) // supress warning for c++17 conditional expression is constant
#endif
    // fourth-dimension is not supposed to be tested currently
    if(Dim::value == 4)
    {
        return;
    }
#if defined(_MSC_VER)
    #pragma warning(pop)
#endif```

sbastrakov:
@jkrude yes, this is the right way. Just please use the Boost variable `BOOST_COMP_MSVC` from `alpaka/core/BoostPredef.hpp` (it's not defined there, but we pass the boost variables in this file) to check for Visual Studio. You can use [this file](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/rand/TinyMT/tinymt32.h) as an example

sbastrakov:
And perhaps put those "warning guards" at start and end of the file, like in that example.

psychocoderHPC (approved):

sbastrakov (approved):

PR #1028, opened 2020-06-06
Simplify templated math tests

BenjaminW3:


test/unit/math/src/math.cpp
```diff
@@ -224,30 +225,9 @@ TEST_CASE("mathOps", "[math] [operator]")
...
-
-    alpaka::meta::forEachType< TestAccs >(
-        ForEachFunctor< float >( ),
-        seed
-    );
+    using Acc = std::tuple_element_t<0u, TestType>;
+    using Functor = std::tuple_element_t<1u, TestType>;
+
+    alpaka::meta::forEachType< DataTypes >(
+        TestTemplate< Acc, Functor >());
```
BenjaminW3, line 232: It is possible to use:
```
alpaka::meta::CartesianProduct<std::tuple, TestAccs, DataTypes, Functors>
```
but then nvcc complains about: `error #456: excessive recursion at instantiation of class`.
Therefore we still need one indirection through `forEachType`.

test/unit/math/src/math.cpp
```diff
@@ -42,19 +66,25 @@ struct TestKernel
...
 struct TestTemplate
 {
-    template < typename TFunctor >
-    auto operator() ( unsigned long seed ) -> void
+    template<
+        typename TData>
+    auto operator()() -> void
     {
+        std::random_device rd{};
+        auto const seed = rd();
```
BenjaminW3, line 80: We now randomly initialize the seed to get more coverage over time. I copied those two lines from the axpy test.

test/unit/math/src/math.cpp
```diff
@@ -15,8 +15,32 @@
...
+        alpaka::test::unit::math::UnaryFunctors,
+        alpaka::test::unit::math::BinaryFunctors
+    >;
+
+using TestAccFunctorTuples =
+    alpaka::meta::CartesianProduct<
+        std::tuple,
+        TestAccs,
+        Functors
+    >;
```
BenjaminW3, line 37: @jkrude Here is the meta programming that generates an entry per functor per accelerator.
You can do something similar for your use case e.g.:
```
using TestAccDataTypeTuples =
    alpaka::meta::CartesianProduct<
        std::tuple,
        alpaka::test::acc::TestAccs,
        DataTypes
    >;
```
together with `TEMPLATE_LIST_TEST_CASE` and `std::tuple_element_t` you can iterate over and get the current types.

BenjaminW3: Depending on the size of template lists you may get issues with the template instantiation depth in nvcc as I did here. So you may still need one layer of `forEachType`.

sbastrakov (approved):

Issue #409, opened 2017-09-26
optimize block shared memory allocation


BenjaminW3:
For each variable in block shared memory, a single allocation is performed.
As far as I can imagine, CUDA compilers have an advantage here, because they could sum up the size of all static block shared memory allocations at compile time.

If a kernel (and the functions called from it) require multiple block shared variables, multiple allocations are exeuted. It would be more optimal to do only one allocation for all those shared variables **(1)**. But we could combine not only allocations within a block but for the whole grid **(2)** or even the whole CPU device **(3)**.
Possible solutions for the optimizations:
* **(1)** we could add a block-shared-memory allocator which uses a preallocated memory pool. We could use MallocMC, but this might be too heavy for this first level of optimization. Alternatively a simple buffer with a hard coded limit from which all allocations are served could be used. If this buffer is fully utilized, the next allocation could either throw an exception (what does CUDA do if the shared memory is full?) or allocate new memory.
* **(2)** same as the above version, but the memory pool is held by the executor and shared by all accelerators / block-shared-memory allocators
* **(3)** same as the above version, but the memory pool is held by the cpu device and shared by all executors / accelerators / block-shared-memory allocators

psychocoderHPC:
This is a patch a of possible solution which use a limited amount of shared memory located in the stack  (CUDA also limits the shared memory per multiprocessor depending of the architecture).
```diff
--- a/thirdParty/alpaka/include/alpaka/block/shared/st/BlockSharedMemStNoSync.hpp
+++ b/thirdParty/alpaka/include/alpaka/block/shared/st/BlockSharedMemStNoSync.hpp
@@ -30,6 +30,7 @@
 
 #include <vector>                           // std::vector
 #include <memory>                           // std::unique_ptr
+#include <array>
 
 namespace alpaka
 {
@@ -50,7 +51,9 @@ namespace alpaka
                     //-----------------------------------------------------------------------------
                     //! Default constructor.
                     //-----------------------------------------------------------------------------
-                    ALPAKA_FN_ACC_NO_CUDA BlockSharedMemStNoSync() = default;
+                    ALPAKA_FN_ACC_NO_CUDA BlockSharedMemStNoSync() : memOffset(0u)
+                    {
+                    }
                     //-----------------------------------------------------------------------------
                     //! Copy constructor.
                     //-----------------------------------------------------------------------------
@@ -73,13 +76,9 @@ namespace alpaka
[...]

PR #1033, opened 2020-06-08
Use member array for block shared memory, avoid malloc

jkelling:
This PR adds block shared memory implementations `BlockSharedMemDynMember` and `BlockSharedMemStMember`, which replace `BlockSharedMemDynBoostAlignedAlloc` and `BlockSharedMemStNoSync` in the backends [AccCpuOmp2Blocks](https://github.com/alpaka-group/alpaka/compare/develop...jkelling:blockSharedMember?expand=1#diff-3337e4cfdef9f4f15dafba232a45d250), [AccCpuSerial](https://github.com/alpaka-group/alpaka/compare/develop...jkelling:blockSharedMember?expand=1#diff-8528393b30e927eea71a9d4bfb86057f) and [AccCpuTbbBlocks](https://github.com/alpaka-group/alpaka/compare/develop...jkelling:blockSharedMember?expand=1#diff-e65bf721c1e9ed77791f535d7e8c8c69) (I did not test TBB).

Here `BlockSharedMemDynMember` contains a member array which provides memory for the dynamic block shared memory and a small object allocator for static block shared memory. This implementation is akin to a patch proposed in https://github.com/alpaka-group/alpaka/issues/409 . The same strategy is also employed in the upcoming  OpenMP and OpenACC backends: https://github.com/jkelling/alpaka/tree/blockSharedMemberOpenacc .

## Features
* The size of the member array can be controlled through the cmake variable/preprocessor macro `ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB`, the default is 30KB.
* `BlockSharedMemStNoSync` has a template parameter `unsigend int TDataAlignBytes` which controls the alignment of allocations. This only works as long as the `TDataAlignBytes` is less or equal to the alignment of the `BlockSharedMemDynMember` instance, because this is where the underlying memory is coming from.
* Commit https://github.com/alpaka-group/alpaka/commit/2392b3b807ec81c2eb2c5df88bdb3a6e03a1e5c8 adds an alpaka option/preprocessor macro `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST` which defaults to `true`. There are asserts inside the alloc() function and the ctor of `BlockSharedMemDynMember` to provide a check if the static array is overflowing in a debug build. However, assert does not work in code offloaded to GPU (i.e. OpenMP, OpenACC), so the point of this option is to enable debug builds with offloading.
* `BlockSharedMemStMember::alloc` calls the default ctor of the allocated type via placement-new. Maybe there could be an alternative to alloc() which is a variadic template which calls a ctor for the allocated object (in a master thread) so the user can get initialized memory directly.

## Benefits

* There are cases, where removing malloc improves performance: https://github.com/ComputationalRadiationPhysics/picongpu/issues/2280 . There are other cases where there is no measurable difference.
* This unifies small-object allocation for static shared memory for multiple, including future backends.

### DIFF
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -28,6 +28,10 @@ namespace alpaka
         {
             namespace dyn
             {
+#if BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED)
+    #pragma warning(push)
+    #pragma warning(disable: 4324)  // warning C4324: structure was padded due to alignment specifier
+#endif
                 //#############################################################################
                 //! Dynamic block shared memory provider using fixed-size
                 //! member array to allocate memory on the stack or in shared
@@ -88,6 +92,9 @@ namespace alpaka
                     mutable std::array<uint8_t, staticAllocBytes> m_mem;
                     unsigned int m_dynSize;
                 };
+#if BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED)
+    #pragma warning(pop)
+#endif
 
                 namespace traits
                 {
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -15,6 +15,10 @@
 #include <type_traits>
 #include <array>
 
+#ifndef ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB
+#define ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB 30
+#endif
+
 namespace alpaka
 {
     namespace block
@@ -27,7 +31,7 @@ namespace alpaka
                 //! Dynamic block shared memory provider using fixed-size
                 //! member array to allocate memory on the stack or in shared
                 //! memory.
-                template<unsigned int TStaticAllocKB = 30>
+                template<unsigned int TStaticAllocKB = ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB>
                 class BlockSharedMemDynMember :
                     public concepts::Implements<ConceptBlockSharedDyn, BlockSharedMemDynMember<TStaticAllocKB>>
                 {
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -33,7 +33,7 @@ namespace alpaka
                 //! member array to allocate memory on the stack or in shared
                 //! memory.
                 template<unsigned int TStaticAllocKiB = ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB>
-                class BlockSharedMemDynMember :
+                class alignas(core::vectorization::defaultAlignment) BlockSharedMemDynMember :
                     public concepts::Implements<ConceptBlockSharedDyn, BlockSharedMemDynMember<TStaticAllocKiB>>
                 {
                 public:
@@ -64,7 +64,7 @@ namespace alpaka
                      *! alignment assumes, that the compiler places the instance
                      *! at architecture-appropriate boundaries.
                      */
-                    template<unsigned int TDataAlignBytes = core::vectorization::defaultAlignment*8>
+                    template<unsigned int TDataAlignBytes = core::vectorization::defaultAlignment>
                     uint8_t* staticMemBegin() const
                     {
                         return m_mem.data() +
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -42,7 +42,9 @@ namespace alpaka
                 {
                 public:
                     //-----------------------------------------------------------------------------
-                    BlockSharedMemDynMember(unsigned int sizeBytes) : m_dynSize(sizeBytes)
+                    BlockSharedMemDynMember(unsigned int sizeBytes)
+                        : m_dynPitch((sizeBytes/core::vectorization::defaultAlignment
+                             + (sizeBytes%core::vectorization::defaultAlignment>0))*core::vectorization::defaultAlignment)
                     {
 #if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
                         ALPAKA_ASSERT(sizeBytes <= staticAllocBytes);
@@ -62,35 +64,24 @@ namespace alpaka
                     uint8_t* dynMemBegin() const {return m_mem.data();}
 
                     /*! \return the pointer to the begin of data after the portion allocated as dynamical shared memory.
-                     *!
-                     *! \tparam TDataAlignBytes Round-up offset to multiple of
-                     *! this number of bytes to maintain alignment. This
-                     *! alignment assumes, that the compiler places the instance
-                     *! at architecture-appropriate boundaries.
                      */
-                    template<unsigned int TDataAlignBytes = core::vectorization::defaultAlignment>
                     uint8_t* staticMemBegin() const
                     {
-                        return m_mem.data() +
-                            (m_dynSize/TDataAlignBytes + (m_dynSize%TDataAlignBytes>0)*TDataAlignBytes);
+                        return m_mem.data() + m_dynPitch;
                     }
 
                     /*! \return the remaining capacity for static block shared memory.
-                     *!
-                     *! \tparam TDataAlignBytes Multiple of bytes the static offset was rounded-up to
                      */
-                    template<unsigned int TDataAlignBytes = 4>
                     unsigned int staticMemCapacity() const
                     {
-                        return staticAllocBytes -
-                            (m_dynSize/TDataAlignBytes + (m_dynSize%TDataAlignBytes>0)*TDataAlignBytes);
+                        return staticAllocBytes - m_dynPitch;
                     }
 
                 private:
                     static constexpr unsigned int staticAllocBytes = TStaticAllocKiB<<10;
 
                     mutable std::array<uint8_t, staticAllocBytes> m_mem;
-                    unsigned int m_dynSize;
+                    unsigned int m_dynPitch;
                 };
 #if BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED)
     #pragma warning(pop)
@@ -115,6 +106,9 @@ namespace alpaka
                             block::shared::dyn::BlockSharedMemDynMember<TStaticAllocKiB> const &mem)
                         -> T *
                         {
+                            static_assert(
+                                core::vectorization::defaultAlignment >= alignof(T),
+                                "Unable to get block shared dynamic memory for types with alignment higher than defaultAlignment!");
                             return reinterpret_cast<T*>(mem.dynMemBegin());
                         }
 #if BOOST_COMP_GNUC
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -44,7 +44,9 @@ namespace alpaka
                     //-----------------------------------------------------------------------------
                     BlockSharedMemDynMember(unsigned int sizeBytes) : m_dynSize(sizeBytes)
                     {
+#if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
                         ALPAKA_ASSERT(sizeBytes <= staticAllocBytes);
+#endif
                     }
                     //-----------------------------------------------------------------------------
                     BlockSharedMemDynMember(BlockSharedMemDynMember const &) = delete;
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -1,0 +1,106 @@
+/* Copyright 2020 Jeffrey Kelling
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/block/shared/dyn/Traits.hpp>
+#include <alpaka/core/Assert.hpp>
+
+#include <type_traits>
+#include <array>
+
+namespace alpaka
+{
+    namespace block
+    {
+        namespace shared
+        {
+            namespace dyn
+            {
+                //#############################################################################
+                //! Dynamic block shared memory provider using fixed-size
+                //! member array to allocate memory on the stack or in shared
+                //! memory.
+                template<unsigned int TStaticAllocKB = 30>
+                class BlockSharedMemDynMember :
+                    public concepts::Implements<ConceptBlockSharedDyn, BlockSharedMemDynMember<TStaticAllocKB>>
+                {
+                    static constexpr unsigned int staticAllocBytes = TStaticAllocKB<<10;
+
+                    mutable std::array<uint8_t, staticAllocBytes> m_mem;
+                    unsigned int m_dynSize;
+
+                public:
+                    //-----------------------------------------------------------------------------
+                    BlockSharedMemDynMember(unsigned int sizeBytes) : m_dynSize(sizeBytes)
+                    {
+                        ALPAKA_ASSERT(sizeBytes <= staticAllocBytes);
+                    }
+                    //-----------------------------------------------------------------------------
+                    BlockSharedMemDynMember(BlockSharedMemDynMember const &) = delete;
+                    //-----------------------------------------------------------------------------
+                    BlockSharedMemDynMember(BlockSharedMemDynMember &&) = delete;
+                    //-----------------------------------------------------------------------------
+                    auto operator=(BlockSharedMemDynMember const &) -> BlockSharedMemDynMember & = delete;
+                    //-----------------------------------------------------------------------------
+                    auto operator=(BlockSharedMemDynMember &&) -> BlockSharedMemDynMember & = delete;
+                    //-----------------------------------------------------------------------------
+                    /*virtual*/ ~BlockSharedMemDynMember() = default;
+
+                    uint8_t* dynMemBegin() const {return m_mem.data();}
+
+                    /*! \return the pointer to the begin of data after protion allocated as dynamical shared memory.
+                     *!
+                     *! \tparam TDataAlignBytes Round-up offset to multiple of
+                     *! this number of bytes to maintain alignment. This
+                     *! alignment assumes, that the compiler places the instance
+                     *! at architecture-appropriate boundaries.
+                     */
+                    template<unsigned int TDataAlignBytes = 4>
+                    uint8_t* staticMemBegin() const
+                    {
+                        return m_mem.data() +
+                            (m_dynSize/TDataAlignBytes + (m_dynSize%TDataAlignBytes>0)*TDataAlignBytes);
+                    }
+
+                    /*! \return the remaining capacity for static block shared memory.
+                     *!
+                     *! \tparam TDataAlignBytes Multiple of bytes the static offset was rounded-up to
+                     */
+                    template<unsigned int TDataAlignBytes = 4>
+                    unsigned int staticMemCapacity() const
+                    {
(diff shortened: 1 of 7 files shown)


include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,146 @@
...
+
+                    };
+                }
+                //#############################################################################
+                //! Static block shared memory provider using a pointer to
+                //! externally allocated fixed-size memory, likely provided by
+                //! BlockSharedMemDynMember.
+                template<unsigned int TDataAlignBytes = 4>
+                class BlockSharedMemStMember :
+                    public detail::BlockSharedMemStMemberImpl<TDataAlignBytes>,
```
BenjaminW3, line 107: What is the reason for the inheritance here? Couldn't `BlockSharedMemStMemberImpl` be merged into `BlockSharedMemStMember`?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,146 @@
...
+                            m_allocdBytes = 0u;
+                        }
+
+                    };
+                }
+                //#############################################################################
+                //! Static block shared memory provider using a pointer to
+                //! externally allocated fixed-size memory, likely provided by
+                //! BlockSharedMemDynMember.
+                template<unsigned int TDataAlignBytes = 4>
```
BenjaminW3: Can we use `core::vectorization::defaultAlignment` instead of the hard coded 4 here? (note: `defaultAlignment` is in bits)

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,146 @@
...
+                        TuniqueId,
+                        BlockSharedMemStMember<TDataAlignBytes>>
+                    {
+                        //-----------------------------------------------------------------------------
+                        static auto allocVar(
+                            block::shared::st::BlockSharedMemStMember<TDataAlignBytes> const &smem)
+                        -> T &
+                        {
+                            smem.template alloc<T>();
+                            return smem.template getLatestVar<T>();
```
BenjaminW3, line 132: Is there a reason for the split here? Can't `alloc` directly return the variable as well?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: Forgot to mention this in the description, sorry.

I need to derive from Impl for the OpenMP 5 and OpenACC backends. Deriving from `BlockSharedMemStMember`, which already implements the shared mem concept confuses the traits, because then there are two base classes of the accelerator which derive from an instance of the concept.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: Yes. This is in order for facilitate derived classes which need to synchronize to do the `alloc()`.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,146 @@
...
+                        //-----------------------------------------------------------------------------
+                        /*virtual*/ ~BlockSharedMemStMemberImpl() = default;
+
+                        template <typename T>
+                        void alloc() const
+                        {
+                            uint8_t* buf = &m_mem[m_allocdBytes];
+                            new (buf) T();
+                            m_allocdBytes += alignPitch<T>();
+#if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
```
BenjaminW3, line 60: Is there some way to find out from the compiler if we are compiling for offload? Some macros? builtins?

jkelling: There are not portable OpenACC or OpenMP 5 things. The only way might be to check for things like `__CUDA_ARCH__`.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,146 @@
...
+                        mutable uint8_t* m_mem;
+#ifndef NDEBUG
+                        const unsigned int m_capacity;
+#endif
+
+                        template<typename T>
+                        static constexpr size_t alignPitch()
+                        {
+                            return (sizeof(T)/TDataAlignBytes + (sizeof(T)%TDataAlignBytes>0))*TDataAlignBytes;
+                        }
```
BenjaminW3: Could you please move the private members down? All other classes in alpaka have it this way.

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
```diff
@@ -0,0 +1,119 @@
...
+                    //-----------------------------------------------------------------------------
+                    auto operator=(BlockSharedMemDynMember const &) -> BlockSharedMemDynMember & = delete;
+                    //-----------------------------------------------------------------------------
+                    auto operator=(BlockSharedMemDynMember &&) -> BlockSharedMemDynMember & = delete;
+                    //-----------------------------------------------------------------------------
+                    /*virtual*/ ~BlockSharedMemDynMember() = default;
+
+                    uint8_t* dynMemBegin() const {return m_mem.data();}
+
+                    /*! \return the pointer to the begin of data after protion allocated as dynamical shared memory.
```
BenjaminW3:
```suggestion
                    /*! \return the pointer to the begin of data after portion allocated as dynamical shared memory.
```

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
```diff
@@ -0,0 +1,119 @@
...
+                //! member array to allocate memory on the stack or in shared
+                //! memory.
+                template<unsigned int TStaticAllocKB = ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB>
+                class BlockSharedMemDynMember :
+                    public concepts::Implements<ConceptBlockSharedDyn, BlockSharedMemDynMember<TStaticAllocKB>>
+                {
+                    static constexpr unsigned int staticAllocBytes = TStaticAllocKB<<10;
+
+                    mutable std::array<uint8_t, staticAllocBytes> m_mem;
+                    unsigned int m_dynSize;
```
BenjaminW3: Private members to the bottom here as well, please.

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
```diff
@@ -0,0 +1,119 @@
...
+#pragma once
+
+#include <alpaka/block/shared/dyn/Traits.hpp>
+#include <alpaka/core/Assert.hpp>
+
+#include <type_traits>
+#include <array>
+
+#ifndef ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB
+#define ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KB 30
```
BenjaminW3: kibi-byte or kilo-byte?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
BenjaminW3: Do we have a different portable way of signaling this error instead of an assertion? Exceptions, exit?

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
jkelling: Kibibyte (1024). I guess there is a wrong comment about this somewhere, I will check.

BenjaminW3: May we want to call it `TStaticAllocKiB` and `ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB`. However, the macro version does not make it easier...

jkelling: The capital 'K' already makes it not be the SI 'kilo', but then, it could also be capitalized only because of camel-case in the template param.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: `exit()` should be as forbidden on GPU as `assert()`, also `exit()` would also require some message to be printed, which one may not want to routinely compile into the GPU Code even in debug mode. I am not sure about `throw`, I have the feeling, that this does not work either.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,147 @@
...
+                    };
+                }
+                //#############################################################################
+                //! Static block shared memory provider using a pointer to
+                //! externally allocated fixed-size memory, likely provided by
+                //! BlockSharedMemDynMember.
+                template<unsigned int TDataAlignBytes = core::vectorization::defaultAlignment*8>
```
BenjaminW3: Shouldn't it be  `core::vectorization::defaultAlignment/8`?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
BenjaminW3: I do not fully get this. Who else is deriving from the same concept and why?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: Ups, yes.

jkelling:
Regarding the alignment: Maybe it would be a good idea to add a `__attribute__ ((aligned (core::vectorization::defaultAlignment/8)))` to `BlockSharedMemDynMember`, but I am note sure how to guard this properly since this is a GCC extension, even though it is also supported by other compilers. On the other hand, since this type is large anyway, compilers will usually align it to some sensible boundary automatically....

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: Things in the [OpenMP 5](https://github.com/jkelling/alpaka/blob/blockSharedMemberOpenacc/include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp#L36) and [OpenACC](https://github.com/jkelling/alpaka/blob/blockSharedMemberOpenacc/include/alpaka/ctx/block/CtxBlockOacc.hpp#L53) backends. They need the same small-object allocators (I actually back-ported this code from there), but they need to synchronize threads to call `alloc` and thus need to implement the traits differently.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: `throw` is not allowed in CUDA kernels. Surprisingly, OpenMP 5 allows `throw` in target and parallel regions as long as the exception is caught inside the same region by the same thread. In the best case, not catching it would yield a runtime error and thus terminate (which calls exit and can lead to another issue, see https://github.com/ComputationalRadiationPhysics/picongpu/issues/3241 ), worst case a missing `try{}catch{}` causes a compile time error.

In any case, I would say `throw` is, apart from being not generally save, carrying the wrong semantics: You put `throw` where you might want to catch an error, not to report a bug in user code (here, exceeding the set shared mem limit).

sbastrakov:
I didn't read it all through, so sorry if this is not what you need. But can't the standard C++11 `alignas` be used here?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: `core::vectorization::defaultAlignment`, the smallest alignment defined https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/core/Vectorize.hpp#L51 is 16, which would not make much sense in bits.

jkelling:
I was not aware of `alignas`. This works, thanks.

jkelling:
@BenjaminW3 there are build issues on windows. I cannot access the log, can you?
https://github.com/alpaka-group/alpaka/runs/749665943

sbastrakov:
@jkelling weird, I can see the log. Here is the relevant warning treated as error:
`D:\a\alpaka\alpaka\include\alpaka/block/shared/st/BlockSharedMemStMember.hpp(57,1): warning C4267: '+=': conversion from 'size_t' to 'unsigned int', possible loss of data [D:\a\alpaka\alpaka\build\test\unit\atomic\atomic.vcxproj]`

jkelling:
Thanks. The warning looks legitimate, strange that clang and gcc do not complain. I pushed a fix.

jkelling:
Adding `alignas` causes a pretty pointless warning in msvc which is escalated to an error: " structure was padded due to alignment specifier".
```
D:\a\alpaka\alpaka\include\alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp(90,1): error C2220: the following warning is treated as an error [D:\a\alpaka\alpaka\build\test\unit\time\time.vcxproj]
D:\a\alpaka\alpaka\include\alpaka/acc/AccCpuSerial.hpp(65): message : see reference to class template instantiation 'alpaka::block::shared::dyn::BlockSharedMemDynMember<30>' being compiled [D:\a\alpaka\alpaka\build\test\unit\time\time.vcxproj]
D:\a\alpaka\alpaka\include\alpaka/acc/AccCpuSerial.hpp(131): message : see reference to class template instantiation 'alpaka::acc::AccCpuSerial<TDim,TIdx>' being compiled [D:\a\alpaka\alpaka\build\test\unit\time\time.vcxproj]
D:\a\alpaka\alpaka\include\alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp(90,1): warning C4324: 'alpaka::block::shared::dyn::BlockSharedMemDynMember<30>': structure was padded due to alignment specifier [D:\a\alpaka\alpaka\build\test\unit\time\time.vcxproj]
```
Can we suppress this warning?

BenjaminW3:
Yes, we should suppress it.

jkelling:
How does this work with MSVC?

BenjaminW3:
Please take a look at: https://github.com/alpaka-group/alpaka/blob/f98a5dbe0296d8debe4e336463dc6dff4821d6e2/test/unit/kernel/src/KernelWithHostConstexpr.cpp#L34-L44

jkelling:
The job linux_nvcc-10.1_gcc-6_release failed with: "Error: retrieving gpg key timed out.". Can someone restart this single job?

BenjaminW3:
Github does not yet allow to restart single jobs.
I can merge it even with the one failing build if the rest gets green.

jkelling:
Three small issues:
1. This patch makes `BlockSharedMemStNoSync` obsolete by replacing it in all places where it is used but it does not delete it. I think, if it should be delete, this could happen in a separate PR, however, I do not think it should be deleted for now as there might be use-cases for it in the future as it goes about managing shared memory in a completely different way: By using malloc `BlockSharedMemStNoSync` does not put a limit on the amount of shared memory available. Kernels that rely on it would not work on GPU in the same way, while alpaka is about portability, hence this  PR. Still it might be useful.
2. `BlockSharedMemStNoSync.hpp` is included here: https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/alpaka.hpp#L54 should this be removed?
3. I am not sure why these classes are explicitly included by  `alpaka.hpp`, should includes of `BlockSharedMemDynMember.hpp` and `BlockSharedMemStMember.hpp` added there?

BenjaminW3:
To 1: I would keep it as it can still be used as a base for user accelerators.

2.&3: `alpaka.hpp` is meant as super-header which simply includes all files in alpaka. It should also include your two new files.

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
```diff
@@ -0,0 +1,128 @@
...
+                    template<
+                        typename T,
+                        unsigned int TStaticAllocKiB>
+                    struct GetMem<
+                        T,
+                        BlockSharedMemDynMember<TStaticAllocKiB>>
+                    {
+#if BOOST_COMP_GNUC
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wcast-align" // "cast from 'unsigned char*' to 'unsigned int*' increases required alignment of target type"
```
BenjaminW3, line 102: This warning is not generated without reason. It is undefined if the pointer resulting from this cast can be accessed or not. There are systems out there where you can access variables only when they are aligned correctly for their size (e.g. 4 byte alignment for `int`). On other systems accessing those variables works but is much slower than aligned access.
We are aligning the class (and so the array) with `defaultAlignment` so we can only correctly access types with this alignment requirement correctly.
 In `BlockSharedMemDynBoostAlignedAlloc` we added a check to verify that it is used correctly:
```
static_assert(
    core::vectorization::defaultAlignment >= alignof(T),
    "Unable to get block shared dynamic memory for types with alignment higher than defaultAlignment!");
```
We should probably add this check here.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -0,0 +1,148 @@
...
+                        mutable uint8_t* m_mem;
+#ifndef NDEBUG
+                        const unsigned int m_capacity;
+#endif
+
+                        template<typename T>
+                        static constexpr unsigned int alignPitch()
+                        {
+                            return (static_cast<unsigned int>(sizeof(T))/TDataAlignBytes
+                                + static_cast<unsigned int>(sizeof(T))%TDataAlignBytes>0)*TDataAlignBytes;
```
BenjaminW3: Why are we computing this as `unsigned int`? Shouldn't this be a `ptrdiff_t` or `size_t`?

jkelling: All offsets in the shared memory buffer are unsigned int because GPUs have 32-bit registers (64-bit calculations require more registers) and, at least for  NVIDIA, GPU shared memory has 32-bit banks. This small-object-allocator method is not the most efficient way of managing shared memory and I want to keep the performance impact as small as possible on GPUs. This is the reason the function has to returns unsigned int, not size_t (which is 64-bit on modern GPUs).

Since this is a constexpr function, this computation could also be done with size_t, I just found it to be more consistent to do the computation in 32-bit instead of doing it in 64-bit and casting afterwards.

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
jkelling: On other architectures, like GPUs (with 32-bit wide banks in shared memory) we want to be able to align types to smaller boundaries. I doubt that  `alignof` could capture this even in theory as it cannot know wether we are storing a value in global or shared memory.

jkelling: Apparently I was wrong about the GPU: If I reinterpret_cast a double pointer to shared memory on a Kepler generation GPU which is not 8-byte aligned, reading from (or writing) to it results int wrong values.

In this case I think, we should solve this problem a bit differently: Each allocation should be aligned to the larger of `TDataAlignBytes` and `alignof(T)`.

We need the assert for Dyn and St and I need to change the alignment of the  St section.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: In https://github.com/alpaka-group/alpaka/pull/1033/commits/fa6b79a26dd249fdf91903f220e8b04b356ca3a5 this function is replaced by one named `allocPitch()` which is not constexpr, so that one should compute in 32-bit.

BenjaminW3 (approved):

BenjaminW3:
@jkelling You can merge it when you are ready.

jkelling:
@BenjaminW3 I do not have write access to merge. But it is ready and can be merged.

psychocoderHPC:
Fom CUDA documentation:

```
H.3. Compute Capability 3.x
H.3.1. Architecture

Devices of compute capability 3.7 add an additional 64 KB of shared memory to each
of the above configurations, yielding 112 KB, 96 KB, and 80 KB shared memory per SM,
respectively. However, the maximum shared memory per thread block remains 48 KB.
```

Maybe we should think about increasing the default to 46kib

psychocoderHPC:
@jkelling I am wondering why this PR is providing the shared memory restriction only for dynampic shared memory and not for static shared memory. 
The alpaka concept has two ways of allocating shared memory but not have two different shared memory storage's with two sizes.

psychocoderHPC:
> @jkelling I am wondering why this PR is providing the shared memory restriction only for dynampic shared memory and not for static shared memory.
> The alpaka concept has two ways of allocating shared memory but not have two different shared memory storage's with two sizes.

Ahh I see now, static shared memory is using the left over of the dynamic shared memory pool: https://github.com/alpaka-group/alpaka/pull/1033/files#diff-3337e4cfdef9f4f15dafba232a45d250R115

Issue #1037, opened 2020-06-09
doxygen build failes


SimeonEhrig:
# Problem

The GitHub `gh-pages` is failing, which updates the doxygen documentation. It was mentioned in #1032. I think the reason is the new Sphinx Doc documentation.

# Background

At the moment, the Sphinx Doc and the doxygen documentation are built and published separately.
- **Sphinx Doc**: is build by the readthedocs builder and published on https://alpaka.readthedocs.io
- **Doxygen API**: is build by a GitHub action job and published on https://alpaka-group.github.io

# Reason

The doxygen api should be integrated or linked in the sphinx doc. During the integration/linking process something was going wrong an the process was not finished. I believe, it was planed to integrate the doxygen build in the sphinx doc but at the moment, the sphinx doc links to https://alpaka-group.github.io. The problem is, that the output type of doxygen is changed from `html` to `xml`, which I think can be better integrated in the Sphinx doc. But the GitHub workflow want to push the `html` output, so it fails because there is not `html` support.

# Solutions

1. **Finish the integration in Sphinx Doc:** Use the .xml files generated by the readthedocs, change the doxygen link from https://alpaka-group.github.io to a internal link (https://alpaka.readthedocs.io) and delete the GitHub workflow. [1]
2. **Keep the GitHub workflow:** Change the doxygen output back to `html`, keep the link in the sphinx doc and the delete everything in the sphinx doc build recipe, which is related to the doxygen build.

[1] A possible solution was posted by @ax3l : https://github.com/alpaka-group/alpaka/pull/990#issuecomment-636128010

@psychocoderHPC @BenjaminW3 @sbastrakov 
What do you think, is the better solution?

sbastrakov:
Thanks for investigating and providing the options @SimeonEhrig ! I do not have a particular preference, so perhaps would prefer the way that gets this issue resolved or worked around quickly. Currently because of this issue CI considers all branches failing.

BenjaminW3:
I would prefer 2. because it will fix the immediate problem. We can still enhance the sphinx doc later.

SimeonEhrig:
> I would prefer 2. because it will fix the immediate problem. We can still enhance the sphinx doc later.

Okay, I use this solution.

[...]

PR #1038, opened 2020-06-10
Fix Doxygen CI build

SimeonEhrig:
solving issue #1037

### DIFF
--- a/docs/Doxyfile
+++ b/docs/Doxyfile
@@ -1172,7 +1172,7 @@ IGNORE_PREFIX          =
 # If the GENERATE_HTML tag is set to YES, doxygen will generate HTML output
 # The default value is: YES.
 
-GENERATE_HTML          = NO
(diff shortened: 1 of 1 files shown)


SimeonEhrig:
@BenjaminW3 
I think I've found a solution. It should be no problem to build html and xml at the same time. Unfortunately the GitHub Action documentation workflow is just running, when it is merged in the develop brunch. So I can't test it online before it is merged (on my offline system, it works).
Is it possible to make the workflow more fine granular, without a lot of works, that the build is running every time and just the push is restricted to the `develop` branch?
Otherwise we have to merge and hope that it works.

BenjaminW3:
I am fine with some hope. It can not get more broken than it is right now.

SimeonEhrig:
I think, it is ready for merge. The CI failures are not related to my change.

PR #1039, opened 2020-06-11
fix the doxygen ci build job Nr. 2

SimeonEhrig:
The PR #1038 doesn't fixed the [problem](https://github.com/alpaka-group/alpaka/runs/760770619).

### DIFF
--- a/script/push_doc.sh
+++ b/script/push_doc.sh
@@ -19,7 +19,7 @@ cd docs/doxygen/html
 git config --global user.email "action@github.com"
 git config --global user.name "GitHub Action"
 
-git add .
+git add --force .
 
 git log -n 3
 


SimeonEhrig:
@BenjaminW3 
I could run the doxygen job in my private repo: https://github.com/SimeonEhrig/alpaka/runs/760845738
I tried to fix the problem, that `git add /docs/doxygen/html` was forbidden by the `.gitignore`. The test ran for a while, but crashed at adding files without an error message. Maybe it didn't catch the error message for `git push`, which can potentially fail because of missing permission. Do you have an idea?

.gitignore
```diff
@@ -1,4 +1,7 @@
 /docs/doxygen/*
+# the GitHub Actions documentation job
+# needs to commit the html output
+!/docs/doxygen/html
```
BenjaminW3: I think the only reason to have the gitignore for `/docs/doxygen/*` above was to exclude the generated doxygen documentation from being pushed. If we now excplicitly include the generated html files we could also directly remove the exclusion.

BenjaminW3:
It should be possible to `git add --force` the files even though they are hidden by `.gitignore`.

.gitignore
SimeonEhrig: Yes, it was the idea of the exclude. But it also excludes the `xml` output. However, I changed to your approach. At the first time, when I chose a solution, I thought modifying the `.gitignore` is the better solution than using the `--force` argument. But you are right, you approach is better suited for normal developer.

SimeonEhrig:
I tested the new solution in my private repository and the GitHub workflow crashed with the same (nondescripting) error message, like the old approach.

SimeonEhrig:
@BenjaminW3 
If you fine with the solution, can you please merge.

PR #1064, opened 2020-07-03
fix CUDA/HIP accelerator concept usage

psychocoderHPC:
`AccGpuUniformCudaHipRt` was not defined to follow the concept
`ConceptAcc`. Since we do not disallow the direct usage of this
accelerator the concept should be set.

This PR fixes compile time issue in #1024. It was not possible to derive the queue type from the accelerator `AccGpuUniformCudaHipRt`.


sbastrakov (commented):
I am not against the change. But I thought we did disallow using this accelerator directly.

BenjaminW3 (approved):

psychocoderHPC:
> I am not against the change. But I thought we did disallow using this accelerator directly.

We do not suggest the usage on the workshop. Currently the unified backend is directly tested and exposed in the same namespace hierarchy to the user.

Issue #995, opened 2020-05-06
std::decay_t causes internal compiler errors in Intel compiler


krzikalla:
In TaskKernelCpuOmp2Blocks.hpp the lines 102 and 122 read

`[this](typename std::decay<TArgs>::type const & ... args)`

The std::decay in these lines produces an internal compiler in all Intel.2018 versions. 
In alpaka 0.3.5 these lines (there in ExecCpuOmp2Blocks.hpp) just read

`[this](TArgs const & ... args)`

Unfortunatly we can't switch to Intel.2019 or a newer version yet (since this causes other internal compiler errors in our code). However, if I just remove the std::decay in alpaka 0.4.1 I get a bunch of compilation errors. What is this std::decay for? If I want to remove it, what other changes do I need to make?

This is somehow related to #843

psychocoderHPC:
For documentation I like to link @ax3l 's comment that it is now possible to add `icc` to travis https://github.com/alpaka-group/alpaka/issues/493#issuecomment-622222393

psychocoderHPC:
> What is this std::decay for?

You can best see it in the [cppreference documentation](https://en.cppreference.com/w/cpp/types/decay). It removes references, universal references and is doing some pointer transformations.

As you wrote I do not think that you can simply remove `std::decay`. 

> If I want to remove it, what other changes do I need to make?

Depends on the compiler errors.

psychocoderHPC:
@krzikalla Is it possible to compile all alpaka tests with icc 2018?

sbastrakov:
I do not remember all details, but as far as I remember that `decay` is indeed needed, we discussed with @psychocoderHPC at that time.

krzikalla:
@psychocoderHPC: 
CMake didn't complete ("Could NOT find alpaka (missing: alpaka_INCLUDE_DIR)")
To reformulate my initial question in präziserem Deutsch: Warum steht std::decay in alpaka an diesen Stellen? Wofür wurde es damals dort hingeschrieben?
@sbastrakov :
It was introduced as part of a patch, which can be traced back to #843. But why?

I mean, you should by no means change it - Intel 2018 is outdated. However maybe I find a hacky workaround for our problem.

psychocoderHPC:
> CMake didn't complete ("Could NOT find alpaka (missing: alpaka_INCLUDE_DIR)")

Remove all temporary cmake files. Than set the environment variable `alpaka_ROOT`. I am not sure this happend.

[...]

PR #1066, opened 2020-07-08
CI: Add Latest Intel Compiler

ax3l:
Add latest `icpc`/`icc` from oneAPI for Linux to CI.

Related to #995

Since it's currently plain impossible to install boost with icc on Linux, we install the header-only variant of boost with GCC and use that (no Boost.Fibers coverage in CI).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,7 +217,7 @@ jobs:
         # icpc
         - name: linux_icpc_release
           os: ubuntu-latest
-          env: {TRAVIS_OS_NAME: linux,   CXX: icpc,    CC: icc, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.15.7, OMP_NUM_THREADS: 2, ALPAKA_CI_STDLIB: libstdc++, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04"}
+          env: {TRAVIS_OS_NAME: linux,   CXX: icpc,    CC: icc, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_CI_CMAKE_VER: 3.15.7, OMP_NUM_THREADS: 2, ALPAKA_CI_STDLIB: libstdc++, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04"}
 
         ## CUDA 9.0
         # nvcc + g++
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,6 +214,11 @@ jobs:
           os: ubuntu-latest
           env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 10,     ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.15.7, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_ACC_CPU_BT_OMP4_ENABLE: OFF}
 
+        # icpc
+        - name: linux_icpc_release
+          os: ubuntu-latest
+          env: {TRAVIS_OS_NAME: linux,   CXX: icpc,    CC: icc, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.66.0, ALPAKA_CI_CMAKE_VER: 3.15.7, OMP_NUM_THREADS: 2, ALPAKA_CI_STDLIB: libstdc++, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04"}
+
         ## CUDA 9.0
         # nvcc + g++
         - name: linux_nvcc-9.0_gcc-5_debug
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,7 +217,7 @@ jobs:
         # icpc
         - name: linux_icpc_release
           os: ubuntu-latest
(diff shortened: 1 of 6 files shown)


ax3l:
Come on Boost, urgh...
```
+ sudo ./bootstrap.sh --with-toolset=intel-linux
Building Boost.Build engine with toolset intel-linux... 
Failed to build Boost.Build build engine
Consult 'bootstrap.log' for more details
```
(the outdated syntax `--with-toolset=intel` gives the same error)

ax3l:
```
which "${CC}"
++ which icc
${CC} -v
/opt/intel/inteloneapi/compiler/latest/linux/bin/intel64/icc
++ icc -v
icc version 2021.1 Beta (gcc version 7.5.0 compatibility)
...
icc -o bootstrap/jam0 command.c compile.c constants.c debug.c execcmd.c frames.c function.c glob.c hash.c hdrmacro.c headers.c jam.c jambase.c jamgram.c lists.c make.c make1.c object.c option.c output.c parse.c pathsys.c regexp.c rules.c scan.c search.c subst.c timestamp.c variable.c modules.c strings.c filesys.c builtins.c class.c cwd.c native.c md5.c w32_getreg.c modules/set.c modules/path.c modules/regex.c modules/property-set.c modules/sequence.c modules/order.c execunix.c fileunix.c pathunix.c
./build.sh: 17: ./build.sh: icc: not found
```

script/install_icpc.sh
```diff
@@ -0,0 +1,39 @@
...
+travis_retry sudo apt-get install -y intel-oneapi-icc
+
+set +eu
+source /opt/intel/inteloneapi/setvars.sh
+set -eu
+
+which "${CXX}"
+${CXX} -v
+which "${CC}"
+${CC} -v
```
BenjaminW3, line 39: Could you try to add those 4 lines in front of the `./bootstrap.sh --with-toolset="intel-linux"`.
Maybe some environment vars get lost in between?

ax3l: I think so, yep.

BenjaminW3: It may be necessary that you have to call `source /opt/intel/inteloneapi/setvars.sh` in `alpaka/script/run.sh` again.
I have not yet removed the split between the docker image creation (start, run install.sh, stop, save as image) and then building and executing the tests (start image, run.sh to build and test). The environment is not correctly preserved in between those steps.
I already started getting rid of this two step setup because we are not caching the docker images anymore with github actions.

ax3l: Looks like the `icc` is there. something funky in bootstrap of bjam is going on.

ax3l:
The boost build system is archaic:
https://github.com/boostorg/build/issues/475

script/install_icpc.sh
ax3l: Yep, the boost build system is archaic:
https://github.com/boostorg/build/issues/475
https://github.com/boostorg/build/pull/635
```
A C++11 capable compiler is required for building the B2 engine.
Toolset 'intel-linux' does not appear to support C++11
```
This does not even make sense, since `icc` already defaults to C++14... The compile feature check is super outdated.

I wonder how far the CMake adoption process in Boost is in 2020...
Update: aha
```
CMake Error at tools/cmake/include/BoostRoot.cmake:13 (message):
  CMake support in Boost is experimental and part of an ongoing development
  effort.  It's not ready for use yet.  Please use b2 (Boost.Build) to build
  and install Boost.
```

ax3l:
@krzikalla how do you install boost with the Intel compilers on Linux?

ax3l:
@BenjaminW3 I am now successfully installing Boost with GCC for this ICC build.
But we might have to preserve the activated compiler environment better or source `source /opt/intel/oneapi/setvars.sh` again in the build.

.github/workflows/ci.yml
```diff
@@ -214,6 +214,11 @@ jobs:
           os: ubuntu-latest
           env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 10,     ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.15.7, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_ACC_CPU_BT_OMP4_ENABLE: OFF}
 
+        # icpc
+        - name: linux_icpc_release
+          os: ubuntu-latest
+          env: {TRAVIS_OS_NAME: linux,   CXX: icpc,    CC: icc, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_CI_CMAKE_VER: 3.15.7, OMP_NUM_THREADS: 2, ALPAKA_CI_STDLIB: libstdc++, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04"}
```
BenjaminW3, line 220: `TRAVIS_OS_NAME` is not necessary anymore.
It has been renamed to `ALPAKA_CI_OS_NAME` and is now automatically added since #1089

BenjaminW3: The order of env variables has been changed and `ALPAKA_CI_STDLIB` has been moved to behind `CC`.

script/ci.sh
```diff
@@ -17,7 +17,7 @@ source ./script/before_install.sh
 
 if [ "$ALPAKA_CI_OS_NAME" = "Linux" ]
 then
-  ./script/docker_install.sh
+  source ./script/docker_install.sh
```
BenjaminW3, line 20: I do not think that this works ;-)

BenjaminW3: With #1109 install.sh and run.sh will be run within a single invocation of `docker run` which should make it much easier to keep settings between the invocations.

script/install_icpc.sh
```diff
@@ -0,0 +1,39 @@
...
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+source ./script/travis_retry.sh
+
+source ./script/set.sh
+
+: "${CXX?'CXX must be specified'}"
```
BenjaminW3, line 17: You also require `CC`.

.github/workflows/ci.yml
BenjaminW3, line 220: May we have to disable the fibers backend because we would be mixing gcc built boost libs with an intel compiler built executable?

ax3l:
Didn't I disabled that already? Please feel free to push on the PR's branch, I am off now for two weeks. 🌴

BenjaminW3:
Merged via #1235

PR #1067, opened 2020-07-08
Simplify parameter deduction in TaskKernel... implementations with auto

sbastrakov:
This is only done when the parameters were already of the correct (decayed) types, as stored in the `m_args` tuple. So in some cases I did not make the change, and believe it would have been wrong to do so.
This also works around Intel internal compiler errors when using `std::decay` on empty template parameter packs #995

### DIFF
--- a/include/alpaka/kernel/TaskKernelCpuFibers.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuFibers.hpp
@@ -122,7 +122,7 @@ namespace alpaka
                 // Get the size of the block shared dynamic memory.
                 auto const blockSharedMemDynSizeBytes(
                     meta::apply(
-                        [&](std::decay_t<TArgs> const & ... args)
+                        [&](auto const & ... args)
                         {
                             return
                                 kernel::getBlockSharedMemDynSizeBytes<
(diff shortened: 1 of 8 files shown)


psychocoderHPC (dismissed):

sbastrakov:
Perhaps we should wait for #1066 to be merged first, and rebase this PR against it to test the change on Intel. On Hemera this prevents the internal compiler error with empty parameter pack.

BenjaminW3 (dismissed):

BenjaminW3:
nvcc does not seem to like this.

sbastrakov:
Yes @BenjaminW3 we spent some time with @psychocoderHPC figuring it out, I will continue tomorrow and fixup this PR.

ax3l:
Feel free to merge your PR first, we can rebase #1066 afterwards. Failing CI in `develop` is always noisy for all other PRs.

sbastrakov:
I did not make much progress. So far I found out that somehow with nvcc / hip the size of derived arguments pack is indeed different due to this PR's change compared to before it, and that in turn causes the displayed error. However, I did not get a good idea yet why does this happen.

sbastrakov:
After investigating the nvcc issues introduced by these changes, I feel rather stuck. So I will close this PR for now. And a much simpler fix for #995 (which was supposed to be a subproduct of this PR) is provided by #1074 .

PR #1070, opened 2020-07-10
Redefine BOOST_COMP_INTEL the same way BOOST_COMP_NVCC is handled

### LINKED ISSUE #1069: Checks for the Intel compilers in the code are incorrect
Alpaka uses `#if BOOST_COMP_INTEL` in a number of places to check for the Intel compiler. However, it does not introduce a redefinition for it, like with `BOOST_COMP_CUDA` and so it is currently the variable directly from Boost.Predef. With that variable, the check should rather be something like `#if defined(BOOST_COMP_INTEL) && defined(BOOST_COMP_INTEL_EMULATED)`. Or alternatively add an alpaka redefinition that takes care of all that, like it is done for CUDA, etc. I will provide a PR with alpaka redefinition.

sbastrakov:
Fixes #1069

### DIFF
--- a/include/alpaka/core/BoostPredef.hpp
+++ b/include/alpaka/core/BoostPredef.hpp
@@ -119,3 +119,11 @@
 #else
     #define BOOST_COMP_CLANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
 #endif
+
+//-----------------------------------------------------------------------------
+// Intel compiler detection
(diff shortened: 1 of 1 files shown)


sbastrakov:
Note: this change is done consistently with the existing code to handle such stuff, and fixes the existing wrong detection of Intel. However, I suggest changing the general way as well, discussed in #1072

BenjaminW3 (approved):

psychocoderHPC (approved):

PR #1071, opened 2020-07-10
fix installation

psychocoderHPC:
alpaka should not be installed if it is used as CMakes `add_subdirectory()`.
The reason is that alpaka is header only and the headers will not be
used after the master project used alpaka for compilation.

note: I labeled this PR as BUG because the installation of not needed files is an unexpected behavior.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -141,38 +141,42 @@ endif()
 
 ################################################################################
 # Installation.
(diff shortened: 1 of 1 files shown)


j-stephan (approved):
Looks good to me. If the CI agrees this can be merged.

sbastrakov (approved):

PR #1074, opened 2020-07-10
 Work around Intel ICE when using std::decay on empty template parameter packs

### LINKED ISSUE #995: std::decay_t causes internal compiler errors in Intel compiler

sbastrakov:
Fixes #995.

This is a simple workaround to get it fixed. My earlier PR #1067 attempted a more general change that gets rid of these `decay`s along the way. However, that one has its own issues, that will take time to resolve. So I will close that one for now.

To do:
- [x] merge #1070, this PR uses a commit from there.

Thanks to @krzikalla for communicating the way to fix.

### DIFF
--- a/include/alpaka/core/BoostPredef.hpp
+++ b/include/alpaka/core/BoostPredef.hpp
@@ -119,3 +119,11 @@
 #else
     #define BOOST_COMP_CLANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
 #endif
+
+//-----------------------------------------------------------------------------
+// Intel compiler detection
+// BOOST_COMP_INTEL_EMULATED is defined by boost instead of BOOST_COMP_INTEL
+#if defined(BOOST_COMP_INTEL) && defined(BOOST_COMP_INTEL_EMULATED)
+    #undef BOOST_COMP_INTEL
+    #define BOOST_COMP_INTEL BOOST_COMP_INTEL_EMULATED
+#endif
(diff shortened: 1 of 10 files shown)


psychocoderHPC (approved):

sbastrakov:
Okay, so #1070 is now merged. And one of commits of this PR is the same as there. I assume it should be all fine right now, without rebasing?

BenjaminW3:
Should be fine. I set `Rebase and merge` as default merge behaviour. This will cleanly get rid of the commit.

psychocoderHPC:
Cc-ing: @jkelling This is important for SPEC

sbastrakov:
Also perhaps the previously wrong Intel compiler identification, fixed in #1070, may be useful for @jkelling .

PR #1077, opened 2020-07-12
Allow to filter specific jobs based on commit message

BenjaminW3:
By adding the word `ci_filter` to the last commit of a branch, only jobs
that are listed in the commit message via their name will be executed.

ci_filter: [linux_nvcc-10.2_gcc-5_debug, linux_clang-7_release_c++17, windows_nvcc-10.0_cl-2017_debug, macos_xcode-11.1_release]

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -491,10 +491,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+      if: (!contains(github.event.head_commit.message, 'ci_filter') || contains(github.event.head_commit.message, matrix.name ))
     - name: build + test
-      if: runner.os == 'Windows'
+      if: (!contains(github.event.head_commit.message, 'ci_filter') || contains(github.event.head_commit.message, matrix.name )) && (runner.os == 'Windows')
       shell: bash
       run: cd ${GITHUB_WORKSPACE} && ./script/ci.sh
     - name: build + test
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -491,10 +491,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+      if: (!contains(github.event.head_commit.message, 'ci_filter') || contains(github.event.head_commit.message, matrix.name ))
```
BenjaminW3, line 494: I tried to add this `if` outside of the `steps` so that it would be required only once but the `matrix.name` does not seem to be available there.

BenjaminW3:
@ax3l This would make it much easier to run your PR #1066 only for the new compiler as long as it is work-in-progress.

BenjaminW3:
You can see the result of this here: https://github.com/BenjaminW3/alpaka/actions/runs/166441351
In PRs the yaml of the target branch is used so those changes do not have an effect on the PR build but only on the branch build in my fork.

sbastrakov (approved):

psychocoderHPC (approved):

SimeonEhrig:
This should be documented in the developer section. It's a really nice feature, but I think, new developers will not find this PR and the feature.

sbastrakov:
This is a good point @SimeonEhrig ! Given that we recently struggled quite a lot with CI speed, this feature is very useful and should be documented.

PR #1078, opened 2020-07-12
Remove duplication of names from github actions matrix

BenjaminW3:
Include section does not need to match matrix in any way anymore.
See: https://github.blog/changelog/2020-04-15-github-actions-new-workflow-features/

PR #1079, opened 2020-07-13
Tidy up the heat equation example

sbastrakov:
Add correction check as return code, as opposed to just printing earlier. Minor style fixes. Update the header which was missed in the original PR #978 .

### DIFF
--- a/example/heatEquation/src/heatEquation.cpp
+++ b/example/heatEquation/src/heatEquation.cpp
@@ -1,4 +1,5 @@
-/* Copyright 2019 Benjamin Worpitz, Matthias Werner
+/* Copyright 2020 Benjamin Worpitz, Matthias Werner, Jakob Krude,
+ *                Sergei Bastrakov
  *
  * This file exemplifies usage of alpaka.
  *
@@ -21,6 +22,7 @@
 #include <algorithm>
 #include <cmath>
 #include <iostream>
+#include <utility>
 
 
 //#############################################################################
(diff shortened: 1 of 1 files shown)


example/heatEquation/src/heatEquation.cpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2019 Benjamin Worpitz, Matthias Werner
+/* Copyright 2020 Benjamin Worpitz, Matthias Werner, Jakob Krude,
+ *                Sergei Bastrakov
```
sbastrakov, line 2: Note: this change of the header was just forgotten in the original PR. This example was developed by @jkrude and me, based on the existing ones.

sbastrakov:
@BenjaminW3 in my last commit 2017bfe6d32cdaab7d86bea466d5dc9bee34f87f I tried to set up filters, to investigate only the 3 CI combinations originally failing. But it seems all combinations were ran again. Could you take a look if I did something wrong?

BenjaminW3:
Maybe #1078 broke this feature. I will have to investigate...

BenjaminW3 (approved):

BenjaminW3:
@sbastrakov I took a look at the `ci_filter` feature again and it seems to work.
The confusing part is that skipped jobs are still shown and they are shown as successful.
I can not make them disappear, but I can make them show up as failed.

sbastrakov:
@BenjaminW3 thanks for looking and explaining. I think in this case it is less confusing to mark skipped as successful than failed (like done now).

BenjaminW3:
The main reason for making them fail now is because else this would allow to make arbitrary PRs green and mergable even when they still have issues. It is not easy for reviewers to see if there are skipped jobs.

Issue #21, opened 2014-12-22
extend device properties


BenjaminW3:
SharedMemorySizeBytes
MaxClockFrequencyHz

Maybe also CoresPerMultiProcessor. This should be the number of cores with fast shared memory. On a two Processor node using OpenMP this should be the number of cores on one of the processors ( x const).
Do we really need this? BlockKernelsCountMax delivers the maximum block size which is more relevant.

sbastrakov:
@krzikalla just mentioned the shared memory size property via email. I guess this should be rather straightforward to implement. If there are no objections to have it added, I could provide a PR in a couple of days.

PR #1084, opened 2020-07-17
Extend AccDevProps with shared memory size per block

sbastrakov:
Implements [this comment](https://github.com/alpaka-group/alpaka/issues/21#issuecomment-659234862). Since this is a property of a particular device, not accelerator, I believe a device trait is a better way to implement it than `AccDevProps`. Perhaps the contents of that struct should be also moved as device traits.

For CPUs perhaps a proper way to implement would be using memory available per NUMA node as discussed in #21. But for now I only made a stub that returns the overall memory size. CUDA / HIP access it properly via device properties.

cc @krzikalla

### DIFF
--- a/include/alpaka/acc/AccCpuFibers.hpp
+++ b/include/alpaka/acc/AccCpuFibers.hpp
@@ -172,8 +172,6 @@ namespace alpaka
                     dev::DevCpu const & dev)
                 -> alpaka::acc::AccDevProps<TDim, TIdx>
                 {
-                    alpaka::ignore_unused(dev);
-
 #ifdef ALPAKA_CI
                     auto const blockThreadCountMax(static_cast<TIdx>(3));
 #else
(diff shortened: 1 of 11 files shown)


include/alpaka/dev/DevCpu.hpp
```diff
@@ -218,6 +218,26 @@ namespace alpaka
...
+                dev::DevCpu>
+            {
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_HOST static auto getSharedMemBytes(
+                    dev::DevCpu const & dev)
+                -> std::size_t
+                {
+                    alpaka::ignore_unused(dev);
+
+                    return dev::cpu::detail::getTotalGlobalMemSizeBytes();
```
psychocoderHPC: Only the accelerator knows the amount of available shared memory., so it should not be a device trait.
~~After looking again into PR #1033 and the help of @sbastrakov we saw that this fixed size shared memory is only used for dynamic shared memory.~~

~~This will now race the question what does this trait should give back. IMO the alpaka concept has only one shared memory which can be used via static and dynamic allocations.~~

psychocoderHPC (changes requested):

BenjaminW3 (dismissed):

include/alpaka/dev/DevCpu.hpp
sbastrakov: Agree, will modify the PR to accomplish it.

sbastrakov: Fixed and rebased.

psychocoderHPC (approved):

psychocoderHPC:
I will merge it even if one test did not start, this is a CI issue we need to keep an eye on.

PR #1086, opened 2020-07-18
Add CUDA 11 support for Windows

BenjaminW3:


BenjaminW3:
The CUDA-only build fails and I am not yet sure why.
Compile errors look like this: `D:\a\alpaka\alpaka\test\unit\acc\src\AccNameTest.cpp(19): note: type is 'TestInit'`

psychocoderHPC:
> The CUDA-only build fails and I am not yet sure why.
> Compile errors look like this: `D:\a\alpaka\alpaka\test\unit\acc\src\AccNameTest.cpp(19): note: type is 'TestInit'`

IMO the problem is that nvcc is pulling a host lambda into the device scope [here](https://github.com/alpaka-group/alpaka/blob/47f80e9493d6b9fcc2108478e0c6c894906a3481/thirdParty/catch2/include/catch2/catch.hpp#L1131-L1134) and/or [here](

The CUDA-only build fails and I am not yet sure why.
Compile errors look like this: D:\a\alpaka\alpaka\test\unit\acc\src\AccNameTest.cpp(19): note: type is 'TestInit'
)

We can try to  guard this section with `#if !defined(__CUDA_ARCH__)` so that the lambda is empty on the device. Time to use you new filter function for the CI.

BenjaminW3:
I tried to replace some code within catch2 depending on `#if defined(__CUDA_ARCH__)` but I was not successful.
Feel free to take this branch and experiment yourself.

BenjaminW3:
Looks good now. Build failure is unrelated. Please approve/merge!

sbastrakov (approved):
Looks good.
I can't merge when CI failed.

PR #1090, opened 2020-07-20
fix warp functions

psychocoderHPC:
nvcc is not able to detect the return value for the warp function implementations.

The error is similar to #1085. Due to the workarounds within the implementations and the precompiler code path selection for CUDA and HIP a simple `decltype` to much code. I looked into the API and used the specified return value for the corresponding function calls.

### DIFF
--- a/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
@@ -59,6 +59,7 @@ namespace alpaka
                 //-----------------------------------------------------------------------------
                 __device__ static auto getSize(
                     warp::WarpUniformCudaHipBuiltIn const & /*warp*/)
+                    -> std::int32_t
                 {
                     return warpSize;
                 }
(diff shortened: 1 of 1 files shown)


sbastrakov (approved):

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -142,6 +150,12 @@ namespace alpaka
                 __device__ static auto ballot(
                     warp::WarpUniformCudaHipBuiltIn const & warp,
                     std::int32_t predicate)
+                    // return type is required by the compiler
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    -> unsigned
+#else
+                    -> uint64_t
+#endif
```
BenjaminW3, line 158: May we want to guard those with some `#if BOOST_COMP_NVCC`?
This would also document the reason whythose are necessary.

BenjaminW3 (approved):

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
psychocoderHPC: IMO we can keep it for clang too, in #1092 it will than use always fixed size types without any side effect. This will guarantee that fixed size types always work.

PR #1092, opened 2020-07-20
use fixed size integral types

psychocoderHPC:
alpaka is mixing fixed size integral types e.g. `unit32_t`, `uint64_t` in the interfaces with build in types `unsigned long`.
This will produce incompatibilities between compiler or platform.
An interfaces specialized for `unsigned long long` will not fit for `uin64_t` if it is defined as `unsigned long` by the compiler.

This PR is switching to fixed size types so that the user knows exactly which type he can use independent of the compiler/OS.
If a native function call requires `unsigned long` or `unsigned long long` and we pass a fixed size type e.g. `uint64_t` we as developer must cast internally. This will move the pain to us as developer instead of the user.

I run into issue with `intrinsics::popcount(acc, warp::activemask(acc))` because active mask was returning `unsigned long` but some interfaces required `unsigned long long`. The type of parameters was different for CPU and GPU implementations.

follow up of: #1090

- [x] the first commit is #1090

### DIFF
--- a/include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
@@ -56,26 +56,26 @@ namespace alpaka
                 //-----------------------------------------------------------------------------
                 __device__ static auto popcount(
                     intrinsic::IntrinsicUniformCudaHipBuiltIn const & /*intrinsic*/,
-                    unsigned int value)
-                -> int
+                    std::uint32_t value)
+                -> std::int32_t
                 {
-#if BOOST_COMP_CLANG
+#if BOOST_COMP_CLANG && BOOST_LANG_CUDA
                     return __popc(static_cast<int>(value));
 #else
-                    return __popc(value);
+                    return __popc(static_cast<unsigned int>(value));
 #endif
                 }
 
                 //-----------------------------------------------------------------------------
                 __device__ static auto popcount(
                     intrinsic::IntrinsicUniformCudaHipBuiltIn const & /*intrinsic*/,
-                    unsigned long long value)
-                -> int
+                    std::uint64_t value)
+                -> std::int32_t
                 {
-#if BOOST_COMP_CLANG
+#if BOOST_COMP_CLANG && BOOST_LANG_CUDA
                     return __popcll(static_cast<long long>(value));
 #else
-                    return __popcll(value);
+                    return __popcll(static_cast<unsigned long long>(value));
 #endif
                 }
             };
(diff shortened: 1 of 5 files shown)


include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
```diff
@@ -56,27 +56,19 @@ namespace alpaka
...
-                -> int
+                    std::uint32_t value)
+                -> std::int32_t
                 {
-#if BOOST_COMP_CLANG
-                    return __popc(static_cast<int>(value));
-#else
-                    return __popc(value);
-#endif
+                    return __popc(static_cast<unsigned int>(value));
```
sbastrakov, line 65: I think here the `#if` guards should stay. The issue is that clang defines a different interface of `__popc` taking a signed int, while all other compilers have unsigned. Same for the other version of popcount.

sbastrakov (changes requested):

include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
psychocoderHPC: HIP and CUDA define the function with unsigned int.

CUDA is unsigned int: https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__INTRINSIC__INT.html#group__CUDA__MATH__INTRINSIC__INT_1g43c9c7d2b9ebf202ff1ef5769989be46
HIP is unsigned int: https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-GUIDE.html

psychocoderHPC: ohh no you are right clang cuda define it as `int` http://clang.llvm.org/doxygen/____clang__cuda__device__functions_8h.html#a2d0eb6e60af727a007e0086d27718517
Than I need to update the guard because hip-clang should use `unsigned`

include/alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp
```diff
@@ -56,26 +56,26 @@ namespace alpaka
...
-                -> int
+                    std::uint64_t value)
+                -> std::int32_t
                 {
-#if BOOST_COMP_CLANG
+#if BOOST_COMP_CLANG && BOOST_LANG_CUDA
                     return __popcll(static_cast<long long>(value));
 #else
-                    return __popcll(value);
+                    return return __popcll(static_cast<unsigned long long>(value));
```
psychocoderHPC: the fast way to return a value 'return return' ^^

psychocoderHPC:
On CI test is not starting but since I pushed all 19h ago I assume this test will not start. IMO we can merge it.

sbastrakov (approved):

sbastrakov:
I agree with @psychocoderHPC . But can't press the button. Please do @BenjaminW3 @psychocoderHPC

PR #1093, opened 2020-07-20
Restructure Sphinx Doc

SimeonEhrig:
The current documentation is not well suited for a Sphinx Doc. The refactoring includes two steps:

1. change the structure
  - introduce new sections: basic, advanced, extra info and development
  - Reorder existing documentation and delete documentation that is no longer relevant
2. change the content of the sections   

# TODO to merge

This points have to do, that the PR is ready for review. The points handles inconsistencies in the documentation and general structure problems.

- [x] Mapping onto specific Hardware: Split CUDA GPUs in General GPUs and CUDA backend section
- [x] Similar Projects
- [x] update cheatsheet build docu and script

# TODO for future PRs

This points should be done in future PRs.

- Abstractions section 
  - The section has to much text and is outdated.
- Library Interface
  - Interface usage should be overworked with the slides from the Alpaka CERN workshop
- Rationale
  - overwork formatting of Implementation Variants
- Backends
  - improve Accelerator Implementation table

ci_filter: [ ]


psychocoderHPC:
for this refactoring the last commit should always contain a CI filter with an empty list. see https://github.com/alpaka-group/alpaka/pull/1077

SimeonEhrig:
> for this refactoring the last commit should always contain a CI filter with an empty list. see #1077

@psychocoderHPC 
ci_filter: [] did not work, maybe ci_filter: [ ]
Can you please cancel the CI

psychocoderHPC:
> > for this refactoring the last commit should always contain a CI filter with an empty list. see #1077
> 
> @psychocoderHPC
> ci_filter: [] did not work, maybe ci_filter: [ ]
> Can you please cancel the CI

You need to add it to the last commit message of the PR not to the github description.

SimeonEhrig:
> > > for this refactoring the last commit should always contain a CI filter with an empty list. see #1077
> > 
> > 
> > @psychocoderHPC
> > ci_filter: [] did not work, maybe ci_filter: [ ]
> > Can you please cancel the CI
> 
> You need to add it to the last commit message of the PR not to the github description.

What did I wrong now?

SimeonEhrig:
@alpaka-group/alpaka-developers 
The restructuring of Sphinx Doc is completed. Most of the work was to change the order of the files, cut and paste, remove some text and change some sentences. An exception is the `Introduction` page, which was completely overworked. Please check the grammar, the spelling (I know I'm not good at it ;-) ) and if a section has lost its correct meaning due to its new position.

SimeonEhrig:
Now, alpaka renders the readthedocs side for each PR, see last entry in the CI list.

For this PR, the link is: https://alpaka--1093.org.readthedocs.build/en/1093/

docs/source/advanced/mapping.rst
```diff
@@ -1,5 +1,24 @@
+Mapping onto Specific Hardware Architectures
+============================================
+
+By providing an accelerator independent interface for kernels, their execution and memory accesses at different hierarchy levels, *alpaka* allows the user to write accelerator independent code that does not neglect performance.
+
+The mapping of the decomposition to the execution environment is handled by the back-ends provided by the *alpaka* library as well as user defined back-ends.
```
sbastrakov, line 6: I think we need to be consistent with spelling: always `backend` or always `back-end`. I am for the former.

SimeonEhrig: Yes, good point. In general, I use `backend`. The version with `-` was written by another person. So I prefer also the former version.

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +--------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | Kernel execution                           | n/a                                           | std::thread(kernel)                                                           | boost::fibers::fiber(kernel)                                                 | ompsetdynamic(0), #pragma omp parallel numthreads(iNumKernelsInBlock)           | #pragma omp target, #pragma omp teams numteams(...) threadlimit(...), #pragma omp distribute, #pragma omp parallel numthreads(...) | cudaConfigureCall, cudaSetupArgument, cudaLaunch |
```
sbastrakov: For the Serial `backend` `Kernel execution` should not be `n/a`, as it can execute kernels. Perhaps the previous line `Lib/API` should be also change accordingly.

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +--------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | ALPAKAFNHOSTACC, ALPAKAFNACC, ALPAKAFNHOST | inline                                        | inline                                                                        | inline                                                                       | inline                                                                          | inline                                                                                                                             | _device__, host, forceinline                     |
```
sbastrakov: Here the underscores seem to have been lost in the alpaka defines.

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +--------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | atomicOp                                   | n/a                                           | std::lockguard< std::mutex >                                                  | n/a                                                                          | #pragma omp critical                                                            | #pragma omp critical                                                                                                               | atomicXXX                                        |
```
sbastrakov: I would not put it as `n/a`, now a reader may think it's not allowed. While it is allowed, it is just same as normal operations.

docs/source/advanced/backends.rst
sbastrakov: In multiple places of this table `shared` and `device` miss one _ in front.

sbastrakov (changes requested):

sbastrakov:
For some reason I can't comment right there. It seems there is wrong caption formatting [just above this](https://github.com/alpaka-group/alpaka/blob/096f52c322985a54a5ea996c84d6a60ccdca3613/docs/source/advanced/mapping.rst#threading-mechanisms) with `### One Block Per Thread`

sbastrakov:
For the sake of the index I suggest changing [this title](https://github.com/alpaka-group/alpaka/blob/096f52c322985a54a5ea996c84d6a60ccdca3613/docs/source/basic/install.rst#alpaka-installation) to just `Intallation`

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,121 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+```
+
+Current restrictions on HCC platform
+++++++++++++++++++++++++++++++++++++
+
+- Workaround for unsupported ``syncthreads_{count|and|or}``.
+  - uses temporary shared value and atomics
+- Workaround for buggy ``hipStreamQuery``, ``hipStreamSynchronize``.
+  - introduces own queue management
+  - ``hipStreamQuery`` and ``hipStreamSynchronize`` did not work in multithreaded environment
```
sbastrakov: This list does not render properly in read the docs.

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,121 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+- Workaround for missing ``cuStreamWaitValue32``.
+  - polls value each 10ms
+- device constant memory not supported yet
+- note, that ``printf`` in kernels is still not supported in HIP
+- exclude ``hipMalloc3D`` and ``hipMallocPitch`` when size is zero otherwise they throw an Unknown Error
+- ``TestAccs`` excludes 3D specialization of Hip back-end for now because ``verifyBytesSet`` fails in ``memView`` for 3D specialization
+- ``dim3`` structure is not available on device (use ``alpaka::vec::Vec`` instead)
+- Constructors' attributes unified with destructors'.
```
sbastrakov: I think either start all entries with a capital letter or all with a small one.

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,121 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
 
 *not available*
+
+
+
+HIP
+```
+
+Current restrictions on HCC platform
+++++++++++++++++++++++++++++++++++++
```
sbastrakov: We need to decide whether such titles have every word (except prepositions and articles) capitalized or not. Now it's a mix in this file. This could be done later as well

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,121 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+  - uses temporary shared value and atomics
+- Workaround for buggy ``hipStreamQuery``, ``hipStreamSynchronize``.
+  - introduces own queue management
+  - ``hipStreamQuery`` and ``hipStreamSynchronize`` did not work in multithreaded environment
+- Workaround for missing ``cuStreamWaitValue32``.
+  - polls value each 10ms
+- device constant memory not supported yet
+- note, that ``printf`` in kernels is still not supported in HIP
+- exclude ``hipMalloc3D`` and ``hipMallocPitch`` when size is zero otherwise they throw an Unknown Error
+- ``TestAccs`` excludes 3D specialization of Hip back-end for now because ``verifyBytesSet`` fails in ``memView`` for 3D specialization
```
sbastrakov: HIP instead of Hip ? Also backend instead of back-end.

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,121 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+- ``dim3`` structure is not available on device (use ``alpaka::vec::Vec`` instead)
+- Constructors' attributes unified with destructors'.
+  - host/device signature must match in HIP(HCC)
```
sbastrakov: I do not understand this point.

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
+Introduction
+============
+
+The *alpaka* library defines and implements an abstract interface for the *hierarchical redundant parallelism* model.
+This model exploits task- and data-parallelism as well as memory hierarchies at all levels of current multi-core architectures.
+This allows to achieve portability of performant codes across various types of accelerators by ignoring specific unsupported levels and utilizing only the ones supported on a specific accelerator.
```
sbastrakov: This question is not for this fragment per se, but the effect is also visible here. In readthedocs it looks visually quite weirdly with the left-side alignment. Do they support alignment by width?

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+All hardware types (multi- and many-core CPUs, GPUs and other accelerators) are treated and can be programmed in the same way.
+The *alpaka* library provides back-ends for *CUDA*, *OpenMP*, *Boost.Fiber* and other methods.
+The policy-based C++ template interface provided allows for straightforward user-defined extension of the library to support other accelerators.
+
+The library name *alpaka* is an acronym standing for **A**\ bstraction **L**\ ibrary for **Pa**\ rallel **K**\ ernel **A**\ cceleration.
+
+Example
+-------
+
+.. literalinclude:: ../../../example/helloWorld/src/helloWorld.cpp
```
sbastrakov, line 16: I think it's fine for now. However, I would prefer the workshop hello world better, as it is more simple and does not have the CI workarounds.

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+   alpaka_add_executable(${_TARGET_NAME} helloWorld.cpp)
+   target_link_libraries(
+     ${_TARGET_NAME}
+     PUBLIC alpaka::alpaka)
+
+You can integrate alpaka into your project via ``find_package()`` in your ``CMakeLists.txt``.
+This requires, that you :doc:`install </basic/install>` alpaka.
+If you do not install alpaka in a default path such as ``/usr/local/`` you have to set the ``CMake`` argument ``-Dalpaka_ROOT=/path/to/alpaka/install``.
+
+The cmake configuration decides which ``alpaka::acc`` is available during compiling. For example, if you configure your ``cmake`` build with the CUDA backend (``cmake -DALPAKA_ACC_GPU_CUDA_ENABLE=ON``), ``cmake`` checks, if the CUDA SDK is available and if it found, the C++ template ``alpaka::acc::AccGpuCudaRt`` is available during compiling.
```
sbastrakov: `which ``alpaka::acc`` is available` - technically it's which templates in this namespace are available. Can we just say `which alpaka accelators are available`? Maybe with a hyperlink to docs on accelerators.

sbastrakov: I would remove `cmake` from ``cmake -DALPAKA_ACC_GPU_CUDA_ENABLE=ON`` as this is not a complete command, and it's already clear we are talking about cmake arguments.

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+
+What is alpaka
+--------------
+
+alpaka is ...
+~~~~~~~~~~~~~
+
+A Abstract Interface
```
sbastrakov: `An`, not `A`

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+   It describing parallel execution on multiple hierarchy levels. It allows to implement a mapping to various hardware architectures but is no optimal mapping itself.
+
+Sustainably
+   *alpaka* decoupled the application from the availability of different accelerator frameworks in different versions, such as OpenMP, CUDA, HIP, etc. (50% on the way to reach full performance portability).
```
sbastrakov: Why in the past? I think `decouples`

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Heterogeneity
+   An identical algorithm / kernel can be executed on heterogeneous parallel systems by selecting the target device. This allows the best performance for each algorithm and/or a good utilization of the system without major code changes.
+
+Maintainability
+   *alpaka* allows to provide a single version of the algorithm / kernel that can be used by all back-ends. There is no need for "copy and paste" kernels with different API calls for different accelerators. All the accelerator dependent implementation details are hidden within the *alpaka* library.
+
+Testability
```
sbastrakov: This list is a little weird with a mix of specific nouns (interface), general category nouns (maintainability) and adjectives (optimizable).

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+Extensible
+   Every concept described by the *alpaka* abstraction can be implemented by users. Therefore it is possible to non-intrusively define new devices, queues, buffer types or even whole accelerator back-ends.
+
+Data Structure Agnostic
+   The user can use and define arbitrary data structures.
+
+alpaka is not ...
+~~~~~~~~~~~~~~~~~
+
+An automatically optimal mapping of algorithms / kernels to various acceleration platforms
```
sbastrakov: Same point here, this one may be written as "Automatically optimizing mapping of ..." to match the style of the following items.

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+Automatically optimizing concurrent data accesses
+   *alpaka* does not provide feature to create optimized memory layouts.
+
+Handling or hiding differences in arithmetic operations
+   For example, due to **different rounding** or different implementations of floating point operations, results can differ slightly between accelerators.
+
+Guaranteeing any determinism of results
+   Due to the freedom of the library to reorder or repartition the threads within the tasks it is not possible or even desired to preserve deterministic results. For example, the non-associativity of floating point operations give non-deterministic results within and across accelerators.
+
+The *alpaka* library is aimed at parallelization within nodes of a cluster.
```
sbastrakov: Maybe add explicitly smth like "... at parallelization on shared memory, i.e. within nodes of a cluster"

docs/source/basic/library.rst
```diff
@@ -1,16 +1,65 @@
-.. highlight:: cpp
+Library Interface
+=================
+
+As described in the chapter about the Abstraction, the general design of the library is very similar to *CUDA* and *OpenCL* but extends both by some points, while not requiring any language extensions.
```
sbastrakov: Add hyperlink for Abstraction docs?

docs/source/basic/library.rst
```diff
@@ -1,16 +1,65 @@
...
+An ``Executor`` can be enqueued into a ``Queue`` and will execute the ``Kernel`` (after all previous tasks in the queue have been completed).
+The ``Kernel`` in turn has access to the ``Accelerator`` it is running on.
+The ``Accelerator`` provides the ``Kernel`` with its current index in the block or grid, their extents or other data as well as it allows to allocate shared memory, execute atomic operations and many more.
+
 
 Interface Usage
-===============
+---------------
 
 Accelerator Executable Functions
```
sbastrakov: Do we call those like that? I would suggest just "device functions", but not sure if that's consistent with the rest

docs/source/info/similar_projects.rst
```diff
@@ -0,0 +1,28 @@
...
+   * https://on-demand.gputechconf.com/supercomputing/2013/presentation/SC3103\_Towards-Performance-Portable-Applications-Kokkos.pdf
+   * https://dx.doi.org/10.3233/SPR-2012-0343
+
+Kokkos provides an abstract interface for portable, performant shared memory-programming. It is a C++ library that offers ``parallel_for``, ``parallel_reduce`` and similar functions for describing the pattern of the parallel tasks. The execution policy determines how the threads are executed. For example, this influences the sizes of blocks of threads or if static or dynamic scheduling should be used. The library abstracts the kernel as a function object that can not have any user defined parameters for its ``operator()``. Inconveniently, arguments have to be stored in members of the function object coupling algorithm and data together. *KOKKOS* provides both, abstractions for parallel execution of code and data management. Multidimensional arrays with a neutral indexing and an architecture dependent layout are available, which can be used, for example, to abstract the underlying hardwares preferred memory access scheme that could be row-major, column-major or even blocked.
```
sbastrakov: Perhaps we should keep neutral here, without judgemental remarks like `Inconveniently`

sbastrakov:
Great work @SimeonEhrig ! Some comments are above. I can have another look through if needed later.

docs/source/advanced/mapping.rst
SimeonEhrig: @sbastrakov 
I greped over the whole project. Most of the time, `back-end` is used. One big exception are the unified HIP/CUDA files, which use `backend`. I would say, we use `back-end`, because it is mostly used in the project, in special in messages in cmake.

sbastrakov: Okay. I think all 3 are not bad, and consistency is most important.

docs/source/advanced/backends.rst
SimeonEhrig: But what should I write? No special function or library is used. Maybe something like `bare`, `pure C++`, `no library` and `C++ function call`?

sbastrakov: I guess something like it. Or Lib/API = standard C++, and kernel execution = sequential

docs/source/advanced/backends.rst
SimeonEhrig: I wrote `not necessary`

docs/source/advanced/backends.rst
SimeonEhrig: I did it for the whole doc. Was not much work.

docs/source/advanced/backends.rst
SimeonEhrig: I did it for the whole doc. Was not much work.

docs/source/advanced/backends.rst
SimeonEhrig: I don't know. Matthias wrote this but I think, the HIP section need a general overwork, because some parts, like the `hipcc` (which is deprecated) are outdated.

docs/source/basic/intro.rst
SimeonEhrig: I add some css code, to justify the text bodies. You can check the result, if I pushed the commit.

docs/source/basic/intro.rst
SimeonEhrig: I agree with you. For a first look the smaller workshop example is more suitable. But then we have the problem of keeping the example up-to-date :-(

docs/source/basic/library.rst
SimeonEhrig: I would say `Accelerator function`, because the text writes about the accelerators the whole time.

docs/source/basic/intro.rst
SimeonEhrig: I agree with you, but I'm not good a finding good headings. This is the reason, why the second part (`alpaka is not ...`) has sentences as headings. Do you have some suggestions?

SimeonEhrig:
Thanks @sbastrakov for the great feedback. I think you have found many old errors and some porting errors from the old markdown files in addition to my new errors :sweat_smile:

docs/source/basic/intro.rst
sbastrakov: Yes, I could help with phrasing these.

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +---------------------------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | ``ALPAKA_FN_HOST_ACC``, ``ALPAKA_FN_ACC``, ``ALPAKA_FN_HOST`` | inline                                        | inline                                                                        | inline                                                                       | inline                                                                          | inline                                                                                                                             | __device__, host, forceinline                    |
```
sbastrakov: `__device__` is now fixed, but the `_` are still missing in `host` and `forceinline`

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +---------------------------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+
+Serial
+``````
+
+The serial accelerator only allows blocks with exactly one thread.
```
sbastrakov: Is that correct? I thought it allows any workdiv, just executes sequentially

SimeonEhrig: Good question. I never tried it. I simply copy and past the text. @psychocoderHPC Do you know, if it is (still) correct?

BenjaminW3: The comment is true. You can not have multiple threads in a block because you can not synchronize those threads because they can not be executed asynchronously.

sbastrakov: Ah, true. Thanks for explaining.

docs/source/basic/intro.rst
SimeonEhrig: We have discussed in a VC that we want to reduce the amount of comments and move the removed comments directly into readthedocs. We also try to remove CI dependent code. See Issue #1101

docs/source/advanced/backends.rst
SimeonEhrig: I opened a new Issue, which contains checking the HIP backend documentation: #1105

docs/source/basic/intro.rst
sbastrakov: Okay, so I think the first five "alpaka is..." items can be reformulated as:

- Abstract
- Sustainable
- Heterogeneous
- Maintainable
- Testable

The last three already have this style and can be left as is

docs/source/basic/intro.rst
sbastrakov: Ànd this one I would write as "alpaka does not..." (instead of `is not`) and then the items as

- Automatically provide an optimal mapping of kernels to various acceleration platforms
- Automatically optimize concurrent data access
- Handle differences in arithmetic operations
- Guarantee determinism of results

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +---------------------------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | getIdx                                                        | n/a                                           | block-kernel: mapping of std::thisthread::getid() grid-block: member variable | block-kernel: mapping of std::thisfiber::getid() grid-block: member variable | block-kernel: ompgetthreadnum() to 3D index mapping grid-block: member variable | block-kernel: ompgetthreadnum() to 3D index mapping grid-block: member variable                                                    | threadIdx, blockIdx                              |
```
sbastrakov: Why is `getIdx` for `Serial` marked as `n/a`? I think that gives a wrong impression that it's not useable.

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+
+Index
++++++
+
+OpenMP only provides a linear thread index. This index is converted to a 3 dimensional index at runtime.
+
+Atomic
+++++++
+
+We can not use '#pragma omp atomic' because braces or calling other functions directly after ``#pragma omp atomic`` are not allowed.
```
sbastrakov: The first '#pragma omp atomic' is not formatted correctly. (The second one is).

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+alpaka uses the mathematical C/C++ array indexing scheme [z][y][x].
+Dimension 0 in this case is z, dimensions 2 is x.
 
-Porting CUDA to *alpaka*
-------------------------
+Furthermore alpaka does not require the indices and extents to be 3-dimensional.
+The accelerators are templatized on and support arbitrary dimensionality.
+NOTE: Currently the CUDA implementation is restricted to a maximum of 3 dimensions!
+
+NOTE: The CUDA-accelerator back-end can change the current CUDA device and will NOT set the device back to the one prior to the invocation of the alpaka function!
```
sbastrakov: This part is mostly a copy of the next section, from line 110. I think it should be removed either here or there.

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,131 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+++++++++++++++++++++++++++++++++++++
+
+- Workaround for unsupported ``syncthreads_{count|and|or}``.
+
+  - Uses temporary shared value and atomics
+
+- Workaround for buggy ``hipStreamQuery``, ``hipStreamSynchronize``.
+
+  - Introduces own queue management
+  - ``hipStreamQuery`` and ``hipStreamSynchronize`` did not work in multithreaded environment
```
sbastrakov: I think better replace `did` with `do`

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,131 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+
+- Workaround for missing ``cuStreamWaitValue32``.
+
+  - Polls value each 10ms
```
sbastrakov: Space missing in `10ms`

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,131 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+- If PATH points to the location of the HIP binaries, the following command should list several relevant environment variables, and also the selected compiler on your ``system-\`hipconfig -f\```
+- Compile and run the `square sample`_, as pointed out in the original `HIP install guide`_.
+
+.. _square sample: https://github.com/ROCm-Developer-Tools/HIP/tree/master/samples/0_Intro/square
+.. _HIP install guide: https://github.com/ROCm-Developer-Tools/HIP/blob/master/INSTALL.md#user-content-verify-your-installation
+
+Compiling Examples with HIP Back End
+++++++++++++++++++++++++++++++++++++
+
+As of now, the back-end has only been tested on the NVIDIA platform.
```
sbastrakov: I don't think it's true, @psychocoderHPC ran it on AMD with HIP and according to him everything works, except peer to peer copies

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,131 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+
+* NVIDIA Platform
+
+  * One issue in this branch of alpaka is that the host compiler flags don't propagate to the device compiler, as they do in CUDA. This is because a counterpart to the ``CUDA_PROPAGATE_HOST_FLAGS`` cmake variable has not been defined in the FindHIP.cmake file.
```
sbastrakov: `in this branch of alpaka` - what branch?

docs/source/basic/abstraction.rst
```diff
@@ -0,0 +1,275 @@
+Abstraction
+===========
+
+.. note::
+
+   Objective of the abstraction is to separate the parallelization strategy from the algorithm itself.
+   Algorithm code written by users should not depend on any parallelization library or specific strategy.
+   This would allow to exchange the parallelization back-end without any changes to the algorithm itself.
```
sbastrakov: `allow exchanging`, or perhaps even better `enable exchanging`

docs/source/basic/intro.rst
sbastrakov: Maybe replace `portability of performant codes` with just `performance portability` ?

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+The cmake configuration decides which alpaka accelerators are available during compiling. For example, if you configure your ``cmake`` build with the CUDA back-end (``-DALPAKA_ACC_GPU_CUDA_ENABLE=ON``), ``cmake`` checks, if the CUDA SDK is available and if it found, the C++ template ``alpaka::acc::AccGpuCudaRt`` is available during compiling.
+
+What is alpaka
+--------------
+
+alpaka is ...
+~~~~~~~~~~~~~
+
+An Abstract Interface
+   It describing parallel execution on multiple hierarchy levels. It allows to implement a mapping to various hardware architectures but is no optimal mapping itself.
```
sbastrakov: `It describing` is not right here. I guess the idea was either `Is describing` or `It describes`

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Sustainably
+   *alpaka* decouple the application from the availability of different accelerator frameworks in different versions, such as OpenMP, CUDA, HIP, etc. (50% on the way to reach full performance portability).
```
sbastrakov: `decouples`

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Heterogeneity
+   An identical algorithm / kernel can be executed on heterogeneous parallel systems by selecting the target device. This allows the best performance for each algorithm and/or a good utilization of the system without major code changes.
+
+Maintainability
+   *alpaka* allows to provide a single version of the algorithm / kernel that can be used by all back-ends. There is no need for "copy and paste" kernels with different API calls for different accelerators. All the accelerator dependent implementation details are hidden within the *alpaka* library.
+
+Testability
+   Due to the easy back-end switch, no special hardware is required for testing the kernels. Even if the simulation itself will always use the *CUDA* back-end, the tests can completely run on a CPU. As long as the *alpaka* library is thoroughly tested for compatibility between the acceleration back-ends, the user simulation code is guaranteed to generate identical results (ignoring rounding errors / non-determinism) and is portable without any changes.
```
sbastrakov: `will always use` should be `always uses`

docs/source/info/similar_projects.rst
```diff
@@ -0,0 +1,28 @@
...
+   * https://on-demand.gputechconf.com/supercomputing/2013/presentation/SC3103\_Towards-Performance-Portable-Applications-Kokkos.pdf
+   * https://dx.doi.org/10.3233/SPR-2012-0343
+
+Kokkos provides an abstract interface for portable, performant shared memory-programming. It is a C++ library that offers ``parallel_for``, ``parallel_reduce`` and similar functions for describing the pattern of the parallel tasks. The execution policy determines how the threads are executed. For example, this influences the sizes of blocks of threads or if static or dynamic scheduling should be used. The library abstracts the kernel as a function object that can not have any user defined parameters for its ``operator()``. Arguments have to be stored in members of the function object coupling algorithm and data together. *KOKKOS* provides both, abstractions for parallel execution of code and data management. Multidimensional arrays with a neutral indexing and an architecture dependent layout are available, which can be used, for example, to abstract the underlying hardwares preferred memory access scheme that could be row-major, column-major or even blocked.
```
sbastrakov: Question: do we need to write such paragraphs as a single line in `.rst` to get alignment by width?

sbastrakov:
Please find some more comments and questions above. I understand the text is mostly reused from the previous documentation, so did not do all my text suggestions, but just a couple (I can do a follow-up PR for that, as it would be quicker that way).

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
-CUDA GPUs
+.. highlight:: bash
+
+Back Ends
```
j-stephan: I would call this either back-end (seems to be standard alpaka terminology) or backend.

docs/source/advanced/backends.rst
j-stephan: It should be `std::this_thread::get_id()`, `boost::this_fiber::get_id()`, `omp_get_num_threads()` (note the switch of `thread` and `num`)

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +---------------------------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | Kernel execution                                              | sequential                                    | std::thread(kernel)                                                           | boost::fibers::fiber(kernel)                                                 | ompsetdynamic(0), #pragma omp parallel numthreads(iNumKernelsInBlock)           | #pragma omp target, #pragma omp teams numteams(...) threadlimit(...), #pragma omp distribute, #pragma omp parallel numthreads(...) | cudaConfigureCall, cudaSetupArgument, cudaLaunch |
```
j-stephan: It should be `omp_set_dynamic`, `pragma omp parallel num_threads`, `pragma omp teams num_teams(...) thread_limit(...)`

docs/source/advanced/backends.rst
```diff
@@ -1,14 +1,108 @@
...
+    +---------------------------------------------------------------+-----------------------------------------------+-------------------------------------------------------------------------------+------------------------------------------------------------------------------+---------------------------------------------------------------------------------+------------------------------------------------------------------------------------------------------------------------------------+--------------------------------------------------+
+    | atomicOp                                                      | not necessary                                 | std::lockguard< std::mutex >                                                  | n/a                                                                          | #pragma omp critical                                                            | #pragma omp critical                                                                                                               | atomicXXX                                        |
```
j-stephan: Should be `std::lock_guard`

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,131 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+
+  - Introduces own queue management
+  - ``hipStreamQuery`` and ``hipStreamSynchronize`` did not work in multithreaded environment
+
+- Workaround for missing ``cuStreamWaitValue32``.
+
+  - Polls value each 10ms
+
+- Device constant memory not supported yet
+- Note, that ``printf`` in kernels is still not supported in HIP
```
j-stephan: The comma is not necessary here

docs/source/advanced/backends.rst
```diff
@@ -434,3 +528,131 @@ The following tables list the functions available in the `CUDA Runtime API <http
...
+  - Host/device signature must match in HIP(HCC)
+
+- A chain of functions must also provide correct host-device signatures
+
+  - E.g. a host function cannot be called from a host-device function
+
+- Recompile your target when HCC linker returned the error:
+  "File format not recognized
+  clang-7: error: linker command failed with exit code 1"
+- If compile-error occurred, the linker still may link, but without the device code
```
j-stephan: The first comma is unnecessary

docs/source/advanced/backends.rst
SimeonEhrig: I would replace it with: `no mapping: every time id 0`

docs/source/advanced/rationale.rst
```diff
@@ -272,7 +272,7 @@ The constant memory is a fast, cached, read-only memory that is beneficial when
 In this case it is as fast as a read from a register.
 
 
-Access to accelerator dependent functionality
+Access to Accelerator Dependent Functionality
```
j-stephan: I believe there should be a dash `-` between `Accelerator` and `Dependent`

docs/source/basic/abstraction.rst
```diff
@@ -0,0 +1,275 @@
...
+   Besides allowing to test different parallelization strategies this also makes it possible to port algorithms to new, yet unsupported, platforms.
+
+Parallelism and memory hierarchies at all levels need to be exploited in order to achieve performance portability across various types of accelerators.
+Within this chapter an abstraction will be derivated that tries to provide a maximum of parallelism while simultaneously considering implementability and applicability in hardware.
```
j-stephan: While I appreciate the use of classical English we should use `derive` instead of `derivated`. The latter is quite outdated.

docs/source/basic/abstraction.rst
```diff
@@ -0,0 +1,275 @@
...
+When it is directly implemented as a graph, at least all depending tasks have to be updated and checked if they are ready to be executed after a task finished.
+Depending on the size of the graph and the number of edges this can be a huge overhead.
+
+*OpenCL* allows to define a task graph in a somewhat different way.
+Tasks can be enqueued into an out-of-order command queue combined with events that have to be finished before the newly enqueued task can be started.
+Tasks in the command queue with unmet dependencies are skipped and subsequent ones are executed.
+The ``CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE`` property of a command queue is an optional feature only supported by few vendors.
+Therefore, it can not be assumed to be available on all systems.
+
+*CUDA* on the other hand does currently (version 7.5) not support such out-of-order queues in any way.
```
j-stephan, line 60: We should probably update the CUDA version here.

docs/source/basic/abstraction.rst
```diff
@@ -0,0 +1,275 @@
...
+This reduces the number of incoming edges of subsequent tasks.
+
+The problem with this model is the inherent overhead and the missing hardware and API support.
+When it is directly implemented as a graph, at least all depending tasks have to be updated and checked if they are ready to be executed after a task finished.
+Depending on the size of the graph and the number of edges this can be a huge overhead.
+
+*OpenCL* allows to define a task graph in a somewhat different way.
+Tasks can be enqueued into an out-of-order command queue combined with events that have to be finished before the newly enqueued task can be started.
+Tasks in the command queue with unmet dependencies are skipped and subsequent ones are executed.
+The ``CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE`` property of a command queue is an optional feature only supported by few vendors.
```
j-stephan, line 57: We really need a source here. `CL_QUEUE_OUT_OF_ORDER_EXEC_MODE_ENABLE` has been present in the OpenCL specification since version 1.0. There is no mention of it being optional as far as I can see.

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+     ${_TARGET_NAME}
+     PUBLIC alpaka::alpaka)
+
+You can integrate alpaka into your project via ``find_package()`` in your ``CMakeLists.txt``.
+This requires, that you :doc:`install </basic/install>` alpaka.
+If you do not install alpaka in a default path such as ``/usr/local/`` you have to set the ``CMake`` argument ``-Dalpaka_ROOT=/path/to/alpaka/install``.
+
+The cmake configuration decides which alpaka accelerators are available during compiling. For example, if you configure your ``cmake`` build with the CUDA back-end (``-DALPAKA_ACC_GPU_CUDA_ENABLE=ON``), ``cmake`` checks, if the CUDA SDK is available and if it found, the C++ template ``alpaka::acc::AccGpuCudaRt`` is available during compiling.
+
+What is alpaka
```
j-stephan: Maybe call this "About alpaka".

docs/source/advanced/backends.rst
SimeonEhrig: This is also a reason, why we have to overwork it. Since @tdd11235813 wrote the documentation there wars a least one release, which also solve some problems with 3D memory. I think, without CI, it is hard to keep the documentation up-to-date.

docs/source/basic/intro.rst
j-stephan: We could also use "alpaka features...". Then we could keep the nouns and just rephrase "Sustainably".

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+   Every concept described by the *alpaka* abstraction can be implemented by users. Therefore it is possible to non-intrusively define new devices, queues, buffer types or even whole accelerator back-ends.
+
+Data Structure Agnostic
+   The user can use and define arbitrary data structures.
+
+alpaka is not ...
+~~~~~~~~~~~~~~~~~
+
+An automatically optimal mapping of algorithms / kernels to various acceleration platforms
+   Except in trivial examples an optimal execution always depends on suitable selected data structure. An adaptive selection of data structures is a separate topic that has to be implemented in a distinct library.
```
j-stephan: Either structure**s** or "a suitable ... structure".

docs/source/basic/intro.rst
```diff
@@ -0,0 +1,90 @@
...
+
+Automatically optimizing concurrent data accesses
+   *alpaka* does not provide feature to create optimized memory layouts.
```
j-stephan, line 78: feature**s**

docs/source/basic/library.rst
```diff
@@ -1,16 +1,65 @@
-.. highlight:: cpp
+Library Interface
+=================
+
+As described in the chapter about the :doc:`Abstraction </basic/abstraction>`, the general design of the library is very similar to *CUDA* and *OpenCL* but extends both by some points, while not requiring any language extensions.
```
j-stephan, line 4: General question: Why are we mentioning OpenCL so often? Except for the experimental SYCL backend we have no relationship to OpenCL.

docs/source/info/similar_projects.rst
```diff
@@ -0,0 +1,28 @@
...
+
+
+`Thrust <https://thrust.github.io/>`_
```
j-stephan, line 26: Is Thrust really a similar library? I feel that Thrust, ArrayFire, ... are much more high-level than alpaka.

j-stephan (changes requested):
First review pass done.

docs/source/info/similar_projects.rst
SimeonEhrig: No, it has something to do with my editor, why the line breaks are missing. I add some.

docs/source/advanced/backends.rst
SimeonEhrig: @psychocoderHPC Sends me a paper, which means, that `back ends` is fine for headings and `back-ends` is fine for the text. But everybody means, it look false, so I will change it.

docs/source/basic/intro.rst
sbastrakov: This is another way. Maybe "alpaka enables..." then?

docs/source/basic/library.rst
sbastrakov: I think because it was like that at the Master thesis used as a basis for these texts

docs/source/basic/abstraction.rst
SimeonEhrig: The problem is, that we have to overwork the whole paragraph, because CUDA 10 provides a graph library.

docs/source/basic/abstraction.rst
SimeonEhrig: Same here. A general overwork of the whole abstraction section is necessary to bring it up to date and make it more readable. But for this PR it was planned to make only some small changes to the text to speed up the merging.

docs/source/basic/abstraction.rst
sbastrakov: Okay, then I think better keep as is now, and later someone provides a follow-up PR to fix. And those follow-up PRs could be smaller, fixing one particular topic / section / etc.

docs/source/info/similar_projects.rst
sbastrakov: Well, similar in one way but not on another. I guess this may be left as is, and clarified later.
But I agree they operate on much higher abstraction level, and it's not easy possible to port between those and alpaka

docs/source/basic/library.rst
SimeonEhrig: Because it was 2015 :sunglasses: . Most parts were copied from Benjamin's master thesis. In my opinion, with overworking the text, we should remove a lot of the OpenCL stuff.

docs/source/info/similar_projects.rst
SimeonEhrig: It's fine by me, but we can remove it. I thought it would be better to keep two rather than one of more than 10 related works (that was CUDA, OpenCL, ...).

SimeonEhrig: I keep it, because it looks better to have two projects and it is not fully unrelated.

SimeonEhrig:
@sbastrakov  @j-stephan 
Can you please check if the headings in the introduction are ok? I think this should be the last item before the merge.

The rest of the outstanding tasks should be done in extra PR.

sbastrakov:
@SimeonEhrig I do not have objections.

sbastrakov (dismissed):

sbastrakov (approved):

j-stephan:
Looks good now. @psychocoderHPC This can be merged.

PR #1099, opened 2020-07-26
Use `apt-get` to install clang

BenjaminW3:
* uses `apt-get` to install clang instead of manual download. This will automatically install correct dependencies.
* removes `ALPAKA_CI_CLANG_DIR` and `ALPAKA_CI_CLANG_LIBSTDCPP_VERSION`


BenjaminW3:
Ready to be merged. Github seems to have issues with reporting the build state of some jobs.

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -235,7 +235,7 @@ namespace alpaka
                 // For equal block thread extent this is easily the nth root of blockThreadCountMax.
                 if(gridBlockExtentSubDivRestrictions == GridBlockExtentSubDivRestrictions::EqualExtent)
                 {
-                    double const fNthRoot(std::pow(blockThreadCountMax, 1.0/static_cast<double>(TDim::value)));
+                    double const fNthRoot(std::pow(static_cast<double>(blockThreadCountMax), 1.0/static_cast<double>(TDim::value)));
```
sbastrakov, line 238: Is this change required for `apt-get` to work??

BenjaminW3: This is not directly related to apt-getting clang. But by doing so, we now get the correct std lib installed automatically instead of always using the one we defined with `ALPAKA_CI_CLANG_LIBSTDCPP_VERSION`, which was an old one. The newer std-lib seems to provide multiple equally matching overloads of `std::pow` and compilation failed.

sbastrakov (approved):

PR #1104, opened 2020-07-28
Change examples (except reduce) to use getValidWorkDiv

### LINKED ISSUE #1100: Use getValidWorkDiv() in examples instead of manually setting it
This will make examples more portable between accelerators, and so there are potentially less guards to write and more portability out-of-the-box for demonstration. I will provide a PR later today or tomorrow.

sbastrakov:
This poses less limitations to use different accelerators.
Closes #1100

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -186,22 +186,23 @@ auto main()
     DevQueue devQueue(devAcc);
     HostQueue hostQueue(devHost);
 
-    // Define the work division
+    // Define the work division for kernels to be run on devAcc and devHost
     using Vec = alpaka::vec::Vec<Dim, Idx>;
     Vec const elementsPerThread(Vec::all(static_cast<Idx>(1)));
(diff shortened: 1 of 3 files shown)


example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -263,7 +258,7 @@ auto main()
 
     alpaka::kernel::exec<Host>(
         hostQueue,
-        workdiv,
+        workDiv,
```
sbastrakov: Note: this change is for consistency with other examples.

BenjaminW3 (dismissed):

sbastrakov:
As hinted by the failing tests, the bufferCopy had another issue that was uncovered with this change. It runs kernels on both host and device, but used the same workDiv for both. Which is wrong, as a valid work div on device is not necessarily valid on host. I now fixed the example to use two separate work divs, each valid for the corresponding device. Rebased.

sbastrakov:
@BenjaminW3 @psychocoderHPC it seems the last CI job is stuck.

BenjaminW3 (approved):

PR #1107, opened 2020-07-31
fix HIP-clang compile

psychocoderHPC:
HIP-clang requires an explicit implemented destructor for the random
number generator.


include/alpaka/rand/RandUniformCudaHipRand.hpp
```diff
@@ -77,6 +77,10 @@ namespace alpaka
                     {
                     }
 
+#if BOOST_COMP_HIP
+                    //-----------------------------------------------------------------------------
+                    ALPAKA_FN_HOST_ACC ~Xor(){}
```
sbastrakov: `= default` does not work?

psychocoderHPC: I changed to `= default`

BenjaminW3 (approved):

psychocoderHPC:
@BenjaminW3 where is the test linux_clang-6_cuda-9.0_debug_analysis coming from? The test is not marked as CI or is this a gihub action job.

BenjaminW3:
It was part of the github actions CI.
It has been removed but is still marked as required. I will fix this.

PR #1108, opened 2020-07-31
fix error handling CUDA/HIP

psychocoderHPC:
Avoid that an error from CUDA/HIP which is ignored by alpaka is throwing with the next API call.

A typical case in alpaka is the error: `cudaErrorNotReady` if we do an event querry.

### DIFF
--- a/include/alpaka/core/UniformCudaHip.hpp
+++ b/include/alpaka/core/UniformCudaHip.hpp
@@ -90,6 +90,11 @@ namespace alpaka
                     {
                         rtCheck(error, ("'" + std::string(cmd) + "' returned error ").c_str(), file, line);
                     }
+                    else
+                    {
+                        // reset the last error to avoid propagation to the next CUDA/HIP API call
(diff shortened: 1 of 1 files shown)


sbastrakov (dismissed):

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -90,6 +90,9 @@ namespace alpaka
                     {
                         rtCheck(error, ("'" + std::string(cmd) + "' returned error ").c_str(), file, line);
                     }
+                    else
+                        // reset the last error to avoid propagation to the next CUDA/HIP API call
```
BenjaminW3, line 95: I would prefer braces here, especially since the comment makes it multiline.

psychocoderHPC: I will change it

BenjaminW3 (approved):

BenjaminW3 (approved):

PR #1110, opened 2020-08-03
set minimal HIP version to 3.5

psychocoderHPC:
- Update to the FindHIP implementation from rocm-3.5
- set HIP 3.5 as minimum requirement
- switch CI base container for HIP to rocm3.5 container
  - test HIP-clang instead of HIP-nvcc
  - remove testing for HIP-nvcc


BenjaminW3 (dismissed):

.github/workflows/ci.yml
```diff
@@ -74,10 +74,10 @@ jobs:
         include:
         ## HIP
         - name: linux_hip_nvcc-9.2_gcc-5_debug_hip_only
-          os: ubuntu-latest
+          os: rocm/pytorch:rocm3.5_ubuntu16.04_py3.6_pytorch
```
psychocoderHPC: @BenjaminW3 I expect this is the place to load a different base container or? I will switch to a container with clean installed hip-clang.

BenjaminW3: I do not think that this works. The `os:` only allows one of the few virtual machine types which github allows.

BenjaminW3: `ALPAKA_CI_DOCKER_BASE_IMAGE_NAME` is the correct one

psychocoderHPC: thx

psychocoderHPC:
@BenjaminW3 I switched now to a rocm container and we test HIP-clang. HIP-clang was always the target for HIP and since rocm3.5 is shipping HIP-clang we do not handle installation.

script/install.sh
```diff
@@ -67,7 +66,7 @@ fi
 if [ "$ALPAKA_CI_OS_NAME" = "Linux" ]
 then
     # Minimize docker image size
-    sudo apt-get --quiet --purge autoremove
+    sudo apt-get --quiet --purge --yes autoremove
     sudo apt-get clean
     rm -rf /var/lib/apt/lists/* /tmp/* /var/tmp/*
 fi
```
BenjaminW3: You can delete this whole `if-fi` block because it is not useful as we do not cache the image anymore.

psychocoderHPC: will do

psychocoderHPC:
I removed the if block and CI passed, ready to merge

BenjaminW3 (approved):

PR #1111, opened 2020-08-03
Explicitly require backends and do not enable them by default

BenjaminW3:
Backends which could not be found by CMake were simply disabled even though they were explicitly requested on the command line. We had multiple such occurences in our CI jobs that were hidden (boost.Fibers and OpenMP was not found correctly in some jobs due to setup problems). Users will probably also run into similar issues.

By not enabling any backend by default we can be sure that enabled backends have been explicitly requested. This allows us to
error when a backend is not found.


script/run_generate.sh
```diff
@@ -74,7 +74,7 @@ cd build/
 
 "${ALPAKA_CI_CMAKE_EXECUTABLE}" -G "${ALPAKA_CI_CMAKE_GENERATOR}" ${ALPAKA_CI_CMAKE_GENERATOR_PLATFORM} \
     -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON \
-    "$(env2cmake BOOST_ROOT)" -DBOOST_LIBRARYDIR="${ALPAKA_CI_BOOST_LIB_DIR}/lib" -DBoost_USE_STATIC_LIBS=ON -DBoost_USE_MULTITHREADED=ON -DBoost_USE_STATIC_RUNTIME=OFF \
+    "$(env2cmake BOOST_ROOT)" -DBOOST_LIBRARYDIR="${ALPAKA_CI_BOOST_LIB_DIR}/lib" -DBoost_USE_STATIC_LIBS=ON -DBoost_USE_MULTITHREADED=ON -DBoost_USE_STATIC_RUNTIME=OFF -DBoost_ARCHITECTURE="-x64" \
```
BenjaminW3, line 77: Without this, newer boost libs have not ben found.

script/install_clang.sh
```diff
@@ -26,6 +26,12 @@ then
     travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install libc++-dev
     travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install libc++abi-dev
 fi
+
+if [ "${ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE}" = "ON" ] || [ "${ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE}" = "ON" ] || [ "${ALPAKA_ACC_CPU_BT_OMP4_ENABLE}" = "ON" ]
+then
+    travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install libomp-dev
```
BenjaminW3, line 32: We forgot to install OpenMP libs when using clang. This simply lead to OpenMP being disabled for clang in CI.

BenjaminW3:
Ready for review!

psychocoderHPC (approved):

psychocoderHPC:
We need to mention this change explicit in the next release. This will change the user workflow completely but IMO this is the right way to go!

PR #1116, opened 2020-08-12
ALPAKA_DECAY_T: Fix INTEL detection, Add PGI

jkelling:
The test `#ifdef BOOST_COMP_INTEL` thinks every compiler is Intel.

### DIFF
--- a/include/alpaka/core/Decay.hpp
+++ b/include/alpaka/core/Decay.hpp
@@ -26,7 +26,7 @@
 // in parameter pack expansion expressions is to avoid warnings from diagnostic
 // tools, and also for brevity.
 //-----------------------------------------------------------------------------
-#ifdef BOOST_COMP_INTEL
+#if BOOST_COMP_INTEL || BOOST_COMP_PGI
     #define ALPAKA_DECAY_T(Type) typename std::decay<Type>::type
 #else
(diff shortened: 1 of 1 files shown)


sbastrakov:
@jkelling I believe that uses alpaka variable from core/BoostPredef, not a Boost counterpart. And should detect properly.

sbastrakov:
Here is the detection: https://github.com/alpaka-group/alpaka/blob/6310424eff951eddedd0af8d26b8aa0012254075/include/alpaka/core/BoostPredef.hpp#L124-L129

jkelling:
Yes, it does. `BOOST_COMP_INTEL` is always defined by boost and not undef'd in `BoostPredef.hpp`.

sbastrakov:
@jkelling sorry, I do not follow. Isn't it what happens on line 127 in the snippet I posted?

jkelling:
Maybe I was not clear enough: There is no problem with the `BOOST_COMP_INTEL` (this PR is not changing anything about that). The problem is the use of `#ifdef` instead of `#if`, which is what this PR changes.

sbastrakov:
Ah, got you. I did not read the changes carefully and thought they meant the other thing. Thanks for explaining!

sbastrakov (approved):

PR #1118, opened 2020-08-12
BlockSharedMemDynMember::staticAllocBytes: -> function

jkelling:
GCC does not allow static const/constexpr data members in types mapped to a targetd device in OpenMP 4.5 and OpenACC. At the moment, GCC is the only compiler which can compile some alpaka examples (helloWorld, vectorAdd) with OpenACC and give something that runs successfully (compile with `-foffload=disabled` and run with `export ACC_DEVICE_TYPE=host`), but this at least provides some syntax checking.

I hoped, that `BlockSharedMemDynMember::staticAllocBytes` would remain private and thus only had this workaround on the OpenACC branch. Because it became `public` the workaround needs to be public, too.

### DIFF
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -28,6 +28,20 @@ namespace alpaka
         {
             namespace dyn
             {
+                namespace detail
+                {
+                    //#############################################################################
+                    //! "namespace" for static constexpr members that should be in BlockSharedMemDynMember
+                    //! but cannot be because having a static const member breaks GCC 10
+                    //! OpenMP target and OpenACC: type not mappable.
+                    template<unsigned int TStaticAllocKiB>
+                    struct BlockSharedMemDynMemberStatic
+                    {
+                        //! Storage size in bytes
+                        static constexpr unsigned int staticAllocBytes = TStaticAllocKiB<<10;
+                    };
+                }
+
 #if BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED)
     #pragma warning(push)
     #pragma warning(disable: 4324)  // warning C4324: structure was padded due to alignment specifier
@@ -47,7 +61,7 @@ namespace alpaka
                              + (sizeBytes%core::vectorization::defaultAlignment>0))*core::vectorization::defaultAlignment)
                     {
 #if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
-                        ALPAKA_ASSERT(sizeBytes <= staticAllocBytes);
+                        ALPAKA_ASSERT(sizeBytes <= staticAllocBytes());
 #endif
                     }
                     //-----------------------------------------------------------------------------
@@ -74,14 +88,15 @@ namespace alpaka
                      */
                     unsigned int staticMemCapacity() const
                     {
-                        return staticAllocBytes - m_dynPitch;
+                        return staticAllocBytes() - m_dynPitch;
                     }
 
                     //! Storage size in bytes
-                    static constexpr unsigned int staticAllocBytes = TStaticAllocKiB<<10;
+                    static constexpr unsigned int staticAllocBytes() {return detail::BlockSharedMemDynMemberStatic<TStaticAllocKiB>::staticAllocBytes;}
 
                 private:
-                    mutable std::array<uint8_t, staticAllocBytes> m_mem;
+
+                    mutable std::array<uint8_t, detail::BlockSharedMemDynMemberStatic<TStaticAllocKiB>::staticAllocBytes> m_mem;
                     unsigned int m_dynPitch;
                 };
 #if BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED)
--- a/include/alpaka/acc/AccCpuOmp2Blocks.hpp
+++ b/include/alpaka/acc/AccCpuOmp2Blocks.hpp
@@ -184,7 +184,7 @@ namespace alpaka
                         // m_threadElemCountMax
                         std::numeric_limits<TIdx>::max(),
                         // m_sharedMemSizeBytes
-                        static_cast<size_t>( acc::AccCpuOmp2Blocks<TDim, TIdx>::staticAllocBytes )};
+                        static_cast<size_t>( acc::AccCpuOmp2Blocks<TDim, TIdx>::staticAllocBytes() )};
                 }
             };
             //#############################################################################
--- a/include/alpaka/acc/AccCpuSerial.hpp
+++ b/include/alpaka/acc/AccCpuSerial.hpp
@@ -178,7 +178,7 @@ namespace alpaka
                         // m_threadElemCountMax
                         std::numeric_limits<TIdx>::max(),
                         // m_sharedMemSizeBytes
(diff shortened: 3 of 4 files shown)


jkelling:
While GCC does not understand, that `static constexpr` data members could be treated just like global constexpr variables, i.e. at compile time, MSVC apparently does not understand `constexpr` functions:
```
D:\a\alpaka\alpaka\include\alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp(85): error C2975: '_Size': invalid template argument for 'std::array', expected compile-time constant expression [D:\a\alpaka\alpaka\build\test\unit\math\math.vcxproj]
  C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\VC\Tools\MSVC\14.16.27023\include\array(20): note: see declaration of '_Size'
  D:\a\alpaka\alpaka\include\alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp(87): note: see reference to class template instantiation 
```
@BenjaminW3 Do you know a fix for this?

jkelling:
Apparently, this is only a problem for `windows_cl-2017`, `windows_cl-2019` works.

BenjaminW3:
Is this a pure c++14 feature? If this is really not supported, we can think about dropping support for MSVC-2017. @sbastrakov What du you think?

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
```diff
@@ -74,14 +74,15 @@ namespace alpaka
...
                     }
 
                     //! Storage size in bytes
-                    static constexpr unsigned int staticAllocBytes = TStaticAllocKiB<<10;
+                    static constexpr unsigned int staticAllocBytes() {return TStaticAllocKiB<<10;}
 
                 private:
-                    mutable std::array<uint8_t, staticAllocBytes> m_mem;
+
+                    mutable std::array<uint8_t, staticAllocBytes()> m_mem;
```
jkelling: MSVC is complaining about this line. With the the function `staticAllocBytes` above only containing a single expression, this is valid C++11 and should not even require C++14.

sbastrakov: I agree it should be working even with C++11. I can take a look locally with my MSVS, but probably some time tomorrow.

sbastrakov: This seems like a VS 2017 bug. A workaround is to explicitly specify the class name for the static method call: instead of `staticAllocBytes()` put `BlockSharedMemDynMember<TStaticAllocKiB>::staticAllocBytes()`. At least on my machine this piece compiles then.

jkelling: Should we just do that then and add a comment, that the class name is given because of a bug in MSVC 2017?

Or should there be compile-time switch testing for MSVC 2017? In this case: how do I check for this MSVC version?

sbastrakov: I think here the lesser evil is to just put type name always and add a comment.
In case you want to go for an explicit test, can use `_MSC_VER` for detection. [Here is a table with values for different MSVS versions](https://dev.to/yumetodo/list-of-mscver-and-mscfullver-8nd)

sbastrakov: Addup: to test, one needs smth like `#if defined(_MSC_VER) && _MSC_VER >= NUMBER_FROM_THE_TABLE`

jkelling: I chose the lesser evil. Thanks.

sbastrakov:
Ugh. Apparently the fix that worked on my Visual Studio does not work for the CI version. Perhaps we would have to resort to an even more ugly workaround, like putting `TStaticAllocKiB<<10` directly into the `std::array`

jkelling:
> Ugh. Apparently the fix that worked on my Visual Studio does not work for the CI version. Perhaps we would have to resort to an even more ugly workaround, like putting `TStaticAllocKiB<<10` directly into the `std::array`

I am very strongly against repeating code because it is very bad for maintainability. I would even prefer a macro to repetition. Since I also prefer to avoid macros, I rather vote for dropping MSVC 2017.

Btw. did you check if the workaround I pushed is exactly the which you tried? Maybe it needs the full qualified name instead of just the class name or something along those lines?

sbastrakov:
@jkelling yes, your version works on my Visual Studio.

Regarding dropping VS 2017. Well, we have a situation where the two compilers have their set of bugs, and we need to workaround something, and potentially drop something else. Then I understand VS 2017 has less priority as a compiler to support. I would not particularly like dropping it, but see the problem.

jkelling:
Pushed a new workaround, which assumes, that the MSVC 2017 bug only affect constexpr functions in template parameter lists and I can thus assign call it to initialize a `static constexp`. Let us see what the CI says.

jkelling:
Why does this not make MSVC 2017 take the `#if` branch?
```c++
#if BOOST_COMP_MSVC && defined(_MSC_VER) && _MSC_VER < 1920
```

jkelling:
Btw,, regarding GCC: This `static constexpr` member thing is not considered a bug by the GCC folks. According to their interpretation of the OpenMP 4.5 standard classes with `static` members are not allowed to be mapped, no exception for `const/constexpr`.

I just noticed, that  we can also work around this using a `static constexpr` member, as long as it is not in a class that we might want to map an instance of to a device. It is totally fine to *use* a `static constexpr` member of a non-mapped type on the device m( .

sbastrakov:
It's same Boost predef confusing names as we saw in another issue.
Should check not just `BOOST_COMP_MSVC` but smth like `#if ( BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED) ) ... ` (the rest of your condition seems good to me).

BenjaminW3:
If you check for `defined(_MSC_VER)` you do not need to check for the `BOOST` variables at all.

jkelling:
MSVC 2017 is passing now.

* a clang 7 test failed with
```
Err:8 http://archive.ubuntu.com/ubuntu bionic-updates/universe amd64 clang-7 amd64 1:7-3~ubuntu0.18.04.1
  Connection failed [IP: 91.189.88.152 80]
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/universe/l/llvm-toolchain-7/clang-7_7-3~ubuntu0.18.04.1_amd64.deb  Connection failed [IP: 91.189.88.152 80]
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```
* a gcc 7 failed with
```
E: Failed to fetch http://archive.ubuntu.com/ubuntu/pool/main/g/gsettings-desktop-schemas/gsettings-desktop-schemas_3.36.0-1ubuntu1_all.deb  Undetermined Error [IP: 91.189.88.152 80]
Fetched 48.2 MB in 1min 30s (538 kB/s)
E: Unable to fetch some archives, maybe run apt-get update or try with --fix-missing?
```

jkelling:
@sbastrakov Can you merge this one? The two failing tests are CI hickups.

sbastrakov:
I agree they are hiccups. But I can't merge until it passes. However @BenjaminW3 can, and @psychocoderHPC as well but he is on vacation.

BenjaminW3:
I am also on vacation😋

PR #1119, opened 2020-08-13
cpu/SysInfo.hpp: Add #else for cpuid; Add PGI

jkelling:
Adds `#else` both for unknown compilers and non-x86, so the difference can be reported by `getCpuName()`.

PR #1120, opened 2020-08-13
Pgi std atomic workaround

jkelling:
This issue seems to have appeared with PGI before, but I noticed it first with NVHPC 20.7.
https://forums.developer.nvidia.com/t/support-for-atomic-in-libstdc-missing/135403/2

Not sure if we should put it this workaround in the code or if we should just wait for the next version...

### DIFF
--- a/test/catch_main/CMakeLists.txt
+++ b/test/catch_main/CMakeLists.txt
@@ -30,6 +30,11 @@ set_target_properties(CatchMain PROPERTIES
 )
 
 target_compile_definitions(CatchMain PUBLIC "CATCH_CONFIG_FAST_COMPILE")
+if (CMAKE_CXX_COMPILER_ID STREQUAL "PGI")
+    # Workaround for STL atomic issue: https://forums.developer.nvidia.com/t/support-for-atomic-in-libstdc-missing/135403/2
+    # still appears in NVHPC 20.7
+    target_compile_definitions(CatchMain PUBLIC "__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1")
+endif()
 if(MSVC)
     target_compile_definitions(CatchMain PUBLIC "CATCH_CONFIG_WINDOWS_CRTDBG")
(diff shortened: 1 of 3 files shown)


jkelling:
Only merge after https://github.com/alpaka-group/alpaka/pull/1119 .

test/catch_main/CMakeLists.txt
```diff
@@ -30,6 +30,10 @@ set_target_properties(CatchMain PROPERTIES
 )
 
 target_compile_definitions(CatchMain PUBLIC "CATCH_CONFIG_FAST_COMPILE")
+if (CMAKE_CXX_COMPILER_ID STREQUAL "PGI")
+    # Workaround for STL atomic issue: https://forums.developer.nvidia.com/t/support-for-atomic-in-libstdc-missing/135403/2
+    target_compile_definitions(CatchMain PUBLIC "__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1")
```
BenjaminW3, line 36: Why do we do both? Shouldn't it be enough to set it either with the build system OR within the code?
Can't we do this directly for the alpaka target publicly?

jkelling: I set it in the code so it works for the Alpaka stand-alone headers.

The flag in the build system only affects the tests, which is required because catch2 itself also includes `<atomic>` and I did not want to touch the catch2 code.

jkelling: An alternative would be to put it into the Alpaka target and the stand-alone headers instead of where `<atomic>` is included in the code.

jkelling: I changed it just add the thing to the alpaka target interface and the CatchMain target. Any other Program using alpaka is likely to include `<atomic>` somewhere else, maybe indirectly through a different library, thus fixing it in the stand-alone headers does not seem very useful after all.

BenjaminW3 (approved):

sbastrakov (approved):

PR #1122, opened 2020-08-14
Add IntrinsicFallback: basic fallback implementations

jkelling:
For OpenMP 5.0 target offload  and OpenACC we need basic implementations of these functions. There is currently no simple way of determining what target device code is being compiled for, we cannot safely decide which compiler buildins to call if any.

This PR adds the `IntrinsicFallback` implementation of the `Intrinsic` concept.

It uses the `ffsFallback` code from `IntrinsicCpu`. The actual implementations are located in a `detail` namespace instead of in the class. `IntrinsicCpu` is changed to use `ffsFallback` from there.

### DIFF
--- a/include/alpaka/intrinsic/IntrinsicCpu.hpp
+++ b/include/alpaka/intrinsic/IntrinsicCpu.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/BoostPredef.hpp>
 #include <alpaka/intrinsic/Traits.hpp>
+#include <alpaka/intrinsic/IntrinsicFallback.hpp>
 
 #include <bitset>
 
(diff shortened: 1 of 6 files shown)


jkelling:
@sbastrakov this is probably mostly interesting for your, since you are working on the intrinsics.

jkelling:
Second commit fixes a warning in gcc 7.5:
```
test/unit/warp/src/Any.cpp:110:78: error: "ALPAKA_ACC_CPU_BT_OMP4_ENABLED" is not defined, evaluates to 0 [-Werror=undef]
 #if BOOST_COMP_GNUC && (BOOST_COMP_GNUC == BOOST_VERSION_NUMBER(7, 5, 0)) && ALPAKA_ACC_CPU_BT_OMP4_ENABLED
                                                                              ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

I do not think this deserves its own PR. Just noticed it when checking some other warnings I got in the CI.

sbastrakov (approved):
Nice!

PR #1125, opened 2020-08-18
Improvements to vectorAdd

jkelling:
This PR Contains different improvement for these two examples:

## ~~helloWorldLambda~~
~~Use a generic lambda: The contract for kernels is, or should be, that the functor passed to the `createTaskKernel`, or similar, function is a template with respect to the accelerator type. This is required for the OpenACC backend, because the context for threads cannot be instanciated on the host and thus is different from the accelerator type, albeit derived from it.~~
Not supported by nvcc with "--expt-extended-lambda".

## vectorAdd
Mostly cosmetic changes:
* Add print time for kernel and copy HtoD.
* Add a limit for the number of printed misscomparing values (20) (very helpful when debugging backends). Still count the number of wrong values and print the count, too.
* `threadElemExtent = 8u` : allows compiler to vectorise, which makes for a better example usage

### DIFF
--- a/example/vectorAdd/src/vectorAdd.cpp
+++ b/example/vectorAdd/src/vectorAdd.cpp
@@ -21,6 +21,7 @@
 #include <random>
 #include <iostream>
 #include <typeinfo>
+#include <chrono>
 
 //#############################################################################
 //! A vector addition kernel.
@@ -116,7 +117,7 @@ auto main()
 
     // Define the work division
     Idx const numElements(123456);
-    Idx const elementsPerThread(3u);
+    Idx const elementsPerThread(8u);
     alpaka::vec::Vec<Dim, Idx> const extent(numElements);
 
     // Let alpaka calculate good block and grid sizes given our full problem extent
@@ -182,13 +183,25 @@ auto main()
         numElements));
 
     // Enqueue the kernel execution task
-    alpaka::queue::enqueue(queue, taskKernel);
+    {
+        const auto beginT = std::chrono::high_resolution_clock::now();
+        alpaka::queue::enqueue(queue, taskKernel);
+        alpaka::wait::wait(queue); // wait in case we are using an asynchronous queue to time actual kernel runtime
+        const auto endT = std::chrono::high_resolution_clock::now();
+        std::cout << "Time for kernel execution: " << std::chrono::duration<double>(endT-beginT).count() << 's' << std::endl;
+    }
 
     // Copy back the result
-    alpaka::mem::view::copy(queue, bufHostC, bufAccC, extent);
-    alpaka::wait::wait(queue);
+    {
+        auto beginT = std::chrono::high_resolution_clock::now();
+        alpaka::mem::view::copy(queue, bufHostC, bufAccC, extent);
+        alpaka::wait::wait(queue);
+        const auto endT = std::chrono::high_resolution_clock::now();
+        std::cout << "Time for HtoD copy: " << std::chrono::duration<double>(endT-beginT).count() << 's' << std::endl;
+    }
 
-    bool resultCorrect(true);
+    int falseResults = 0;
+    static constexpr int MAX_PRINT_FALSE_RESULTS = 20;
     for(Idx i(0u);
         i < numElements;
         ++i)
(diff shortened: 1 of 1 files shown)


example/vectorAdd/src/vectorAdd.cpp
```diff
@@ -182,13 +183,23 @@ auto main()
         numElements));
 
     // Enqueue the kernel execution task
-    alpaka::queue::enqueue(queue, taskKernel);
+    {
+        const auto beginT = std::chrono::high_resolution_clock::now();
+        alpaka::queue::enqueue(queue, taskKernel);
+        const auto endT = std::chrono::high_resolution_clock::now();
+        std::cout << "Time for kernel execution: " << std::chrono::duration<double>(endT-beginT).count() << 's' << std::endl;
+    }
```
sbastrakov: This is only for the blocking queue. Which it is in the example, but a user may change that when playing around.

example/vectorAdd/src/vectorAdd.cpp
```diff
@@ -182,13 +183,23 @@ auto main()
...
+    {
+        auto beginT = std::chrono::high_resolution_clock::now();
+        alpaka::mem::view::copy(queue, bufHostC, bufAccC, extent);
+        const auto endT = std::chrono::high_resolution_clock::now();
+        std::cout << "Time for HtoD copy: " << std::chrono::duration<double>(endT-beginT).count() << 's' << std::endl;
+    }
 
-    bool resultCorrect(true);
+    int falseResults = 0;
+    static constexpr int NUM_FALSE_RESULTS = 20;
```
sbastrakov: Perhaps `MAX_NUM_FALSE_RESULTS` would be more descriptive here?

sbastrakov:
Since it uses commits of #1118 which is not merged yet, I added [WIP].

example/vectorAdd/src/vectorAdd.cpp
jkelling: It is not a maximum for `falseResults`, but the maximum number to be printed. Maybe `MAX_PRINT_FALSE_RESULTS`?

sbastrakov: I like your suggestion

jkelling:
Apparently, generic lambdas do not always work in cuda 9.1? The job linux_nvcc-9.1_gcc-5_debug_analysis failed with:
```
/home/runner/work/alpaka/alpaka/example/helloWorldLambda/src/helloWorldLambda.cpp(129): error #3002-D: __host__ __device__ extended lambdas cannot be generic lambdas
```
but linux_clang-8_cuda-9.1_release succeeded.

Many other jobs are currently failing with:
```
+ wget --no-verbose https://cmake.org/files/v3.17/cmake-3.17.3-Linux-x86_64.tar.gz
Unable to establish SSL connection.
```

jkelling:
The generic lambda problem seems to only exist in cuda.9.1. 9.2 works: https://github.com/alpaka-group/alpaka/pull/1125/checks?check_run_id=998076582 .

jkelling:
The instance in which the generic lambda works with nvcc is linux_nvcc-9.2_gcc-7_release_extended_lambda_off , which appears to to be the only one where `--expt-extended-lambd` is turned off. In all other cases it fails, even with cuda 11.

I will remove the `helloWorldLambda` commit from this PR then and disable the example fro OpenACC.

example/vectorAdd/src/vectorAdd.cpp
jkelling: Added `alpaka::wait::wait(queue)`, here and for the copy further down.

jkelling:
Again, two random CI fails:
```
+ wget --no-verbose https://cmake.org/files/v3.16/cmake-3.16.8-Linux-x86_64.tar.gz
failed: Connection timed out.
failed: Connection timed out.
[...]
failed: Connection timed out.
##[error]Process completed with exit code 4.
```
and one during setup:
```
Download action repository 'actions/checkout@v1'
##[warning]Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/v1'. Error: Response status code does not indicate success: 401 (Unauthorized).
##[warning]Back off 20.357 seconds before retry.
##[warning]Failed to download action 'https://api.github.com/repos/actions/checkout/tarball/v1'. Error: Response status code does not indicate success: 401 (Unauthorized).
##[warning]Back off 29.732 seconds before retry.
##[error]Response status code does not indicate success: 401 (Unauthorized).
```

I prefer not to run it again, because there are other CI workflows to run.

sbastrakov (approved):

BenjaminW3 (approved):

PR #789, opened 2019-05-15
Add experimental SYCL backend

j-stephan:
This PR is intended to provide Alpaka with a working SYCL backend. In the beginning this work will focus on (Xilinx) FPGAs but will hopefully support other device types as well in the future.

Open problems that require discussion:

- [x] Atomics require a special SYCL data type (e.g. `atomic<float>`). This is different from CUDA and the other backends as far as I can see and might require substantial work to abstract the differences away.
- [x] SYCL's inter-thread synchronisation routines are not as elaborate as CUDA's. We can only mirror `__syncthreads` but probably not the rest.
- [x] Alpaka really loves raw pointers. This is fine on all existing backends right now but SYCL uses `accessors` for kernel data access which can be transformed into a `multi_ptr` class type if required. I'm not quite sure yet how this can be implemented in a way that doesn't break Alpaka.
- [x] SYCL doesn't permit direct passing of parameters to kernels. Instead kernels are written as either `operator()` or lambdas and parameters are passed as class members or are captured by the lambda. I haven't figured out yet if this approach works with the Alpaka way.

Further tasks:

- [x] Complete the backend!
- [x] Support all SYCL device types (`host`, `cpu`, `gpu`, `accelerator`, `custom`) and additional selectors (`automatic`, `all`)
- [x] Compile and execute the vector addition example.
- [x] Add CMake support. There are multiple SYCL implementations which can happily co-exist so this should be considered in the `CMakeLists.txt`

PR #1126, opened 2020-08-19
OpenMP 5 target offload backend

jkelling:
This PR turns the existing OpenMP4 backend into one with target offload and renames it.

## AccCpuOmp4 -> AccOmp5

* Practically, most of this might also work with OpenMP 4.0, depending on how the compiler interprets the standard. Also there is currently no compiler which actually supports OpenMP 5.0 (Clang 11 claims to, but the support is incomplete). Some compilers claiming OpenMP 4.5/4.0 may work partially, or we can push their interpretation of the standard into a direction that works for us. This is why the `_OPENMP` guards are unchanged (still OpenMP 4.0). Technically, C++ TMP requires OpenMP 5.0 to undisputably work and there are some other problems in the earlier versions, which is why the backend is named Omp5.
* Since this can run on any device, I removed the `Cpu`(`Gpu`) part alltogether and replaced this bit by `ANY` in the cmake option.

## CMake

* Adds option (string) `ALPAKA_OFFLOAD_MAX_BLOCK_SIZE` which set the number blocks per thread that the backend reports as maximum.
  * If the kernels are invoked with more threads per block than the runtime will actually run, some work will not get done (=> wrong results). There is a check in place from the old Omp4 backend in ALPAKA_DEBUG mode printing a warning. Some runtimes will also warn about this.
* Adds option (bool) `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST` which tells the code, that the target it is compuled for is the host, thus allowing things like asserts in target code. There is no OpenMP standard way for the code to get information about the target at compile that that I am aware of.

## CI

* Because compiler support is very limited, only this backend can only be tested with clang 10 in the CI. I changed the `linux_clang-10_release` job to include it and added `-omp5` to the name to mark it.
* Had to change the clang install script to pull in the requirements for target offload for versions >= 10.
  * Because some clang versions are given as e.g. "4.0" in ci.yml, I had to use this condition to check the version:
    https://github.com/jkelling/alpaka/blob/4c2de74a549b3e998b0cfb0bf8eea860d7bdc298/script/install_clang.sh#L21
      ```bash
     [ "$( echo "${ALPAKA_CI_CLANG_VER} < 10" | bc )" == 1 ]
     ```
    Should one always assume, that version numbers can be float when comparing, or should we rather, or can we even, change version number to be only major version ints?
* `ALPAKA_OFFLOAD_MAX_BLOCK_SIZE=1` because the runtime in the CI will not run more than one thread per team even though `OMP_NUM_THREADS=4` and it is running multiple teams.

## Backend Properties

* `AccCpuOmp4` -> `AccOmp5`
* Adds `DevOmp5`, to handle OMP device ids. Also `PltfOmp5`
* Adds `BufOmp5` for OMP target allocations.
* Currently uses `QueueGenericThreads*<DevOmp5>`. An asynchronous queue could later be implemented based on OMP tasking constructs.

## Tests

* Turns `cpu::detail::EventHostManualTriggerCpuImpl` into a template with respect to the device: `DevCpu` or `DevOmp5`.


jkelling:
The CI job "linux_clang-10_release" is not going to happen, because I renamed it to "linux_clang-10_release-omp5", why is this listed as required?

BenjaminW3:
The selection which jobs are required is part of repository settings. They are mentioned there by name.
This is somehow fragile but it is what github offers at the moment. When we want to rename a job, we have to change the settings as well. I can do this, once this PR is ready or has been merged.

BenjaminW3:
Clang changed the scheme of their ubuntu packet names from two numbers seperated by a dot (`clang-4.0`) to single number (`clang-9`). That is the reason for the different numbers in the yml file. `4.0` without quotes seems to be interpreted as floating point number and forwarded as `4` which does not match the packet names anymore.

BenjaminW3:
Could we use this PR as a general discussion board but split off some uncritical changes into independent PRs that can easily be merged?
The first change I would like to get merged independently is the change you did to `EventHostManualTriggerCpuImpl`. I may find more things later.

thirdParty/catch2/include/catch2/catch.hpp
```diff
@@ -11392,6 +11393,8 @@ FP step(FP start, FP direction, uint64_t steps) {
 // But without the subtraction to allow for INFINITY in comparison
 bool marginComparison(double lhs, double rhs, double margin) {
     return (lhs + margin >= rhs) && (rhs + margin >= lhs);
+    // auto ulpDiff = std::abs((FP)(lc.i - rc.i));
+    // return ulpDiff <= maxUlpDiff;
```
BenjaminW3: Could you pleas remove those comments?

thirdParty/catch2/include/catch2/catch.hpp
```diff
@@ -11469,14 +11472,14 @@ namespace Floating {
...
-            write(ret, step(m_target, static_cast<double>( INFINITY), m_ulps));
+            write(ret, step(m_target,  std::numeric_limits<double>::infinity(), m_ulps));
         } else {
             // We have to cast INFINITY to float because of MinGW, see #1782
-            write(ret, step(static_cast<float>(m_target), static_cast<float>(-INFINITY), m_ulps));
-            ret << ", ";
-            write(ret, step(static_cast<float>(m_target), static_cast<float>( INFINITY), m_ulps));
+            write(ret, step(static_cast<float>(m_target), -std::numeric_limits<float>::infinity(), m_ulps));
+            ret << ", ";                                                                        
+            write(ret, step(static_cast<float>(m_target),  std::numeric_limits<float>::infinity(), m_ulps));
```
BenjaminW3: Could you create a PR with those changes to the original catch2 repository?

thirdParty/catch2/include/catch2/catch.hpp
```diff
@@ -1298,7 +1299,7 @@ struct AutoReg : NonCopyable {
                 void reg_tests(){\
                     int index = 0;\
                     using expander = int[];\
-                    (void)expander{(Catch::AutoReg( Catch::makeTestInvoker( &TestName<Types>::test ), CATCH_INTERNAL_LINEINFO, #ClassName, Catch::NameAndTags{ Name " - " + std::string(INTERNAL_CATCH_STRINGIZE(TmplList)) + " - " + std::to_string(index), Tags } ), index++, 0)... };/* NOLINT */ \
+                    (void)expander{0, (Catch::AutoReg( Catch::makeTestInvoker( &TestName<Types>::test ), CATCH_INTERNAL_LINEINFO, #ClassName, Catch::NameAndTags{ Name " - " + std::string(INTERNAL_CATCH_STRINGIZE(TmplList)) + " - " + std::to_string(index), Tags } ), index++, 0)... };/* NOLINT */ \
```
BenjaminW3: What exactly is this doing?

thirdParty/catch2/include/catch2/catch.hpp
```diff
@@ -11342,7 +11343,7 @@ namespace {
             return lhs == rhs;
         }
 
-        auto ulpDiff = std::abs(lc - rc);
+        auto ulpDiff = std::abs((FP)(lc - rc));
```
BenjaminW3: This should probably also be added to the catch2 original.

test/common/devCompileOptions.cmake
```diff
@@ -139,6 +139,10 @@ ELSE()
         # as they are stored as members. Therefore, the padding warning is triggered by the calling code
         # and does not indicate a failure within alpaka.
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wno-padded")
+        # Triggers for all instances of ALPAKA_DEBUG_MINIMAL_LOG_SCOPE and similar macros followed by semicolon
+        IF(CLANG_VERSION_MAJOR GREATER 7)
+            LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wno-extra-semi-stmt")
+        ENDIF()
```
BenjaminW3: Can we add this in a separate PR?

script/install_clang.sh
```diff
@@ -18,7 +18,13 @@ source ./script/set.sh
 : "${ALPAKA_CI_STDLIB?'ALPAKA_CI_STDLIB must be specified'}"
 : "${CXX?'CXX must be specified'}"
 
-travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install clang-${ALPAKA_CI_CLANG_VER}
+if [ "$( echo "${ALPAKA_CI_CLANG_VER} < 10" | bc )" == 1 ]
+then
+    travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install clang-${ALPAKA_CI_CLANG_VER}
+else
+    travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install \
+        clang-tools-${ALPAKA_CI_CLANG_VER} llvm-${ALPAKA_CI_CLANG_VER} libomp-${ALPAKA_CI_CLANG_VER}-dev
```
BenjaminW3: We are already installing `libomp-dev` below now (this probably was not the case when you created this branch).
We should probably only install one version.

include/alpaka/queue/QueueOmp5NonBlocking.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#include <alpaka/core/Omp5.hpp>
+#include <alpaka/queue/QueueGenericThreadsNonBlocking.hpp>
+
+#include <stdexcept>
+#include <memory>
+#include <functional>
+#include <mutex>
+#include <condition_variable>
+#include <thread>
```
BenjaminW3: I am pretty sure we can get rid of all (really all) includes except `QueueGenericThreadsNonBlocking.hpp` here and in the blocking version.

include/alpaka/idx/MapIdx.hpp
```diff
@@ -157,5 +157,138 @@ namespace alpaka
...
+            static_assert(TidxDimIn > 0u, "The dimension of the input vector has to be greater than zero!");
+
+            return
+                detail::MapIdxPitch<
+                    TidxDimOut,
+                    TidxDimIn>
+                ::mapIdxPitch(
+                    idx,
+                    pitch);
+        }
```
BenjaminW3: What exactly differentiates those `pitch` versions from the `extent` versions?
It might be confusing for users to see those new `pitch` versions.
We should discuss those additions together with some unit tests in a separate PR.

test/common/include/alpaka/test/acc/TestAccs.hpp
```diff
@@ -108,16 +108,29 @@ namespace alpaka
                     typename TIdx>
                 using AccCpuOmp2ThreadsIfAvailableElseInt = int;
 #endif
-#if defined(ALPAKA_ACC_CPU_BT_OMP4_ENABLED) && !defined(ALPAKA_CUDA_CI)
+#if defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) && !defined(TEST_UNIT_KERNEL_KERNEL_STD_FUNCTION) \
+        && !( BOOST_COMP_GNUC && ( \
+                /* tests excluded because of GCC10 Oacc / Omp5 target symbol bug with multiple units */ \
```
BenjaminW3: What exactly is the issue here?

test/unit/rand/src/RandTest.cpp
```diff
@@ -90,6 +90,9 @@ class RandTestKernel
 
 #if !defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && \
   !defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#ifndef ALPAKA_ACC_ANY_BT_OMP5_ENABLED
+        // TODO: These ifdefs are wrong: They will reduce the test to the
+        // smallest common denominator from all enabled backends
```
BenjaminW3, line 95: This is what I would expect. If we configured alpaka with mulitple backends, only the tests which are supported by all of them can run.

cmake/alpakaCommon.cmake
```diff
@@ -234,24 +235,27 @@ endif()
...
-        if(ALPAKA_ACC_CPU_BT_OMP4_ENABLE)
+        # Clang versions starting from 3.9 support OpenMP 5.0 only when given the corresponding flag
+        if(ALPAKA_ACC_ANY_BT_OMP5_ENABLE)
             target_link_options(alpaka INTERFACE $<$<CXX_COMPILER_ID:AppleClang,Clang>:-fopenmp-version=40>)
         endif()
     else()
         message(FATAL_ERROR "Optional alpaka dependency OpenMP could not be found!")
+        set(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OFF CACHE BOOL "Enable the OpenMP 2.0 CPU grid block back-end" FORCE)
+        set(ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OFF CACHE BOOL "Enable the OpenMP 2.0 CPU block thread back-end" FORCE)
+        set(ALPAKA_ACC_ANY_BT_OMP5_ENABLE OFF CACHE BOOL "Enable the OpenMP 5.0 CPU block and thread back-end" FORCE)
```
BenjaminW3: Please remove. This is not necessary anymore as we error now above.

cmake/alpakaCommon.cmake
```diff
@@ -234,24 +235,27 @@ endif()
...
+if(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR ALPAKA_ACC_ANY_BT_OMP5_ENABLE)
     find_package(OpenMP)
 
     if(OpenMP_CXX_FOUND)
-        if(ALPAKA_ACC_CPU_BT_OMP4_ENABLE)
-            if(OpenMP_CXX_VERSION VERSION_LESS 4.0)
-                message(FATAL_ERROR "OpenMP 4.0 is required but not supported!")
+         if(ALPAKA_ACC_ANY_BT_OMP5_ENABLED)
+           if(OpenMP_CXX_VERSION VERSION_LESS 4.0)
+               set(ALPAKA_ACC_ANY_BT_OMP5_ENABLE OFF CACHE BOOL "Enable the OpenMP 5.0 CPU block and thread back-end" FORCE)
```
BenjaminW3: We now error when a requested backend is not available. Please adapt this line here.

README_OMP5.md
```diff
@@ -0,0 +1,190 @@
+# Building configuring for the OMP5 backend
+
+To make the build system enable the OpenMP5 backend (internally called
+`CPU_BT_OMP4` for now), one has to tell CMake explicitly about the OpenMP
```
BenjaminW3: `CPU_BT_OMP4` still true?

README_OMP5.md
```diff
@@ -0,0 +1,190 @@
...
+cmake -DOpenMP_CXX_VERSION=5 \
+  -DALPAKA_ACC_ANY_BT_OMP5_ENABLE=on \
+  -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=on \
+  -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=off \
+  -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=off \
+  -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=off \
+  -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE=off \
+  -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE=off \
+  -DALPAKA_ACC_GPU_CUDA_ENABLE=off \
+  -DALPAKA_ACC_GPU_HIP_ENABLE=off \
```
BenjaminW3: It should not be necessary anymore to disable any backends on the command line now anymore. This is the default now. 
`cmake -DOpenMP_CXX_VERSION=5 -DALPAKA_ACC_ANY_BT_OMP5_ENABLE=ON` should be enough.

.github/workflows/ci.yml
```diff
@@ -64,7 +64,7 @@ env:
   ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: ON
   ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: ON
   ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: ON
-  ALPAKA_ACC_CPU_BT_OMP4_ENABLE: ON
+  ALPAKA_ACC_ANY_BT_OMP5_ENABLE: OFF
```
BenjaminW3, line 67: As you are defaulting to `OFF` here, you can remove all the other mentions of `ALPAKA_ACC_ANY_BT_OMP5_ENABLE: OFF` below.

thirdParty/catch2/include/catch2/catch.hpp
jkelling: It makes sure the list is never empty. This is a workaround for a bug in IBM XL, which cannot deal with empty initializer lists. See: https://github.com/alpaka-group/alpaka/pull/1126/commits/8da30a1429ef8df4b820c0c3a5e4d784d79f1306

test/common/devCompileOptions.cmake
jkelling: https://github.com/alpaka-group/alpaka/pull/1131

include/alpaka/idx/MapIdx.hpp
jkelling: They use pitches in order to work with pitched views. They are used by a memset kernel which is also part of this PR. I created a single PR for both, because the belong together: https://github.com/alpaka-group/alpaka/pull/1132

thirdParty/catch2/include/catch2/catch.hpp
jkelling: Removed.

include/alpaka/queue/QueueOmp5NonBlocking.hpp
jkelling: Also need `DevOmp5.hpp`, removed the rest.

cmake/alpakaCommon.cmake
jkelling: Removed.

README_OMP5.md
jkelling: No, this got changed during renaming. The interjection in parenthesis is obsolete now, removed.

README_OMP5.md
jkelling: Also need to add `-DBUILD_TESTING=on -Dalpaka_BUILD_EXAMPLES=on` now (the README is currently for testers not users). Updated.

.github/workflows/ci.yml
jkelling: Done.

test/unit/rand/src/RandTest.cpp
jkelling: On the level of whole tests, they are only disabled for single backends, i.e. the test will still be done for the backends which support it even of non-supporting backend are enabled. This did not seem viable at the level of granularity, but I am not sure, hence the discussion.

cmake/alpakaCommon.cmake
jkelling: > The selection which jobs are required is part of repository settings. They are mentioned there by name.
> This is somehow fragile but it is what github offers at the moment. When we want to rename a job, we have to change the settings as well. I can do this, once this PR is ready or has been merged.

Ok. My rationale about changing this name is, that this is the only job with omp5 enabled, and will be one of only a few for a while. If for a future PR clang-10 fails while others succeed, it may be good to see, that it includes the omp5  backend.

jkelling:
> The selection which jobs are required is part of repository settings. They are mentioned there by name.
> This is somehow fragile but it is what github offers at the moment. When we want to rename a job, we have to change the settings as well. I can do this, once this PR is ready or has been merged.

Ok. My rationale about changing this name is, that this is the only job with omp5 enabled, and will be one of only a few for a while. If for a future PR clang-10 fails while others succeed, it may be good to see, that it includes the omp5 backend.

test/common/include/alpaka/test/acc/TestAccs.hpp
jkelling: For example, the intrinsic test produces the following linker errors (both with OpenMP5 target and OpenACC, the example below is OpenACC):
```
`_ZNK6alpaka6kernel14TaskKernelOaccISt17integral_constantImLm1EElNS_3mem4view12MemSetKernelEJRKhPhRNS_3vec3VecIS3_lEESD_EEclERKNS_3dev7DevOaccE._omp_fn.0' referenced in section `.gnu.offload_funcs' of CMakeFiles/intrinsic.dir/src/Popcount.cpp.o: defined in discarded section `.text._ZNK6alpaka6kernel14TaskKernelOaccISt17integral_constantImLm1EElNS_3mem4view12MemSetKernelEJRKhPhRNS_3vec3VecIS3_lEESD_EEclERKNS_3dev7DevOaccE._omp_fn.0[_ZNK6alpaka6kernel14TaskKernelOaccISt17integral_constantImLm1EElNS_3mem4view12MemSetKernelEJRKhPhRNS_3vec3VecIS3_lEESD_EEclERKNS_3dev7DevOaccE]' of CMakeFiles/intrinsic.dir/src/Popcount.cpp.o
```
There are two compilarion units in this order:
1. Ffs.cpp.o
2. Popcount.cpp.o

MemSetKernel is used in both, the symbols are missing for the second unit. They are always missing for all but the first unit.

I just tested this with a fresh gcc master build. There it appears fixed, so we can expect this to be fixed in GCC 11. Adding another check... .

script/install_clang.sh
jkelling: Actualy, I missed this bit. Changed the install script to rely less on the clang version, but take into account whether OMP5 is enbabled: https://github.com/alpaka-group/alpaka/pull/1126/commits/912f6db4e27beb5b2c3840857dee96c44af7a493

test/common/include/alpaka/test/acc/TestAccs.hpp
jkelling: New check: https://github.com/alpaka-group/alpaka/pull/1126/commits/d20db3eeb5bc541aa7e234bb3ea273e2f0e2d489

cmake/alpakaCommon.cmake
jkelling: Done. Also added a warning for OpenMP < 5.0: https://github.com/alpaka-group/alpaka/pull/1126/commits/7fcec9209569693dfd5755a377581eb16663e5e4

jkelling:
> Could we use this PR as a general discussion board but split off some uncritical changes into independent PRs that can easily be merged?
> The first change I would like to get merged independently is the change you did to `EventHostManualTriggerCpuImpl`. I may find more things later.

https://github.com/alpaka-group/alpaka/pull/1133

cmake/alpakaCommon.cmake
```diff
@@ -234,20 +235,22 @@ endif()
...
+            elseif(OpenMP_CXX_VERSION VERSION_LESS 5.0)
+                message(WARNING "OpenMP < 5.0, for ALPAKA_ACC_ANY_BT_OMP5_ENABLE 5.0 is recommended.")
             endif()
         endif()
 
         target_link_libraries(alpaka INTERFACE OpenMP::OpenMP_CXX)
 
-        # Clang versions starting from 3.9 support OpenMP 4.0 only when given the corresponding flag
-        if(ALPAKA_ACC_CPU_BT_OMP4_ENABLE)
+        # Clang versions starting from 3.9 support OpenMP 5.0 only when given the corresponding flag
```
BenjaminW3:
```suggestion
        # Clang versions starting from 3.9 support OpenMP 4.0 and higher only when given the corresponding flag
```

cmake/alpakaCommon.cmake
```diff
@@ -234,20 +235,22 @@ endif()
...
+        if(ALPAKA_ACC_ANY_BT_OMP5_ENABLE)
             target_link_options(alpaka INTERFACE $<$<CXX_COMPILER_ID:AppleClang,Clang>:-fopenmp-version=40>)
```
BenjaminW3, line 254: Do we need to use `-fopenmp-version=50`? But which clang versions support this?
As far as I understood your PR, we do not use OpenMP 5.0 features, so we do not need to increase the version yet.

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -58,36 +55,36 @@ namespace alpaka
...
                 atomic::AtomicOmpBuiltIn     // thread atomics
             >,
             public math::MathStdLib,
-            public block::shared::dyn::BlockSharedMemDynAlignedAlloc,
-            public block::shared::st::BlockSharedMemStMasterSync,
+            public block::shared::dyn::BlockSharedMemDynMember<>,
+            public block::shared::st::BlockSharedMemStOmp5,
             public block::sync::BlockSyncBarrierOmp,
-            public intrinsic::IntrinsicCpu,
+            // cannot determine which intrinsics are save to use (depends on target), using fallback
```
BenjaminW3:
```suggestion
            // cannot determine which intrinsics are safe to use (depends on target), using fallback
```

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -147,36 +139,45 @@ namespace alpaka
...
                 //-----------------------------------------------------------------------------
                 ALPAKA_FN_HOST static auto getAccDevProps(
-                    dev::DevCpu const & dev)
+                    dev::DevOmp5 const & dev)
                 -> acc::AccDevProps<TDim, TIdx>
                 {
+                    alpaka::ignore_unused(dev);
+
+#if defined(ALPAKA_OFFLOAD_MAX_BLOCK_SIZE) && ALPAKA_OFFLOAD_MAX_BLOCK_SIZE>0
+                    auto const blockThreadCount = ALPAKA_OFFLOAD_MAX_BLOCK_SIZE;
```
BenjaminW3: Is `ALPAKA_OFFLOAD_MAX_BLOCK_SIZE` meant to only reduce the `blockThreadCount` when there are too many available?
Should we then do `std::min(::omp_get_max_threads(), ALPAKA_OFFLOAD_MAX_BLOCK_SIZE)`?

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -147,36 +139,45 @@ namespace alpaka
...
+                    auto const blockThreadCount = ::omp_get_max_threads();
+#endif
 #ifdef ALPAKA_CI
-                    auto const blockThreadCountMax(alpaka::core::clipCast<TIdx>(std::min(4, ::omp_get_max_threads())));
+                    auto const blockThreadCountMax(alpaka::core::clipCast<TIdx>(std::min(4, blockThreadCount)));
+                    auto const gridBlockCountMax(alpaka::core::clipCast<TIdx>(std::min(4, ::omp_get_max_threads())));
 #else
-                    auto const blockThreadCountMax(alpaka::core::clipCast<TIdx>(::omp_get_max_threads()));
+                    auto const blockThreadCountMax(alpaka::core::clipCast<TIdx>(blockThreadCount));
+                    auto const gridBlockCountMax(alpaka::core::clipCast<TIdx>(::omp_get_max_threads())); //! \todo fix max block size for target
```
BenjaminW3: What do you want to do with the TODO here?

include/alpaka/core/Omp5.hpp
```diff
@@ -0,0 +1,72 @@
...
+#include <alpaka/offset/Traits.hpp>
+#include <alpaka/extent/Traits.hpp>
+#include <alpaka/idx/Traits.hpp>
+#include <alpaka/vec/Vec.hpp>
+#include <alpaka/meta/IntegerSequence.hpp>
+#include <alpaka/meta/Metafunctions.hpp>
+
+#include <omp.h>
+
+#include <array>
```
BenjaminW3: where do we need `<array>` here?

include/alpaka/core/Omp5.hpp
```diff
@@ -0,0 +1,72 @@
...
+    #error If ALPAKA_ACC_ANY_BT_OMP5_ENABLED is set, the compiler has to support OpenMP 4.0 or higher!
+#endif
+
+#include <alpaka/elem/Traits.hpp>
+#include <alpaka/offset/Traits.hpp>
+#include <alpaka/extent/Traits.hpp>
+#include <alpaka/idx/Traits.hpp>
+#include <alpaka/vec/Vec.hpp>
+#include <alpaka/meta/IntegerSequence.hpp>
+#include <alpaka/meta/Metafunctions.hpp>
```
BenjaminW3: This block of includes also seems unused.

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -0,0 +1,365 @@
...
+            std::shared_ptr<omp5::detail::DevOmp5Impl> m_spDevOmp5Impl;
+        };
+    }
+
+    namespace dev
+    {
+        namespace traits
+        {
+            //#############################################################################
+            //! The CUDA RT device name get trait specialization.
```
BenjaminW3: There are multiple `CUDA RT` in the comments in this file.

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -0,0 +1,365 @@
...
+                dev::DevOmp5>
+            {
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_HOST static auto getFreeMemBytes(
+                    dev::DevOmp5 const & dev)
+                -> std::size_t
+                {
+                    alpaka::ignore_unused(dev); //! \todo query device
+                    std::size_t freeInternal((6ull<<30));
+                    // std::size_t totalInternal(0u);
```
BenjaminW3: Any ideas here and above?

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -0,0 +1,365 @@
...
+            {
+                //-----------------------------------------------------------------------------
+                ALPAKA_FN_HOST static auto currentThreadWaitFor(
+                    dev::DevOmp5 const & dev)
+                -> void
+                {
+                    ALPAKA_DEBUG_FULL_LOG_SCOPE;
+
+                    generic::currentThreadWaitForDevice(dev);
+// #pragma omp taskwait
```
BenjaminW3: ?

thirdParty/catch2/include/catch2/catch.hpp
jkelling: These changes to catch2 are only required for some special cases:
* This one when compiling for offload to nvptx using clang.
* The other one with IMB's XL compiler.
Neither should be an issue in the CI. I will test this theory by pushing a version of this branch without these changes later.

These patches to not apply to the current catch2 master. I will have to open issues rather than a PR.

cmake/alpakaCommon.cmake
jkelling: I know that clang 11 claims to support OpenMP 5.0 (it is incomplete), I did not check where exactly clang 10 is at.

This PR is not explicitly using OpenMP 5.0, but it relies on the compiler being lenient with some of OpenMP < 5.0's restrictions, chiefly the `static constexpr` member thing: https://github.com/alpaka-group/alpaka/pull/1118 .

jkelling: The Math test uses `static constexpr` members, which is why it is disabled for GCC.

include/alpaka/dev/DevOmp5.hpp
jkelling: The OpenMP standard does not offer a way to query this. AFAIK, there is also no way to figure what type of device the target is, so we could check if we are targeting the CPU, then check the host memory, or if the we are targeting the GPU and then ask the CUDA/HIP runtime (which would be problematinc any because it would become a dependency). Thought of a workaround like this was the reason for the `\todo`. I think this can just not be supported. What is the alpaka policy to report, that this number is not available?

include/alpaka/idx/bt/IdxBtOmp5BuiltIn.hpp
```diff
@@ -0,0 +1,145 @@
...
+                    typename TWorkDiv>
+                static auto getIdx(
+                    idx::bt::IdxBtOmp5BuiltIn<dim::DimInt<1u>, TIdx> const & idx,
+                    TWorkDiv const &)
+                -> vec::Vec<dim::DimInt<1u>, TIdx>
+                {
+                    alpaka::ignore_unused(idx);
+                    return vec::Vec<dim::DimInt<1u>, TIdx>(static_cast<TIdx>(omp_get_thread_num()));
+                }
+            };
```
BenjaminW3: Could you add this 1D specialization to `alpaka::idx::bt::IdxBtOmp` and get rif of `IdxBtOmp5BuiltIn` completely?
This could also be a separate PR.

include/alpaka/mem/view/Traits.hpp
```diff
@@ -115,8 +115,9 @@ namespace alpaka
                         -> idx::Idx<TView>
                         {
                             return
-                                extent::getExtent<dim::Dim<TView>::value - 1u>(view)
-                                * sizeof(elem::Elem<TView>);
+                                static_cast<idx::Idx<TView>>(
+                                    static_cast<unsigned int>(extent::getExtent<static_cast<unsigned int>(dim::Dim<TView>::value) - 1u>(view))
```
BenjaminW3: What exactly made it necessary to cast the non-type-template parameters here?

include/alpaka/mem/view/ViewPlainPtr.hpp
```diff
@@ -349,6 +350,39 @@ namespace alpaka
...
+                    template<
+                        typename TElem,
+                        typename TExtent>
+                    static auto createStaticDevMemView(
+                        TElem * pMem,
+                        dev::DevOmp5 const & dev,
+                        TExtent const & extent)
+#ifdef BOOST_NO_CXX14_RETURN_TYPE_DEDUCTION
+                    -> alpaka::mem::view::ViewPlainPtr<dev::DevOmp5, TElem, alpaka::dim::Dim<TExtent>, alpaka::idx::Idx<TExtent>>
+#endif
```
BenjaminW3: This `#ifdef` can be removed now.

include/alpaka/pltf/PltfOmp5.hpp
```diff
@@ -0,0 +1,105 @@
...
+                ALPAKA_FN_HOST static auto getDevByIdx(
+                    std::size_t devIdx)
+                -> dev::DevOmp5
+                {
+                    ALPAKA_DEBUG_FULL_LOG_SCOPE;
+
+                    std::size_t const devCount(static_cast<std::size_t>(::omp_get_num_devices()));
+                    int devIdxOmp5 = static_cast<int>(devIdx);
+                    if(devIdx >= devCount)
+                    { // devIdx param must be unsigned, take take this case to use the initial device
```
BenjaminW3: "take take"?
The other platforms throw an exception in the case of an invalid `devIdx`. We should do this here as well.

include/alpaka/pltf/PltfOmp5.hpp
```diff
@@ -0,0 +1,105 @@
...
+                //-----------------------------------------------------------------------------
+                //! \param devIdx device id, less than GetDevCount, will be set to omp_get_initial_device() otherwise
+                ALPAKA_FN_HOST static auto getDevByIdx(
+                    std::size_t devIdx)
+                -> dev::DevOmp5
+                {
+                    ALPAKA_DEBUG_FULL_LOG_SCOPE;
+
+                    std::size_t const devCount(static_cast<std::size_t>(::omp_get_num_devices()));
+                    int devIdxOmp5 = static_cast<int>(devIdx);
```
BenjaminW3, line 92: What is the mapping between alpaka `devIdx` and OpenMP? Is `devIdx == 0` always the host and `devIdx > 0` the offloading devices?
Can we document this?

script/docker_ci.sh
```diff
@@ -57,9 +65,13 @@ if [ ! -z "${ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE+x}" ]
...
 fi
-if [ ! -z "${ALPAKA_ACC_CPU_BT_OMP4_ENABLE+x}" ]
+if [ ! -z "${ALPAKA_ACC_ANY_BT_OMP5_ENABLE+x}" ]
+then
+    ALPAKA_DOCKER_ENV_LIST+=("--env" "ALPAKA_ACC_ANY_BT_OMP5_ENABLE=${ALPAKA_ACC_ANY_BT_OMP5_ENABLE}")
+fi
+if [ ! -z "${ALPAKA_OFFLOAD_MAX_BLOCK_SIZE+x}" ]
 then
-    ALPAKA_DOCKER_ENV_LIST+=("--env" "ALPAKA_ACC_CPU_BT_OMP4_ENABLE=${ALPAKA_ACC_CPU_BT_OMP4_ENABLE}")
+    ALPAKA_DOCKER_ENV_LIST+=("--env" "ALPAKA_OFFLOAD_MAX_BLOCK_SIZE=${ALPAKA_OFFLOAD_MAX_BLOCK_SIZE}")
```
BenjaminW3, line 74: Is `ALPAKA_OFFLOAD_MAX_BLOCK_SIZE` limited to OpenMP for now?
Should we rename it to `ALPAKA_OMP_OFFLOAD_MAX_BLOCK_SIZE`?

test/common/include/alpaka/test/acc/TestAccs.hpp
BenjaminW3: Do we need this whole test disabling stuff in this PR at all? We are not enabling any gcc builds so we can simply wait for gcc-11 and only enable it there.

include/alpaka/dev/DevOmp5.hpp
BenjaminW3: Any non-zero value might be too much on a given hardware.
Only pessimistic answers can be valid answers. As we do not have any hint about the real sizes, zero seems to be the only safe return value.

include/alpaka/pltf/PltfOmp5.hpp
jkelling: OpenMP target has target devices with `0 <= devIdxOmp5 < devCount`. In addition, there is an "initial device", which has a `devIdxOmp5` which is implementation defined, all compilers I saw have `devIdxOmp5 < 0`. The trait's parameter is size_t, so we cannot map this directly, so I decided to use `devIdx > devCount`. Apparently, they want change this in OpenMP 5.1 to mandate `initialDevice == devCount`.

script/docker_ci.sh
jkelling: It is also used by OpenAcc.

include/alpaka/mem/view/Traits.hpp
jkelling: I got warnings there and it was either cast `1u` to `Idx` or cast the rest to `unsigned int`, I went with the latter. I think there are cases where `idx::Idx<TView>` is signed, maybe creating warnings because of the unsigned value returned by `sizeof`.

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -0,0 +1,365 @@
...
+                        // Register this queue on the device.
+                        m_queues.push_back(spQueue);
+                    }
+
+                    int iDevice() const {return m_iDevice;}
+
+                private:
+                    std::mutex mutable m_Mutex;
+                    std::vector<std::weak_ptr<queue::IGenericThreadsQueue<DevOmp5>>> mutable m_queues;
+                    int m_iDevice = 0;
```
BenjaminW3: Can we make this `const` and remove the `=0`? It is always initialized in the constructor.

include/alpaka/pltf/PltfOmp5.hpp
BenjaminW3: `devIdx` and `devCount` are in general totally unrelated to `devIdxOmp5`. We could have an arbitrary mapping between those.
I would imagine that we always artificially add one to the value returned by `omp_get_num_devices()` to account for the default-device when giving the user the number of available devices. `devIdx == omp_get_num_devices()` would request this default device. All other indices would directly map to the `devIdxOmp5` and `devIdx > omp_get_num_devices()` would result in the exception.
We could also change this mapping to make `devIdx == 0` the default-device and offset all others by one.

include/alpaka/acc/AccOmp5.hpp
jkelling: Yes, this is sensible.

include/alpaka/acc/AccOmp5.hpp
jkelling: This is basically a todo for a later OpenMP version: On CPU we would want `blockThreadCountMax * gridBlockCountMax` to equal the number of CPU threads available, so we would write:
```c++
auto const gridBlockCountMax(alpaka::core::clipCast<TIdx>(::omp_get_max_threads() / blockThreadCountMax));
```
However, If we did this, we would always run with one block (if `ALPAKA_OFFLOAD_MAX_BLOCK_SIZE` does not say otherwise) even on GPU, because `omp_get_max_threads()` appears to return the number of threads per block/team when the target is a GPU (at least XL does that, which is the only compiler with which I was actually able to run any of this on a GPU).

What we really want to do is `gridBlockCountMax =  omp_get_max_teams()`, but this function does not exist.

jkelling: The XL behavior may not be conforming to OpenMP 5.0, but I could not test with any other compiler, so any fix for this would at this point be theoretical. The way it is currently handled does not break too much. It can be fixed once it is clear how compilers handle this.

BenjaminW3: Could you add parts of your answer to an extended comment?

include/alpaka/core/Omp5.hpp
jkelling: No.

include/alpaka/mem/view/Traits.hpp
BenjaminW3: The `getExtent<static_cast<unsigned int>` seems to be valid.
The outer `static_cast<idx::Idx<TView>>` makes it explicit what the `return` will do anyways.
However, the `static_cast<unsigned int>(extent::getExtent<` is dangerous because we should not change / clip the values returned by `getExtent`.

include/alpaka/dev/DevOmp5.hpp
jkelling: Done.

include/alpaka/pltf/PltfOmp5.hpp
jkelling: Including the initial device in `devCount` would break performance of programs which map ranks to device in a wrap-around fashion or which try to use all devices reported by alpaka. The normal use case would be only use the actual target devices and not the host, i.e. `0 <= devIdx < devCount`. Only if there are not target devices available it makes sense include the host and thus have `devCount = 1` instead of `0`, merging OpenMP behavior with normal alpaka-backend behavior. There is no alpaka interface for using the initial device/host through the backend. Doing this, when there are actual target devices available, is purely an OpenMP target thing, thus only this case should deviate from normal alpaka behavior. If the user code is not aware which backend is being used, it should be able expect the same behavior from all backends, as long as it uses the backends properly through the interface.

We could limit this to using the initial device only if `devIdx == omp_get_num_devices()`, because this can be elegantly checked, and throw on other values.

This would of course still lead to the same unexpected behavior if user code were to iterate the devices like this:
```c++
try
{
 for(int a; ;++a)
 {
  getDevice(a);
  //...
 }
}
catch (out_of_range){}
```
but I think we do not need to cater to bad style.

BenjaminW3: > We could limit this to using the initial device only if devIdx == omp_get_num_devices(), because this can be elegantly checked, and throw on other values.

Sounds good

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -0,0 +1,365 @@
...
+        //! The Omp5 device handle.
+        class DevOmp5 :
+            public concepts::Implements<wait::ConceptCurrentThreadWaitFor, DevOmp5>,
+            public concepts::Implements<ConceptDev, DevOmp5>
+        {
+            friend struct pltf::traits::GetDevByIdx<pltf::PltfOmp5>;
+
+        protected:
+            //-----------------------------------------------------------------------------
+            DevOmp5(int iDevice) :
```
jkelling: Just noticed, that this ctor is protected, not private. I do not currently see a use case for deriving from this class and using this ctor in the derived class. So, I will remove the `protected:` from this PR.

include/alpaka/dev/DevOmp5.hpp
jkelling: Either this is left over from the old OpenMP4, or I was thinking about implementing an OpenMP tesking-based queue. Removed.

test/common/include/alpaka/test/acc/TestAccs.hpp
jkelling: The same stuff is in the OpenACC PR, where it is needed, because there is currently no other option than gcc10 in the CI for this one. So, it is mostly there for symmetry, although one could try to enable this with gcc10 in the CI, too. The reason I did not is because I can only compile but not run on my local machine. This is because it only compiles with `-foffload=disable` (i.e. no GPU) which will make it generate code for the host only but at runtime it sees a GPU and allocates the memory there, making the tests segfault. In the CI there is no GPU, so it will either work, because it allocates on the host or also fail, because it fails to allocate anything on target. I did not test this, but wanted to keep the option open.

On the other hand: could we make a compile-only Ci job?

BenjaminW3: Yes, currently we have the variable `ALPAKA_CI_ANALYSIS` which selects whether the alpaka tests should be run or if the self-contained-header build (and some other small things) should be done. This was necessary in Travis CI because both things combined took too long. This may not be necessary now with github actions but this is still available.

BenjaminW3: I just took a look at the gcc release dates and it lloks like gcc-11 will not be released before May 2021. So waiting for it does not seem to be an option.

include/alpaka/mem/view/Traits.hpp
jkelling: Maybe test causing this issue were changed in the meantime, if I revert this hunk, I can now build without warnings.

Should I remove this change or just reduce it to the innermost cast to `unsigned int`?

BenjaminW3: I would completely remove it

include/alpaka/idx/bt/IdxBtOmp5BuiltIn.hpp
jkelling: https://github.com/alpaka-group/alpaka/pull/1140

include/alpaka/acc/AccOmp5.hpp
jkelling: https://github.com/alpaka-group/alpaka/pull/1126/commits/fbf0df054a673698127df84c73342514bbcb6290

include/alpaka/pltf/PltfOmp5.hpp
jkelling: https://github.com/alpaka-group/alpaka/pull/1126/commits/2db34b82d4586a5ea853005aba532e9b5ddfccf2

In addition to the above condition, it also uses the initial device if `devCount == 0 && devIdx == 1`, that is when `devIdx` is equal to the return value of `getDevCount()`, which is this case differs from `omp_get_num_devices()`.

include/alpaka/pltf/PltfOmp5.hpp
jkelling: https://github.com/alpaka-group/alpaka/commit/2db34b82d4586a5ea853005aba532e9b5ddfccf2#diff-0af16922c3afe4ac9df8b1a1c77c3229R84

Changed the docstring a bit.

test/common/include/alpaka/test/acc/TestAccs.hpp
jkelling: I added an `ALPAKA_CI_ANALYSIS` job with gcc 10 which includes OMP5.

.github/workflows/ci.yml
```diff
@@ -81,85 +81,88 @@ jobs:
       matrix:
         include:
         ### Analysis builds
+        - name: linux_gcc-10_debug_analysis-omp5
```
BenjaminW3: Could you please use an underscore here and in the renamed clang job?
```suggestion
        - name: linux_gcc-10_debug_analysis_omp5
```

jkelling: Done,.

include/alpaka/mem/buf/omp5/Set.hpp
```diff
@@ -0,0 +1,116 @@
...
+                        std::uint8_t const & byte,
+                        TExtent const & extent)
+                    {
+                        using Idx = typename idx::traits::IdxType<TExtent>::type;
+                        auto pitch = view::getPitchBytesVec(view);
+                        auto byteExtent = extent::getExtentVec(extent);
+                        byteExtent[TDim::value-1] *= static_cast<Idx>(sizeof(elem::Elem<TView>));
+                        constexpr auto lastDim = TDim::value - 1;
+
+                        if(pitch[0] <= 0)
```
BenjaminW3: Could you please add braces around the if block?

include/alpaka/mem/buf/omp5/Set.hpp
```diff
@@ -0,0 +1,116 @@
...
+                                        vec::Vec<TDim, Idx>::ones(),
+                                        vec::Vec<TDim, Idx>::ones()),
+                                    view::MemSetKernel(),
+                                    byte,
+                                    reinterpret_cast<std::uint8_t*>(alpaka::mem::view::getPtrNative(view)),
+                                    byteExtent,
+                                    pitch
+                                ); // NOP if size is zero
+
+                        std::cout << "Set TDim=" << TDim::value << " pitch=" << pitch << " byteExtent=" << byteExtent << std::endl;
```
BenjaminW3, line 92: Could you please guard this log line with `#if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL`?

include/alpaka/mem/buf/omp5/Set.hpp
BenjaminW3: What exactly is this pitch check doing here?
Shouldn't we be checking the `byteExtent.prod()`?
Furthermore, negative pitches are a valid thing. Some image formats describe their buffers with a pointer to the back and pitch being negative.

include/alpaka/mem/buf/omp5/Set.hpp
```diff
@@ -0,0 +1,116 @@
...
+                    ALPAKA_FN_HOST static auto createTaskSet(
+                        TView & view,
+                        std::uint8_t const & byte,
+                        TExtent const & extent)
+                    {
+                        using Idx = typename idx::traits::IdxType<TExtent>::type;
+                        auto pitch = view::getPitchBytesVec(view);
+                        auto byteExtent = extent::getExtentVec(extent);
+                        byteExtent[TDim::value-1] *= static_cast<Idx>(sizeof(elem::Elem<TView>));
+                        constexpr auto lastDim = TDim::value - 1;
```
BenjaminW3, line 73: Could you switch those two lines and reuse `lastDim` in the previous line?

include/alpaka/mem/buf/omp5/Set.hpp
```diff
@@ -0,0 +1,116 @@
...
+
+                        if(pitch[0] <= 0)
+                            return kernel::createTaskKernel<acc::AccOmp5<TDim,Idx>>(
+                                    workdiv::WorkDivMembers<TDim, Idx>(
+                                        vec::Vec<TDim, Idx>::ones(),
+                                        vec::Vec<TDim, Idx>::ones(),
+                                        vec::Vec<TDim, Idx>::ones()),
```
BenjaminW3: Isnt't this the "nothing to do"-case? Shouldn't this be zeros?

include/alpaka/mem/buf/omp5/Copy.hpp
```diff
@@ -0,0 +1,545 @@
...
+                                for(int i = lastDim - 1; i >= 0; --i)
+                                {
+                                    dstExtentFull[i] = m_dstPitchBytes[i]/m_dstPitchBytes[i+1];
+                                    srcExtentFull[i] = m_srcPitchBytes[i]/m_srcPitchBytes[i+1];
+                                }
+
+                                // std::cout << "copy " << TDim::value << "d\textent=" << m_extent
+                                //     << "\tdstExtentFull=" << dstExtentFull << " (p " << m_dstPitchBytes
+                                //     << " )\tsrcExtentFull=" << srcExtentFull << " (p " << m_srcPitchBytes
+                                //     << " )\telementSize=" << elementSize << std::endl;
```
BenjaminW3: remove those comments?

include/alpaka/mem/buf/omp5/Copy.hpp
```diff
@@ -0,0 +1,545 @@
...
+                                m_iSrcDevice(iSrcDevice),
+                                m_extent(VecFromDimTrait<
+                                        TDim, size_t, TExtent,
+                                        MyGetExtent>::vecFromDimTrait(extent)),
+                                m_dstPitchBytes(VecFromDimTrait<
+                                        TDim, size_t, TViewDst,
+                                        MyGetPitch>::vecFromDimTrait(viewDst)),
+                                m_srcPitchBytes(VecFromDimTrait<
+                                        TDim, size_t, TViewSrc,
+                                        MyGetPitch>::vecFromDimTrait(viewSrc)),
```
BenjaminW3: What is the difference to the `mem::view::getPitchBytesVec(viewSrc)`, that is used in ` TaskCopyCpuBase` (and `getExtentVec` for above)?

include/alpaka/mem/buf/BufOmp5.hpp
```diff
@@ -0,0 +1,713 @@
...
+                    typename TExtent>
+                ALPAKA_FN_HOST BufOmp5(
+                    dev::DevOmp5 const & dev,
+                    TElem * const pMem,
+                    TExtent const & extent) :
+                        m_spBufImpl(std::make_shared<omp5::detail::BufOmp5Impl<TElem, TDim, TIdx>>(dev, pMem, extent))
+                {}
+
+                    BufOmp5(const BufOmp5&) = default;
+                    BufOmp5(BufOmp5 &&) = default;
```
BenjaminW3, line 150: indentation seems to be wrong

include/alpaka/mem/buf/BufOmp5.hpp
```diff
@@ -0,0 +1,713 @@
...
+                        dev::DevOmp5 m_dev;
+                        vec::Vec<TDim, TIdx> m_extentElements;
+                        vec::Vec<TDim, TIdx> m_pitchBytes;
+                        TElem* m_pMem;
+
+                            BufOmp5Impl(const BufOmp5Impl&) = delete;
+                            BufOmp5Impl(BufOmp5Impl &&) = default;
+                        BufOmp5Impl& operator=(const BufOmp5Impl&) = delete;
+                        BufOmp5Impl& operator=(BufOmp5Impl&&) = default;
+                            ~BufOmp5Impl()
```
BenjaminW3: indentation wrong here and above as well

include/alpaka/kernel/TaskKernelOmp5.hpp
```diff
@@ -0,0 +1,386 @@
...
+#endif
+                // We have to make sure, that the OpenMP runtime keeps enough threads for executing a block in parallel.
+                TIdx const maxOmpThreadCount(static_cast<TIdx>(::omp_get_max_threads()));
+                // The number of blocks in the grid.
+                TIdx const gridBlockCount(gridBlockExtent.prod());
+                // The number of threads in a block.
+                TIdx const blockThreadCount(blockThreadExtent.prod());
+
+#if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
+                if(maxOmpThreadCount < blockThreadExtent.prod())
```
BenjaminW3, line 142: braces here please as well.

test/unit/atomic/CMakeLists.txt
```diff
@@ -20,5 +20,6 @@ target_link_libraries(
     PRIVATE common)
 
 set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER "test/unit")
+target_compile_definitions(${_TARGET_NAME} PRIVATE "-DTEST_UNIT_ATOMIC")
```
BenjaminW3, line 23: Should we do this confitionally only when `ALPAKA_ACC_ANY_BT_OMP5_ENABLE` is set? This would also automatically document why this is necessary.

test/integ/separableCompilation/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 #
 
-if((NOT ALPAKA_ACC_GPU_CUDA_ENABLE AND NOT ALPAKA_ACC_GPU_HIP_ENABLE) OR
+if((NOT ALPAKA_ACC_GPU_CUDA_ENABLE AND NOT ALPAKA_ACC_GPU_HIP_ENABLE AND NOT ALPAKA_ACC_ANY_BT_OMP5_ENABLE ) OR
```
BenjaminW3, line 11: separable compilation is not supported? Is this also the known linker error that will be fixed with gcc-11?

test/unit/kernel/src/KernelStdFunction.cpp
```diff
@@ -7,6 +7,11 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+#define TEST_UNIT_KERNEL_KERNEL_STD_FUNCTION
+// clang thinks a macro is unused when only used as the second operand of an &&
+// where the first operand evaluates to false, so we use it here:
+TEST_UNIT_KERNEL_KERNEL_STD_FUNCTION
```
BenjaminW3, line 13: Why is this clang workaround not necessary for the other tests?
Should we remove one `KERNEL` from the name?

jkelling: 1. Maybe because there the macros are defined via `-D` instead of in the code and are thus maybe not triggering warnings.
2. This one only disables a part of what would be called `TEST_UNIT_KERNEL`, namely the `KERNEL_STD_FUNCTION` part. This is also why this is defined in the source file and not via `target_compile_definitions`.

test/integ/separableCompilation/CMakeLists.txt
jkelling: The issue here is that functions which are not defined inline have to be enclosed in
```c++
#pragma omp declare target
...
#pragma omp end declare target
```
so target code is generated and it can be called  in target regions. This cannot be worked into the `ALPAKA_FN_HOST_DEVICE` macro. Note, even if there would only be one pragma in front required, like in OpenACC, we could still not map this to the macro, because this would break template declarations:
```c++
template<class T>
ALPAKA_FN_DEVICE void f(int);
```
would expand to
```c++
template<class T>
_Pragma("acc routine seq") void f(int);
```
but you cannot place the pragma between the template and the function part.

test/unit/atomic/CMakeLists.txt
jkelling: Doing this would require this line to change again in https://github.com/alpaka-group/alpaka/pull/1127 , because they need to be excluded there.

I would rather add a line like this to all test, so each test can be identified by the code in accelerator selection code [test/common/include/alpaka/test/acc/TestAccs.hpp](https://github.com/alpaka-group/alpaka/pull/1126/files#diff-c93c0e3f72fa0a6945dfff2600feafd4). Other new backends, like [SYCL](https://github.com/alpaka-group/alpaka/pull/789) may also want to exclude some test at some stages of development. This could be a saparate PR, then.

include/alpaka/mem/buf/omp5/Set.hpp
jkelling: I noticed, that the existing pitch calculations always lead to `pitch[0]` containing `byteExtent.prod()`, so we would not need to calculate it here.

I was not aware of the possibility of negative pitches, could we check `pitch[0] != 0` instead?

BenjaminW3: yes, should be the correct thing to do.

include/alpaka/mem/buf/BufOmp5.hpp
jkelling: Habit (indenting ctors and dtors), sorry. Fixed.

include/alpaka/mem/buf/omp5/Copy.hpp
jkelling: This version is casting to a `Vec<size_t>`. The class stores the extents with this type because they are directly used as arguments to the omp runtime functions, which expect arrays of `size_t`. An alternative to this may be to add a cast operator to `Vec<..>`.

BenjaminW3: You can find it here: https://github.com/alpaka-group/alpaka/blob/d1f7047ae2116cc6a83674ce352adabbb263ad04/test/unit/vec/src/VecTest.cpp#L91-L113

jkelling: Thanks; I put it in.

jkelling:
Somehow adding that one CI job "linux_gcc-10_debug_analysis_omp5" made the number of check go up from 101 to 107. Did this create some other jobs implicitly due to the "matrix" mode?

include/alpaka/mem/buf/omp5/Set.hpp
jkelling: A kernel with a zero workdiv currently leads to a divide-by-zero exception. I did not take zeros to be a valid workdiv.

As it is now, the kernel will run, see that the extent is zero and terminate. With zeros, the `TaskKernelOmp5::operator()` needs to check this and return without executing.

jkelling: I pushed this change: https://github.com/alpaka-group/alpaka/pull/1126/commits/5626aae8e287102b91a52daef51dbe86212eecb0

jkelling: Amended: https://github.com/alpaka-group/alpaka/pull/1126/commits/50666f6941193635a1d1f4f226de3ec5261b7655
(Make comment doc string)

BenjaminW3:
Did you rebase which might have added some new builds to your branch?

jkelling:
> Did you rebase which might have added some new builds to your branch?

A diff to develop shows many changes, but mostly removing `ALPAKA_ACC_CPU_BT_OMP4_ENABLE=OFF`. "linux_gcc-10_debug_analysis_omp5" is the only one added.

jkelling:
... or were jobs recently added to develop? I did rebase onto the current state of the develop branch.

BenjaminW3:
I think some Windows+CUDA 11 builds have been added

include/alpaka/mem/buf/omp5/Set.hpp
jkelling: @psychocoderHPC , @BenjaminW3 : Do you agree with this behavior of `TaskKernelOmp5::operator()`, i.e. return without doing anything and not raising an error if the grid is empty?

include/alpaka/dev/DevOmp5.hpp
jkelling: @psychocoderHPC : Reporting zero here breaks PIConGPU via pmacc, which apparently checks this and then fails to allocate memory because there are only 0B left. How do you suggest to resolve this? Would you change this to be more optimistic (basically `INT64_MAX`) or should this be handled in pmacc?

jkelling: @BenjaminW3 Maybe there should be a trait added to check if the value returned by this one can be trusted on the given backend or should be ignored.

include/alpaka/mem/buf/omp5/Set.hpp
BenjaminW3: I am OK with this

include/alpaka/dev/DevOmp5.hpp
BenjaminW3: Or return an optional

jkelling: Using optional would be a breaking change. With this:
```c++
template<class TDev>
ALPAKA_FN_HOST static auto getMemBytes(TDev const & dev)
 -> boost::optional<std::size_t>
{...}
```
All uses would need to change to:
```c++
auto memBytes = getMemBytes(dev);
if(memBytes)
{
 if(*memBytes < howMuchWeNeed)
// ...
}
else
{
// hope there is enough
}
```

BenjaminW3: But better explicitly break the interface to make users aware than hiding the still breaking change by reporting UINT_MAX or similar unknown values.
Lets wait for an answer from @psychocoderHPC. I would still prefer returning zero.

jkelling: I am with you on the zero vs. UINT_MAX issue.

Adding another trait would however be less breaking than returning an optional: Using optional would require all code using any backend to change. Adding another trait, would allow legacy code work unchanged with all backends other than this one. Only code that wants to support this new backend would have to add a check with the new trait ( `deviceKnowsAvailableMem` or something ), which with time would become all code, but it could happen more gradually.

psychocoderHPC: Is it possible to start a kernel to figure out which type of device is used? Than we could use the return value of this kernel to select they way how to calculate the available memory.

jkelling: I do not see any standard way to get the type of device being used. We could add a cmake option to select the target which could then define a macro and set the appropriate compiler flags. Something like this should become part of the OpenMP cmake module at some point. This would, however, not work when the stand-alone headers are used.

Even if we could determine what type of device is being used, there would still be no guarantee, that we can determine a value, because we need code for each device type, so we would have nothing in a new environment.

We need a way to tell the program, that we cannot get this information as a fallback.

jkelling: Also: If this would be a runtime decision, all strategies for possible targets need to be compiled. While it is possible to leave out runtimes which are not available at compile time, this would make CUDA a dependency if you want to compile for OpenMP 5 targeting nvptx,for example. Usually this dependency would not be explicit and it might sometimes be hidden away in the compiler.

test/integ/separableCompilation/CMakeLists.txt
BenjaminW3: Is this missing support for separable compilation documented somewhere?

jkelling: No, I only disabled the test. Which would be the place to document this?

include/alpaka/dev/DevOmp5.hpp
jkelling: I had a discussion with @psychocoderHPC 
* The PIConGPU/pmacc issue is not central and can be worked around very easily
* He brought up the option of raising an exception in the specialization, which would only break application if they use this backend and the root cause would be easy to track down. Also these functions are not pefrormance critical.
* He would be OK with just returning zero. I will add a commit to this PR documenting the possibility of this trait returning 0, which then means, that the real value is unknown. I check this issue for the OpenACC backend: There it is possible to get these values (works with NHVPC 20.7+GPU). However GCC just returns 0, which maybe because of a lack of support or disabled offloading, and a lack of support in the runtime for the host. This means, 0 return values can also occur for this trait when support is technically there.

jkelling: Added https://github.com/alpaka-group/alpaka/pull/1126/commits/ab9b8dbb094a67703bc627076f1652a6da12681e

I suggest to move forward with this state and optionally create an issue in case we want to consider more invasive handling (optional, exception, ...).

BenjaminW3:
I do not think that more than 100 commits bring any value. Could you please reduce the number of commits (maybe to one)?

jkelling:
I can squash after all issues are resolved.

test/integ/separableCompilation/CMakeLists.txt
BenjaminW3: It should probably be part of `README_OMP5.md`

jkelling: Added: https://github.com/alpaka-group/alpaka/pull/1126/commits/9e78aa35b9138cc5a096d5e597833554c9662623
(also some updates on compiler status)

BenjaminW3 (approved):

BenjaminW3:
@psychocoderHPC Could you please also review this PR?

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -58,36 +55,36 @@ namespace alpaka
...
             typename TIdx>
-        class AccCpuOmp4 final :
+        class AccOmp5 final :
             public workdiv::WorkDivMembers<TDim, TIdx>,
-            public idx::gb::IdxGbRef<TDim, TIdx>,
+            public idx::gb::IdxGbLinear<TDim, TIdx>,
             public idx::bt::IdxBtOmp<TDim, TIdx>,
             public atomic::AtomicHierarchy<
-                atomic::AtomicStdLibLock<16>,   // grid atomics
+                atomic::AtomicOmpBuiltIn,   // grid atomics
```
psychocoderHPC, line 74: Is OpenMP guarantee that atomic is atomic between grids?
If not than it would not be allowed to use atomics in two parallel running kernel where the same memory is accessed. We need to at at least a warning about unexpected behaviour into the documentation.

psychocoderHPC (changes requested):
There is only one open question about atomic functions between grids.

include/alpaka/acc/AccOmp5.hpp
jkelling: It depends:
OpenMP 5 2.17.7 . Binding
```
If the size of x is 8, 16, 32, or 64 bits and x is aligned to a multiple of its size, the binding thread set
for the atomic region is all threads on the device. Otherwise, the binding thread set for the
atomic region is all threads in the contention group. atomic regions enforce exclusive access
with respect to other atomic regions that access the same storage location x among all threads in
the binding thread set without regard to the teams to which the threads belong.
```

sbastrakov: Did I understand the quoted fragment right that in case the size and alignment conditions are met, atomicity is guaranteed for all threads on a device (so between grids in alpaka speech), while if they are not met, only between threads of an openmp team (not sure what it corresponds to in alpaka for this backend)?

jkelling: That is how I understood it.

psychocoderHPC (approved):

psychocoderHPC:
Welcome in the jungle of alpaka backends

PR #1128, opened 2020-08-19
Fix and improve block shared mem st member sanity checks

sbastrakov:
First, the check was inaccurate: it is fine to allocate exactly the whole size, but the check did not allow it.

Then, moved the check to be before the unsafe part, instead of after. So that now a wrong usage triggers the check rather than crashes the program before it. Note that the case of `m_capacity = 0` and `m_mem = nullptr` is automatically covered now.

Edit: after discussion below, also removed placement `new` and added additional debug asserts.

### DIFF
--- a/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
+++ b/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
@@ -58,7 +58,7 @@ namespace alpaka
                             new (buf) T();
                             m_allocdBytes += sizeof(T);
 #if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
-                            ALPAKA_ASSERT(m_allocdBytes < m_capacity);
+                            ALPAKA_ASSERT(m_allocdBytes <= m_capacity);
 #endif
                         }
 
--- a/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
+++ b/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
@@ -59,8 +59,6 @@ namespace alpaka
                         void alloc() const
                         {
                             m_allocdBytes = allocPitch<T>();
-                            uint8_t* buf = &m_mem[m_allocdBytes];
-                            new (buf) T();
                             m_allocdBytes += sizeof(T);
 #if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
                             ALPAKA_ASSERT(m_allocdBytes <= m_capacity);
--- a/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
+++ b/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
@@ -35,7 +35,12 @@ namespace alpaka
                     public:
                         //-----------------------------------------------------------------------------
 #ifndef NDEBUG
-                        BlockSharedMemStMemberImpl(uint8_t* mem, unsigned int capacity) : m_mem(mem), m_capacity(capacity) {}
+                        BlockSharedMemStMemberImpl(uint8_t* mem, unsigned int capacity) : m_mem(mem), m_capacity(capacity)
+                        {
+#ifdef ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST
+                            ALPAKA_ASSERT( ( m_mem == nullptr ) == ( m_capacity == 0u ) );
+#endif
+                        }
 #else
                         BlockSharedMemStMemberImpl(uint8_t* mem, unsigned int) : m_mem(mem) {}
 #endif


include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -54,12 +54,12 @@ namespace alpaka
                         void alloc() const
                         {
                             m_allocdBytes = allocPitch<T>();
+#if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
+                            ALPAKA_ASSERT(m_allocdBytes + sizeof(T) <= m_capacity);
```
sbastrakov: Here need to add `sizeof(T)`, as it was correctly done in the original version.

sbastrakov:
Note: the failing CI configuration is due to fetching, unrelated to these changes.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -54,12 +54,12 @@ namespace alpaka
...
+#endif
                             uint8_t* buf = &m_mem[m_allocdBytes];
                             new (buf) T();
```
jkelling: I was actually thinking about removing this placement new, because `alloc()` did not initialize the memory in any backend. While type allocated here should be default constructible, in some exotic cases default constructing causes problems because of compiler bugs (e.g. generated default ctor calls memset()).

jkelling: Maybe we could just remove the `new` and not move the assert.

sbastrakov: Maybe, I do not have a particular opinion regarding `new`.
But I think my change has to be applied anyways, as already `m_mem[m_allocdBytes]` is potentially dangerous and should happen after the debug check.

jkelling: Ahh, you are referring to the case `m_mem == nullptr`. Maybe this should be checked in the `BlockSharedMemStMember` ctor explicitly. The assert here relies on `m_capacity` telling the truth, which we should no rely on if we want to catch bugs like this.

jkelling: Regarding `new`: it is pointless here, because it is not required, thus applications cannot rely on it, and might cause issues which can be avoided by just not having it. I wanted to start a discussion later about maybe introducing an additional interface that does alloc and has a master thread call `new`, maybe `allocEmplace(Args... && args)`.

(Btw. I cannot edit here :( )

jkelling: Or maybe we do not want to check in the ctor, if the assume passing `nullptr` to be valid usage, where you are then just not allowed to call malloc. In this case your check should be enough.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -54,12 +54,12 @@ namespace alpaka
                         void alloc() const
                         {
                             m_allocdBytes = allocPitch<T>();
+#if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
+                            ALPAKA_ASSERT(m_allocdBytes + sizeof(T) <= m_capacity);
+#endif
                             uint8_t* buf = &m_mem[m_allocdBytes];
```
jkelling: While we are discussing this code: Which version would you consider is more readable, the current
```c++
&m_mem[m_allocdBytes]
```
or
```c++
m_mem + m_allocdBytes
```
?

sbastrakov: I find both understandable with a slight favor towards +.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
sbastrakov: Yes, `nullptr` and no allocations should be a valid usage. And the check can help (in debug mode) with the case of nullptr and attempt to allocate. This is what actually happened with me debugging recently.

sbastrakov: With the `new` being not required: I see your point. Do not really know, and it's unrelated to this PR.

@psychocoderHPC @BenjaminW3 what are your thoughts?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: I do also lean towards `+`, although it look a bit C-ish, but it is more explicit about what we are doing here. I think `+` would also be save with `m_mem = nullptr`.

sbastrakov: Ah, you are right. I will change then

jkelling:
Would you like to also add the following check to the debug ctor:
```c++
                            ALPAKA_ASSERT( ( m_mem == nullptr ) == ( m_capacity == 0 ) );
#endif
```
? Just to make sure the capacity is zero when a `nullptr` was passed and to document, that `nullptr` is valid under this condition.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -54,12 +59,12 @@ namespace alpaka
...
                             m_allocdBytes = allocPitch<T>();
-                            uint8_t* buf = &m_mem[m_allocdBytes];
-                            new (buf) T();
-                            m_allocdBytes += sizeof(T);
 #if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
-                            ALPAKA_ASSERT(m_allocdBytes < m_capacity);
+                            ALPAKA_ASSERT(m_allocdBytes + sizeof(T) <= m_capacity);
 #endif
+                            uint8_t* buf = m_mem + m_allocdBytes;
+                            new (buf) T();
```
jkelling: I am encourage you to remove the placement new, because no other backend's shared memory has this feature, it was a bad idea. It is related ti this PR, because removing it makes it possible to leave the ASSERT in the back and not have to add `sizeof(T)` twice.

sbastrakov: Done.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
sbastrakov: Okay, I implemented suggestion of @jkelling and removed the placement `new`.

BenjaminW3 (approved):

sbastrakov:
@BenjaminW3 could you merge please?

PR #1132, opened 2020-08-24
Add set kernel

jkelling:
This PR adds a kernel which implement device memset in a generic way for use by OpenMP 5.0 and OpenAcc backends.

Also introduces ~~`MapIdxPitch`~~`MapIdxPitchBytes` which is similar to `MapIdxExtend` but works based on pitches to support views.

### DIFF
--- a/include/alpaka/idx/MapIdx.hpp
+++ b/include/alpaka/idx/MapIdx.hpp
@@ -157,5 +157,138 @@ namespace alpaka
                     idx,
                     extent);
         }
+
+        namespace detail
+        {
+            //#############################################################################
+            //! Maps a linear index to a N dimensional index.
+            template<
+                std::size_t TidxDimOut,
+                std::size_t TidxDimIn,
+                typename TSfinae = void>
+            struct MapIdxPitch;
+            //#############################################################################
+            //! Maps a N dimensional index to the same N dimensional index.
+            template<
+                std::size_t TidxDim>
+            struct MapIdxPitch<
+                TidxDim,
+                TidxDim>
+            {
+                //-----------------------------------------------------------------------------
+                // \tparam TElem Type of the index values.
+                // \param idx Idx to be mapped.
+                // \param pitch Spatial pitch (in elems) to map the index to
+                // \return A N dimensional vector.
+                ALPAKA_NO_HOST_ACC_WARNING
+                template<
+                    typename TElem>
+                ALPAKA_FN_HOST_ACC static auto mapIdxPitch(
+                    vec::Vec<dim::DimInt<TidxDim>, TElem> const & idx,
+                    vec::Vec<dim::DimInt<TidxDim>, TElem> const & pitch)
+                -> vec::Vec<dim::DimInt<TidxDim>, TElem>
+                {
+                    alpaka::ignore_unused(pitch);
+
+                    return idx;
+                }
+            };
+            //#############################################################################
+            //! Maps a 1 dimensional index to a N dimensional index.
+            template<
+                std::size_t TidxDimOut>
+            struct MapIdxPitch<
+                TidxDimOut,
+                1u,
+                typename std::enable_if<TidxDimOut != 1u>::type>
+            {
+                //-----------------------------------------------------------------------------
+                // \tparam TElem Type of the index values.
+                // \param idx Idx to be mapped.
+                // \param pitch Spatial pitch (in elems) to map the index to
+                // \return A N dimensional vector.
+                ALPAKA_NO_HOST_ACC_WARNING
+                template<
+                    typename TElem>
+                ALPAKA_FN_HOST_ACC static auto mapIdxPitch(
+                    vec::Vec<dim::DimInt<1u>, TElem> const & idx,
+                    vec::Vec<dim::DimInt<TidxDimOut>, TElem> const & pitch)
+                -> vec::Vec<dim::DimInt<TidxDimOut>, TElem>
+                {
+                    auto idxNd(vec::Vec<dim::DimInt<TidxDimOut>, TElem>::all(0u));
+
+                    constexpr std::size_t lastIdx(TidxDimOut - 1u);
+
+                    TElem tmp = idx[0u];
+                    for(std::size_t d(0u); d < lastIdx; ++d)
+                    {
+                        idxNd[d] = static_cast<TElem>(tmp / pitch[d+1]);
+                        tmp %= pitch[d+1];
+                    }
+                    idxNd[lastIdx] = tmp;
+
+                    return idxNd;
+                }
+            };
+            //#############################################################################
+            //! Maps a N dimensional index to a 1 dimensional index.
+            template<
+                std::size_t TidxDimIn>
+            struct MapIdxPitch<
+                1u,
+                TidxDimIn,
+                typename std::enable_if<TidxDimIn != 1u>::type>
+            {
+                //-----------------------------------------------------------------------------
+                // \tparam TElem Type of the index values.
+                // \param idx Idx to be mapped.
+                // \param pitch Spatial pitch (in elems) to map the index to
+                // \return A 1 dimensional vector.
+                ALPAKA_NO_HOST_ACC_WARNING
+                template<
+                    typename TElem>
+                ALPAKA_FN_HOST_ACC static auto mapIdxPitch(
+                    vec::Vec<dim::DimInt<TidxDimIn>, TElem> const & idx,
+                    vec::Vec<dim::DimInt<TidxDimIn>, TElem> const & pitch)
+                -> vec::Vec<dim::DimInt<1u>, TElem>
+                {
+                    constexpr auto lastDim = TidxDimIn - 1;
+                    TElem idx1d(idx[lastDim]);
+                    for(std::size_t d(0u); d < lastDim; ++d)
+                    {
+                        idx1d = static_cast<TElem>(idx1d + pitch[d+1] * idx[d]);
+                    }
+                    return {idx1d};
+                }
+            };
+        }
+
+        //#############################################################################
+        //! Maps a N dimensional index to a N dimensional position.
+        //!
+        //! \tparam TidxDimOut Dimension of the index vector to map to.
+        //! \tparam TidxDimIn Dimension of the index vector to map from.
+        //! \tparam TElem Type of the elements of the index vector to map from.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<
+            std::size_t TidxDimOut,
+            std::size_t TidxDimIn,
+            typename TElem>
+        ALPAKA_FN_HOST_ACC auto mapIdxPitch(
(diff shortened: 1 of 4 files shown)


include/alpaka/idx/MapIdx.hpp
```diff
@@ -157,5 +157,138 @@ namespace alpaka
...
+            static_assert(TidxDimIn > 0u, "The dimension of the input vector has to be greater than zero!");
+
+            return
+                detail::MapIdxPitch<
+                    TidxDimOut,
+                    TidxDimIn>
+                ::mapIdxPitch(
+                    idx,
+                    pitch);
+        }
```
BenjaminW3, line 293: Could you please add a unit test for this functionality?
The test for the existing `mapIdx` is not very good but we should try to improve.

BenjaminW3:
From the pure naming, I do not get what the pitch version does differently.
Could you give me some examples for easier understanding? Those could also be used as unit tests.

include/alpaka/idx/MapIdx.hpp
jkelling: Had to wrap my head around this myself first, wrote this a while ago. I pushed a test case.

What `SetKernel` uses it for is the following: You have grid `threadIdx` and want to map this to a linear index relative to the base pointer of an `alpaka::mem::view::ViewSubView` which has non-zero offset. The `Nd` to `1d` version gives this index, observing, that you skip chunks of the addresspace due to the offset without having to do check extent and offset separately in each dimension. The `1d` to `Nd` version does the opposite for completeness sake but it not used anywhere, I think.

test/unit/idx/src/MapIdxPitch.cpp
```diff
@@ -0,0 +1,56 @@
...
+    auto const devAcc = alpaka::pltf::getDevByIdx<Acc>(0u);
+    alpaka::mem::view::ViewPlainPtr<Dev, Elem, Dim, Idx> parentView( nullptr, devAcc, extentNd );
+
+    auto const offset(Vec::all(4u));
+    auto const extent(Vec::all(4u));
+    auto const idxNd(Vec::all(2u));
+    alpaka::mem::view::ViewSubView<Dev, Elem, Dim, Idx> view( parentView, extent, offset );
+    auto pitch = alpaka::mem::view::getPitchBytesVec(view);
+    for(Idx a = 0; a < Dim::value; ++a)
+        pitch[a] /= sizeof(Elem);
```
BenjaminW3: I am not sure if we can make the assumption that the pitch is cleanly dividable by the element size.
The `Elem` type is something that is only known on the alpaka side. The allocation size is defined by `extent*sizeof(Elem)`. CUDA then may introduce some padding to make the pitch better match the hardware.
If we have a custom struct that has a size of 3 bytes and allocate a 2-dimenstional buffer of size 2x2, then CUDA may give us a buffer with pitch 8 instead of 6. However, 8 is not cleanly dividable by 3.

include/alpaka/mem/buf/SetKernel.hpp
```diff
@@ -0,0 +1,80 @@
...
+                    constexpr auto lastDim = dim::Dim<TAcc>::value - 1;
+                    const auto lastIdx = idx +
+                        std::min(threadElemExtent[lastDim], static_cast<Idx>(extent[lastDim]-idxThreadFirstElem[lastDim]));
+
+                    if ([&idxThreadFirstElem, &extent](){
+                            for(auto i = 0u; i < dim::Dim<TAcc>::value; ++i)
+                                if(idxThreadFirstElem[i] >= extent[i])
+                                    return false;
+                            return true;
+                        }())
```
BenjaminW3: May we be able to express this via an element-wise vector comparison and a logical fold-expression?
`if((idxThreadFirstElem < extent).foldrAll(std::logical_and<bool>()))`?

include/alpaka/mem/buf/SetKernel.hpp
```diff
@@ -0,0 +1,80 @@
...
+                    {
+                        // assuming elements = {1,1,...,1,n}
```
BenjaminW3: I do not get this comment

include/alpaka/mem/buf/SetKernel.hpp
```diff
@@ -0,0 +1,80 @@
...
+                        for(; idx<lastIdx; ++idx)
+                        {
+                            dst[idx] = val;
```
BenjaminW3: Given my example where the element size is not a multiple of the pitch, it is invalid to index into a multidimensional buffer outside of the first row.
We can not use `dst` from the outside here directly. We have to calculate the base pointer for each thread manually.

BenjaminW3: You can find the corresponding code here: https://github.com/alpaka-group/alpaka/blob/develop/test/common/include/alpaka/test/mem/view/Iterator.hpp#L141-L153
`m_pitchBytes` -> `pitch`
`currentIdxDimx` -> `idxThreadFirstElem`

BenjaminW3:
Thank you for adding the test. This made it much easier to understand what those methods are going to do ;-)

test/unit/idx/src/MapIdxPitch.cpp
jkelling: The pitch expected by `MapIdxPitch` is not in bytes but in elements, the same unit the index is in, because it does not know the element size. Not sure if this problem should be solved inside `MapIdxPitch` or outside...

include/alpaka/mem/buf/SetKernel.hpp
jkelling: This wants to say that, that threadElemExtent must be of the form {1,1,...,1,n}. I will move this to the doc string, with better wording.

include/alpaka/mem/buf/SetKernel.hpp
jkelling: I think we can.

According to cppreference.com, `std::logical_and` "Effectively calls operator&& on type T", thus with `foldr` calling `f(t0, foldr(f, t1, ts...))` this should short-circuit just like the loop, right? (Not important on GPU, but somewhat useful on CPU).

BenjaminW3: I am not sure. This is also what I thought about and my conclusion was that there will be no short circuiting because it is a function call and not an inline &&. I would still prefer the shorter code here.

include/alpaka/mem/buf/SetKernel.hpp
jkelling: Why is this iterator part of the test tree and not of alpaka? I could implement `SetKernel` using this iterator instead of MapIdxPitch. Could we move this into alpaka?

test/unit/idx/src/MapIdxPitch.cpp
jkelling: MapIdxPitch does fundamentally not work with padded buffers: One could give it information about `TElem` so it could take the padding into account, but then it would have to return either an index in bytes, which would produce weird code in the caller, or return a pointer instead of an index, which would basically turn it into an iterator.

include/alpaka/mem/buf/SetKernel.hpp
jkelling: Specifically in the case of `SetKernel` the current implementation does work because memset always sets memory byte-wise. The limit checking the row-extent in line 60,61 ensures that it does not write into the padding space.

include/alpaka/mem/buf/SetKernel.hpp
jkelling: It is already there: https://github.com/alpaka-group/alpaka/pull/1132/files/5757fb1795984bf9f5db4c17953e24bdadcfc2d5#diff-608534e792f3910755a21033659f815aR31

include/alpaka/mem/buf/SetKernel.hpp
BenjaminW3: The Iterator is not part of alpaka because nobody moved it there and it was only used for testing yet.
You can move it to `alpaka/mem/view` and add a short test. This would allow to remove `mapIdxPitchBytes`.

jkelling: There would also be some work on the iterator's interface required, I would prefer not to do this now, even though it would probably be a good addition to alpaka.

The iterator would add a bit of overhead compare to the current solution. As memset is exclusively working with bytes in any interface, a solution tailored for bytes seems appropriate. I changed the name and the doc strings to avoid confusion. Also, `SetKernel` not longer has `TElem` as a template argument but now works on `std::uint8_t`, which is what the `createTaskSet` trait requires.

BenjaminW3 (dismissed):

include/alpaka/mem/buf/SetKernel.hpp
```diff
@@ -0,0 +1,73 @@
...
+                    std::uint8_t * dst,
+                    TExtent extent,
+                    TPitch pitch) const
+                -> void
+                {
+                    using Idx = typename idx::traits::IdxType<TExtent>::type;
+                    auto const gridThreadIdx(alpaka::idx::getIdx<alpaka::Grid, alpaka::Threads>(acc));
+                    auto const threadElemExtent(alpaka::workdiv::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc));
+                    auto const idxThreadFirstElem = idx::getIdxThreadFirstElem(acc, gridThreadIdx, threadElemExtent);
+                    auto idx = idx::mapIdxPitchBytes<1u, dim::Dim<TAcc>::value>(idxThreadFirstElem, pitch)[0];
```
BenjaminW3, line 57: I find the name `idx` confusing now. Could this be renamed to `offset`?

BenjaminW3: On the other hand we are comparing and adding other index related variables, so it might be ok.

include/alpaka/mem/buf/SetKernel.hpp
```diff
@@ -0,0 +1,73 @@
...
+                    constexpr auto lastDim = dim::Dim<TAcc>::value - 1;
+                    const auto lastIdx = idx +
+                        std::min(threadElemExtent[lastDim], static_cast<Idx>(extent[lastDim]-idxThreadFirstElem[lastDim]));
+
+                    if( (idxThreadFirstElem < extent).foldrAll(std::logical_and<bool>()) )
+                    {
+                        for(; idx<lastIdx; ++idx)
+                        {
+                            dst[idx] = val;
```
BenjaminW3: We are not doing a real array access here. I would prefer a pointer calculation here `dst + idx` but this is only personal taste.

include/alpaka/mem/buf/SetKernel.hpp
jkelling: I prefer idx, because of the mapping from `gridThreadIdx`.

include/alpaka/mem/buf/SetKernel.hpp
jkelling: Changed.

BenjaminW3 (approved):

PR #1141, opened 2020-09-18
activate `-Wconversion` for tests 

psychocoderHPC:
* change interface `getBlockSharedMemDynSizeBytes`
  - trait is now always returning  `std::size_t`, before the return type was the index type of the accelerator
* add `-Wconversion` to the developer GCC options
* restrict types which can be used as index type for the accelerator
  - remove tests for accelerator index types `uint16_t` and `int16_t`

The shared memory traits and implementation mixed before the usage of `size_t`, `unsigned int` and accelerator index type.
This behaviour is changed by using always similar to the STL `size_t` as return value for the size.

Reasons for the index type restriction of an accelerator:
An operator e.g. `+`,`-` with types smaller than integer can return an `integer`. see https://stackoverflow.com/a/39061103
This behaviour results into sign errors/warning on every place in the code were we calculate with index types. 
It should not be an problem to remove the usage of 16bit and 8bit if we can increase the code quality/stability because of this restriction.

An additional motivation to check for conversion issues is that DLR require that alpaka compiles without conversion warnings/errors.

### DIFF
--- a/include/alpaka/atomic/Op.hpp
+++ b/include/alpaka/atomic/Op.hpp
@@ -38,8 +38,15 @@ namespace alpaka
                 {
                     auto const old(*addr);
                     auto & ref(*addr);
+#if BOOST_COMP_GNUC 
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wconversion"
+#endif
                     ref += value;
                     return old;
+#if BOOST_COMP_GNUC
+    #pragma GCC diagnostic pop
+#endif
                 }
             };
             //#############################################################################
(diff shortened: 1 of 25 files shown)


include/alpaka/atomic/Op.hpp
```diff
@@ -38,8 +38,15 @@ namespace alpaka
                 {
                     auto const old(*addr);
                     auto & ref(*addr);
+#if BOOST_COMP_GNUC 
+    #pragma GCC diagnostic push
+    #pragma GCC diagnostic ignored "-Wconversion"
```
psychocoderHPC, line 43: alpaka atomic operations support types smaller than the size of int. In that case the compiler is always warning about type conversions.
This PR is ignoring the error at this place but in general we need to take care which types can be used for atomic operations. 
This depends on the alpaka backend and should be handled in another PR.

psychocoderHPC:
This PR can be reviewed the current CI issues are unrelated, I try to fix this issues in a new PR.

BenjaminW3:
I already have a fix for the Mac issues in #1142

psychocoderHPC:
Could you please merge it without the OSX CI. #1142 is now merged to dev, it saves us few hours CI tests.
Except I should change something, in this case I will rebase.

BenjaminW3 (approved):

PR #1144, opened 2020-09-22
Remove namespace alpaka::block::sync and alpaka::block::shared

bernhardmgruber:
See #1034

### DIFF
--- a/example/reduce/src/kernel.hpp
+++ b/example/reduce/src/kernel.hpp
@@ -80,7 +80,7 @@ struct ReduceKernel
                                   TFunc func) const -> void
     {
         auto &sdata(
-            alpaka::block::shared::st::allocVar<cheapArray<T, TBlockSize>,
+            alpaka::block::st::allocVar<cheapArray<T, TBlockSize>,
                                                 __COUNTER__>(acc));
 
         const uint32_t blockIndex(static_cast<uint32_t>(
@@ -125,7 +125,7 @@ struct ReduceKernel
         if (threadIndex < n)
             sdata[threadIndex] = result;
 
-        alpaka::block::sync::syncBlockThreads(acc);
+        alpaka::block::syncBlockThreads(acc);
 
         // --------
         // Level 2: block + warp reduce, reading from shared memory
@@ -155,7 +155,7 @@ struct ReduceKernel
                     func(sdata[threadIndex],
                          sdata[threadIndex + currentBlockSizeUp]);
 
-            alpaka::block::sync::syncBlockThreads(acc);
+            alpaka::block::syncBlockThreads(acc);
         }
 
         // store block result to gmem
--- a/include/alpaka/acc/AccCpuFibers.hpp
+++ b/include/alpaka/acc/AccCpuFibers.hpp
@@ -78,9 +78,9 @@ namespace alpaka
                 atomic::AtomicNoOp         // thread atomics
             >,
             public math::MathStdLib,
-            public block::shared::dyn::BlockSharedMemDynAlignedAlloc,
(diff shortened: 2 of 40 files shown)


bernhardmgruber:
There is an alternative and I think I like it better: We could just add this to the end of alpaka.hpp:
```
namespace alpaka::block {
  using namespace alpaka::block::sync;
  using namespace alpaka::block::shared;
}
```
This avoids inline namespaces. What do you think?

bernhardmgruber:
After a discussion with @j-stephan I think we should really get rid of some inner namespaces.

j-stephan:
> After a discussion with @j-stephan I think we should really get rid of ~~some~~ **most** (if not **all**) inner namespaces.

FTFY. This is IMO especially true for the folder-namespace relation. Namespaces are intended to logically structure the code (and prevent name clashes), not mirror the file system. Which we should also talk about by the way, I don't really see the need to hide (for example) `QueueUniformCudaHipRtBase.hpp` in its own directory. See also:

https://medium.com/@steve_verm/matching-namespaces-to-folders-is-an-anti-pattern-e8edb3a6b8e5
https://arne-mertz.de/2016/06/organizing-directories-namespaces/

psychocoderHPC:
The hiding of `QueueUniformCudaHipRtBase.hpp` is coming from the problem of header only libraries. 
The question is how you provide a simple way to the user to see which parts are user interface and what is "internal" stuff.
In a "normal" C++ program you have header you expose and the implementation is in cpp files.
There is no need that the user knows about `QueueUniformCudaHipRtBase`, therefore it is in an other folder. 
I agree that it is not obvious that this could be a reason.

j-stephan:
>  The question is how you provide a simple way to the user to see which parts are user interface and what is "internal" stuff.

`namespace detail`

> There is no need that the user knows about QueueUniformCudaHipRtBase, therefore it is in an other folder.

How often does the user look at the folder structure of his libraries? How is this problem solved in the standard library or boost?

bernhardmgruber:
While the discussion might be fruitful, are there any specific issues with the proposed changes in this PR? If not, can you please proceed to merge it? :)

psychocoderHPC:
> There is an alternative and I think I like it better: We could just add this to the end of alpaka.hpp:
> 
> ```
> namespace alpaka::block {
>   using namespace alpaka::block::sync;
>   using namespace alpaka::block::shared;
> }
> ```
> 
> This avoids inline namespaces. What do you think?

IMO inline namespaces are the better way. With inline namespaces there is no different if you include `alpaka.hpp" or files directly, both ways will be equal where your suggestion depends on how you use alpaka.

psychocoderHPC:
@bernhardmgruber Why do you removed the namespaces instead of using inline namespaces?

I am asking because this change will break all existing codes and require a stronger refactoring on the user side. All atomic function and shared mem usages must be updated. With a `inline` namespace this change would not require any user code refactorings.

psychocoderHPC (changes requested):
I am not sure if it is required to break the interface for the user. Lets wait way @bernhardmgruber was suggesting this way instead of using inline namespaces.

bernhardmgruber:
> @bernhardmgruber Why do you removed the namespaces instead of using inline namespaces?
> 
> I am asking because this change will break all existing codes and require a stronger refactoring on the user side. All atomic function and shared mem usages must be updated. With a `inline` namespace this change would not require any user code refactorings.

Because inline namespaces do not work in most cases for this reason: https://github.com/alpaka-group/alpaka/issues/1034#issuecomment-696846048
Thus, the only options are `using` declarations or namespace removal.

psychocoderHPC (approved):

PR #1145, opened 2020-09-22
Using namespace alpaka::dev and alpaka::pltf into alpaka

bernhardmgruber:
See #1034

This PR simplifies the API on the user's side. But it is just one of many options we have. This PR is the least disruptive, since I do not change the existing namespace hierarchy inside of alpaka. Is this something we like and we care for?

Alternatively, we can also *remove* the namespaces `dev` and `pltf` and spill all their entities into the `alpaka` namespace directly.

### DIFF
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -205,3 +205,8 @@
 // vec
 #include <alpaka/vec/Vec.hpp>
 #include <alpaka/vec/Traits.hpp>
+
+namespace alpaka {
+    using namespace alpaka::dev;
+    using namespace alpaka::pltf;
+}
--- a/test/unit/warp/src/Activemask.cpp
+++ b/test/unit/warp/src/Activemask.cpp
@@ -83,13 +83,13 @@ public:
 TEMPLATE_LIST_TEST_CASE( "activemask", "[warp]", alpaka::test::acc::TestAccs)
 {
     using Acc = TestType;
-    using Dev = alpaka::dev::Dev<Acc>;
-    using Pltf = alpaka::pltf::Pltf<Dev>;
+    using Dev = alpaka::Dev<Acc>;
+    using Pltf = alpaka::Pltf<Dev>;
     using Dim = alpaka::dim::Dim<Acc>;
     using Idx = alpaka::idx::Idx<Acc>;
 
-    Dev const dev(alpaka::pltf::getDevByIdx<Pltf>(0u));
-    auto const warpExtent = alpaka::dev::getWarpSize(dev);
+    Dev const dev(alpaka::getDevByIdx<Pltf>(0u));
+    auto const warpExtent = alpaka::getWarpSize(dev);
     if (warpExtent == 1)
     {
         Idx const gridThreadExtentPerDim = 4;
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -179,8 +179,8 @@ auto main()
     using HostQueue = alpaka::queue::Queue<Host, HostQueueProperty>;
 
     // Select devices
-    auto const devAcc = alpaka::pltf::getDevByIdx<Acc>(0u);
-    auto const devHost = alpaka::pltf::getDevByIdx<Host>(0u);
+    auto const devAcc = alpaka::getDevByIdx<Acc>(0u);
+    auto const devHost = alpaka::getDevByIdx<Host>(0u);
 
     // Create queues
     DevQueue devQueue(devAcc);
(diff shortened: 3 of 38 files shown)


test/unit/warp/src/Activemask.cpp
```diff
@@ -83,13 +83,13 @@ class ActivemaskMultipleThreadWarpTestKernel
...
-    using Pltf = alpaka::pltf::Pltf<Dev>;
+    using Dev = alpaka::Dev<Acc>;
+    using Pltf = alpaka::Pltf<Dev>;
     using Dim = alpaka::dim::Dim<Acc>;
     using Idx = alpaka::idx::Idx<Acc>;
 
-    Dev const dev(alpaka::pltf::getDevByIdx<Pltf>(0u));
-    auto const warpExtent = alpaka::dev::getWarpSize(dev);
+    Dev const dev(alpaka::getDevByIdx<Pltf>(0u));
+    auto const warpExtent = alpaka::getWarpSize(dev);
```
sbastrakov, line 92: Just as a note. The changed version looks visually better to me, and looking through the changed part it mostly never became confusing. This line however may be.

bernhardmgruber: Hmm. I wonder how this can be confusing? I call `getWarpSize` and I pass in a `Dev` called `dev`. Of course I need to know what a warp is. But how could it not be clear, that I query a warp size of a device?

sbastrakov: To me, it is a subjective mismatch that warp is a really specific and low-level thing and feels weird to access it "directly" via `alpaka::` first-level namespace.

bernhardmgruber: I think I understand your concern, but I also do not think that separating lower-level functionality into different namespaces is a solution. I also cannot recall any C++ library that does that. I am also worried that deciding what is low level and what not might be subjective as well. I think **if** sub namespaces are used, then they should be used to separate domains (cf. std::filesystem, std::chrono or std::ranges).

bernhardmgruber:
Generally, do we prefer the `using` declarations to just removing the namespaces? I think I would rather remove the namespaces, because then we can avoid users using two different spellings: `alpaka::dev::DevCpu` and `alpaka::DevCpu`.

test/unit/warp/src/Activemask.cpp
BenjaminW3: >  I think if sub namespaces are used, then they should be used to separate domains (cf. std::filesystem, std::chrono or std::ranges).

At least from my perspective, this is exactly what is done in alpaka. The only difference is the granularity.

bernhardmgruber: Well, to some degree yes. Alpaka is trying to group the right things together. But the point is that there are far too many namespaces. We discussed this in length in VCs and in #1034. This is also the feedback we get from e.g. CERN.

And also in the standard library, `std::filesystem` is already so long, that you can frequently see an alias like `namespace fs = std::filesystem` in many examples on the web.

psychocoderHPC:
> Generally, do we prefer the `using` declarations to just removing the namespaces? I think I would rather remove the namespaces, because then we can avoid users using two different spellings: `alpaka::dev::DevCpu` and `alpaka::DevCpu`.

The disadvantage of `using namespaces` is that you need to include the file where using is defined, else your like and feel will be different. 
So IMO in this case removing will be better **BUT** I have a general issue:
Since our developer base increases we have many different opinions, what is good but results into many user interface breaking changes. I miss a clear goal where we like to break the user interface. IMO if we break it now so much we should have a clear vision what we like to change additional to that to not require with each release that the user is refactoring there app on mostly all sides.

bernhardmgruber:
> The disadvantage of `using namespaces` is that you need to include the file where using is defined, else your like and feel will be different.

That's a very valid concern. You could move the `using` declarations closer to the entities used. But generally this approach is probably brittle.

> Since our developer base increases we have many different opinions, what is good but results into many user interface breaking changes. I miss a clear goal where we like to break the user interface. IMO if we break it now so much we should have a clear vision what we like to change additional to that to not require with each release that the user is refactoring there app on mostly all sides.

The vision for the my namespace/API refactoring effort is here: https://github.com/alpaka-group/alpaka/issues/1034#issuecomment-665011587

We have been talking about refactoring the namespace usage for some months now and the discussion stalled again. I finally decided to start applying changes so the proposed changes and ideas do not get lost in time.

psychocoderHPC:
@bernhardmgruber Could you please update your description. The description is not showing what you did. I assume you used before `using` and later decided to remove namespaces.

include/alpaka/alpaka.hpp
```diff
@@ -205,3 +205,8 @@
 // vec
 #include <alpaka/vec/Vec.hpp>
 #include <alpaka/vec/Traits.hpp>
+
+namespace alpaka {
```
psychocoderHPC, line 209: This looks outdated since you removed the namespaces.

psychocoderHPC (changes requested):

include/alpaka/alpaka.hpp
bernhardmgruber: No, this is the current state of the PR. This PR has 2 using declarations in alpaka.hpp.

bernhardmgruber:
@psychocoderHPC the description is up-to-date. This PR uses `using` declarations. However, I would prefer removing the namespaces instead. I think I will close this PR and create a new one.

PR #1096, opened 2020-07-23
CMake: First draft of first-class CUDA support

j-stephan:
This is an early draft PR for integrating CMake's first class CUDA support, intended for discussion. Functionality-wise it is mostly the same as our old FindCUDA-based approach. Things that are different:

* No more appending of compiler flags to weird lists. Instead we can now use `target_compile_options`, `target_compile_definitions` and so on.
* The host compiler checks were removed. There is currently no way to identify the host compiler with CMake's first class CUDA support, see https://gitlab.kitware.com/cmake/cmake/-/issues/20901 for reference. Note that the host compiler no longer has to be identical to `CMAKE_CXX_COMPILER`.
* Verbose builds for debugging purposes are now triggered by `cmake --build --verbose` instead of setting an `ALPAKA_DEBUG` level. This is because `CUDA_VERBOSE_BUILD` is no longer available.

TODO:

* Set examples and test cases to CMake 3.18.

PR #1146, opened 2020-09-26
Use CMake first-class CUDA support

BenjaminW3:
This is an improved version of the first-class CMake CUDA support PR #1096.

* removes support for CMake 3.15.7, 3.16.8, 3.17.3 and enforces 3.18.0+
* removes tests for CUDA builds using libc++ instead of libstdc++ (I was not able to get this working)

required changes:
* If you want to select the CUDA compiler explicitly, you can specify `CMAKE_CUDA_COMPILER` or set `CUDACXX` to the path to the CUDA compiler e.g.. `nvcc` or `clang++`. The default seems to be `nvcc`.
* If you selected `CMAKE_CUDA_COMPILER: clang++` and want to specify a specific CUDA toolkit different from the default that CMake takes, you have to specify `CUDAToolkit_ROOT`.
* `ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION` has been replaced by `CMAKE_CUDA_SEPARABLE_COMPILATION`
* `ALPAKA_CUDA_ARCH="35;72"` has ben replaced by `CMAKE_CUDA_ARCHITECTURES="35;72"`


.github/workflows/ci.yml
```diff
@@ -211,253 +211,253 @@ jobs:
...
         - name: linux_clang-10_cuda-10.1_release
           os: ubuntu-latest
-          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 10,     ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.74.0, ALPAKA_CI_CMAKE_VER: 3.18.0,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CUDA_VERSION: "10.1", ALPAKA_CUDA_COMPILER: clang,                                               ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: OFF}
+          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 10,     ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.74.0, ALPAKA_CI_CMAKE_VER: 3.18.1,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CUDA_VERSION: "10.1", CMAKE_CUDA_COMPILER: clang,                                                ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: OFF}
```
bernhardmgruber, line 384: Is there a reason for switching std library here? Or just to have more variance on the CI tests?

.github/workflows/ci.yml
```diff
@@ -211,253 +211,253 @@ jobs:
...
         - name: linux_clang-9_cuda-10.0_debug
           os: ubuntu-latest
-          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 9,      ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.17.3,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CUDA_VERSION: "10.0", ALPAKA_CUDA_COMPILER: clang,                                               ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: OFF}
+          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 9,      ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.18.1,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CUDA_VERSION: "10.0", CMAKE_CUDA_COMPILER: clang,                                                ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: OFF}
```
bernhardmgruber: Is there a reason for switching std library here? Or just to have more variance on the CI tests?

.github/workflows/ci.yml
bernhardmgruber: See previous comment.

.github/workflows/ci.yml
```diff
@@ -211,253 +211,253 @@ jobs:
...
+          env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 5,        ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.66.0, ALPAKA_CI_CMAKE_VER: 3.18.1, OMP_NUM_THREADS: 3, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04"}
         - name: linux_gcc-6_debug_c++17
           os: ubuntu-latest
-          env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 6,        ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.70.0, ALPAKA_CI_CMAKE_VER: 3.16.5, OMP_NUM_THREADS: 2, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_CXX_STANDARD: 17}
+          env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 6,        ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.70.0, ALPAKA_CI_CMAKE_VER: 3.18.1, OMP_NUM_THREADS: 2, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_CXX_STANDARD: 17, CMAKE_CXX_EXTENSIONS: OFF}
```
bernhardmgruber: Is turning CXX_EXTENSIONS OFF specifically required for this PR? Or just again to give the tests more variety?

cmake/alpakaCommon.cmake
```diff
@@ -261,278 +261,146 @@ endif()
 #-------------------------------------------------------------------------------
 # Find CUDA.
 if(ALPAKA_ACC_GPU_CUDA_ENABLE)
+    include(CheckLanguage)
+    check_language(CUDA)
 
-    if(NOT DEFINED ALPAKA_CUDA_VERSION)
-        set(ALPAKA_CUDA_VERSION 9.0)
-    endif()
```
bernhardmgruber, line 319: Maybe I missed this, but what happens if I do not set `ALPAKA_CUDA_VERSION`?

Also see subsequent comment.

cmake/alpakaCommon.cmake
```diff
@@ -261,278 +261,146 @@ endif()
...
-            if(CUDA_VERSION VERSION_LESS 10.3)
-                set(ALPAKA_CUDA_ARCH "30" CACHE STRING "GPU architecture")
-            else()
-                set(ALPAKA_CUDA_ARCH "35" CACHE STRING "GPU architecture")
-            endif()
-            set(ALPAKA_CUDA_COMPILER "nvcc" CACHE STRING "CUDA compiler")
-            set_property(CACHE ALPAKA_CUDA_COMPILER PROPERTY STRINGS "nvcc;clang")
+    if(CMAKE_CUDA_COMPILER)
+        enable_language(CUDA)
+        find_package(CUDAToolkit ${ALPAKA_CUDA_VERSION} REQUIRED)
```
bernhardmgruber: Which CUDAToolkit version is searched here, if I do not set `ALPAKA_CUDA_VERSION`? Will this even error?

cmake/alpakaCommon.cmake
```diff
@@ -261,278 +261,146 @@ endif()
...
 
-            option(ALPAKA_CUDA_FAST_MATH "Enable fast-math" ON)
-            option(ALPAKA_CUDA_FTZ "Set flush to zero for GPU" OFF)
-            option(ALPAKA_CUDA_SHOW_REGISTER "Show kernel registers and create PTX" OFF)
-            option(ALPAKA_CUDA_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps (folder: nvcc_tmp)" OFF)
-            option(ALPAKA_CUDA_NVCC_EXPT_EXTENDED_LAMBDA "Enable experimental, extended host-device lambdas in NVCC" ON)
-            option(ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION "Enable separable compilation in NVCC" OFF)
+        target_compile_features(alpaka INTERFACE cuda_std_${ALPAKA_CXX_STANDARD})
```
bernhardmgruber: I still find it weird that we do not pass `CMAKE_CXX_STANDARD` here. But this can be a topic for a different discussion.

cmake/alpakaCommon.cmake
```diff
@@ -261,278 +261,146 @@ endif()
...
+            if(ALPAKA_CUDA_FAST_MATH)
+                target_compile_options(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:--use_fast_math>)
+            endif()
 
-                set(CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
+            if(ALPAKA_CUDA_FTZ)
+                target_compile_options(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:--ftz=true>)
+            else()
+                target_compile_options(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:--ftz=false>)
+            endif()
```
bernhardmgruber: Without being a CMake expert: Can we refactor this into something like:
```
 target_compile_options(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:--ftz=${ALPAKA_CUDA_FTZ}>)
```

bernhardmgruber (commented):
I looked over the changeset and could not find something glaringly wrong. The changes seem reasonable.

.github/workflows/ci.yml
BenjaminW3: This PR is currently based on #1153 where this originates from.

cmake/alpakaCommon.cmake
BenjaminW3: `ALPAKA_CUDA_FTZ` is currently either `ON` or `OFF` and not `true`/`false`. There is probably a nice way to write this if-else as generator expression but this is a purely stylistic change that is not necessary for this PR. If someone has time to refactor this, please provide the proposed syntax.

.github/workflows/ci.yml
BenjaminW3: This PR is currently having some issues in some build types with the `stdlib` parameter given to nvcc. Some CUDA versions have not been tested with libc++ so yes, this is for more variance.

cmake/alpakaCommon.cmake
BenjaminW3: I am currently experimenting with getting rid of this completely. The new way of doing this seems to set `CUDAToolkit_ROOT` env variable to the CUDA toolkit of your choice and else CMake will select one automatically.

cmake/alpakaCommon.cmake
BenjaminW3: See above

cmake/alpakaCommon.cmake
BenjaminW3: I hopefully got rid of the variable now.
If you want to select the CUDA compiler explicitly, you can specify `CMAKE_CUDA_COMPILER: nvcc` or `CMAKE_CUDA_COMPILER: clang`. The default seems to be `nvcc`. If you want to select a specific version different from the default that CMake takes, you have to specify `CUDACXX`.
If you selected `CMAKE_CUDA_COMPILER: clang` and want to specify a specific CUDA toolkit different from the default that CMake takes, you have to specify `CUDAToolkit_ROOT`.

cmake/alpakaCommon.cmake
BenjaminW3: With the new native CMake support, `ALPAKA_CXX_STANDARD` is not necessary anymore. alpaka would define its minimum requirements as `cxx_std_14` and `cuda_std_14`. The user is then able to overwrite this with `CMAKE_CXX_STANDARD=17` and `CMAKE_CUDA_STANDARD=17`. This is not as easy as before but the natively supported variant. I am not sure what happens when the user specifies two different versions but my expectation would be that CMake also forwards those different flags to nvcc/gcc respectively.
However, we still have to support HIP. There we explicitly have to pass the `std` flag ourselves. We would have to decide on a version there. We could pass the value of `CMAKE_CXX_STANDARD` to HIP if it is set and else use 14 as default.

However, I would like to do such a big change later independently.

cmake/alpakaCommon.cmake
bernhardmgruber: Thank you for the elaborated reply! The new approach also seems more reasonable to me, since I usually want to select a specific compiler instead of a CUDA version (with the selected compiler entailing a CUDA version).

.github/workflows/ci.yml
bernhardmgruber: OK I see. Is there a dependency of this PR on #1153 ? If so please state it in this PR's description.

cmake/alpakaCommon.cmake
bernhardmgruber: Makes perfect sense to delay such a change! Thank you for the explanation!

bernhardmgruber:
Could you please rebase to get rid of the conflicts so that the CI can start working? Thank you!

SimeonEhrig:
> This is an improved version of the first-class CMake CUDA support PR #1096.
> 
> Notes:
> 
>     * If you want to select the CUDA compiler explicitly, you can specify `CMAKE_CUDA_COMPILER: nvcc` or `CMAKE_CUDA_COMPILER: clang`. The default seems to be `nvcc`.
> 
>     * If you want to select a specific CUDA compiler version different from the default that CMake takes, you have to specify `CUDACXX`.
> 
>     * If you selected `CMAKE_CUDA_COMPILER: clang` and want to specify a specific CUDA toolkit different from the default that CMake takes, you have to specify `CUDAToolkit_ROOT`.
> 
>     * `ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION` has been replaced by `CMAKE_CUDA_SEPARABLE_COMPILATION`
> 
>     * `ALPAKA_CUDA_ARCH="35;72"` has ben replaced by `CMAKE_CUDA_ARCHITECTURES="35 72"`

@BenjaminW3 
Can you please also update the Sphinx Doc documentation: https://github.com/alpaka-group/alpaka/blob/develop/docs/source/advanced/cmake.rst

https://alpaka.readthedocs.io/en/latest/advanced/cmake.html

psychocoderHPC:
I converted this PR to draft to avoid accidentally merging

BenjaminW3:
I now got rid of most of the issues. Still not working is:
* nvcc + cl.exe (windows) builds are failing
* clang (6, 7 and 8) as CUDA compiler fails to be detected correctly by CMake - newer versions work

@alpaka-group/alpaka-maintainers I would be happy if someone could further experiment around with those failing builds.

psychocoderHPC:
> I now got rid of most of the issues. Still not working is:
> 
>     * nvcc + cl.exe (windows) builds are failing
> 
>     * clang (6, 7 and 8) as CUDA compiler fails to be detected correctly by CMake - newer versions work
> 
> 
> @alpaka-group/alpaka-maintainers I would be happy if someone could further experiment around with those failing builds.

I will bring this topic in the next developer meeting tomorrow 10am up. @BenjaminW3 do you want to join ?

script/run_generate.sh
```diff
@@ -72,20 +76,23 @@ fi
...
-    "$(env2cmake ALPAKA_CUDA_FAST_MATH)" "$(env2cmake ALPAKA_CUDA_FTZ)" "$(env2cmake ALPAKA_CUDA_SHOW_REGISTER)" "$(env2cmake ALPAKA_CUDA_KEEP_FILES)" "$(env2cmake ALPAKA_CUDA_NVCC_EXPT_EXTENDED_LAMBDA)" "$(env2cmake ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION)" \
+    "$(env2cmake ALPAKA_ACC_GPU_CUDA_ENABLE)" "$(env2cmake ALPAKA_ACC_GPU_CUDA_ONLY_MODE)" "$(env2cmake CMAKE_CUDA_ARCHITECTURES)" "$(env2cmake CMAKE_CUDA_COMPILER)" \
+    "$(env2cmake ALPAKA_CUDA_FAST_MATH)" "$(env2cmake ALPAKA_CUDA_FTZ)" "$(env2cmake ALPAKA_CUDA_SHOW_REGISTER)" "$(env2cmake ALPAKA_CUDA_KEEP_FILES)" "$(env2cmake ALPAKA_CUDA_NVCC_EXPT_EXTENDED_LAMBDA)" "$(env2cmake CMAKE_CUDA_SEPARABLE_COMPILATION)" \
     "$(env2cmake ALPAKA_ACC_GPU_HIP_ENABLE)" "$(env2cmake ALPAKA_ACC_GPU_HIP_ONLY_MODE)" "$(env2cmake ALPAKA_HIP_PLATFORM)" \
     "$(env2cmake ALPAKA_EMU_MEMCPY3D)" \
     "$(env2cmake ALPAKA_DEBUG)" "$(env2cmake ALPAKA_CI)" "$(env2cmake ALPAKA_CI_ANALYSIS)" "$(env2cmake ALPAKA_CXX_STANDARD)" \
     ".."
 
+more "CMakeFiles/CMakeOutput.log"
+more "CMakeFiles/CMakeError.log"
```
bernhardmgruber: Since at least 1 windows build fails because `more` could not be found by bash: Could you maybe use `cat` here instead?

bernhardmgruber:
If it helps in any way, I just tried your branch on my Windows 10 with latest VS2019, CMake 3.18.2 and nvcc 11 and it worked just fine:

```
PS C:\dev\alpaka_ben\build> cmake -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DCMAKE_TOOLCHAIN_FILE="D:\dev\vcpkg\scripts\buildsystems\vcpkg.cmake" ..
-- Building for: Visual Studio 16 2019
-- Selecting Windows SDK version 10.0.17763.0 to target Windows 10.0.19042.
-- The CXX compiler identification is MSVC 19.27.29112.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Community/VC/Tools/MSVC/14.27.29110/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Boost: D:/dev/vcpkg/installed/x64-windows/include (found suitable version "1.70.0", minimum required is "1.65.1") found components: fiber context
-- Looking for a CUDA compiler
-- Looking for a CUDA compiler - C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.0/bin/nvcc.exe
-- The CUDA compiler identification is NVIDIA 11.0.221
-- Detecting CUDA compiler ABI info
-- Detecting CUDA compiler ABI info - done
-- Check for working CUDA compiler: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.0/bin/nvcc.exe - skipped
-- Detecting CUDA compile features
-- Detecting CUDA compile features - done
-- Found CUDAToolkit: C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.0/include (found version "11.0.221")
-- nvcc is used as CUDA compiler
-- ALPAKA_ACC_GPU_CUDA_ENABLED
-- Configuring done
-- Generating done
-- Build files have been written to: C:/dev/alpaka_ben/build
```

I also tried the exact same CMake command line as the build slave `windows_nvcc-11.1_cl-2019_debug` and it finished successfully, whereas the build slave fails.

script/run_generate.sh
BenjaminW3: I added this yesterday only for enhanced debugging. I will replace it with something platform independent.

BenjaminW3:
Good to know that it works locally for you. I will try to do some changes to the way we install cuda in CI.

script/run.sh
```diff
@@ -59,25 +59,30 @@ fi
...
     then
-        export PATH="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${ALPAKA_CUDA_VERSION}\bin":$PATH
-        export CUDA_PATH="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${ALPAKA_CUDA_VERSION}"
+        export PATH="/c/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v${ALPAKA_CI_CUDA_VERSION}/bin":$PATH
+        export CUDA_PATH="/c/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v${ALPAKA_CI_CUDA_VERSION}"
+
+        which nvcc
+        nvcc -V
+
+        export CMAKE_CUDA_COMPILER=\"$(which nvcc)\"
```
bernhardmgruber: I played around a bit and I think you can revert your changes to PATH and CUDA_PATH and also remove the definition of CMAKE_CUDA_COMPILER.

In addition, you need this:
`export CUDA_PATH_V${ALPAKA_CI_CUDA_VERSION/./_}="C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v${ALPAKA_CI_CUDA_VERSION}"`

script/run.sh
```diff
@@ -113,29 +117,35 @@ then
...
 
     which "${CXX}"
     ${CXX} -v
+fi
+
+CMAKE_CUDA_FLAGS=
+if [ ! -z "${CMAKE_CXX_FLAGS+x}" ]
+then
+    CMAKE_CUDA_FLAGS=${CMAKE_CXX_FLAGS}
+fi
```
bernhardmgruber: That seems weird to me.

BenjaminW3: Our build scripts provide some `CMAKE_CXX_FLAGS` from the outside to be applied to all the code that is built by CMake later. This does not work with the builtin CUDA support because this will not use `CMAKE_CXX_FLAGS` but `CMAKE_CUDA_FLAGS`.
As I did not want to make the CI scripts adapt to the compiler, I simply fill both of the variables with the same content.

script/run.sh
BenjaminW3: Thanks I will try this out. But first I have to rebase this branch and get it working again.

bernhardmgruber:
So there is a second big issue. All compile options that you add for MSVC's cl.exe seem to end up on nvcc's command line. Since they all start with a `/`, nvcc thinks these are input files and complains. I played a bit here and removed all options for MSVC, that nvcc chokes on: https://github.com/bernhardmgruber/alpaka/commit/6c7321de8a7b5f18bdeb95a1e4932a14b2035da6 However, that is not a solution. But it makes nvcc happy :) I leave it to you to find a better solution for this! I guess you need some of those CMake generator expressions to make sure the flags end up at MSVC. Or maybe prefix all the options with `-Xcompiler`.

script/run.sh
bernhardmgruber: I am not sure if this is also part of the problem described below (CXX flags that end up on nvcc's command line). You should make sure that nvcc does not get any MSVC CXX flags.

bernhardmgruber:
The remaining failing Windows builds now show, that we still ned /bigobj. And there is the recently introduced bug, that makes nvcc fail compiling something inside the MS standard library. Looks almost done!

script/run.sh
BenjaminW3: I solved both issues by using generator expressions.

BenjaminW3:
It looks like there is a bug in CMake (at least 3.19.0) when it tries to find the supported `CMAKE_CUDA_ARCHITECTURES` of a combination of MSVC and NVCC:
```
+ cmake --verbose --log-level=trace -G 'Visual Studio 16 2019' -A x64 -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON '-DBOOST_ROOT=D:\a\alpaka\alpaka/boost' '-DBOOST_LIBRARYDIR=D:\a\alpaka\alpaka/boost_libs/lib' -DBoost_USE_STATIC_LIBS=ON -DBoost_USE_MULTITHREADED=ON -DBoost_USE_STATIC_RUNTIME=OFF -DBoost_ARCHITECTURE=-x64 -DCMAKE_BUILD_TYPE=Release '' '' '' '' -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=OFF -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE=OFF -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE=OFF -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=OFF -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=OFF -DALPAKA_ACC_ANY_BT_OMP5_ENABLE=OFF -DALPAKA_ACC_ANY_BT_OACC_ENABLE=OFF '' -DTBB_ROOT=tbb -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DALPAKA_ACC_GPU_CUDA_ONLY_MODE=ON -DCMAKE_CUDA_ARCHITECTURES=35 '' '' '' '' '' '' '' -DALPAKA_ACC_GPU_HIP_ENABLE=OFF -DALPAKA_ACC_GPU_HIP_ONLY_MODE=OFF '' '' -DALPAKA_CI=GITHUB -DALPAKA_CI_ANALYSIS=OFF '' ..
16.8.1.52902-- The CXX compiler identification is MSVC 19.28.29333.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files (x86)/Microsoft Visual Studio/2019/Enterprise/VC/Tools/MSVC/14.28.29333/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Found Boost: D:/a/alpaka/alpaka/boost (found suitable version "1.74.0", minimum required is "1.65.1")  missing components: fiber context
-- Looking for a CUDA compiler
-- Looking for a CUDA compiler - C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v11.1/bin/nvcc.exe
-- The CUDA compiler identification is NVIDIA 11.1.74
CMake Error at C:/Program Files/CMake/share/cmake-3.19/Modules/CMakeDetermineCUDACompiler.cmake:572 (message):
  The CMAKE_CUDA_ARCHITECTURES:

    35

  do not all work with this compiler.  Try:
See also "D:/a/alpaka/alpaka/build/CMakeFiles/CMakeOutput.log".

    52

  instead.
  cmake/alpakaCommon.cmake:277 (enable_language)
  CMakeLists.txt:85 (include)
```
It is already reported here: https://discourse.cmake.org/t/incorrect-cuda-architecture-detection/2018/2

For now I will try to work around this by selecting exactly the single version that CMake thinks that is supported but ultimately this is a regression.

BenjaminW3:
@bernhardmgruber did you try setting `CMAKE_CUDA_ARCHITECTURES` locally and got a similar broken result?

bernhardmgruber:
No, I do not specify `CMAKE_CUDA_ARCHITECTURES`. I just tried your branch and ran:
```bash
cmake -DCMAKE_TOOLCHAIN_FILE="D:\dev\vcpkg\scripts\buildsystems\vcpkg.cmake" -DALPAKA_ACC_GPU_CUDA_ENABLE=ON ..
cmake --build . --target kernel
```
And that configured, generated and compiled successfully.

IDK, should we drop setting `CMAKE_CUDA_ARCHITECTURES` from the CI?

psychocoderHPC:
> IDK, should we drop setting `CMAKE_CUDA_ARCHITECTURES` from the CI?

Note only: We need a way in CMake to define the architecture or do you suggest removing it only for the WINDOWS CI?

BenjaminW3:
For now I set it to the only version that CMake thinks that is supported for MSVC. No need to remove it.

BenjaminW3:
The CMake bug above for Visual Studio Architecture selection has been fixed in CMake 3.18.x and 3.19.x: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5531
Now we have to wait for them to be released.

psychocoderHPC:
> The CMake bug above for Visual Studio Architecture selection has been fixed in CMake 3.18.x and 3.19.x: https://gitlab.kitware.com/cmake/cmake/-/merge_requests/5531
> Now we have to wait for them to be released.

Could you not apply for the older CMake versions a fix that `sm_50` or whatever is the minimal compute architecture?

BenjaminW3:
The issue is that CMake does not allow any other architecture than the nvcc default of the respective CUDA version when using Visual Studio, so setting a minumum does not help. You can not select higher versions, you can not select multiple versions.
I think we have to wait for the patch releases.

script/run.sh
```diff
@@ -114,29 +118,34 @@ then
...
 
-        if [ -z "${CMAKE_EXE_LINKER_FLAGS+x}" ]
+        if [ "${CXX}" == "clang++" ]
         then
-            export CMAKE_EXE_LINKER_FLAGS=
+            CMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS} -stdlib=libc++"
+            CMAKE_CUDA_FLAGS="${CMAKE_CUDA_FLAGS} -stdlib=libc++"
+        #elif [ "${CXX}" == "g++" ]
+        #then
+        #    CMAKE_CXX_FLAGS="${CMAKE_CXX_FLAGS} -nostdinc++ -nodefaultlibs -lc++ -lc++abi -lm -lc -lgcc_s -lgcc -I<libcxx-install-prefix>/include/c++/v1"
```
bernhardmgruber: Is this still needed?

BenjaminW3:
With CMake 3.19.2 the issue with the MSVC builds is fixed.

Now the only remaining problem is that CMake is not happy with clang 6, 7 and 8 as CUDA compilers.
This is still the same issue we had in the original PR: https://github.com/alpaka-group/alpaka/pull/1096#issuecomment-692218860

Can anyone help here?

bernhardmgruber:
@BenjaminW3 I would like to express my gratitude that you are looking into native CUDA support! I would be super happy if this could land in alpaka!
I use native CUDA support in LLAMA as well and since i have alpaka examples, I always end up with a mix of native CUDA and FindCUDA. I would be glad if that mess could go away!
Thank you for your time!

BenjaminW3:
Now that the support for clang 6, 7 and 8 as CUDA compiler is officially gone, I was able to get all builds green!

script/run.sh
```diff
@@ -59,25 +59,30 @@ fi
...
     elif [ "$ALPAKA_CI_OS_NAME" = "Windows" ]
     then
-        export PATH="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${ALPAKA_CUDA_VERSION}\bin":$PATH
-        export CUDA_PATH="C:\Program Files\NVIDIA GPU Computing Toolkit\CUDA\v${ALPAKA_CUDA_VERSION}"
+        export PATH="/c/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v${ALPAKA_CI_CUDA_VERSION}/bin":$PATH
+        export CUDA_PATH="/c/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v${ALPAKA_CI_CUDA_VERSION}"
+
+        which nvcc
+        nvcc -V
+
```
bernhardmgruber, line 84: Is that still needed?

test/common/CMakeLists.txt
```diff
@@ -22,18 +22,26 @@ add_library(${_COMMON_TARGET_NAME} INTERFACE)
 target_include_directories(${_COMMON_TARGET_NAME} INTERFACE ${_COMMON_INCLUDE_DIRECTORY})
 
 include(${_COMMON_COMPILE_OPTIONS_FILE})
+message(STATUS "ALPAKA_DEV_COMPILE_OPTIONS: ${ALPAKA_DEV_COMPILE_OPTIONS}")
```
bernhardmgruber, line 25: Do we still need this?

test/common/CMakeLists.txt
```diff
@@ -22,18 +22,26 @@ add_library(${_COMMON_TARGET_NAME} INTERFACE)
...
     # CUDA driver API is used by EventHostManualTrigger
-    target_link_libraries(${_COMMON_TARGET_NAME} INTERFACE "${CUDA_CUDA_LIBRARY}")
+    target_link_libraries(${_COMMON_TARGET_NAME} INTERFACE CUDA::cuda_driver)
     target_compile_definitions(${_COMMON_TARGET_NAME} INTERFACE "CUDA_API_PER_THREAD_DEFAULT_STREAM")
 endif()
+if(ALPAKA_ACC_GPU_CUDA_ENABLE AND CMAKE_CUDA_COMPILER_ID STREQUAL "Clang")
+    # nvcc supports werror starting with 10.2
+    if(CUDA_VERSION GREATER_EQUAL 10.2)
+        message(STATUS "adding -Werror=all-warnings")
+        target_compile_options(${_COMMON_TARGET_NAME} INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:-Werror=all-warnings>)
```
bernhardmgruber, line 42: Should we add this via `alpaka_set_compiler_options`?

bernhardmgruber (commented):
Generally, looks very good!

test/common/CMakeLists.txt
BenjaminW3: I do not think so. `alpaka_set_compiler_options` is used for options that are added to the alpaka target which are transitively applied to user code. This setting only applies to our internal unit tests.

test/common/CMakeLists.txt
BenjaminW3: I would like to keep this for debugging purposes. It only writes the additional compiler options we use for the tests.

script/run.sh
BenjaminW3: no, I will remove it

script/run.sh
BenjaminW3: We do the same above for linux and it is good for debugging, so I would keep it.

bernhardmgruber: Ok!

test/common/CMakeLists.txt
bernhardmgruber: Sounds good to me!

bernhardmgruber (dismissed):
LGTM!

Still, I would like someone more familiar with the alpaka CMakeLists have a look as well! @psychocoderHPC could you also please have a quick look? Thanks!

.github/workflows/ci.yml
```diff
@@ -24,7 +24,7 @@ on: [push, pull_request]
 # ALPAKA_CI                                     : {GITHUB}
 # ALPAKA_CI_DOCKER_BASE_IMAGE_NAME              : {ubuntu:16.04, ubuntu:18.04, ubuntu:20.04}
 # ALPAKA_CI_BOOST_BRANCH                        : {boost-1.65.1, boost-1.66.0, boost-1.67.0, boost-1.68.0, boost-1.69.0, boost-1.70.0, boost-1.71.0, boost-1.72.0, boost-1.73.0, boost-1.74.0}
-# ALPAKA_CI_CMAKE_VER                           : {3.15.7, 3.16.9, 3.17.5, 3.18.6, 3.19.7, 3.20.0}
```
j-stephan, line 27: Do we have a policy on which patch versions we use? With this change we will test `3.18.6`, `3.19.7` and `3.20.0`, but the latest versions are `3.18.6`, `3.19.8` and `3.20.2`.

CMakeLists.txt
```diff
@@ -117,24 +117,12 @@ endif()
...
-    endif()
-
     if(ALPAKA_ACC_GPU_HIP_ENABLE)
         # export HIPCC flags to parent scope in case alpaka is another project's subdirectory
         set(HIP_HIPCC_FLAGS ${HIP_HIPCC_FLAGS} PARENT_SCOPE)
         set(HIP_NVCC_FLAGS ${HIP_NVCC_FLAGS} PARENT_SCOPE)
         set(HIP_VERBOSE_BUILD ${HIP_VERBOSE_BUILD} PARENT_SCOPE)
     endif()
-
-    if((ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE) AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
```
j-stephan, line 135: Shouldn't we keep this for the HIP + nvcc case?

cmake/alpakaCommon.cmake
```diff
@@ -313,257 +313,146 @@ endif()
...
+            if(ALPAKA_ACC_ANY_BT_OMP5_ENABLE)
+                message(FATAL_ERROR "Clang as a CUDA compiler does not support OpenMP 5!")
+            endif()
 
-                if(ALPAKA_CUDA_SHOW_REGISTER STREQUAL ON)
-                    alpaka_set_compiler_options(DEVICE target alpaka "-Xcuda-ptxas=-v")
-                endif()
+            # libstdc++ since version 7 when GNU extensions are enabled (e.g. -std=gnu++11)
+            # use `__CUDACC__` to avoid defining overloads using non-standard `__float128`.
```
j-stephan: There is a `s` missing here (`use` -> `uses`).

test/unit/math/CMakeLists.txt
```diff
@@ -15,8 +15,7 @@ append_recursive_files_add_to_src_group("src/" "src/" "hpp" _FILES_HEADER)
 
 if(ALPAKA_ACC_GPU_CUDA_ENABLE)
     list(REMOVE_ITEM
-            CUDA_NVCC_FLAGS "--ftz=true" "--prec-div=false" "--prec-sqrt=false" "--fmad=true" "--use_fast_math" "-use_fast_math")
-
+            CMAKE_CXX_FLAGS "--ftz=true" "--prec-div=false" "--prec-sqrt=false" "--fmad=true" "--use_fast_math" "-use_fast_math")
```
j-stephan, line 18: Shouldn't this be `CMAKE_CUDA_FLAGS`?

j-stephan: Shouldn't this be `CMAKE_CUDA_FLAGS`?

script/run.sh
```diff
@@ -59,25 +59,30 @@ fi
...
+        export CUDA_PATH_V${ALPAKA_CI_CUDA_VERSION/./_}="C:/Program Files/NVIDIA GPU Computing Toolkit/CUDA/v${ALPAKA_CI_CUDA_VERSION}"
```
j-stephan, line 85: (I'm no Windows expert, so this is pure curiosity:) Is there a reason why we have the `C:/Program Files/` variant here and the `/c/Program Files/` variant above?

cmake/alpakaCommon.cmake
```diff
@@ -313,257 +313,146 @@ endif()
...
-                # correctly propagate OpenMP flags
-                # This can be removed once we support CMake's first class CUDA support.
-                target_compile_options(alpaka INTERFACE ${OpenMP_CXX_FLAGS})
-            endif()
-
-            target_link_libraries(alpaka INTERFACE ${CUDA_CUDART_LIBRARY})
-            target_include_directories(alpaka INTERFACE ${CUDA_INCLUDE_DIRS})
         endif()
+
+        target_link_libraries(alpaka INTERFACE CUDA::cudart)
```
j-stephan, line 452: Is this required when nvcc is the CUDA compiler? (Also the next line)

j-stephan (changes requested):
Looks great, thanks for the effort! A few minor comments:

script/run.sh
bernhardmgruber: My guess is that the `run.sh` script is executed in a bash on Windows, which uses `/c/Program Files/` path convensions, whereas every normal Windows tool that will run after bash uses the normal `C:/Program Files` path conventions.

cmake/alpakaCommon.cmake
bernhardmgruber: I encountered this as well with LLAMA. If you just enable the CUDA language, you get a compiler able to compile CUDA. But you will not automatically get the proper include directories set for the target. So e.g. you IDE cannot find `<cuda_runtime.h>`. nvcc might still find the headers without this line though.

script/run.sh
sbastrakov: `/c/Program Files/` generally doesn't work on Windows. However it would sometimes work, like inside certain terminals or some tools may understand it.

.github/workflows/ci.yml
BenjaminW3: Every time I add a new version, I simply update the patch releases of the previous releases. I did not want to mix such changes in here.

test/unit/math/CMakeLists.txt
BenjaminW3: To be honest, I do not know.
CI was happy. If you want, I could try to remove it completely because if it was wrong after my change, it is unnecessary.

j-stephan: Yes, that's a good idea. If it doesn't break the CI we should remove this since it seems unnecessary.

psychocoderHPC: please use the new compiler option manipulations from #1289 (`alpaka_set_compiler_options`).

I do not understand why do we remove flags here?

psychocoderHPC: > I do not understand why do we remove flags here?

Ahh now I get it, it is for the math test only.

CMakeLists.txt
BenjaminW3: I kept it for HIP for now but someone with more knowledge about it should check this in the future.

cmake/alpakaCommon.cmake
BenjaminW3: fixed

j-stephan (approved):
LGTM.

j-stephan:
Rerunning the CI because one of the required jobs was cancelled and I couldn't access its log for some reason. If all goes well I'll hit the merge button.

BenjaminW3:
We should try to give it some usage in other projects. I am pretty sure that we will find some issues.

bernhardmgruber:
Just tried in LLAMA, works fine!

PR #1147, opened 2020-09-26
Add support for CUDA 11.1

BenjaminW3:
and update CUDA 10.0 installation to the latest version.

No changes were necessary.

We should think about removing support for some compiler/toolchain versions because the number of tested combinations is getting very large.


psychocoderHPC:
I think we still support CUDA 9.2.
I will ping DLR again that they should check if there are still open BUGs and if not we should release a new version of alpaka and than skip CUDA 9.X

psychocoderHPC (approved):

PR #1153, opened 2020-10-01
Test CMAKE_CXX_EXTENSIONS=OFF

BenjaminW3:
* Test that alpaka also builds with `CMAKE_CXX_EXTENSIONS=OFF` which makes CMake use `-std=c++17` instead of `-std=gnu17` or similar.
* Fix that we do not pass empty parameters e.g. `-DCMAKE_CXX_EXTENSIONS=` to CMake.
* Remove the `__float128=void` workaround. It is not necessary anymore.


BenjaminW3:
The failing mac build did not start. The PR can be merged without this.

script/run_generate.sh
```diff
@@ -72,10 +72,10 @@ fi
 mkdir -p build/
 cd build/
 
-"${ALPAKA_CI_CMAKE_EXECUTABLE}" -G "${ALPAKA_CI_CMAKE_GENERATOR}" ${ALPAKA_CI_CMAKE_GENERATOR_PLATFORM} \
+"${ALPAKA_CI_CMAKE_EXECUTABLE}" --verbose -G "${ALPAKA_CI_CMAKE_GENERATOR}" ${ALPAKA_CI_CMAKE_GENERATOR_PLATFORM}\
```
bernhardmgruber, line 75: Adding this flag feels not necessary for this PR to work. Is this a functional change? Or purely more output from cmake?

bernhardmgruber (approved):

script/run_generate.sh
psychocoderHPC: @BenjaminW3 Could you check this question so that I know if I can merge this PR.

BenjaminW3: Ah, I overlooked this. This change was part of @j-stephan s initial PR for first class CUDA support with CMake because there we lose the option to improve the debug output only for the `FindCUDA` script.
I found it useful for investigation here even without the CUDA change so I added it here.
As this is only in CI, it does not have an influence on our users.

psychocoderHPC (approved):

PR #1173, opened 2020-10-09
Remove some mem related namespaces

bernhardmgruber:
* remove namespace alpaka::buf::cpu
* remove namespace alpaka::buf::omp5
* remove namespace alpaka::buf
* remove namespace alpaka::alloc
* remove namespace alpaka::view::uniform_cuda_hip
* rename memory allocation function/trait alloc/Alloc to malloc/Malloc
* rename buffer allocation function alloc to allocBuf

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -220,7 +220,7 @@ auto main()
     // The `alloc` method returns a reference counted buffer handle.
     // When the last such handle is destroyed, the memory is freed automatically.
     using BufHost = alpaka::Buf<Host, Data, Dim, Idx>;
-    BufHost hostBuffer(alpaka::alloc<Data, Idx>(devHost, extents));
+    BufHost hostBuffer(alpaka::allocBuf<Data, Idx>(devHost, extents));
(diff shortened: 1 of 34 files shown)


bernhardmgruber:
With the removal of the `buf` namespace, we now have two functions `alloc`:

1. `auto alpaka::alloc(TAlloc const & alloc, std::size_t const & sizeElems) -> T *`
    Is a kind of aligned memory allocator. The returned pointer must be passed to `alpaka::free()`.
2. `auto alpaka::alloc(TDev const & dev, TExtent const & extent = TExtent())`
    Creates buffers, which destroy themselves using their dtor.

I propose to rename one of them. Either the first one to `alpaka::malloc()` or the second one to `alpaka::allocBuf` or `alpaka::createBuf`. What do you think?

sbastrakov:
Or maybe rename them both? As otherwise the remaining `alloc` may be still confusing. I like the suggested names.

psychocoderHPC (changes requested):
There are some compile issues in the CI: e.g. https://github.com/alpaka-group/alpaka/pull/1173/checks?check_run_id=1232144344#step:5:4380

psychocoderHPC (approved):

PR #1182, opened 2020-10-21
Remove namespace alpaka::block and some inner namespaces

bernhardmgruber:
See #1034

Also renames the namespace `alpaka::block::op` to `alpaka::blockOp` to not be confused with `alpaka::op` containing atomic operators. I think we should debate this!

### DIFF
--- a/include/alpaka/block/sync/Traits.hpp
+++ b/include/alpaka/block/sync/Traits.hpp
@@ -58,8 +58,8 @@ namespace alpaka
     }
 
     //-----------------------------------------------------------------------------
-    //! Defines operation functors.
-    namespace op
+    //! Defines block operation functors.
+    namespace blockOp
     {
         //#############################################################################
         //! The addition function object.
--- a/include/alpaka/block/sync/Traits.hpp
+++ b/include/alpaka/block/sync/Traits.hpp
@@ -16,134 +16,129 @@
 
 namespace alpaka
 {
+    struct ConceptBlockSync{};
+
     //-----------------------------------------------------------------------------
-    //! The grid block specifics
-    namespace block
+    //! The block synchronization traits.
+    namespace traits
     {
-        struct ConceptBlockSync{};
-
-        //-----------------------------------------------------------------------------
-        //! The block synchronization traits.
-        namespace traits
-        {
-            //#############################################################################
-            //! The block synchronization operation trait.
-            template<
-                typename TBlockSync,
-                typename TSfinae = void>
-            struct SyncBlockThreads;
-
-            //#############################################################################
-            //! The block synchronization and predicate operation trait.
-            template<
-                typename TOp,
-                typename TBlockSync,
-                typename TSfinae = void>
-            struct SyncBlockThreadsPredicate;
-        }
-
-        //-----------------------------------------------------------------------------
-        //! Synchronizes all threads within the current block (independently for all blocks).
-        //!
-        //! \tparam TBlockSync The block synchronization implementation type.
-        //! \param blockSync The block synchronization implementation.
-        ALPAKA_NO_HOST_ACC_WARNING
+        //#############################################################################
+        //! The block synchronization operation trait.
         template<
-            typename TBlockSync>
-        ALPAKA_FN_ACC auto syncBlockThreads(
-            TBlockSync const & blockSync)
-        -> void
-        {
-            using ImplementationBase = concepts::ImplementationBase<ConceptBlockSync, TBlockSync>;
-            traits::SyncBlockThreads<
-                ImplementationBase>
-            ::syncBlockThreads(
-                blockSync);
-        }
+            typename TBlockSync,
+            typename TSfinae = void>
+        struct SyncBlockThreads;
 
-        //-----------------------------------------------------------------------------
-        //! Defines operation functors.
-        namespace op
-        {
-            //#############################################################################
-            //! The addition function object.
-            struct Count
-            {
-                enum { InitialValue = 0u};
-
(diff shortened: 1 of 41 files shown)


include/alpaka/block/sync/Traits.hpp
```diff
@@ -16,134 +16,129 @@
...
-        //! \param blockSync The block synchronization implementation.
-        ALPAKA_NO_HOST_ACC_WARNING
-        template<
-            typename TBlockSync>
-        ALPAKA_FN_ACC auto syncBlockThreads(
-            TBlockSync const & blockSync)
-        -> void
+    //-----------------------------------------------------------------------------
+    //! Defines block operation functors.
+    namespace blockOp
```
bernhardmgruber, line 62: Are we cool with this? I did not want to rename to `alpaka::op` because this namespace already contains the atomic operations and then users could expect to be able to call:
```
alpaka::atomicOp<alpaka::op::LogicalAnd>()
```
which is an error, because `LogicalAnd` is a block operation and not an atomic operation. This would work:
```
alpaka::atomicOp<alpaka::op::And>()
```

j-stephan: Wouldn't it make more sense to rename `alpaka::op::And` to `alpaka::atomicAnd` and then remove the entire `op` namespace? We wouldn't need `blockOp` then.

bernhardmgruber: I would very much just have explicit atomic functions, such as: `alpaka::atomicAnd(acc, T* addr, T value) -> T`. So we cannot name the functor that implements the atomic operation the same. We could name it `alpaka::AtomicAnd` though. And `alpaka::atomicAnd()` just calls `alpaka::atomicOp<alpaka::AtomicAnd>()`.

Afterwards, we can rename `alpaka::blockop::LogicalAnd` into `alpaka::BlockAnd`, which is then used as `alpaka::syncBlockThreadsPredicate<alpaka::BlockAnd>(acc, predicate)`.

What do you think?

bernhardmgruber: We could also make the case again that there should be namespaces for the functors like `alpaka::atomic::And` and `alpaka::block::And`. However, I would leave the user API functions outside, so a user calls `alpaka::atomicOp<alpaka::atomic::And>()` and `alpaka::syncBlockThreadsPredicate<alpaka::block::And>(acc, predicate)`.

sbastrakov: I am very much for introducing renamings like `alpaka::atomicAnd` (not in this PR of course). However, with this naming in mind, I find having `alpaka::AtomicAnd` to denote the operation confusing. As normally in such situations we follow the naming that capitalized means functor and small case means the corresponding function, that would not be the case here.

sbastrakov (approved):

include/alpaka/block/sync/Traits.hpp
bernhardmgruber: Here is a renaming of the atomic functions: #1185

`alpaka::AtomicAnd` would be the functor. The function using the functor is called `alpaka::atomicOp<Functor>(...)`.

sbastrakov: Yes, but then this function is also renamed as `alpaka::atomicAnd()`, which is connected but does not exactly match the meaning of `alpaka::AtomicAnd` ? Or did I get your idea wrong?

bernhardmgruber: The current state of alpaka before this PR:
* we have functors like `alpaka::op::And` that can be used as template argument to functions like `alpaka::atomicOp`.
* we have functors like `alpaka::block::op::And` that can be used as template argument to functions like `alpaka::block::syncBlockThreadsPredicate`

The state of alpaka after this PR:
* we have functors like `alpaka::op::And` that can be used as template argument to functions like `alpaka::atomicOp`.
* we have functors like `alpaka::blockOp::And` that can be used as template argument to functions like `alpaka::syncBlockThreadsPredicate`

The state of alpaka after this PR and PR #1185:
* we have functors like `alpaka::op::And` that can be used as template argument to functions like `alpaka::atomicOp`.
* additionally, we have functions like `alpaka::atomicAnd` that just call `alpaka::atomicOp<alpaka::op::And>`
* we have functors like `alpaka::blockOp::And` that can be used as template argument to functions like `alpaka::syncBlockThreadsPredicate`

The state of alpaka after this PR and PR #1185 and renaming the atomic functors:
* we have functors like `alpaka::AtomicAnd` that can be used as template argument to functions like `alpaka::atomicOp`.
* additionally, we have functions like `alpaka::atomicAnd` that just call `alpaka::atomicOp<alpaka::AtomicAnd>`
* we have functors like `alpaka::blockOp::And` that can be used as template argument to functions like `alpaka::syncBlockThreadsPredicate`

The state of alpaka after this PR and PR #1185 and renaming the atomic functors and also renaming the block sync functors:
* we have functors like `alpaka::AtomicAnd` that can be used as template argument to functions like `alpaka::atomicOp`.
* additionally, we have functions like `alpaka::atomicAnd` that just call `alpaka::atomicOp<alpaka::AtomicAnd>`
* we have functors like `alpaka::BlockAnd` that can be used as template argument to functions like `alpaka::syncBlockThreadsPredicate`

sbastrakov: Thanks for a detailed explanation. Turned out, I was still confused about some things and did not realize it. Nice!

bernhardmgruber: I created PR #1186 for the remaining changes.

PR #1187, opened 2020-10-29
Change block memory size back to be stored as 32 bit

sbastrakov:
The change to 64 bit was introduced as part of #1141.
It was later discovered that this can lead to performance detriment on GPUs, as discussed in https://github.com/alpaka-group/alpaka/commit/6b13898a8cece81b4222b739027f77bc3aab4b15#commitcomment-43688079
Revert the storage back to 32 bit, but now use a fixed-width type.

### DIFF
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -13,6 +13,7 @@
 #include <alpaka/core/Assert.hpp>
 #include <alpaka/core/Vectorize.hpp>
 
+#include <cstdint>
 #include <type_traits>
 #include <array>
 
@@ -32,7 +33,8 @@ namespace alpaka
         struct BlockSharedMemDynMemberStatic
         {
             //! Storage size in bytes
-            static constexpr std::size_t staticAllocBytes = TStaticAllocKiB<<10;
+            static constexpr std::uint32_t staticAllocBytes =
+                static_cast<std::uint32_t>(TStaticAllocKiB<<10u);
         };
     }
 
@@ -51,11 +53,10 @@ namespace alpaka
     public:
         //-----------------------------------------------------------------------------
         BlockSharedMemDynMember(std::size_t sizeBytes)
-            : m_dynPitch((sizeBytes / core::vectorization::defaultAlignment
-                    + (sizeBytes % core::vectorization::defaultAlignment > 0u)) * core::vectorization::defaultAlignment)
+            : m_dynPitch(getPitch(sizeBytes))
         {
 #if (defined ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) && (! defined NDEBUG)
-            ALPAKA_ASSERT(sizeBytes <= staticAllocBytes());
+            ALPAKA_ASSERT(static_cast<std::uint32_t>(sizeBytes) <= staticAllocBytes());
 #endif
         }
         //-----------------------------------------------------------------------------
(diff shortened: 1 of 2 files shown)


sbastrakov:
So my fix had a few shortcomings that resulted in failing builds. I've now pushed another commit, that tries to fix those. Since the `Acc` implementations take shared memory size as `size_t` in the constructor, and this is totally logical, I've decided to keep this part, and convert some of the implementation interface back, also to be `size_t`. And only made internal members 32 bit with conversions and only certain methods returning 32 bit. I will squash the commits later, after the issues are solved.

include/alpaka/kernel/Traits.hpp
```diff
@@ -120,15 +121,15 @@ namespace alpaka
...
+        return static_cast<std::size_t>(
             traits::BlockSharedMemDynSizeBytes<
                 TKernelFnObj,
                 TAcc>
             ::getBlockSharedMemDynSizeBytes(
                 kernelFnObj,
                 blockThreadExtent,
                 threadElemExtent,
-                args...);
+                args...));
```
sbastrakov: This seems redundant, was done before I came to the current state.

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
```diff
@@ -78,20 +79,29 @@ namespace alpaka
...
             */
-        std::size_t staticMemCapacity() const
+        std::uint32_t staticMemCapacity() const
         {
             return staticAllocBytes() - m_dynPitch;
         }
 
         //! Storage size in bytes
-        static constexpr std::size_t staticAllocBytes() {return detail::BlockSharedMemDynMemberStatic<TStaticAllocKiB>::staticAllocBytes;}
+        //! Returns a 32-bit type for register efficiency on GPUs
```
jkelling: Should be `\return` instead of prose "Returns" for Dxygen, even though it does not actually describe what the meaning of its return value is. Maybe it should do both, though.
```c++
//! \returns size of statically allocated memory available for both dynamic and static shared memory. Value is of a 32-bit type for register efficiency on GPUs

File: include/alpaka/block/shared/st/BlockSharedMemStMember.hpp [hunk 3]
```diff
@@ -29,7 +29,9 @@ namespace alpaka
         public:
             //-----------------------------------------------------------------------------
 #ifndef NDEBUG
-            BlockSharedMemStMemberImpl(uint8_t* mem, std::size_t capacity) : m_mem(mem), m_capacity(capacity)
+            BlockSharedMemStMemberImpl(uint8_t* mem, std::size_t capacity) :
```
jkelling, line 32: This call also happens on GPU and thus technically also wastes registers with the 64-bit value. However, this call happens at the beginning of the kernel in a non-thread-parallel section before the payload logic, where probably not many registers are used and thus will likely not increase the overall footprint of non-tiny kernels.

So, I can live with 64-bit here if it serves to simplify the interface.

jkelling (changes requested):

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
psychocoderHPC: I agree if it makes the interfaces simpler and avoids a lot of cast usage keep it 64bit. This part will not effect the register usage. 
It is only active in the debug build and it consumes only a temporary register at the beginning of the kernel.

sbastrakov: Okay, so this one I will keep as is.

include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
sbastrakov: Fixed and rebased.

include/alpaka/kernel/Traits.hpp
```diff
@@ -23,6 +23,7 @@
     #include <alpaka/workdiv/Traits.hpp>
 #endif
 
+#include <cstdint>
```
jkelling: This header should not be required here: `std::size_t` is not defined in there. Is it left over from an intermediate state with `uni32_t` in the interface?

jkelling (changes requested):

include/alpaka/kernel/Traits.hpp
sbastrakov: You are right, fixed.

jkelling (approved):

jkelling:
The failing checks are CI hickups. @psychocoderHPC can you please merge?

psychocoderHPC (approved):

PR #1196, opened 2020-11-09
fix bug in `isPinned()` and `pin()`

psychocoderHPC:
With the renaming of the namespaces in #1173 some calls to the traits
are broken.
Instead of calling the trait the local member function is called again.


psychocoderHPC:
PR can be merged, Issue is in the CI

sbastrakov (approved):

PR #1197, opened 2020-11-09
rename shared memory functions

psychocoderHPC:
Since we refactored in the last pull requests the namespaces block function to allocate shared memory are named strange because the hierarchy information provided by the namespaces get lost.
I renamed the functions/traits based on the naming we used for block synchronization functions e.g. `syncBlockThreadsPredicate`

- rename `freeMem` to `freeStaticBlockSharedMem`
- rename `allocVar` to `allocStaticBlockSharedVar`
- rename `getMem` to `getDynBlockSharedMem`
- rename coresponding structs to implement the traits
- update documentation

IMO this example is showing why I prefere namespaces to group objects within a hierarchy and build hierarchies via namespeces instead of mangle such information into names.


psychocoderHPC:
Note: I do not use the abbreviation of `Static` in the names because `St` is to less to expand it to the word `static`

bernhardmgruber:
While I agree that the current names are not good, I think the proposed names are too long and complicated.

AFAIK `freeStaticBlockSharedMem` is never called by the user, so I do not care too much about its naming.

As for `allocStaticBlockSharedVar` and `getDynBlockSharedMem`: these functions allocate shared memory to a kernel. AFAIK there are no other scopes for these allocated variables then the block level. So I would omit this information from the function name. We could mention this in the documentation though. Furthermore, shared memory is a concept known to many GPU programming people anyway so i do not see a large benefit of reminding them that shared memory works on the block level.

The distinction between allocating a variable and a piece of memory feels unnecessary to me.

So that leaves us with:
* `allocStaticShared<T, __COUNTER__>()`
* `allocDynamicShared<T>(size)`

And I would argue now that these calls could be distinguished by overload resolution:

* `allocShared<T, __COUNTER__>()`
* `allocShared<T>(size)`

However, I can see a benefit of making the difference more clear, since that has runtime implications.

psychocoderHPC:
I do not like to remove the word `Block` because we have `syncBlockThreadsPredicate`, `BlockAnd`, ....
I like to be consistent in the naming.

`allocShared<T>(size)` is not the right naming because this is no allocation, this function returns a already allocated memory region so with your naming schema it should be `getShared<T>(size)`.

IMO the names are too short because there is also inter-process shared memory which is currently not covered by alpaka but I like to avoid confision.

bernhardmgruber:
> I do not like to remove the word `Block` because we have `syncBlockThreadsPredicate`, `BlockAnd`, ....
> I like to be consistent in the naming.

OK. So if we say this variable or memory lives in block scope, then there is no need anymore to say it is shared ;)

> `allocShared<T>(size)` is not the right naming because this is no allocation, this function returns a already allocated memory region so with your naming schema it should be `getShared<T>(size)`.

Fair point. So how about a `getBlockVar<T, __COUNTER__>()` and an `allocBlockMem<T>(size)`.

> IMO the names are too short because there is also inter-process shared memory which is currently not covered by alpaka but I like to avoid confision.

That sounds scary! One more reason to not mention `shared` in the functions then!

psychocoderHPC:
@sbastrakov mentioned in the offline discussion that `shared memory` is part of the alpaka model. I am fine with `BlockMem` but then it will not reflect the alpaka model anymore :-(

The corresponding function to ask for the dynamic shared memory of a kernel is named `getBlockSharedMemDynSizeBytes` thus the current naming in this PR is IMO fitting best to the interfaces we have in alpaka.

psychocoderHPC:
@BenjaminW3 @sbastrakov @j-stephan What is your opinion?

j-stephan:
> mentioned in the offline discussion that shared memory is part of the alpaka model. I am fine with BlockMem but then it will not reflect the alpaka model anymore :-(

Perhaps we should talk about removing shared memory from alpaka's model then. IMO this is a relic from CUDA that doesn't make sense in its current form for either CPUs (because they don't have shared memory at all) or FPGAs (because they have different types of shared memory).

sbastrakov:
Well, my original point was just that the naming used in API should match the naming used by the current alpaka model, at least for "keywords" like blocks, threads, shared memory, etc.

If we are speaking in general about alpaka's shared memory. I agree it's awkward anywhere but on GPUs, however it is essential on GPUs.

psychocoderHPC:
Let us find a solution on the next alpaka meeting. I like the idea to change the name of the shared memory (I said the same already to @sbastrakov) but keep in mind that this mean we need to update the full documentation and all images.

BenjaminW3 (dismissed):

psychocoderHPC:
`getDynSharedMem` ,  `declareSharedVar` and `freeSharedVars`

If possible we will drop `freeSharedVars`

psychocoderHPC:
PR is now ready, `freeSharedVars` can not be removed, so I kept it as it is.

bernhardmgruber (dismissed):

psychocoderHPC:
I force pushed again to rebase and integrate the latest Xcode fix from the dev branch.

bernhardmgruber (dismissed):

include/alpaka/block/shared/dyn/Traits.hpp
```diff
@@ -27,20 +27,26 @@ namespace alpaka
...
         template<typename T, typename TBlockSharedMemDyn, typename TSfinae = void>
-        struct GetMem;
+        struct GetDynSharedMem;
     } // namespace traits
 
     //-----------------------------------------------------------------------------
-    //! Returns the pointr to the block shared dynamic memory.
+    //! Get block shared dynamic memory.
+    //!
+    //! The available size of the memory is can be defined by specializing the trait
```
j-stephan: Typo ("is can be")

include/alpaka/kernel/TaskKernelOmp5.hpp
```diff
@@ -205,7 +205,7 @@ namespace alpaka
                     }
 
                     // After a block has been processed, the shared memory has to be deleted.
-                    // freeMem(acc);
+                    // freeSharedVars(acc);
```
j-stephan: Why the comment? Can this be removed?

j-stephan (changes requested):
Looks good so far, only two minor comments.

include/alpaka/kernel/TaskKernelOmp5.hpp
jkelling: It is not required in this backend and thus could be removed. The trait `freeSharedVars` would do nothing here. The same is true for `TaskKernelCpuOmp2Blocks`, `TaskKernelCpuSerial`, `TaskKernelCpuTbbBlocks` and `TaskKernelOacc`.

jkelling:
> PR is now ready, `freeSharedVars` can not be removed, so I kept it as it is.

@psychocoderHPC Could you briefly summarize what it is that keeps us from removing this trait (moving any functionality into the `Acc`/`BlockShared*` destructor) at this time? We should keep what you found for later discussion.

As far as I am aware, all `TaskKernel*` only use their `acc` instances for a single block, making a call to free and having the dtor clean up equivalent. Is this trait use by some other code outside of this context?

psychocoderHPC:
> > PR is now ready, `freeSharedVars` can not be removed, so I kept it as it is.
> 
> @psychocoderHPC Could you briefly summarize what it is that keeps us from removing this trait (moving any functionality into the `Acc`/`BlockShared*` destructor) at this time? We should keep what you found for later discussion.
> 
> As far as I am aware, all `TaskKernel*` only use their `acc` instances for a single block, making a call to free and having the dtor clean up equivalent. Is this trait use by some other code outside of this context?

The shared memory operations are independent of kernels. The methods to allocate and free shared memory is called from `alpaka::kernel::*`. So I stopped looking into it because I can not assume any accelerator or implementation of shared memory there.

include/alpaka/block/shared/dyn/Traits.hpp
psychocoderHPC: corrected

BenjaminW3 (approved):

PR #1198, opened 2020-11-09
fix documentation

psychocoderHPC:
The documentation is not updated for all renaming we did during the
namespace refactoring.

### DIFF
--- a/docs/source/basic/cheatsheet.rst
+++ b/docs/source/basic/cheatsheet.rst
@@ -135,12 +135,12 @@ Allocate a buffer in device memory
 Enqueue a memory copy from host to device
   .. code-block:: c++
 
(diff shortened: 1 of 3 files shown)


bernhardmgruber:
Thx for spotting these!

But now I wonder: We have `alpaka::memcpy` now, but the `memset` equivalent is called `alpaka::set`? For sure I thought about renaming that one to `alpaka::memset` as well. Should we consider this rename?

psychocoderHPC:
> alpaka::set

should be `memset` my fault

psychocoderHPC:
renamed `set` into `memset`

BenjaminW3 (approved):

sbastrakov (approved):

PR #1192, opened 2020-11-04
fix travis_retry

psychocoderHPC:
`travis_retry` script function was not working because `pipefail` always
stoped the shell if a command fails.

- add missing color definitions, required else an error fired
- fix command execution to allow error handling

PR #1200, opened 2020-11-09
fix CI: use `travis_retry` for `git clone`

psychocoderHPC:
follow up of #1192

Guard `git clone` by `travis_retry` to avoid that the CI fails during cloning.
The folder into which we are cloning must be removed to avoid cloning into an existing folder.


BenjaminW3 (approved):

psychocoderHPC:
The current CI error is something with xcode and unrelated to this PR.

sbastrakov (approved):

PR #1206, opened 2020-11-11
CI: update used Xcode versions

psychocoderHPC:
- remove unavailable tests for Xcode 11.1 (reason for current CI
crashes)
- update Xcode versions we are testing to 11.2.1 - 12.2.0
- update README.md
- remove `-Wpoison-system-directories` for macOSX if CI is used 
  - Xcode 12.X compile error: `error: include location '/usr/local/include' is unsafe for cross-compilation` introduced by a `-isystem /usr/local/include` during compile (since nobody of us has a MAC and it is not our primary target I disabled it for all Xcode versions in the CI)
- always show all available Xcode versions in the OSX CI


BenjaminW3 (dismissed):

psychocoderHPC:
looks like this option requires Xcode 12. I will check which clang version it is and disable the warning only for the coresponding clang.

j-stephan:
Looks like all the tests passed successfully. Is this ready to merge or are you planning additional work?

script/install.sh
```diff
@@ -53,6 +53,9 @@ then
     if [ "${CXX}" == "clang++" ] ;then source ./script/install_clang.sh ;fi
 elif [ "$ALPAKA_CI_OS_NAME" = "macOS" ]
 then
+    echo "### list all applications ###"
+    ls "/Applications/"
+    echo "### end list all applications ###"
```
bernhardmgruber, line 58: Is this some kind of debug output? Do you still need it? Can it be removed?

psychocoderHPC: yes please keep it else we need to fire dummy CI tasks to get this information

j-stephan (approved):

bernhardmgruber (approved):

PR #1127, opened 2020-08-19
OpenACC backend

jkelling:
~*Builds on https://github.com/alpaka-group/alpaka/pull/1126  and will be rebased accordingly.*~

This PR adds an OpenACC offloading backend.

## CMake

* Adds cmake option `ALPAKA_ACC_ANY_BT_OACC_ENABLE`.
  * Compiler flags enabling OpenMP and OpenACC can be mutually exclusive, thus this backend cannot be enabled at the same time as any of the OpenMP backends. A CMake warning for this case is in place.

## CI

* Adds CI job linux_gcc-10_release-oacc which only enables this backend and `ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE`
* All tests with multiple compilation units are disabled for OpenACC+GCC because of a bug where GCC messes up symbols of target regions in some way.
* The test math is disabled for OpenACC+GCC because GCC does not allow types with `static constexpr` members to be mapped.
* In terms of compiler support: Most examples and tests can be compiled for tesla devices using NVHPC 20.7 (others trigger ICEs). Most compilable tests also succeed on GPU.

## Backend Properties

* Adds `AccOacc`, analogous to `AccOmp5`
* Adds `DevOacc` to handle OpenACC device ids, type and `acc_set_device_num`.
* Adds `BufOacc` for OpenACC device allocations.
* Uses `QueueGenericThreads*<DevOacc>`.
* ~Adds `CtxThreadOacc` to represent the context of a thread/worker.~ A per-thread instance [of `AccOacc`] is required because there is no intrinsic way in OpenACC to get a worker id and thus a loop counter ist used.
* Adds `CtxBlockOacc` to represent block-shared state, like shared memory and infrastructure for synchronization.

~While `AccOacc` is a handle for the backend on the host side, it doe not implement all concepts other `Acc*` type do. Concepts for traits which can only be used in device code, such as `getWorkDiv` or `getIdx` are only implemented by `CtxThreadOacc` which derives from `AccOacc`.~ Each instance of ~`CtxThreadOacc`~`AccOacc` holds a pointer to a block-shared instance of `CtxBlockOacc`.

## Quirks

* OpenACC does not support intra-block(gang) synchronization. `SyncBlockThreads<>::syncBlockThreads` had to be implemented using atomics: https://github.com/jkelling/alpaka/blob/openacc/include/alpaka/ctx/block/CtxBlockOacc.hpp#L114
[...]

PR #1209, opened 2020-11-16
README.md: Add OpenACC, OpenMP 5.0 GPU

jkelling:
Suggested by @BenjaminW3 in https://github.com/alpaka-group/alpaka/issues/1127#issuecomment-727188487_

Note:
1. The formate is a bit different fro OpenMP 5 and OpenACC: They take two lines because the can be compiled for CPU or GPU, but this is nothing Alpaka can know or control at compile time.
2. The version column for `OpenMP 5.0+`  states `OpenMP 4.0+`. There should be footnote here, but I do not know how to place it. The text would be:
"
The OpenMP 5 backend may work with an OpenMP 4.0+ compiler, depending on that compiler interpretation of the standard, only OpenMP 5.0+ clearly fullfills all requirements.
"


BenjaminW3 (dismissed):

jkelling:
Changed the required version for OpenMP 5.0 to 5.0.

psychocoderHPC (approved):

PR #1212, opened 2020-11-19
Add CONTRIBUTING.md with formatting instructions

bernhardmgruber:
This is (almost) the same file as in LLAMA and mallocMC.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,20 @@
+# Contributing
+
+## Formatting
+
+Please format your code before before opening pull requests using clang-format 11 and the .clang-format file placed in the repository root.
+
+### Visual Studio and CLion
+Suport for clang-format is built-in since Visual Studio 2017 15.7 and CLion 2019.1.
+The .clang-format file in the repository will be automatically detected and formatting is done as you type, or triggered when pressing the format hotkey.
+
+### Bash
+First install clang-format-11. Instructions therefore can be found on the web.
+To format your changes since branching off develop, you can run this command in bash:
+```
+git clang-format-11 develop
+```
+To format all code in your working copy, you can run this command in bash:
+```
+find -iname *.cpp -o -iname *.hpp | xargs clang-format-11 -i
+```


bernhardmgruber:
@psychocoderHPC mentioned somewhere else, that we should spell out the specific version of clang-format we use. I am afraid that this might cause quite an amount of maintenance every time we upgrade to a new version of clang-format, so I refrained from that for now.

sbastrakov (dismissed):

BenjaminW3:
I second @psychocoderHPC`s opinion that we should mention the specific version somewhere.
clang-format versions are incompatible (new options, removed options, same options format differently).
All commercial projects I know pin the version of the formatting tool in use.

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Bash
+First install clang-format. Instructions therefore can be found on the web.
+To format your changes since branching off develop, you can run this command in bash:
+```
+git clang-format develop
+```
+To format all code in your working copy, you can run this command in bash:
+```
+find -iname *.cpp -o -iname *.hpp | xargs clang-format -i
```
psychocoderHPC: Could you please write `clang-format-11`

psychocoderHPC (dismissed):

psychocoderHPC:
@bernhardmgruber could you somewhere also add a phrase that clang format 11 is mandentory.

bernhardmgruber:
Very well then, I added the clang-format version.

BenjaminW3 (approved):

CONTRIBUTING.md
```diff
@@ -0,0 +1,20 @@
...
+
+### Bash
+First install clang-format-11. Instructions therefore can be found on the web.
+To format your changes since branching off develop, you can run this command in bash:
+```
+git clang-format-11 develop
+```
+To format all code in your working copy, you can run this command in bash:
+```
+find -iname *.cpp -o -iname *.hpp | xargs clang-format-11 -i
```
psychocoderHPC, line 19: This command is wrong, it must be `find -iname "*.cpp" -o -iname "*.hpp" | xargs clang-format-11 -i`

psychocoderHPC: `*` is in the bash evaluated but this should define only a pattern for a file.

bernhardmgruber: Funnily enough, the command works just fine in my bash.

psychocoderHPC: Ohh let's check if this issue is coming from my `zsh`

psychocoderHPC: Mhh it is the zshell which do not like `*` only.

j-stephan: Yes, `zsh` needs either the quotes or a backslash. I prefer the quotes because it is portable and looks okay.

bernhardmgruber: Very well then, you have my permission to fix that then :)

psychocoderHPC: I can do it tomorrow.

PR #1204, opened 2020-11-10
[clang-format] Add .clang-format file

j-stephan:
This is the .clang-format file used for #1194 and #1203.

Application: From the top-level alpaka directory, execute

```bash
find example include test -name '*.hpp' -o -name '*.cpp' | xargs clang-format -i --style=file
```
I'm not certain about the best place in the docs to include the command. Any suggestions?

PR #1214, opened 2020-11-19
TestAccs: Fix OMP5/OACC GCC<11 checks

### LINKED ISSUE #1211: clangFormat broke indetation of complex compile-time guards
Here:
https://github.com/alpaka-group/alpaka/blob/474412ec54b43e95bb2de7039b1b4f53a744b5f3/test/common/include/alpaka/test/acc/TestAccs.hpp#L89
and here:
https://github.com/alpaka-group/alpaka/blob/474412ec54b43e95bb2de7039b1b4f53a744b5f3/test/common/include/alpaka/test/acc/TestAccs.hpp#L109

jkelling:
* Fix OMP5/OACC GCC<11 
* alos fix check of ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST

Closes https://github.com/alpaka-group/alpaka/issues/1211

### DIFF
--- a/test/common/include/alpaka/test/acc/TestAccs.hpp
+++ b/test/common/include/alpaka/test/acc/TestAccs.hpp
@@ -83,15 +83,14 @@ namespace alpaka
 #if defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) && !defined(TEST_UNIT_KERNEL_KERNEL_STD_FUNCTION)                         \
     && !(                                                                                                             \
         BOOST_COMP_GNUC                                                                                               \
-        && (((BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(11, 0, 0))                                                       \
-                 && /* tests excluded because of GCC10 Oacc / Omp5 target symbol bug with multiple units */           \
-                 defined(TEST_UNIT_BLOCK_SHARED)                                                                      \
-             || defined(TEST_UNIT_BLOCK_SYNC) || defined(TEST_UNIT_WARP) || defined(TEST_UNIT_INTRINSIC)              \
-             || defined(TEST_UNIT_KERNEL) || defined(TEST_UNIT_MEM_VIEW))                                             \
+        && (((BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(11, 0, 0)) /* tests excluded because of GCC10 Oacc / Omp5 target \
+                                                                   symbol bug with multiple units */                  \
+             && (defined(TEST_UNIT_BLOCK_SHARED) || defined(TEST_UNIT_BLOCK_SYNC) || defined(TEST_UNIT_WARP)          \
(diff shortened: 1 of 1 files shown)


jkelling:
@j-stephan : as the clang-format file is not yet merged, is the workflow to apply it to this PR documented somewhere?

j-stephan:
Yes, the updated documentation file in #1204 shows how to use it.

test/common/include/alpaka/test/acc/TestAccs.hpp
```diff
@@ -83,15 +83,14 @@ namespace alpaka
 #if defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) && !defined(TEST_UNIT_KERNEL_KERNEL_STD_FUNCTION)                         \
     && !(                                                                                                             \
         BOOST_COMP_GNUC                                                                                               \
-        && (((BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(11, 0, 0))                                                       \
-                 && /* tests excluded because of GCC10 Oacc / Omp5 target symbol bug with multiple units */           \
-                 defined(TEST_UNIT_BLOCK_SHARED)                                                                      \
-             || defined(TEST_UNIT_BLOCK_SYNC) || defined(TEST_UNIT_WARP) || defined(TEST_UNIT_INTRINSIC)              \
-             || defined(TEST_UNIT_KERNEL) || defined(TEST_UNIT_MEM_VIEW))                                             \
+        && (((BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(11, 0, 0)) /* tests excluded because of GCC10 Oacc / Omp5 target \
```
jkelling, line 86: clang-format forces this longer inline comment to the end of the line an aligns it. This looks strange.

j-stephan: Maybe put it above the `&&`? Like so:

```c++
BOOST_COMP_GNUC
 /* tests excluded because of GCC10 Oacc / Omp5 target symbol bug with multiple units */
&& (((BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(11, 0, 0))
```

jkelling: It does not belong before the `&&`, because that first condition, which it refers to, is for the GCC 10 issue, there is a second condition which is for GCC in general, separate from the "multiple units"-bug referred to in the comment.

If put in the place where it is in the old version of the code, clang-format would evidently not move it, as this already went through clang format. I did not want to leave it after the following `&&`, because placing it there may have let to forgetting the parenthesis in the first place.

jkelling: Btw.: I think in your example the comment would be put after `BOOST_COMP_GNUC`, which would look less bad but would be more unsuitable.

jkelling: I can either
1. leave it as it is now (in which case this PR can be merged), or
2. return the comment to it former place (old state in diff).

How do people feel about this?

BenjaminW3 (approved):

PR #1218, opened 2020-11-23
clang-format tinymt32.h

bernhardmgruber:
This file was forgotton during the recent clang-format of the code base.

### DIFF
--- a/include/alpaka/rand/TinyMT/tinymt32.h
+++ b/include/alpaka/rand/TinyMT/tinymt32.h
@@ -6,6 +6,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
+// clang-format off
 #ifndef TINYMT32_H
 #define TINYMT32_H
 /**


BenjaminW3:
I am pretty sure that we explicitly did not want to format this file so that it is easier to apply patches from the original.
Could we add a .clang-format file that disables formatting for this folder instead?

bernhardmgruber:
Well, that sounds like this is a third party file then. Why is it not in the thirdParty folder, where formatting is disabled? I would move it there then.

BenjaminW3:
You have to ask @ax3l about this. We had to do some changes to this file. Maybe that is the reason.

bernhardmgruber:
@ax3l, why is tinymt32.h not in the thirdParty folder of alpaka?

Btw, in the same folder there is also Engine.hpp, so we cannot just place a .clang-format file in the same folder to disable formatting, Then we can also not format Engine.hpp.

bernhardmgruber:
@psychocoderHPC thinks it is not in thirdParty, because it is installed with the alpaka headers.

Remaining options: move it into a subdirectory and place a clang-format file there. Or even shorter: add a `// clang-format off` at the beginning of the file.

bernhardmgruber:
I revered the format and added a `// clang-format off` at the beginning of the file.

psychocoderHPC (approved):

PR #1219, opened 2020-11-23
Enable OpenMP backends for more Visual Studio builds

BenjaminW3:
`ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE` backend historically had been disabled
due to the compilation times but without travis there is no hard boundary
anymore that we could reach.

This also fixes the mention of `ALPAKA_ACC_CPU_BT_OMP4_ENABLE` by replacing
it with `ALPAKA_ACC_ANY_BT_OMP5_ENABLE`


.github/workflows/ci.yml
```diff
@@ -185,43 +185,43 @@ jobs:
...
+          env: {CXX: cl.exe,  CC: cl.exe, ALPAKA_CI_CL_VER: 2017,                                   CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.72.0, ALPAKA_CI_CMAKE_VER: 3.18.5,                     ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CUDA_VERSION: "11.1", ALPAKA_CUDA_ARCH: "75;80", ALPAKA_ACC_GPU_CUDA_ONLY_MODE: ON,                                                                ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE: OFF, ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: OFF, ALPAKA_ACC_ANY_BT_OMP5_ENABLE: OFF}
```
psychocoderHPC: Please keep `35` that we have a check of the minimal supported ISA set.

psychocoderHPC (changes requested):

.github/workflows/ci.yml
BenjaminW3: Fixed. This slipped through from the CMake first-class CUDA branch where I split this off. There the architecture selection seems to be broken within CMake.

BenjaminW3:
Ready to be reviewed/merged!

psychocoderHPC (approved):

Issue #1205, opened 2020-11-11
bad performance with OpenMP


psychocoderHPC:
DLR reported via mail that they see a performance decrease when the OpenMP backend (block parellel) is used and the load per block is very different.

Following this discussion: https://stackoverflow.com/questions/42970700/openmp-dynamic-vs-guided-scheduling
we should think about changeing the OpenMP scheduling strategy from current used [guided](https://github.com/alpaka-group/alpaka/blob/7ba4c30a4947b97169b3216436ca23e2db2f5935/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp#L201) to `schedule(auto)` or `schedule(runtime)` and select a schedular based on the grid size.

sbastrakov:
I was also concerned about the `guided` schedule in the past. And we actually checked it a few month ago using @kloppstock 's code, which did not show the difference. However, of course it does not mean the difference doesn't exist for any code.

Can we maybe think of adding a cmake option that controls it, and by default use something reasonable?

j-stephan:
What do you guys think of some scheduling hints in the alpaka API? I'm thinking of something like

```c++
class Kernel
{
public:
    using SchedulerHint = alpaka::ImbalancedHint;
};
```

And if the kernel class doesn't have a hint assume the current way by default.

sbastrakov:
Oh yes, that can also be an option. And I actually like it, if we do it as a trait. Because then by default it can use our default option, and a user (kernel developer) does not have to, but can specialize it.

(we can of course also check if a kenel class has internally defined type with a fixed name like `SchedulerHint` and wrap it into a trait, however i feel just plain trait would be more alpaka way)

psychocoderHPC:
solved by #1223

PR #1223, opened 2020-11-27
Add an option to set OpenMP schedule for the Omp2Blocks backend

sbastrakov:
After looking at implementations, we could only really set a schedule to the `Omp2Blocks` accelerator, as others using OpenMP (in various ways) set number of threads equal to number of tasks and so do not employ a schedule.

I've so far checked this with manually modifying one of the examples, and both internal member and trait specialization work.

TODOs:

- [x] add unit tests for the schedule and `meta::Void`, maybe separate void in another PR
- [x] ~~add docs~~ we do not have docs for OpenMP backends at all (and the ones existing are really outdated), so I added the usage illustration and explanation to the new example, and somewhat to tests
- [x] maybe add an example
- [x] fix commit message (all text editors ~~are broken on the dev machine~~ turned out not working when llvm was spack-loaded for clang-format)

However, the implementation itself is ready for review.

Solves #1205. Needs to go to release 0.6.0


include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+            enum Kind
+            {
+                Static = 1,
+                Dynamic = 2,
+                Guided = 3,
+                Auto = 4
+            };
+
+            //! Integer representation of schedule kind.
+            //!
```
sbastrakov, line 41: I'm now confused. Isn't clang-format taking care of that?

include/alpaka/kernel/Traits.hpp
```diff
@@ -83,6 +85,84 @@ namespace alpaka
...
+            //! Functor to get OpenMP schedule for kernel classes with
+            //! ompSchedule static member.
+            //! That member is never odr-used by alpaka.
+            template<class TKernel>
+            struct GetOmpSchedule<TKernel, meta::Void<decltype(TKernel::ompSchedule)>>
+            {
+                ALPAKA_FN_HOST_ACC auto operator()() const
+                {
+                    return TKernel::ompSchedule;
+                }
```
sbastrakov: Note: here have to go with a functor, as we can't rely that a user-defined member is `constexpr`. And for non-constexpr can't initialize it in place, as it is not a literal type.

include/alpaka/core/OmpSchedule.hpp
j-stephan: Could you clarify a bit? What should clang-format do here?

j-stephan:
Thanks! I'll do the review on Monday.

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+            //! Not stored as Kind to reduce casts when converting to and from omp_sched_t
+            uint32_t kind;
+
+            //! Chink size.
+            //!
+            //! Same as in OpenMP, value 0 corresponds to default chunk size
+            int chunkSize;
+
+            constexpr Schedule(Kind kind = Guided, int chunkSize = 0)
+                : kind(static_cast<uint32_t>(kind))
```
psychocoderHPC: There are few errors about name shadowing errors: https://github.com/alpaka-group/alpaka/pull/1223/checks?check_run_id=1464801629#step:5:4290 with `kind` and `chunkSize`.
In PIConGU we use `m_varname` in such case for class member and the human readable name `varname` for method parameter to expose nice names in the userinterface.

psychocoderHPC (changes requested):

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -156,15 +169,17 @@ namespace alpaka
                 *static_cast<WorkDivMembers<TDim, TIdx> const*>(this),
                 blockSharedMemDynSizeBytes);
 
-            // NOTE: schedule(static) does not improve performance.
+            auto oldSchedule = omp::getSchedule();
+            omp::setSchedule(schedule);
```
psychocoderHPC: Is this function and `omp::setSchedule` defined for omp < 3.0?

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+        {
+            //#############################################################################
+            //! Mirror of the OpenMP enum omp_sched_t with the corresponding integer values
+            enum Kind
+            {
+                Static = 1,
+                Dynamic = 2,
+                Guided = 3,
+                Auto = 4
+            };
```
bernhardmgruber: Why is this not an enum class?

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+        {
+            // Getting a runtime schedule requires OpenMP 3.0 or newer
+#if defined _OPENMP && _OPENMP >= 200805
+            auto result = Schedule{};
+            omp_sched_t kind;
+            omp_get_schedule(&kind, &result.chunkSize);
+            result.kind = static_cast<uint32_t>(kind);
+            return result;
+#else
+            return Schedule{};
```
bernhardmgruber, line 79: This creates a `Schedule` with `kind` and `chunkSize` 0. What does this mean? A kind of `0` does not exist in the enum.

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+            //!
+            //! Not stored as Kind to reduce casts when converting to and from omp_sched_t
+            uint32_t kind;
+
+            //! Chink size.
+            //!
+            //! Same as in OpenMP, value 0 corresponds to default chunk size
+            int chunkSize;
+
+            constexpr Schedule(Kind kind = Guided, int chunkSize = 0)
```
bernhardmgruber: Why do we need a ctor? `Schedule` is just an aggregate.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -181,6 +196,8 @@ namespace alpaka
                 // After a block has been processed, the shared memory has to be deleted.
                 freeSharedVars(acc);
             }
+            // Restore the old schedule to not interfere with the rest
+            omp::setSchedule(oldSchedule);
```
bernhardmgruber: This should be called by a scope guard to guarantee that this function is called in case of exceptions. (like `freeSharedVars` throws)

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -92,6 +93,17 @@ namespace alpaka
...
+            // Get the OpenMP schedule.
+            auto const schedule(meta::apply(
+                [&](ALPAKA_DECAY_T(TArgs) const&... args) {
+                    return getOmpSchedule<AccCpuOmp2Blocks<TDim, TIdx>>(
+                        m_kernelFnObj,
+                        blockThreadExtent,
+                        threadElemExtent,
+                        args...);
+                },
+                m_args));
```
bernhardmgruber: I do not get why we need to pass so many arguments around. All these extents and args are ignored later down the call chain anyway. You could just replace this by: `const auto schedule = GetOmpSchedule<decltype(m_kernelFnObj)>{}();`

What am I missing here?

bernhardmgruber (changes requested):

include/alpaka/core/OmpSchedule.hpp
BenjaminW3: Maybe for the default parameters?

bernhardmgruber: Well, then just directly initialize the members:
```c++
struct Schedule {
    Kind kind = Kind::Guided;
    int chunkSize = 0;
};
```

include/alpaka/core/OmpSchedule.hpp
sbastrakov: I don't think it actually shadows anything with empty constructor body, but will fix. Here addintg `m_` is not very nice, as this is POD, and so public members are part of interface.

include/alpaka/core/OmpSchedule.hpp
sbastrakov: So that one can write `Schedule::Static` at construction (in a user code), and not `Schedule::Kind::Static`

include/alpaka/core/OmpSchedule.hpp
sbastrakov: I started with your suggestion, But it felt like too many casts when this struct is used  (as then there is a double cast between this enum and an OpenMP one via int in the same expression), and so I went for storng internally as int and converting in a constructor. This is the only purpose of it.

include/alpaka/core/OmpSchedule.hpp
sbastrakov: Default `kind` is not 0, it's from the default constructor. Which is `Guided` to match the existing alpaka behavior.
`chunkSize == 0` is OpenMP special value to use internal default (for given schedule), there is a comment at line 48 about it.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: Good point, will fix

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: They are for a user (developer of a kernel) if they specialize the trait and not just provide a static member. E.g. if they want to distinguish between large and small jobs, or smth else. Maybe they even have an internal parameter that defines schedule and so could translate it to alpaka there.

Of course, for alpaka's default logic of just setting some schedule depending on the kernel type only these parameters are not used

sbastrakov: I think from the existing code and commeents it's just not clear yet that there are two ways to provide a schedule. We need to evaluate if it's clear in the finished version.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: That is my function. And it was introduced exactly for the reason to hide checks like if we have OpenMP and OpenMP is at least 3.0

include/alpaka/core/OmpSchedule.hpp
sbastrakov: I was not sure how to format this comment, i.e. whether to separate comment "head" with an empty line. And thought it would be auto-formatted. But I guess that was just stupid thinking on my side

include/alpaka/core/OmpSchedule.hpp
bernhardmgruber: You are right! That was me thinking in aggregates! So `Schedule{}` is fine then!

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+                Guided = 3,
+                Auto = 4
+            };
+
+            //! Integer representation of schedule kind.
+            //!
+            //! Not stored as Kind to reduce casts when converting to and from omp_sched_t
+            uint32_t kind;
+
+            //! Chink size.
```
j-stephan: Typo.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -156,15 +169,17 @@ namespace alpaka
...
                 blockSharedMemDynSizeBytes);
 
-            // NOTE: schedule(static) does not improve performance.
+            auto oldSchedule = omp::getSchedule();
+            omp::setSchedule(schedule);
 #    if _OPENMP < 200805 // For OpenMP < 3.0 you have to declare the loop index (a signed integer) outside of the loop
                          // header.
             std::intmax_t iNumBlocksInGrid(static_cast<std::intmax_t>(numBlocksInGrid));
             std::intmax_t i;
+            // OpenMP 2.0 does not support schedule(runtime), stick to guided
```
j-stephan: Are you sure? The OpenMP 2.0 specification does mention `schedule(runtime)`. See https://www.openmp.org/wp-content/uploads/cspec20.pdf, page 13. Is this maybe confused with `schedule(auto)`?

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+        //-----------------------------------------------------------------------------
+        //! Get the OpenMP schedule that is applied when the runtime schedule is used.
+        //!
+        //! When executed without OpenMP or with OpenMP < 3.0, returns default
+        //! schedule.
+        //!
+        //! \return Schedule object.
+        ALPAKA_FN_HOST inline auto getSchedule()
+        {
+            // Getting a runtime schedule requires OpenMP 3.0 or newer
```
j-stephan, line 72: See the `runtime` <-> `auto` comment in `TaskKernelCpuOmp2Blocks.hpp`.

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -0,0 +1,94 @@
...
+            omp_get_schedule(&kind, &result.chunkSize);
+            result.kind = static_cast<uint32_t>(kind);
+            return result;
+#else
+            return Schedule{};
+#endif
+        }
+
+        //-----------------------------------------------------------------------------
+        //! Set the OpenMP schedule that is applied when the runtime schedule is used.
```
j-stephan: See the `runtime` <-> `auto` comment in `TaskKernelCpuOmp2Blocks.hpp`.

include/alpaka/core/OmpSchedule.hpp
j-stephan: Will the cast really have a noticeable performance impact? I doubt that users will do high-frequency scheduler switching. Staying with `enum` will eliminate this issue anyway as soon as we transition to C++17 because then we can just write `enum (class) Kind : std::uint32_t`.

include/alpaka/core/OmpSchedule.hpp
j-stephan: Why not put the `enum class` outside of the struct? I don't think we can expect many more `kind`s in the `omp` namespace, the specification itself only mentions it with regard to scheduling.

j-stephan (changes requested):
Great, this looks really helpful!

include/alpaka/core/OmpSchedule.hpp
sbastrakov: No, performance does not matter here. Just a (subjective) matter of how clean the implementation looks

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: Thanks for clarifying. So my comment was imprecise. In 2.0 `schedule(runtime)` exists but only takes the value from an environment variable. So one can't set it at runtime with `omp_set_schedule` (and that requires 3.0) and this is what my comments should have been talking about. I will fix the comments.

include/alpaka/core/OmpSchedule.hpp
sbastrakov: Yes, I can put it outside as well. I don't think this makes all that much difference, I just liked most the version of writing (in a user code) smth like `... ompSchedule = alpaka::omp::Schedule{ alpaka::omp::Schedule::Dynamic };` . We could consider alternatives so that instead of `alpaka::omp::Schedule::Dynamic` there would be something else

sbastrakov: Okay, I've also realized we can't define our own mirror of `omp_sched_t` enum, since implementations of OpenMP are allowed to extend the standard values with their own and I don't think alpaka should disable this possibility by only allowing the standard values. I've now switched to representing as just ints and documented the decision in comments.

sbastrakov:
Thanks for the reviews. I've continued working on it today, and just pushed the current version. Also added unit tests and a new example.

However, on the `dev` machine it seems like the OpenMP runtime ignores the schedule I am setting (even though it does support these functions and does call them), and instead uses a schedule defined by the environment variable `OMP_SCHEDULE`. I am so far confused by how this happens, since this environment variable should have lower priority.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: From the example and unit tests it should hopefully be more clear now. Docs TBD

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -156,15 +170,38 @@ namespace alpaka
...
                          // header.
             std::intmax_t iNumBlocksInGrid(static_cast<std::intmax_t>(numBlocksInGrid));
             std::intmax_t i;
-#        pragma omp for nowait schedule(guided)
+#        pragma omp for nowait schedule(runtime)
             for(i = 0; i < iNumBlocksInGrid; ++i)
 #    else
-#        pragma omp for nowait schedule(guided)
+#        pragma omp for nowait schedule(runtime)
             for(TIdx i = 0; i < numBlocksInGrid; ++i)
```
sbastrakov, line 206: Note: with thie PR, the Omp2Blocks accelerator will always use `runtime`. For the implementations that only has 2.0 support (so for Microsoft) this means the default changes from guided to whatever the default for the environment variable. I think this is okay, since for 3.0 and higher the default will stay guided as it was.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
bernhardmgruber: Ok, I think I got it now! You want to give users the ability to specialize a trait so they have access to the kernel arguments for determining what kind of schedule to use. I think this is fine then! Thx!

sbastrakov:
Update: I've found the reason why my example was not working. I was accidentally setting the variable inside the parallel region, and I guess the its value may have been captured when the started. With moving it outside it now works as expected. I will finish this PR today.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2019 Benjamin Worpitz, Bert Wesarg, René Widera
+/* Copyright 2019-2020 Benjamin Worpitz, Bert Wesarg, René Widera,
+ *                     Sergei Bastrakov
```
sbastrakov: Followed the old mentality of 80 characters per line, should have continued on the previous line

include/alpaka/kernel/Traits.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2019 Axel Huebl, Benjamin Worpitz, René Widera
+/* Copyright 2019-2020 Axel Huebl, Benjamin Worpitz, René Widera,
+ *                     Sergei Bastrakov
```
sbastrakov: Same with 80 symbols here

sbastrakov:
Ready for review. (So many code lines for such a small feature)

So today after fighting with tests and example and thinking more about all it, I think I've come to a solution that is more clean than my original implementation in some details. Now the OpenMP2 blocks backend always uses `omp schedule(runtime)` regardless of the OpenMP version and if schedule is set via an alpaka abstraction.

By default, so if no schedule is specialized either way for a kernel, it will behave exactly like a pure OpenMP parallel for: 

- for OpenMP runtime < 3.0 it will use the environment variable `OMP_SCHEDULE` that a user could set externally, and OpenMP implementation sets the default value otherwise
-  for OpenMP runtime >= 3.0 it will take the internal control variable (ICV) `run-sched-var` which is by default equal to `OMP_SCHEDULE`, and can be set by a user externally via calling `omp_set_schedule()` on their side

So the default behavior should be completely transparent, in a way that alpaka does not change anything in how a user controls their OpenMP setup.

Then if the schedule is set by a user for a kernel:

- for OpenMP runtime < 3.0 this setting will have no effect, will still use `OMP_SCHEDULE`
- for OpenMP runtime >= 3.0 the set schedule will be used for this kernel, will restore the original state afterwards (including if the kernel throws)

It is possible to set the schedule for all kernels at once, as a kind of user application level setting. This is demonstrated in the example (also with some explanation) and in the unit tests. Then a user could still further partially specialize for a particular kernel.

example/openMPSchedule/src/openMPSchedule.cpp
```diff
@@ -0,0 +1,163 @@
...
+    {
+        //! Schedule trait specialization for our kernel.
+        //! When uncommented, that would have priority over the ompSchedule member.
+        //! This is the most general way to define a schedule, while that member is a shortcut for the common case.
+        //! In case neither the trait nor the member are provided, alpaka does not set any runtime schedule and the
+        //! schedule used is defined by omp_set_schedule() called on the user side, or otherwise by the OMP_SCHEDULE
+        //! environment variable.
+        /*
+        template<typename TAcc>
+        struct OmpSchedule<OpenMPScheduleKernel, TAcc>
```
psychocoderHPC: Could we provide this as a second example or the second file in this example. 
This code will be soon outdated and untested.

psychocoderHPC (changes requested):
Please also fix the code formatting. CI is failing because of the wrong formated code.

psychocoderHPC:
looks like we have some issues with OSX and in windows: 

OSX error: https://github.com/alpaka-group/alpaka/pull/1223/checks?check_run_id=1481511372#step:5:1768
Windows (runtime error) : https://github.com/alpaka-group/alpaka/pull/1223/checks?check_run_id=1481511720#step:4:3654

example/openMPSchedule/src/openMPSchedule.cpp
```diff
@@ -0,0 +1,163 @@
...
+    {
+        // For simplicity assume 1d index space throughout this example
+        using Idx = alpaka::Idx<TAcc>;
+        Idx const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+
+        // Print work distribution between threads for illustration
+        printf(
+            "alpaka global thread index %u is processed by OpenMP thread %d\n",
+            static_cast<std::uint32_t>(globalThreadIdx),
+            omp_get_thread_num());
```
psychocoderHPC, line 48: OSX bug: If OpenMP is not available e.g. if the OMP backend is not enabled then you can not call a native OpenMP function. 
This print should be guarded with `#if`

psychocoderHPC:
I do not understand why our Montecarlo test is failing in windows:

```
 test 5
      Start  5: monteCarloIntegration

5: Test command: D:\a\alpaka\alpaka\build\example\monteCarloIntegration\Release\monteCarloIntegration.exe
5: Test timeout computed to be: 1500
5: exact result (pi / 4): 0.785398
5: final result: 0.785369
5: error: 2.9542e-05
```
Your condition to pass the test is https://github.com/alpaka-group/alpaka/blob/1ffe300fd28d023473686af5a75d75701dbe301e/example/monteCarloIntegration/src/monteCarloIntegration.cpp#L149

` 2.9542e-05 > 0.001 ` should be false and return `EXIT_SUCCESS`.
Do I miss something?

psychocoderHPC:
The new OpenMP test is failing under Windoes: 
https://github.com/alpaka-group/alpaka/pull/1223/checks?check_run_id=1481511582#step:4:3908

example/openMPSchedule/src/openMPSchedule.cpp
sbastrakov: Ah, yes. I originally guarded the whole thing but then changed it. Will change back

example/openMPSchedule/src/openMPSchedule.cpp
sbastrakov: Good point, will change

psychocoderHPC:
@sbastrakov could you please have a look into this PR today?

sbastrakov:
Ugh, I made a stupid mistake with guarding in the example again. Will fix tomorrow.

psychocoderHPC:
Looks like windows has still some issues: https://github.com/alpaka-group/alpaka/pull/1223/checks?check_run_id=1524473770#step:4:3074

And OpenAcc is missing the main function: https://github.com/alpaka-group/alpaka/pull/1223/checks?check_run_id=1524474259#step:5:3640

sbastrakov:
@psychocoderHPC yes, that was due to wrong guarding. I've just pushed a fix.

sbastrakov:
It took long, but finally all CI issues are resolved. Ready for re-review @psychocoderHPC @j-stephan (and of course others if you are interested)

include/alpaka/kernel/Traits.hpp
```diff
@@ -83,6 +85,88 @@ namespace alpaka
...
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored                                                                                  \
+        "-Wdocumentation" // clang does not support the syntax for variadic template arguments "args,..."
+#endif
+            //-----------------------------------------------------------------------------
+            //! \param kernelFnObj The kernel object for which the schedule should be returned.
+            //! \param blockThreadExtent The block thread extent.
+            //! \param threadElemExtent The thread element extent.
+            //! \tparam TArgs The kernel invocation argument types pack.
+            //! \param args,... The kernel invocation arguments.
```
psychocoderHPC, line 149: is it not possible to write only `args`. The macros around the comments make it hard to read the file

psychocoderHPC (approved):

include/alpaka/kernel/Traits.hpp
sbastrakov: I agree. This pattern is also present in [other places](https://github.com/alpaka-group/alpaka/blob/a8e26171e86b65754e66f5f3b9f9ef0ac23729d6/include/alpaka/kernel/Traits.hpp#L54-L69) where I copied it from. I will make another PR that implements your suggestion. Not only it makes the code more readable, it is also more technically accurate, as `args` here is all arguments already and `args...` is just pack extension

psychocoderHPC:
@j-stephan Could you please check it again, if all is fine I would merge it.

j-stephan (approved):
Looks good to me!

PR #1215, opened 2020-11-19
Update to Catch 2.13.3

j-stephan:
Our last update happened last year. Since then there have been multiple fixes for MSVC (which is currently troubling #1203). This draft PR is a test to see if it helps our test suite in this regard.

PR #1228, opened 2020-12-07
Make test/common a CMake INTERFACE library

j-stephan:
This PR contains two commits:

1. Bump the Catch2 version required by CMake to 2.13.3 (this was forgotten in #1215).
2. Make `test/common` an `INTERFACE` library. `test/common/src/Dummy.cpp` stated that CMake doesn't support header-only targets, but this is no longer true.

### DIFF
--- a/test/CMakeLists.txt
+++ b/test/CMakeLists.txt
@@ -20,8 +20,8 @@ endif()
 add_subdirectory(common)
 
 if(ALPAKA_ACC_GPU_CUDA_ENABLE AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
-    # NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
-    get_target_property(_COMMON_COMPILE_OPTIONS common COMPILE_OPTIONS)
+    # NVCC does not incorporate the INTERFACE_COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
+    get_target_property(_COMMON_COMPILE_OPTIONS common INTERFACE_COMPILE_OPTIONS)
     # If the property does not exist, the variable is set to NOTFOUND.
     if(_COMMON_COMPILE_OPTIONS)
         string(REPLACE ";" " " _COMMON_COMPILE_OPTIONS_STRING "${_COMMON_COMPILE_OPTIONS}")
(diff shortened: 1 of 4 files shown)


bernhardmgruber:
Some Visual Studio developers will now be very unhappy to see the `test/common` library gone from their Solution Explorer ;)

That is btw why we also have the `alpakaIde` target, so it shows up in Visual Studio as a project. It is a different workaround to achieve the same: making the files appear in Visual Studio.

psychocoderHPC (approved):

psychocoderHPC (dismissed):

psychocoderHPC:
There are a few issues in the CI.

j-stephan:
@bernhardmgruber Thanks for the hint, I keep forgetting about the peculiarities of IDEs. At least there is hope: We can drop the dummy targets once we have fully transitioned to CMake >= 3.19 (so in 1 or 2 years): `INTERFACE` targets can now have `PUBLIC` and `PRIVATE` sources, which should make them appear as regular projects in Visual Studio and XCode. For now I'll add an dummy IDE target.

@psychocoderHPC CMake chokes on the `FOLDER` property for `INTERFACE` targets for all CMake versions < 3.19. Looks like an undocumented change in 3.19 to me because I can't find anything in the release notes. I think this should also be solved by the IDE target (since that will have all the source code assigned to), so I'll remove this property from the actual target.

sbastrakov:
@bernhardmgruber is right about Visual Studio and I also encountered that. It is definitely unpleasant, and for Windows-only users may come as a surprise, but honestly it is not that big of an issue in my opinion, as long as the standard settings provided by cmake fit. It becomes a pain when settings need to be changed manually, like I have to do for PIConGPU, but luckily not for alpaka.

sbastrakov:
But I think it is better if we are consistent, so if alpaka decides to provide these targets for IDE to do it for all meaningful targets

bernhardmgruber:
> We can drop the dummy targets once we have fully transitioned to CMake >= 3.19 (so in 1 or 2 years): `INTERFACE` targets can now have `PUBLIC` and `PRIVATE` sources, which should make them appear as regular projects in Visual Studio and XCode.

That is great news! I hated these dummy cpp files for the `xxxIde` targets just to see my files in VS' Solution Explorer.

Btw, VS also has the Open Folder mode, where you just view the filesystem and VS just relies on the CMake files directly for building. There you also do not need any workarounds. However, in this mode not all VS builtin tools are suppored, like background code analysis etc. Also: The Open Folder mode is the same as CLion does, which works far better in this regard.

sbastrakov:
Agree @bernhardmgruber , whenever I tried opening a folder with VS something did not work properly. But that was some time ago, so hopefuly they are improving.

bernhardmgruber (approved):

j-stephan:
This can be merged by the way...

sbastrakov (approved):

PR #1231, opened 2020-12-17
Add CI runs for CUDA 11.2

bernhardmgruber:


psychocoderHPC:
We have not updated warning suppressions for CUDA 11
https://github.com/alpaka-group/alpaka/blob/ec56e2e3f51dbe394e6a394a2a9b952f3c2bf066/cmake/alpakaCommon.cmake#L499-L505
and should also provide correct compiler supports for CUDA 11: https://github.com/alpaka-group/alpaka/blob/ec56e2e3f51dbe394e6a394a2a9b952f3c2bf066/cmake/alpakaCommon.cmake#L377-L423

psychocoderHPC:
We can derive most of the information from https://gist.github.com/ax3l/9489132

bernhardmgruber:
> We have not updated warning suppressions for CUDA 11
> https://github.com/alpaka-group/alpaka/blob/ec56e2e3f51dbe394e6a394a2a9b952f3c2bf066/cmake/alpakaCommon.cmake#L499-L505

And just checked one of the CUDA 11.2 runs and there are no warnings generated. So why suppress anything?
See e.g. https://github.com/alpaka-group/alpaka/pull/1231/checks?check_run_id=1570822602

> and should also provide correct compiler supports for CUDA 11:
> https://github.com/alpaka-group/alpaka/blob/ec56e2e3f51dbe394e6a394a2a9b952f3c2bf066/cmake/alpakaCommon.cmake#L377-L423

Why do we even provide this information at all? Some CUDA headers have this check anyway, so you get an error during compilation. Is it really worth the effort to maintain this list in our cmake as well?

bernhardmgruber:
I added your diagnostics for unsupported compilers. But I still believe we should not maintain this list and just rely on nvcc erroring.

sbastrakov:
I think such checks are useful and worth extra manual work because

- it is unfortunately not common knowledge and may be not easy to find as a centralized info (like in Axel's gist); definitely not reasonable to expect from a user to know it by heart
- we need this info ourselves to know what alpaka supports and what it doesn't, since it is by nature having lots of possible combinations
- for something we know it does not work it is better to give a hard error, than for it to work e.g. for user's helloWorld example but not their real thing

bernhardmgruber:
So what do we do now? According to @ax3l's table, CUDA 11.2 does not support clang 4 and 5. But clang 4 and 5 work successfully with CUDA 11.2 in our CI (see commit 8326fe5). So do we allow them now or forbid them?

psychocoderHPC:
> So what do we do now? According to @ax3l's table, CUDA 11.2 does not support clang 4 and 5. But clang 4 and 5 work successfully with CUDA 11.2 in our CI (see commit [8326fe5](https://github.com/alpaka-group/alpaka/commit/8326fe5b2a6a85d75c5c3ac245bef57d5bd1d1fd)). So do we allow them now or forbid them?

The table is wrong. I posted yesterday the restrictions for all important compilers and clang 3.3+ is supported https://gist.github.com/ax3l/9489132#gistcomment-3564951

No need to change our CI

BenjaminW3 (approved):

PR #1241, opened 2021-01-15
fix OpenPower CPU name detection

psychocoderHPC:
For non x86 systems we assert if `getCpuName()`.
This PR changes the behavior and returns `unknown` for non x86 systems.

This PR is going directly in the release 0.6.0 and will be backported after the release.

# Reason for changing the behavior

I do not think that it is required to assert the code only because we do not know the name of the CPU. 
We can later add for OpenPower and ARM methods to query the CPU name, until this is implemented I do not see a reason to follow the old behavior.

### DIFF
--- a/include/alpaka/dev/cpu/SysInfo.hpp
+++ b/include/alpaka/dev/cpu/SysInfo.hpp
@@ -124,7 +124,7 @@ namespace alpaka
                 }
                 return std::string(cpuBrandString);
 #else
-                assert(false); // if we got here there is a bug in the cpuid() function
+                return std::string("unknown");
 #endif
             }
             //-----------------------------------------------------------------------------


j-stephan (approved):

sbastrakov (approved):

j-stephan:
After a quick Google search this seems to be the way to go on PowerPC.

macOS:

```c++
#include <sys/sysctl.h>
#include <sys/types.h>

auto proc = unsigned int{};
auto len = std::size_t{4};

sysctlbyname("hw.cpusubtype", &proc, &len, NULL, 0);
switch(proc)
{
    // proc can have a lot of different values that need to be handled manually
}
```

Linux:

```c++
#include <sys/auxv.h>

/* A pointer to a string (PowerPC and MIPS only).  On PowerPC, this identifies the real platform;
   may differ from AT_PLATFORM.  On MIPS, this identifies the ISA level (since Linux 5.7).*/
auto base_platform_string = reinterpret_cast<const char*>(getauxval(AT_BASE_PLATFORM));

/* A pointer to a string that identifies the hardware platform that the program is running on. 
   The dynamic linker uses this in the interpretation of rpath values.*/
auto platform_string = reinterpret_cast<const char*>(getauxval(AT_PLATFORM));
```

Do we care about the BSDs? They'd require special handling, too.

I can't implement and test this myself since I don't have access to OpenPOWER systems.

psychocoderHPC:
> After a quick Google search this seems to be the way to go on PowerPC.
> 
> macOS:
> 
> ```c++
> #include <sys/sysctl.h>
> #include <sys/types.h>
> 
> auto proc = unsigned int{};
> auto len = std::size_t{4};
> 
> sysctlbyname("hw.cpusubtype", &proc, &len, NULL, 0);
> switch(proc)
> {
>     // proc can have a lot of different values that need to be handled manually
> }
> ```
> 
> Linux:
> 
> ```c++
> #include <sys/auxv.h>
> 
> /* A pointer to a string (PowerPC and MIPS only).  On PowerPC, this identifies the real platform;
>    may differ from AT_PLATFORM.  On MIPS, this identifies the ISA level (since Linux 5.7).*/
> auto base_platform_string = reinterpret_cast<const char*>(getauxval(AT_BASE_PLATFORM));
> 
> /* A pointer to a string that identifies the hardware platform that the program is running on. 
>    The dynamic linker uses this in the interpretation of rpath values.*/
> auto platform_string = reinterpret_cast<const char*>(getauxval(AT_PLATFORM));
> ```
> 
> Do we care about the BSDs? They'd require special handling, too.
> 
> I can't implement and test this myself since I don't have access to OpenPOWER systems.

No BSD is IMO not important. Let us test this after the release.

PR #1248, opened 2021-01-27
Enable ADL for math functions

### LINKED ISSUE #855: Math functions are not ADL-friendly
The implementation uses hardcoded functions from `std` in situations like [this](https://github.com/ComputationalRadiationPhysics/alpaka/blob/f3d363bc581e378a4ef6440ec423c7575066b556/include/alpaka/math/exp/ExpStdLib.hpp#L49). It is done consistently for all math functions, I will use `exp` as an example.

This way is not friendly for ADL in case the argument is a user-defined type with `std::is_arithmetic` trait enabled and a custom `exp` implemented, since that implementation would not be found. A [standard approach](https://en.cppreference.com/w/cpp/language/adl) to look for the user-defined first and have `std::exp` as a fallback is to put `using std::exp;` inside the function and then call just `exp`. It looks to me we can do that and I made a toy example which works this way. In terms of implementation, it is awkward for us since we also need the right version of `exp` for return type deduction (before C++14), and it is not allowed to put `using std::exp;` just inside the `struct` (only inside a namespace or function body). So it looks like we would need another wrapper function to resolve all that. And then also to avoid conflicts between our function `exp` and the one we are ADL-looking for.

bernhardmgruber:
Pre-refactoring: Replace static functions of math trait implementations with `operator()`. We need this so we make unqualified calls without name lookup immediately finding the enclosing member function.

All math traits now have a default implementation which does an unqualified call to the corresponding math function. This allows to find overloads for user provided types which are not covered by alpaka or any backend. Integral and floating-point types are covered by the backends.

### DIFF
--- a/include/alpaka/math/abs/Traits.hpp
+++ b/include/alpaka/math/abs/Traits.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/core/Unused.hpp>
 
 #include <type_traits>
 
@@ -29,7 +30,16 @@ namespace alpaka
             //#############################################################################
             //! The abs trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Abs;
+            struct Abs
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not supported by the
+                    // backend and we could not find abs(TArg) in the namespace of your type.
+                    return abs(arg);
+                }
+            };
         } // namespace traits
 
         //-----------------------------------------------------------------------------
--- a/include/alpaka/math/abs/Traits.hpp
+++ b/include/alpaka/math/abs/Traits.hpp
@@ -44,7 +44,7 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC auto abs(T const& abs_ctx, TArg const& arg)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptMathAbs, T>;
-            return traits::Abs<ImplementationBase, TArg>::abs(abs_ctx, arg);
+            return traits::Abs<ImplementationBase, TArg>{}(abs_ctx, arg);
         }
     } // namespace math
 } // namespace alpaka
--- a/include/alpaka/math/acos/Traits.hpp
+++ b/include/alpaka/math/acos/Traits.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/core/Unused.hpp>
 
 #include <type_traits>
 
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The acos trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Acos;
+            struct Acos
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not supported by the
+                    // backend and we could not find acos(TArg) in the namespace of your type.
+                    return acos(arg);
+                }
+            };
         } // namespace traits
 
         //-----------------------------------------------------------------------------
--- a/include/alpaka/math/acos/Traits.hpp
+++ b/include/alpaka/math/acos/Traits.hpp
@@ -45,7 +45,7 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC auto acos(T const& acos_ctx, TArg const& arg)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptMathAcos, T>;
-            return traits::Acos<ImplementationBase, TArg>::acos(acos_ctx, arg);
+            return traits::Acos<ImplementationBase, TArg>{}(acos_ctx, arg);
         }
     } // namespace math
 } // namespace alpaka
--- a/include/alpaka/math/asin/Traits.hpp
+++ b/include/alpaka/math/asin/Traits.hpp
@@ -45,7 +45,7 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC auto asin(T const& asin_ctx, TArg const& arg)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptMathAsin, T>;
-            return traits::Asin<ImplementationBase, TArg>::asin(asin_ctx, arg);
+            return traits::Asin<ImplementationBase, TArg>{}(asin_ctx, arg);
         }
     } // namespace math
 } // namespace alpaka
--- a/include/alpaka/math/asin/Traits.hpp
+++ b/include/alpaka/math/asin/Traits.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/core/Unused.hpp>
 
 #include <type_traits>
 
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The asin trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Asin;
+            struct Asin
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not supported by the
+                    // backend and we could not find asin(TArg) in the namespace of your type.
+                    return asin(arg);
+                }
+            };
         } // namespace traits
 
         //-----------------------------------------------------------------------------
--- a/include/alpaka/math/atan/Traits.hpp
+++ b/include/alpaka/math/atan/Traits.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/core/Unused.hpp>
 
 #include <type_traits>
 
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The atan trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Atan;
+            struct Atan
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not supported by the
+                    // backend and we could not find atan(TArg) in the namespace of your type.
+                    return atan(arg);
+                }
+            };
         } // namespace traits
 
         //-----------------------------------------------------------------------------
--- a/include/alpaka/math/atan/Traits.hpp
+++ b/include/alpaka/math/atan/Traits.hpp
@@ -41,7 +41,7 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC auto atan(T const& atan_ctx, TArg const& arg)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptMathAtan, T>;
-            return traits::Atan<ImplementationBase, TArg>::atan(atan_ctx, arg);
+            return traits::Atan<ImplementationBase, TArg>{}(atan_ctx, arg);
         }
     } // namespace math
 } // namespace alpaka
--- a/include/alpaka/math/atan2/Traits.hpp
+++ b/include/alpaka/math/atan2/Traits.hpp
@@ -44,7 +44,7 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC auto atan2(T const& atan2_ctx, Ty const& y, Tx const& x)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptMathAtan2, T>;
-            return traits::Atan2<ImplementationBase, Ty, Tx>::atan2(atan2_ctx, y, x);
+            return traits::Atan2<ImplementationBase, Ty, Tx>{}(atan2_ctx, y, x);
         }
     } // namespace math
 } // namespace alpaka
--- a/include/alpaka/math/atan2/Traits.hpp
+++ b/include/alpaka/math/atan2/Traits.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/core/Unused.hpp>
 
 #include <type_traits>
 
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The atan2 trait.
             template<typename T, typename Ty, typename Tx, typename TSfinae = void>
-            struct Atan2;
+            struct Atan2
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, Ty const& y, Tx const& x)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not supported by the
+                    // backend and we could not find atan2(Tx, Ty) in the namespace of your type.
+                    return atan2(y, x);
+                }
+            };
         } // namespace traits
 
         //-----------------------------------------------------------------------------
--- a/include/alpaka/math/cbrt/Traits.hpp
+++ b/include/alpaka/math/cbrt/Traits.hpp
@@ -11,6 +11,7 @@
 
 #include <alpaka/core/Common.hpp>
 #include <alpaka/core/Concepts.hpp>
+#include <alpaka/core/Unused.hpp>
 
 #include <type_traits>
 
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The cbrt trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Cbrt;
+            struct Cbrt
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not supported by the
+                    // backend and we could not find cbrt(TArg) in the namespace of your type.
+                    return cbrt(arg);
+                }
+            };
         } // namespace traits
 
         //-----------------------------------------------------------------------------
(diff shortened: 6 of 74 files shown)


include/alpaka/math/abs/Traits.hpp
```diff
@@ -44,7 +54,7 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC auto abs(T const& abs_ctx, TArg const& arg)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptMathAbs, T>;
-            return traits::Abs<ImplementationBase, TArg>::abs(abs_ctx, arg);
+            return traits::Abs<ImplementationBase, TArg>{}(abs_ctx, arg);
```
sbastrakov, line 57: I think i don't fully understand it. Now a user calls that `alpaka::abs(acc, arg)` declared on line 54 in their kernel. It in turn does ADL on line 40 to switch to user's `abs()` definition when applicable. But when and how do we switch to backend-specific calls such as `struct Abs<AbsUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_floating_point<TArg>::value>>` ?

sbastrakov: I think the schema of how these things are organized is worth documenting. As seen from my question, it's probably not that simple to grasp, even though I discovered lack of ADL there and tried a simple fix (with a different way than yours)

include/alpaka/math/abs/Traits.hpp
```diff
@@ -29,7 +30,16 @@ namespace alpaka
             //#############################################################################
             //! The abs trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Abs;
+            struct Abs
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/acos/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The acos trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Acos;
+            struct Acos
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/asin/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The asin trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Asin;
+            struct Asin
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/atan/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The atan trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Atan;
+            struct Atan
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/atan2/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The atan2 trait.
             template<typename T, typename Ty, typename Tx, typename TSfinae = void>
-            struct Atan2;
+            struct Atan2
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, Ty const& y, Tx const& x)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/cbrt/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The cbrt trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Cbrt;
+            struct Cbrt
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/ceil/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The ceil trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Ceil;
+            struct Ceil
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/cos/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The cos trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Cos;
+            struct Cos
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/erf/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The erf trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Erf;
+            struct Erf
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/exp/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The exp trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Exp;
+            struct Exp
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/floor/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The floor trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Floor;
+            struct Floor
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/fmod/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The fmod trait.
             template<typename T, typename Tx, typename Ty, typename TSfinae = void>
-            struct Fmod;
+            struct Fmod
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, Tx const& x, Ty const& y)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/log/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The log trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Log;
+            struct Log
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/max/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The max trait.
             template<typename T, typename Tx, typename Ty, typename TSfinae = void>
-            struct Max;
+            struct Max
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, Tx const& x, Ty const& y)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/min/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The min trait.
             template<typename T, typename Tx, typename Ty, typename TSfinae = void>
-            struct Min;
+            struct Min
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, Tx const& x, Ty const& y)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/trunc/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The trunc trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Trunc;
+            struct Trunc
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/tan/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The tan trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Tan;
+            struct Tan
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/pow/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The pow trait.
             template<typename T, typename TBase, typename TExp, typename TSfinae = void>
-            struct Pow;
+            struct Pow
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TBase const& base, TExp const& exp)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/remainder/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The remainder trait.
             template<typename T, typename Tx, typename Ty, typename TSfinae = void>
-            struct Remainder;
+            struct Remainder
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, Tx const& x, Ty const& y)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/round/Traits.hpp
```diff
@@ -27,17 +28,44 @@ namespace alpaka
             //#############################################################################
             //! The round trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Round;
+            struct Round
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/rsqrt/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The rsqrt trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Rsqrt;
+            struct Rsqrt
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/sin/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The sin trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Sin;
+            struct Sin
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/sincos/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The sincos trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct SinCos;
+            struct SinCos
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg, TArg& result_sin, TArg& result_cos)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

include/alpaka/math/sqrt/Traits.hpp
```diff
@@ -27,7 +28,16 @@ namespace alpaka
             //#############################################################################
             //! The sqrt trait.
             template<typename T, typename TArg, typename TSfinae = void>
-            struct Sqrt;
+            struct Sqrt
+            {
+                ALPAKA_FN_HOST_ACC auto operator()(T const& ctx, TArg const& arg)
+                {
+                    alpaka::ignore_unused(ctx);
+                    // This is an ADL call. If you get a compile error here then your type is not suppored by the
```
j-stephan: Typo: suppored instead of supported

j-stephan (changes requested):
Nice work! Requested changes are:

1. Clarification regarding @sbastrakov's question.
2. A typo that propagated through copy & paste.

include/alpaka/math/sqrt/Traits.hpp
bernhardmgruber: Are you bored? :D One comment would have been enough!

include/alpaka/math/abs/Traits.hpp
bernhardmgruber: The user calls `alpaka::math::abs(acc, value);` and the function on line 54 is picked. On line 57 we instantiate `traits::Abs<ImplementationBase, TArg>`. Notice that the template arguments are the backend and the type on which to compute the abs. If there is any specialization of `traits::Abs`, it will be picked. For `float` and `double`, such specializations exist for all backends. However, if there is no specialization, the primary template on line 33 will be instantiated which proceeds to do an ADL call.

We could document that somewhere. Where could be a good place?

j-stephan:
Could you rebase on `develop`? These old CI tests are blocking my merge button.

bernhardmgruber:
Rebased

j-stephan:
Re-requesting review from @sbastrakov: Has your question been sufficiently answered?

include/alpaka/math/abs/Traits.hpp
sbastrakov: Thanks for explaining @bernhardmgruber . Where to document is a good question. I am not sure, maybe as a new subsection [here](https://alpaka.readthedocs.io/en/latest/dev/details.html). Btw I've also found out [the last point about naming here](https://alpaka.readthedocs.io/en/latest/dev/style.html#traits) is violated by this PR. However, I remember Benjamin himself recommending this new naming you used, so should be no issue there.

bernhardmgruber: As you said, @BenjaminW3 mentioned in some comment somewhere, that we should have picked `operator()` for trait implementations.
Furthermore, I had to change the function name to make it work. Consider this:
```c++
struct Acos
{
    ALPAKA_FN_HOST_ACC auto acos(T const& ctx, TArg const& arg) // instead of operator()
    {
        alpaka::ignore_unused(ctx);
        // This is an ADL call. If you get a compile error here then your type is not supported by the
        // backend and we could not find acos(TArg) in the namespace of your type.
        return acos(arg);
    }
};
```
<details>
  <summary>Guess what would happen :)</summary>
  An infinite recursion :D
</details>

sbastrakov (dismissed):
My question was answered. I still think it's worth to add a brief explanation for documentation (see discussion above), but will leave it to @bernhardmgruber to decide.

bernhardmgruber:
I have added documentation for this feature.

sbastrakov (approved):

psychocoderHPC (approved):

psychocoderHPC:
@j-stephan Is your request resolved?

j-stephan (approved):

j-stephan:
Yes. You're free to hit the merge button.

PR #1254, opened 2021-02-08
Fix declareVar<BlockSharedMemStOmp5>(): only thread 0 should alloc

jkelling:
Bug when running more than one thread per block, probably a regression introduced during the refactor for the OMP5 backend PR.

### DIFF
--- a/include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
+++ b/include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
@@ -45,8 +45,10 @@ namespace alpaka
             static auto declareVar(BlockSharedMemStOmp5 const& smem) -> T&
             {
 #    pragma omp barrier
-                smem.alloc<T>();
-#    pragma omp barrier
+#    pragma omp single
+                {
+                    smem.alloc<T>();
+                }
                 return smem.getLatestVar<T>();
             }
         };


jkelling:
@psychocoderHPC Only on Windows run failed, which only shows: "This check failed".

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -45,7 +45,8 @@ namespace alpaka
             static auto declareVar(BlockSharedMemStOmp5 const& smem) -> T&
             {
 #    pragma omp barrier
-                smem.alloc<T>();
+                if(::omp_get_thread_num() == 0)
```
psychocoderHPC: Why do we need a barrier before?
Can we not use `#pragma omp single` or will this create issues with the counter in `smem`?

psychocoderHPC (changes requested):

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
sbastrakov: I think a barrier here was so that the allocation happens for all threads at once, not to provide exclusivity for a thread that `single` would. I am not familiar with this piece to judge how it should be working

jkelling: @sbastrakov was faster: Yes, the barrier before is required, because after the `alloc` and subsequent barrier all threads read the new offset to get the pointer to the new shared memory. If one `declareVar()` call is followed by another, the allocating thread could get here before some other thread read the previous offset.

`single` implies a barrier after. It may me a good idea to use it, however, together with a barrier before: `single` might in some cases be more efficient than a barrier.

psychocoderHPC (dismissed):

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
jkelling: Updated to use `omp single`.

psychocoderHPC: Only to be sure that we do not break anything, it the barrier after the single mandatory and guaranteed in the standard?

psychocoderHPC: found this in the standard:

```
The single construct specifies that the associated structured block is executed by only one of the
threads in the team (not necessarily the master thread), in the context of its implicit task. The other
threads in the team, which do not execute the block, wait at an implicit barrier at the end of the
single construct unless a nowait clause is specified.
```

psychocoderHPC (approved):

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
jkelling: Yes, this paragraph guarantees that there will be barrier.

PR #1258, opened 2021-02-24
fix CPU static shared memory implementation

### LINKED ISSUE #1257: shared memory allocation issue
Allocating shared memory with CPU accelerators within a function which is called from a loop will result into allocation issues due to the limited shared memory per block (64kib)

```
Assertion `m_allocdBytes <= m_capacity' failed.
```

pseudo example:
```
foo()
{
  auto& sharedMem
    = alpaka::declareSharedVar<char * [1024], __COUNTER__>(acc);
}

main()
{
  for(int i = 0; i < 1000; ++i)
    foo();
}
```

For CUDA this is not a problem because calling `declareSharedVar` with the same id, here provided by `__COUNTER__` will reverence always to the same memory and only allocated shared memory once.
This strange CUDA behavior was in the past the reason why we introduced the unique id for `declareSharedVar`.
Reusing shared memory could be intended by the user e.g. in mallocMC https://github.com/alpaka-group/mallocMC/blob/00df9a5b7c33ae7ca70fcab9cc77951c098dfb6b/src/include/mallocMC/distributionPolicies/XMallocSIMD.hpp#L126

We do not declare in alpaka how `declareSharedVar` should be used or what is forbidden. Based on the definition we can fix this issue.

psychocoderHPC:
fix #1257

Using `declareSharedVar` with the same template signature within a kernel should always return the same object, located at the same memory address. This was not the case on CPU and was not checked in the test cases.

- use fixed sizes memory pool for all CPU accelerator to provide static shared memory
- remove `BlockSharedMemDynAlignedAlloc.hpp`,
`BlockSharedMemStNoSync.hpp` and `BlockSharedMemStMasterSync.hpp`
- add a synched shared memory pool `BlockSharedMemStMemberMasterSync.hpp`

The internal workflows on how static memory allocation during the kernel execution behaves is slightly changed.
If a thread is executing more than one block then the static shared memory is only allocated once during the execution of the first block.
All followed blocks will reuse the already allocated chunks. This is safe because the size of the allocations is fixed for the full kernel.


psychocoderHPC:
@jkelling Could you please check this PR with OpenAcc and OpenMP5.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -122,18 +42,25 @@ namespace alpaka
...
             static auto declareVar(BlockSharedMemStMember<TDataAlignBytes> const& smem) -> T&
             {
-                smem.template alloc<T>();
-                return smem.template getLatestVar<T>();
+                auto* data = smem.template getVar<T>(TuniqueId);
+
+                if(!data)
+                {
+                    smem.template alloc<T>(TuniqueId);
+                    data = smem.template getVar<T>(TuniqueId);
```
jkelling: `BlockSharedMemStMember` can only be used in accelerators which only run one thread per block, thus `getLatestVar` would be sufficient here, no need to iterate through all variables again.

include/alpaka/block/shared/st/BlockSharedMemStMemberMasterSync.hpp
```diff
@@ -0,0 +1,95 @@
...
+            std::function<void()> fnSync,
+            std::function<bool()> fnIsMasterThread)
+            : detail::BlockSharedMemStMemberImpl<TDataAlignBytes>(mem, capacity)
+            , m_syncFn(fnSync)
+            , m_isMasterThreadFn(fnIsMasterThread)
+        {
+        }
+
+        std::function<void()> m_syncFn;
+        std::function<bool()> m_isMasterThreadFn;
```
jkelling, line 43: Are you sure your want to use `std::function`, i.e. a function pointer, here? This could also be solved using a template predicate.

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -44,12 +44,20 @@ namespace alpaka
...
+                if(!data)
                 {
-                    smem.alloc<T>();
+#    pragma omp barrier
+#    pragma onmp single
+                    {
+                        blockSharedMemSt.template alloc<T>(TuniqueId);
+                    }
+                    // lookup for the data chunk allocated by the master thread
+                    data = blockSharedMemSt.template getVar<T>(TuniqueId);
```
jkelling: Due to the `barrier` before the `single` no second allocation can happen before a thread reaches this line, so `getLatestVar`would be sufficient.

The barrier is placed at the start because it guards again crosstalk with the previous `declareVar`, but is actually only protects the calle to `getVar`/`getLatestVar` after the alloc, so it could also be placed there if you prefer.

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -44,12 +44,20 @@ namespace alpaka
...
                 }
-                return smem.getLatestVar<T>();
+                ALPAKA_ASSERT(data != nullptr);
```
jkelling: This assert should be guarded by a
```c++
```
otherwise it will break offloading in debug builds.

Maybe we need an `ALPAKA_ASSERT_OFFLOAD` or `ALPAKA_ASSET_DEVICE` which does this.

include/alpaka/ctx/block/CtxBlockOacc.hpp
```diff
@@ -228,10 +228,17 @@ namespace alpaka
...
-                });
-                return smem.template getLatestVar<T>();
+                auto* data = smem.template getVar<T>(TuniqueId);
+
+                if(!data)
+                {
+                    traits::SyncBlockThreads<CtxBlockOacc<TDim, TIdx>>::masterOpBlockThreads(smem, [&data, &smem]() {
+                        smem.template alloc<T>();
+                    });
+                    data = smem.template getVar<T>(TuniqueId);
```
jkelling: Here `getLatestVar` would be sufficient again. `SyncBlockThreads` acts like a `omp single`, except, that there is also a barrier in front, so the code is equivalent to the OpenMP5 version.

jkelling (changes requested):

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
psychocoderHPC: This is required to fulfill the requirement that some allocations of shared memory (same template signature) will not allocate with each call new memory.

psychocoderHPC: > BlockSharedMemStMember can only be used in accelerators which only run one thread per block, thus getLatestVar would be sufficient here, no need to iterate through all variables again.

Get the last variable is not allowed to be called from outside anymore. The reason is that the last element in the memory blob is always the message header. 
IMO the performance drawback is very tiny.

include/alpaka/block/shared/st/BlockSharedMemStMemberMasterSync.hpp
psychocoderHPC: It is equal to the implementation before. It based on the synchronized allocation method we have already in alpaka.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: This what the first `getVar` is for. The second one happens after a new allocation was already found to be required.

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
psychocoderHPC: Same as my answer above. This is not possible because of the inlined dictionary inside the memory blob we use. Possible would be to return the memory from the function `blockSharedMemSt.template alloc<T>(TuniqueId);`

include/alpaka/ctx/block/CtxBlockOacc.hpp
psychocoderHPC: Same as before.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
psychocoderHPC: The message header is aligned and the last element in the memory blob, therefore I do not know the starting address of the last allocated variable.

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
jkelling: This would not work because `alloc` is only called by the master.

psychocoderHPC: Yes, would only work for accelerators where the block extent is limited to one thread.

psychocoderHPC:
I guarded the assert. The `getLastVar` request is not possible because of the mentioned reasons (the last entry in the memory blob is always a metadata header).

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
```diff
@@ -0,0 +1,190 @@
...
+            //!
+            //! \tparam T type should fit into the chunk
+            //! \param byteOffset Current byte offset.
+            //! \result Byte offset to the end of the user data chunk.
+            template<typename T>
+            std::uint32_t varChunkSize(uint32_t byteOffset) const
+            {
+                static_assert(
+                    core::vectorization::defaultAlignment >= alignof(T),
+                    "Unable to get block shared static memory for types with alignment higher than defaultAlignment!");
```
jkelling, line 150: I believe we should check against `TDataAlignBytes` here, which only happens to default to `core::vectorization::defaultAlignment`.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
```diff
@@ -0,0 +1,190 @@
...
+                    metaDataPtr = reinterpret_cast<MetaData*>(m_mem + off);
+                }
+
+                // Last header, variable with id not found.
+                if(metaDataPtr->offset == 0)
+                {
+                    return nullptr;
+                }
+
+                return reinterpret_cast<T*>(&m_mem[off - sizeof(T)]);
```
jkelling, line 121: Will this not break if the alignment of `T` is smaller than that of `MetaData`, i.e. if there are padding bytes before the address `metaDataPtr`?

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -44,12 +44,22 @@ namespace alpaka
...
             {
-#    pragma omp barrier
-#    pragma omp single
+                auto* data = blockSharedMemSt.template getVar<T>(TuniqueId);
+
+                if(!data)
                 {
-                    smem.alloc<T>();
+#    pragma omp barrier
+#    pragma onmp single
```
jkelling: There is a superfluous `n` here.

jkelling:
Can we rename `getLatestVar` to `getLastHeader` and make it not a template, because it will only ever return `MetaData`. The last variable should always be a header in this imlementation? Then add a public function:
```c++
template<typename T>
T& getLatestVar() const
{
 return *reinterpret_cast<T*>(
   reinterpret_cast<char*>(&getLastHeader()) - sizeof(T);
  );
}
```
and use this to get rid of the additional iteration through the list?

Note, that this code has the same issue with smaller alignments than `MetaData` which I mentioned above.

psychocoderHPC:
> Can we rename `getLatestVar` to `getLastHeader` and make it not a template, because it will only ever return `MetaData`. The last variable should always be a header in this imlementation? Then add a public function:
> 
> ```c++
> template<typename T>
> T& getLatestVar() const
> {
>  return *reinterpret_cast<T*>(
>    reinterpret_cast<char*>(&getLastHeader()) - sizeof(T);
>   );
> }
> ```
> 
> and use this to get rid of the additional iteration through the list?
> 
> Note, that this code has the same issue with smaller alignments than `MetaData` which I mentioned above.

No this will not work because the header can be aligned and you can not find the unaligned start if you subtract the header size only. 
Never the less I moved this PR into draft because I think I know now how to solve the problem.

include/alpaka/acc/AccCpuFibers.hpp
```diff
@@ -14,9 +14,8 @@
 // Base classes.
 #    include <alpaka/atomic/AtomicHierarchy.hpp>
 #    include <alpaka/atomic/AtomicNoOp.hpp>
-#    include <alpaka/atomic/AtomicStdLibLock.hpp>
```
psychocoderHPC, line 17: deleted one line too much.

psychocoderHPC:
The PR can now be review.

@jkelling I refactored the PR so that know it is again possible to get the last allocated variable without traversing the dictionary a second time.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: Note: @jkelling's comment was from before the refactoring.
In the refactored implementation `off` is pointing to the end of the variable, therefore `off - sizeof(T)` will point to an aligned offset.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: I have not changed this code but this check is valid.
I think you assumed at the moment where you wrote this code that `m_mem` is default aligned https://github.com/alpaka-group/alpaka/blob/ed3c123b22684b9a342506d4d46a03a8ac6bd9bf/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp#L98-L102
Therefore it is possible to do the alignment based on an offset relative to `m_mem`.
It would be possible to remove the assert fully if we always compute the alignment based on `m_mem + byte_offset`.

Changing the assert to `>= TDataAlignBytes` would mean you can with OpenMP5 only handle aligned data up to `4` because of https://github.com/alpaka-group/alpaka/blob/ed3c123b22684b9a342506d4d46a03a8ac6bd9bf/include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp#L31

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
jkelling: @psychocoderHPC ping "onmp -> omp"

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
jkelling: Good point. I would actually chalk this up as a bug in the OMP5 ( [and OpenACC](https://github.com/alpaka-group/alpaka/blob/ed3c123b22684b9a342506d4d46a03a8ac6bd9bf/include/alpaka/ctx/block/CtxBlockOacc.hpp#L43) backends: I thought about 4-byte banks on GPUs and never came back to fix this. The intention of the template parameter was to set the alignment of the memory pool, which would limit the alignment of the variables because alignment was done relative to this address. If we can safely align the pointer address, we can indeed remove the assert which would make the template parameter obsolete, too.

... Actually, the template parameter `TDataAlignBytes` is already obsolete, because the alignment of the memory is hardcoded:
https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp#L49

Please remove `TDataAlignBytes` and the assert.

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -44,12 +44,21 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             static auto declareVar(BlockSharedMemStOmp5 const& smem) -> T&
```
jkelling, line 45: Below: use `smem` instead of  `blockSharedMemSt`.

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -59,7 +68,6 @@ namespace alpaka
             //-----------------------------------------------------------------------------
             static auto freeVars(BlockSharedMemStOmp5 const& mem) -> void
```
jkelling: Remove unused  parameter name `mem`.

jkelling (changes requested):

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: `TDataAlignBytes` is used to enforce a minimum alignment. Also, the static assert in the current version is correct, as you linked the dynamic shared memory object is aligned to `core::vectorization::defaultAlignment`. Therefore we can not guarantee any alignment larger than this.
To have a minimum alignment of 4byte for OpenMP which can be offloaded to GPUs sound valid, if this should be changed we should do it after this PR in a separate PR to not hide this change within this PR.

psychocoderHPC:
@jkelling Thanks for the review, I fixed all except https://github.com/alpaka-group/alpaka/pull/1258#discussion_r582788979

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
jkelling: Right. I missed the `std::max`. As a minimum alignment it should stay. I have to say, the name is a bad choice then. Could you change it to `TMinDataAlignBytes` in this PR @psychocoderHPC ?

include/alpaka/ctx/block/CtxBlockOacc.hpp
```diff
@@ -228,10 +228,17 @@ namespace alpaka
...
-                traits::SyncBlockThreads<CtxBlockOacc<TDim, TIdx>>::masterOpBlockThreads(smem, [&smem]() {
-                    smem.template alloc<T>();
-                });
-                return smem.template getLatestVar<T>();
+                auto* data = smem.template getVarPtr<T>(TuniqueId);
+
+                if(!data)
+                {
+                    traits::SyncBlockThreads<CtxBlockOacc<TDim, TIdx>>::masterOpBlockThreads(smem, [&data, &smem]() {
+                        smem.template alloc<T>();
```
jkelling: `TuniqueId` is missing in the call to `alloc`.

psychocoderHPC:
@alpaka-group/alpaka-maintainers please review this PR, @jkelling points are addressed and implemented.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -118,18 +43,25 @@ namespace alpaka
...
-                return smem.template getLatestVar<T>();
+                auto* data = smem.template getVarPtr<T>(TuniqueId);
+
+                if(!data)
+                {
+                    smem.template alloc<T>(TuniqueId);
+                    data = smem.template getLatestVarPtr<T>();
+                }
+                ALPAKA_ASSERT(data != nullptr);
+                return std::ref(*data);
```
bernhardmgruber: Why are you constructing a `std::reference_wrapper<T>` here, that immediately converts to a `T&` again? Why not just `return *data`?

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
```diff
@@ -118,18 +43,25 @@ namespace alpaka
...
         struct FreeSharedVars<BlockSharedMemStMember<TDataAlignBytes>>
         {
             //-----------------------------------------------------------------------------
-            static auto freeVars(BlockSharedMemStMember<TDataAlignBytes> const& mem) -> void
+            static auto freeVars(BlockSharedMemStMember<TDataAlignBytes> const&) -> void
             {
-                mem.free();
+                // shared memory block data will be reused
             }
         };
```
bernhardmgruber, line 66: From a quick look at the alpaka code base, it seems that was the last use of `FreeSharedVars` that actually does something. So we can remove the entire trait and all specializations. Mind that `AccOacc` has an implementation of `FreeSharedVars`, but it delegates again `freeSharedVars`. So either it recurses indefinitely or it does nothing ;)

include/alpaka/block/shared/st/BlockSharedMemStMemberMasterSync.hpp
```diff
@@ -0,0 +1,95 @@
...
+                    {
+                        blockSharedMemSt.template alloc<T>(TuniqueId);
+                    }
+
+                    blockSharedMemSt.m_syncFn();
+                    // lookup for the data chunk allocated by the master thread
+                    data = blockSharedMemSt.template getLatestVarPtr<T>();
+                }
+                ALPAKA_ASSERT(data != nullptr);
+                return std::ref(*data);
```
bernhardmgruber: Why not `return *data`?

include/alpaka/block/shared/st/BlockSharedMemStOmp5.hpp
```diff
@@ -44,22 +44,31 @@ namespace alpaka
...
+                    {
+                        smem.template alloc<T>(TuniqueId);
+                    }
+                    // lookup for the data chunk allocated by the master thread
+                    data = smem.template getLatestVarPtr<T>();
                 }
-                return smem.getLatestVar<T>();
+                ALPAKA_ASSERT_OFFLOAD(data != nullptr);
+
+                return std::ref(*data);
```
bernhardmgruber: Why not `return *data`?

include/alpaka/ctx/block/CtxBlockOacc.hpp
```diff
@@ -228,10 +228,17 @@ namespace alpaka
...
+
+                if(!data)
+                {
+                    traits::SyncBlockThreads<CtxBlockOacc<TDim, TIdx>>::masterOpBlockThreads(smem, [&data, &smem]() {
+                        smem.template alloc<T>(TuniqueId);
+                    });
+                    data = smem.template getLatestVarPtr<T>();
+                }
+                ALPAKA_ASSERT(data == nullptr);
+                return std::ref(*data);
```
bernhardmgruber: Why not `return *data`?

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
```diff
@@ -0,0 +1,168 @@
...
+            }
+
+            //! Offset in bytes relative to m_mem to next free data area.
+            //! The last aligned before the free area is always a meta data header.
+            mutable std::uint32_t m_allocdBytes = 0u;
+
+            //! Memory layout
+            //! |Header|Padding|Variable|Padding|Header|....uninitialized Data ....
+            //! Size of padding can be zero if data after padding is already aligned.
+            mutable uint8_t* m_mem;
```
bernhardmgruber: I wonder if the implementation would be easier, if we stored the `MetaData` in a separate array. Thus no alignment would be needed for `MetaData`s and finding variables by id could be faster.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
```diff
@@ -0,0 +1,168 @@
...
+            //! \param byteOffset Current byte offset.
+            //! \result Byte offset to the end of the data chunk, relative to m_mem..
+            template<typename T>
+            std::uint32_t varChunkEnd(uint32_t byteOffset) const
+            {
+                static_assert(
+                    core::vectorization::defaultAlignment >= alignof(T),
+                    "Unable to get block shared static memory for types with alignment higher than defaultAlignment!");
+                constexpr std::uint32_t align = static_cast<std::uint32_t>(std::max(TMinDataAlignBytes, alignof(T)));
+                return (byteOffset / align + (byteOffset % align > 0u)) * align + static_cast<uint32_t>(sizeof(T));
```
bernhardmgruber, line 152: This seems needlessly complex, try:
`(byteOffset + align - 1) / align * align`

And since I have seen this code in multiple places by now, consider moving it into a separate function.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
bernhardmgruber: I think `m_mem` does not need to be `mutable`.

bernhardmgruber (changes requested):

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
jkelling: We do not know how many headers we need at compile time and we do not want to block more shared memory than from being used by the kernel than absolutely necessary.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: both are equivalent where the current implementation has the advantage that it is not overflowing when byte offset is near to 4 billion. But for this implementation, the overflow is not a problem at all.

> And since I have seen this code in multiple places by now, consider moving it into a separate function.

This can be done in a separate PR, this should not be mixed with this fix.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: It should be mutable because the accelerator is inheriting from this class and is a constant object in all kernels.

include/alpaka/ctx/block/CtxBlockOacc.hpp
psychocoderHPC: It was in old implementations like this. https://github.com/alpaka-group/alpaka/pull/1258/files/55e205b395a5b780430780ee256e0a716e175feb#diff-33fbcb49de69fef7bf79c86c0106d8497046ffd67dfc8305c1912987a6a16585L81

I tried to understand why it was returning a `std::ref`. Since I was not sure I keep it. @BenjaminW3 could you explain why the old implementation returned a `std::ref`

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
jkelling: `m_mem` is never modified. It does not need to be mutable to point to mutable memory. The actual underlying memory must be mutable, but this is [here](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp#L104).

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
psychocoderHPC: Because from the concept it is correct to have a `free` method to flush or delete the memory. I already deleted it but reactivated the code to keep the door open that someone can implement the shared memory with other allocators.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
sbastrakov: I think here, as we indeed immediately convert to `T&`, it's indeed better to use just `return *data`

include/alpaka/block/shared/st/BlockSharedMemStMemberMasterSync.hpp
sbastrakov: Here as well.

include/alpaka/ctx/block/CtxBlockOacc.hpp
sbastrakov: I think with the signature of the function returning `T&` explicitly, so as `auto ... -> T&`, it's better to remove `std::ref` from return of such funcitons. It indeed does nothing and just causes confusion. If there is a function what just has `auto` as return type and returns `std::ref` it would be another story.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
jkelling: @psychocoderHPC you could straight-ip put it down as:
```c++
uint8_t* const m_mem;
```

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
bernhardmgruber: I think this is a case of YAGNI.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
bernhardmgruber: Created issue: https://github.com/alpaka-group/alpaka/issues/1261

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
bernhardmgruber: I considered adding `const` but I do not know if inheriting classes need to be movable, which no longer works with const data members.

bernhardmgruber: IIUC these are CPU implementations, so it should not be an issue to have a separate 1KiB block for `MetaData`s. That alone would allow 128 shared memory variable declarations in a kernel. I think that could simplify the code a lot, but it is just a suggestion.

jkelling: E.g.:
https://github.com/alpaka-group/alpaka/blob/ed3c123b22684b9a342506d4d46a03a8ac6bd9bf/include/alpaka/acc/AccCpuSerial.hpp#L112
... no.

bernhardmgruber: Fine! So `uint8_t* const m_mem;` is probably a good choice then.

psychocoderHPC: right I mixed the pointer with the value to store how many bytes we already allocated.
I will change it.

psychocoderHPC: I prefer the integrated dictionary. Adding 1KiB will introduce another implicit restriction. And will only allow holding 85 variables. (4 byte id and 8 byte pointer)

jkelling: `BlockSharedMemStMember` was, and `BlockSharedMemStMemberImpl` now is, primarily intended to be used  by the OpenMP5 and OpenACC backends, i.e. ultimately on GPU. They have replaced the old CPU shared memory implementations because they avoid malloc, improve memory locality (creating a chunk of memory which might just end up in L1) and limit the amount of shared memory per block to the same amount as on GPU, thereby forcing improved portability.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: It was required when the CPU accelerator were allocating shared memory dynamically on the heap. It is obsolete at the moment.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
bernhardmgruber: > I prefer the integrated dictionary.

Fine, keep it then!

> Adding 1KiB will introduce another implicit restriction. And will only allow holding 85 variables. (4 byte id and 8 byte pointer)

`MetaData` is defined as:
```c++
struct MetaData {
  uint32_t id = std::numeric_limits<uint32_t>::max();
  uint32_t offset = 0;
};
```
that makes it 8 bytes in size. 1KiB / 8B = 128.

> BlockSharedMemStMember was, and BlockSharedMemStMemberImpl now is, primarily intended to be used by the OpenMP5 and OpenACC backends, i.e. ultimately on GPU. They have replaced the old CPU shared memory implementations.

If they replaced all CPU implementations, then they *are* intended for those as well. But they new implementation is definitely better than any `malloc()`-based one (I did not know the old implemenation). Locality is rightfully important!

bernhardmgruber (approved):

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
bernhardmgruber: Go ahead, leave it in if you want. I would drop it, decreasing the legacy.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
jkelling: > If they replaced all CPU implementations, then they are intended for those as well.

Indeed, they are now. Just wanted to point out their "historical" reason for being. They have to keep working in that context as well as possible.

include/alpaka/block/shared/st/BlockSharedMemStMember.hpp
jkelling: I have actually no strong opinion on dropping or leaving (slightly leaning towards dropping, actually), except, that this change would belong into a separate PR.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: > that makes it 8 bytes in size. 1KiB / 8B = 128.

Yes, you are right, I was thinking about an id and pointer instead of keeping the current format. 
Never the less I will keep inlined dictionary.

bernhardmgruber:
@jkelling you still have several unresolved conversations. Could you please mark them resolved, if @psychocoderHPC has addressed the issues? I think then we can hit the green button :)

jkelling:
> @jkelling you still have several unresolved conversations. Could you please mark them resolved, if @psychocoderHPC has addressed the issues? I think then we can hit the green button :)

I do not see buttons to resolve the conversations. (Maybe because I am not a maintainer?) Everything has been addressed. @psychocoderHPC as the submitter can resolve the conversations.

PR #1266, opened 2021-03-01
fix queue test

psychocoderHPC:
Calling `alpaka::empty()` can return false negatives, therefore using the result and enforce a return value is not possible and will result in random false results.

By adding a loop around the condition and wait between the querries we can control when we latest expect a valid result.

The behavior of `alpaka::empty()` is documented in #1265. The reason for the false negatives of `empty` is that the used method `(hip|cuda)StreamQuery` can result `(hip|cuda)ErrorNotReady` which will be mapped by alpaka to false (queue not empty).
Mapping an undefined status to `false` is IMO correct.

### DIFF
--- a/test/unit/queue/src/QueueTest.cpp
+++ b/test/unit/queue/src/QueueTest.cpp
@@ -15,6 +15,7 @@
 
 #include <catch2/catch.hpp>
 
+#include <atomic>
 #include <future>
 #include <thread>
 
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
After this PR I can reactivate callbacks for the HIP backend. https://github.com/ROCm-Developer-Tools/HIP/issues/1205 is fixed in HIP but our test is broken (fixed with this PR)

test/unit/queue/src/QueueTest.cpp
```diff
@@ -82,9 +104,9 @@ TEMPLATE_LIST_TEST_CASE(
     using Fixture = alpaka::test::QueueTestFixture<DevQueue>;
     Fixture f;
 
-    bool CallbackFinished = false;
+    volatile bool CallbackFinished = false;
```
bernhardmgruber: I think this is a wrong use of `volatile`. You need a synchronization primitive because `CallbackFinished` is potentially accessed from 2 threads. I would use `std::atomic<bool>`.

psychocoderHPC: `volatile` is correct to avoid that the value is cached but I agree that using `std::atomic` makes much more sense because of the access from two threads.

bernhardmgruber: You are right in the sense that `volatile` avoids load/store optimizations. This might be correct in this case if there are memory fences somewhere inside `alpaka::enqueue` or `alpaka::wait`. But `std::atomic` is definitely safe, independent on what alpaka or the backends are doing.

test/unit/queue/src/QueueTest.cpp
```diff
@@ -26,6 +26,28 @@ using TestQueues = alpaka::meta::Concatenate<
 #endif
     >;
 
+//! Equivalent too CHECK but conditions with false negatives
```
sbastrakov: Typo: should be `to` instead of `too`. I would also rewrite as `Equivalent to CHECK but accounts for potential false negatives`

test/unit/queue/src/QueueTest.cpp
```diff
@@ -26,6 +26,28 @@ using TestQueues = alpaka::meta::Concatenate<
...
+//!
+//! This is required when checking if an event or a queue is finished/empty.
+//! \warning This macro can be used on host only!
+//!
+//! \param count number of checks until the check will use the last result of the condition
```
sbastrakov: In the implementation it only applies for negative results. The first positive result will be used.

sbastrakov (changes requested):
Just some minor text suggestions.

bernhardmgruber (dismissed):

sbastrakov (approved):

bernhardmgruber (approved):

psychocoderHPC:
ready to merge

PR #1271, opened 2021-03-09
Add a new example to demonstrate specializaing kernels for a particular backend

### LINKED ISSUE #1267: Add an example demonstrating specialization of a kernel for a particular backend
We kinda advertize this design, and alpaka made a choice to use trait-based implementation to enable it from the very beginning. This arguably makes alpaka a bit more hard to develop and may appear overly complicated to users in exchange for more flexibility and potential for fine-grained optimization on the level of a single kernel or a single device-level alpaka function. So we should add a new example that demonstrates doing so on a technical level.

sbastrakov:
When developing it, I was considering between a really simplistic example that only demonstrate the technical way of doing so, without a particular use case, or a more full fledged one with actual optimization. I decided to go with the former, as it is much easier to understand by a user who want to apply this technique in their code. So it only prints which version is executed. However in the comments I added a brief hint that this way one can use it also for a particular function, not necessarily the whole kernel.

Closes #1267.


example/kernelSpecialization/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+#
```
sbastrakov, line 1: Hmm. I copied this from another example, and it seems everywhere in cmakelists of examples there is this emptyish starting line. Is it a typo that got copied around? @psychocoderHPC

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -0,0 +1,115 @@
...
+    //-----------------------------------------------------------------------------
+    //! Implementation for the general case
+    //!
+    //! It will be called when no specialization is a better fit.
+    template<typename TAcc>
+    ALPAKA_FN_ACC auto operator()(TAcc const& acc) const
+    {
+        // For simplicity assume 1d thread indexing
+        auto const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        if(!globalThreadIdx)
```
SimeonEhrig: Can you make it more explicit, that thread 0 is running the printf function? At the first time, when I read the code, I thought the if statement do some error checking because of the `!` operator.

sbastrakov: Agree, will change.

sbastrakov: Fixed.

example/kernelSpecialization/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
...
+cmake_minimum_required(VERSION 3.15)
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+################################################################################
+# Project.
+
+set(_TARGET_NAME kernelSpecialization)
+
+project(${_TARGET_NAME})
```
bernhardmgruber, line 31: I always wondered why we define our own variable, when we can just use `${PROJECT_NAME}` after the call to `project(...)`.

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -0,0 +1,115 @@
...
+
+//#############################################################################
+//! Kernel to illustrate specialization for a particular accelerator
+//!
+//! It has a general operator() implementation and a specialized one for the CUDA accelerator.
+//! When running the kernel on a CUDA device, the specialized version of operator() is called.
+//! Otherwise the general version is called.
+//! The same technique can be applied for any function called from inside the kernel,
+//! thus allowing specialization of only relevant part of the code.
+//! It can be useful for optimization or accessing specific functionality not abstracted by alpaka.
```
bernhardmgruber: Speaking of "specializing" is right from what we want to achieve, but the mechanism in C++ used to do so is not specialization but overloading. I wonder if we should rephrase the text in this regard.

```
//! It has a *generic* operator() implementation and an *overload* for the CUDA accelerator.
//! When running the kernel on a CUDA device, the *overload* of operator() with the *CUDA accelerator* is called.
//! Otherwise the *generic* version is called.
//! The same technique can be applied for any function called from inside the kernel,
//! thus allowing specialization of only relevant part of the code.
//! It can be useful for optimization or accessing specific functionality not abstracted by alpaka.
```

Does this sound better?

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -0,0 +1,115 @@
...
+struct Kernel
+{
+    //-----------------------------------------------------------------------------
+    //! Implementation for the general case
+    //!
+    //! It will be called when no specialization is a better fit.
```
bernhardmgruber: replace `specialization is a better fit` with `overload is a better match`

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -0,0 +1,115 @@
...
+    template<typename TAcc>
+    ALPAKA_FN_ACC auto operator()(TAcc const& acc) const
+    {
+        // For simplicity assume 1d thread indexing
+        auto const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        if(globalThreadIdx == 0u)
+            printf("Running the general kernel implementation\n");
+    }
+
+    //! Simple partial specialization for the CUDA accelerator
```
bernhardmgruber: This is unfortunately wrong, we have an overload here. A partial specialization would look like this:
```c++
ALPAKA_FN_ACC auto operator()<alpaka::AccGpuCudaRt<TDim, TIdx>>(alpaka::AccGpuCudaRt<TDim, TIdx> const& acc) const
```
However, partial specialization of members of class templates is forbidden :/

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -0,0 +1,115 @@
...
+    //!
+    //! We have to guard it with #ifdef as the types of alpaka accelerators are only conditionally available.
+    //! Specialization for other accelerators is similar, with another template name instead of AccGpuCudaRt.
```
bernhardmgruber: Maybe replace "Specialization" with "Overloading". But here I actually like the word, because specializing is what the user wants to.

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -0,0 +1,115 @@
...
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    template<typename TDim, typename TIdx>
+    ALPAKA_FN_ACC auto operator()(alpaka::AccGpuCudaRt<TDim, TIdx> const& acc) const
+    {
+        // This specialization is used when the kernel is run on the CUDA accelerator.
```
bernhardmgruber: overload

example/kernelSpecialization/CMakeLists.txt
sbastrakov: I don't really know, taken it from other examples. @psychocoderHPC do you know?

example/kernelSpecialization/src/kernelSpecialization.cpp
sbastrakov: You are right. I was actually mixing things up originally, and so some of the comment were left out of my original misinterpretation. Thanks for pointing out!

sbastrakov:
Thanks @bernhardmgruber for your comments, I agree with them and will do the suggested fixes. Do you think the "specializing" in the name of the example is fine, or it should also be "overloading"?

bernhardmgruber:
I think the term "specializing" is what the user wants to do. Overloading is just the vehicle for it. So I think "specializing" is fine for the example name and in some places in the documentation.

sbastrakov:
@bernhardmgruber I added a commit addressing your suggestions.

example/kernelSpecialization/src/kernelSpecialization.cpp
```diff
@@ -23,18 +23,24 @@
...
+//! It has a generic operator() implementation and an overload for the CUDA accelerator.
+//! When running the kernel on a CUDA device, the corresponding overload of operator() is called.
+//! Otherwise the generic version is called.
 //! The same technique can be applied for any function called from inside the kernel,
 //! thus allowing specialization of only relevant part of the code.
 //! It can be useful for optimization or accessing specific functionality not abstracted by alpaka.
+//!
+//! This kernel demonstrates the simplest way to achieve the effect by function overloading.
+//! Note that it does not perform specialization (in C++ template meaning) of function templates.
+//! We use word "specialization" as it represents a use case of having a special version for a particular accelerator.
```
bernhardmgruber: I think there is a *the* missing:

```suggestion
//! We use the word "specialization" as it represents a use case of having a special version for a particular accelerator.
```

bernhardmgruber (dismissed):
One minor fix and the rest looks good to me!

example/kernelSpecialization/src/kernelSpecialization.cpp
sbastrakov: Fixed, added it to the last commit to not have too many commits.

bernhardmgruber: Perfect!

bernhardmgruber (approved):

example/kernelSpecialization/CMakeLists.txt
psychocoderHPC: I do not know why it is there but if it is in all examples I am fine to keep it.
If we like to fix it I would suggest a separate PR to fix it in all examples/tests.

psychocoderHPC (approved):

example/kernelSpecialization/CMakeLists.txt
sbastrakov: Sure, if we fix it should be fixed everywhere. It's actually part of all alpaka cmake files, not just for examples. So I guess it's a formatting style then

example/kernelSpecialization/CMakeLists.txt
psychocoderHPC: I do not know. When you compile this example within alpaka is `${PROJECT_NAME}` then `alpaka` or the name of this example?

sbastrakov: Name of example.

Or wdym inside alpaka? It only supports out-of-source builds iirc.

PR #1273, opened 2021-03-14
Added warp::shfl functionality.

frobnitzem:
These changes implement calls to:
 * https://rocmdocs.amd.com/en/latest/Programming_Guides/HIP-GUIDE.html#warp-shuffle-functions
 * https://developer.nvidia.com/blog/using-cuda-warp-level-primitives

I only added __shfl for now, but there are also __shfl_up, etc. which could be cut/paste in a similar way. Apparently these were also needed by #18 .

### DIFF
--- a/include/alpaka/warp/Traits.hpp
+++ b/include/alpaka/warp/Traits.hpp
@@ -50,6 +50,11 @@ namespace alpaka
             struct Ballot;
 
             //#############################################################################
+            //! The shfl warp swizzling trait.
+            template<typename TWarp, typename TSfinae = void>
+            struct Shfl;
+
+            //#############################################################################
             //! The active mask trait.
             template<typename TWarp, typename TSfinae = void>
             struct Activemask;
@@ -149,6 +154,33 @@ namespace alpaka
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
             return traits::Ballot<ImplementationBase>::ballot(warp, predicate);
+        }
+
+        //-----------------------------------------------------------------------------
+        //! Broadcasts data from one thread to all members of the warp.
+        //! Similar to MPI_Bcast, but using srcLane instead of root.
+        //!
+        //! \tparam TWarp The warp implementation type.
+        //! \param  warp The warp implementation.
+        //! \param  value The value to broadcast (only meaningful from threadIdx == srcLane)
+        //! \param  srcLane The source lane sending value.
+        //! \return val from the thread index srcLane.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename TWarp>
+        ALPAKA_FN_ACC auto shfl(TWarp const& warp, int value, int srcLane)
+        {
+            using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
+            return traits::Shfl<ImplementationBase>::shfl(warp, value, srcLane);
+        }
+
+        //-----------------------------------------------------------------------------
+        //! shfl for float vals
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename TWarp>
+        ALPAKA_FN_ACC auto shfl(TWarp const& warp, float value, int srcLane)
+        {
+            using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
+            return traits::Shfl<ImplementationBase>::shfl(warp, value, srcLane);
         }
     } // namespace warp
 } // namespace alpaka
--- a/include/alpaka/warp/Traits.hpp
+++ b/include/alpaka/warp/Traits.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2020 Sergei Bastrakov
+/* Copyright 2020-2021 Sergei Bastrakov, David M. Rogers
  *
  * This file is part of Alpaka.
  *
@@ -157,30 +157,48 @@ namespace alpaka
         }
 
         //-----------------------------------------------------------------------------
-        //! Broadcasts data from one thread to all members of the warp.
-        //! Similar to MPI_Bcast, but using srcLane instead of root.
+        //! Exchange data between threads within a warp.
         //!
-        //! \tparam TWarp The warp implementation type.
-        //! \param  warp The warp implementation.
-        //! \param  value The value to broadcast (only meaningful from threadIdx == srcLane)
-        //! \param  srcLane The source lane sending value.
+        //! Effectively executes:
+        //!
+        //!     __shared__ int32_t values[warpsize];
+        //!     values[threadIdx.x] = value;
+        //!     __syncthreads();
+        //!     return values[(srcLane + width*floor(threadIdx.x/width))%width];
+        //!
+        //! However, it does not use shared memory.
+        //!
+        //! Notes:
+        //! * The programmer must ensure that all threads calling this
+        //!   function (and the srcLane) are executing the same line of code.
+        //!   In particular it is not portable to write if(a) {shfl} else {shfl}.
+        //!
+        //! * Commonly used with width = warpsize (the default), (returns values[srcLane])
+        //!
+        //! * Width must be a power of 2.
+        //!
+        //! \tparam TWarp   warp implementation type
+        //! \param  warp    warp implementation
+        //! \param  value   value to broadcast (only meaningful from threadIdx == srcLane)
+        //! \param  srcLane source lane sending value
+        //! \param  width   number of threads receiving a single value
         //! \return val from the thread index srcLane.
         ALPAKA_NO_HOST_ACC_WARNING
         template<typename TWarp>
-        ALPAKA_FN_ACC auto shfl(TWarp const& warp, int value, int srcLane)
+        ALPAKA_FN_ACC auto shfl(TWarp const& warp, std::int32_t value, std::int32_t srcLane, std::int32_t width = 0)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
-            return traits::Shfl<ImplementationBase>::shfl(warp, value, srcLane);
+            return traits::Shfl<ImplementationBase>::shfl(warp, value, srcLane, width ? width : getSize(warp));
         }
 
         //-----------------------------------------------------------------------------
         //! shfl for float vals
         ALPAKA_NO_HOST_ACC_WARNING
         template<typename TWarp>
-        ALPAKA_FN_ACC auto shfl(TWarp const& warp, float value, int srcLane)
+        ALPAKA_FN_ACC auto shfl(TWarp const& warp, float value, std::int32_t srcLane, std::int32_t width = 0)
         {
             using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
-            return traits::Shfl<ImplementationBase>::shfl(warp, value, srcLane);
+            return traits::Shfl<ImplementationBase>::shfl(warp, value, srcLane, width ? width : getSize(warp));
         }
     } // namespace warp
 } // namespace alpaka
--- a/include/alpaka/warp/WarpSingleThread.hpp
+++ b/include/alpaka/warp/WarpSingleThread.hpp
@@ -92,6 +92,22 @@ namespace alpaka
                     return predicate ? 1u : 0u;
                 }
             };
+
+            //#################################################################
+            template<>
+            struct Shfl<WarpSingleThread>
+            {
+                //-------------------------------------------------------------
+                static auto shfl(warp::WarpSingleThread const& /*warp*/, int val, int srcLane)
+                {
+                    return val;
+                }
+                //-------------------------------------------------------------
+                static auto shfl(warp::WarpSingleThread const& /*warp*/, float val, int srcLane)
+                {
+                    return val;
+                }
+            };
         } // namespace traits
     } // namespace warp
 } // namespace alpaka
--- a/include/alpaka/warp/WarpSingleThread.hpp
+++ b/include/alpaka/warp/WarpSingleThread.hpp
@@ -98,12 +98,12 @@ namespace alpaka
             struct Shfl<WarpSingleThread>
             {
                 //-------------------------------------------------------------
-                static auto shfl(warp::WarpSingleThread const& /*warp*/, int val, int srcLane)
+                static auto shfl(warp::WarpSingleThread const& /*warp*/, int val, int /*srcLane*/)
                 {
                     return val;
                 }
                 //-------------------------------------------------------------
-                static auto shfl(warp::WarpSingleThread const& /*warp*/, float val, int srcLane)
+                static auto shfl(warp::WarpSingleThread const& /*warp*/, float val, int /*srcLane*/)
                 {
                     return val;
                 }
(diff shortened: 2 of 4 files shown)


include/alpaka/warp/WarpSingleThread.hpp
```diff
@@ -92,6 +92,22 @@ namespace alpaka
                     return predicate ? 1u : 0u;
                 }
             };
+
+            //#################################################################
+            template<>
+            struct Shfl<WarpSingleThread>
+            {
+                //-------------------------------------------------------------
```
bernhardmgruber, line 100: I was never a fan of these section comments, because I think they just add noise. But everybody else is free to convince me ;) I know we have a lot of these in alpaka.

include/alpaka/warp/WarpSingleThread.hpp
```diff
@@ -92,6 +92,22 @@ namespace alpaka
...
+                static auto shfl(warp::WarpSingleThread const& /*warp*/, int val, int /*srcLane*/)
```
bernhardmgruber: I have seen unused argments in alpaka usually named and cast to `void` inside the function body. I guess the argument was better documentation.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -142,6 +142,31 @@ namespace alpaka
...
+                    -> float
+                {
+#    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+                    return __shfl_sync(activemask(warp), val, srcLane, getSize(warp));
+#    else
+                    return __shfl(val, srcLane, getSize(warp));
+#    endif
+                }
+                //-------------------------------------------------------------
+                __device__ static auto shfl(warp::WarpUniformCudaHipBuiltIn const& warp, int val, int srcLane) -> int
```
bernhardmgruber: Here you specify an explicit return value `int`, which you did not do for the other `shfl` overload. I guess this was an oversight. The other overload returns `float`, does it? Could you add the return type please?

test/unit/warp/src/Shfl.cpp
```diff
@@ -0,0 +1,110 @@
...
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TAcc>
+    ALPAKA_FN_ACC auto operator()(TAcc const& acc, bool* success) const -> void
+    {
+        std::int32_t const warpExtent = alpaka::warp::getSize(acc);
+        ALPAKA_CHECK(*success, warpExtent == 1);
+
+        ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 12, 0) == 12);
+        ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 42, -1) == 42);
+        // ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 3.3f, 0) == 3.3f);
```
bernhardmgruber: Why is this test disabled?

include/alpaka/warp/WarpSingleThread.hpp
sbastrakov: I think we only mandatory do it when the arguments are conditionally used (e.g. inside an `#ifdef` guard), to avoid a warning when the condition is not met. This one is probably fine.

include/alpaka/warp/WarpSingleThread.hpp
sbastrakov: Me neither. @j-stephan maybe we can add this quesion to the meeting points to discuss it?

Of course, this PR just follows the prevailing style.

test/unit/warp/src/Shfl.cpp
```diff
@@ -0,0 +1,110 @@
+/* Copyright 2021 David M. Rogers
```
sbastrakov, line 1: Please also add yourself and 2021 to the other modified files

include/alpaka/warp/WarpSingleThread.hpp
j-stephan: Yes, we'll talk about it tomorrow. For now these should be kept to match the rest of alpaka.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
sbastrakov: I think `float` is specified there, just formatted in another way

sbastrakov (commented):
Thanks for the PR!

Looking at the HIP and CUDA documentation, they also have the last parameter `width` which is by default `= warpSize`. What is the reason for omitting it from your API? If you need access to this value, there is `getSize()` in `Traits.hpp`.

One thing that we discussed quite a bit when bringing in the support for warp functions #1003 (and also intrinsics #1004, #1018 ) was to go for fix-width types vs. just `int`. Here it's always a bit awkward as alpaka tries using fixed-width everywhere, but the backend APIs like CUDA operate with `int`s. @psychocoderHPC what do you think?

test/unit/warp/src/Shfl.cpp
frobnitzem: This was an oversight.  Unrelated, note that this thread discusses shfl with invalid srcLane: https://github.com/ROCm-Developer-Tools/HIP-CPU/issues/14

include/alpaka/warp/WarpSingleThread.hpp
frobnitzem: I'm being consistent with Ballot.

frobnitzem:
> Thanks for the PR!
> 
> Looking at the HIP and CUDA documentation, they also have the last parameter `width` which is by default `= warpSize`. What is the reason for omitting it from your API? If you need access to this value, there is `getSize()` in `Traits.hpp`.
> 
> One thing that we discussed quite a bit when bringing in the support for warp functions #1003 (and also intrinsics #1004, #1018 ) was to go for fix-width types vs. just `int`. Here it's always a bit awkward as alpaka tries using fixed-width everywhere, but the backend APIs like CUDA operate with `int`s. @psychocoderHPC what do you think?

These can be set to int32_t, since the CUDA documentation specifies "Template type T" with 4 or 8-byte values, while HIP says int or float. Templates seem to give awful errors from inputs other than int or float, however, since the casting is not unique.

With regard to warpsize, I had not thought about using that feature, but it is possible to use groups of consecutive threads smaller than the warpsize as "mini-warps", so I'll add it.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
bernhardmgruber: Oh yeah, sorry! :D

psychocoderHPC:
> One thing that we discussed quite a bit when bringing in the support for warp functions #1003 (and also intrinsics #1004, #1018 ) was to go for fix-width types vs. just `int`. Here it's always a bit awkward as alpaka tries using fixed-width everywhere, but the backend APIs like CUDA operate with `int`s. @psychocoderHPC what do you think?

Yes please expose always fixed with types to the user. If needed please cast the input values to whatever the native API function call requires. Fixed with types are required to guarentee portability between different compiler and platforms.

psychocoderHPC:
> With regard to warpsize, I had not thought about using that feature, but it is possible to use groups of consecutive threads smaller than the warpsize as "mini-warps", so I'll add it.

Yes CUDA allows shuffling between less than warp size threads. The problem in apaka would be that our CPU backends has a warpsize of one. One case where it is useful to shuffle between 4 lanes is if you run some kind of emulated sse4 instructions on the GPU e.g. for cryptographic algorithms. This would not be portable to alpaka CPU backends, therefore I would maybe restrict alpaka to support for shuffling in a full warp only.

include/alpaka/warp/Traits.hpp
```diff
@@ -150,5 +155,32 @@ namespace alpaka
             using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
             return traits::Ballot<ImplementationBase>::ballot(warp, predicate);
         }
+
+        //-----------------------------------------------------------------------------
+        //! Broadcasts data from one thread to all members of the warp.
+        //! Similar to MPI_Bcast, but using srcLane instead of root.
+        //!
```
psychocoderHPC, line 161: IMO we need to add to the documentation that this function `shfl` is collective what means all threads need to call the function and also from the same code branch.
The reason is that for CUDA the implementation is using `activemask` and for HIP all threads in a warp needs to call the function. Using `activemask` means if threads from the if and else branch call the function they will not see each other.

frobnitzem: I updated these docs to include this warning.

test/unit/warp/src/Shfl.cpp
bernhardmgruber: I think this is done.

include/alpaka/warp/WarpSingleThread.hpp
bernhardmgruber: We discussed to remove those section comments in the future, but you can keep them in this PR.

test/unit/warp/src/Shfl.cpp
```diff
@@ -0,0 +1,124 @@
...
+        for(int width = 1; width < warpExtent; width *= 2)
+        {
+            for(int idx = 0; idx < width; idx++)
+            {
+                int off = width * (threadIdxInWarp / width);
+                ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, threadIdxInWarp, idx, width) == idx + off);
+                ALPAKA_CHECK(
+                    *success,
+                    alpaka::warp::shfl(acc, 4.0f - float(threadIdxInWarp), idx, width) == 4.0f - float(idx + off));
+            }
```
bernhardmgruber: We have a misscompilation here because the warning on a floating point comparison is turned into an error. Despite the error/warning, I believe this should work though in strict floating-point modes ;) But since alpaka compiles with fast-math enabled, should we do an epsilon comparison here?

include/alpaka/warp/Traits.hpp
sbastrakov: I think I forgot to add a similar warning to the previously existing warp collectives. You comment also alllies to those, right @psychocoderHPC ?

test/unit/warp/src/Shfl.cpp
sbastrakov: FYI we build with all warnings on and all warnings are errors. So it's a pain sometimes to make all compilers happy.

If you need approximate floating-point comparisoin, there is `Approx` from catch2, we use it in some other alpaka unit tests. But in this case I guess it should be exact actually?

frobnitzem: It should be exact, since the computation is done the same way, just on different threads. I could cast to uint32_t and compare that instead, but who knows if the compiler will complain about the cast.

test/unit/warp/src/Shfl.cpp
```diff
@@ -0,0 +1,130 @@
...
+    ALPAKA_FN_ACC auto operator()(TAcc const& acc, bool* success) const -> void
+    {
+        std::int32_t const warpExtent = alpaka::warp::getSize(acc);
+        ALPAKA_CHECK(*success, warpExtent == 1);
+
+        ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 12, 0) == 12);
+        ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 42, -1) == 42);
+        float ans = alpaka::warp::shfl(acc, 3.3f, 0);
+        float expect = 3.3;
+        ALPAKA_CHECK(*success, CAST(ans) == CAST(expect));
```
bernhardmgruber: While that may work on most compilers, it is certainly undefined behavior because we are reading a float value as an int32. At that point a `memcmp` is probably the more correct solution.

However, I think I prefer disabling the warning here. I think we know what we are doing here and we expect the exact same float to be shuffled around, so the comparison should not need an epsilon. 

```c++
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wfloat-equal"
ALPAKA_CHECK(*success, ans == expect);
#pragma GCC diagnostic pop
```

I don't know if we need a similar escape hatch for the other compilers.

psychocoderHPC: Compiler specific pragmas should be guarded: see https://github.com/alpaka-group/alpaka/blob/a78a70adb27c5285fd8752f96c11db1487a20906/include/alpaka/atomic/Op.hpp#L31-L34

test/unit/warp/src/Shfl.cpp
```diff
@@ -0,0 +1,127 @@
...
+    template<typename TAcc>
+    ALPAKA_FN_ACC auto operator()(TAcc const& acc, bool* success) const -> void
+    {
+        std::int32_t const warpExtent = alpaka::warp::getSize(acc);
+        ALPAKA_CHECK(*success, warpExtent == 1);
+
+        ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 12, 0) == 12);
+        ALPAKA_CHECK(*success, alpaka::warp::shfl(acc, 42, -1) == 42);
+        float ans = alpaka::warp::shfl(acc, 3.3f, 0);
+        ALPAKA_CHECK(*success, alpaka::math::abs(acc, ans - 3.3f) < 1e-8f);
```
bernhardmgruber, line 35: Not the solution I had in mind, but LGTM!

bernhardmgruber (dismissed):

include/alpaka/warp/Traits.hpp
psychocoderHPC: @sbastrakov Yes this should be added to other warp functions too. Currently, only CUDA allows calling warp functions from different branches. It is fine if all threads of the warp are in the same branch but as soon as the threads diverge the behavior is undefined (for HIP and CUDA devices before sm_70) .

test/unit/warp/src/Shfl.cpp
psychocoderHPC: Could we define a misc function in [this folder](https://github.com/alpaka-group/alpaka/tree/develop/test/common/include/alpaka/test) and use the function here?.

```C++
ALPAKA_FN_INLINE bool floatEqual(TAcc const & acc, T const & v0, T const & v1, T const epsilon = std::numeric_limits<T>::epsilon() )
{
  return alpaka::math::abs(acc, v0 - v1) < episilon;
}
```

frobnitzem:
It seems one of the tests spuriously failed on a network error.

psychocoderHPC:
> It seems one of the tests spuriously failed on a network error.

We mostly need to ignore these tests. Github is not providing us with a feature to restart only a single job :-(

sbastrakov:
So should we merge this one then?

psychocoderHPC:
@frobnitzem Could you please address https://github.com/alpaka-group/alpaka/pull/1273#discussion_r597785985 IMO it should be a separate function else it looks like a line with many magic numbers.

psychocoderHPC (changes requested):
I would like to block this PR until https://github.com/alpaka-group/alpaka/pull/1273#issuecomment-805637599 is implemented.

test/unit/warp/src/Shfl.cpp
psychocoderHPC: You could also use `std::numeric_limits<float>::epsilon()` directly instead of the magic number `1.e-8`
This woild be simpler then including the helper function.

psychocoderHPC: I pushed a fix to this PR which used `std::numeric_limits::epsilon()`

psychocoderHPC (approved):

bernhardmgruber (approved):

BenjaminW3 (approved):

PR #1275, opened 2021-03-17
Remove section comments

bernhardmgruber:
As discussed during the last alpaka VC, this PR is just a mechanical removal of section comments of the patterns:

```c++
//-----------------------------------------------------------------------------
//#############################################################################
```

I did a bit of manual formatting on top.

PR #1279, opened 2021-03-26
Rework implementation of OpenMP schedule support

### LINKED ISSUE #1255: Implementation of user-defined scheduling for OpenMP should be changed
The current implementation has several flaws:

* It doesn't work, if you are already inside a forked region. Eg. QueueCpuOmp2CollectiveImpl won't work as expected. Some OpenMP implementations might work nevertheless., but it is not backed up by the standard.
* It inflicts performance issues:
  * There is always the need to request the runtime schedule. This requires some loads, tests and branches, which esp. with a lot of small loops is undesirable.
  * Especially with static scheduling a lot of compile-time optimization potential is wasted.

The scheduler should be statically dispatched. That is, there have to be several OpenMP for loops in alpaka (one for each schedule kind and one without anything) and then a template magic decides, which loop is compiled. After all, the schedule kind of a loop is a static property almost always (TBH I can't think of any reasonable special case).

sbastrakov:
As discussed in #1255, the previous implementation always used schedule(runtime) and called appropriate `omp_set_schedule()`.
This could have been suboptimal for two reasons.
First, it did not give control while inside the parallel section.
Second, it may have hindered some optimizations made by compilers.

The new implementation performs explicit complile-time dispatch into appropriate `schedule()` clauses.
This solves the first shortcoming and hopefully helps the second.

In terms of the user interface, the only change is that when defining as a member in kernel, it is no longer
`ompSchedule` of type `alpaka::omp::Schedule`, but two separate members `ompScheduleKind` and `ompScheduleChunkSize`.
The trait interface is unchanged and still has a higher priority.
See the changes in example and tests.

It is a bit ugly due to combination of copy-pasted implementations for compilte-time dispatch and existing ifdef-guards that were inside and got cloned now. I actually tried a lot to limit the ugliness, but that's how it is now.

@krzikalla is it possible that you test this branch for your codebase?

### DIFF
--- a/include/alpaka/core/OmpSchedule.hpp
+++ b/include/alpaka/core/OmpSchedule.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2020 Sergei Bastrakov
+/* Copyright 2020-2021 Sergei Bastrakov
  *
  * This file is part of Alpaka.
  *
@@ -26,33 +26,33 @@ namespace alpaka
         //! whether OpenMP is enabled.
         struct Schedule
         {
-            //! Special schedule setting which does not change the internal control variables of OpenMP
-            constexpr static std::uint32_t NoSchedule = 0u;
-            //! Integers corresponding to the mandatory OpenMP omp_sched_t enum values as of version 5.0.
-            constexpr static std::uint32_t Static = 1u;
-            constexpr static std::uint32_t Dynamic = 2u;
-            constexpr static std::uint32_t Guided = 3u;
-            constexpr static std::uint32_t Auto = 4u;
-            //! Each schedule value can be combined with monotonic using + or |.
-            constexpr static std::uint32_t Monotonic = 0x80000000u;
+            //! Schedule kinds corresponding to arguments of OpenMP schedule clause
+            //!
+            //! Kinds also present in omp_sched_t enum have the same integer values.
+            //! It is enum, not enum class, for shorter usage as omp::Schedule::[kind] and to keep interface of 0.6.0.
+            enum Kind
+            {
+                // Corresponds to not setting schedule
+                NoSchedule,
+                Static = 1u,
+                Dynamic = 2u,
+                Guided = 3u,
+                // Auto supported since OpenMP 3.0
+#if defined _OPENMP && _OPENMP >= 200805
+                Auto = 4u,
+#endif
+                Runtime
+            };
 
             //! Schedule kind.
-            //!
-            //! We cannot simply use type omp_sched_t since this struct is agnostic of OpenMP. We also cannot create an
-            //! own mirror enum, since OpenMP implementations are allowed to extend the range of values beyond the
-            //! standard ones defined above. So we have to accept and store any uint32_t value, and for non-standard
-            //! values a user has to ensure the underlying implementation supports it.
-            std::uint32_t kind;
+            Kind kind;
 
             //! Chunk size. Same as in OpenMP, value 0 corresponds to default chunk size. Using int and not a
             //! fixed-width type to match OpenMP API.
             int chunkSize;
 
-            //! The provided value myKind has to be supported by the underlying OpenMP implementation.
-            //! It does not have to be one of the constants defined above.
-            //! A default-constructed schedule does not affect internal control variables of OpenMP.
-            //! The constructor is constexpr to simplify creation of static constexpr ompSchedule in user code.
-            ALPAKA_FN_HOST constexpr Schedule(std::uint32_t myKind = NoSchedule, int myChunkSize = 0)
+            //! Create a schedule with the given kind and chunk size
+            ALPAKA_FN_HOST constexpr Schedule(Kind myKind = NoSchedule, int myChunkSize = 0)
                 : kind(myKind)
                 , chunkSize(myChunkSize)
             {
@@ -72,7 +72,7 @@ namespace alpaka
             omp_sched_t ompKind;
             int chunkSize = 0;
             omp_get_schedule(&ompKind, &chunkSize);
-            return Schedule{static_cast<std::uint32_t>(ompKind), chunkSize};
+            return Schedule{static_cast<Schedule::Kind>(ompKind), chunkSize};
 #else
             return Schedule{};
 #endif
@@ -86,7 +86,7 @@ namespace alpaka
         //! Note that calling from inside a parallel region does not have an immediate effect.
         ALPAKA_FN_HOST inline void setSchedule(Schedule schedule)
         {
-            if(schedule.kind != Schedule::NoSchedule)
+            if((schedule.kind != Schedule::NoSchedule) && (schedule.kind != Schedule::Runtime))
             {
 #if defined _OPENMP && _OPENMP >= 200805
                 omp_set_schedule(static_cast<omp_sched_t>(schedule.kind), schedule.chunkSize);
--- a/include/alpaka/core/OmpSchedule.hpp
+++ b/include/alpaka/core/OmpSchedule.hpp
@@ -37,11 +37,11 @@ namespace alpaka
                 Static = 1u,
                 Dynamic = 2u,
                 Guided = 3u,
-                // Auto supported since OpenMP 3.0
+            // Auto supported since OpenMP 3.0
 #if defined _OPENMP && _OPENMP >= 200805
                 Auto = 4u,
 #endif
-                Runtime
+                Runtime = 5u
             };
 
             //! Schedule kind.
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -98,14 +98,25 @@ namespace alpaka
             }
         };
 
+        //! Helper trait to check if TSchedule is not a type originating from OmpSchedule trait definition
+        //!
+        //! Is well-formed for those types (not defining the trait), ill-formed otherwise.
+        //! When well-formed, result in a non-void type to not collide with void-based SFINAE also used.
+        //!
+        //! \tparam TSchedule The schedule type.
+        template<typename TSchedule>
+        using HasOmpScheduleNotDefined = std::enable_if_t<!std::is_same<TSchedule, omp::Schedule>::value, char>;
+
         //! Executor of parallel OpenMP loop with the static schedule
         //!
         //! The default implementation is for the kernels that do not set chunk size in any way.
         //!
+        //! Note: HasOmpScheduleNotDefined<TSchedule> should not be necessary, but is required by some old compilers.
+        //!
         //! \tparam TKernel The kernel type.
         //! \tparam TSchedule The schedule type (not necessarily omp::Schedule).
-        template<typename TKernel, typename TSchedule, typename TSfinae>
-        struct ParallelForImpl<TKernel, TSchedule, omp::Schedule::Static, TSfinae>
+        template<typename TKernel, typename TSchedule>
+        struct ParallelForImpl<TKernel, TSchedule, omp::Schedule::Static, HasOmpScheduleNotDefined<TSchedule>>
         {
             //! Run parallel OpenMP loop
             //!
@@ -176,13 +187,19 @@ namespace alpaka
             }
         };
 
-        //! Helper trait to check if TKernel has an public (not necessarily static) member ompScheduleChunkSize
+        //! Helper trait to check if TKernel has not defined OmpSchedule trait and has a public (not necessarily
+        //! static) member ompScheduleChunkSize
         //!
         //! Is well-formed for those types, ill-formed otherwise.
         //!
+        //! Note: HasOmpScheduleNotDefined<TSchedule> part should not be necessary, but is required by some old
+        //! compilers.
+        //!
         //! \tparam TKernel The kernel type.
-        template<typename TKernel>
-        using HasScheduleChunkSize = meta::Void<decltype(std::declval<TKernel&>().ompScheduleChunkSize)>;
+        //! \tparam TSchedule The schedule type.
+        template<typename TKernel, typename TSchedule>
+        using HasScheduleChunkSize
+            = meta::Void<decltype(std::declval<TKernel&>().ompScheduleChunkSize), HasOmpScheduleNotDefined<TSchedule>>;
 
         //! Executor of parallel OpenMP loop with the static schedule
         //!
@@ -191,7 +208,7 @@ namespace alpaka
         //! \tparam TKernel The kernel type.
         //! \tparam TSchedule The schedule type (not necessarily omp::Schedule).
         template<typename TKernel, typename TSchedule>
-        struct ParallelForImpl<TKernel, TSchedule, omp::Schedule::Static, HasScheduleChunkSize<TKernel>>
+        struct ParallelForImpl<TKernel, TSchedule, omp::Schedule::Static, HasScheduleChunkSize<TKernel, TSchedule>>
         {
             //! Run parallel OpenMP loop
             //!
@@ -228,10 +245,11 @@ namespace alpaka
         //!
         //! The default implementation is for the kernels that do not set chunk size in any way.
         //!
+        //! Note: HasOmpScheduleNotDefined<TSchedule> should not be necessary, but is required by some old compilers.
         //! \tparam TKernel The kernel type.
         //! \tparam TSchedule The schedule type (not necessarily omp::Schedule).
-        template<typename TKernel, typename TSchedule, typename TSfinae>
-        struct ParallelForImpl<TKernel, TSchedule, omp::Schedule::Dynamic, TSfinae>
+        template<typename TKernel, typename TSchedule>
+        struct ParallelForImpl<TKernel, TSchedule, omp::Schedule::Dynamic, HasOmpScheduleNotDefined<TSchedule>>
         {
             //! Run parallel OpenMP loop
             //!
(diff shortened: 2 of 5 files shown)


sbastrakov:
**Note to reviewers**. It is probably easiest to read the big bulk of changes in `TaskKernelCpuOmp2Blocks.hpp` starting from the end of the `namespace detail` there and go upwards. There are comments explaining the dispatch process.

When looking at the "leafs" of dispatch in specializations of `ParallelForImpl` keep in mind there are only two cases. One is only schedule kind matters, but not chunk, it is the case for `NoSchedule`, `Auto`, `Runtime`. There only one specialization is sufficient. The second, more complicated, case is `Static`, `Dynamic`, `Guided`. Because of 3 ways of setting chunk size, they all have 3 specializations each.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,694 @@
...
+            //! \param loopBody The loop body functor instance, takes iteration index as input.
+            //! \param numIterations The number of loop iterations.
+            template<typename TLoopBody, typename TIdx>
+            ALPAKA_FN_HOST void operator()(
+                TKernel const& kernel,
+                TLoopBody&& loopBody,
+                TIdx const numIterations,
+                TSchedule const&)
+            {
+/// TODO remove debug output_iterator
```
BenjaminW3: TODO remove debug output_iterator

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,694 @@
...
+                    break;
+#    endif
+                case omp::Schedule::Runtime:
+                    ParallelForImpl<TKernel, omp::Schedule, omp::Schedule::Runtime>{}(
+                        kernel,
+                        std::forward<TLoopBody>(loopBody),
+                        numIterations,
+                        schedule);
+                    break;
+                    /// TODO: what do we do by default? Should not happen, but may due to copy-paste
```
BenjaminW3: Please remove the comment. Compiler should find those cases when the kind member is an `enum class`

include/alpaka/core/OmpSchedule.hpp
```diff
@@ -27,33 +27,33 @@ namespace alpaka
...
-            constexpr static std::uint32_t Dynamic = 2u;
-            constexpr static std::uint32_t Guided = 3u;
-            constexpr static std::uint32_t Auto = 4u;
-            //! Each schedule value can be combined with monotonic using + or |.
-            constexpr static std::uint32_t Monotonic = 0x80000000u;
+            //! Schedule kinds corresponding to arguments of OpenMP schedule clause
+            //!
+            //! Kinds also present in omp_sched_t enum have the same integer values.
+            //! It is enum, not enum class, for shorter usage as omp::Schedule::[kind]
+            enum Kind
```
BenjaminW3, line 33: please use `enum class`

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: Ah thanks

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: It is not enum class, as explained at its definition

include/alpaka/core/OmpSchedule.hpp
sbastrakov: I think it makes usage more wordy in this case, and also will break the 0.6.0 interface

BenjaminW3: Ah, OK. Not breaking the interface is a good reason

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,694 @@
...
+                        schedule);
+                    break;
+#    endif
+                case omp::Schedule::Runtime:
+                    ParallelForImpl<TKernel, omp::Schedule, omp::Schedule::Runtime>{}(
+                        kernel,
+                        std::forward<TLoopBody>(loopBody),
+                        numIterations,
+                        schedule);
+                    break;
```
bernhardmgruber: The implementation with all the `ParallelForImpl` specializations feels very repretitive. I think I would inline all the `ParallelForImpl` instances into the `switch` statement and declare the loop induction variables outside the `switch` statement. That gets rid of all the specializations, comments, extra argument lists, documentation etc.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,694 @@
...
+                TKernel const& kernel,
+                TLoopBody&& loopBody,
+                TIdx const numIterations,
+                TSchedule const&)
+            {
+/// TODO remove debug output_iterator
+#    pragma omp single
+                {
+                    std::cout << "ParallelForImpl for Static schedule, chunk size from member\n";
+                }
```
bernhardmgruber: Debug output?

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,694 @@
...
```
bernhardmgruber: Debug output?

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
psychocoderHPC: What do you mean by `inline`? Do you suggest moving the loop implementation to the switch?

psychocoderHPC: If `ParallelForImpl` would contain a static method it is also possible to create an array with function pointers to the implementation and than call 
```
functionPtrArray[omp::Schedule::Runtime](kernel,
                        std::forward<TLoopBody>(loopBody),
                        numIterations,
                        schedule);
``` 

instead of using a switch and repeating the code to call the implementation.

bernhardmgruber: Yes. I believe this could dramatically shorten the code.

test/unit/kernel/src/KernelWithOmpSchedule.cpp
```diff
@@ -28,43 +25,47 @@ struct KernelWithOmpScheduleBase
...
-struct KernelWithMemberOmpSchedule : KernelWithOmpScheduleBase
+// Kernel that sets the schedule chunk size via non-constexpr ompScheduleChunkSize.
+struct KernelWithStaticMemberOmpScheduleChunkSize : KernelWithOmpScheduleBase
 {
-    static const alpaka::omp::Schedule ompSchedule;
+    static const int ompScheduleChiunkSize;
 };
 // In this case, the member has to be defined externally
-const alpaka::omp::Schedule KernelWithMemberOmpSchedule::ompSchedule = expectedSchedule;
+const int KernelWithStaticMemberOmpScheduleChunkSize::ompScheduleChiunkSize = 2;
```
psychocoderHPC: type `ompScheduleChiunkSize` should be `ompScheduleChunkSize`

This typo is also in other parts of this diff.

psychocoderHPC (changes requested):

test/unit/kernel/src/KernelWithOmpSchedule.cpp
```diff
@@ -76,7 +77,7 @@ struct KernelWithTraitOmpSchedule : KernelWithOmpScheduleBase
 struct KernelWithMemberAndTraitOmpSchedule : KernelWithOmpScheduleBase
 {
     // Set to be different from expected so that it this is used the test would fail
```
krzikalla: it -> if

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,694 @@
...
+            //! \param loopBody The loop body functor instance, takes iteration index as input.
+            //! \param numIterations The number of loop iterations.
+            template<typename TLoopBody, typename TIdx>
+            ALPAKA_FN_HOST void operator()(
+                TKernel const&,
+                TLoopBody&& loopBody,
+                TIdx const numIterations,
+                TSchedule const&)
+            {
+#    if _OPENMP < 200805 // For OpenMP < 3.0 you have to declare the loop index (a signed integer) outside of the loop
```
krzikalla, line 85: I would also suggest to remove the #if's here and go with the OPENMP<200805 branch only.
In addition you could create two macros, which take the schedule kind as an argument. One macro without chunksize and one with chunksize. Then the impls get one liner (somewhat difficult to read and debug, but this is alpaka, there is a lot difficult to read and debug). That's up to the alpaka team.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
krzikalla: Inlining the ParallelForImpl won't work, as you need them also for the static dispatch. As far as I understand, this switch-case code here will be rarely used - normally you use the static dispatch. Nevertheless I''d not use function pointers, since it makes it more difficult for the compiler to optimize.

sbastrakov: I don't understands your proposal @bernhardmgruber @psychocoderHPC , could you elaborate?

This part is used only when schedule information is set "run-time" via trait specialization. So wrt optimization I don't think it matters much here. The optimized-for-schedule versions are `ParallelForImpl`  ones, they are fully specialized wrt all supported schedules and will be used directly, without `switch`, in case there is no trait (so either use the member or no schedule)

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: Yes, I forgot to remove some of it.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: > I would also suggest to remove the #if's here and go with the OPENMP<200805 branch only.

In terms of shortening the code, which is also now copied 10 times I agree. However this may be against the spirit of optimization, which is the only reason for this PR? The other branch allows using an actual data type and not wrapping it into the definitely wide enough type.

With the macros I actually had the same idea initially, but then decided it's not worth it. But ultimately i think it does not make a big difference here: not much simplification, not much less readability.

test/unit/kernel/src/KernelWithOmpSchedule.cpp
sbastrakov: And in addition to it, this comment was outdated at all, will remove

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
sbastrakov: @psychocoderHPC what do you think about the suggestion to remove `if`-guards?

BenjaminW3: I would keep them 😉

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
bernhardmgruber: Oh sorry, I did not see the second callsite to `ParallelForImpl`. The code is fine then.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
psychocoderHPC: I am not sure, keep it for now and we should discuss this in the alpaka dev meeting. Maybe can find a more elegant way.
I would like to drop `200805` but I heard that Microsoft is only providing OMP2.

psychocoderHPC: maybe we should add a macro as @krzikalla suggested

```C++
#    if _OPENMP < 200805 // For OpenMP < 3.0 you have to declare the loop index (a signed integer) outside of the loop
                         // header.
#define ALPAKA_OMP_LOOP(iterations, ...)  \
                std::intmax_t iNumBlocksInGrid(static_cast<std::intmax_t>(iterations)); \
                std::intmax_t i; \
                __pragma(omp for nowait __VA_ARGS__) \
                for(i = 0; i < iNumBlocksInGrid; ++i) 
#    else
#define ALPAKA_OMP_LOOP(iterations, ...)  \
                __pragma(omp for nowait __VA_ARGS__) \
                for(TIdx i = 0; i < iterations; ++i) \
#    endif
```

But I would do it in a separate PR.

BenjaminW3: Microsoft has started adding support for OpenMP 3.0 see here: https://devblogs.microsoft.com/cppblog/improved-openmp-support-for-cpp-in-visual-studio/

As we only support the latest greatest Visual Studio version we could try to use the new command line switch to enable it.

bernhardmgruber: MS is in a mix of OpenMP 2.0 with some OpenMP 3.0 features. They also allow using the LLVM OpenMP runtime now. It definitely looks like there is some movement recently: https://devblogs.microsoft.com/cppblog/improved-openmp-support-for-cpp-in-visual-studio/
But for the time being, it's still OpenMP 2.0.

psychocoderHPC:
@sbastrakov could you please rebase this PR, I merged #1275 to remove the section comments.

Please check the formating of your PR too, the CI was showing formatting issues.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -830,12 +830,12 @@ namespace alpaka
 
             // Body of the OpenMP parallel loop to be executed.
             // Index type is auto since we have a difference for OpenMP 2.0 and later ones
-            auto loopBody = [&](auto i) {
+            auto loopBody = [&](auto currentIndex) {
```
sbastrakov, line 881: Note: this is due to `i` shadowing warning in some compilers, as `i` is also used as an index of `for` loops calling this lambda. I would imagine it's a false alarm tho.

sbastrakov:
So for the Intel compiler there is some weird error, I guess on the compiler side
```
/home/runner/work/alpaka/alpaka/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp(96): internal error: assertion failed: find_assoc_pragma: pragma not found (il.c, line 25380 in find_assoc_pragma)

                      loopBody(i);
```

sbastrakov:
It seems gcc 5 and 6 sometimes get confused with template substitution priority. I believe the code is correct, and it's the fault of these compilers. However I think i've found a not too bad fix by adding some conditions explicitly.

psychocoderHPC (changes requested):
From the code side, it looks good :star: , @sbastrakov could you work on the CI issues.

sbastrakov:
@psychocoderHPC yes, on it. I just love how all compilers interpret complex template matching differently.

sbastrakov:
Okay, so the template issues are finally resolved.

Now there are two CI issues remaining:

- Intel compiler version 20 seems to be unable to work with labmdas called inside OpenMP regions. It is both in CI which uses that version and on Hemera only on version 20 (works fine with earlier ones). Also seems to be same as [here](https://community.intel.com/t5/Intel-C-Compiler/compiler-2020-1-217-internal-error-assertion-failed-find-assoc/td-p/1207649). I do not know how to reasonably implement this feature without lambdas. @psychocoderHPC what should be the course of action? I would suggest not supporting the affected compiler versions, as this is a pretty obvious and gigantic bug. #1288 
- Mac OS CI is failing at preparation stage due to inability to get TBB. This may be a random glitch as it worked just yesterday and I made no changes there. Edit: this was a glitch, now works fine.

psychocoderHPC:
> * @psychocoderHPC what should be the course of action? I would suggest not supporting the affected compiler versions, as this is a pretty obvious and gigantic bug

This is only an option if we know that this will be fixed in the next version of INTEL compiler. If it stays for years in we need to find a workaround even if it is in the worst case something with macros or copy-pasted code.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -38,12 +39,702 @@
...
+                std::intmax_t iNumBlocksInGrid(static_cast<std::intmax_t>(numIterations));
+                std::intmax_t i;
+#        pragma omp for nowait schedule(static)
+                for(i = 0; i < iNumBlocksInGrid; ++i)
+#    else
+#        pragma omp for nowait schedule(static)
+                for(TIdx i = 0; i < numIterations; ++i)
+#    endif
+                {
+                    loopBody(i);
```
psychocoderHPC: for the INTEL compiler bug, it could help to create before `#pragma omp for nowait schedule(static)` a copy of the lambda function and then declare the copy explicit private.

Another possible workaround would be to apply the arguments for the functor here instead of somewhere before. Maybe the compiler has problems to move some arguments into the parallel for.

psychocoderHPC: I was able to work around the example https://community.intel.com/t5/Intel-C-Compiler/compiler-2020-1-217-internal-error-assertion-failed-find-assoc/m-p/1227009#M37871 with the following code on hemera.

```C++
template < typename F>
void foo(const F& f)
{
  #pragma omp parallel for schedule(dynamic) collapse(2) 
  for (unsigned int v = 0; v < 0; ++v){
   for (unsigned int t = 0; t < 0; ++t){
      auto x = [&f]( const auto & i){ f(i);};
      x(0);
   }
  }
}

void f() {
 foo( [](const auto &indices_2) {} );
}
```

psychocoderHPC (approved):

bernhardmgruber (approved):
Generally LGTM. I have not fully verified the implementation but I trust the tests.

PR #1281, opened 2021-03-27
Added warp info. to cheatsheet. Fixed some other names there.

frobnitzem:
While coding, I noticed the documentation needed updates.


bernhardmgruber (approved):
Thank you for catching the inconsistencies! I removed some of the namespaces a while back to make the API leaner and it seems I did not catch some of the mentionings in the documentation.

Also thanks for the little warp-level example.

PR #1282, opened 2021-03-29
fix static shared memory alignment

### LINKED ISSUE #1272: static assert on alignment
In https://github.com/alpaka-group/alpaka/pull/824/files static assertions related to allocation of more than core::vectorization::defaultAlignment were removed and the code was corrected.

However, in the new file alpaka-0.6.0/include/alpaka/block/shared/st/BlockSharedMemStMember.hpp this static assertion was reintroduced although the following lines seem to handle this allocation correctly:

`
constexpr std::uint32_t align = static_cast<std::uint32_t>(std::max(TDataAlignBytes, alignof(T)));
return (m_allocdBytes / align + (m_allocdBytes % align > 0u)) * align;
`

I suppose that the static_assert got there only by copy&paste and could/should be removed.

psychocoderHPC:
fix #1272

For CPU accelerators there was a limitation that only objects with a `alignof(T) < core::vectorization::defaultAlignment` could be created in shared memory.
By using the pointer of the shared memory data instead of the relative offset to the base pointer we can relex this restriction. Additionally, we avoid that our alignment requires a pointer with an alignment of `core::vectorization::defaultAlignment`, this was the case in the original implementation.

### DIFF
--- a/include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
+++ b/include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
@@ -132,7 +132,7 @@ namespace alpaka
 
             //! Byte offset to the end of the memory chunk
             //!
-            //! Calculate bytes required to store a type with a aligned starting offset.
+            //! Calculate bytes required to store a type with a aligned starting address in m_mem.
             //! Start offset to the origin of the user data chunk can be calculated with `result - sizeof(T)`.
             //! The padding is always before the origin of the user data chunk and can be zero byte.
             //!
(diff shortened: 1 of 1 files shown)


include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
```diff
@@ -142,11 +142,10 @@ namespace alpaka
...
             {
-                static_assert(
-                    core::vectorization::defaultAlignment >= alignof(T),
-                    "Unable to get block shared static memory for types with alignment higher than defaultAlignment!");
-                constexpr std::uint32_t align = static_cast<std::uint32_t>(std::max(TMinDataAlignBytes, alignof(T)));
-                return (byteOffset / align + (byteOffset % align > 0u)) * align + static_cast<uint32_t>(sizeof(T));
+                size_t const ptr = reinterpret_cast<size_t>(m_mem + byteOffset);
+                constexpr size_t align = std::max(TMinDataAlignBytes, alignof(T));
+                size_t const newPtrAdress = ((ptr + align - 1u) / align) * align + sizeof(T);
+                return static_cast<uint32_t>(newPtrAdress - reinterpret_cast<size_t>(m_mem));
```
bernhardmgruber, line 148: This smells a bit like `std::align`: https://en.cppreference.com/w/cpp/memory/align

But your implementation looks good as well.

bernhardmgruber (approved):

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
psychocoderHPC: You are right but the special requirement is that we can get the address from other threads too. There is no guarantee how `std::align` is implemented therefore IMO the only safe variant is to implement it on our own.

psychocoderHPC:
The failing CI job can be ignored, it is a CI issue.

include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
bernhardmgruber: Good point! Leave it then as it is.

bernhardmgruber:
LGTM

We need an admin to merge though because of the failing test.

Issue #1020, opened 2020-06-02
math test failed with Clang as CUDA compiler


SimeonEhrig:
```bash
cmake -DBoost_USE_STATIC_LIBS=ON -DBoost_USE_MULTITHREADED=ON -DBoost_USE_STATIC_RUNTIME=OFF -DCMAKE_BUILD_TYPE=Release -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE=OFF -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE=OFF -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=OFF -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=OFF -DALPAKA_ACC_CPU_BT_OMP4_ENABLE=OFF -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DALPAKA_ACC_GPU_HIP_ENABLE=OFF -DALPAKA_DEBUG=0 -DALPAKA_CUDA_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON ..
```

```bash
build/test/unit/math/math
using seed: 1337

testing:
 3 - accelerators !
17 - unary math operators
6 - binary math operators
testing with two data types
total 2 * accelerators * (unary + binary) * capacity


~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
math is a Catch v2.11.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
mathOps
-------------------------------------------------------------------------------
/alpaka/test/unit/math/src/math.cpp:173
...............................................................................

/alpaka/test/unit/math/src/math.cpp:144: FAILED:
  REQUIRE( results(i) == Approx(std_result) )
with expansion:
  -inf == Approx( inf )
with messages:
  Operator: OpExp
  Type: d
  The args buffer: 
  capacity: 1000
  0: [ 0, ]
  1: [ 1.797693134862316e+308, ]
  2: [ -1.797693134862316e+308, ]
[...]

PR #1285, opened 2021-04-12
Set default values for CUDA and HIP fast math flags to off

sbastrakov:
We discussed it on a meeting a few weeks ago. It feels not right that alpaka sets fast math by default, and thus potentially modifies behavior of a client code. It may be surprising, as comes by default somewhere inside our cmake. The flags themselves are of course useful, this PR merely sets their default values to off.

Also add the missing HIP flag to documentation.

Note that this change may modify the behavior of all alpaka clients who didn't set this option explicitly.

@psychocoderHPC I was actually unsure how the CI sets this variable: it seems to me, also with the default value, so also changes the CI?


sbastrakov:
Also, wrt to math tests potentially failing with fast-math enables, there is already [a note about it](https://github.com/alpaka-group/alpaka/blob/fe5c3f92388ee193784c5d41e13f49f01d09c074/test/unit/math/src/math.cpp#L147-L154). Potentially connected to #1020.

docs/source/advanced/cmake.rst
```diff
@@ -238,6 +238,11 @@ ALPAKA_ACC_GPU_HIP_ONLY_MODE
 
      Only back-ends using HIP can be enabled in this mode.
 
+ALPAKA_HIP_FAST_MATH
+  .. code-block::
+
+     Enable fast-math.
```
psychocoderHPC: Could you please add a warning box: `The default value is changed with alpaka 0.7.0 to "OFF"`

Later we need to add a note to the changelog into the breaking changes section too.

cmake/alpakaCommon.cmake
```diff
@@ -561,7 +561,7 @@ if(ALPAKA_ACC_GPU_HIP_ENABLE)
             set(ALPAKA_HIP_COMPILER "hipcc" CACHE STRING "HIP compiler")
             set_property(CACHE ALPAKA_HIP_COMPILER PROPERTY STRINGS "hipcc")
 
-            option(ALPAKA_HIP_FAST_MATH "Enable fast-math" ON)
+            option(ALPAKA_HIP_FAST_MATH "Enable fast-math" OFF)
```
psychocoderHPC, line 564: suggestion: Instead of hard set this option we should maybe use the compiler default. If we use cmake options with three states: `AUTO, OFF, ON` we could use `AUTO` as default.

Why I suggest it: IMO the goal of the alpaka's build system should be to set compiler options only if necessary e.g. workarounds or if the user explicit ask for it.
This would also solve all discussions about the default value.

psychocoderHPC (changes requested):

docs/source/advanced/cmake.rst
psychocoderHPC: We should discuss this today in the dev meeting.

cmake/alpakaCommon.cmake
bernhardmgruber: If we enable `AUTO` as an option, could we not remove the option entirely then? Users can always add the necessary flags themselves.

psychocoderHPC: It is possible, IMO not preferred, to remove it but these options are supported for the user. If you integrate alpaka into your project and you do not like to maintain a list of options per compilers, supporting it in alpaka is a multiplication so you do not add those options to any project.
I am also happy that we have those fast switches because I am jumping daily between different compilers and can not keep all options in my head.
An additional point is that it is not obvious how to provide these flags to the compiler because you need to know how alpaka is handling compile options internally.

bernhardmgruber: I also think that keeping the option is easier for users. I just wondered if we could remove some of our developer complexity. Anyway, let's keep the flag.

bernhardmgruber: Btw: I would not need the `AUTO` option. Having just a bool with `OFF` as default is fine for me!

docs/source/advanced/cmake.rst
sbastrakov: On the VC we decided to make this addition.

cmake/alpakaCommon.cmake
sbastrakov: On the VC we decided to generally follow the `ON - OFF -AUTO` approach with cmake options. I made a new issue #1286 for it.

bernhardmgruber (approved):

psychocoderHPC (approved):

sbastrakov:
Okay, now CI passed no problem. Ready to be merged!

PR #1289, opened 2021-04-14
cmake: refactor compiler option manipulation

### LINKED ISSUE #1286: Make CMake options feel more like compiler options
We discussed it on a VC before and after #1285 . There are two directions to improve alpaka CMake options concerning compiler flags:

- Avoid enabling options influencing user code by defaut. For fast math, #1285 is a fix, but there may be more of this kind, we need to check. The reason is that a user may not expect alpaka to hard-set it by default, and therefore not disable it when needed.
- Generally try to go to the ON / OFF / AUTO system, where a certain aspect can be hard-enabled, hard-disabled, or left by default. Again, how compiler flags are mostly operating.

psychocoderHPC:
solve #1286

* define unified cmake option:
  - `ALPAKA_FAST_MATH`
  - `ALPAKA_FTZ`
* rename  cmake option:
  - `ALPAKA_CUDA_NVCC_EXPT_EXTENDED_LAMBDA` into `ALPAKA_CUDA_EXPT_EXTENDED_LAMBDA`
* remove cmake option `ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION`
  This option is by default available from the CUDA find module under the name `CUDA_SEPARABLE_COMPILATION`
* update documentation
* add new cmake function `alpaka_compiler_option` and `alpaka_set_compiler_options`
* Use helper function `alpaka_set_compiler_options` to set compiler flags.

The cmake helper function `alpaka_set_compiler_options()` is documenting added compiler flags and a summary about all flags alpaka is setting will be printed at the end of cmake output.

```
cmake ../alpaka/ -DCMAKE_INSTALL_PREFIX=$HOME/lib/alpaka -Dalpaka_BUILD_EXAMPLES=ON -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON -DALPAKA_FAST_MATH=ON -DALPAKA_ACC_GPU_CUDA_ENABLE=ON -DALPAKA_FTZ=OFF
-- ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
-- ALPAKA_ACC_GPU_CUDA_ENABLED

List of compiler flags added by alpaka
device compiler:
    --expt-relaxed-constexpr;--generate-code=arch=compute_35,code=sm_35;--generate-code=arch=compute_35,code=compute_35;-std=c++14;--use_fast_math;--ftz=false;-Xcudafe=--display_error_number;-Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored

-- Configuring done
-- Generating done
-- Build files have been written to: /home/rwidera/workspace/buildAlpaka
```

The options ftz and fast-math will be implemented in a separate PR for the host compiler too.


cmake/alpakaCommon.cmake
```diff
@@ -45,6 +88,10 @@ option(ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE "Enable the OpenMP 2.0 CPU block threa
 option(ALPAKA_ACC_ANY_BT_OMP5_ENABLE "Enable the OpenMP 5.0 CPU block and block thread back-end" OFF)
 option(ALPAKA_ACC_ANY_BT_OACC_ENABLE "Enable the OpenACC block and block thread back-end" OFF)
 
+# Unified compiler options
+alpaka_compiler_option(FAST_MATH "Enable fast-math" DEFAULT)
+alpaka_compiler_option(FTZ "Set flush to zero for GPU" DEFAULT)
```
psychocoderHPC: remove `for GPU`

cmake/alpakaCommon.cmake
```diff
@@ -285,12 +332,9 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
             set_property(CACHE ALPAKA_CUDA_COMPILER PROPERTY STRINGS "nvcc;clang")
 
-            option(ALPAKA_CUDA_FAST_MATH "Enable fast-math" OFF)
-            option(ALPAKA_CUDA_FTZ "Set flush to zero for GPU" OFF)
-            option(ALPAKA_CUDA_SHOW_REGISTER "Show kernel registers and create PTX" OFF)
-            option(ALPAKA_CUDA_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps 'CMakeFiles/<targetname>.dir'" OFF)
-            option(ALPAKA_CUDA_NVCC_EXPT_EXTENDED_LAMBDA "Enable experimental, extended host-device lambdas in NVCC" ON)
-            option(ALPAKA_CUDA_NVCC_SEPARABLE_COMPILATION "Enable separable compilation in NVCC" OFF)
+            alpaka_compiler_option(CUDA_SHOW_REGISTER "Show kernel registers and create device ASM" DEFAULT)
+            alpaka_compiler_option(CUDA_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps 'CMakeFiles/<targetname>.dir'" OFF)
```
psychocoderHPC: should be `DEFAULT`

cmake/alpakaCommon.cmake
```diff
@@ -285,12 +332,9 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
+            alpaka_compiler_option(CUDA_EXPT_EXTENDED_LAMBDA "Enable experimental, extended host-device lambdas in CUDA with nvcc" ON)
```
psychocoderHPC, line 340: note: I think this must be `ON` else some of our examples or tests will not run.

docs/source/advanced/cmake.rst
```diff
@@ -261,3 +246,16 @@ ALPAKA_HIP_KEEP_FILES
...
+ALPAKA_HIP_SHOW_REGISTER
+  .. code-block::
+
+     Show the number of used kernel registers during compilation and create PTX.
+     Option is only available if ALPAKA_HIP_PLATFORM=nvcc.
+
+ALPAKA_HIP_SHOW_CODELINES
+  .. code-block::
+
+     Show kernel lines in cuda-gdb and cuda-memcheck. If ALPAKA_KEEP_FILES
```
psychocoderHPC: should be `ALPAKA_HIP_KEEP_FILES`

docs/source/advanced/cmake.rst
```diff
@@ -180,42 +193,23 @@ ALPAKA_CUDA_COMPILER
...
-
-     Enable separable compilation in NVCC.
-
 https://developer.nvidia.com/blog/separate-compilation-linking-cuda-device-code/
 
 ALPAKA_CUDA_SHOW_CODELINES
   .. code-block::
 
-     Show kernel lines in cuda-gdb and cuda-memcheck. If ALPAKA_CUDA_KEEP_FILES
+     Show kernel lines in cuda-gdb and cuda-memcheck. If ALPAKA_KEEP_FILES
```
psychocoderHPC: should be `ALPAKA_CUDA_KEEP_FILES`

cmake/alpakaCommon.cmake
```diff
@@ -855,11 +893,28 @@ if((ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE) AND ALPAKA_CUDA_COM
...
+    alpaka_set_compiler_options(HOST set CMAKE_CXX_FLAGS "${_ALPAKA_COMPILE_OPTIONS_STRING}")
 
     # Append CMAKE_CXX_FLAGS_[Release|Debug|RelWithDebInfo] to CMAKE_CXX_FLAGS
     # because FindCUDA only propagates the latter to nvcc.
     string(TOUPPER "${CMAKE_BUILD_TYPE}" build_config)
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${CMAKE_CXX_FLAGS_${build_config}}")
+    alpaka_set_compiler_options(HOST set CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_${build_config}}")
 endif()
 
+if(ALPAKA_COMPILER_OPTIONS_DEVICE OR ALPAKA_COMPILER_OPTIONS_DEVICE OR ALPAKA_COMPILER_OPTIONS_HOST_DEVICE)
```
bernhardmgruber:
```suggestion
if(ALPAKA_COMPILER_OPTIONS_DEVICE OR ALPAKA_COMPILER_OPTIONS_HOST OR ALPAKA_COMPILER_OPTIONS_HOST_DEVICE)
```

cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,49 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
...
+# operation - operation used to append given parameter: set, list, or target
+# name - name of the variable or target
+# ... - parameter to appended to the variable or target 'name'
+function(alpaka_set_compiler_options scope operation name)
+    if(scope STREQUAL HOST)
+        set(ALPAKA_COMPILER_OPTIONS_HOST ${ALPAKA_COMPILER_OPTIONS_HOST} ${ARGN} PARENT_SCOPE)
+    elseif(scope STREQUAL DEVICE)
+        set(ALPAKA_COMPILER_OPTIONS_DEVICE ${ALPAKA_COMPILER_OPTIONS_DEVICE} ${ARGN} PARENT_SCOPE)
+    elseif(scope STREQUAL HOST_DEVICE)
+        set(ALPAKA_COMPILER_OPTIONS_HOST_DEVICE ${ALPAKA_COMPILER_OPTIONS_HOST_DEVICE} ${ARGN} PARENT_SCOPE)
```
psychocoderHPC: `ALPAKA_COMPILER_OPTIONS_HOST_DEVICE` can be removed and instead the flags can be pushed to `ALPAKA_COMPILER_OPTIONS_HOST` and `ALPAKA_COMPILER_OPTIONS_DEVICE`

cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,49 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
...
+# This method is automatically documenting all parameter added to the compiler into the variables
+# ALPAKA_COMPILER_OPTIONS_HOST, ALPAKA_COMPILER_OPTIONS_DEVICE, and ALPAKA_COMPILER_OPTIONS_HOST_DEVICE.
+#
+# scope - which compiler is effected: DEVICE, HOST, or HOST_DEVICE
+# operation - operation used to append given parameter: set, list, or target
+# name - name of the variable or target
+# ... - parameter to appended to the variable or target 'name'
+function(alpaka_set_compiler_options scope operation name)
+    if(scope STREQUAL HOST)
+        set(ALPAKA_COMPILER_OPTIONS_HOST ${ALPAKA_COMPILER_OPTIONS_HOST} ${ARGN} PARENT_SCOPE)
```
bernhardmgruber, line 38: I prefer `list(APPEND ...)`, but I guess that does not work if the list does not exist and cannot inject into the parent scope.

cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,49 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
...
+    elseif(scope STREQUAL DEVICE)
+        set(ALPAKA_COMPILER_OPTIONS_DEVICE ${ALPAKA_COMPILER_OPTIONS_DEVICE} ${ARGN} PARENT_SCOPE)
+    elseif(scope STREQUAL HOST_DEVICE)
+        set(ALPAKA_COMPILER_OPTIONS_HOST_DEVICE ${ALPAKA_COMPILER_OPTIONS_HOST_DEVICE} ${ARGN} PARENT_SCOPE)
+    else()
+        message(FATAL_ERROR "alpaka_set_compiler_option 'scope' unknown, value must be 'HOST', 'DEVICE', or 'HOST_DEVICE'.")
+    endif()
+    if(operation STREQUAL list)
```
bernhardmgruber: I would rename this option to be clearer to the reader about what is meant at the callsite. E.g. `list_append` or maybe just `append`.

cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,49 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
...
+        set(${name} ${${name}} ${ARGN} PARENT_SCOPE)
+    elseif(operation STREQUAL set)
+        foreach(arg IN LISTS ARGN)
+            set(tmp "${tmp} ${arg}")
+        endforeach()
+        set(${name} "${${name}} ${tmp}" PARENT_SCOPE)
+    elseif(operation STREQUAL target)
```
bernhardmgruber: Consider also extending this operation's name a bit, e.g. `for_target`.

cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,49 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
...
+    else()
+        message(FATAL_ERROR "alpaka_set_compiler_option 'scope' unknown, value must be 'HOST', 'DEVICE', or 'HOST_DEVICE'.")
+    endif()
+    if(operation STREQUAL list)
+        set(${name} ${${name}} ${ARGN} PARENT_SCOPE)
+    elseif(operation STREQUAL set)
+        foreach(arg IN LISTS ARGN)
+            set(tmp "${tmp} ${arg}")
+        endforeach()
+        set(${name} "${${name}} ${tmp}" PARENT_SCOPE)
```
bernhardmgruber: Since this also appends to the variable `${name}` I fail to understand why we need to `set` operation. Can we not handle this with `list` instead?

cmake/alpakaCommon.cmake
```diff
@@ -855,11 +893,28 @@ if((ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE) AND ALPAKA_CUDA_COM
                  TARGET alpaka
                  PROPERTY INTERFACE_COMPILE_OPTIONS)
     string(REPLACE ";" " " _ALPAKA_COMPILE_OPTIONS_STRING "${_ALPAKA_COMPILE_OPTIONS_PUBLIC}")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${_ALPAKA_COMPILE_OPTIONS_STRING}")
+    alpaka_set_compiler_options(HOST set CMAKE_CXX_FLAGS "${_ALPAKA_COMPILE_OPTIONS_STRING}")
```
bernhardmgruber: Why would `alpaka_set_compiler_options(HOST list CMAKE_CXX_FLAGS ${_ALPAKA_COMPILE_OPTIONS_STRING})` be wrong here?

cmake/alpakaCommon.cmake
psychocoderHPC: `set` is equal but required because I need to move the result into the parent scope.

cmake/alpakaCommon.cmake
psychocoderHPC: I am currently on the way to rename `operation` into `type`. After that you define the type of the element you give into the method: `var`, `list`, or `target`

cmake/alpakaCommon.cmake
psychocoderHPC: A variable that is not a list is concatenated by spaces. A list is `;` separated.

cmake/alpakaCommon.cmake
psychocoderHPC: `CMAKE_CXX_FLAGS` is not a list. It is a string with space-separated entries.

cmake/alpakaCommon.cmake
psychocoderHPC: @bernhardmgruber I pushed an update. Instead of providing the operation to `alpaka_set_compiler_options()` the user is now providing the type of the parameter `name`. I also extended the documentation of the function `alpaka_set_compiler_options()` to avoid confusion between a list and a variable.

cmake/alpakaCommon.cmake
psychocoderHPC: The new command is now `alpaka_set_compiler_options(HOST var CMAKE_CXX_FLAGS ${_ALPAKA_COMPILE_OPTIONS_STRING})`

Documentation of `var` is: `var: is spaces sperated`

cmake/alpakaCommon.cmake
bernhardmgruber: Fair enough.

cmake/alpakaCommon.cmake
bernhardmgruber: OK!

bernhardmgruber (approved):

PR #1291, opened 2021-04-15
Fix CtxBlockOacc: SyncBlockThreads

jkelling:
The bug fixed here makes threads skip the seconds wait loop allowing them to run into a later barrier before `m_syncCounter` has been prepared for that by the master thread.

### DIFF
--- a/include/alpaka/ctx/block/CtxBlockOacc.hpp
+++ b/include/alpaka/ctx/block/CtxBlockOacc.hpp
@@ -113,8 +113,8 @@ namespace alpaka
                 }
 #    pragma acc atomic capture
                 {
-                    ++m_syncCounter[slot];
-                    sum = m_syncCounter[slot];
+                    ++m_syncCounter[slot + 1];
+                    sum = m_syncCounter[slot + 1];
                 }
                 while(sum < workerNum)
                 {


psychocoderHPC:
@jkelling I assume we should backport this fix to 0.6.1?

jkelling:
> @jkelling I assume we should backport this fix to 0.6.1?

Yes, please.

include/alpaka/ctx/block/CtxBlockOacc.hpp
```diff
@@ -113,8 +113,8 @@ namespace alpaka
                 }
 #    pragma acc atomic capture
                 {
-                    ++m_syncCounter[slot];
-                    sum = m_syncCounter[slot];
+                    ++m_syncCounter[slot + 1];
+                    sum = m_syncCounter[slot + 1];
```
psychocoderHPC, line 117: IMO there is a data race between this line and line 106. @jkelling I contacted you over Mattermost for a VC to peer review it on a shared screen.

psychocoderHPC (changes requested):

include/alpaka/ctx/block/CtxBlockOacc.hpp
psychocoderHPC: Ahh line 106 is setting `nextSlot` and not `slot`, then it should be ok.

psychocoderHPC (approved):

include/alpaka/ctx/block/CtxBlockOacc.hpp
jkelling: Yes, `nextSlot` is for the next barrier. The loop in line 119 is protecting that .. or not without this fix.

PR #1294, opened 2021-04-19
CMake CUDA: dev compile options not propagated

psychocoderHPC:
The compile options for the test cases are not propagated to all tests.

Note: I am not sure if this also happens with the other variables for the CPU. 
I have currently no time to check this. 

This fix is required for CUDA 11.3.

Tested with cmake@3.18.4

- [x] rebase against #1302


bernhardmgruber (dismissed):
I don't know enough CMake to understand why this is necessary. I also do not believe it does any harm.

sbastrakov (approved):

PR #1249, opened 2021-01-28
Add Accessor, an interface for accessing buffers in kernels

bernhardmgruber:
This PR adds a new class template `Accessor`:

```c++
template<typename TMemoryHandle, typename TElem, typename TBufferIdx, std::size_t TDim, typename TAccessModes>
struct Accessor;
```

An accessor is an interface to access the data stored by a memory object.
Memory objects in alpaka are currently all objects fulfulling the view concept, which includes buffers.

An accessor is created via one of the following calls:
```c++
auto a = accessWith<Mode1, Mode2, ...>(memObj); // where ModeN is an access tag
auto a = access(memObj); // read/write
auto a = readAccess(memObj); // read
auto a = writeAccess(memObj); // write
```

Currently, alpaka defines the access modes via the tags `ReadAccess`, `WriteAccess` and `ReadWriteAccess`.
However, the user is free to add their own tags and specialize `Accessor` based on them.

The access functions construct an accessor by retrieving a memory handle from the memory object, which is stored in the accessor.
An accessor additionally has an element type `TElem` which is the type of the instances returned on access.
This type may be mutated depending on the access modes.
E.g. a read/write accessor will return a `TElem&` whereas a read accessor will just return `TElem`.
The `TBufferIdx` is the integral type used for indexing and index computations.
`TDim` is the integral dimension of the accessor.
`TAccessModes` is either one access tag or a `std::tuple` of access tags. In the latter case, the accessor behaves as if only the first access tag was specified.

Accessors with multiple access tags can be further constrained to a single access tag (contained in the list of existing tags) by again calling one of the access functions on them.

The alias `BufferAccessor` specifies the accessor type for the buffer object of a given accelerator and may simplify specifying a kernel's interface.

For data access, the accessor overloads `operator()` and `operator[]`, which can be called with either `TDim` integrals or a corresponding `Vec`.

Releated issue: https://github.com/alpaka-group/alpaka/issues/38

PR #1306, opened 2021-04-22
Add Front and Contains type list meta functions

bernhardmgruber:
As suggested by @BenjaminW3, this PR splits out the two added meta functions form the accessor PR #1249.

### DIFF
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -146,6 +146,7 @@
 #include <alpaka/meta/NdLoop.hpp>
 #include <alpaka/meta/Set.hpp>
 #include <alpaka/meta/Transform.hpp>
+#include <alpaka/meta/TypeListOps.hpp>
 #include <alpaka/meta/Void.hpp>
 // offset
 #include <alpaka/offset/Traits.hpp>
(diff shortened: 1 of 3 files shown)


BenjaminW3:
Could you please create a new file per function? They have nothing in common with the boolean metafunctions in this header.

bernhardmgruber:
I think creating a new file per function is an overkill, since the funtions are really simple. However, there is no good existing place either. Metafunctions.hpp was the best existing choice IMO. What do other people think?

psychocoderHPC (dismissed):

psychocoderHPC:
@bernhardmgruber Maybe we should move it into a separate file as @BenjaminW3  suggested. 
`meta/ListOp.hpp` or `meta/ListSearch.hpp`
How is the file in mp11 called?

bernhardmgruber:
mp11 groups algorithms, as shown on the left side in their docs: https://www.boost.org/doc/libs/develop/libs/mp11/doc/html/mp11.html

I think for alpaka we could also have one header file containing type list operations. This header should include the content of `Apply.hpp`, `CartesianProduct.hpp`, `Concatenate.hpp`, `Filter.hpp`, `InheritFromList.hpp`, `Transform.hpp` and `Unique.hpp`. Maybe `TypeList.hpp` or `TypeListOps.hpp` is a good name.

We could also have a dedicated type list template, so `Accessor` does not need to use `std::tuple` for the access mode tag list.

psychocoderHPC (dismissed):

psychocoderHPC (approved):

PR #1308, opened 2021-04-26
Add NVHPC to CI

### LINKED ISSUE #203: EventCpu.hpp:284:32: warning: declaration of ‘lk’ shadows a previous local [-Wshadow]
```
alpaka/include/alpaka/event/EventCpu.hpp:284:32: warning: declaration of ‘lk’ shadows a previous local [-Wshadow]
                                 std::lock_guard<std::mutex> lk(spEventCpuImpl->m_Mutex);
                                ^
alpaka/include/alpaka/event/EventCpu.hpp:248:30: note: shadowed declaration is here
                     std::lock_guard<std::mutex> lk(spEventCpuImpl->m_Mutex);
                              ^
```
As this gets repeated on every object compilation, I find it crucial to eliminate these warnings somehow, either through cmake or by renaming the second `lk` inside the lambda with `lk2` or some more meaningful name even.

jkelling:
This PR adds CI jobs for nvc++ based on NVIDIA's docker image version 21.3.

* adds `linux_nvhpc-21.7_debug_analysis_oacc_multicore` OpenACC build and test targeting multi core cpus
* adds (gitlab) `.base_oacc_nvhpc` and `linux_nvhpc-21.7_cuda-11.4_oacc`: OpenACC build and test targeting NVIDIA GPUs
* ~~waiting for nvc++ CUDA support~~ making this one about OpenACC
* disables additional `TestTime` for OpenACC due lack of support in OpenACC

~~Relates to https://github.com/alpaka-group/llama/issues/203 (something got mixed up here)~~

## Edit:
This PR now also fixes some final issues in the OpenACC backend and adds some NVHPC 21.7 workarounds.

Also:
* ~~This PR adds a portable implementation of the random distributions in `alpaka/rand/RandAlpaka.hpp` (this part can be moved to a separate PR on request.)~~
* Disables 8 and 16 bit atomic tests. I suggest to either rebasing after the atomics have been cleaned-up or merging before and cleaning up atomics in general later. See https://github.com/alpaka-group/alpaka/issues/1402

## Todo

- [x]  includes commits from https://github.com/alpaka-group/alpaka/pull/1387 , so that one must be merged first.
- [x] merge after math: #1446 
- [x] merge after RandAlpaka: #1444


test/unit/meta/src/ApplyTupleTest.cpp
```diff
@@ -42,8 +42,10 @@ TEST_CASE("invoke", "[meta]")
     // invoke a free function
     REQUIRE(9 == alpaka::meta::invoke(abs_num, -9));
 
+#if !BOOST_COMP_PGI // nvc++: "a lambda is not allowed in an unevaluated expression"
     // invoke a lambda
     REQUIRE(42 == alpaka::meta::invoke([]() { return abs_num(-42); }));
+#endif
```
bernhardmgruber, line 48: Interesting observation. Lambdas in unevaluated contexts are a C++20 feature. Where is that lambda?

jkelling: My guess is, that this is a compiler bug and the lambda is not actually unevaluated.

nvc++ does not give comprehensive backtraces of instantiations like gcc would, so I do not know if the problem i somewhere deep inside catch2. Considering, that the very similar tests, with the `REQUIRE` and `meta::invoke` do compile, I think the lambda in question is the one which is the argument to `meta::invoke`.

bernhardmgruber: Checking the implementation of `alpaka::meta::invoke`, the lambda's type is actually used in unevaluated contexts (the `decltype` expressions in the return types of `invoke_impl`. Have you tried removing all the `decltype` expressions and replacing the `auto` return type by `decltype(auto)`?

jkelling: I think you are right, meaning the compiler may be correct in complaining, albeit a bit pedantic.

On the one hand the explicit return types look like just an artifact from the C++11 days. On the other, they do serve a purpose for SFINAE: When removing them, a call with a lambda leads to the last overload being used, which tries to call `<lambda_type>.<first arg as member pointer>` in the function body, which is an error. Having this expression on the `decltype` return type eliminates this variant from overload resolution.

So, you are saying, that alpaka already implicitly uses C++20? @psychocoderHPC what do you think about this?

psychocoderHPC: > On the one hand the explicit return types look like just an artifact from the C++11

In some places in alpaka explicit return types are required, else the compiler can not detect the return type.

> So, you are saying, that alpaka already implicitly uses C++20? @psychocoderHPC what do you think about this?

I am not aware that we use C++20, no compiler threw a warning or error.

psychocoderHPC: That `invoke` is using explicit return type is realy looking like a leftover from C++11. I would try what @bernhardmgruber  suggested.

https://github.com/alpaka-group/alpaka/blob/fe5c3f92388ee193784c5d41e13f49f01d09c074/include/alpaka/meta/ApplyTuple.hpp#L26-L56

jkelling: I already tried that, it breaks SFINAE, see above.

jkelling: We could replace the explicit return type by some `enable_if` but this would just move the lambda from one unevaluated context to another, as I understand it.

jkelling: I am back to thinking, that this is a compiler bug.

All compilers, including nvc++, seem to agree, that it is not allowed to have literal lambda (lambda-expression) in an unevaluated context, before c++20 . However, they also agree, that it is ok if you put any variable that is a lambda.
```c++
#include <utility>

template<class T, class... Args>
auto invoke(T&& t, Args... args)
	-> decltype(t(std::forward<Args...>(args...)))
	// OK
{
	return t(std::forward<Args...>(args...));
}

template<class... Args>
auto invokeL(Args... args)
	-> decltype([](int a){return a;}(std::forward<Args...>(args...)))
	// error: lambda-expression in unevaluated context
{
	return ([](int a){return a;}(std::forward<Args...>(args...)));
}

int b = 5;
auto lambda = [b = b](int a) mutable {return b += a;}; // making sure this thing has a closure

template<class... Args>
auto invokeLL(Args... args)
	-> decltype(lambda(std::forward<Args...>(args...)))
	// OK
{
	return (lambda(std::forward<Args...>(args...)));
}

int main()
{
	int b = 5;
	invoke([&b](int a){return a+b;}, 0);
	invokeL(0);
	invokeLL(0);
}
```

I do not see a lambda-expression in an unevaluated context in the in the involved Alpaka code.

bernhardmgruber: The behavior you describe here is fully conformant. You cannot define a lambda expression inside an unevaluated context, that is the `decltype(...)` expression here. The closure object created from the lambda expression, the `lambda` here, as well as any derived objects or types, can be used inside unevaluated contexts.

bernhardmgruber: > I do not see a lambda-expression in an unevaluated context in the in the involved Alpaka code.

You are right, there shouldn't be one. Still, you can remove all the `decltype` return types and replace `auto` by `decltype(auto)` to try to work around the bug.

jkelling: I cannot, because:

> On the one hand the explicit return types look like just an artifact from the C++11 days. On the other, they do serve a purpose for SFINAE: When removing them, a call with a lambda leads to the last overload being used, which tries to call `<lambda_type>.<first arg as member pointer>` in the function body, which is an error. Having this expression on the `decltype` return type eliminates this variant from overload resolution.

bernhardmgruber: Sorry, you are right! While you can remove it for some overloads, you cannot distinguish the member pointers (data member vs. function). Another call for C++17.

script/gitlabci/job_base.yml
```diff
@@ -47,3 +47,29 @@
...
+    ALPAKA_ACC_GPU_CUDA_ENABLE: "OFF"
+    ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: "OFF"
+    ALPAKA_ACC_ANY_BT_OACC_ENABLE: "ON"
+    ALPAKA_OFFLOAD_MAX_BLOCK_SIZE: 256
+    ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST: "OFF"
+    CMAKE_CXX_FLAGS: "-acc -ta=tesla"
+  extends: .base
+  tags:
+    - cuda
+    - intel
```
jkelling: @psychocoderHPC I added this gitlab-ci job. Will this run automatically at HZDR fron this PR, or does it need to be triggered manually? In any case: Where can I see the results?

jkelling: Never mind, found it. Apparently, I did not add the job, just the base. @psychocoderHPC : can you maybe kill the previous CI job I produced with wrong configs?

SimeonEhrig: I killed the GitLab CI pipeline.

jkelling: Thanks, but that was the latest one. This one should run. There are some skipped ones...

jkelling: Thanks. Can you please restart the latest one? That one should run...

SimeonEhrig: I restarted the last one and the other was canceled by anybody else.

include/alpaka/atomic/AtomicOaccExtended.hpp
```diff
@@ -114,7 +119,12 @@ namespace alpaka
                 T const& value) -> T
             {
                 return detail::criticalOp(
-                    [&]() {
+                    [&
+#    if BOOST_COMP_PGI
```
psychocoderHPC, line 122: Why do we not always capture the ptr by value to avoid the `#if` but keep the comment that nobody removed the capturing?

psychocoderHPC (dismissed):

include/alpaka/atomic/AtomicOaccExtended.hpp
jkelling: I hope, that this bug gets fixed some time. Once there is a compiler version without this bug, and hopefully some others, we can move the CI to that version, make it the minimum supported version of NVHPC and remove the workaround. At the moment I would consider NVHPC OpenACC, like all offloading compilers, work in progress.

jkelling:
I am dissappointed by the failure in the [gitlab CI](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1643888627): The tests and examples compile and run with NVHPC 21.7 on my local machine :/ .

jkelling:
@psychocoderHPC this PR is ready. CI passed in the last iteration. Rebased onto develop and clean-up the branch now.

https://github.com/alpaka-group/alpaka/pull/1308/commits/1a944bcf0b631dbb5ba3f8588977027f5a71789f is somewhat forstalling https://github.com/alpaka-group/alpaka/pull/1431 , but it is required for the CI to pass on this one. I would prefer to merge this PR as https://github.com/alpaka-group/alpaka/pull/1431 is not ready yet, which then has to make sure not to break OpenACC.

.github/workflows/ci.yml
```diff
@@ -82,6 +82,7 @@ jobs:
   ci:
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
+    container: ${{ matrix.container }}
```
bernhardmgruber: I don't see where you set a key `container` in your changes below. Why is this needed?

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -782,10 +782,6 @@ using TestAccs = alpaka::test::EnabledAccs<alpaka::DimInt<1u>, std::size_t>;
 TEMPLATE_LIST_TEST_CASE("atomicOperationsWorking", "[atomic]", TestAccs)
 {
     using Acc = TestType;
-    TestAtomicOperations<Acc, unsigned char>::testAtomicOperations();
-    TestAtomicOperations<Acc, char>::testAtomicOperations();
-    TestAtomicOperations<Acc, unsigned short>::testAtomicOperations();
-    TestAtomicOperations<Acc, short>::testAtomicOperations();
```
bernhardmgruber, line 788: I think this is too harsh, as now we are no longer testing 8 and 16 bit atomics on CPUs, which support these. Please also synchronize with @tonydp03, who is also working on this in #1431.

bernhardmgruber (commented):
Consider also splitting off the random number extensions and synchronize with @sliwowitz, who is also working on this in #1319 and has recently brought in several PRs on that topic.

.github/workflows/ci.yml
jkelling: Right. This is left over from an earlier attempt to select the image. Alpak CI is using `ALPAKA_CI_DOCKER_BASE_IMAGE_NAME=` for this.
```suggestion
```

test/unit/atomic/src/AtomicTest.cpp
jkelling: This is not supposed to do redundant work. This is just a temporary measure following one path of the currently inconsistent handling of atomic tests.  #1431 is expected to fix this. I would like to have this PR merged first so that #1431 can the requirements herein into account.

jkelling:
> Consider also splitting off the random number extensions and synchronize with @sliwowitz, who is also working on this in #1319 and has recently brought in several PRs on that topic.

That PR adds the philox PRNG. @sliwowitz told me, that there is no PR for the RNG redesign yet. The introduction of `RandAlpaka` here follows the existing style of RNG, so can be refactored like the rest when the new design is finalized. It does add a new requirement for the redesign: A generic PRNG that works on accelerators. This is only a fallback until something better is available, as it uses the CPU implementation of TinyMT which is portable. While it could be split into a separate PR, because it is touch alpaka's RNG lib, which is not at the core of this PR, it fits as part of this PR, because it is only required to make the RNG tests work for OpenACC on GPU.

In any case: This PR should be merged before the redesign of the alpaka RNG lib to make sure that redesign does not break the OpenACC (and by extension OpenMP5) backend.

test/unit/atomic/src/AtomicTest.cpp
bernhardmgruber: Could you plese then comment the tests out and add a comment that those should be reenabled later for the targets supporting them? I want to avoid getting these tests lost when #1431 rebases later (whenever that is).

jkelling:
> This is only a fallback until something better is available, as it uses the CPU implementation of TinyMT which is portable. While it could be split into a separate PR, because it is touch alpaka's RNG lib, which is not at the core of this PR, it fits as part of this PR, because it is only required to make the RNG tests work for OpenACC on GPU.

This is not completely correct: It does add a Box-Muller transform which would could be used by  #1319 and thus addresses https://github.com/alpaka-group/alpaka/issues/1441 . This gives it merit as a separate PR.

jkelling:
I created https://github.com/alpaka-group/alpaka/pull/1444 which covers only the first commit of this PR.

@sliwowitz please rather review that one.

script/gitlabci/job_base.yml
psychocoderHPC: You can push any time new commits or force push to your PR. The HZDR CI will automatically kill old instances of the same PR.

psychocoderHPC (dismissed):

psychocoderHPC:
@jkelling Could you review you inital PR description

```
This PR adds a portable implementation of the random distributions in alpaka/rand/RandAlpaka.hpp (this part can be moved to a separate PR on request.)
```

Looks like this is outdated. Please notify me if this PR can be merged or has dependencies on other PRs.

jkelling:
> Looks like this is outdated. Please notify me if this PR can be merged or has dependencies on other PRs.

Added two prior PRs.

bernhardmgruber:
@jkelling please rebase!

jkelling:
@bernhardmgruber Thanks for the ping. Done.

jkelling:
Two CI Jobs failed:
* [linux_nvcc-11.3_clang-6_release_cuda_only](https://github.com/alpaka-group/alpaka/runs/4149884482?check_suite_focus=true) with `./script/run.sh: line 50: cmake: command not found`
* [linux_nvcc-11.2_gcc-7_debug](https://github.com/alpaka-group/alpaka/runs/4149883947?check_suite_focus=true) with, i guess, `which nvcc` returning `1`.

These are strange for just CI glitches (usually apt or the like fails).

This PR touches
* [script/docker_ci.sh](https://github.com/alpaka-group/alpaka/pull/1308/files#diff-4dc84d2789fe0cd5abcb451f6b00831520c64f1e73ba7c79b4e01258d6c7215f)
* [script/run_generate.sh](https://github.com/alpaka-group/alpaka/pull/1308/files#diff-26b3156cae89013b6a3b6158c28cd2665235bbacd36f6ca3b014a718c949f019)
* [script/run_tests.sh](https://github.com/alpaka-group/alpaka/pull/1308/files#diff-3454a861a4856a5928f5fa96ed68aa6b67e2fa326a58b964f52f54acb80fb762)

Does anyone how those changes might cause this?

Or maybe just try re-running the CI, in case it is a glitch?

sliwowitz:
The linux_nvcc-11.3_clang-6_release_cuda_only run failed to install a bunch of necessary stuff including git (whose invocation is the actual failure hidden under `./script/run.sh: line 50: cmake: command not found`) - you can see it in the log on line 894. Similar failures follow, e.g. `wget` fails to get the the `cuda-repo-XXX.deb` file at line 1552. 

The linux_nvcc-11.2_gcc-7_debug run was also unable to download packages - see lines 2312-2408, I suppose that's the actual cause of the ultimate error.

I'm no CI expert, but this looks a lot like a temporary internet connection problem on the side of the CI runners, so I'd say re-run.

sliwowitz (approved):

j-stephan:
I have restarted the CI.

test/unit/atomic/src/AtomicTest.cpp
j-stephan: Has this been resolved? @jkelling @bernhardmgruber

j-stephan: Ah, sorry for the noise. I should have read further...

j-stephan (approved):

test/unit/atomic/src/AtomicTest.cpp
jkelling: Yes, these lines are in, but commented, with a note for the atomic test rework.
https://github.com/alpaka-group/alpaka/pull/1308/files#diff-97505ba324687f46490362096f3ee6b11b77b812ca5e918fa54e956304865c81

PR #1309, opened 2021-05-04
Fix OpenMP schedule support

sbastrakov:
There is a bug in #1279 and so currently in `develop` and 0.6.1 release branch, reported by DLR. Faulty template specialization logic results in some valid combinations not compiling. I accidentally introduced it while fighting old compilers in the CI and then the tests were not trying out that combination, so it went through. The code is quite complicated as well and now became even slightly more due to the fix.

This PR in the current state attemps to fix and try how CI likes it. The unit tests need to be extended as well, I will do it in this PR.

### DIFF
--- a/example/openMPSchedule/src/openMPSchedule.cpp
+++ b/example/openMPSchedule/src/openMPSchedule.cpp
@@ -58,6 +58,7 @@ struct OpenMPScheduleMemberKernel : public OpenMPScheduleDefaultKernel
     static constexpr auto ompScheduleKind = alpaka::omp::Schedule::Static;
 
     //! Member to set OpenMP chunk size, can be non-static and non-constexpr.
+    //! Defining kind as member and not defining chunk will result in default chunk size for the kind.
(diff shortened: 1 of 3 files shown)


sbastrakov:
cc @krzikalla

sbastrakov:
Note for reviewers. After making a fix I reordered partial specializations as this makes more sense in the fixed version. However, github diff doesn't understand it well (since implementations are mostly copy-pasted). So the actual change is in partial specialization arguments, and some additional traits needed for it, not in the bodies of the loops like one may think from the diff. If that helps, I can maybe restore the old order.

sbastrakov:
I added a fix for build issue on some compilers, hopefully it works now. I also added all combinations of use to unit tests, so that the thing missed originally would not be missed. However, these unit tests compile extremely slowly on some compilers and maybe I will need to reduce the number of combinations. Will see from how the CI goes.

BenjaminW3 (approved):

psychocoderHPC (approved):

PR #1310, opened 2021-05-07
Add support for clang 11

BenjaminW3:


README.md
```diff
@@ -70,18 +70,18 @@ Supported Compilers
...
+| std::thread |:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:x:|:white_check_mark:|:white_check_mark:|
+| Boost.Fiber |:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:x:|:x:|:white_check_mark:|
+|TBB|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:x:|:white_check_mark:|:x:|
+|CUDA (nvcc)|:white_check_mark: <br/> (CUDA 9.0-11.3)|:white_check_mark: <br/> (CUDA 9.2-11.3) |:white_check_mark: <br/> (CUDA 10.1-11.3) |:white_check_mark: <br/> (CUDA 11.0-11.3)|:x:|:white_check_mark: <br/> (CUDA 9.1-11.3)|:white_check_mark: <br/> (CUDA 10.1-11.3)|:white_check_mark: <br/> (CUDA 11.0-11.3)|:white_check_mark: <br/> (CUDA 11.1-11.3)|:white_check_mark: <br/> (CUDA 11.1-11.3)| - |:x:|:white_check_mark: <br/> (CUDA 10.1-11.3)|
```
bernhardmgruber, line 82: The `x` in the gcc 10.3 column makes it look really bad now, because gcc 10.2 worked fine AFAIK. Maybe we should split the column for gcc 10.1 - 10.2, and gcc 10.3? Or maybe put a checkmark into the cell with the note: except gcc 10.3. What do others think?

bernhardmgruber (commented):
I don't like how the README.md now shows that gcc 10.x does not work with nvcc. But the rest looks fine to me!

README.md
psychocoderHPC: We can only mention that gcc 10.2 is working if we add tests for 10.2

bernhardmgruber: Since we do not test the minor releases of compilers, it is right to say that the latest gcc 10 does not work. Your proposed changes are therefore correct.

bernhardmgruber (approved):

README.md
```diff
@@ -70,18 +70,18 @@ Supported Compilers
...
+|CUDA (clang) | - | - | - | - | - | - | - | :white_check_mark: <br/> (CUDA 9.2-10.1) | :white_check_mark: <br/> (CUDA 9.2-10.1) | :white_check_mark: <br/> (CUDA 10.0-10.2) | - | - | - |
```
j-stephan, line 83: @BenjaminW3 Do you remember why the clang-11 entry was changed to CUDA 10.0 as minimum? Technically previous CUDA versions should work, too.

PR #1318, opened 2021-05-25
Rename "generator" to "engine" in rand

sliwowitz:
As discussed in an earlier meeting, *engine* is a better name than *generator*. It's also consistent with the STL which refers to these as engines (even though it does not use such a namespace directly).


j-stephan (changes requested):
The changes look good so far, but the test cases need to be updated as well.

Could you also squash your commits into one?

sliwowitz:
> The changes look good so far, but the test cases need to be updated as well.

Sorry, forgot about these. fixing it now.

> Could you also squash your commits into one?

Sure. I thought it would be squashed after the review upon merging to develop. Is it possible to squash on the go as I add further commits to this already opened PR?

j-stephan:
Yes, you can squash them once the PR is ready to be merged.

bernhardmgruber (approved):

j-stephan (approved):
LGTM.

Issue #710, opened 2018-12-10
Counter based random number generators?


tdd11235813:
I am looking for an efficient way to obtain the same pseudo random numbers being independent of the backend and parallelization. The focus is on parallel RNG, so it can be easily mapped to serial execution.

__Counter/Index based RNGs__ ([1](https://en.wikipedia.org/wiki/Counter-based_random_number_generator_%28CBRNG%29), [2](https://streamhpc.com/blog/2016-03-18/random-numbers-in-parallel-computing-generation-and-reproducibility-part-1/), [3](https://www.iaria.org/conferences2015/filesADVCOMP15/parallel-rng-advcomp15.pdf)) are also lightweight (stateless) and fast in the parallel environment, and therefore used in parallel Monte-Carlo etc.
Other RNGs also offer streams and sub-streams like MRG32k3a with "fast enough" jump aheads and also might offer better quality than CBRNGs, but have states, are slower (e.g., for CBRNG skip-ahead is O(1)) and might be more tricky to use it correctly in parallel environment.

With rocRAND/hipRAND and cuRAND we have RNGs suited for parallel reproducible random number generation, and they also include counter based RNGs. It looks like alpaka does not provide wrapper for it and the host side does not even have support for it.
[Random123](http://www.thesalmons.org/john/random123/) could be a library for this gap, because it offers CBRNGs and has CPU as well as CUDA and OpenCL backends, although CPU would be the interesting part for now.

Related to: #390 (unify RNG selection), #610 (Monte Carlo example), upcoming N-body project for mephisto (with real physics ;))

PS: Philox is an CBRNG, and here is [rocRAND's implementation](https://github.com/ROCmSoftwarePlatform/rocRAND/blob/master/library/include/rocrand_philox4x32_10.h)

Edit: just want to add that pregenerating random numbers and using them via a table-based approach is not useful, when you have to generate a huge number of them.

j-stephan:
We discussed this in today's meeting. @sliwowitz is currently working on a separate RNG library on top of alpaka that will adress this issue. This is therefore WONTFIX and will be closed once the new RNG library is public.

PR #1319, opened 2021-05-25
Philox random number generator

sliwowitz:
Implementation of the counter-based Philox PRNG. 

Its default variant (4x32x10) with saved counter state would be a good default for consistent random numbers across all accelerators, sine its state is acceptably small: 352 bits = 44 bytes), and its behavior is consistent with usual expectations (it is compatible with STL distributions). 

A version that only outputs vectors of the full length of the Philox counter (4 numbers of size 32 bits in the default 4x32x10 variant) can be used to further cut the size of the saved state down to 192 bits = 24 bytes at the expense of usage convenience. The user must consume the whole vector and deal with the result themself.

It is further possible to use the generator in a completely stateless mode by initializing the engine object with custom values of *Counter* and *Key* at request for a new number. If there is a need for such use case, a better API might be sought.

This is a draft looking for comments on the overall idea. There are still some missing pieces:

- No API for stateless operation (tracked in #1443)
- No Philox-specific tests yet (the current rand test pass)
- Performance is sub-optimal, especially with CUDA (to be done as the last task)
- Box-Muller states should be implemented at some point for fast normal distribution generation (tracked in #1441)

related issues: #390  and #710


bernhardmgruber:
I would like to suggest to reorganize the commits. Please squash the clang-format into the commit where you forgot to format. Furthermore, the renaming of `generator` to `engine` could be sent in as a separate PR.

example/randomCells2D/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright 2014-2020 Benjamin Worpitz, Jan Stephan
```
bernhardmgruber: I am usually against copyright headers. But if we have them, shouldn't you put your name there and 2021? Or did you copy the file from somewhere? Maybe at least add yourself.

example/randomCells2D/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+# THE SOFTWARE IS PROVIDED “AS IS” AND ISC DISCLAIMS ALL WARRANTIES WITH
+# REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
+# SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
+# IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+#
+
+################################################################################
```
bernhardmgruber, line 19: I thought I had killed those section comments.

example/randomCells2D/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+# Find alpaka.
+
+if(NOT TARGET alpaka::alpaka)
+    option(USE_ALPAKA_SOURCE_TREE "Use alpaka's source tree instead of an alpaka installation" OFF)
+
+    if(USE_ALPAKA_SOURCE_TREE)
+        # Don't build the examples recursively
+        set(alpaka_BUILD_EXAMPLES OFF)
+        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../.." "${CMAKE_BINARY_DIR}/alpaka")
+    else()
```
bernhardmgruber, line 44: Adding a subdirectory from your parent directory feels wrong. I would argue, `find_package(alpaka ...)` should be enough, but I am not the CMake expert here.

example/randomCells2D/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+cmake_minimum_required(VERSION 3.18)
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+################################################################################
+# Project.
+
+set(_TARGET_NAME randomCells2D)
+
+project(${_TARGET_NAME})
```
bernhardmgruber, line 31: I have seen this in many places here in alpaka. You can just set `project(randomCells2D)` and then later use the built-in variable `${PROJECT_NAME}`, instead of creating your own variable.

example/randomCells2D/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+alpaka_add_executable(
+    ${_TARGET_NAME}
+    src/randomCells2D.cpp)
+target_link_libraries(
+    ${_TARGET_NAME}
+    PUBLIC alpaka::alpaka)
+
+set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER example)
+
+add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME})
```
bernhardmgruber, line 61: This is an example and not a test. Or do we add examples as tests in alpaka? That would be weird.

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
+#include <alpaka/alpaka.hpp>
+#include <alpaka/example/ExampleDefaultAcc.hpp>
+
+#include <alpaka/rand/RandPhilox.hpp>  //XXX: I don't like this.
```
bernhardmgruber: Add it to alpaka.hpp then.

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+#include <cstdint>
+#include <cstdlib>
+#include <iostream>
+
+    unsigned constexpr NUM_CALCULATIONS = 256;
+    unsigned constexpr NUM_X = 100;
+    unsigned constexpr NUM_Y = 100;
```
bernhardmgruber: Weird indentation. Is that from clang-format?

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+
+    struct InitRandomKernel
+    {
+        template<typename TAcc, typename TExtent, typename TRandEngine>
+        ALPAKA_FN_ACC auto operator()(TAcc const& acc, TExtent const extent, TRandEngine* const states) const -> void
+        {
+            auto idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+            auto linearIdx = alpaka::mapIdx<1u>(idx, extent)[0];
```
bernhardmgruber:
```suggestion
            auto const idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
            auto const linearIdx = alpaka::mapIdx<1u>(idx, extent)[0];
```

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        template<typename TAcc, typename TExtent>
+        ALPAKA_FN_ACC auto operator()(
+            TAcc const& acc,
+            TExtent const extent,
+            alpaka::rand::Philox4x32x10<TAcc>* const states,
+            float* const cellsA) const -> void
+        {
+            auto idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+            auto linearIdx = alpaka::mapIdx<1u>(idx, extent)[0];
```
bernhardmgruber:
```suggestion
            auto const idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
            auto const linearIdx = alpaka::mapIdx<1u>(idx, extent)[0];
```

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        template<typename TAcc, typename TExtent>
+        ALPAKA_FN_ACC auto operator()(
+            TAcc const& acc,
+            TExtent const extent,
+            alpaka::rand::Philox4x32x10Vector<TAcc>* const states,
+            float* const cellsB) const -> void
+        {
+            auto idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+            auto linearIdx = alpaka::mapIdx<1u>(idx, extent)[0];
```
bernhardmgruber:
```suggestion
            auto const idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
            auto const linearIdx = alpaka::mapIdx<1u>(idx, extent)[0];
```

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+            // Setup generator and distribution.
+            alpaka::rand::Philox4x32x10Vector<TAcc> engine(states[linearIdx]);
+            unsigned constexpr resultVectorSize = std::tuple_size_v<decltype(engine())>;
```
bernhardmgruber: Do you really need to go through `std::tuple_size_v`? Does `engine()` return a `std::array`? Because then you could use:
`constexpr auto resultVectorSize = decltype(engine())::size();`

Consider `engine` also having some kind of `size()` member, if that information is so critical.

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+    alpaka::wait(queue);
+
+    for(Idx x = 0; x < numX; ++x)
+    {
+        for(Idx y = 0; y < numY; ++y)
+        {
+            ptrBufHostA[y * numX + x] = 0;
+            ptrBufHostB[y * numX + x] = 0;
+        }
+    }
```
bernhardmgruber: Is this something a `std::memset` or an `alpaka::memset` could do?

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,187 @@
...
+    float avgA = 0;
+    float avgB = 0;
+    for(Idx x = 0; x < numX; ++x)
+    {
+        for(Idx y = 0; y < numY; ++y)
+        {
+            avgA += ptrBufHostA[y * numX + x];
+            avgB += ptrBufHostB[y * numX + x];
+        }
+    }
```
bernhardmgruber: Isn't this just an unnecessary 2D version of `std::reduce(ptrBufHostA, ptrBufHostA + numX * numY)`?

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -0,0 +1,220 @@
...
+        , bufAccResult{alpaka::allocBuf<float, Idx>(alpaka::getDevByIdx<Acc>(0u), extentResult)}
+    {
+    }
+};
+
+/// Save the results file in CSV format. One line for each "point"; line length is the number of "rolls".
+void saveToFile(std::string filename, float const* buffer, Box const& box)
+{
+    std::ofstream output;
+    output.open(filename, std::ios::out);
```
bernhardmgruber:
```suggestion
    std::ofstream output(filename);
```

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -0,0 +1,220 @@
...
+    std::cout << "Writing " << filename << std::endl;
+    auto lineLength = box.extentResult[0] / box.extentRand[0];
+    for(Box::Idx i = 0; i < box.extentResult[0]; ++i)
+    {
+        output << buffer[i] << ((i + 1) % lineLength ? ", " : "\n");
```
bernhardmgruber: I saw people writing TSV (tab separated values) a while back at CERN. That was unfamiliar at first, but i nicely eliminates the problem with locales, where e.g. en_US OS uses ',' as separator und '.' as decimal point, while 'de_DE` uses ';' as separater because it uses ',' as decimal point. So on a computer with a German OS, you will write a wrong file and have troubles opening these files in e.g. Excel. Thus I prefer TSV files in the meantime, they eliminate one of the two problems.

bernhardmgruber: A bit of googling just showed me that in Russian and Czechia `,` is also used as decimal separator. Might be relevant for some team members :)

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -0,0 +1,220 @@
...
+    // alpaka::memcpy(box.queue, box.bufHostRand, box.bufAccRand, box.extentRand);
+    // alpaka::wait(box.queue);
+
+    // Set up the pointers to the results buffers
+    float* const ptrBufHostResult{alpaka::getPtrNative(box.bufHostResult)};
+    float* const ptrBufAccResult{alpaka::getPtrNative(box.bufAccResult)};
+
+    // Initialise the results buffer to zero
+    for(Box::Idx i = 0; i < box.extentResult[0]; ++i)
+        ptrBufHostResult[i] = 0;
```
bernhardmgruber, line 266: memset

include/alpaka/rand/Philox/PhiloxBaseArrayLike.hpp
```diff
@@ -0,0 +1,106 @@
...
+                 * Advances the internal value of this->state.counter
+                 *
+                 * @param offset number of N-vectors to skip
+                 */
+                ALPAKA_FN_HOST_ACC void skip4(uint64_t offset)
+                {
+                    Counter& counter = static_cast<TImpl*>(this)->state.counter;
+                    Counter temp = counter;
+                    counter[0] += lo32(offset);
+                    counter[1] += hi32(offset) + (counter[0] < temp[0] ? 1 : 0);
```
bernhardmgruber: `lo32` and `hi32` have a bit too short of a name for my taste. Furthermore, I guess this code assumes that `counter[i]` have 32 bits? If yes, place `static_assert` this.

include/alpaka/rand/Philox/helpers/metaprog.hpp
```diff
@@ -0,0 +1,109 @@
...
+             */
+            template<unsigned N>
+            struct repeat_n
+            {
+                template<typename Fn>
+                constexpr void operator()(Fn&& fn) const
+                {
+                    static_for<0, static_cast<int>(N)>()([&](int i) { fn(); });
+                }
+            };
```
bernhardmgruber: This is basically a weaker `boost::mp11::for_each`. The entire file should be moved into `alpaka/meta`. Furthermore, I want to again raise the question whether alpaka should adopt a proper metaprogramming library like boost::mp11, so we do not need to write and maintain such utilities ourselves.

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
```diff
@@ -0,0 +1,90 @@
...
+                ALPAKA_FN_HOST_ACC auto nRounds(Counter const& counter_in, Key const& key_in) -> Counter
+                {
+                    Key key{key_in};
+                    Counter counter = TBackend::singleRound(counter_in, key);
+                    helpers::repeat_n<numRounds>()(
+                        [&]()
+                        {
+                            key = TBackend::bumpKey(key);
+                            counter = TBackend::singleRound(counter, key);
+                        });
```
bernhardmgruber: Why can't this be a `for` loop? Potentially with `#pragma unroll`. This looks like a premature optimization.

include/alpaka/rand/Philox/PhiloxBaseCudaPlain.hpp
```diff
@@ -0,0 +1,106 @@
...
+            template<typename TParams, typename TImpl>
+            class PhiloxBaseCudaPlain
+            {
+                static_assert(TParams::counterSize == 4, "GPU Philox implemented only for counters of width == 4");
+
+            public:
+                using Counter = uint4;
+                using Key = uint2;
+
+                static constexpr uint64_t WEYL_64_0 = 0x9E3779B97F4A7C15; /* golden ratio */
```
bernhardmgruber: Maybe add a comment where this numbers are taken from.

include/alpaka/rand/Philox/PhiloxBaseStdArray.hpp
```diff
@@ -0,0 +1,57 @@
...
+                /** Single round of the Philox shuffle
+                 *
+                 * @param counter state of the counter
+                 * @param key value of the key
+                 * @return shuffled counter
+                 */
+                ALPAKA_FN_HOST_ACC auto singleRound(Counter const& counter, Key const& key)
+                {
+                    auto [H0, L0] = mulhilo32(counter[0], this->MULTIPLITER_4x32_0);
+                    auto [H1, L1] = mulhilo32(counter[2], this->MULTIPLITER_4x32_1);
```
bernhardmgruber: If this PR lives long enough, we can get C++17 in time.

include/alpaka/rand/Philox/PhiloxBaseTraits.hpp
```diff
@@ -0,0 +1,80 @@
...
+                 * @tparam TParams Philox algorithm parameters
+                 * @tparam TImpl engine type implementation (CRTP)
+                 */
+                template<typename TAcc, typename TParams, typename TImpl>
+                struct PhiloxBaseTraits<
+                    TAcc,
+                    TParams,
+                    TImpl,
+                    typename std::enable_if<
+                        std::is_same_v<typename alpaka::traits::DevType<TAcc>::type, alpaka::DevUniformCudaHipRt>>>
```
bernhardmgruber: Instead of using `enable_if`, can you specialize on `TAcc` directly? Something like:
```c++
template<...>
struct PhiloxBaseTraits<alpaka::AccUniformCudaHipRt<...>, ...> { ... };
```

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+#pragma once
+
+#include <alpaka/rand/Philox/PhiloxBaseTraits.hpp>
+#include <alpaka/rand/Philox/mulhilo.hpp>
+
+#include <utility>
+
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED) || defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+#    define CUDA_OLD_COMPAT
```
bernhardmgruber: That is dangerous, because someone could include `alpaka/rand/Philox/helpers/cuintArray.hpp`. If this is really needed, maybe set it in cmake for the entirety of alpaka.

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,162 @@
...
+                        return state.result.x;
+                    case 1:
+                        return state.result.y;
+                    case 2:
+                        return state.result.z;
+                    case 3:
+                        return state.result.w;
+                    default: // Should never get here
+                        return 0u;
+                    }
```
bernhardmgruber: I am crossing my fingers for your optimizer! That could be really inefficient. I generally wondered, why you use the vector types in CUDA an not just something like `std::array` but for CUDA.

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,162 @@
...
+                    : state{
+                        {0, 0, 0, 0},
+                        {static_cast<uint32_t>(lo32(seed)), static_cast<uint32_t>(hi32(seed))},
+                        {0, 0, 0, 0},
+                        0}
+                {
+                    this->skipSubsequence(subsequence);
+                    skip(offset);
+                    advanceState();
+                }
```
bernhardmgruber: I think I would have written something like this:
```suggestion
                    : state{} // bulk zero initialize
                {
                    state.seedLow = lo32(seed); // however seedLow is called
                    state.seedHigh = hi32(seed);
                    this->skipSubsequence(subsequence);
                    skip(offset);
                    advanceState();
                }
```
I would be too afraid that I might miss initializing a value in `state` and have a bug.

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,162 @@
...
+
+                // XXX:
```
bernhardmgruber: What's that comment?

include/alpaka/rand/Philox/mulhilo.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <alpaka/core/Common.hpp>
+
+#include <array>
+
+namespace alpaka
+{
+    namespace rand
+    {
+        /// Get high 32 bits of a 64-bit number
+        ALPAKA_FN_HOST_ACC constexpr static std::uint64_t hi32(std::uint64_t x)
```
bernhardmgruber: Why is the result `uint64_t`, when you get 32 bits out?

bernhardmgruber (changes requested):

example/randomCells2D/CMakeLists.txt
sliwowitz: fixed in next iteration

example/randomCells2D/CMakeLists.txt
sliwowitz: the CMakeLists.txt files in examples still have them

example/randomCells2D/CMakeLists.txt
sliwowitz: I've just modified CMakeLists.txt from another example so that they have the same strcutre. If this is to be changed, we should IMHO do it for all *example* files in a separate PR.

example/randomCells2D/CMakeLists.txt
sliwowitz: Same as above - should be a separate PR for all example files. I'd say the supporting files should be consistent. Now this is what I've seen in other place in alpaka, so that's what I'm using too.

example/randomCells2D/CMakeLists.txt
sliwowitz: Doesn't seem that weird to me. You run tests, and you also check if the examples compile I guess.

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: fixed in next iteration

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: fixed in next iteration

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: fixed in next iteration

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: fixed in next iteration

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: fixed in next iteration

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: Maybe. But `std::memset` assigns bytes so I'd have to calculate the correct extent. `alpaka::memset` takes queue as a parameter so I guess it's supposed to be used on the device side?

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: Yes, it is. I'll use std::reduce in the 1D version. The 2D version is now more of a placeholder for some more interesting algorithm that would actually show something happening in 2D. I haven't figured out a nice simple but interesting example yet.

example/randomStrategies/src/randomStrategies.cpp
sliwowitz: Yeah. In Czechia, we're used to using '.' as decimal separator while programming or doing data analysis. Commas are too much trouble so you usually expect your files to have a decimal point (unless it comes from MS Excel). I use some "en" locale on my machines so I haven't encountered this for quite a while. I'll switch to TSV for now.

example/randomStrategies/src/randomStrategies.cpp
sliwowitz: fixed in next iteration

example/randomStrategies/src/randomStrategies.cpp
sliwowitz: I still don't like losing the type and treating `ptrBufHostResult` as a byte array. If someone copies the example with memset but tries to initialize to non-zero, they'd be in for a nasty surprise.

include/alpaka/rand/Philox/helpers/metaprog.hpp
sliwowitz: Yes. I didn't know where to keep these, I'll move it to alpaka/meta. I'd vote for mp11 (or something similar) though.

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
sliwowitz: I did it becuse all the other Philox implementations do it as well. I'll run some performance tests when the PR is ready.

include/alpaka/rand/Philox/PhiloxBaseCudaPlain.hpp
sliwowitz: fixed in next iteration

include/alpaka/rand/Philox/PhiloxSingle.hpp
sliwowitz: You're right. The cuintArray is still a temporary idea. It would simplify the implementation and possibly help to get a cleaner API, but I don't know if there are any performance or other unforeseen implications.

include/alpaka/rand/Philox/PhiloxSingle.hpp
sliwowitz: Well, we talked about it - std::array doesn't really work with CUDA, and this is just a tiny array with 4 numbers in it, so device arrays would be an overkill. I've looked into the cuRAND sources, and they seem to use the unit4 type as well.

include/alpaka/rand/Philox/PhiloxSingle.hpp
sliwowitz: No idea :-)

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: The issue here is that `engine()` won't return `std::array` on CUDA. Putting some sort of `size` member is a good idea. But this also touches the subject of *distributions* as they have to take the result of `engine()`, and we haven't decided what to do with the distributions yet.

example/randomCells2D/src/randomCells2D.cpp
bernhardmgruber: Ah, I see. Because the buffer could have a pitched allocation. Ok fine!

include/alpaka/rand/Philox/PhiloxBaseArrayLike.hpp
sliwowitz: I have no issues with renaming them. Any suggestions?

sliwowitz: There's more assumptions at play here - Counter has 4 elements, Key has 2 elements, and the elements are all 32 bit. I want to provide all four [reference Philox variants](https://www.thesalmons.org/john/random123/releases/latest/docs/group__PhiloxNxW.html) - with 32/64-bit elements, and a 2/4-element counter. (With a customizable number of rounds - 10 for a safe default, 7 for improved performance but less safety margin for those who know what they're doing).

I'll fix this when the other variants are added - that is after I validate the default one.

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: In the end, I won't use `std::reduce` even in the 1D version yet. The *hemera* toolchain seems convinced that `error: namespace "std" has no member "reduce"` 😢

bernhardmgruber: Funny. I had no problems with `std::reduce` when I ran on hemera 3 weeks ago with gcc 10.2 and C++17.

sliwowitz: thx. I was on gcc/9 because of cuda. I've somehow missed that I can go cuda/11.2 + gcc/10.2

include/alpaka/rand/Philox/PhiloxSingle.hpp
sliwowitz: I don't really see the problem here - you could forget to initialize the seed as is in the second example anyway.

include/alpaka/rand/Philox/mulhilo.hpp
sliwowitz: that's a good question - I probably just copied it from the reference implementation. changed it to return a 32bit number as expected

sliwowitz: changed to return unit32_t

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,244 @@
...
+#include <tuple>
+
+unsigned constexpr NUM_CALCULATIONS = 256;
+unsigned constexpr NUM_X = 1237;
+unsigned constexpr NUM_Y = 2131;
+
+/// Selected PRNG engine for single-value operation
+template<typename TAcc>
+// using RandomEngineSingle = alpaka::rand::engine::uniform_cuda_hip::Xor;
+using RandomEngineSingle = alpaka::rand::Philox4x32x10<TAcc>;
```
bernhardmgruber:
```suggestion
template<typename TAcc>
using RandomEngineSingle = alpaka::rand::Philox4x32x10<TAcc>;
// using RandomEngineSingle = alpaka::rand::engine::uniform_cuda_hip::Xor;
```

bernhardmgruber: swap those lines please for readability.

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -0,0 +1,244 @@
...
+
+        float sum = 0;
+        static_assert(NUM_CALCULATIONS % resultVectorSize == 0);
+        for(unsigned numCalculations = 0; numCalculations < NUM_CALCULATIONS / resultVectorSize; ++numCalculations)
+        {
+            auto result = dist(engine);
+            for(unsigned i = 0; i < resultVectorSize; ++i)
+            {
+                sum += result[i];
+            }
```
bernhardmgruber, line 114: My dream interface:

```suggestion
            for(auto result : dist(engine))
            {
                sum += result;
            }
```

Would need `begin`/`end` for your array types.

include/alpaka/meta/IsArrayOrVector.hpp
```diff
@@ -0,0 +1,111 @@
...
+         * @tparam T a type to check
+         */
+        template<typename T>
+        struct IsArrayOrVector
+        {
+            enum
+            {
+                value = false
+            };
+        };
```
bernhardmgruber:
```suggestion
        struct IsArrayOrVector : std::false_type {};
```

include/alpaka/meta/IsArrayOrVector.hpp
```diff
@@ -0,0 +1,111 @@
...
+         * @tparam A vector allocator
+         */
+        template<typename T, typename A>
+        struct IsArrayOrVector<std::vector<T, A>>
+        {
+            enum
+            {
+                value = true
+            };
+        };
```
bernhardmgruber:
```suggestion
        struct IsArrayOrVector<std::vector<T, A>> : std::true_type {};
```

bernhardmgruber: Applies to more places in this PR.

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
```diff
@@ -0,0 +1,93 @@
...
+                    Key key{key_in};
+                    Counter counter = TBackend::singleRound(counter_in, key);
+
+                    // TODO: Possible premature optimization. Check real performance.
+                    meta::RepeatN<numRounds>()(
+                        [&]()
+                        {
+                            key = TBackend::bumpKey(key);
+                            counter = TBackend::singleRound(counter, key);
+                        });
```
bernhardmgruber: Can we just have a simple loop here? If this is really a problem, we can always add a `#pragma unroll` later or unroll with metaprogramming as you do now. That would reduce the complexity a fair bit.

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
```diff
@@ -0,0 +1,93 @@
...
+                    // TODO: Should the key be returned as well??? i.e. should the original key be bumped?
```
bernhardmgruber, line 164: What about this? Does this still need to be addressed?

include/alpaka/rand/Philox/PhiloxBaseCudaPlain.hpp
bernhardmgruber: I still don't see where these are coming from :) Please add a link to the paper, website, etc.

include/alpaka/rand/Philox/PhiloxBaseStdArray.hpp
```diff
@@ -0,0 +1,48 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/rand/Philox/PhiloxBaseArrayLike.hpp>
+#include <alpaka/rand/Philox/mulhilo.hpp>
+
+//#include "alpaka_rand/rand/detail/PhiloxBaseArrayLike.hpp"
+
```
bernhardmgruber:
```suggestion
```

include/alpaka/rand/Philox/PhiloxConstants.hpp
```diff
@@ -0,0 +1,52 @@
...
+             * The numbers are taken from the reference Philox implementation:
+             *
+             * J. K. Salmon, M. A. Moraes, R. O. Dror and D. E. Shaw, "Parallel random numbers: As easy as 1, 2, 3,"
+             * SC '11: Proceedings of 2011 International Conference for High Performance Computing, Networking,
+             * Storage and Analysis, 2011, pp. 1-12, doi: 10.1145/2063384.2063405.
+             *
+             * @tparam TParams basic Philox algorithm parameters
+             */
+            template<typename TParams>
+            class PhiloxConstants
```
bernhardmgruber, line 35: How do these constants relate to the one in `include/alpaka/rand/Philox/PhiloxBaseCudaPlain.hpp`?

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,160 @@
...
+                 * @return The next random number
+                 */
+                ALPAKA_FN_HOST_ACC auto nextNumber()
+                {
+                    state.position++;
+                    if(state.position == TParams::counterSize)
+                    {
+                        advanceState();
+                    }
+#ifdef PLAIN_GPU_VECTOR_TYPES
```
bernhardmgruber: Where is this macro coming from?

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,160 @@
...
+                    case 0:
+                        return state.result.x;
+                    case 1:
+                        return state.result.y;
+                    case 2:
+                        return state.result.z;
+                    case 3:
+                        return state.result.w;
+                    default: // Should never get here
+                        return 0u;
```
bernhardmgruber: Consider adding an `assert(false && "...");`. I think we even have an alpaka specific assert.

include/alpaka/rand/Philox/helpers/cuintArray.hpp
```diff
@@ -0,0 +1,140 @@
...
+    using value_type = double;
+
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE double4_array(std::initializer_list<double> init)
+    {
+        auto it = std::begin(init);
+        x = *it++;
+        y = *it++;
+        z = *it++;
+        w = *it++;
+    }
```
bernhardmgruber: Theoretically, this cannot work because you mark the ctor `ALPAKA_FN_HOST_ACC` and `std::initializer_list` is not available in device code. Practically, nvcc just does whatever it wants.

include/alpaka/rand/Philox/helpers/cuintArray.hpp
```diff
@@ -0,0 +1,140 @@
...
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE constexpr double& operator[](int const k)
+    {
+        return k == 0 ? x : (k == 1 ? y : (k == 2 ? z : w));
+    }
+
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE const double& operator[](int const k) const
```
bernhardmgruber:
```suggestion
    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE constexpr const double& operator[](int const k) const
```

include/alpaka/rand/Philox/helpers/cuintArray.hpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+};
+
+
+/// Helper struct providing [] subscript access to CUDA double4 type
+struct double4_array : public double4
+{
+    using value_type = double;
+
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE double4_array(std::initializer_list<double> init)
```
bernhardmgruber:
```suggestion
    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE constexpr double4_array(std::initializer_list<double> init)
```

bernhardmgruber: Same for `uint4_array` and `uint2_array`.

include/alpaka/rand/Philox/helpers/cuintArray.hpp
bernhardmgruber: Same for `uint4_array` and `uint2_array`.

include/alpaka/rand/Philox/helpers/cuintArray.hpp
```diff
@@ -0,0 +1,140 @@
...
+    }
+    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE const uint32_t& operator[](int const k) const
+    {
+        return k == 0 ? x : y;
+    }
+};
+
+namespace std
+{
+    /// Specialization of std::tuple_size for \a float4_array
```
bernhardmgruber: Do we need that comment? It's obvious what is happening here.

include/alpaka/rand/RandPhilox.hpp
```diff
@@ -0,0 +1,231 @@
...
+
+            ALPAKA_FN_HOST_ACC UniformReal(T min, T max) : _min(min), _max(max), _range(_max - _min)
+            {
+            }
+
+            template<typename TEngine>
+            ALPAKA_FN_HOST_ACC auto operator()(TEngine& engine)
+            {
+                // Since it's possible to get a host-only engine here, the call has to go through proxy 
+                return static_cast<T>(EngineCallHostAccProxy<TEngine>{}(engine)) / engine.max() * _range + _min;
```
bernhardmgruber: I think we should not have this hack. Either the compiler yields a wrong diagnostic and we should silence it. Or you mixed up some host/device annotations. Or we need to document that host side engines cannot be passed to device side distributions etc. Or there is a bug.

include/alpaka/rand/RandStdLib.hpp
```diff
@@ -47,6 +48,9 @@ namespace alpaka
                 //! size of state: 19937 bytes
                 class MersenneTwister
                 {
+                public:
+                    std::mt19937 m_State;
+
```
bernhardmgruber: The entire changeset of this file looks like a refactoring that could come in a separate PR, probably before this one.

bernhardmgruber (changes requested):
It looks like you refactored the interface of existing RNGs to be STL compatible. That is good work! We should add a test for this however. E.g. using alpaka RNG engines with STL distributions. This refactoring should also be split off this PR and come as a separate PR.

This PR is already quite hard to review, if you can split anything else off as well and propose it via a separate PR, it would help a lot!

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: Agreed, but as you said, the PR is already quite long as it is. Introducing the uint4 et. al. array wrappers without the generator facility makes little sense as they are not currently used anywhere. On the other hand, this PR needs the wrappers in order to work on CUDA. I would opt for merging this PR first, and improving the wrappers later.

include/alpaka/rand/Philox/helpers/cuintArray.hpp
sliwowitz: These are doxygen comments. I can drop them of course but in general I think keeping them makes for better looking doxygen docs.

include/alpaka/meta/IsArrayOrVector.hpp
sliwowitz: done in next iteration

include/alpaka/meta/IsArrayOrVector.hpp
sliwowitz: done in next iteration

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: yes

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
sliwowitz: done. this will need some performance tests, but let's leave those for later

include/alpaka/rand/Philox/helpers/cuintArray.hpp
sliwowitz: done (will be in another PR)

include/alpaka/rand/Philox/helpers/cuintArray.hpp
sliwowitz: I don't know how to do that - constexpr constructor must initialize direct base class "uint4", and I want that class to be initialized with the values obtained by walking through the iterator. Also the underlying types do not have viable constructors, and are instead obtained via `make_uint4(x,y,z,w)` and similar functions. 

I'll look into that in a different PR

include/alpaka/rand/RandStdLib.hpp
sliwowitz: done

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
```diff
@@ -0,0 +1,172 @@
...
+    {
+        namespace engine
+        {
+            /** Philox algorithm parameters
+             *
+             * @tparam TCounterSize number of elements in the counter
+             * @tparam TWidth width of one counter element (in bits)
+             * @tparam TRounds number of S-box rounds
+             */
+            // TODO: Not sure where to put this
```
bernhardmgruber: Well, make a choice and then remove the comment please. I am sure the next person reading this will have even less of a clue.

include/alpaka/rand/Philox/PhiloxBaseCudaArray.hpp
```diff
@@ -0,0 +1,82 @@
...
+    {
+        namespace engine
+        {
+            namespace traits
+            {
+                template<typename TScalar>
+                struct PhiloxResultContainerTraits
+                {
+                    using type = void;
+                };
```
bernhardmgruber: I would prefer to either not define the type (just declare), or put a `static_assert` inside.

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,138 @@
...
+            struct PhiloxStateSingle
+            {
+                using Counter = TCounter;
+                using Key = TKey;
+
+                Counter counter; ///< Counter array
+                Key key; ///< Key array
+                Counter result; ///< Intermediate result array
+                std::uint32_t position; ///< Pointer to the active intermediate result element
+                // TODO: Box-Muller states
```
bernhardmgruber, line 38: What needs to be done here? Please remove the TODO or explain in more detail what is missing.

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -0,0 +1,138 @@
...
+            };
+
+            /** Philox engine generating a single number
+             *
+             * This engine's operator() will return a single number. Since the result is the same size as the counter,
+             * and so it contains more than one number, it has to be stored between individual invocations of
+             * operator(). Additionally a pointer has to be stored indicating which part of the result array is to be
+             * returned next.
```
bernhardmgruber, line 46: Is this really user facing documentation? This sounds more like notes for developers.

include/alpaka/rand/Philox/PhiloxVector.hpp
```diff
@@ -0,0 +1,115 @@
...
+                    this->advanceCounter(state.counter);
+                    return this->nRounds(state.counter, state.key);
+                }
+
+                /** Skips the next \a offset vectors
+                 *
+                 * Unlike its counterpart in \a PhiloxSingle, this function advances the state in multiples of the
+                 * counter size thus skipping the entire array of numbers.
+                 */
+                // XXX: Skips the whole vector! Is it enough to document this behavior?
```
bernhardmgruber: Please remove this comment. You documented the behavior.

include/alpaka/rand/RandPhilox.hpp
```diff
@@ -0,0 +1,232 @@
...
+            }
+            ALPAKA_FN_HOST_ACC constexpr result_type max()
+            {
+                return std::numeric_limits<result_type>::max();
+            }
+            ALPAKA_FN_HOST_ACC result_type operator()()
+            {
+                return engineVariant();
+            }
+            // TODO: https://en.cppreference.com/w/cpp/named_req/RandomNumberEngine
```
bernhardmgruber: Do you mean that the remaining features of the `RandomNumberEngine` concepts are missing? If yes, please say so here.

include/alpaka/rand/RandPhilox.hpp
```diff
@@ -0,0 +1,232 @@
...
+            }
+
+            ALPAKA_FN_HOST_ACC UniformReal(T min, T max) : _min(min), _max(max), _range(_max - _min)
+            {
+            }
+
+            template<typename TEngine>
+            ALPAKA_FN_HOST_ACC TResult operator()(TEngine& engine)
+            {
+                // FIXME: quick hack for 4-vectors, to be changed after API decision.
```
bernhardmgruber: So has the API decision been taken yet? Do we still need this FIXME?

include/alpaka/rand/RandPhilox.hpp
```diff
@@ -0,0 +1,232 @@
...
+                auto result = engine();
+                T scale = static_cast<T>(1) / engine.max() / _range;
```
bernhardmgruber:
```suggestion
                T scale = static_cast<T>(1) / (engine.max() * _range);
```

include/alpaka/rand/RandPhilox.hpp
```diff
@@ -0,0 +1,232 @@
...
+                return ret;
+            }
+
+        private:
+            const T _min;
+            const T _max;
+            const T _range;
+        };
+
+        /// TEMP: Distributions to be decided on later. The generator should be compatible with STL as of now.
```
bernhardmgruber: So is this finished now? Or work in progress? If it is the latter, you could remove it from the PR. Or document what is missing or needs be changed later. If the work is finished, then please remove the comment.

bernhardmgruber (changes requested):
You have a lot of TODOs left in your code. Please have a look at them and either remove them if they are fixed, or specify more clearly what needs to be done. It might just happen that someone else needs to work on this code at some point, so it should be clear what is missing.

include/alpaka/rand/Philox/PhiloxSingle.hpp
sliwowitz: This is the doxygen documentation. I was about to move this somewhere else, but then we talked about doxygen in the meeting and it seemed that these docs (where they exist at all) are targeted at the developers as well.

include/alpaka/rand/RandPhilox.hpp
sliwowitz: No decision has been made - we barely started a discussion :-) I'll remove the comment and create a github issue instead

include/alpaka/rand/RandPhilox.hpp
sliwowitz: wow. this actually should have been  `T scale = static_cast<T>(1) / engine.max() * _range;` no idea how I got the `/` there at all.

include/alpaka/rand/Philox/PhiloxSingle.hpp
bernhardmgruber: After today's meeting, this is fine I guess!

include/alpaka/rand/Philox/PhiloxBaseArrayLike.hpp
bernhardmgruber: Ok.

sliwowitz: Actually I remembered this conversation, forgot I didn't want to rename these, and renamed them anyway in the current version :-D Now the master function is `multiplyAndSplit64to32` and then there's `high32Bits` and `low32Bits`.

bernhardmgruber (approved):

bernhardmgruber:
@jkelling I am fine with merging this. Does one of your PRs needs to come first or can you rebase on this one?

bernhardmgruber:
I contacted @jkelling and @sliwowitz offline and they are fine with merging this one first.

PR #1320, opened 2021-05-26
fix HIP compiler options not recognized (backport to 0.6.1)

psychocoderHPC:
This PR is against 0.6.1, the fix for dev and 0.7.0-rc1 will follow soon.

Compiler options for hipcc must be provided via the cmake variable HIP_HIPCC_FLAGS. CMAKE_CXX_FLAGS is not taken into account.

Bug was found with PIConGPU: https://github.com/ComputationalRadiationPhysics/picongpu/issues/3624#issuecomment-848517408

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -853,7 +853,8 @@ if(TARGET alpaka)
 endif()
 
 # NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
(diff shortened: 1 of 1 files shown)


j-stephan (approved):

j-stephan:
It seems that MSVC 2019 doesn't like its own `[[nodiscard]]` functions...

psychocoderHPC:
ready to merge, one CI test failed because the macOSX instance not started correctly

bernhardmgruber (approved):

PR #1322, opened 2021-05-26
set CMAKE_BUILD_TYPE

psychocoderHPC:
alpaka is always building with a undefined CMAKE_BUILD_TYPE.
This PR sets `Release` as default build type and provides selections for `Debug` and `RelWithDebInfo`.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -14,6 +14,14 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
 #-------------------------------------------------------------------------------
 # Options.
 
+# CMake build option
+set(alpaka_BUILD_TYPE "Release;RelWithDebInfo;Debug")
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the build type for the project" FORCE)
+endif()
+set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS "${alpaka_BUILD_TYPE}")
+unset(alpaka_BUILD_TYPE)
+
 # Compiler options
 function(alpaka_compiler_option name description default)
     if(NOT DEFINED name)
@@ -772,7 +780,8 @@ if(TARGET alpaka)
 endif()
 
 # NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
-if(ALPAKA_ACC_GPU_HIP_ENABLE AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
+# For hipcc we need to propagate the CMAKE_CXX_FLAGS to HIP_HIPCC_FLAGS.
+if(ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE)
     get_property(_ALPAKA_COMPILE_OPTIONS_PUBLIC
                  TARGET alpaka
                  PROPERTY INTERFACE_COMPILE_OPTIONS)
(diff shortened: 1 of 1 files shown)


cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,14 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
 #-------------------------------------------------------------------------------
 # Options.
 
+# CMake build option
+set(alpaka_BUILD_TYPE "Release;RelWithDebInfo;Debug")
```
bernhardmgruber, line 18: Is `MinSizeRel` omitted on purpose?

cmake/alpakaCommon.cmake
```diff
@@ -14,6 +14,14 @@ include(CMakePrintHelpers) # for easier printing of variables and properties
...
+if(NOT CMAKE_BUILD_TYPE)
+    set(CMAKE_BUILD_TYPE "Release" CACHE STRING "Choose the build type for the project" FORCE)
+endif()
```
bernhardmgruber, line 21: Does that mean that if I `find_package(alpaka)` from a LLAMA example, which does not specify a `CMAKE_BUILD_TYPE`, alpaka will set the build type of my LLAMA example? Because that might be invasive.

j-stephan:
Which issue are you trying to solve here? The build type would only affect our test cases and examples, not alpaka itself.

cmake/alpakaCommon.cmake
psychocoderHPC: `MinSizeRel` in to an option in CMake.

psychocoderHPC:
> Which issue are you trying to solve here? The build type would only affect our test cases and examples, not alpaka itself.

It will be set if you use alpaka with `add_subdirectory, during the installation for tests and examples but if alpaka it installed it should be used too. The build type is only set if the app or user is not setting the type before.

cmake/alpakaCommon.cmake
psychocoderHPC: If LLAMA is setting something else as buildtype and you define your own build types your app setting will not be overwritten.
If your app is keeping CMake in an undefined build type it will be set by alpaka.

cmake/alpakaCommon.cmake
psychocoderHPC: Sry my fault, `CMAKE_CXX_FLAGS_MINSIZEREL` is a valid variable and therefore `MinSizeRel`. 
`MinSizeRel` should be added to this PR too.

bernhardmgruber: I am sorry, but I do not understand your comment :S

cmake/alpakaCommon.cmake
bernhardmgruber: LLAMA is not setting anything, because the build type is up to the user.
> If your app is keeping CMake in an undefined build type it will be set by alpaka.
So alpaka is injecting a build type now, depending on whether the LLAMA examples find it or not. I find that dangerous.

I would like to propose to set the build type only for alpaka tests and examples and not when the library is included.

psychocoderHPC: > I would like to propose to set the build type only for alpaka tests and examples and not when the library is included.

We can but this means that your application is not compiled with a defined CMake build type.

@j-stephan Do you also think we should keep the undefined behavior?

j-stephan: Well, the CMake developers don't seem opposed to having a default type: https://blog.kitware.com/cmake-and-the-default-build-type/

So I'm not opposed to this.

We just need to make sure that this applies to other generators (MSVC...), too, that don't obey `CMAKE_BUILD_TYPE`.

cmake/alpakaCommon.cmake
```diff
@@ -772,7 +780,8 @@ if(TARGET alpaka)
 endif()
 
 # NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
-if(ALPAKA_ACC_GPU_HIP_ENABLE AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
+# For hipcc we need to propagate the CMAKE_CXX_FLAGS to HIP_HIPCC_FLAGS.
+if(ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE)
```
j-stephan, line 784: How did this get here? Isn't this already part of #1324?

psychocoderHPC: You are right, this should not be part of this PR, maybe I branched from the wrong branch.

Let us decide if we set a default type or not, depending on the result I will remove this section or we close the PR.

cmake/alpakaCommon.cmake
bernhardmgruber: > We can but this means that your application is not compiled with a defined CMake build type.

That is true and it is usually bad. But I think alpaka should not fix this for downstream projects, because it is surprising.

However, if you truly want this then go ahead. I don't mind much for LLAMA. Let's see if someone else complains later.

Maybe the new `CMakePresets.txt` could also be used to address this.

psychocoderHPC:
I will close this PR, let us maybe later decide if we should fix this strange behavior.

PR #1321, opened 2021-05-26
fix HIP compiler options not recognized (backport 0.7.0-rc1)

psychocoderHPC:
Backport to 0.7.1-rc1 similar to #1324

Compiler options for hipcc must be provided via the cmake variable HIP_HIPCC_FLAGS. CMAKE_CXX_FLAGS is not taken into account.

PR #1324, opened 2021-05-26
fix HIP compiler options not recognized

psychocoderHPC:
Compiler options for hipcc must be provided via the cmake variable
HIP_HIPCC_FLAGS. CMAKE_CXX_FLAGS is not taken into account.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -771,8 +771,8 @@ if(TARGET alpaka)
     endif()
 endif()
 
-# NVCC does not incorporate the COMPILE_OPTIONS of a target but only the CMAKE_CXX_FLAGS
-if(ALPAKA_ACC_GPU_HIP_ENABLE AND ALPAKA_CUDA_COMPILER MATCHES "nvcc")
+# For hipcc we need to propagate the CMAKE_CXX_FLAGS to HIP_HIPCC_FLAGS.
+if(ALPAKA_ACC_GPU_HIP_ENABLE)
     get_property(_ALPAKA_COMPILE_OPTIONS_PUBLIC
                  TARGET alpaka
                  PROPERTY INTERFACE_COMPILE_OPTIONS)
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
https://github.com/alpaka-group/alpaka/pull/1321#discussion_r639556414 should be addressed

j-stephan:
By the way: Have you seen that HIP can now be linked in a much easier way?

https://rocmdocs.amd.com/en/latest/Installation_Guide/Using-CMake-with-AMD-ROCm.html

`target_link_libraries(alpaka INTERFACE hip::host hip::device)`

psychocoderHPC:
> By the way: Have you seen that HIP can now be linked in a much easier way?
> 
> https://rocmdocs.amd.com/en/latest/Installation_Guide/Using-CMake-with-AMD-ROCm.html
> 
> `target_link_libraries(alpaka INTERFACE hip::host hip::device)`

No I have not seen it, thanks for the link! 
Never the less I would push this PR first and at a later point check a refactoring.
If we use targets it is not clear if all works out of the box or if something requires workarounds.

psychocoderHPC:
@j-stephan THis PR is now only handling `hipcc`. I will check the usage of targets for `hipcc` separately.

cmake/alpakaCommon.cmake
```diff
@@ -771,18 +771,22 @@ if(TARGET alpaka)
...
+if(ALPAKA_ACC_GPU_HIP_ENABLE)
     get_property(_ALPAKA_COMPILE_OPTIONS_PUBLIC
                  TARGET alpaka
                  PROPERTY INTERFACE_COMPILE_OPTIONS)
     string(REPLACE ";" " " _ALPAKA_COMPILE_OPTIONS_STRING "${_ALPAKA_COMPILE_OPTIONS_PUBLIC}")
     alpaka_set_compiler_options(HOST var CMAKE_CXX_FLAGS "${_ALPAKA_COMPILE_OPTIONS_STRING}")
 
     # Append CMAKE_CXX_FLAGS_[Release|Debug|RelWithDebInfo] to CMAKE_CXX_FLAGS
-    # because FindCUDA only propagates the latter to nvcc.
+    # because cmake HIP modules do not use this variables
```
```suggestion
    # because cmake HIP modules do not use the former variables
```

j-stephan (dismissed):
Looks good to me, maybe one suggestion:

j-stephan (approved):

ax3l:
> No I have not seen it, thanks for the link!

Hah, I wrote that to you on Slack on May 27th, too :laughing: 

Be aware of https://github.com/ROCm-Developer-Tools/HIP/pull/2280, landing in ROCm 4.4 as it seems.

psychocoderHPC:
> Hah, I wrote that to you on Slack on May 27th, too 

Could be but this information got lost because my brain runs out of memory :exploding_head:

psychocoderHPC:
> Be aware of ROCm-Developer-Tools/HIP#2280, landing in ROCm 4.4 as it seems.

If I understand the problem this should not be too important for our projects because we do not mix languages.

What I need to check is if this warning
```
clang-12: warning: argument unused during compilation: '-amdgpu-function-calls=false'
```
is also gone with the linked PR, else I need to open an issue for HIP.

PR #1327, opened 2021-05-26
Add missing <limits> header and fix integer namespace

j-stephan:
This PR fixes a compilation error in `block/shared/st/detail/BlockSharedMemStMemberImpl.hpp` which appeared on my system while only the TBB back-end was enabled (`numeric_limits` is no member of `std`).

While doing this I noticed that only some of the fixed-size integers were correctly assigned to the `std` namespace. This is now fixed as well.

Once merged, this also needs to be backported to 0.6.1 and 0.7.

### DIFF
--- a/include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
+++ b/include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
@@ -16,6 +16,7 @@
 #include <algorithm>
 #include <cstdint>
 #include <functional>
+#include <limits>
 #include <type_traits>
 
 namespace alpaka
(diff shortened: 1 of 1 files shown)


BenjaminW3 (dismissed):

psychocoderHPC:
@j-stephan It looks like we need to remove our shipped cmake TBB files too (as you mentioned yesterday offline) to pass the CI: https://github.com/alpaka-group/alpaka/pull/1327/checks?check_run_id=2679766501#step:5:1716

bernhardmgruber:
> @j-stephan It looks like we need to remove our shipped cmake TBB files too (as you mentioned yesterday offline) to pass the CI: https://github.com/alpaka-group/alpaka/pull/1327/checks?check_run_id=2679766501#step:5:1716

I used `find_package(TBB)` in LLAMA without providing a find module and that worked. So I guess you can drop alpaka's one.

psychocoderHPC:
> > @j-stephan It looks like we need to remove our shipped cmake TBB files too (as you mentioned yesterday offline) to pass the CI: https://github.com/alpaka-group/alpaka/pull/1327/checks?check_run_id=2679766501#step:5:1716
> 
> I used `find_package(TBB)` in LLAMA without providing a find module and that worked. So I guess you can drop alpaka's one.

This depends on the system. Unluckily not all systems shipping the find CMake modules.
#1329 should fix the TBB issues.

psychocoderHPC (dismissed):

bernhardmgruber (dismissed):

bernhardmgruber (approved):

bernhardmgruber (dismissed):

psychocoderHPC:
PR can be rebased against develop to fix the CI issues.

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #1336, opened 2021-06-07
Add implementation of complex numbers

### LINKED ISSUE #734: Abstraction of complex numbers
As I am working on the second project by now that requires the use of complex numbers in an alpaka kernel, I agreed with @psychocoderHPC , @tdd11235813 and @sbastrakov that it probably is a good idea to provide some abstractions for using complex numbers in alpaka.
This is due to the following. By now, there is a standardized interface for complex numbers in `std::complex<RealType>` which ships with abs, exp and trigonometric functions. Unfortunately, the STL version of std::complex is not usable in CUDA code as all its functionality is not marked as a `__device__` function. However, in [thrust/complex.h][thrust-complex], there exists an implementation which is suitable for all accelerators and furthermore binary compatible to the `std::complex` implementations (and other complex types like `cuComplex/cuDoubleComplex`). The only problem is that this is only available by default when CUDA is enabled. So, it is not possible to use this directly when only relying on alpaka by now.
To solve this, there are two potential solutions:
* Provide a copy of a modified [thrust/complex.h][thrust-complex] wrapped in alpaka. This is the preferable solution we agreed on initially, but I inspected it and found out that it is not a mere copy of std::complex with `__device__` added, but a complete reimplementation of all the functions including the trigonometric ones. So this might be problematic due to license issues.
* Provide an easy solution to switch between `std::complex` and `thrust::complex`. This would look like some trait deciding which one to use depending on the accelerator accompanied by complex math functions like they already exist in the form of `alpaka::math`. This is possible as the classes std::complex and thrust::complex share the same interface. But, this might fail for future accelerators as there might be platforms where none of the above implementations is available.

Feel free to add some other solutions.
I have already implemented the second for a small project, but it is very incomplete as I only needed the base functionality of the complex class itself, two functions and two accelerator types.
What do you think?

[thrust-complex]: https://github.com/thrust/thrust/blob/master/thrust/complex.h

sbastrakov:
This PR adds implementation of alpaka's own Complex class template. It is basically `std::complex<T>` but its methods also work on device. Math functions are (WIP now) done via alpaka traits, same way as for real numbers.

TODO:
- [x] Math traits. For now I've done only one for Abs, just to demonstrate how I would like to do it, and to discuss.
- [x] Unit tests.
- [x] Figure out what to do with SYCL math implementation.

Resolve #734.


include/alpaka/math/abs/AbsStdLib.hpp
```diff
@@ -29,17 +30,15 @@ namespace alpaka
         {
             //! The standard library abs trait specialization.
             template<typename TArg>
-            struct Abs<
-                AbsStdLib,
-                TArg,
-                std::enable_if_t<std::is_arithmetic<TArg>::value && std::is_signed<TArg>::value>>
```
sbastrakov: Note: these requirements did not work for `std::complex<T>` either, I think they were just wrong, but didn't make any difference before.

include/alpaka/math/abs/AbsStdLib.hpp
```diff
@@ -29,17 +30,15 @@ namespace alpaka
...
+            struct Abs<AbsStdLib, TArg>
             {
-                ALPAKA_FN_HOST auto operator()(AbsStdLib const& abs_ctx, TArg const& arg)
+                ALPAKA_FN_HOST auto operator()(AbsStdLib const& abs_ctx, ConvertedArg<TArg> const& arg)
```
sbastrakov: So the idea for the host side is to convert `alpaka::Complex<T>` into `std::complex<T>` and just call the standard functions like we do for real numbers. Unfortunately, it didn't work as is, since typecast didn't get along with template substitution. To avoid having to specialize for `Complex` separately (which would have been needed for each math function), I've wrapped this type.

include/alpaka/math/abs/AbsStdLib.hpp
```diff
@@ -29,17 +30,15 @@ namespace alpaka
...
                 {
                     alpaka::ignore_unused(abs_ctx);
                     return std::abs(arg);
                 }
             };
+
```
sbastrakov: Accidental change

include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
```diff
@@ -83,6 +84,16 @@ namespace alpaka
...
+            //! The CUDA built in abs Complex specialization.
+            template<typename T>
+            struct Abs<AbsUniformCudaHipBuiltIn, Complex<T>>
+            {
+                __device__ auto operator()(AbsUniformCudaHipBuiltIn const& abs_ctx, Complex<T> const& arg)
+                {
+                    alpaka::ignore_unused(abs_ctx);
+                    return ::sqrt(arg.real() * arg.real() + arg.imag() * arg.imag());
+                }
+            };
```
sbastrakov: Since core CUDA doesn't have any complex type (only CUDA libraries do), we will have to just implement everything manually

sbastrakov: Probably `::sqrt` is wrong for `T == float`, and I should have called alpaka's `sqrt()` instead

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,333 @@
...
+        //! Get the real part
+        constexpr ALPAKA_FN_HOST_ACC T real() const
+        {
+            return m_real;
+        }
+
+        //! Stub for get the real part (does nothing, added for compatibility with std::complex)
+        constexpr ALPAKA_FN_HOST_ACC void real(T)
+        {
+        }
```
bernhardmgruber: I think that is supposed to be the setter, so you are missing a `m_real = ...;`. See: https://en.cppreference.com/w/cpp/numeric/complex/real
Same for `imag()`.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,333 @@
...
+            return *this;
+        }
+
+        //! Division assignment with a complex number
+        template<typename U>
+        ALPAKA_FN_HOST_ACC Complex& operator/=(Complex<U> const& other)
+        {
+            return *this *= Complex{
+                       static_cast<T>(other.real() / (other.real() * other.real() + other.imag() * other.imag())),
+                       static_cast<T>(-other.imag() / (other.real() * other.real() + other.imag() * other.imag()))};
```
bernhardmgruber, line 169: Here is a subtlety: part of the computation is done in Us, another part in Ts. I think this needs some thinking what is right here.

include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
SimeonEhrig: Do you mean with cuda complex type a class or struct (I think so)? There are a lot of complex types in CUDA, but every complex type is a alias of float2.

bernhardmgruber (commented):
I am wondering whether the complex type should be templated on the accelerator. Because then you could just use `std::complex` everywhere on the CPU and only supply a different implementation for CUDA/HIP. I think this is also what @sliwowitz does with his RNG. AFAIK there is a different Philox implementation depending on the Acc.

include/alpaka/math/Complex.hpp
sbastrakov: True, I was thinking of smth else. Thanks for pointing out

include/alpaka/math/Complex.hpp
sbastrakov: I think it's supposed to be mixed in cppreference definition. As an alternative, I could first cast the other argument to `Complex<T>` and call the same method with matching `T`.

include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
sbastrakov: There are in CUDA libraries, not in CUDA itself as far as I understand

sbastrakov:
During the VC @bernhardmgruber suggested if it is woth to have `Acc` as an additional template parameter of `Complex`. Then it would be possible to partially specialize there, maybe for CUDA/HIP enabled vs. disabled. Then device-side functions would change from `Complex<T>` like now to `Complex<TAcc, T>` or `TComplex`, and generally it becomes futher away from `std::complex<T>`

sbastrakov:
Notes from @SimeonEhrig on the VC. In the example we need to demostrate more: how to cast between buffer of `alpaka::Complex` and `std::complex`, how to call math operations on host (either cast to `std::complex` and call, or have a kernel called on `Host` acc). I agree, will extend the example.

include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
sbastrakov: There is indeed `cuComplex` which is an alias for `float2`. Weirdly, it is not searchable in the programming guide.

include/alpaka/math/abs/AbsStdLib.hpp
sbastrakov: Ah, now after CI failed I understand the reason. After I removed that, some compilers issue a warning that `std::abs` call was removed as it is called for unsigned.

include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
SimeonEhrig: Attention, there are more complex types in the CUDA libraries, e.g. `cuFFT`
```C
typedef cuComplex cufftComplex;
typedef cuDoubleComplex cufftDoubleComplex;
```

But should be all a typedef of `float2`.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,333 @@
...
+        template<typename U>
+        ALPAKA_FN_HOST_ACC Complex& operator+=(Complex<U> const& other)
+        {
+            m_real += static_cast<T>(other.real());
+            m_imag += static_cast<T>(other.imag());
+            return *this;
+        }
+
+        //! Subtraction assignment with a real number
+        ALPAKA_FN_HOST_ACC Complex& operator-=(T const& other)
```
psychocoderHPC, line 121: I think we can define the operators constexpr too?

sbastrakov: I defined same as `std::complex` in cppreference for C++14. It says these operators will become `constexpr` since C++20.

psychocoderHPC: Sry, I missed that this is the `operator-=` and others are defined constexpr.

include/alpaka/math/ConvertedArg.hpp
```diff
@@ -0,0 +1,55 @@
...
+            } // namespace detail
+
+            //! Wrapper for argument types in StdLib implementations of math functions
+            //!
+            //! Defines the type of the argument after conversion (if conversion is necessary).
+            //! Helps provide conversion and avoid duplication of implementations.
+            //!
+            //! \tparam TArg original argument type
+            template<typename TArg>
+            using ConvertedArg = typename detail::ConvertedArg<TArg>::type;
```
BenjaminW3: From the name it is not clear for the user what is converted to which type and why? When should the user specialize this trait?
Some more explicit naming might help.

sbastrakov: Yes, I was struggling to find a fitting name here. It's also not for the users of alpaka, only for its developers.

SimeonEhrig:
I tested complex numbers with vikunja: https://github.com/alpaka-group/vikunja/pull/54
The API usage is nice and it works well with the CPU backend. But I have a problem with the CUDA backend. If I compile it for the CUDA backend, I get the following error:

```
/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/cuda-11.0.2-mtxiuzc35ykbmjk4fnpmhwnjbgquvcv5/bin/nvcc -forward-unknown-to-host-compiler -ccbin=/usr/bin/c++ -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=47 -DALPAKA_DEBUG=0 -DALPAKA_DEBUG_OFFLOAD_ASSUME_HOST -DALPAKA_OFFLOAD_MAX_BLOCK_SIZE=256 -DBOOST_ALL_NO_LIB -I../vikunja/include -I/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/cuda-11.0.2-mtxiuzc35ykbmjk4fnpmhwnjbgquvcv5/targets/x86_64-linux/include -I../alpaka/include -isystem=/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/boost-1.74.0-lqyr7ou7iafiwvq5fb6ipp62o6l2gqlf/include -isystem=/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/cuda-11.0.2-mtxiuzc35ykbmjk4fnpmhwnjbgquvcv5/include --generate-code=arch=compute_52,code=[compute_52,sm_52] -Xcompiler=-fPIE --expt-extended-lambda --expt-relaxed-constexpr -Xcudafe=--display_error_number -Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored -Xcompiler -pthread -std=c++17 -MD -MT CMakeFiles/complexVikunja.dir/main.cpp.o -MF CMakeFiles/complexVikunja.dir/main.cpp.o.d -x cu -c ../main.cpp -o CMakeFiles/complexVikunja.dir/main.cpp.o
../alpaka/include/alpaka/math/abs/Traits.hpp(54): error #3050: calling a __device__ function("operator()") from a __host__ __device__ function("abs") is not allowed
          detected during instantiation of "auto alpaka::math::abs(const T &, const TArg &) [with T=alpaka::AccGpuCudaRt<alpaka::DimInt<1UL>, uint64_t>, TArg=alpaka::Complex<float>]" 
../main.cpp(114): here

../alpaka/include/alpaka/math/abs/Traits.hpp(54): error #3050: calling a __device__ function("operator()") from a __host__ __device__ function("abs") is not allowed
          detected during instantiation of "auto alpaka::math::abs(const T &, const TArg &) [with T=alpaka::AccGpuCudaRt<alpaka::DimInt<1UL>, uint64_t>, TArg=alpaka::Complex<float>]" 
../main.cpp(114): here

2 errors detected in the compilation of "../main.cpp".
ninja: build stopped: subcommand failed.
```

This line is causing the error
https://github.com/alpaka-group/vikunja/blob/28c82eada54e042adc316a57c9ef172c21bc1834/example/complex/main.cpp#L117

The problem has to do with the specialization of the `acc` object. Maybe I use it wrong.

sbastrakov:
Thanks for reporting. I will try.

psychocoderHPC:
I discussed @SimeonEhrig issue with him offline, it could be he must change https://github.com/alpaka-group/vikunja/blob/28c82eada54e042adc316a57c9ef172c21bc1834/example/complex/main.cpp#L116 to

```C++
auto transform = [] ALPAKA_FN_HOST_ACC(auto const& acc, alpaka::Complex<Data> const& a) -> Data
```

to enforce the compile to evaluate the lambda on the device side only. IMO the issue is coming from the two-phase compiling of the cuda compiler.

SimeonEhrig:
After an additional discussion with @psychocoderHPC we found a solution. Now I use a `struct` with templated `operator()` function for the transform (I was not aware, that is already working :sweat_smile: ). Now the template injection is working correctly and the whole application too.

I uploaded the change: https://github.com/alpaka-group/vikunja/blob/77ed0ce48b33721d3d5a90705968817682a98121/example/complex/main.cpp#L19

Therefore, I'm fine with the complex numbers.

sbastrakov:
Just pushed a mostly finished version. Will hopefully fix the failing tests and do some polish tomorrow, also will clean up commit history.

include/alpaka/math/ConvertedArg.hpp
sbastrakov: This is no longer needed, removed in the current version.

sbastrakov:
@SimeonEhrig there was an issue in my original code that caused a mixup of host and host-device math. Now I believe it is fixed for everything but `arg` and `conj`, however so far i do not see what's exactly wrong with them

So perhaps your original code was fine and the error was on my side, I do not know.

SimeonEhrig:
> @SimeonEhrig there was an issue in my original code that caused a mixup of host and host-device math. Now I believe it is fixed for everything but `arg` and `conj`, however so far i do not see what's exactly wrong with them

Looks like, I was not affected of this bug. Nevertheless I tested you latest commit (bd35e7beb96305d3b51284204aeef9584c2c20ce) with my vikunja example and following acc's successful.
* AccCpuSerial
* AccCpuOmp2Blocks
* AccGpuCudaRt
* AccGpuHipRt

include/alpaka/math/arg/Traits.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+        //! The math traits.
+        namespace traits
+        {
+            //! The arg trait.
+            template<typename T, typename TArgument, typename TSfinae = void>
+            struct Arg
+            {
+                //!@todo why is this needed here, but not in other functions?
+                ALPAKA_NO_HOST_ACC_WARNING
```
sbastrakov: I must be missing something very simple. But i have no idea why is there a warning generated here about host and device functions (unless I add a guard). The function works as intended (on host and device, for real, complex, and other types (the other are with ADL)), as tests show and I also checked manually. And in similar situations other functions do not generate such a warning. If anyone has an idea, it's welcome.

Btw the same is for another new math function/trait added in this PR, `conj()`

test/unit/math/src/math.cpp
```diff
@@ -125,12 +133,47 @@ struct TestTemplate
...
+
+    //! Approximate comparison of complex numbers
+    template<typename T>
+    static bool isApproxEqual(alpaka::Complex<T> const& a, alpaka::Complex<T> const& b)
+    {
+        // Consider all infinite values equal, @see comment at isFinite()
+        if(!isFinite(a) && !isFinite(b))
+            return true;
+        // For the same reason use relative difference comparison with quite a large epsilon
+        auto const eps = static_cast<T>(1.0e4) * std::numeric_limits<T>::epsilon();
```
sbastrakov: Oh damn, I forgot to change back this part

j-stephan:
Shifting to 0.9. This is a feature too important to put into a release on such short notice.

psychocoderHPC:
@sbastrakov Please do not forget to rebase before pushing updates.

j-stephan:
What is the state here?

psychocoderHPC:
@sbastrakov Could you finish this PR?

sbastrakov:
I updated the old PR state to match the current alpaka `develop`. Math tests build and run on the dev system. However I believe the previously encountered issues with accuracy threshold for unit tests on some CI configurations still remain (since i didn't do anything to fix them yet).

sbastrakov:
SYCL backend is not updated yet, added a checkbox.

sbastrakov:
CI passed, I just need to clean up history a bit, since it took some effort to make it pass.

Ready for review @j-stephan @bernhardmgruber

example/complex/CMakeLists.txt
```diff
@@ -0,0 +1,60 @@
+#
+# Copyright 2014-2022 Erik Zenker, Benjamin Worpitz, Jan Stephan, Sergei Bastrakov
```
bernhardmgruber:
```suggestion
# Copyright 2022 Erik Zenker, Benjamin Worpitz, Jan Stephan, Sergei Bastrakov
```

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+namespace alpaka
+{
+    //! Implementation of a complex number useable on host and device.
+    //!
+    //! It follows the layout of std::complex and so array-oriented access.
+    //! Inside the class template implements all methods and operators as std::complex<T>.
+    //! Additionally, provides an implicit conversion to and from std::complex<T>.
+    //! All methods besides operators << and >> are host-device.
+    //! Note that it does not provide non-member functions of std::complex besides the operators.
+    //! Those are provided the same way as alpaka math functions for real numbers.
```
bernhardmgruber:
```suggestion
    //! It follows the layout of std::complex and so array-oriented access.
    //! The class template implements all methods and operators as std::complex<T>.
    //! Additionally, it provides an implicit conversion to and from std::complex<T>.
    //! All methods besides operators << and >> are host-device.
    //! It does not provide non-member functions of std::complex besides the operators.
    //! Those are provided the same way as alpaka math functions for real numbers.
```

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+    //!
+    //! Note that unlike most of alpaka, this is a concrete type template, and not merely a concept.
```
bernhardmgruber, line 31: I would not need that comment. But you can keep it.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+    //!
+    //! Naming and order of the methods match https://en.cppreference.com/w/cpp/numeric/complex.
+    //!
+    //! @tparam T type of the real and imaginary part: float, double, or long double.
+    template<typename T>
+    class Complex
+    {
```
bernhardmgruber, line 42:
```suggestion
    //! @tparam T type of the real and imaginary part: float, double, or long double.
    template<typename T>
    class Complex
    {
        static_assert(std::is_floating_point_v<T>);
```
Maybe enforce that `T` is a floating point in addition to the comment.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+        //! Conversion to std::complex
+        constexpr ALPAKA_FN_HOST_ACC operator std::complex<T>() const
+        {
+            return std::complex<T>{m_real, m_imag};
+        }
+
+        //! Assignment
+        ALPAKA_FN_HOST_ACC Complex& operator=(Complex const&) = default;
+
```
bernhardmgruber, line 79:
```suggestion
```
Just use the compiler provided one. Or are you getting a warning if you don't provide a defaulted one?

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+        {
+            m_imag = value;
+        }
+
+        //! Addition assignment with a real number
+        ALPAKA_FN_HOST_ACC Complex& operator+=(T const& other)
+        {
+            m_real += other;
+            return *this;
+        }
```
bernhardmgruber, line 109: Why not `constexpr` the operators as well?

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+            return *this *= Complex{
+                       static_cast<T>(other.real() / (other.real() * other.real() + other.imag() * other.imag())),
+                       static_cast<T>(-other.imag() / (other.real() * other.real() + other.imag() * other.imag()))};
+        }
+
+    private:
+        //! Real and imaginary parts, storage enables array-oriented access
+        T m_real, m_imag;
+    };
+
```
bernhardmgruber, line 176: Maybe add:
```c++
#ifdef __cpp_lib_is_layout_compatible
static_assert(std::is_layout_compatible_v<Complex<float>, std::complex<float>>);
#endif
```
here or in one of the unit tests.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+    //!
+    //! @{
+    //!
+
+    //! Unary plus (added for compatibility with std::complex)
+    template<typename T>
+    ALPAKA_FN_HOST_ACC Complex<T> operator+(Complex<T> const& val)
+    {
+        return val;
+    }
```
bernhardmgruber, line 189: Prefer to define the arithmetic operators as hidden friends inside the class. Applies to all operators here.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+    //! Inequality of two complex numbers.
+    //!
+    //! @note this and other versions of operator != should be removed since C++20, as so does std::complex
+    template<typename T>
+    constexpr ALPAKA_FN_HOST_ACC bool operator!=(Complex<T> const& lhs, Complex<T> const& rhs)
+    {
+        return !math::floatEqualExactNoWarning(lhs.real(), rhs.real())
+            || !math::floatEqualExactNoWarning(lhs.imag(), rhs.imag());
```
bernhardmgruber:
```suggestion
        return !(lhs == rhs);
```
Reuse `==` when implementing `!=`. Applies to the operator `!=` implementations as well.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+
+    //! @}
+
+    //! Host-only output of a complex number
+    template<typename T, typename TChar, typename TTraits>
+    std::basic_ostream<TChar, TTraits>& operator<<(std::basic_ostream<TChar, TTraits>& os, Complex<T> const& x)
+    {
+        os << x.operator std::complex<T>();
+        return os;
+    }
```
bernhardmgruber, line 345: You really want to support wide streams? :D

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+        }
+
+        //! Addition assignment with a complex number
+        template<typename U>
+        ALPAKA_FN_HOST_ACC Complex& operator+=(Complex<U> const& other)
+        {
+            m_real += static_cast<T>(other.real());
+            m_imag += static_cast<T>(other.imag());
+            return *this;
+        }
```
bernhardmgruber, line 118:
```suggestion
        ALPAKA_FN_HOST_ACC Complex& operator+=(Complex const& other)
        {
            m_real += other.real();
            m_imag += other.imag();
            return *this;
        }
```
For `operator+=` you support converting from complex numbers with different FP type, but e.g. `operator+` does not support his conversion. I would keep it stricter and only allow a matching complex number in arithmetic operations.

include/alpaka/math/Complex.hpp
```diff
@@ -0,0 +1,540 @@
...
+        //! Copy constructor
+        constexpr ALPAKA_FN_HOST_ACC Complex(Complex const& other) = default;
+
+        //! Constructor from Complex of another type
+        template<typename U>
+        constexpr ALPAKA_FN_HOST_ACC Complex(Complex<U> const& other)
+            : m_real(static_cast<T>(other.real()))
+            , m_imag(static_cast<T>(other.imag()))
+        {
+        }
```
bernhardmgruber, line 64:
```suggestion
        template<typename U>
        explicit constexpr ALPAKA_FN_HOST_ACC Complex(Complex<U> const& other)
            : m_real(static_cast<T>(other.real()))
            , m_imag(static_cast<T>(other.imag()))
        {
        }
```
You may consider making conversion between complex numbers of different FP type explicit.

test/unit/math/src/Functor.hpp
```diff
@@ -51,11 +70,22 @@ namespace alpaka
         ALPAKA_NO_HOST_ACC_WARNING                                                                                    \
         template<                                                                                                     \
             typename TAcc = std::nullptr_t,                                                                           \
-            typename... TArgs, /* SFINAE: Enables if called from host. */                                             \
+            typename TArg1, /* SFINAE: Enables if called from host. */                                                \
+            typename std::enable_if<std::is_same<TAcc, std::nullptr_t>::value, int>::type = 0>                        \
```
bernhardmgruber:
```suggestion
            std::enable_if_t<std::is_same_v<TAcc, std::nullptr_t>, int> = 0>                        \
```

include/alpaka/math/Complex.hpp
sbastrakov: I think i had a warning when using a compiler-provided one, but will check for the current code version.

include/alpaka/math/Complex.hpp
sbastrakov: I wanted to make them as close to `std::complex<>` as possible, and there the operators become `constexpr` in C++20.

include/alpaka/math/Complex.hpp
sbastrakov: I probably would for a new class, but here mirrored [how cppreference describes `std::complex`](https://en.cppreference.com/w/cpp/numeric/complex/operator_arith2)

SimeonEhrig:
@sbastrakov If you are finish with applying the suggestions, I would test the PR again with the vikunja example: https://github.com/alpaka-group/vikunja/pull/54

include/alpaka/math/Complex.hpp
sbastrakov: I guess it's quite obvious, but after seeing our traits for math a new user might have expected a similar generalized complex implementation. So i'll prefer to keep it.

include/alpaka/math/Complex.hpp
sbastrakov: [This page](https://en.cppreference.com/w/cpp/utility/feature_test) lists  `__cpp_lib_is_layout_compatible` as C++20.

include/alpaka/math/Complex.hpp
sbastrakov: No, but i mirrored [how `std::complex` does it](https://en.cppreference.com/w/cpp/numeric/complex/operator_ltltgtgt)

include/alpaka/math/Complex.hpp
sbastrakov: I re-checked, that's what `std::complex` does, only assignment-operation has mixed-type support

include/alpaka/math/Complex.hpp
sbastrakov: It's non-explicit [here](https://en.cppreference.com/w/cpp/numeric/complex/complex), so i did the same

sbastrakov:
@bernhardmgruber thanks for your review! As i answered in a few points, the choices were made simply to mirror interface of `std::complex<T>`. The idea is, that it's as smooth as possible to transition from `std::complex<T>` to `alpaka::Complex<T>` when porting a code base to alpaka. So personally i would also do what you suggest in some places, but think here it's not worth to deviate from the `std::` implementation unless there is a good reason.

I resolved the points i implemented in my local branch, and kept open and answered the points I think should not be implemented.

include/alpaka/math/Complex.hpp
bernhardmgruber: Which is perfectly fine! When the macro is not defined in C++17, then the `static_assert` is just bypassed.

include/alpaka/math/Complex.hpp
bernhardmgruber: Some tools warn if you provide a copy constructor but now copy assignment operator. So this might be a reason to have it.

include/alpaka/math/Complex.hpp
bernhardmgruber: Ok, that explains the choice. I would still make it `constexpr` because it makes the class more powerful and also forward compatible to C++20.

sbastrakov: So perhaps i will just add a `@todo` to the header doxygen of the class to switch to C++20 signatures of `std::complex` when we transition

include/alpaka/math/Complex.hpp
sbastrakov: Okay, i'll implement it

include/alpaka/math/Complex.hpp
bernhardmgruber: It's fine! It's just really the first time I see someone use `basic_ostream` instead of `ostream`.

include/alpaka/math/Complex.hpp
bernhardmgruber: Ok then!

include/alpaka/math/Complex.hpp
bernhardmgruber: Fine!

bernhardmgruber:
@sbastrakov the PR is fine as is. I only suggested a few enhancements to improve the implementation. I would really like to see the `constexpr` and layout comaptible check. But if you don't want to add it, so be it.

include/alpaka/math/Complex.hpp
sbastrakov: Yes, i actually now faced it. So will keep as it was.

include/alpaka/math/Complex.hpp
sbastrakov: Discussed in mattermost, will keep C++17 interface and add a note on motivation of it.

include/alpaka/math/Complex.hpp
bernhardmgruber: Ok!

test/unit/math/src/mathComplexFloat.cpp
```diff
@@ -34,3 +35,17 @@ TEMPLATE_LIST_TEST_CASE("mathOpsComplexFloat", "[math] [operator]", TestAccFunct
     auto testTemplate = TestTemplate<Acc, Functor>{};
     testTemplate.template operator()<alpaka::Complex<float>>();
 }
+
+#ifdef __has_include
+#    if __has_include(<version>)
+#        include <version>
+#    endif
+#endif
```
bernhardmgruber: `__cpp_lib_is_layout_compatible` is also defined inside `<type_traits>`, so you can just include that header instead, which you will need anyway for `std::is_layout_compatible_v`.

test/unit/math/src/mathComplexFloat.cpp
```diff
@@ -0,0 +1,51 @@
...
+#ifdef __cpp_lib_is_layout_compatible
+TEMPLATE_LIST_TEST_CASE("mathOpsComplex", "[layout]", TestAccs)
+{
+    // Check both float and double here to avoid duplicating the wrapper code
+    REQUIRE(std::is_layout_compatible_v<alpaka::Complex<float>, std::complex<float>>);
+    REQUIRE(std::is_layout_compatible_v<alpaka::Complex<double>, std::complex<double>>);
```
bernhardmgruber:
```suggestion
    STATIC_REQUIRE(std::is_layout_compatible_v<alpaka::Complex<float>, std::complex<float>>);
    STATIC_REQUIRE(std::is_layout_compatible_v<alpaka::Complex<double>, std::complex<double>>);
```
make it a compile-time check.

SimeonEhrig:
> @sbastrakov If you are finish with applying the suggestions, I would test the PR again with the vikunja example: [alpaka-group/vikunja#54](https://github.com/alpaka-group/vikunja/pull/54)

I tested the example with Serial CPU, OpenMP Grid-Block, CUDA and HIP backkend. Looks good.

sbastrakov:
Damn, now 1 test that have passed yesterday fails. Even tho i'm sure i didn't make any changes that should have affected it.

I now loosened the bounds a little bit, hopefully will pass now. Also implemented @bernhardmgruber 's latest review suggestion.

bernhardmgruber (approved):

sbastrakov:
@bernhardmgruber @j-stephan ready to be merged

PR #1338, opened 2021-06-09
native hip clang support

### LINKED ISSUE #1337: Drop HIP-nvcc support
I am currently testing using HIP targets for alpaka as suggested by @j-stephan https://github.com/alpaka-group/alpaka/pull/1324#issuecomment-848680830

I found out that if we like to use targets we need to use `find_package(HIP CONFIG)` else there is no target `hip::device` available.
Wenn using the normal old cmake find module the build system is using the `hipcc` wrapper while using `CONFIG` will enforce the usage of a native clang with AMD support.
I prefer to switch using the native clang and drop using `hipcc` to allow clean target usage and modern CMake for HIP.

The disadvantage when using the native `clang` is that we can not use HIP nvcc support anymore because this would require `hipcc`.

I would like to drop `HIP-nvcc` support for alpaka in the current dev (upcoming 0.8.0). 
IMO this is not a problem because we can use NVIDIA GPUs via clang(native) or nvcc(native). 
I do not see that it is useful to maintain `HIP-nvcc` anymore. HIP-nvcc is currently not tested in the CI and maybe not working anyway.

@alpaka-group/alpaka-maintainers If someone has another opinion please feel free to provide reasons to keep HIP-nvcc.

psychocoderHPC:
fix #1337

Use native clang without the wrapper `hipcc` for the AMD HIP backend.

- drop HIP-nvcc support
- **remove** `CMAKE_ALPAKA_HIP_PLATFORM`
- setting `CMAKE_MODULE_PATH` for HIP is not required anymore
- remove C++ compile option forwarding workarounds
- use modern CMake targets for HIP
- fix implicit casts and handle unused variables

# todo
- [x] merge #1324 first


psychocoderHPC:
ready for review

cmake/alpakaCommon.cmake
```diff
@@ -467,173 +459,49 @@ endif()
...
-                find_package(rocrand REQUIRED CONFIG
-                    HINTS "${HIP_ROOT_DIR}/rocrand"
-                    HINTS "/opt/rocm/rocrand")
-                if(rocrand_FOUND)
-                    target_include_directories(alpaka INTERFACE ${rocrand_INCLUDE_DIRS})
-                    # ATTENTION: rocRand libraries are not required by alpaka
-                else()
-                    MESSAGE(FATAL_ERROR "Could not find rocRAND (also searched in: HIP_ROOT_DIR=${HIP_ROOT_DIR}/rocrand).")
-                endif()
+        alpaka_compiler_option(HIP_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps 'CMakeFiles/<targetname>.dir'" OFF)
```
bernhardmgruber, line 471: Could we unify this with the CUDA KEEP FILES? Or not, because CUDA is now using CMake native CUDA support?

bernhardmgruber (dismissed):
Looks good to me in general, without understanding every detail. But if the CI completes I am fine with this.

cmake/addExecutable.cmake
```diff
@@ -25,18 +25,6 @@ MACRO(ALPAKA_ADD_EXECUTABLE In_Name)
         ADD_EXECUTABLE(
             ${In_Name}
             ${ARGN})
-
-    ELSEIF(ALPAKA_ACC_GPU_HIP_ENABLE)
```
j-stephan, line 29: General question: Now that we have removed both CUDA's and HIP's special treatment with regard to `add_executable`, do we still need this file at all?

cmake/alpakaCommon.cmake
```diff
@@ -467,173 +459,49 @@ endif()
 # Find HIP.
 if(ALPAKA_ACC_GPU_HIP_ENABLE)
 
-    if(NOT DEFINED ALPAKA_HIP_VERSION)
-        set(ALPAKA_HIP_VERSION 4.0)
-    endif()
+    # minimal supported HIP version
+    set(_ALPAKA_HIP_MIN_VER 4.0)
```
j-stephan, line 463: What is the purpose of `_ALPAKA_HIP_MIN_VER`? Do we need this in other places than `find_package`?

cmake/alpakaCommon.cmake
```diff
@@ -467,173 +459,49 @@ endif()
...
+    find_package(hip "${_ALPAKA_HIP_MIN_VER}")
```
j-stephan, line 464: Does this work without explicit `CONFIG` mode?

BenjaminW3 (dismissed):

cmake/alpakaCommon.cmake
```diff
@@ -467,173 +459,49 @@ endif()
...
-                else()
-                    MESSAGE(FATAL_ERROR "Could not find rocRAND (also searched in: HIP_ROOT_DIR=${HIP_ROOT_DIR}/rocrand).")
-                endif()
+        alpaka_compiler_option(HIP_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps 'CMakeFiles/<targetname>.dir'" OFF)
 
-                if(ALPAKA_FAST_MATH STREQUAL ON)
-                    alpaka_set_compiler_options(DEVICE list HIP_HIPCC_FLAGS "-ffast-math")
-                endif()
+        if(ALPAKA_FAST_MATH STREQUAL ON)
+            alpaka_set_compiler_options(DEVICE target alpaka "-ffast-math")
```
j-stephan, line 474: Do we still need `alpaka_set_compiler_options` with enabled HIP targets?

cmake/alpakaCommon.cmake
```diff
@@ -467,173 +459,49 @@ endif()
...
+        if(rocrand_FOUND)
+            target_link_libraries(alpaka INTERFACE roc::rocrand)
+        else()
+            MESSAGE(FATAL_ERROR "Could not find rocRAND (also searched in: HIP_ROOT_DIR=${HIP_ROOT_DIR}/rocrand).")
+        endif()
 
-            # # HIP random numbers
-            FIND_PACKAGE(hiprand REQUIRED CONFIG
+        # # HIP random numbers
+        FIND_PACKAGE(hiprand REQUIRED CONFIG
```
j-stephan: Well, if you're touching this line anyway, please convert it to `lower_case`.

cmake/alpakaCommon.cmake
```diff
@@ -774,24 +613,7 @@ if(TARGET alpaka)
...
-    # because cmake HIP modules do not use the former variables
-    string(TOUPPER "${CMAKE_BUILD_TYPE}" build_config)
-    alpaka_set_compiler_options(HOST var CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS_${build_config}}")
-
-    # additionally CMAKE_CXX_FLAGS must be propagated to HIP_HIPCC_FLAGS
-    string(REPLACE " " ";" CMAKE_CXX_FLAGS_AS_LIST "${CMAKE_CXX_FLAGS}")
-    alpaka_set_compiler_options(DEVICE list HIP_HIPCC_FLAGS "${CMAKE_CXX_FLAGS_AS_LIST}")
-endif()
-
+# Compile options summery
```
```suggestion
# Compile options summary
```

docs/source/advanced/cmake.rst
```diff
@@ -233,8 +233,7 @@ ALPAKA_CUDA_SHOW_REGISTER
 HIP
 ---
 
-To enable the HIP backend please provide the path to the CMake find module `FindHIP.cmake`.
-The path can be given via an environment variable `CMAKE_MODULE_PATH` or by providing the CMake flag `-DCMAKE_MODULE_PATH=<path>`.
+To enable the HIP backend please extent ``CMAKE_PREFIX_PATH`` with the path to the HIP installation.
```
j-stephan:
```suggestion
To enable the HIP back-end please extend ``CMAKE_PREFIX_PATH`` with the path to the HIP installation.
```

script/install_hip.sh
```diff
@@ -18,3 +18,29 @@ source ./script/set.sh
...
+
+sudo update-alternatives --install /usr/bin/clang clang ${ROCM_PATH}/llvm/bin/clang 50
+sudo update-alternatives --install /usr/bin/clang++ clang++ ${ROCM_PATH}/llvm/bin/clang++ 50
+sudo update-alternatives --install /usr/bin/cc cc ${ROCM_PATH}/llvm/bin/clang 50
+sudo update-alternatives --install /usr/bin/c++ c++ ${ROCM_PATH}/llvm/bin/clang++ 50
+
+export LD_LIBRARY_PATH=${ROCM_PATH}/lib64:${ROCM_PATH}/hiprand/lib:${LD_LIBRARY_PATH}:${ROCM_PATH}/llvm/lib
+export CMAKE_PREFIX_PATH=${ROCM_PATH}:${ROCM_PATH}/hiprand:${CMAKE_PREFIX_PATH:-}
+
+# calls nvcc or clang
```
j-stephan: Do we still call nvcc?

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
```diff
@@ -693,7 +694,7 @@ namespace alpaka
                 std::cerr << "[Workaround] polling of device-located value in stream, as hipStreamWaitValue32 is not "
                              "available.\n";
 #    endif
-                while(hostMem < 0x01010101u)
+                while(hostMem < 0x01010101)
```
j-stephan, line 697: Why the change of type here?

j-stephan (changes requested):
Great work! A few remarks:

cmake/alpakaCommon.cmake
psychocoderHPC: It can be potentially unified with CUDA but not with all CPU backends. Never the less I would not integrate it into this PR because this is out of the scope.

cmake/addExecutable.cmake
psychocoderHPC: IMO yes, to mark the cpp files as cuda files and in the near future we need the same for HIP. As soon as HIP becomes a native language in CMake.

cmake/alpakaCommon.cmake
psychocoderHPC: We use the same pattern for boost too.  This will separate defining the minimum number from `find_package()` and increases the readability because it is not only a magic number `find_package(hip 4.0)`.
This is helpful for someone who is not familiar with CMake.

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
bernhardmgruber: I guess because `hostMem` is also a signed integer. The previous version probably triggers a warning.

cmake/alpakaCommon.cmake
psychocoderHPC: Yes if you configured your environment correctly and extent `CMAKE_PREFIX_PATH` or one of the other default CMAke ways to say where HIP is installed. The CMake handling is equal to all other dependencies.

cmake/alpakaCommon.cmake
psychocoderHPC: Yes this is tracking the flags we give to the compiler and reports a summery to the user.

script/install_hip.sh
psychocoderHPC: no, I will change the comment

test/common/include/alpaka/test/event/EventHostManualTrigger.hpp
psychocoderHPC: implicit comparison between `int` and `unsigned int`. `hostMem` is `int` therefore the hex code must be `int` too.

psychocoderHPC:
I updated this PR and answered your questions.

bernhardmgruber (approved):

j-stephan (approved):

PR #1339, opened 2021-06-10
CI: fix warp test

psychocoderHPC:
Fix undefined behavior because we shifted `static_cast<T>(1)<<n` where `n == sizeof(T)`.
If `n` is known at compile time the compiler is throwing a warning.
In the case where we use it, we assumed `1<<32` for 32bit is always `0`, this is wrong!

Unfortunately, we need to backport this to 0.6.1 and 0.7.0 :-(

### DIFF
--- a/test/unit/warp/src/Activemask.cpp
+++ b/test/unit/warp/src/Activemask.cpp
@@ -14,6 +14,7 @@
 
 #include <catch2/catch.hpp>
 
+#include <climits>
 #include <cstdint>
 
 class ActivemaskSingleThreadWarpTestKernel
@@ -51,7 +52,9 @@ public:
 
         auto const actual = alpaka::warp::activemask(acc);
         using Result = decltype(actual);
-        Result const allActive = (Result{1} << static_cast<Result>(warpExtent)) - 1;
+        Result const allActive = static_cast<size_t>(warpExtent) == sizeof(Result) * CHAR_BIT
+            ? ~Result{0u}
+            : (Result{1} << warpExtent) - 1u;
         Result const expected = allActive & ~(Result{1} << inactiveThreadIdx);
         ALPAKA_CHECK(*success, actual == expected);
     }
--- a/test/unit/warp/src/Ballot.cpp
+++ b/test/unit/warp/src/Ballot.cpp
@@ -14,6 +14,7 @@
 
 #include <catch2/catch.hpp>
 
+#include <climits>
 #include <cstdint>
 
 class BallotSingleThreadWarpTestKernel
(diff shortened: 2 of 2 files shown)


test/unit/warp/src/Activemask.cpp
```diff
@@ -51,7 +51,9 @@ class ActivemaskMultipleThreadWarpTestKernel
 
         auto const actual = alpaka::warp::activemask(acc);
         using Result = decltype(actual);
-        Result const allActive = (Result{1} << static_cast<Result>(warpExtent)) - 1;
+        Result const allActive = static_cast<size_t>(warpExtent) == sizeof(Result) * 8u
```
bernhardmgruber:
```suggestion
        Result const allActive = static_cast<size_t>(warpExtent) == sizeof(Result) * CHAR_BIT
```
I was very confused what the `8u` meant until I noticed you wanted to convert a size in bytes to bits. With `CHAR_BIT` this is more clear and i supports architectures where 1byte != 8bit :)

test/unit/warp/src/Activemask.cpp
```diff
@@ -51,7 +51,9 @@ class ActivemaskMultipleThreadWarpTestKernel
...
+            ? ~static_cast<Result>(0u)
```
bernhardmgruber:
```suggestion
            ? ~Result{0}
```
I think this should be the same?

test/unit/warp/src/Ballot.cpp
```diff
@@ -41,7 +41,11 @@ class BallotMultipleThreadWarpTestKernel
         std::int32_t const warpExtent = alpaka::warp::getSize(acc);
         ALPAKA_CHECK(*success, warpExtent > 1);
 
-        ALPAKA_CHECK(*success, alpaka::warp::ballot(acc, 42) == (std::uint64_t{1} << warpExtent) - 1);
+        using BallotResultType = decltype(alpaka::warp::ballot(acc, 42));
+        BallotResultType const allActive = static_cast<size_t>(warpExtent) == sizeof(BallotResultType) * 8u
```
bernhardmgruber:
```suggestion
        BallotResultType const allActive = static_cast<size_t>(warpExtent) == sizeof(BallotResultType) * CHAR_BIT
```

test/unit/warp/src/Ballot.cpp
```diff
@@ -41,7 +41,11 @@ class BallotMultipleThreadWarpTestKernel
...
+            ? ~static_cast<BallotResultType>(0u)
```
bernhardmgruber:
```suggestion
            ? ~BallotResultType{0}
```

bernhardmgruber:
> Unfortunately, we need to backport this to 0.6.1 and 0.7.0 :-(

Why do we need to backport this? It's a bug in the unit tests and not in the library.

psychocoderHPC:
> > Unfortunately, we need to backport this to 0.6.1 and 0.7.0 :-(
> 
> Why do we need to backport this? It's a bug in the unit tests and not in the library.

If you run alpaka tests with the release branches they will fail. In that case, the user does not know if the test is broken or the implementation of the function.

The problem is that our CI is not doing runtime tests for our most important targets CUDA and HIP (**this is a huge issue**). Maybe HIP was passing in the past because the left shift was following our assumption but with the HIP 4.2 it is not the case anymore. Maybe it is broken for a long time but I do not run the alpaka tests very often.

psychocoderHPC:
I opened #1340 to push HZDR CI integration.

psychocoderHPC:
@bernhardmgruber I added your suggestions

bernhardmgruber (dismissed):

psychocoderHPC:
PR passed the CI now

bernhardmgruber (approved):

PR #1341, opened 2021-06-15
Fix outdated comments in the OpenMP schedule example

sbastrakov:
In the schedule implementation fix #1309, the comments in the example were outdated. Thanks to DLR for reporting this.

Regarding the other reported issue, there is a possible situation that one sets an `ompScheduleChunkSize` member but not `ompScheduleKind`. In this case `ompScheduleChunkSize` would be doing nothing, in accordance to how it is documented. I see that there can be a typo causing it, but I do not see any easy way to check and issue a warning there. Internally, this part is already very complicated, and also compile-time-consuming. So I think it is more practical to keep it as is.

### DIFF
--- a/example/openMPSchedule/src/openMPSchedule.cpp
+++ b/example/openMPSchedule/src/openMPSchedule.cpp
@@ -29,7 +29,7 @@
 //!
 //! Prints distribution of alpaka thread indices between OpenMP threads.
 //! Its operator() is reused in other kernels of this example.
-//! Sets no schedule explicitly, so the default is used, controlled by the OMP_SCHEDULE environment variable.
+//! Sets no schedule explicitly, so no schedule() clause is used.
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
@sbastrakov I think https://github.com/alpaka-group/alpaka/pull/1341/files#diff-3b259106a02535ed19d81b9083c681059b100275c5f8803c30beef98ca99d74aR32 is outdated too. I would say `OMP_SCHEDULE` is not used if no schedular is defined.

sbastrakov:
You are right, I missed that.

sbastrakov:
@psychocoderHPC CI passed. I've fixed the thing you noted.

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #1342, opened 2021-06-16
TaskKernelOacc: copyin(all used local vars)

jkelling:
Because of the `default(present)` clause, which is required, because alpaka device buffer contain device pointer and should not be mapped to device again, all local variables use in the device code must be mapped manually.

Please backport to 0.6.1.


jkelling:
Ping @psychocoderHPC

psychocoderHPC (approved):

PR #1345, opened 2021-06-22
Showcase and discuss CRP clang-tidy 

bernhardmgruber:
This PR tests the clang-tidy file from https://github.com/ComputationalRadiationPhysics/contributing/blob/master/formatting-tools/.clang-tidy. It just includes the result of running run-clang-tidy -fix, with no additional manual changes.

### DIFF
--- a/test/integ/mandelbrot/src/mandelbrot.cpp
+++ b/test/integ/mandelbrot/src/mandelbrot.cpp
@@ -22,44 +22,44 @@
 //#define ALPAKA_MANDELBROT_TEST_CONTINOUS_COLOR_MAPPING  // Define this to enable the continuous color mapping.
 
 //! Complex Number.
-template<typename T>
+template<typename TT>
 class SimpleComplex
 {
 public:
     ALPAKA_NO_HOST_ACC_WARNING
-    ALPAKA_FN_HOST_ACC SimpleComplex(T const& a, T const& b) : r(a), i(b)
+    ALPAKA_FN_HOST_ACC SimpleComplex(TT const& a, TT const& b) : m_r(a), m_i(b)
     {
     }
     ALPAKA_NO_HOST_ACC_WARNING
     ALPAKA_FN_INLINE
-    ALPAKA_FN_HOST_ACC auto absSq() const -> T
+    ALPAKA_FN_HOST_ACC auto abs_sq() const -> TT
     {
-        return r * r + i * i;
+        return m_r * m_r + m_i * m_i;
     }
     ALPAKA_NO_HOST_ACC_WARNING
     ALPAKA_FN_HOST_ACC auto operator*(SimpleComplex const& a) -> SimpleComplex
     {
-        return SimpleComplex(r * a.r - i * a.i, i * a.r + r * a.i);
+        return SimpleComplex(m_r * a.m_r - m_i * a.m_i, m_i * a.m_r + m_r * a.m_i);
     }
     ALPAKA_NO_HOST_ACC_WARNING
     ALPAKA_FN_HOST_ACC auto operator*(float const& a) -> SimpleComplex
     {
-        return SimpleComplex(r * a, i * a);
+        return SimpleComplex(m_r * a, m_i * a);
     }
     ALPAKA_NO_HOST_ACC_WARNING
     ALPAKA_FN_HOST_ACC auto operator+(SimpleComplex const& a) -> SimpleComplex
     {
-        return SimpleComplex(r + a.r, i + a.i);
+        return SimpleComplex(m_r + a.m_r, m_i + a.m_i);
     }
     ALPAKA_NO_HOST_ACC_WARNING
     ALPAKA_FN_HOST_ACC auto operator+(float const& a) -> SimpleComplex
     {
-        return SimpleComplex(r + a, i);
+        return SimpleComplex(m_r + a, m_i);
     }
 
-public:
-    T r;
-    T i;
+
+    TT m_r;
+    TT m_i;
 };
 
 //! A Mandelbrot kernel.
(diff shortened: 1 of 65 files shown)


BenjaminW3:
I really would like to add automated clang-tidy checks to our CI in the same way we already have clang-format. I am not yet sure about the `readability-identifier-naming` things though.

bernhardmgruber:
Sure, we should have CI checks for clang-tidy. I did not look into this yet.

The `readability-identifier-naming` is the result of all the discussions and votings we had on coding style over the last year, specifically this one: https://github.com/ComputationalRadiationPhysics/contributing/pull/42

SimeonEhrig:
@bernhardmgruber Like you already mentioned yesterday in the VC, our big amount of compiler combinations could be a problem. The `kernel.ompScheduleChunkSize` was only changed in one of the `#if #else` branches.

https://github.com/alpaka-group/alpaka/blob/7bf07b027f1b55839aa21abf32bcd52d13d137f1/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp#L223-L232

SimeonEhrig:
@BenjaminW3 What is the time impact of clang-tidy compare to normal compiling?

BenjaminW3:
I can not quantify clang-tidy compilation time increase. However, we should get most of its advantages by adding a single additional job to our CI, which should not increase the overall time considerably.

SimeonEhrig:
> I can not quantify clang-tidy compilation time increase. However, we should get most of its advantages by adding a single additional job to our CI, which should not increase the overall time considerably.

Sorry, I pinged the wrong version. I wanted to ping @bernhardmgruber :sweat_smile: 
Nevertheless, I'm not sure if a single job is enough. If clang-tidey respect the config done by CMake we have to run multiple configurations (CPU Serial, CUDA, Linux, Windows ...) to cover the complete code. In this case, maybe it makes sense to extends the existing jobs. The alternative would be to setup a second series of CI jobs (maybe using the configs from the existing CI jobs).

psychocoderHPC:
We have the header validation test, where we check if all includes are available, I am not sure if we perform this test for each accelerator but if this could be an entry to use clang-tidy.

test/integ/mandelbrot/src/mandelbrot.cpp
```diff
@@ -22,44 +22,44 @@
 //#define ALPAKA_MANDELBROT_TEST_CONTINOUS_COLOR_MAPPING  // Define this to enable the continuous color mapping.
 
 //! Complex Number.
-template<typename T>
+template<typename TT>
```
SimeonEhrig, line 25: I thought, we decided, that a single char is fine for a template parameter :thinking:

test/unit/math/src/sincos.cpp
```diff
@@ -17,34 +17,34 @@
 #include <type_traits>
 
 // https://en.cppreference.com/w/cpp/types/numeric_limits/epsilon
-template<typename TAcc, typename FP>
```
SimeonEhrig, line 20: What is going on here?

test/unit/mem/view/src/ViewStaticAccMem.cpp
```diff
@@ -24,27 +24,27 @@ using Idx = std::uint32_t;
 // from a different compilation unit and should be moved to a common header.
 // Here they are used to silence clang`s -Wmissing-variable-declarations warning
 // that forces every non-static variable to be declared with extern before the are defined.
-extern ALPAKA_STATIC_ACC_MEM_CONSTANT Elem g_constantMemory2DInitialized[3][2];
-extern ALPAKA_STATIC_ACC_MEM_CONSTANT Elem g_constantMemory2DUninitialized[3][2];
+extern ALPAKA_STATIC_ACC_MEM_CONSTANT Elem g_constant_memory2_d_initialized[3][2];
```
SimeonEhrig: The automatism failed here.

SimeonEhrig:
Most of the changes looks fine for me. In special, I'm really lazy to keep function, class and variable names consistent :sweat_smile: 

If we apply clang-tidy, we need also document the API changes.

test/integ/mandelbrot/src/mandelbrot.cpp
bernhardmgruber: Yes we did. The PR is quite old, so maybe that was not in yet.

test/unit/math/src/sincos.cpp
bernhardmgruber: It renamed the template arg to `TFp` according to our coding conventions.

test/unit/mem/view/src/ViewStaticAccMem.cpp
bernhardmgruber: I did. Some of these can be fixed manually and clang-tidy is not reformatting them again.

bernhardmgruber:
I think we need to discuss how we apply these at some point. Because we had votes on the coding style more than a year ago, but somehow LLAMA is the only project that actually checks some of the rules in the CI. Nobody picked them up otherwise. A coding style is worthless if it is not automatically enforced. But I also don't have the time to follow up on this for every project. That's for the professional support.

test/integ/mandelbrot/src/mandelbrot.cpp
SimeonEhrig: Okay, I found the PR: https://github.com/ComputationalRadiationPhysics/contributing/pull/54

test/unit/math/src/sincos.cpp
SimeonEhrig: Sure, but I can't reconstruct the rule, which was applied on this. Looks like, it was to short and something needed to be appended :thinking: Nevertheless, this will be also ignored by the new Clang-tidy file.

https://github.com/ComputationalRadiationPhysics/contributing/pull/54

test/unit/mem/view/src/ViewStaticAccMem.cpp
SimeonEhrig: I know. It should be only a remark.

SimeonEhrig:
> I think we need to discuss how we apply these at some point. Because we had votes on the coding style more than a year ago, but somehow LLAMA is the only project that actually checks some of the rules in the CI. Nobody picked them up otherwise. A coding style is worthless if it is not automatically enforced. But I also don't have the time to follow up on this for every project. That's for the professional support.

I want to also introduce an (automatic) clang-tidy check in vikunja before the next release, because of possible API changes: https://github.com/alpaka-group/vikunja/issues/47

I have only fear, that there are more changes necessary than renaming :-( On the other side, I'm pretty lazy, therefore it would great, if the computer can do my job :wink:

bernhardmgruber:
Here is a ticket to remind us of this effort: #1786. Since this PR is stale, I am closing it.

PR #1347, opened 2021-06-23
.gitignore for Eclipse project files & directories

sliwowitz:
For C/C++ projects, Eclipse stores the project configuration in the top level directory in two files, `.project` and `.cproject`, and local user settings in `.settings` directory. (I realize I might be the only one using Eclipse, but still...)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,9 @@ cmake-build-*/
 *.kdev?
 spack-build*
 build/
+
+# Eclipse project files
+/.project
+/.cproject
+/.settings/
+


.gitignore
```diff
@@ -25,3 +25,9 @@ cmake-build-*/
 *.kdev?
 spack-build*
 build/
+
+# Eclipse project files
+/.project
```
j-stephan, line 30: Is the preceeding `/` correct? Naïvely I'd say that this looks up the file in the root file system, but I'm no expert on `.gitignore`.

bernhardmgruber: I think this just ignores `.project` files in the root of the repository.

bernhardmgruber (approved):

.gitignore
sliwowitz: Yes. Leading `/` looks only for matching files located in the same directory as the `.gitignore` file itself. Without the `/`, the line would match the files in sub-directories as well.

j-stephan (approved):

psychocoderHPC (approved):

PR #1349, opened 2021-06-25
Refactor include and checks for CUDA/HIP math functions

### LINKED ISSUE #1335: Duplication of CUDA/HIP include checks?
While working on traits for my ongoing complex number implementation, I've noticed all math function implementation files for CUDA/HIP have [the same piece as "include-and-check"](https://github.com/alpaka-group/alpaka/blob/6861f7b31ef0725cb117d523b94fbe2c53b60b81/include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp#L16-L38). Maybe it is duplicated in some other files as well, but even among the math files it is present too many times. Is there a reason it is not separated into a special include file implementing just that once, and then included when needed?

sbastrakov:
Those checks were duplicated in multiple files.
Move this logic to a new header file.

Resolves #1335 .

### DIFF
--- a/include/alpaka/core/CudaHipMath.hpp
+++ b/include/alpaka/core/CudaHipMath.hpp
@@ -17,13 +17,5 @@
 #endif
 
 #if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
-#    if BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 0, 0)
-#        include <cuda_runtime_api.h>
-#    else
-#        if BOOST_COMP_HIP
-#            include <hip/math_functions.h>
-#        else
-#            include <math_functions.hpp>
-#        endif
-#    endif
+#    include <hip/math_functions.h>
 #endif
--- a/include/alpaka/core/CudaHipMath.hpp
+++ b/include/alpaka/core/CudaHipMath.hpp
@@ -1,0 +1,29 @@
+/* Copyright 2019-2021 Benjamin Worpitz, René Widera, Sergei Bastrakov
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/BoostPredef.hpp>
+#include <alpaka/core/UniformCudaHip.hpp>
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+#    include <cuda_runtime.h>
+#endif
+
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#    if BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 0, 0)
+#        include <cuda_runtime_api.h>
+#    else
+#        if BOOST_COMP_HIP
+#            include <hip/math_functions.h>
+#        else
+#            include <math_functions.hpp>
+#        endif
+#    endif
+#endif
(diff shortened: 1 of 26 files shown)


include/alpaka/core/CudaHipMath.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/BoostPredef.hpp>
+#include <alpaka/core/UniformCudaHip.hpp>
+
```
sbastrakov, line 14: Note: in the [original duplicated piece](https://github.com/alpaka-group/alpaka/blob/6861f7b31ef0725cb117d523b94fbe2c53b60b81/include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp#L16-L38) there was also checks and errors for backend and compiler mismatch. However, exactly the same checks are done in `alpaka/core/UniformCudaHip.hpp`, so I did not put them in this file as well.

include/alpaka/math/MathUniformCudaHipBuiltIn.hpp
```diff
@@ -11,31 +11,6 @@
...
-#                include <hip/math_functions.h>
-#            else
-#                include <math_functions.hpp>
-#            endif
-#        endif
-#        if !BOOST_LANG_HIP
-#            error If ALPAKA_ACC_GPU_HIP_ENABLED is set, the compiler has to support HIP!
-#        endif
-#    endif
-
```
sbastrakov, line 38: I think this whole piece should not have been here at all. All the header files do the check, and this file by itself does not add anything. So here i just removed it

include/alpaka/core/CudaHipMath.hpp
```diff
@@ -0,0 +1,29 @@
...
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+#    include <cuda_runtime.h>
+#endif
+
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#    if BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(9, 0, 0)
+#        include <cuda_runtime_api.h>
+#    else
```
bernhardmgruber: Didn't we drop targetting CUDA via HIP? Because then we can remove this branch.

bernhardmgruber (changes requested):

include/alpaka/core/CudaHipMath.hpp
bernhardmgruber: Fine for me!

include/alpaka/math/MathUniformCudaHipBuiltIn.hpp
bernhardmgruber: Fine for me!

include/alpaka/core/CudaHipMath.hpp
psychocoderHPC: well spotted, yes CUDA support via hip was removed with #1338. This workaround can be removed.

sbastrakov: Thanks! Will remove then

sbastrakov: Wait, but what exactly should be removed? I am a bit confused about this part

bernhardmgruber: I guess we can replace the entire block by:
```c++
#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
#    include <hip/math_functions.h>
#endif
```

sbastrakov: Okay, so then there was not much point in moving it all into a separate header. But i guess it doesn't hurt anyways, just doesn't help as much as I originally thought.

psychocoderHPC: There must be a branch for CUDA and one for HIP.

psychocoderHPC (approved):

bernhardmgruber (approved):

PR #1351, opened 2021-06-29
OpenMP Target fixes: assert, printf

jkelling:
This PR
* fixes some places in which `ALPAKA_ASSERT_OFFLOAD` should be used instead of `ALPAKA_ASSERT` because they occur in device code,
* adds a non-empty definition of `ALPAKA_ASSERT_OFFLOAD` for `__AMDGCN__`,
* defines `ALPAKA_ASSERT_OFFLOAD` as `alpaka::ignore_unused(EXPRESSION)`  when empty to avoid triggering unused argument warnings, and
* defines `printf(...)` as empty macro for `__AMDGCN__` because it is not yet supported in Clang or AOMP offloading compilers and it will still take some time until it is.

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -78,7 +78,7 @@ struct TestBufferKernel
 
         for(size_t i(linearizedGlobalThreadIdx[0]); i < extents.prod(); i += globalThreadExtent.prod())
         {
-            ALPAKA_ASSERT(data[linIdxToPitchedIdx<2>(i, pitch)] == i);
+            ALPAKA_ASSERT_OFFLOAD(data[linIdxToPitchedIdx<2>(i, pitch)] == i);
         }
     }
 };
--- a/include/alpaka/acc/AccOmp5.hpp
+++ b/include/alpaka/acc/AccOmp5.hpp
@@ -15,6 +15,12 @@
 #        error If ALPAKA_ACC_ANY_BT_OMP5_ENABLED is set, the compiler has to support OpenMP 4.0 or higher!
 #    endif
 
+// Kill printf in AMD GPU code because of missing compiler support
+#    ifdef __AMDGCN__
+#        include <cstdio> // the define breaks <cstdio> if it is included afterwards
+#        define printf(...)
+#    endif
+
 // Base classes.
 #    include <alpaka/atomic/AtomicHierarchy.hpp>
 #    include <alpaka/atomic/AtomicOmpBuiltIn.hpp>
--- a/include/alpaka/core/Assert.hpp
+++ b/include/alpaka/core/Assert.hpp
@@ -20,6 +20,13 @@
 
 #ifdef ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST
 #    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) ALPAKA_ASSERT(EXPRESSION)
+#elif defined __AMDGCN__ && (!defined NDEBUG)
+#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
+        do                                                                                                            \
+        {                                                                                                             \
+            if(!(EXPRESSION))                                                                                         \
+                __builtin_trap();                                                                                     \
+        } while(false)
 #else
 #    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)
 #endif
@@ -38,11 +45,8 @@ namespace alpaka
                 ALPAKA_NO_HOST_ACC_WARNING
                 ALPAKA_FN_HOST_ACC static auto assertValueUnsigned(TArg const& arg) -> void
                 {
-#ifdef NDEBUG
                     alpaka::ignore_unused(arg);
-#else
-                    ALPAKA_ASSERT(arg >= 0);
-#endif
+                    ALPAKA_ASSERT_OFFLOAD(arg >= 0);
                 }
             };
             template<typename TArg>
@@ -78,11 +82,8 @@ namespace alpaka
                 ALPAKA_NO_HOST_ACC_WARNING
                 ALPAKA_FN_HOST_ACC static auto assertGreaterThan(TRhs const& lhs) -> void
                 {
-#ifdef NDEBUG
                     alpaka::ignore_unused(lhs);
-#else
-                    ALPAKA_ASSERT(TLhs::value > lhs);
-#endif
+                    ALPAKA_ASSERT_OFFLOAD(TLhs::value > lhs);
                 }
             };
             template<typename TLhs, typename TRhs>
--- a/include/alpaka/idx/bt/IdxBtOmp.hpp
+++ b/include/alpaka/idx/bt/IdxBtOmp.hpp
@@ -59,7 +59,7 @@ namespace alpaka
             {
                 alpaka::ignore_unused(idx);
                 // We assume that the thread id is positive.
-                ALPAKA_ASSERT(::omp_get_thread_num() >= 0);
+                ALPAKA_ASSERT_OFFLOAD(::omp_get_thread_num() >= 0);
                 // \TODO: Would it be faster to precompute the index and cache it inside an array?
                 return mapIdx<TDim::value>(
                     Vec<DimInt<1u>, TIdx>(static_cast<TIdx>(::omp_get_thread_num())),


psychocoderHPC:
@jkelling This PR is still a draft, do you like to release it?

include/alpaka/core/Assert.hpp
```diff
@@ -20,8 +20,12 @@
 
 #ifdef ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST
 #    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) ALPAKA_ASSERT(EXPRESSION)
+#elif defined __AMDGCN__ && (!defined NDEBUG)
+#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
+        if(!(EXPRESSION))                                                                                             \
+        __builtin_trap()
```
jkelling: Somehow this passes the clang-tidy check, but the indentation seems odd here... .

Maybe it is the missing semicolon at the end. If this is added empty expression warnings are raised because invocation sites add another one.

Should this stay this way?

j-stephan: Would it work to put everything in one line here? Otherwise you can turn clang-format off for this particular line and indent by hand.

```suggestion
        // clang-format off
        if(!(EXPRESSION))                                                                                             \
            __builtin_trap()
        //clang-format on
```

jkelling: Having everything on one line was my first version, I think, but clang-format changed it.

The clang format comments would need to surround the whole macro definition, the suggestion breaks it. Maybe inline comments (`/* clang-format off */`) would work?

I am against turning off clang format in this place, because the style demands alignment of these line continuation way to the right, which is just too tedious without clang format.

Workable options would be:
1. Check if the clang-format style can be adjusted to fix this.
2. Adding  `-Wno-extra-semi-stmt` in cmake and adding a semicolon.
3. ignore it and move on

psychocoderHPC: Disabling clang format in edge cases is fine.

1. Changing the format only for an edge case is not the best way, you would always find some strange cases
2. Is not possible because we support stand allows header too and can not enforce the user to disable warnings

The common way to write a macro and enforce a semicolon is

```C++
#define MYMACRO do { \
  if(!(EXPRESSION))  \
      __builtin_trap(); \
while(false)
```

jkelling: That `while(false)` does not look save: Would this not swallow the next statement if there is no semicolon without raising an error? I guess some stand-alone no-op statement would be more safe...

j-stephan: The inline disabling of clang-format should work, too. Maybe that is the best approach here.

jkelling: Ok, I put the inline clang-format switch in.

psychocoderHPC: `do {}while(false)` is the standard expression you use in such a case, it is a common pattern.
The advantage of the do-while is that is works well with if conditions too.

jkelling: That would work, does not swallow statements and raises an error if the semicolon is missing.

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -15,6 +15,12 @@
 #        error If ALPAKA_ACC_ANY_BT_OMP5_ENABLED is set, the compiler has to support OpenMP 4.0 or higher!
 #    endif
 
+// Kill printf in AMD GPU code because of missing compiler support
+#    ifdef __AMDGCN__
```
psychocoderHPC, line 19: This is disabling `printf` for HIP too but I assume OpenMP offloading can not be enabled together with the HIP backend.

jkelling: Enabling both of these at the same time is not a use case I had considered. So, at the moment this would not be possible, or you have to go without `printf` in your HIP kernels, too.

jkelling: Maybe it would also work with:
```c++
#if defined __AMDGCN__ && !defined __HIP__
```
This would assume, that the OpenMP HSA device code compiler does not define `__HIP__` even is HIP enabled otherwise.

This interaction is pure speculation though: We have no compiler that can properly offload OpenMP, let alone does that *and* HIP. So, there would be no way to test this modification. It is likely, that this workaround becomes obsolete,, by `printf` becoming available in clang HSA before anyone tries to run OpenMP target and HIP together.

include/alpaka/core/Assert.hpp
jkelling: I added the `do{} while(false)` version.

jkelling:
The `alpaka::unused(EXPRESSION)` breaks calling code which contain identifiers not declared in release mode. I added guards for these invocation sites.

There appears to be gap in the CI here:
```
-DCMAKE_BUILD_TYPE=Release
-DALPAKA_OFFLOAD_ASSUME_HOST=OFF
```
must be active to catch this issue.

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -78,7 +78,9 @@ struct TestBufferKernel
 
         for(size_t i(linearizedGlobalThreadIdx[0]); i < extents.prod(); i += globalThreadExtent.prod())
         {
-            ALPAKA_ASSERT(data[linIdxToPitchedIdx<2>(i, pitch)] == i);
+#ifndef NDEBUG
```
psychocoderHPC: Mhh you mentioned in your last comment that you added this but I do not understand why this is required. If `#ifndef NDEBUG` is required it should be hidden in the macro and not where it is called.

include/alpaka/core/Assert.hpp
```diff
@@ -20,8 +20,15 @@
...
+#elif defined __AMDGCN__ && (!defined NDEBUG)
+#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
+        if(!(EXPRESSION))                                                                                             \
+            __builtin_trap();                                                                                         \
+        do                                                                                                            \
+        {                                                                                                             \
+        } while(false)
 #else
-#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)
+#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) alpaka::ignore_unused(EXPRESSION)
```
psychocoderHPC: This macro should be empty. Normal `assert` is also not performing anything if `NDEBUG` is defined.

IMO this macro should remove the expression if `NDEBUG` is defined and guards around  `ALPAKA_ASSERT_OFFLOAD()` must be removed.

psychocoderHPC (changes requested):

example/bufferCopy/src/bufferCopy.cpp
psychocoderHPC: IN this special case maybe removing the full loop and not only the assert with `NDEBUG` guard would be fine but all other places where the assert is called should not contain guard around.

jkelling: See other thread.

include/alpaka/core/Assert.hpp
jkelling: This change is the reason the why the `#ifndef NDEBUG` is required in some places: it leaves the expression in place to not trigger unused argument warnings. In exchange leaving the expression in place makes it so, that if the expression uses some identifier not declared when `defined NDEBUG` then the whole macro needs to be guarded.

Maybe the better solution would to not add `alpaka::ingore_unsused` and leave this case to the calling code to deal with. I think such a guard is in place for `ALPAKA_ASSERT`, so this would indeed be more consistent.

jkelling: Removing `alpaka::ingore_unused` breaks the debug build, because the existing guards for unused parameters assume, that the parameter will be used in debug mode, but with `ALPAKA_OFFLOAD_ASSUME_HOST=OFF` this would then not be the case, hence the definition as `alpaka::ingore_unused(EXPRESSION)`.

jkelling: The only way to reconcile this would be to have an unconditional `alpaka::ingore_unused(param)` at the affected cal sites... .

example/bufferCopy/src/bufferCopy.cpp
jkelling: Having the guard inside is more closer to the original version with `ALPAKA_ASSERT`. The should be optimized out when empty.

include/alpaka/core/Assert.hpp
```diff
@@ -81,7 +88,7 @@ namespace alpaka
 #ifdef NDEBUG
                     alpaka::ignore_unused(lhs);
 #else
-                    ALPAKA_ASSERT(TLhs::value > lhs);
+                    ALPAKA_ASSERT_OFFLOAD(TLhs::value > lhs);
```
jkelling, line 86: @psychocoderHPC for example, here the `alpaka::ignore_unused(lhs)` would need to become unconditional for `ALPAKA_ASSERT_OFFLOAD` to work without the `alpaka::ignore_unsused` fallback definition (or check `ALPAKA_OFFLOAD_ASSUME_HOST`, which would defeat the purpose of the macro).

include/alpaka/core/Assert.hpp
```diff
@@ -20,8 +20,15 @@
 
 #ifdef ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST
 #    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) ALPAKA_ASSERT(EXPRESSION)
+#elif defined __AMDGCN__ && (!defined NDEBUG)
+#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
+        if(!(EXPRESSION))                                                                                             \
+            __builtin_trap();                                                                                         \
+        do                                                                                                            \
+        {                                                                                                             \
+        } while(false)
```
psychocoderHPC, line 29: please move line`25,26` into the `do` body. This is a common pattern for marcos and make the macro save to be used within `if` condition branches.

psychocoderHPC:
```C++
#define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) \
do \
{ \
if(!(EXPRESSION)) \
  __builtin_trap(); \
} \
while(false)
```
```

psychocoderHPC (changes requested):

include/alpaka/core/Assert.hpp
jkelling: Ahh, Ok. Done.

psychocoderHPC (dismissed):

jkelling:
Rebased on to develop.

psychocoderHPC (approved):

PR #1352, opened 2021-06-30
test: remove support for ubuntu:16.04

### LINKED ISSUE #1348: CI: Ubuntu 16.04 is deprecated
Ubuntu 16.04 images will be removed from GitHub Actions in September: https://github.blog/changelog/2021-04-29-github-actions-ubuntu-16-04-lts-virtual-environment-will-be-removed-on-september-20-2021/

Time to upgrade our CI runners which still use 16.04.

BenjaminW3:
fixes #1348 

LTS support ended April 2021.


j-stephan:
It seems that this PR uncovered a bug in `linux_nvcc-9.1_clang-4_debug`. For whatever reason, our `CC=clang` is ignored and nvcc pulls in `gcc` instead, which is too new for the CUDA version.

bernhardmgruber:
> It seems that this PR uncovered a bug in `linux_nvcc-9.1_clang-4_debug`. For whatever reason, our `CC=clang` is ignored and nvcc pulls in `gcc` instead, which is too new for the CUDA version.

This might be related to switching to native CUDA support in CMake. If you set `CXX` then CMake will use that as normal C++ compiler. CUDA will *NOT* use that, but check for the variable `CMAKE_CUDA_HOST_COMPILER` and if this is not set, fallback to the default platform compiler, and *NOT* `CXX`. So I guess for CUDA builds we should set `CMAKE_CUDA_HOST_COMPILER` as well.

psychocoderHPC:
> It seems that this PR uncovered a bug in `linux_nvcc-9.1_clang-4_debug`. For whatever reason, our `CC=clang` is ignored and nvcc pulls in `gcc` instead, which is too new for the CUDA version.

For alpaka we do not use `CC` if we set the environment variable `CXX` must be set.

bernhardmgruber (dismissed):

BenjaminW3:
I tried adding `CMAKE_CUDA_HOST_COMPILER=clang++` but it still does not work. I also tried using ubuntu-20.04 instead of 18.04 but it does not provide clang-4.0 by default anymore.
May we drop cuda-9.1+clang-4.0?

psychocoderHPC:
> I tried adding `CMAKE_CUDA_HOST_COMPILER=clang++` but it still does not work. I also tried using ubuntu-20.04 instead of 18.04 but it does not provide clang-4.0 by default anymore.
> May we drop cuda-9.1+clang-4.0?

Yes I am for dropping clang 4.0, both clang 4.0 and cuda 9.1 are not our primary targets anymore.

psychocoderHPC (approved):

PR #1356, opened 2021-07-09
Prepare for 0.7.0 release

j-stephan:
* Incorporate 0.6.1 changelog into 0.7 changelog
* Set release date to 2021-08-03
* Update version number from `0.7.0-rc1` to `0.7.0`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
-## [0.7.0] - 2021-07-09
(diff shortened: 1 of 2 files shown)


sbastrakov (approved):

psychocoderHPC:
IMO 0.7.0 should provide a changelog relative to 0.6.0. The reason is that there is no need that bug fixes will be in a sequential order with other releases.
In PIConGPU we are very explicit and mention the origin of the changelog: https://github.com/ComputationalRadiationPhysics/picongpu/blob/master/CHANGELOG.md#changes-to-040

Never the less I am fine with this PR because 0.6.1 was released before 0.7.0.

psychocoderHPC (approved):

PR #1358, opened 2021-07-13
Add AtomicOaccExtended

jkelling:
Ternary operator is not allowed in atomic capture by the OpenACC standart. This is actually like OpenMP < 5.1, where critical regions are used to implement atomics requiring them (cas, min, max, inc, dec).

The OpenACC standart states "The user should not attempt to implement barrier synchronization, critical sections or locks [...]" We already had a barrier, this commit adds a critical region (which is basically a lock) ... for completeness.


BenjaminW3 (dismissed):

jkelling:
The `linux_hip_clang_*_hip_only` test fail to find `rocrand`, which is unrelated to this PR.

@psychocoderHPC Please also add an `openacc` label.

psychocoderHPC:
Please rebase this PR to pull the fix for HIP

include/alpaka/acc/AccOacc.hpp
```diff
@@ -62,40 +60,32 @@ namespace alpaka
...
-        public concepts::Implements<ConceptBlockSync, AccOacc<TDim, TIdx>>,
-        public concepts::Implements<ConceptIdxGb, AccOacc<TDim, TIdx>>
+    template<typename TDim, typename TIdx>
+    class AccOacc final
+        : public bt::IdxBtLinear<TDim, TIdx>
+        , public math::MathStdLib
+        , public rand::RandStdLib
+        , public TimeStdLib
+        , public warp::WarpSingleThread
+        ,
```
j-stephan, line 70: Is this a result of clang-format doing weird things?

j-stephan (approved):

include/alpaka/acc/AccOacc.hpp
j-stephan: @jkelling Ping.

jkelling: After adding the concepts at the end and removing the `AtomicHierarchy`, I just rand git-clang-format of clang 11 on it. The CI clang-format appears to be happy with it, too.

I do not recall if changed anything else that could have triggered clang-format to change the comma position, although I prefer this style because adding and removing base-classes at the end by changing only one line.

If you are asking about the lone comma specifically, this is because the comment was before the base it belongs to in the old version, clang-format cannot change this.

PR #1360, opened 2021-07-14
Make alpaka/test headers installable

### LINKED ISSUE #1357: Bundle test/common/include with install.
When writing software that uses alpaka, it would be nice to use the header files from `alpaka/test`.  However, those aren't installed by alpaka.

Currently, the `CMakeLists.txt` only says this about tests:
```
if(BUILD_TESTING)
    add_subdirectory("test/")
endif()
```

Why not always add the `test/common/include` to the installed header list?
e.g. somewhere around `set(_ALPAKA_INCLUDE_DIRECTORY "${_ALPAKA_ROOT_DIR}/include")`

SimeonEhrig:
- move all test headers to the include folder
- test headers can be install via cmake option
- headers are not installed by default, because they are not designed
 for productive code
- built-in examples and tests are independent of the cmake option

address Issue #1357


psychocoderHPC:
Please run clang-format

This command contains some useless file endings because it is for PIConGPU but fits to alpaka too.
```
find share tests include test  -iname "*.def" \
  -o -iname "*.h" -o -iname "*.cpp" -o -iname "*.cu" \
  -o -iname "*.hpp" -o -iname "*.tpp" -o -iname "*.kernel" \
  -o -iname "*.loader" -o -iname "*.param" -o -iname "*.unitless" \
  | xargs clang-format-11 -i
```

SimeonEhrig:
@psychocoderHPC @sbastrakov Ready for review

CMakeLists.txt
```diff
@@ -42,6 +42,8 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 option(alpaka_BUILD_EXAMPLES "Build the examples" OFF)
 
 option(BUILD_TESTING "Build the testing tree." OFF)
+
+option(alpaka_INSTALL_TEST_HEADER "Install headers of the namespace alpaka::test. Attention, headers are not designed for productive code, see documentation." OFF)
```
sbastrakov: I think you meant `production`, not `productive`

sbastrakov: If you change, the same line is present in `cmake.rst`

sbastrakov (commented):
Nice! I only have one remark with wording, see above. I didn't test it, only looked at the changes

CMakeLists.txt
SimeonEhrig: Yes your are right. It makes more sense. I fixed it.

BenjaminW3 (approved):

PR #1379, opened 2021-08-05
Add memory fences to kernel language

### LINKED ISSUE #14: add threadFence<>
- [ ] __threadfence_block()
Waits until all GM and SM accesses made by the calling thread prior to __threadfence_block() are visible to all threads in the thread block.
- [ ] __threadfence()
Waits until all GM and SM accesses made by the calling thread prior to the call are visible to:
• threads in the thread block for shared memory accesses and
• threads in the device for global memory accesses.
Prevents the compiler from optimizing by caching shared memory writes in registers.
It does not synchronize the threads and it is not necessary for all threads to actually reach this instruction.
  gcc: __sync_synchronize();
  intel: _mm_mfence();
  MSVC: MemoryBarrier();
  OpenMP: #pragma omp flush
- [ ] __threadfence_system()
waits until all GM and SM accesses made by the calling thread prior to __threadfence_system() are visible to:
• threads in the thread block for shared memory accesses,
• all threads in the device for global memory accesses,
• host threads for page-locked host memory accesses.

j-stephan:
This PR adds memory fences / barriers (also called thread fences) to alpaka's kernel language. Note that these fences provide the same guarantee as CUDA's, i.e. `LoadLoad` and `StoreStore` ordering but **not** `LoadStore` ordering (as is common on CPUs).

Closes #14.


test/unit/mem/fence/CMakeLists.txt
```diff
@@ -0,0 +1,27 @@
...
+                ${_FILES_SOURCE})
+
+        target_link_libraries(
+                ${_TARGET_NAME}
+                PRIVATE common)
+
+        set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER "test/unit")
+
+        add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} ${_ALPAKA_TEST_OPTIONS})
+endif()
```
psychocoderHPC: missing empty line

psychocoderHPC (changes requested):
Only a missing line.

psychocoderHPC:
@j-stephan Could you add a section into the documentation add document the new feature (mostly what you have already in the oxygen) and examples or a description where it is required to use `ALPAKA_DEVICE_VOLATILE`

j-stephan:
Note to potential reviewers: I'm aware of the clang-format, macOS/TBB and OpenACC errors, I'll fix them on Monday (as well as the documentation).

The failing OpenMP 5 test is because the necessary `#pragma` was introduced in clang 11 and we're still testing with clang 10. We can either use OpenMP 2.x's behaviour for older compilers or raise the minimum clang version to 11 for OpenMP 5.

docs/source/basic/cheatsheet.rst
```diff
@@ -256,6 +256,12 @@ Atomic operations
          AtomicAdd, AtomicSub, AtomicMin, AtomicMax, AtomicExch,
          AtomicInc, AtomicDec, AtomicAnd, AtomicOr, AtomicXor, AtomicCas
 
+Memory fences on block- or device level (guarantees LoadLoad and StoreStore ordering)
+  .. code-block:: c++
+
+     mem_fence(acc, memory_scope::Block{});
+     mem_fence(acc, memory_scope::Device{});
```
bernhardmgruber, line 263: Can we not reuse the same hierarchy tags as for the atomics?

include/alpaka/mem/fence/MemFenceOacc.hpp
```diff
@@ -0,0 +1,72 @@
...
+                 *
+                 * As noted in the OpenACC 3.1 spec, section 1.2, programmers should not attempt to implement any form
+                 * of explicit synchronization themselves.
+                 *
+                 * We therefore do not implement a memory fence for OpenACC and instead throw a compile-time error
+                 * which informs the user to use another back-end.
+                 */
+                static_assert(
+                    std::is_same<TMemScope, unsupported>::value,
+                    "Memory fences are not available in the OpenACC back-end");
```
bernhardmgruber: I don't get why you compare to `unsupported` here. Are you trying to create a dependent context for the condition of `static_assert`? If yes, try:
```suggestion
                static_assert(!sizeof(TMemScope), "Memory fences are not available in the OpenACC back-end");
```
See: https://devblogs.microsoft.com/oldnewthing/20200311-00/?p=103553

include/alpaka/mem/fence/MemFenceCpu.hpp
```diff
@@ -0,0 +1,72 @@
...
+    class MemFenceCpu : public concepts::Implements<ConceptMemFence, MemFenceCpu>
+    {
+    public:
+        MemFenceCpu() = default;
+        MemFenceCpu(MemFenceCpu const&) = delete;
+        auto operator=(MemFenceCpu const&) -> MemFenceCpu& = delete;
+        MemFenceCpu(MemFenceCpu&&) = delete;
+        auto operator=(MemFenceCpu&&) -> MemFenceCpu& = delete;
+        ~MemFenceCpu() = default;
+    };
```
bernhardmgruber, line 29: Why do we need to delete all the copy/move operations? This is just a tag type we mixin into the accelerator.
```suggestion
    struct MemFenceCpu : public concepts::Implements<ConceptMemFence, MemFenceCpu> {};
```

include/alpaka/mem/fence/Traits.hpp
```diff
@@ -0,0 +1,66 @@
...
+        //! Memory fences are observed by all threads in the same block.
+        struct Block
+        {
+        };
+
+        //! Memory fences are observed by all threads on the device.
+        struct Device
+        {
+        };
+    } // namespace memory_scope
```
bernhardmgruber, line 32: Do we also need a `System` scope? CUDA has `threadfence_system`.

j-stephan: Nope, because we cannot support this for all back-ends. The problem: A CUDA system fence would also synchronize with memory used by OpenMP, but there is no way to issue an OpenMP fence and synchronize with GPU memory.

include/alpaka/mem/fence/MemFenceOacc.hpp
j-stephan: Neat trick, thanks!

docs/source/basic/cheatsheet.rst
j-stephan: We could, but I was afraid this would cause confusion. We can only fence within a block (which mirrors `ThreadAtomic`) or the whole device (`GridAtomic`), but not between blocks of the same grid (`BlockAtomic`).

include/alpaka/mem/fence/MemFenceCpu.hpp
j-stephan: I just mirrored what we are doing elsewhere in alpaka (see the intrinsics for example). But perhaps we should review the other instances of this, too.

include/alpaka/mem/fence/Traits.hpp
bernhardmgruber: ok!

docs/source/basic/cheatsheet.rst
bernhardmgruber: Yeah I see, better have a separate hierarchy!

include/alpaka/mem/fence/MemFenceCpu.hpp
sbastrakov: FYI for intrinsics I also mirrored how it was done elsewhere, without much thinking on that

j-stephan: Thanks for the info. Guess we'll have to do an alpaka code review at some point. However, this is out of scope for this PR.

test/unit/mem/fence/CMakeLists.txt
j-stephan: Done.

j-stephan:
Ready to merge.

include/alpaka/core/Common.hpp
```diff
@@ -102,3 +102,17 @@
...
+//! This macro disables memory optimizations for annotated device memory.
+//!
+//! Example:
+//!   ALPAKA_DEVICE_VOLATILE float* ptr;
+//!
+//! This is useful for pointers, (shared) variables and shared memory which are used in combination with
+//! the alpaka::mem_fence() function. It ensures that memory annotated with this macro will always be written directly
+//! to memory (and not to a register or cache because of compiler optimizations).
+#if(BOOST_LANG_CUDA || BOOST_LANG_HIP)
+#    define ALPAKA_DEVICE_VOLATILE volatile
```
psychocoderHPC, line 116: If you only like to activate on the GPU device but not on cpu you must write 

```
#if(BOOST_LANG_CUDA && BOOST_ARCH_PTX) || (BOOST_LANG_HIP && defined(__HIP_DEVICE_COMPILE__) && __HIP_DEVICE_COMPILE__ == 1)
```

psychocoderHPC (dismissed):

include/alpaka/core/Common.hpp
j-stephan: Done.

cmake/alpakaCommon.cmake
```diff
@@ -279,18 +279,22 @@ if(ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR A
...
-            elseif(OpenMP_CXX_VERSION VERSION_LESS 5.0)
-                message(WARNING "OpenMP < 5.0, for ALPAKA_ACC_ANY_BT_OMP5_ENABLE 5.0 is recommended.")
+            if(OpenMP_CXX_VERSION VERSION_LESS 5.0)
+                message(FATAL_ERROR "ALPAKA_ACC_ANY_BT_OMP5_ENABLE requires compiler support for OpenMP 5.0.")
+
+                if((${CMAKE_CXX_COMPILER_ID} STREQUAL "AppleClang") AND (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 12.0.5))
+                    message(FATAL_ERROR "The OpenMP 5.0 back-end requires Xcode 12.5 or later")
+                elseif((${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang") AND (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 11.0))
+                    message(FATAL_ERROR "The OpenMP 5.0 back-end requires clang 11.0 or later")
+                endif()
```
bernhardmgruber, line 289: Have we agreed on dropping support for OpenMP 4+? I am fine, but I think we did not have a discussion.

docs/source/basic/library.rst
```diff
@@ -115,6 +115,68 @@ The ``alpaka::getWorkDiv`` and the ``alpaka::getIdx`` functions both return a ve
 They are parametrized by the origin of the calculation as well as the unit in which the values are calculated.
 For example, ``alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)`` returns a vector with the extents of the grid in units of threads.
 
+Memory fences
+`````````````
+
+**Note**: Memory fences should not be mistaken for synchronization functions between threads. They solely enforce the
+ordering of certain memory operations (see below) and how other threads observe this order. If you need to rely on the
+results of memory operations being visible to other threads please use ``alpaka::syncBlockThreads`` or atomic functions
+instead.
```
bernhardmgruber: I don't think the wording is fully clear here, because a fence *is* used to influence the order of how changes from one thread become visible to another thread. I would add some emphasis:

```suggestion
**Note**: Memory fences should not be mistaken for synchronization functions between threads.
They solely enforce the ordering of certain memory operations (see below) and restrict the possibilities of how other threads *can* observe this order.
If you need to guarantee that the results of memory operations are visible to other threads you must use ``alpaka::syncBlockThreads`` or atomic functions instead.
```

include/alpaka/mem/fence/MemFenceCpu.hpp
bernhardmgruber: Then please fork off an issue next time! There is a button right here in the `...` menu on github.
I did it now: #1396

cmake/alpakaCommon.cmake
j-stephan: Well, it's either this or no support for `#pragma omp flush acq_rel` and having to resort to older OpenMP standards. But admittedly I just assumed that we'd be okay with this since we're also okay with raising the required compiler version quite significantly.

docs/source/basic/library.rst
j-stephan: I don't really see a difference between those sentences, sorry. I'm also not certain that fences restrict "the possibilities of how other threads can observe this order" since there aren't any possibilities left once a fence is issued.

bernhardmgruber: >I'm also not certain that fences restrict "the possibilities of how other threads can observe this order" since there aren't any possibilities left once a fence is issued.

Of course there are possibilities:

```
int a = 0, b = 0;
void thread1() {
    a = 1;
    fence();
    b = 3;
}
void thread2() {
    read(a);
    fence();
    read(b);
}
```

Thread 2 can observe a=0/b=0 or a=1/b=0 or a=1/b=3. However, the possibility of a=0/b=3 is not possible because of the fence. The fence thus restricts the 4 possible variable assignments to 3. The behavior of thread 2 is not deterministic by the fence. That is also why I wrote that thread 2 **can** observe the behavior, because in the case of a=0/b=0 it does not observe the changes from thread 1.

docs/source/basic/library.rst
```diff
@@ -115,6 +115,68 @@ The ``alpaka::getWorkDiv`` and the ``alpaka::getIdx`` functions both return a ve
...
+
+        /* Possible results at this point:
+         * a == 1 && b == 2
+         * a == 10 && b == 2
+         * a == 10 && b == 20
+         *
+         * but NOT:
+         * a == 1 && b == 20
+         */
+    }
```
bernhardmgruber, line 179: I am starting to think that this code is undefined in C++, as it is missing an atomic operation. See: https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence

> Fence-fence synchronization
> A release fence FA in thread A synchronizes-with an acquire fence FB in thread B, if
> There exists an atomic object M,

This code does not have an atomic, and volatile does replace this behavior.

docs/source/basic/library.rst
j-stephan: You're talking about other threads observing different **results**, but the **order of instructions** is still the same. With the fence it is guaranteed that a is written before b, but the observed results can be different.

bernhardmgruber: I see. Then please rephrase at least:
```suggestion
**Note**: Memory fences should not be mistaken for synchronization functions between threads. They solely enforce the
ordering of certain memory instructions (see below) and restrict how other threads can observe this order. If you need to rely on the
results of memory operations being visible to other threads please use ``alpaka::syncBlockThreads`` or atomic functions
instead.
```
This way we are specific about instructions. And a fence does not enforce how other threads observe results, it merely restricts what observations are possible.

docs/source/basic/library.rst
j-stephan: Ugh, I thought I could do this without having to call another atomic operation (on godbolt the correct CPU instructions are generated by `std::atomic_thread_fence`), but you're right.

bernhardmgruber (approved):

PR #1380, opened 2021-08-06
CI: test CUDA 11.4

### LINKED ISSUE #1377: CUDA 11.4
CUDA 11.4.1 is already released we should add tests to our CI.

https://developer.nvidia.com/cuda-downloads

psychocoderHPC:
fix: #1377

CI test CUDA 11.4.

- update README.txt
- CI:
  - add CUDA 11.4 test
  - add rules to install CUDA 11.4


.github/workflows/ci.yml
```diff
@@ -571,6 +580,53 @@ jobs:
           os: ubuntu-latest
           env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 11,     ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.72.0, ALPAKA_CI_CMAKE_VER: 3.19.7,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CI_CUDA_VERSION: "11.3", CMAKE_CUDA_COMPILER: nvcc, CMAKE_CUDA_ARCHITECTURES: "86",              ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF}
 
+        ## CUDA 11.4
```
j-stephan, line 583: I'm a bit sceptical about adding additional tests to the matrix. Our CI is already pretty slow, adding more tests only makes it more complicated. Maybe we can remove CUDA 9.0 (and possibly 9.1) so that we have roughly the same amount of tests?

j-stephan (changes requested):

.github/workflows/ci.yml
psychocoderHPC: We need to test CUDA 11.4 there is no way around it. I also thought about removing maybe CUDA 9.x, this is something we can do in a follow-up PR, that's something I would like to discuss in the developer meeting first.
IMO dropping compiler support should not be CI motivated!

j-stephan (approved):

BenjaminW3 (approved):

bernhardmgruber (approved):

Issue #1220, opened 2020-11-24
Discussion: Reduce legacy support / future support


j-stephan:
In today's alpaka VC we raised the question of dropping some legacy stuff in favour of C++17 some time **after** the next release. We currently aim to do this by mid-2021. However, this might affect some of our users outside of HZDR who are hereby invited to participate in this discussion. @fwyzard, @bertwesarg and the DLR come to mind, but everyone should feel free to notify other non-HZDR alpaka users they know of.

Current candidates for removal are (feel free to expand this list):

* old CUDA versions (everything older than CUDA 10, maybe CUDA 10 as well, point releases that are not the latest versions)
* old MSVC versions (everything older than the latest stable version)

Things to consider:

* Will this seriously break existing workflows?
* Will (older) HPC systems no longer be supported?

Another question that was raised today is if we want to support Apple's future ARM-based products. Right now we have no developers / users (that I know of) who could help us.

fwyzard:
Thanks for the heads up.
CMS is already using c++17, CUDA 11 and gcc 8 or later - so I think we should be fine.

j-stephan:
DLR responded per mail:

* They don't care about MSVC
* They require CUDA 10.x but don't care about older releases
* They wouldn't like a full transition to C++17 because this affects their users who still work with older codes.

bernhardmgruber:
> DLR responded per mail:
> * They wouldn't like a full transition to C++17 because this affects their users who still work with older codes.

That statement is a bit vague. Does that mean we need to stick with C++14 fully? Or are they just afraid of C++17? Do they tolerate some C++17 features?

> * They require CUDA 10.x but don't care about older releases

That kind of binds us to C++14 when compiled with nvcc.

Could you still please ask them for a minimum host compiler version they would like to suport? Thank you!

j-stephan:
> That statement is a bit vague. Does that mean we need to stick with C++14 fully? Or are they just afraid of C++17? Do they tolerate some C++17 features?

Translation of the sentence in question (original in German):

> The switch to C++17 is more critical. We wouldn't want this so soon because of alpaka's header-only concept. This would cascade through to our users.

Sounds like a hard C++14 limit to me.

> Could you still please ask them for a minimum host compiler version they would like to suport? Thank you!

Will do.

j-stephan:
DLR responded. They require compatibility for:

* CUDA 10.2.89
* CUDA 10.1.105
* gcc >= 5.5.0
[...]

PR #1375, opened 2021-08-05
Add runtime tests on Gitlab CI

SimeonEhrig:
The main goal of this PR is to generate work with finding runtime bugs ;-)

# About

The alpaka repository is mirrored on GitLab where we registered custom GitLab CI runners. The runner provides different hardware configurations (12/16 cores x86 CPU, Nvidia GPU, AMD GPU, ARM CPU and Power 8 CPU). This allows us to automatically run runtime tests with all accelerator backends instead of just the CPU backend as the current CI can only do.

This PR contains only two basic jobs, to initial set up the GitLab CI.

PR #1295, opened 2021-04-19
CUDA 11.3 support

psychocoderHPC:
fix #1293 

- fix compile issues shown up with CUDA 11.3
- CI add CUDA 11.3
- update documentation

- [x] rebase against #1294

PR #1385, opened 2021-08-10
Purge legacy software support

j-stephan:
This PR reduces the size of our CI by removing legacy software support. Based on the discussion in #1220:

* removed gcc < 7
* removed clang < 5
* removed CUDA < 9.2
* removed MSVC < 2019
* removed discontinued Xcode checks except for 11.3.1 (macOS 10.14) and 12.4 (macOS 10.15)
* added clang 12
* added Xcode 12.5.1 (macOS 11)

This PR also introduces a change in `install_clang.sh` which ensures that we are installing the matching libc++ for the chosen clang version.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,9 +16,9 @@ on: [push, pull_request]
 # NOTE: Testing the full matrix is not practical.
 # Therefore we aim to have each value been set in at lest one job.
 # CXX                                           : {g++, clang++}
-#   [g++] ALPAKA_CI_GCC_VER                     : {5, 6, 7, 8, 9, 10}
-#   [clang++] ALPAKA_CI_CLANG_VER               : {4.0, 5.0, 6.0, 7, 8, 9, 10, 11}
-#   [cl.exe] ALPAKA_CI_CL_VER                   : {2017, 2019}
+#   [g++] ALPAKA_CI_GCC_VER                     : {7, 8, 9, 10}
+#   [clang++] ALPAKA_CI_CLANG_VER               : {5.0, 6.0, 7, 8, 9, 10, 11}
+#   [cl.exe] ALPAKA_CI_CL_VER                   : {2019}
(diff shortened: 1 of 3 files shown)


j-stephan:
Personally I'm also in favour of removing CUDA 10.x and 11.x that are not the latest versions.

SimeonEhrig:
> Personally I'm also in favour of removing CUDA 10.x and 11.x that are not the latest versions.

We should keep CUDA 10.1, because until Clang 10, Clang as CUDA Compiler supports only CUDA 10.1 or older.
https://gist.github.com/ax3l/9489132#clang--x-cuda

psychocoderHPC:
> Personally I'm also in favour of removing CUDA 10.x and 11.x that are not the latest versions.

As I am writing in the developer chat we should keep CUDA 10.X and 11.X. NVIDIA is doing breaking changes in minor versions. 
An example can be found in #1295.

After we use the HZDR CI #1375 we could think about reducing the test matrix for the PULL requests and perform a more complex test matrix each night or once per week.

j-stephan:
It seems that macOS 11 is only available as a non-public preview now. I have requested preview access, let's see what happens.

SimeonEhrig:
> After we use the HZDR CI #1375 we could think about reducing the test matrix for the PULL requests and perform a more complex test matrix each night or once per week.

I was also thinking if maybe we can work with different test sets for pull requests. Maybe a default set `quick` that contains e.g. only one test per Acc compiler combination and a set `full` that contains the complete test matrix and is executed when the PR is ready to merge.

bernhardmgruber (dismissed):

psychocoderHPC:
I removed that github enforces that tests you removed in this PR are required before PRs can be merged.
This PR must be rebased against the develop branch after #1404 is merged to the dev branch.

psychocoderHPC:
PR can now be rebased against develop.

bernhardmgruber (approved):

PR #1386, opened 2021-08-10
fix `ALPAKA_STATIC_ACC_MEM_CONSTANT` memory

psychocoderHPC:
- Usage of `ALPAKA_STATIC_ACC_MEM_CONSTANT` and `ALPAKA_STATIC_ACC_MEM_GLOBAL`  was broken for clang CUDA and HIP
- HIP: fix `hipGetSymbolAddress` interface was wrongly used
- remove tests for direct initialization
- update documentation

# Usage of both memory types changed!!
Remove the possibility to initialize both types of memory during the definition. (more information)[https://github.com/alpaka-group/alpaka/pull/1386#issuecomment-896962596)

Co-authored-by: Simeon Ehrig <s.ehrig@hzdr.de>

### DIFF
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -76,6 +76,20 @@
 //! In contrast to ordinary variables, you can not define such variables
 //! as static compilation unit local variables with internal linkage
 //! because this is forbidden by CUDA.
+//!
+//! \attention It is not allowed to initialize the variable together with the declaration.
+//!            To initialize the variable alpaka::createStaticDevMemView and alpaka::memcpy must be used.
+//! \code{.cpp}
+//! ALPAKA_STATIC_ACC_MEM_GLOBAL int foo;
+//!
+//! void initFoo() {
+//!     auto extent = alpaka::Vec<alpaka::DimInt<1u>, size_t>{1};
+//!     auto viewFoo = alpaka::createStaticDevMemView(&foo, device, extent);
+//!     int initialValue = 42;
+//!     alpaka::ViewPlainPtr<DevHost, int, alpaka::DimInt<1u>, size_t> bufHost(&initialValue, devHost, extent);
+//!     alpaka::memcpy(queue, viewGlobalMemUninitialized, bufHost, extent);
+//! }
+//! \endcode
 #if((BOOST_LANG_CUDA && BOOST_COMP_CLANG) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX) || BOOST_LANG_HIP)
 #    define ALPAKA_STATIC_ACC_MEM_GLOBAL __device__
 #else
@@ -97,6 +111,20 @@
 //! In contrast to ordinary variables, you can not define such variables
 //! as static compilation unit local variables with internal linkage
 //! because this is forbidden by CUDA.
+//!
+//! \attention It is not allowed to initialize the variable together with the declaration.
+//!            To initialize the variable alpaka::createStaticDevMemView and alpaka::memcpy must be used.
+//! \code{.cpp}
+//! ALPAKA_STATIC_ACC_MEM_CONSTANT int foo;
+//!
+//! void initFoo() {
+//!     auto extent = alpaka::Vec<alpaka::DimInt<1u>, size_t>{1};
+//!     auto viewFoo = alpaka::createStaticDevMemView(&foo, device, extent);
+//!     int initialValue = 42;
+//!     alpaka::ViewPlainPtr<DevHost, int, alpaka::DimInt<1u>, size_t> bufHost(&initialValue, devHost, extent);
+//!     alpaka::memcpy(queue, viewGlobalMemUninitialized, bufHost, extent);
+//! }
+//! \endcode
 #if((BOOST_LANG_CUDA && BOOST_COMP_CLANG) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX) || BOOST_LANG_HIP)
 #    define ALPAKA_STATIC_ACC_MEM_CONSTANT __constant__
 #else
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -76,7 +76,7 @@
 //! In contrast to ordinary variables, you can not define such variables
 //! as static compilation unit local variables with internal linkage
 //! because this is forbidden by CUDA.
-#if(BOOST_LANG_CUDA && BOOST_ARCH_PTX) || (BOOST_LANG_HIP && (BOOST_ARCH_HSA || BOOST_ARCH_PTX))
+#if((BOOST_LANG_CUDA && BOOST_COMP_CLANG) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX) || BOOST_LANG_HIP)
 #    define ALPAKA_STATIC_ACC_MEM_GLOBAL __device__
 #else
 #    define ALPAKA_STATIC_ACC_MEM_GLOBAL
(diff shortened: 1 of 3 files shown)


include/alpaka/core/Common.hpp
```diff
@@ -97,7 +97,7 @@
 //! In contrast to ordinary variables, you can not define such variables
 //! as static compilation unit local variables with internal linkage
 //! because this is forbidden by CUDA.
-#if(BOOST_LANG_CUDA && BOOST_ARCH_PTX) || (BOOST_LANG_HIP && (BOOST_ARCH_HSA || BOOST_ARCH_PTX))
```
psychocoderHPC, line 100: `BOOST_ARCH_PTX` is derived from `CUDA_ARCH` which is not defined on the host side, therefore the guard on host side was always false for CUDA, therefore the qualifier of `ALPAKA_STATIC_ACC_MEM_CONSTANT` was wrong.
For unknown reason, CUDA accepted it but this is against any CUDA example code and IMO we had luck that it worked before.

j-stephan (dismissed):

psychocoderHPC:
I am fighting with some HIP specifics, if under HIP 

```
ALPAKA_STATIC_ACC_MEM_CONSTANT Elem g_constantMemory2DInitialized[3][2] = {{0u, 1u}, {2u, 3u}, {4u, 5u}};
```

... is used where the variable is default initialized then it is only initialized for the device but not on the host side. 
In case we have multiple accelerators enabled the variable is default initialized with zeros :-(

SimeonEhrig:
> I am fighting with some HIP specifics, if under HIP
> 
> ```
> ALPAKA_STATIC_ACC_MEM_CONSTANT Elem g_constantMemory2DInitialized[3][2] = {{0u, 1u}, {2u, 3u}, {4u, 5u}};
> ```
> 
> ... is used where the variable is default initialized then it is only initialized for the device but not on the host side.
> In case we have multiple accelerators enabled the variable is default initialized with zeros :-(

What do you mean with it is not initialized on the host? Do you mean, if we ask for the memory address (`auto const viewConstantMemInitialized = alpaka::createStaticDevMemView(&g_constantMemory2DInitialized[0u][0u], devAcc, extent);`) and  we copy constant memory data to a host buffer afterwards, we get only `0`s?

psychocoderHPC:
Our tests in alpaka are testing accessing contact memory initialized at compile time. If two accelerators are enabled the variable must be defined on host and device. For CUDA we work around it by guarding `__consttant__` with `#ifdef __CUDA_ARCH__`, this is accepted by `clang++ -x cuda` and `nvcc`.
Currently, I have not found an equivalent workaround for HIP.

```C++
#include "cuda_runtime.h"
#include <cstdio>
#include <cstdint>

// not allowed to access from host, same is true for __device__ variables
// alpaka is only producing valid code because we guard the qualifier `__contant__`
// if the guard is removed and __constant__ is always added the variable can not be read on the host side
#ifdef __CUDA_ARCH__
__constant__
#endif
int gGPUConstantMemBuffer = {42};

#define CHECK_HIP(x) check_hip(x, __FILE__, __LINE__)

static int check_hip(const long long int error, const char* file, int line)
{
  if (error == cudaSuccess) {
    return (0);
  }
  printf("HIP Error: %lld / %s (%s:%d)\n", error, cudaGetErrorString((cudaError_t)error), file, line);
  return 1;
}

__global__
void hello(int* x)
{
        *x = gGPUConstantMemBuffer;
        printf("device value = %i\n", gGPUConstantMemBuffer);
}

int main(int, char**)
{
    int* devValue;
    CHECK_HIP(cudaMalloc(&devValue,sizeof(int)));
    int x = 0;
    dim3 dimBlock( 1, 1 );
    dim3 dimGrid( 1, 1 );
    hello<<<dim3(dimGrid), dim3(dimBlock), 0, 0>>>( devValue);
    CHECK_HIP(cudaMemcpy( &x, devValue, sizeof(int), cudaMemcpyDeviceToHost));
    printf("host value = %i but should be %i\n", gGPUConstantMemBuffer,x);
    CHECK_HIP(cudaDeviceSynchronize());

    return 0;
}
```

psychocoderHPC:
I figuard out that 

```
#ifdef __HIP_DEVICE_COMPILE__
__constant__
#endif
int gGPUConstantMemBuffer = {42};
```

is equivalent to the CUDA workaround, I will test it with alpaka again.

SimeonEhrig:
> I figuard out that
> 
> ```
> #ifdef __HIP_DEVICE_COMPILE__
> __constant__
> #endif
> int gGPUConstantMemBuffer = {42};
> ```
> 
> is equivalent to the CUDA workaround, I will test it with alpaka again.

What is the code doing? Looks strange. Looks like, that you want to remove the `__constant__` keyword, if the host compiler of hip is compiling the code.

psychocoderHPC:
> What is the code doing? Looks strange. Looks like, that you want to remove the `__constant__` keyword, if the host compiler of hip is compiling the code.

It is creating a variable on host and device. The strange thing is that this is valid for with `nvcc` and `clang` but if you like to get the symbol address via

```
int* devPtrConstantMem;
    int *sym = &gGPUConstantMemBuffer;
    if (CHECK_HIP(cudaGetSymbolAddress((void**)&devPtrConstantMem, *sym))) {
        printf("Error getting ptr to constant memory\n");
        return 1;
    }
```

It works only with `nvcc` but NOT with `clang (CUDA or HIP)`. Since we do not have CUDA runtime tests we never realized it because nobody is using clang CUDA :-(
Now I am aware and try to solve the issue, somehow.

psychocoderHPC:
The feature `ALPAKA_STATIC_ACC_MEM_CONSTANT` and `ALPAKA_STATIC_ACC_MEM_GLOBAL` was broken for clang CUDA too :-(

I removed now the possibility to initialize both types of memory during the declaration. 
The behavior of how such type of memory should be used in alpaka will be changed with this PR.
I was not able to find any valid way to keep the initialization without making the usage and declaration extremely complicated.
I will extend the documentation to describe how to use the memory.

psychocoderHPC:
@j-stephan ready for review

j-stephan (dismissed):

include/alpaka/core/Common.hpp
```diff
@@ -76,7 +76,21 @@
...
+//!
+//! \attention It is not allowed to initialize the variable together with the declaration.
```
sbastrakov, line 80: Should be `definition`, not `declaration`? It is generally not allowed at declarations that are not definitions

sbastrakov: Same is also below

psychocoderHPC: updated

include/alpaka/core/Common.hpp
```diff
@@ -76,7 +76,21 @@
...
+//!            To initialize the variable alpaka::createStaticDevMemView and alpaka::memcpy must be used.
```
bernhardmgruber, line 81:
```suggestion
//!            To initialize the variable alpaka::createStaticDevMemView an alpaka::memcpy must be used.
```

include/alpaka/core/Common.hpp
```diff
@@ -97,7 +111,21 @@
...
```
bernhardmgruber, line 116:
```suggestion
//!            To initialize the variable alpaka::createStaticDevMemView an alpaka::memcpy must be used.
```

bernhardmgruber (approved):
Please fix the two typos, the rest LGTM.

include/alpaka/core/Common.hpp
psychocoderHPC: `and` is fine at this point. You must create the view and use `alpaka::memcpy` to initialize the instance.

bernhardmgruber: Ok then I misread the sentence.

PR #1387, opened 2021-08-10
Add third_party STLTuple, use in TaskKernel{Oacc,Omp5}

### LINKED ISSUE #1381: OpenMP 5 back-end broken for clang 12
Compiling the alpaka tests with clang 12 and an activated OpenMP 5 back-end leads to compilation errors:

```
In file included from /home/stepha27/workspace/caravan/alpaka/test/integ/sharedMem/src/sharedMem.cpp:10:
In file included from /home/stepha27/workspace/caravan/alpaka/include/alpaka/alpaka.hpp:112:
/home/stepha27/workspace/caravan/alpaka/include/alpaka/kernel/TaskKernelOmp5.hpp:191:29: error: Type 'std::tuple<int *>' is not trivially copyable and not guaranteed to be mapped correctly [-Werror,-Wopenmp-mapping]
                            argsD);
                            ^~~~~
```

The culprit is this member variable of `TaskKernelOmp5`:

https://github.com/alpaka-group/alpaka/blob/c938c8045bb9592fc1ee6a05a440d3cfc279c2d3/include/alpaka/kernel/TaskKernelOmp5.hpp#L206

I didn't test this with clang 11 yet. In clang 10 this works because our CI doesn't complain.

jkelling:
This PR uses a `tuple` [implementation](https://github.com/codeplaysoftware/computecpp-sdk/blob/master/include/stl-tuple/STLTuple.hpp) which `is_trivially_copyable` if all component types are [suggested](https://github.com/alpaka-group/alpaka/issues/1381#issuecomment-895853621) by @j-stephan , thereby removing the clang warning about `std::tuple<of, anything>` not being trivially copyable and thus potentially not save to map to target.

1. The included library is included in `third_party`, so that the include directory can be added as `SYSTEM` to portably silence warning immanating fro this file. The file is currently not installed, so currently an installed alpaka would not work. I need directions on how this should be handled:
   * installing a third-party library outside of `include/alpaka` would be problematic, because this could break other installations of that third-party lib in the same prefix.
   * installing the file in `alpaka/include` would change its path between build and install, breaking includes (except if cmake `configure(FILE...)` is used)
   * placing it directly inside `include/alpaka` in this source tree would make it impossible to use `target_include_directories( ... SYSTEM ... )` on it to silence warnings with all compilers, which would make it necessary to change the file.
2. [include/alpaka/core/Tuple.hpp](https://github.com/jkelling/alpaka/blob/14cf2d69e39a06da63959c8d5fee95a59fd24d8f/include/alpaka/core/Tuple.hpp) is added to provide an `apply` implementation working for this `Tuple` type (taken from #789). While it is located inside `include/alpaka` (and uses alpaka style) it currently adds to the namespaces from `STLTuple`, because it logcally extents it. To resolve this inconsistency some policy decisions must be taken first:
   1. Point 1. above.
   2. It is located in `core/` because `Tuple` is provided through it. It would belong int `meta/` instead of `core/` as it only adding `apply()` if `STLTuple` were to be moved into the `include/alpaka` tree.
   3. In `meta/` it would somewhat collide with the existing `ApplyTuple.hpp`.
   4. It could make `ApplyTuple.hpp` if `STLTuple` were to be used exclusively internally instead of `std::tuple`. Is this possible and desirable?

This is note marked as draft, because it is done as far as I know how to do it at the moment and needs discussion.

closes #1381

### DIFF
--- a/include/alpaka/core/STLTuple/LICENSE.txt
+++ b/include/alpaka/core/STLTuple/LICENSE.txt
@@ -1,0 +1,33 @@
+/* Copyright 2016 Codeplay Software Ltd.
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+   Copyright 2016 Codeplay Software Ltd.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use these files except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   For your convenience, a copy of the License has been included in this
+   repository.
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+   The files samples/gaussian_blur/stb_image.h and stb_image_write.h are
+   in the Public Domain. Originally authored by Sean Barrett.
+   https://github.com/nothings/stb
+
+   The file samples/smart_pointer/stack_allocator.hpp is (c) Charles Salvia
+   and is available under the MIT License.
+   https://github.com/charles-salvia/charles
+
--- a/include/alpaka/core/STLTuple/STLTuple.hpp
+++ b/include/alpaka/core/STLTuple/STLTuple.hpp
@@ -1,0 +1,302 @@
+/***************************************************************************
+ *
+ *  Copyright (C) 2018 Codeplay Software Limited
+ *  Licensed under the Apache License, Version 2.0 (the "License");
+ *  you may not use this file except in compliance with the License.
+ *  You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *  For your convenience, a copy of the License has been included in this
+ *  repository.
+ *
+ *  Unless required by applicable law or agreed to in writing, software
+ *  distributed under the License is distributed on an "AS IS" BASIS,
+ *  WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *  See the License for the specific language governing permissions and
+ *  limitations under the License.
+ *
+ * STLTuple.h
+ *
+ * \brief:
+ *  Minimal implementation of std::tuple that is standard layout.
+ *
+  Authors:
+ *
+ *    Mehdi Goli    Codeplay Software Ltd.
+ *    Ralph Potter  Codeplay Software Ltd.
+ *    Luke Iwanski  Codeplay Software Ltd.
+ *
+ **************************************************************************/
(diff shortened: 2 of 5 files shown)


jkelling:
@j-stephan @psychocoderHPC  @BenjaminW3 @sbastrakov ... any thoughts how to treat and where to put `STLTuple.cpp`?

jkelling:
@j-stephan would including this eliminate an external dependency in #789 ? Or do you need more from that library?

j-stephan:
Ah sorry, I wanted to respond yesterday but then other stuff interfered and I forgot :-(

> installing the file in `alpaka/include` would change its path between build and install, breaking includes (except if cmake `configure(FILE...)` is used)

This is probably solvable through CMake's generator expressions (`$<BUILD_INTERFACE:...>` and `$<INSTALL_INTERFACE:...>`).

> In meta/ it would somewhat collide with the existing ApplyTuple.hpp.

Can't we merge the two files?

> It could make `ApplyTuple.hpp` if `STLTuple` were to be used exclusively internally instead of `std::tuple`. Is this possible and desirable?

I don't see any immediate problems with this, but I would like for the other maintainers to chime in on this one.

> @j-stephan would including this eliminate an external dependency in #789 ? Or do you need more from that library?

IIRC that would indeed eliminate the external dependency.

jkelling:
> > installing the file in `alpaka/include` would change its path between build and install, breaking includes (except if cmake `configure(FILE...)` is used)
> 
> This is probably solvable through CMake's generator expressions (`$<BUILD_INTERFACE:...>` and `$<INSTALL_INTERFACE:...>`).

The key issue is not installing the header from the `third_party` directory.  The issue is moving in into `include/alpaka` while under `third_party` it is located in `include/stl-tuple` because this is its place in the original library. This would require the `#include <>` in `Tuple.hpp` to change between build and install. We cannot install to `include/stl-tuple` because this would conflict with an installation of the original library.

> > In meta/ it would somewhat collide with the existing ApplyTuple.hpp.
> 
> Can't we merge the two files?

It is an option to place the code from `Tuple.hpp` in `ApplyTuple.hpp`.

There is also the option to move the code from `STLTuple.hpp` into `Tuple.hpp`, so that is becomes part of alpaka. To make this work we need to fix the code to not trigger warning and to conform with alpaka's coding style. The license should allow this, I just to not know what the copyright notice would need to look like. Apart form that this appears to be the simplest solution.

psychocoderHPC:
@jkelling This PR has in the title still WIP, is this PR ready for a review and merge?

jkelling:
> @jkelling This PR has in the title still WIP, is this PR ready for a review and merge?

No. I did not get to implementing the approach discussed in the meeting. I will convert it to draft until that is done.

jkelling:
Ping @psychocoderHPC for review.

@j-stephan : after this is merged, SYCL can `#    include <alpaka/core/Tuple.hpp>` and use `alpaka::core::Tuple`. Maybe you could try to rebase the SCYL branch and see if this works for your before the merge.

j-stephan:
I don't know if I'll get my hands on SYCL in the next days, but I don't expect any problems with the new include. And it is only used in one place so the amount of required changes is quite small.

jkelling:
MSVC was not happy with the `STLTuple` code. Added a suppressed warning, as, I guess, we do not want to touch this code more than absolutely necessary. CI should pass in this round.

psychocoderHPC (approved):

PR #1393, opened 2021-08-11
Implement tuple interface for Vec to allow structured bindings

### LINKED ISSUE #1390: alpaka::Vec cannot be destructured
`alpaka::Vec<alpaka::DimInt<N>, ...>` is a simple type holding `N` numbers, so it should be reasonable to be able to decompose it into N such values:

```c++
using Vec = alpaka::Vec<alpaka::DimInt<2>, int>;
Vec v(1, 2);
auto [x, y] = v; // error: cannot decompose private member 'm_data' of 'alpaka::Vec<std::integral_constant<unsigned long, 2>, int>'
```
However, that does not work because `alpaka::Vec` does not implement the necessary tuple protocol. Please consider adding this to alpaka:
```c++
namespace alpaka
{
    template<size_t I, typename TDim, typename TVal>
    auto get(Vec<TDim, TVal> v)
    {
        return v[I];
    }
} // namespace alpaka

namespace std
{
    template<typename TDim, typename TVal>
    struct tuple_size<alpaka::Vec<TDim, TVal>> : integral_constant<size_t, TDim::value>
    {
    };

    template<size_t I, typename TDim, typename TVal>
    struct tuple_element<I, alpaka::Vec<TDim, TVal>>
    {
        using type = TVal;
    };
} // namespace std
```

bernhardmgruber:
Fixes #1390.

### DIFF
--- a/test/unit/vec/src/VecTest.cpp
+++ b/test/unit/vec/src/VecTest.cpp
@@ -280,3 +280,24 @@ TEMPLATE_LIST_TEST_CASE("vecNDConstructionFromNonAlpakaVec", "[vec]", alpaka::te
         REQUIRE(nonAlpakaVec[d] == alpakaVec[d]);
     }
 }
+
+#ifdef __cpp_structured_bindings
+TEST_CASE("structuredBindings", "[vec]")
+{
+    using Dim = alpaka::DimInt<2u>;
+    using Idx = std::size_t;
+    using Vec = alpaka::Vec<Dim, Idx>;
+
+    Vec vec{Idx{1}, Idx{2}};
+
+    const auto [y, x] = vec;
+    CHECK(y == 1);
+    CHECK(x == 2);
+
+    auto& [y2, x2] = vec;
+    y2++;
+    x2++;
+    CHECK(vec[0] == 2);
+    CHECK(vec[1] == 3);
+}
+#endif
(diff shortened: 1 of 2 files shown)


BenjaminW3 (dismissed):

test/unit/vec/src/VecTest.cpp
```diff
@@ -280,3 +280,22 @@ TEMPLATE_LIST_TEST_CASE("vecNDConstructionFromNonAlpakaVec", "[vec]", alpaka::te
...
+
+TEST_CASE("structuredBindings", "[vec]")
+{
+    using Dim = alpaka::DimInt<2u>;
+    using Idx = std::size_t;
+    using Vec = alpaka::Vec<Dim, Idx>;
+
+    Vec vec{Idx{1}, Idx{2}};
+
+    const auto [y, x] = vec;
```
j-stephan, line 293: Remember that we're still on C++14 by default. You'll need to enable C++17 explicitly for the test case. Maybe also create a new test target so that we don't require C++17 for the rest of VecTest.

j-stephan (changes requested):

test/unit/vec/src/VecTest.cpp
bernhardmgruber: I disabled the test if structured bindings are not supported.

j-stephan (approved):

PR #1398, opened 2021-08-12
create view factory methods for views

psychocoderHPC:
alpaka is providing a helper factory method to create a view to static constant and global memory via `createStaticDevMemView()`.
Such a simple method is missing to create views from a pointer, `std::vector`,  or `std::array`.

This PR provides new factory methods to create Views or SubView without the required knowledge of the device type by using the device instants and derive dimensionality from the extend, equally to `createStaticDevMemView()`.
The cheatsheet is extended with the new methods.

```C++
int* ptr = nullptr;
using View = alpaka::ViewPlainPtr<Dev, int, Dim, Idx>;
View dst(ptr, dev, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));
```

is becoming

```C++
int* ptr = nullptr;
auto src = alpaka::createView(ptr, dev, Vec(1u));
```

- [x] rebase against #1404

### DIFF
--- a/include/alpaka/mem/view/Traits.hpp
+++ b/include/alpaka/mem/view/Traits.hpp
@@ -20,8 +20,10 @@
 #include <alpaka/queue/Traits.hpp>
 #include <alpaka/vec/Vec.hpp>
 
+#include <array>
 #include <iosfwd>
 #include <type_traits>
+#include <vector>
 
 namespace alpaka
 {
@@ -101,6 +103,14 @@ namespace alpaka
         //! The static device memory view creation trait.
         template<typename TDev, typename TSfinae = void>
         struct CreateStaticDevMemView;
+
+        //! The device memory view creation trait.
+        template<typename TDev, typename TSfinae = void>
+        struct CreateViewPlainPtr;
+
+        //! The sub view creation trait.
+        template<typename TDev, typename TSfinae = void>
+        struct CreateSubView;
     } // namespace traits
 
     //! Gets the native pointer of the memory view.
@@ -330,6 +340,19 @@ namespace alpaka
                 return getPitchBytes<Tidx>(pitch);
             }
         };
+
+        //! Calculate the pitches purely from the extents.
+        template<typename TElem, typename TDim, typename TIdx>
+        ALPAKA_FN_HOST static auto calculatePitchesFromExtents(Vec<TDim, TIdx> const& extent)
+        {
+            Vec<TDim, TIdx> pitchBytes(Vec<TDim, TIdx>::all(0));
+            pitchBytes[TDim::value - 1u] = extent[TDim::value - 1u] * static_cast<TIdx>(sizeof(TElem));
+            for(TIdx i = TDim::value - 1u; i > static_cast<TIdx>(0u); --i)
+            {
+                pitchBytes[i - 1] = extent[i - 1] * pitchBytes[i];
+            }
+            return pitchBytes;
+        }
     } // namespace detail
     //! \return The pitch vector.
     template<typename TPitch>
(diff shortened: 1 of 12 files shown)


include/alpaka/mem/view/Traits.hpp
```diff
@@ -353,4 +378,100 @@ namespace alpaka
     {
         return traits::CreateStaticDevMemView<TDev>::createStaticDevMemView(pMem, dev, extent);
     }
+
+    //! Creates a view to a pointer device
```
sbastrakov: Did you mean `device pointer`?

include/alpaka/mem/view/Traits.hpp
```diff
@@ -353,4 +378,100 @@ namespace alpaka
...
+    //!
+    //! \param pMem Pointer to memory.
+    //! \param dev Device from where pMem can be accessed.
+    //! \param extent Number of elements represented by the pMem.
+    //!               Using a multi dimensional extent will result in a multi dimension view to the memory represented
+    //!               by pMem.
+    //! \return A view to device memory.
+    template<typename TElem, typename TDev, typename TExtent>
+    auto createView(TElem* pMem, TDev const& dev, TExtent const& extent)
```
sbastrakov: This is subjective.
device being the second parameter is not logical imo. I would put it first, as the pointer and extent/pitch are interconnected

include/alpaka/mem/view/Traits.hpp
```diff
@@ -353,4 +378,100 @@ namespace alpaka
...
+
+    //! Creates a view to a device std::vector
+    //!
+    //! \param vec Pointer to memory.
+    //! \param dev Device from where vec can be accessed.
+    //! \return A view to device memory.
+    template<typename TElem, typename TAllocator, typename TDev>
+    auto createView(std::vector<TElem, TAllocator>& vec, TDev const& dev)
+    {
+        return createView(vec.data(), dev, extent::getExtent(vec));
```
sbastrakov: Not for this PR, since it would require C++17. But theoretically [std::data](https://en.cppreference.com/w/cpp/iterator/data) would be nice to unify all contiguous containers. (Of course, we may have our version of it for C++14, but not worth it imo). Maybe add a TODO when switching to 17?

include/alpaka/mem/view/Traits.hpp
```diff
@@ -353,4 +378,100 @@ namespace alpaka
...
+    //! \return A view to device memory.
+    template<typename TElem, typename TDev, typename TExtent, typename TPitch>
+    auto createView(TElem* pMem, TDev const& dev, TExtent const& extent, TPitch const& pitch)
+    {
+        return traits::CreateViewPlainPtr<TDev>::createViewPlainPtr(pMem, dev, extent, pitch);
+    }
+
+    //! Creates a view to a device std::vector
+    //!
+    //! \param vec Pointer to memory.
```
sbastrakov: `vec` is not a pointer, same in the other version below.

I think we should also clarify somewhere in this doxygen comment that not the vector object, but the memory it holds must be accessible on device

include/alpaka/mem/view/Traits.hpp
```diff
@@ -353,4 +378,100 @@ namespace alpaka
...
+    //! \return A view to device memory.
+    template<typename TElem, typename TAllocator, typename TDev, typename TExtent>
+    auto createView(std::vector<TElem, TAllocator>& vec, TDev const& dev, TExtent const& extent)
+    {
+        return createView(vec.data(), dev, extent);
+    }
+
+    //! Creates a view to a device std::array
+    //!
+    //! \param array Pointer to memory.
```
sbastrakov: `array` is also not a pointer, same in the other version below

sbastrakov (changes requested):
Very nice, just some small stuff below

bernhardmgruber (dismissed):
I gave it a high level review and this PR LGTM. I also appreciate the new functionality.

Please still address @sbastrakov's comments.

include/alpaka/mem/view/Traits.hpp
psychocoderHPC: I know I keep the order equal to `createStaticDevMemView`

include/alpaka/mem/view/Traits.hpp
psychocoderHPC: > vec is not a pointer,

copy past bug^^

include/alpaka/mem/view/Traits.hpp
psychocoderHPC: I will change the order but it will then have a different order than the direct creation of `ViewPlainPtrView`, never the less I prevare to have the device as the first param too.

include/alpaka/mem/view/Traits.hpp
```diff
@@ -353,4 +376,104 @@ namespace alpaka
...
+
+    //! Creates a view to a device std::vector
+    //!
+    //! \param dev Device from where vec can be accessed.
+    //! \param vec std::vector container. The data hold by the container the must be accessible from the given device.
+    //! \return A view to device memory.
+    template<typename TDev, typename TElem, typename TAllocator>
+    auto createView(TDev const& dev, std::vector<TElem, TAllocator>& vec)
+    {
+        //! \todo With C++17 uste std::data() and add support for all contagious container
```
sbastrakov: Lol, `contagious` containers with zombies.
Should be `contiguous`.

sbastrakov: Also, typo in `use`

sbastrakov (changes requested):
Thanks for fixing, just two more typos introduced in your last changes

include/alpaka/mem/view/Traits.hpp
psychocoderHPC: > Lol, contagious containers with zombies.

Maybe I read to much Covid news ^^

I fixed the typo

sbastrakov (dismissed):

sbastrakov (approved):

PR #1400, opened 2021-08-13
HIP: fix peer to peer memcopy

psychocoderHPC:
Fix HIP p2p memcopy between devices within a single host system.
This PR additionally refactored the CUDA p2p implementation. 
For a long time CUDA is not requiring explicit p2p copy calls, `cudaMemcpy*` is detecting if a copy operation must be performed between different devices and will to all required configurations.

note: To test p2p memcpy #1399 is required too. Never the less this PR can be merged before #1399.

### DIFF
--- a/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp
+++ b/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp
@@ -45,49 +45,6 @@ namespace alpaka
 {
     namespace detail
     {
-        //! Not being able to enable peer access does not prevent such device to device memory copies.
-        //! However, those copies may be slower because the memory is copied via the CPU.
-        inline auto enablePeerAccessIfPossible(const int& devSrc, const int& devDst) -> void
-        {
-            ALPAKA_ASSERT(devSrc != devDst);
-
-#    if BOOST_COMP_CLANG
-#        pragma clang diagnostic push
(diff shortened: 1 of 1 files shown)


include/alpaka/mem/buf/uniformCudaHip/Copy.hpp
```diff
@@ -364,13 +250,14 @@ namespace alpaka
...
-            Idx m_srcWidth; // required for 3D peer copy
+            Idx m_dstWidth;
+            Idx m_srcWidth;
 
             Idx m_extentHeight;
-            Idx m_dstHeight; // required for 3D peer copy
-            Idx m_srcHeight; // required for 3D peer copy
-
+#    if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
+            Idx m_dstHeight;
```
j-stephan, line 258: Is there a reason we're not putting the width here, too? Also below for the 3D case.

psychocoderHPC: width is used somewhere in a call. All attributes which are guarded by the define are only active for debugging but are not required for any api call.

psychocoderHPC:
@j-stephan Can we merge this PR?

sbastrakov (approved):

j-stephan:
Yes, sorry for the delay. I somehow missed your response to my review.

PR #1407, opened 2021-09-06
Run CUDA 9.2 tests on HZDR CI

psychocoderHPC:
Copy github actions CUDA 9.2 jobs over to HZDR gitlab CI.
Reuse install script we already use for the Github actions CI.

HZDR CI:
  - CUDA: use nvidia cuda container (do not install CUDA)
  - use Ubuntu 18.04 for CUDA 9.2 (Ubuntu 20.04 container does not exists)
  - do not test `ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE` together with CUDA
(reduce compile time)
    - OpenMP + CUDA is still tested with `ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE`

Sometimes it is required to make a step back to go forward.
Therefore I used nvidia container sinstead of the alpaka containers provided by @SimeonEhrig.
The alpaka containers provide `avx512` compile time errors we workaround only for GCC 5: https://github.com/alpaka-group/alpaka/blob/9ac94b8faf85a9bceb3c260ddb3a32599644d99f/script/before_install.sh#L85-L103
It looks like the alpaka clang-cuda versions are built against GCC5 headers.
Switching back to the alpaka headers should be implemented in a follow-up PR. The advantage of first using the current existing script and then switching to the specialized containers is that we can reuse all our scripts and need only small changes to avoid installing already shipped software or compilers.

### DIFF
--- a/script/before_install.sh
+++ b/script/before_install.sh
@@ -32,6 +32,7 @@ fi
 
 #-------------------------------------------------------------------------------
 # Boost.
+echo $ALPAKA_CI_BOOST_BRANCH
 ALPAKA_CI_BOOST_BRANCH_MAJOR=${ALPAKA_CI_BOOST_BRANCH:6:1}
 echo ALPAKA_CI_BOOST_BRANCH_MAJOR: "${ALPAKA_CI_BOOST_BRANCH_MAJOR}"
 ALPAKA_CI_BOOST_BRANCH_MINOR=${ALPAKA_CI_BOOST_BRANCH:8:2}
@@ -40,7 +41,7 @@ echo ALPAKA_CI_BOOST_BRANCH_MINOR: "${ALPAKA_CI_BOOST_BRANCH_MINOR}"
 #-------------------------------------------------------------------------------
 # CUDA
 export ALPAKA_CI_INSTALL_CUDA="OFF"
-if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" ]
+if [[ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" && -z "${GITLAB_CI+x}" ]]
 then
     export ALPAKA_CI_INSTALL_CUDA="ON"
 fi
@@ -85,9 +86,9 @@ fi
 # GCC-5.5 has broken avx512vlintrin.h in Release mode with NVCC 9.X
 #   https://gcc.gnu.org/bugzilla/show_bug.cgi?id=76731
 #   https://github.com/tensorflow/tensorflow/issues/10220
-if [ "${ALPAKA_CI_INSTALL_CUDA}" == "ON" ]
+if [ "${ALPAKA_CI_INSTALL_CUDA}" == "ON"  ]
 then
-    if [ "${CXX}" == "g++" ]
+    if [[ "${CXX}" == "g++"* ]]
     then
         if (( "${ALPAKA_CI_GCC_VER_MAJOR}" == 5 ))
         then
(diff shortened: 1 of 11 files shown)


psychocoderHPC:
@alpaka-group/alpaka-maintainers ready for a review and merge

script/before_install.sh
```diff
@@ -40,7 +41,7 @@ echo ALPAKA_CI_BOOST_BRANCH_MINOR: "${ALPAKA_CI_BOOST_BRANCH_MINOR}"
 #-------------------------------------------------------------------------------
 # CUDA
 export ALPAKA_CI_INSTALL_CUDA="OFF"
-if [ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" ]
+if [[ "${ALPAKA_ACC_GPU_CUDA_ENABLE}" == "ON" && -z "${GITLAB_CI+x}" ]]
```
SimeonEhrig, line 44: Why do you use the `nvidia/cuda:{version}-devel-ubuntu{version}` container and not the `nvidia/cuda:{version}-base-ubuntu{version}` container and install cuda afterwards, like you suggested for the original alpaka container?

script/gitlabci/job_base.yml
```diff
@@ -1,40 +1,48 @@
...
-    - mkdir -p ${ALPAKA_CI_CMAKE_DIR}
-    - source ./script/install_cmake.sh
+    - source ./script/before_install.sh
+    - source ./script/install.sh
     - source ./script/run.sh
+  interruptible: true
 
-.base_gcc:
+.base_cuda_gcc:
+  image: nvidia/cuda:${ALPAKA_CI_CUDA_VERSION}-devel-ubuntu${ALPAKA_CI_UBUNTU_VER}
```
SimeonEhrig, line 14: What is wrong with the alpaka container? Can you please open an issue, that it can be fixed: https://gitlab.com/hzdr/crp/alpaka-group-container

script/before_install.sh
psychocoderHPC: discussed offline: using base container and `apt` is more expensive. Container are cached but apt is not cached in our CI.
Never the less the next step will be switching to alpaka container instead of NVIDIA container.

script/gitlabci/job_base.yml
psychocoderHPC: offline discussed: https://github.com/alpaka-group/alpaka/pull/1407#issue-727718589 clang is using gcc 5 header and therefore compiling with clang cuda is not possible. 
Maybe we need to build a new alpaka container version. Integration of alpaka container will be the next step.
This PR is for using the full existing alpaka toolchain.

j-stephan:
> It looks like the alpaka clang-cuda versions are built against GCC5 headers.

You can specify alternative headers clang should use:

```
clang++ --gcc-toolchain=/usr/local/gcc/6.4.0
```

SimeonEhrig:
> > It looks like the alpaka clang-cuda versions are built against GCC5 headers.
> 
> You can specify alternative headers clang should use:
> 
> ```
> clang++ --gcc-toolchain=/usr/local/gcc/6.4.0
> ```

Does you know, if `g++` also use the `libstdc++` version of the default host compiler, means for example if you compile with `g++-8` in Ubuntu 16.04, it selects the `libstdc++` of the `g++-5`? If yes, we should implement an mechanism to manual specific it.

SimeonEhrig (approved):

j-stephan:
> Does you know, if g++ also use the libstdc++ version of the default host compiler, means for example if you compile with g++-8 in Ubuntu 16.04, it selects the libstdc++ of the g++-5? If yes, we should implement an mechanism to manual specific it.

GCC should pull in the libstdc++ that is part of its installation. So g++-5 will link to libstdc++-5 and g++-8 to libstdc++-8.

bernhardmgruber (approved):

PR #1416, opened 2021-10-08
fix warning: annotated defaulted constructors

psychocoderHPC:
```
alpaka/workdiv/WorkDivMembers.hpp(57): warning: __device__ annotation is ignored on a function("operator=") that is explicitly defaulted on its first declaration
```

Shown with CUDA 11.2 and @j-stephan's example https://github.com/infn-esc/esc21/blob/master/hands-on/alpaka_exercises/computePi/computePi.cu

### DIFF
--- a/example/reduce/src/iterator.hpp
+++ b/example/reduce/src/iterator.hpp
@@ -47,7 +47,7 @@ public:
     //! Constructor.
     //!
     //! \param other The other iterator object.
-    ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE Iterator(const Iterator& other) = default;
+    Iterator(const Iterator& other) = default;
 
     //! Compare operator.
     //!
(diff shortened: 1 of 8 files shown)


bernhardmgruber (approved):

sbastrakov (approved):

j-stephan:
By the way, this doesn't occur with a CMake-based build because we silence this specific warning. I guess someone encountered this before? This is the flag we add:

```
-Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored
```

psychocoderHPC:
> By the way, this doesn't occur with a CMake-based build because we silence this specific warning. I guess someone encountered this before? This is the flag we add:
> 
> ```
> -Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored
> ```

Thanks for catching the reason why we missed this warnings. If we do not break HIP IMO we should merge it.

j-stephan (approved):
We should probably remove the suppression of the warning in a follow-up PR. I don't see a reason why we should keep it.

PR #1417, opened 2021-10-12
Update to clang-format 12

bernhardmgruber:


include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
```diff
@@ -127,36 +130,36 @@ namespace alpaka
 // useful because it is meant for cases where multiple iterations are executed by one thread but in our case a 1:1
 // mapping is required. Therefore we use 'omp parallel' with the specified number of threads in a block.
 #    pragma omp parallel num_threads(iBlockThreadCount)
-                {
+                    {
                     // The guard is for gcc internal compiler error, as discussed in #735
 #    if(!BOOST_COMP_GNUC) || (BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(8, 1, 0))
 #        pragma omp single nowait
```
j-stephan, line 136: Shouldn't the `#pragma` be in line with the following block instead of the preprocessor macros? Or is this something introduced in clang-format 13?

j-stephan (approved):
Looks good to me. Only one (non-blocking) question:

include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
bernhardmgruber: Maybe. There was on option `IndentPragma` in the `.clang-format` file, which was introduced during `clang-format-12`, but pulled out again before the release. (See: https://reviews.llvm.org/D92753)
I removed that option from the `.clang-format` file.

j-stephan: Can you add a note to our clang-format repository at CRP? We should keep track of this.

PR #1422, opened 2021-10-13
Add special handling for serial CPU memfence

### LINKED ISSUE #1403: Memory fences should not do anything for the block scope on AccCpuSerial
As pointed out by the CERN-CMS folks our current implementation of memory fences needlessly issues a memory fence for the block scope of `AccCpuSerial`. It would be sufficient to just call an empty function here since fences don't serve a purpose without multiple threads.

For the device scope we should keep the existing behaviour because the user might want to synchronize between kernels.

j-stephan:
Closes #1403.

### DIFF
--- a/include/alpaka/mem/fence/MemFenceCpuSerial.hpp
+++ b/include/alpaka/mem/fence/MemFenceCpuSerial.hpp
@@ -1,0 +1,52 @@
+/* Copyright 2021 Jan Stephan
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/Concepts.hpp>
+#include <alpaka/mem/fence/Traits.hpp>
+
+#include <atomic>
+
+namespace alpaka
+{
+    //! The serial CPU memory fence.
+    class MemFenceCpuSerial : public concepts::Implements<ConceptMemFence, MemFenceCpuSerial>
+    {
+    };
+
+    namespace traits
+    {
+        template<>
+        struct MemFence<MemFenceCpuSerial, memory_scope::Block>
+        {
+            static auto mem_fence(MemFenceCpuSerial const&, memory_scope::Block const&)
+            {
+                /* Nothing to be done on the block level for the serial case. */
+            }
+        };
+
+        template<typename TMemScope>
+        struct MemFence<MemFenceCpuSerial, TMemScope>
+        {
+            static auto mem_fence(MemFenceCpuSerial const&, TMemScope const&)
+            {
+                /* Enable device fences because we may want to synchronize with other (serial) kernels. */
+
+                static std::atomic<int> dummy{42};
+
+                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
+                 * these dummy operations to ensure this.*/
+                auto x = dummy.load(std::memory_order_relaxed);
+                std::atomic_thread_fence(std::memory_order_acq_rel);
+                dummy.store(x, std::memory_order_relaxed);
+            }
+        };
+    } // namespace traits
+} // namespace alpaka
--- a/include/alpaka/acc/AccCpuSerial.hpp
+++ b/include/alpaka/acc/AccCpuSerial.hpp
@@ -22,7 +22,7 @@
 #    include <alpaka/idx/gb/IdxGbRef.hpp>
 #    include <alpaka/intrinsic/IntrinsicCpu.hpp>
 #    include <alpaka/math/MathStdLib.hpp>
-#    include <alpaka/mem/fence/MemFenceCpu.hpp>
+#    include <alpaka/mem/fence/MemFenceCpuSerial.hpp>
 #    include <alpaka/rand/RandStdLib.hpp>
 #    include <alpaka/time/TimeStdLib.hpp>
 #    include <alpaka/warp/WarpSingleThread.hpp>
@@ -69,7 +69,7 @@ namespace alpaka
         public BlockSharedMemStMember<>,
         public BlockSyncNoOp,
         public IntrinsicCpu,
-        public MemFenceCpu,
+        public MemFenceCpuSerial,
         public rand::RandStdLib,
         public TimeStdLib,
         public warp::WarpSingleThread,
@@ -99,7 +99,7 @@ namespace alpaka
             , BlockSharedMemDynMember<>(blockSharedMemDynSizeBytes)
             , BlockSharedMemStMember<>(staticMemBegin(), staticMemCapacity())
             , BlockSyncNoOp()
-            , MemFenceCpu()
+            , MemFenceCpuSerial()
             , rand::RandStdLib()
             , TimeStdLib()
             , m_gridBlockIdx(Vec<TDim, TIdx>::zeros())
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -124,6 +124,7 @@
 #include <alpaka/mem/buf/BufUniformCudaHipRt.hpp>
 #include <alpaka/mem/buf/Traits.hpp>
 #include <alpaka/mem/fence/MemFenceCpu.hpp>
+#include <alpaka/mem/fence/MemFenceCpuSerial.hpp>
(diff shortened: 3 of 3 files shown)


include/alpaka/mem/fence/MemFenceCpuSerial.hpp
```diff
@@ -0,0 +1,52 @@
...
+        };
+
+        template<typename TMemScope>
+        struct MemFence<MemFenceCpuSerial, TMemScope>
+        {
+            static auto mem_fence(MemFenceCpuSerial const&, TMemScope const&)
+            {
+                /* Enable device fences because we may want to synchronize with other (serial) kernels. */
+
+                static std::atomic<int> dummy{42};
```
bernhardmgruber, line 42: A local `static` variable with an initializer will implicitely introduce an atomic test into this function, which might be additional overhead. That variable should be moved outside the function, but I guess that would require C++17 to stay header-only.

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
```diff
@@ -0,0 +1,52 @@
...
+                /* Nothing to be done on the block level for the serial case. */
+            }
+        };
+
+        template<typename TMemScope>
+        struct MemFence<MemFenceCpuSerial, TMemScope>
+        {
+            static auto mem_fence(MemFenceCpuSerial const&, TMemScope const&)
+            {
+                /* Enable device fences because we may want to synchronize with other (serial) kernels. */
```
bernhardmgruber, line 40: Are you sure? Do memory fences in CUDA kernels synchronize with other CUDA kernels? The way I interpret the CUDA programming guide, there are no guarantees between kernels. So I would stick to the same behavior on CPU kernels and have this function empty.

bernhardmgruber: I interpret `memory_scope::Device` as kernel wide, as the CUDA implementation uses `__threadfence` and not `__threadfence_system`.

bernhardmgruber (changes requested):

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
j-stephan: CUDA's `__threadfence()` is observed by all threads in the device, not a single kernel. See the corresponding section here: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#memory-fence-functions

Note that it explicitly talks about `device`, not `kernel`.

Which makes sense - how would you guarantee observability across all kernel threads without guaranteeing the same on the whole device? You don't know on which SM your blocks are going to execute, you must make it observable on the entire device.

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
j-stephan: Do you suggest a C++14 conforming change or is this something we need to keep in mind?

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
bernhardmgruber: I think I got a bit confused by this SO answer: https://stackoverflow.com/a/6546246/2406044. Re-reading the programming guide actually looks like `__threadfence` will synchronize between kernels on the same device. So I assume the fence on the CPU is needed to have the same behavior.

j-stephan: > So I assume the fence on the CPU is needed to have the same behavior.

Yup. This behaviour was also suggested by CMS, so it seems intuitive from a user POV.

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
bernhardmgruber: I suggested a simplification and performance improvement with C++17.

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+                static std::atomic<int> dummy{42};
+
+                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
+                 * these dummy operations to ensure this.*/
+                auto x = dummy.load(std::memory_order_relaxed);
+                std::atomic_thread_fence(std::memory_order_acq_rel);
+                dummy.store(x, std::memory_order_relaxed);
```
bernhardmgruber, line 48: Have you actually checked the correctness of your implementation? This code compiles down to an empty function with clang: https://godbolt.org/z/a6nbes7xz

j-stephan: Yes. This is because `std::memory_order_acq_rel` does not correspond to an actual hardware instruction on x86. If you change the architecture to something else you will see a non-empty function. Example for RISC-V:

```
mem_fence():                          # @mem_fence()
        fence.tso       
        ret
```

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
j-stephan: Okay. Since we are currently on C++14 I guess this is resolved for now?

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
bernhardmgruber: Fair enough, but the body is also empty for ARM64: https://godbolt.org/z/dW7En6qq3 Is that correct too? I wonder if clang is smart and realizes that the atomic is not visible outside the function and the code inside therefore has no visible side effects.

j-stephan: It is not empty. `dmb ish` is a memory barrier. https://developer.arm.com/documentation/100941/0100/Barriers

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
bernhardmgruber: Maybe. I don't see the lock for the initialization of the static variable being created in godbolt. But I guess it's fine because the integer will be put preinitialized in the data segment. I still think there is something wrong with this code, but let's discuss this in the thread below.

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
bernhardmgruber: Oh. Then I must have misread the tooltop in godbolt. Sorry!

bernhardmgruber:
I am still wondering whether the implementation is correct, so I created this little test based on the alpaka documentation:
```c++
void mem_fence() {
    static std::atomic<int> dummy{42};
    auto x = dummy.load(std::memory_order_relaxed);
    std::atomic_thread_fence(std::memory_order_acq_rel);
    dummy.store(x, std::memory_order_relaxed);
}

unsigned vars[2]{2, 1};

void a() {
    vars[0]++;
    mem_fence();
    vars[1]++;
}

void b() {
    auto const a = vars[0];
    mem_fence();
    auto const b = vars[1];
    if(a == b) report_fail(a, b);
}

constexpr auto n = 1000000;
int main() {
    std::thread ta([]{ for (auto i = 0; i < n; i++) a(); });
    std::thread tb([]{ for (auto i = 0; i < n; i++) b(); });
    ta.join();
    tb.join();
}
```
Godbolt: https://godbolt.org/z/s481MEojd

I would argue that `report_fail` should never be called, as `a` should always be ahead of `b`. However, if you press recompile on Godbolt a couple of times, you can see that `report_fail` is called many times. Maybe my example is be wrong. But I still think there is an issue.

j-stephan:
Remember that we are talking about memory fences, not synchronization. The memory fence in your example guarantees that `vars[0]++` happens before `var[1]++` and that `a` is loaded before `b`. There is no relationship between `vars[0]++` and the loading of `a`.

For an actual test case you need to switch the loading of `a`and `b` as follows:

```c++
void b()
{
    auto const a = vars[1];
    mem_fence();
    auto const b = vars[0];
    if(a == b) report_fail(a, b);
}
```

Thanks to the memory fence it is always ensured that `vars[0]++` happens before `vars[1]++` and also before `b` is loaded. Thus `a` and `b` can never be the same.

j-stephan:
Ready to merge.

psychocoderHPC (approved):

bernhardmgruber:
@j-stephan I think I confused something there indeed. Thank you for the explanation!

bernhardmgruber (approved):

PR #1423, opened 2021-10-15
fix CMake native CUDA compiler support

psychocoderHPC:
With the switch to the native CMake CUDA support in #1146 alpaka was not setting anymore the host compiler for CUDA.
We did this in older releases: https://github.com/alpaka-group/alpaka/blob/5dca322c06ab59abc8a69edc363859898b1b7713/cmake/alpakaCommon.cmake#L471

CMake CUDA compiler is always falling back to the system C++ compiler instead of using the CMake CXX compiler.

The bug exists already in alpaka 0.7.0 and therefore should be ported back to 0.7.1.

I found this bug during the update of PIConGPU to alpaka 0.7.0. nvcc was falling back to GCC 5 and therefore I got boot compile issues.

- Additionally, this PR fixes a bug introduced with #1386 where a macro guard is based on the CUDA host compiler instead of the device compiler. This bug became visible with the fix from the first commit.
- - remove CUDA <= 11.3 + clang 11 CI tests (CUDA is not supporting clang 11 before CUDA 11.4)

CC-ing: @SimeonEhrig I know you fought with a similar bug yesterday


SimeonEhrig:
@psychocoderHPC I have already the solution and I wanted to open this issue on Monday ;-)
You have to set the cmake `CMAKE_CUDA_COMPILER` to `nvcc`, otherwise the CMake variable is `CMAKE_CUDA_HOST_COMPILER` ignored.

See example here: https://github.com/alpaka-group/vikunja/blob/66bd45a05771855c06cc812eae7bcd7ea079cf2c/script/run_test.sh#L41
and doc here (in the note area): https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_HOST_COMPILER.html

psychocoderHPC:
> CMAKE_CUDA_HOST_COMPILER

Setting `cmake -DCMAKE_CUDA_HOST_COMPILER=g++-9` is not working because `check_language(CUDA)` resets what ever is in `CMAKE_CUDA_HOST_COMPILER`.

SimeonEhrig:
> > CMAKE_CUDA_HOST_COMPILER
> 
> Setting `cmake -DCMAKE_CUDA_HOST_COMPILER=g++-9` is not working because `check_language(CUDA)` resets what ever is in `CMAKE_CUDA_HOST_COMPILER`.

What's happen, if you execute `check_language(CUDA)` in your project CMake, than set `CMAKE_CUDA_HOST_COMPILER` and include `alpaka` after this?

```cmake
check_language(CUDA)
set(CMAKE_CUDA_HOST_COMPILER g++9)
find_package(alpaka)
```

psychocoderHPC:
> > > CMAKE_CUDA_HOST_COMPILER
> > 
> > 
> > Setting `cmake -DCMAKE_CUDA_HOST_COMPILER=g++-9` is not working because `check_language(CUDA)` resets what ever is in `CMAKE_CUDA_HOST_COMPILER`.
> 
> What's happen, if you execute `check_language(CUDA)` in your project CMake, than set `CMAKE_CUDA_HOST_COMPILER` and include `alpaka` after this?
> 
> ```cmake
> check_language(CUDA)
> set(CMAKE_CUDA_HOST_COMPILER g++9)
> find_package(alpaka)
> ```

This works too if you also activate CUDA after you set the host compiler and is my workaround in PIConGPU until alpaka is handling the workaround: https://github.com/ComputationalRadiationPhysics/picongpu/pull/3868/commits/a77233a4ccaf2e85fc714c4f88fa3a1cea2dfb9d
Never the less I like to avoid as best as possible to handle backend specific stuff in the user application.

SimeonEhrig:
> Never the less I like to avoid as best as possible to handle backend specific stuff in the user application.

I agree with you. This was the reason, why I wanted to the issue at Monday and not yesterday evening, because I wanted to discuss how we handle this. We should also check, if the CUDA CMake support has more "special" behavior like this.

psychocoderHPC:
> > Never the less I like to avoid as best as possible to handle backend specific stuff in the user application.
> 
> I agree with you. This was the reason, why I wanted to the issue at Monday and not yesterday evening, because I wanted to discuss how we handle this. We should also check, if the CUDA CMake support has more "special" behavior like this.

I directly added debug output in cmake and inspected the cmake implementation.
In the file `cmake-3.18/Modules/CMakeDetermineCUDACompiler.cmake` they do

```
    set(extra_compiler_variables)
    if(${lang} STREQUAL CUDA)
      set(extra_compiler_variables "set(CMAKE_CUDA_HOST_COMPILER \\\"\${CMAKE_CUDA_HOST_COMPILER}\\\")")
    endif()
```

This is written into `result.cmake`, they do not check if the user already sets `CMAKE_CUDA_HOST_COMPILER` and there overwrite it hard.

j-stephan:
I found this explanation from a CMake developer:

> The purpose of `check_language` is to detect the cuda compiler and host compiler from the environment and report them back.  In a use case where you're setting `CMAKE_CUDA_HOST_COMPILER` explicitly then you should set `CMAKE_CUDA_COMPILER` explicitly too.  It's the case where `CMAKE_CUDA_COMPILER` is not set that this check runs, and in that case we should let `check_language` detect the host compiler that goes with the detected cuda compiler.

https://gitlab.kitware.com/cmake/cmake/-/issues/20676

j-stephan (changes requested):
I think the whole section should be rewritten into something like this:

```cmake
# Save the user-defined host compiler (if any)
set(_ALPAKA_CUDA_HOST_COMPILER ${CMAKE_CUDA_HOST_COMPILER})

check_language(CUDA)
if(CMAKE_CUDA_COMPILER)
    if(${_ALPAKA_CUDA_HOST_COMPILER})
        set(CMAKE_CUDA_HOST_COMPILER ${_ALPAKA_CUDA_HOST_COMPILER})
    elseif(NOT $ENV{CUDAHOSTCXX})
        set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
    endif()

    enable_language(CUDA)
    find_package(CUDAToolkit REQUIRED)
    # etc.
endif()
```

This also needs an extra guard for clang since the host compiler is only meaningful when `CMAKE_CUDA_COMPILER` is `nvcc`.

psychocoderHPC:
> I think the whole section should be rewritten into something like this:
> 
> ```cmake
> # Save the user-defined host compiler (if any)
> set(_ALPAKA_CUDA_HOST_COMPILER ${CMAKE_CUDA_HOST_COMPILER})
> 
> check_language(CUDA)
> if(CMAKE_CUDA_COMPILER)
>     if(${_ALPAKA_CUDA_HOST_COMPILER})
>         set(CMAKE_CUDA_HOST_COMPILER ${_ALPAKA_CUDA_HOST_COMPILER})
>     elseif(NOT $ENV{CUDAHOSTCXX})
>         set(CMAKE_CUDA_HOST_COMPILER ${CMAKE_CXX_COMPILER})
>     endif()
> 
>     enable_language(CUDA)
>     find_package(CUDAToolkit REQUIRED)
>     # etc.
> endif()
> ```
> 
> This also needs an extra guard for clang since the host compiler is only meaningful when `CMAKE_CUDA_COMPILER` is `nvcc`.

~~It is not only about keeping the state of `CMAKE_CUDA_HOST_COMPILER` the user provides, CMake native CUDA support is searching for the CUDA compiler and is not using by default the CXX as host cuda compiler. In old alpaka before 0.7.0 we always set the host compiler in alpaka to avoid that CUDA is using a compiler which e.g. not supports C++14.
Setting the host compiler was removed in #1146  and therefore CUDA is depending on your environment silently falling back to a host compiler which is not supporting C++14. We do not see this behavior in our CI because we use containers with only one host compiler and this host compiler is relatively new.~~

[update] I was miss interpreting your suggested patch. The patch looks good!

psychocoderHPC:
@j-stephan I implemented your suggestion, including a small fix on how to check the environment variable. It is required to ship the first commit with an additional fix which became now visible. In #1386 I used the wrong variable to check which CUDA compiler is used.

include/alpaka/core/Common.hpp
```diff
@@ -90,7 +90,7 @@
 //!     alpaka::memcpy(queue, viewGlobalMemUninitialized, bufHost, extent);
 //! }
 //! \endcode
-#if((BOOST_LANG_CUDA && BOOST_COMP_CLANG) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX) || BOOST_LANG_HIP)
+#if((BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX) || BOOST_LANG_HIP)
```
j-stephan: Did Boost rename this macro?

cmake/alpakaCommon.cmake
```diff
@@ -324,6 +324,21 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
...
     if(CMAKE_CUDA_COMPILER)
+        # Save the user-defined host compiler (if any)
+        set(_ALPAKA_CUDA_HOST_COMPILER ${CMAKE_CUDA_HOST_COMPILER})
+
+        check_language(CUDA)
+        if(CMAKE_CUDA_COMPILER)
+            # Use user selected CMake CXX compiler or CMAKE_CUDA_HOST_COMPILER as cuda host compiler to avoid fallback to the default system CXX host compiler.
+            # CMAKE_CUDA_HOST_COMPILER is reset by check_language(CUDA) therefore definition passed by the user via -DCMAKE_CUDA_HOST_COMPILER are
+            # ignored by CMake (looks like a CMake bug).
+            if(${_ALPAKA_CUDA_HOST_COMPILER})
```
j-stephan: This section still needs to be guarded against clang-cuda.

j-stephan (changes requested):

include/alpaka/core/Common.hpp
psychocoderHPC: No this name is provided by alpaka. The check before was saying:
true - if CUDA + HOST clang is used.
Now it is
true - if CUDA + DEVICE side clang is used

If `BOOST_COMP_CLANG` is used it means the CXX compiler is clang, this variable is shipped by boost.
`BOOST_COMP_CLANG_CUDA` is created by alpaka und means that the device compiler is clang.

cmake/alpakaCommon.cmake
psychocoderHPC: It is not required to remove this for clang. I realized I moved the fix to the wrong line. I need to move all up.

psychocoderHPC: Decided to guard it, as suggested.

j-stephan (dismissed):

psychocoderHPC:
This PR is triggering many issues we head because the system compiler was used as the host compiler.
We added CI tests for CUDA 11.1 - 11.3 together with clang 11. This is not supported by CUDA, therefore the third PR will update the test matrix in README and remove the unsupported tests.

See this table to find all supported CUDA nvcc + Clang version: https://gist.github.com/ax3l/9489132

j-stephan:
Is this ready to merge?

j-stephan (approved):

bernhardmgruber (approved):

PR #1431, opened 2021-10-25
Atomic tests enabled and fix for float/double

### LINKED ISSUE #1402: Inconsistent Atomic support handling/testing
This is both an issue and a question.

## Question: How does the `atomic` test pass for the CUDA/HIP backend?

The atomic test tests 8-bit atomic operations:
https://github.com/alpaka-group/alpaka/blob/a972d5eb00c36c4806f8162dc217eae9beafa2c3/test/unit/atomic/src/AtomicTest.cpp#L785-L786
The CUDA backend does not implement these because CUDA does not support those (I could also not find 16-bit, even though CUDA supports those, but this would be a different issue). I am confused about this because, the atomic test does not exclude CUDA from them 8-bit case and CUDA provides a "fallback" specialization which would have a failing `static_assert` unless there is something special about the `THierarchy` argument:
https://github.com/alpaka-group/alpaka/blob/a972d5eb00c36c4806f8162dc217eae9beafa2c3/include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp#L777-L790

## Issue: inconsistent testing

While 8-bit operations are not supported by all backends (i.e. CUDA) but are still tested, the floating point cases have been deactivated in the test, because they are apparently not supported everywhere:

https://github.com/alpaka-group/alpaka/blob/a972d5eb00c36c4806f8162dc217eae9beafa2c3/test/unit/atomic/src/AtomicTest.cpp#L798-L800

## My Problem

In the OpenMP5 and OpenACC backends, 8 and 16 bit atomics are not supported when targeting GPUs, due to missing support in the underlying hardware (see CUDA). How can this be handled in agreement with alpaka's `atomic` test?

tonydp03:
This Pull request enables back all the atomic tests. The 8- and 16-bit tests are skipped for the CUDA/HIP backend; the float and double tests for atomics are re-enabled but not run for bitwise operations (and, or, xor).

### DIFF
--- a/test/unit/atomic/src/AtomicTest.cpp
+++ b/test/unit/atomic/src/AtomicTest.cpp
@@ -9,6 +9,7 @@
 
 #include <alpaka/atomic/Traits.hpp>
 #include <alpaka/core/Unused.hpp>
+#include <alpaka/math/FloatEqualExact.hpp>
 #include <alpaka/test/KernelExecutionFixture.hpp>
 #include <alpaka/test/acc/TestAccs.hpp>
 
@@ -16,6 +17,24 @@
 
 #include <climits>
 #include <type_traits>
+
+template<typename T1, typename T2>
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals(T1 a, T2 b)
+{
+    return a == b;
+}
+
+template<>
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals<float, float>(float a, float b)
+{
+    return alpaka::math::floatEqualExactNoWarning(a, b);
+}
+
+template<>
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals<double, double>(double a, double b)
+{
+    return alpaka::math::floatEqualExactNoWarning(a, b);
+}
 
 ALPAKA_NO_HOST_ACC_WARNING
 template<typename TAcc, typename T>
@@ -27,14 +46,14 @@ ALPAKA_FN_ACC auto testAtomicAdd(TAcc const& acc, bool* success, T operandOrig) 
     {
         operand = operandOrig;
         T const ret = alpaka::atomicOp<alpaka::AtomicAdd>(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
     {
         operand = operandOrig;
         T const ret = alpaka::atomicAdd(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
 }
 
@@ -48,14 +67,14 @@ ALPAKA_FN_ACC auto testAtomicSub(TAcc const& acc, bool* success, T operandOrig) 
     {
         operand = operandOrig;
         T const ret = alpaka::atomicOp<alpaka::AtomicSub>(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
     {
         operand = operandOrig;
         T const ret = alpaka::atomicSub(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
 }
 
@@ -69,14 +88,14 @@ ALPAKA_FN_ACC auto testAtomicMin(TAcc const& acc, bool* success, T operandOrig) 
     {
         operand = operandOrig;
         T const ret = alpaka::atomicOp<alpaka::AtomicMin>(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
     {
         operand = operandOrig;
         T const ret = alpaka::atomicMin(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
 }
 
@@ -90,14 +109,14 @@ ALPAKA_FN_ACC auto testAtomicMax(TAcc const& acc, bool* success, T operandOrig) 
     {
         operand = operandOrig;
         T const ret = alpaka::atomicOp<alpaka::AtomicMax>(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
     {
         operand = operandOrig;
         T const ret = alpaka::atomicMax(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
 }
 
@@ -111,14 +130,14 @@ ALPAKA_FN_ACC auto testAtomicExch(TAcc const& acc, bool* success, T operandOrig)
     {
         operand = operandOrig;
         T const ret = alpaka::atomicOp<alpaka::AtomicExch>(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
     {
         operand = operandOrig;
         T const ret = alpaka::atomicExch(acc, &operand, value);
-        ALPAKA_CHECK(*success, operandOrig == ret);
-        ALPAKA_CHECK(*success, operand == reference);
+        ALPAKA_CHECK(*success, equals(operandOrig, ret));
+        ALPAKA_CHECK(*success, equals(operand, reference));
     }
 }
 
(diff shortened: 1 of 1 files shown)


test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -317,6 +361,8 @@ class AtomicTestKernel<alpaka::AccGpuUniformCudaHipRt<TDim, TIdx>, T>
 };
 #endif
 
+
+
```
bernhardmgruber: Please don't add any unnecessary whitespace.

bernhardmgruber (changes requested):
I think you did not run clang-format on your PR. Please do so. I have seen you use VSCode. It should have a formatting hotkey as well, see e.g.: https://code.visualstudio.com/docs/cpp/cpp-ide.
Post-commit, you can run `git clang-format HEAD~1` to reformat the changes from the last commit, then amend and force push.

bernhardmgruber:
Regarding the failing tests. You might need to disable the floating point comparison warnings around the offending lines:
```c++
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wfloat-equal"
    ...
#pragma GCC diagnostic pop
```

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -582,7 +654,10 @@ class AtomicTestKernel<
         alpaka::ignore_unused(operandOrig);
 
         // All other types are not supported by CUDA atomic operations.
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wfloat-equal"
         ALPAKA_CHECK(*success, true);
+#    pragma GCC diagnostic pop
```
bernhardmgruber: Are you sure this is required here? There is no float comparison involved here.

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -803,7 +878,10 @@ class AtomicTestKernel<
         alpaka::ignore_unused(operandOrig);
 
         // All other types are not supported by HIP atomic operations.
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wfloat-equal"
         ALPAKA_CHECK(*success, true);
+#    pragma GCC diagnostic pop
```
bernhardmgruber: See above.

bernhardmgruber (changes requested):

test/unit/atomic/src/AtomicTest.cpp
tonydp03: I put it everywhere around ALPAKA_CHECK to see if it worked - I was wondering if those pragmas were working for every compiler, but as far as I see, they don't. So should we put other pragmas around those lines, for all the compilers? Or maybe it's better to modify the library `Op.hpp` such that whenever there's a comparison warnings should be ignored if it's between floating point numbers?

tonydp03: Like it is done in

https://github.com/alpaka-group/alpaka/blob/878dc184d92bba0bbbf6379e5b9886142aed7e9a/include/alpaka/atomic/Op.hpp#L190-L197

At the end, in the Op.hpp library the comparison happens only in AtomicCas and atomicDec if I saw correctly. A possibility would be to add pragmas around those lines only, but for every compiler?

bernhardmgruber: Well, that's a good first test locally. But please apply them only where they are really needed.
As you said you are going needed to `#ifdef` them for the right compiler, like:
```c++
#if BOOST_COMP_GNUC
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wfloat-equal"
#endif
...
#if BOOST_COMP_GNUC
#endif
```
Also check if that works, because I think clang is also producing the warning.

I think we should not suppress the warning in general, because it is useful in other places. But we can debate this. @j-stephan?

tonydp03: Maybe we should discuss this offline and see if there could be a better solution rather than adding all this stuff in the code just to suppress warnings?

tonydp03: We might want to comment out these tests for float?
https://github.com/alpaka-group/alpaka/blob/0dcc191e502509fa658600a4abaed49711fff0a8/test/unit/atomic/src/AtomicTest.cpp#L317-L325
It seems that `atomicInc`, `atomicDec` and `atomicCas` are not supported with floats and they perform float comparison, which is what it might cause the warning. But we would still need pragmas for the ALPAKA_CHECK with the float comparisons).

bernhardmgruber: Sure, comment out these atomics for floating point types! Add the pragmas in the other places please.

bernhardmgruber:
You may want to replace all the diagnostic disabling by a new function that @sliwowitz will bring in soon: #1440.

tonydp03:
> You may want to replace all the diagnostic disabling by a new function that @sliwowitz will bring in soon: #1440.

I just saw it's already been merged. What would be the best way, just closing the repo and adding the atomic test changes to the clean repository to open a new PR? (Since this is a draft one...)

bernhardmgruber:
> > You may want to replace all the diagnostic disabling by a new function that @sliwowitz will bring in soon: #1440.
> 
> I just saw it's already been merged. What would be the best way, just closing the repo and adding the atomic test changes to the clean repository to open a new PR? (Since this is a draft one...)

Rebase on `develop` and sort out potential conflicts, then remove the pragmas and use @sliwowitz's function. You might also want to start new from `develop` and apply the non-pragma changes and then. In any case, I would force push to the same remote branch name here and continue on this PR.

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -17,6 +18,24 @@
...
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals(T1 a, T2 b)
+{
+    return a == b;
+}
+
+template<>
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals<float, float>(float a, float b)
+{
+    return alpaka::math::floatEqualExactNoWarning(a, b);
+}
```
bernhardmgruber:
```suggestion
ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals(float a, float b)
{
    return alpaka::math::floatEqualExactNoWarning(a, b);
}
```

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -17,6 +18,24 @@
...
+
+template<>
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals<double, double>(double a, double b)
+{
+    return alpaka::math::floatEqualExactNoWarning(a, b);
+}
```
bernhardmgruber:
```suggestion
ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals(double a, double b)
{
    return alpaka::math::floatEqualExactNoWarning(a, b);
}
```

test/unit/atomic/src/AtomicTest.cpp
bernhardmgruber: prefer function overload to function template specialization.

bernhardmgruber (changes requested):

test/unit/atomic/src/AtomicTest.cpp
tonydp03: Changes committed and pushed.

bernhardmgruber (dismissed):

bernhardmgruber:
@jkelling the only failing test is with nvc++ and I don't get the error message. Could you maybe have a look please? Thx!

psychocoderHPC:
I think the CI issue is this: https://gitlab.com/hzdr/crp/alpaka/-/jobs/1769570585#L3295

```
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 50)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 68)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 108)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 126)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 144)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 50)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 68)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 108)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 126)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 144)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 50)
NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 68)
```

psychocoderHPC:
I assume the CI issue is coming from testing 8bit and 16bit atomics with OpenAcc. @jkelling are 8 and 16bit atomics with OpenAcc supported?

jkelling:
> I assume the CI issue is coming from testing 8bit and 16bit atomics with OpenAcc. @jkelling are 8 and 16bit atomics with OpenAcc supported?

Technically, the standard does not define what data types atomics can work on, only which operations are supported. So, whatever the target supports is supported.

NVIDA GPUs do *not* support them, hence CUDA does not, hence NVHPC targeting nvptx does not. This is why https://github.com/alpaka-group/alpaka/commit/731e80d0ff3ba7ffea3985745639d4e4e5153d4b disabled these tests. This PR should apply the same exclusion mechanism established herein for CUDA on OpenACC and OpenMP5 (unfortunately we do not have a GPU test for this yet, but this would also fail for HSA).

jkelling:
> ```
> NVC++-S-0155-Invalid atomic capture.  (/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp: 50)
> [...]
>```

Note, that there is a bug in NVHPC which always prints the main source file there. The line number actually refers to [AtomicOaccBuiltIn.hpp:50](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/atomic/AtomicOaccBuiltIn.hpp#L50), which is an atomic capture. The reason for the error is the 8 or 16 bit datatype.

I am actually a surprised about the two errors in the NVHPC x86 test:
```
NVC++-S-0155-Invalid atomic capture.  (/home/runner/work/alpaka/alpaka/test/unit/atomic/src/AtomicTest.cpp: 46)
NVC++-S-0155-Invalid atomic capture.  (/home/runner/work/alpaka/alpaka/test/unit/atomic/src/AtomicTest.cpp: 52)
```
For one, these passed for me locally with 8 and 16 bit enabled. The other thing is that [AtomicOaccBuiltIn.hpp:46](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/atomic/AtomicOaccBuiltIn.hpp#L46) would point to the directive itself, while the nvptx errors always point to the end of the block. Could be, that it refers to a different file. The other one could mean [AtomicOaccExtended.hpp:50](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/atomic/AtomicOaccExtended.hpp#L50). Let us see if these go away too when 8 and 16 bit is disable for OpenACC.

test/unit/atomic/src/AtomicTest.cpp
```diff
@@ -759,6 +823,24 @@ class AtomicTestKernel<
 };
 #endif
 
+#ifdef ALPAKA_ACC_ANY_BT_OACC_ENABLED
+template<typename TDim, typename TIdx, typename T>
+    class AtomicTestKernel < alpaka::AccOacc<TDim, TIdx>,
+    T,
+    std::enable_if_t<sizeof(T) <= 2> // disable 8-bit and 16-bit tests
```
bernhardmgruber:
```suggestion
    std::enable_if_t<sizeof(T) <= 2>> // disable 8-bit and 16-bit tests
```
You miss an angle bracket here.

tonydp03: Right, commit fixed.

BenjaminW3 (approved):

bernhardmgruber (approved):

PR #1432, opened 2021-10-25
Renamed tests

### LINKED ISSUE #1429: Cmake reading binary file for atomic test
After building the unit test for atomics, a binary file called "atomic" is created in the `build/test/unit/atomic/` directory. If the source code for the Atomic tests is then modified, the compiler fails to rebuild the application as it tries to read the binary file from the above mentioned directory instead of the std library header <atomic> (included in the source code).

Screenshot 2021-10-22 at 11 58 57

It has been verified that this also happen if including the std library header <queue> in the source code of the unit test called `queue`.

A quick and easy solution might be to rename the target name of all the tests, i.e.: from `atomic` to `test-atomic` or `atomicTest` in the CMake files, to preven this from happening for other tests as well.

tonydp03:
This PR renames all the binary files created when running tests with the logic "testnameTest". This avoid the bug of Cmake reading the binary file for (i.e.) the atomic test, as the std header <atomic> is included in the source file. The same issue might happen with other tests (i.e. if the queue test included the std header <queue> in the source file).

### DIFF
--- a/test/analysis/headerCheck/CMakeLists.txt
+++ b/test/analysis/headerCheck/CMakeLists.txt
@@ -10,7 +10,7 @@
 
 if(NOT ALPAKA_CI OR (ALPAKA_CI AND ALPAKA_CI_ANALYSIS))
 
-    set(_TARGET_NAME "headerCheck")
+    set(_TARGET_NAME "headerCheckTest")
 
     #---------------------------------------------------------------------------
     # Create source files.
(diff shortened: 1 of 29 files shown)


bernhardmgruber (approved):
I think I would prefer the suffix `Tests` instead of `Test`, because each executable runs multiple tests. Furthermore, I think you could squash your two commits into one. I don't see a reason to have them separate.

Otherwise, LGTM.

sbastrakov (approved):

PR #1433, opened 2021-10-26
Add Accessor for views and buffers (only feature)

bernhardmgruber:
* add access mode tags for read, write and read-write access
* add readAccess()/writeAccess()/access()
* only allow writing via WriteAccess accessors
* add Accessor specific tests
* add example for an accessor with a projection
* allow constraining Accessors
* add BuildAccessor customization point
* view accessors are separate from generic Accessor functionality
* add MemoryHandle trait to query memory handle of an accessor
* add BufferAccessor convenience alias
* add brief documentation

This PR is split off from #1249 and only proposes the feature without changing any of the tests and examples.

### DIFF
--- a/docs/source/basic/cheatsheet.rst
+++ b/docs/source/basic/cheatsheet.rst
@@ -148,6 +148,11 @@ Get a raw pointer to a buffer or view initialization, etc.
      DataType* raw = view::getPtrNative(bufHost);
      DataType* rawViewPtr = view::getPtrNative(hostView);
 
+Get an accessor to a buffer and the accessor's type (experimental)
+  .. code-block:: c++
+
+     experimental::BufferAccessor<Acc, Elem, N, AccessTag> a = experimental::access(buffer);
+
 Allocate a buffer in device memory
   .. code-block:: c++
 
--- a/docs/source/basic/library.rst
+++ b/docs/source/basic/library.rst
@@ -212,6 +212,38 @@ It does not return raw pointers but reference counted memory buffer objects that
 Additionally the memory buffer objects know their extents, their pitches as well as the device they reside on.
 This allows buffers that possibly reside on different devices with different pitches to be copied only by providing the buffer objects as well as the extents of the region to copy (``alpaka::memcpy(bufDevA, bufDevB, copyExtents``).
 
+Accessors
+`````````
+
+An accessor is an interface to access the data stored by a buffer or, more generally, a view.
+Accessors take care of the multidimensionality of their underlying buffers when indexed, including pitched allocations.
+It is created via one of the following calls:
+
+.. code-block:: cpp
+
+   auto accessor = alpaka::experimental::access(buffer);      // read/write
+   auto accessor = alpaka::experimental::readAccess(buffer);  // read
+   auto accessor = alpaka::experimental::writeAccess(buffer); // write
+
+Accessors have many template parameter and users are adviced to use the ``BufferAccessor`` convenience alias to get the type of an accessor for a buffer of a given accelerator.
+Example kernel with 3D accessor:
+
+.. code-block:: c++
+
+   struct Kernel {
+      template<typename Acc>
+      ALPAKA_FN_ACC void operator()(Acc const & acc, alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::WriteAccess> data) const {
+         ...
+         for(Idx z = 0; z < data.extents[0];  ++z)
+             for(Idx y = 0; y < data.extents[1]; ++y)
+                 for(Idx x = 0; x < data.extents[2];  ++x)
+                     data(z, y, x) = 42.0f;
+      }
+   };
+   ...
+   alpaka::exec<Acc>(queue, workDiv, kernel, alpaka::experimental::writeAccess(buffer));
+
+
 Kernel Execution
 ````````````````
 
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Axel Huebl, Benjamin Worpitz, Erik Zenker, Matthias Werner, René Widera
+/* Copyright 2019-2021 Axel Huebl, Benjamin Worpitz, Erik Zenker, Matthias Werner, René Widera, Bernhard Manfred Gruber
  *
  * This file is part of alpaka.
  *
@@ -131,7 +131,9 @@
 #include <alpaka/mem/fence/MemFenceOmp5.hpp>
 #include <alpaka/mem/fence/MemFenceUniformCudaHipBuiltIn.hpp>
 #include <alpaka/mem/fence/Traits.hpp>
+#include <alpaka/mem/view/Accessor.hpp>
 #include <alpaka/mem/view/Traits.hpp>
+#include <alpaka/mem/view/ViewAccessor.hpp>
 #include <alpaka/mem/view/ViewCompileTimeArray.hpp>
 #include <alpaka/mem/view/ViewPlainPtr.hpp>
 #include <alpaka/mem/view/ViewStdArray.hpp>
--- a/include/alpaka/mem/view/Accessor.hpp
+++ b/include/alpaka/mem/view/Accessor.hpp
@@ -1,0 +1,192 @@
+/* Copyright 2021 Bernhard Manfred Gruber
+
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED “AS IS” AND ISC DISCLAIMS ALL WARRANTIES WITH
+ * REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
+ * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
+ * IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#pragma once
+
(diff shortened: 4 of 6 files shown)


docs/source/basic/cheatsheet.rst
```diff
@@ -148,6 +148,11 @@ Get a raw pointer to a buffer or view initialization, etc.
      DataType* raw = view::getPtrNative(bufHost);
      DataType* rawViewPtr = view::getPtrNative(hostView);
 
+Get an accessor to a buffer and the accessor's type
+  .. code-block:: c++
+
+     BufferAccessor<Acc, Elem, N, AccessTag> a = access(buffer);
```
SimeonEhrig: Is the namespace `experimental::` required, because in the examples above, `view::` is also used.

docs/source/basic/library.rst
```diff
@@ -212,6 +212,35 @@ It does not return raw pointers but reference counted memory buffer objects that
...
+Accessors have many template parameter and users are adviced to use the ``BufferAccessor`` convenience alias to get the type of an accessor for a buffer of a given accelerator.
+Example kernel with 3D accessor:
+
+.. code-block:: c++
+
+   struct Kernel {
+      template<typename Acc>
+      ALPAKA_FN_ACC void operator()(Acc const & acc, alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::ReadAccess> data) const {
+         ...
+         data(z, y, x) = 42.0f;
```
SimeonEhrig: In my opinion, the example should be extend by the use of `extents`. 

```
for(auto z = 0; z < data.extents[0];  ++z)
   for(auto y = 0; y < data.extents[1]; ++y)
     for(auto x = 0; x < data.extents[2];  ++x)
       data(z, y, x) = 42.0f;
```

By the way, does `accessors` respect the pitching? You should document this also in small sentence.

docs/source/basic/library.rst
```diff
@@ -212,6 +212,35 @@ It does not return raw pointers but reference counted memory buffer objects that
...
+   auto accessor = alpaka::experimental::writeAccess(buffer); // write
+
+Accessors have many template parameter and users are adviced to use the ``BufferAccessor`` convenience alias to get the type of an accessor for a buffer of a given accelerator.
+Example kernel with 3D accessor:
+
+.. code-block:: c++
+
+   struct Kernel {
+      template<typename Acc>
+      ALPAKA_FN_ACC void operator()(Acc const & acc, alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::ReadAccess> data) const {
```
SimeonEhrig: I think, this should be `alpaka::experimental::writeAccess`

If yes, the `alpaka::exec` needs also be modified.

docs/source/basic/cheatsheet.rst
bernhardmgruber: Yes it is required until the feature is stable enough to remove the `experimental` namespace. I was not sure how to deal with the cheatsheet in general, because I am not sure if we should advertise the feature too much already before it is fully baked.

docs/source/basic/library.rst
bernhardmgruber: Good point! Yes the accessor does respect the pitch, that is the whole point of it :D

docs/source/basic/library.rst
bernhardmgruber: Good catch! Thx!

docs/source/basic/library.rst
```diff
@@ -212,6 +212,38 @@ It does not return raw pointers but reference counted memory buffer objects that
...
+
+An accessor is an interface to access the data stored by a buffer or, more generally, a view.
+Accessors take care of the multidimensionality of their underlying buffers when indexed, including pitched allocations.
+It is created via one of the following calls:
+
+.. code-block:: cpp
+
+   auto accessor = alpaka::experimental::access(buffer);      // read/write
+   auto accessor = alpaka::experimental::readAccess(buffer);  // read
+   auto accessor = alpaka::experimental::writeAccess(buffer); // write
```
SimeonEhrig, line 226:
```suggestion
   auto accessor = alpaka::experimental::ReadWriteAccess(buffer);      // read/write
   auto accessor = alpaka::experimental::ReadAccess(buffer);  // read
   auto accessor = alpaka::experimental::WriteAccess(buffer); // write
```

SimeonEhrig:
I will never develop without accessors alpaka applications again. It's like moving from C to C++ :grin: 

It's like I used:
https://github.com/SimeonEhrig/my-alpaka-examples/blob/master/matmul_accessors/matmul.cpp

docs/source/basic/library.rst
bernhardmgruber: I get the suggestion to rename `access` to `readWriteAccess`. But these are functions, so I would like to keep them lowercase.

SimeonEhrig: Sorry, I mixed up the function, where you get the accessor on the host side with the accessor types in an alpaka kernel. Therefore the documentation is correct and my suggested change is wrong.

There is only one small point. I mixed up the host and kernel accessor, because the only difference between the host and the device version of the read and write accessor is the case of the first letter. Should we think about, to make the names more unique or do you think, this is no problem?

bernhardmgruber: I think you mixed up the second point as well. There is no difference between host and device accessors. And there shouldn't be one. You always produce and `Accessor` from an existing view or accessor with `access(...)`.

SimeonEhrig: Yes, I mixed every point up. So this means `access`, `readAccess` and `writeAccess` are factory methods? In this case, I would suggest to replace the `auto` with actual type to document the relationship.

```suggestion
   alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::ReadWriteAccess> accessor = 
     alpaka::experimental::access(buffer);
   alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::ReadAccess> accessor = 
     alpaka::experimental::readAccess(buffer);  
   alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::WriteAccess> accessor = 
     alpaka::experimental::writeAccess(buffer); 
```

bernhardmgruber: I would like to avoid that as well, because `access` only returns a `BufferAccessor` for the current alpaka buffers. It returns something else for views I think (a different instantiation of `Accessor` than the `BufferAccessor` alias). Also SYCL might change the landscape here. Users should definitely use `auto` in there code.

SimeonEhrig: Okay, that's one point. But my motivation for the concrete type instead of `auto` is that I know what I'm getting and I can see in the documentation what the member function and variable of the type are and where I can use them. But this is not possible if the return type is not specified. That's why @sbastrakov suggests to use a "concepts-structs".

> E.g. this functor: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/fields/MaxwellSolver/Yee/StencilFunctor.hpp . Btw note that the T_Curl template parameter is also from Curl "concept". It's used here: https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/include/picongpu/fields/MaxwellSolver/FDTD/FDTD.kernel

In our case, maybe we can create a concept like:

```c++
struct ConceptAccessor {
public:
   //! Store dimension and size of each dimension
   Vec<...> extends
};
``` 

Than you can write in the documentation, that the return type has to fulfill the `ConceptAccessor` and point to the doxygen documentation.

And as bonus, if the some guy makes pressure, that we should switch to C++20 (and know, who it is ;-)), we can easily replace the concept-structs with real c++ concepts.

bernhardmgruber: The return type of `access` et al. is a an object which is callable on the number of dimensions and has a member `extends`, which is subscriptable  in the number of dimensions. This should also be obvious by the example below (library.rst:237) so I think an additional concept or concept struct is of little value.

I also intentionally keep the documentation of this feature short for now, because it is experimental and we for sure want to change it some times again. I am almost certain it will change after merging when @j-stephan will rebase SYCL on top of it. So I would like to wait with more user side information until the details are fleshed out.

SimeonEhrig (approved):

docs/source/basic/library.rst
```diff
@@ -212,6 +212,38 @@ It does not return raw pointers but reference counted memory buffer objects that
...
+
+Accessors have many template parameter and users are adviced to use the ``BufferAccessor`` convenience alias to get the type of an accessor for a buffer of a given accelerator.
```
j-stephan, line 228: Two typos in this line:

```suggestion
Accessors have many template parameters and users are advised to use the ``BufferAccessor`` convenience alias to get the type of an accessor for a buffer of a given accelerator.
```

Not something that needs to be fixed right now but we should keep it in mind for the main accessor PR.

docs/source/basic/library.rst
```diff
@@ -212,6 +212,38 @@ It does not return raw pointers but reference counted memory buffer objects that
...
+Example kernel with 3D accessor:
+
+.. code-block:: c++
+
+   struct Kernel {
+      template<typename Acc>
+      ALPAKA_FN_ACC void operator()(Acc const & acc, alpaka::experimental::BufferAccessor<Acc, float, 3, alpaka::experimental::WriteAccess> data) const {
```
j-stephan, line 235: There should probably be a line break somewhere here.

j-stephan (approved):
Looks good to me. Things to keep in mind:

docs/source/basic/library.rst
bernhardmgruber: thx!

docs/source/basic/library.rst
bernhardmgruber: thx!

bernhardmgruber:
I fixed the typos and linebreak here: #1459.

docs/source/basic/library.rst
```diff
@@ -212,6 +212,38 @@ It does not return raw pointers but reference counted memory buffer objects that
...
+         ...
+         for(Idx z = 0; z < data.extents[0];  ++z)
+             for(Idx y = 0; y < data.extents[1]; ++y)
+                 for(Idx x = 0; x < data.extents[2];  ++x)
+                     data(z, y, x) = 42.0f;
```
fwyzard, line 240: This doesn't seem like a good example ?
I would assume that one wants different threads to write to different positions inside the buffer, not that all threads write to the whole buffer (without using atomic operations).

bernhardmgruber: For sure it can be improved! But it is good enough to showcase the API in a small amount of LOCs. More elaborate examples can be found once some of the standard alpaka examples are rewritten using accessors, see #1458.

PR #1434, opened 2021-10-27
Modify existing PRNGs to faclitate Philox implementation

sliwowitz:
* Made generators in RandStdLib.hpp implement the STL UniformRandomBitGenerator concept.
 * Renamed public members according to our current conventions.
 * Changed constructor parameters to (seed, subsequence, offset) in anticipation of merging the Philox family generators.

### DIFF
--- a/include/alpaka/rand/RandStdLib.hpp
+++ b/include/alpaka/rand/RandStdLib.hpp
@@ -15,6 +15,7 @@
 #include <alpaka/rand/Traits.hpp>
 
 #include <cstdint>
+#include <limits>
 #include <random>
 #include <type_traits>
 
@@ -48,6 +49,9 @@ namespace alpaka
                 class MersenneTwister
                 {
                 public:
+                    std::mt19937 state;
+
+                public:
                     MersenneTwister() = default;
 
                     ALPAKA_FN_HOST MersenneTwister(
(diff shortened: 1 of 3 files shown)


include/alpaka/rand/RandStdLib.hpp
```diff
@@ -55,12 +59,24 @@ namespace alpaka
...
+                    // STL UniformRandomBitGenerator concept interface
+                    using result_type = decltype(state)::result_type;
+                    ALPAKA_FN_HOST constexpr result_type min()
+                    {
+                        return 0;
+                    }
+                    ALPAKA_FN_HOST constexpr result_type max()
+                    {
+                        return std::numeric_limits<result_type>::max();
+                    }
```
bernhardmgruber: Shouldn't this ask `state` for min and max? Maybe it's also easier if you just inherit from `std::mt19937`.

sliwowitz: You're right. It should. Inheriting *might* be an option, but it would change the API significantly as you could not access the `state` member at all (now it's just renamed from `m_State` to `state` which should be a trivial change in any dependent codebase)

bernhardmgruber: Fair point on the `state` member, since that is part of the public API.

bernhardmgruber (dismissed):

bernhardmgruber (approved):

PR #1435, opened 2021-10-28
Wrapper adding operator[] to Cuda vector types

sliwowitz:
This wrapper add `operator[]` to CUDA vector types such as `int4`, `float3` etc. It is a user convenience for the API of the vector variants of the _Philox_ random generator. In a perfect world, these would return a `std::array`. Unfortunately many older `nvcc`s do not support `std::array`s in device code. Their vector types are a good match for the _Philox_ generator, but they're structs with members `.x`, `.y`, etc. This syntax doesn't make much sense for a result of a random number generator, thus this wrapper.

The limitation is that it does not provide all possible type-size combination. If more are needed, they have to be manually added as new `struct CudaVectorArrayTypeTraits<T, N>` specializations. I don't know how to generate these automatically without resorting to preprocessor macro black magic, and given the wrappers will hopefully soon become obsolete, I included only `int`, `unsigned`, `float` and `double` in sizes `1` to `4`.

### DIFF
--- a/include/alpaka/meta/CudaVectorArrayWrapper.hpp
+++ b/include/alpaka/meta/CudaVectorArrayWrapper.hpp
@@ -1,0 +1,251 @@
+/* Copyright 2021 Jiri Vyskocil
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED) || defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+
+#    include <alpaka/core/Common.hpp>
+#    include <alpaka/core/Unused.hpp>
+
+#    include <functional>
+#    include <initializer_list>
+#    include <numeric>
+#    include <type_traits>
+
+#    if(__cplusplus > 201402L)
+#        define CXX17_CONSTEXPR constexpr
+#    else
+// This is mostly needed because of nvcc 9 and should be removed upon moving to C++17.
+// Maybe some others need it too. GNU STL library uses the same trick.
+#        define CXX17_CONSTEXPR
+#    endif
+
+namespace alpaka
+{
+    namespace meta
+    {
+        namespace detail
+        {
+            template<typename TScalar, unsigned N>
+            struct CudaVectorArrayTypeTraits;
+
+            template<>
+            struct CudaVectorArrayTypeTraits<float, 1>
+            {
+                using type = float1;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<float, 2>
+            {
+                using type = float2;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<float, 3>
+            {
+                using type = float3;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<float, 4>
+            {
+                using type = float4;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<double, 1>
+            {
+                using type = double1;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<double, 2>
+            {
+                using type = double2;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<double, 3>
+            {
+                using type = double3;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<double, 4>
+            {
+                using type = double4;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<unsigned, 1>
+            {
+                using type = uint1;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<unsigned, 2>
+            {
+                using type = uint2;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<unsigned, 3>
+            {
+                using type = uint3;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<unsigned, 4>
+            {
+                using type = uint4;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<int, 1>
+            {
+                using type = int1;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<int, 2>
+            {
+                using type = int2;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<int, 3>
+            {
+                using type = int3;
+            };
+
+            template<>
+            struct CudaVectorArrayTypeTraits<int, 4>
+            {
+                using type = int4;
+            };
+        } // namespace detail
+
+        /// Helper struct providing [] subscript access to CUDA vector types
+        template<typename TScalar, unsigned N>
+        struct CudaVectorArrayWrapper;
+
+        template<typename TScalar>
+        struct CudaVectorArrayWrapper<TScalar, 4> : public detail::CudaVectorArrayTypeTraits<TScalar, 4>::type
+        {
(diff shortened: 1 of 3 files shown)


test/unit/meta/src/IsArrayOrVectorTest.cpp
```diff
@@ -0,0 +1,67 @@
...
+float arrayFloat[4] = {1.0f, 2.0f, 3.0f, 4.0f};
+
+float notAnArrayFloat = 15.0f;
+float* notAnArrayFloatPointer = &notAnArrayFloat;
+std::string notAnArrayString{"alpaka"};
+
+TEST_CASE("isArrayOrVector", "[meta]")
+{
+    constexpr bool isArrayOrVectorStdArray = alpaka::meta::IsArrayOrVector<std::array<int, 10>>::value;
+    static_assert(isArrayOrVectorStdArray, "alpaka::meta::IsArrayOrVector failed!");
```
bernhardmgruber:
```suggestion
    STATIC_REQUIRE(alpaka::meta::IsArrayOrVector<std::array<int, 10>>::value);
```
Catch2 can do that for you. Please apply in other places as well.

include/alpaka/meta/IsArrayOrVector.hpp
```diff
@@ -0,0 +1,72 @@
...
+namespace alpaka
+{
+    namespace meta
+    {
+        /** Checks whether T is an array or a vector type
+         *
+         * @tparam T a type to check
+         */
+        template<typename T>
+        struct IsArrayOrVector : std::false_type
```
bernhardmgruber: I don't understand the semantic and use of that trait. Why only vector or array? What about `std::deque`? Is this supposed to be a concept? Would `llama::Array` fulfill that? What decisions are you going to take based on this information?

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -0,0 +1,252 @@
...
+            {
+                using type = int4;
+            };
+        } // namespace detail
+
+        /// Helper struct providing [] subscript access to CUDA vector types
+        template<typename TScalar, unsigned N>
+        struct CudaVectorArrayWrapper
+        {
+        };
```
bernhardmgruber:
```suggestion
        struct CudaVectorArrayWrapper;
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -0,0 +1,252 @@
...
+namespace alpaka
+{
+    namespace meta
+    {
+        namespace detail
+        {
+            template<typename TScalar, unsigned N>
+            struct CudaVectorArrayTypeTraits
+            {
+            };
```
bernhardmgruber:
```suggestion
            struct CudaVectorArrayTypeTraits;
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -0,0 +1,252 @@
...
+                this->x = *it++;
+                this->y = *it++;
+                this->z = *it++;
+                this->w = *it++;
+            }
+            ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE CXX17_CONSTEXPR value_type& operator[](int const k) noexcept
+            {
+                assert(k > 0 && k < 4);
+                return k == 0 ? this->x : (k == 1 ? this->y : (k == 2 ? this->z : this->w));
+            }
```
bernhardmgruber: Why can't we make this `constexpr` before C++17?

bernhardmgruber (changes requested):

bernhardmgruber:
Please rebase this PR on the recent develop!

include/alpaka/meta/CudaVectorArrayWrapper.hpp
sliwowitz: Because *nvcc 9* will then fail with `qualifiers dropped in binding reference of type "unsigned int &" to initializer of type "const unsigned int"` when you do e.g. `unsigned x = myVector[2]`. The newer compiler I tried will accept the `constexpr` even in C++14 mode, but there's those infinite other combinatins in the CI. I was looking for anything useful on that matter and found out that GNU STL libstdc++ uses this exact trick in their `std::array` implementation which seemed good enough for me.

test/unit/meta/src/IsArrayOrVectorTest.cpp
sliwowitz: Didn't know that. I used to work with Boost.Test so Catch2 is new to me. Thanks!

include/alpaka/meta/IsArrayOrVector.hpp
sliwowitz: I use it to select a correct scalar/vector implementation of distributions - something like `UniformReal<PRNG_Result>` where `PRNG_Result` is usually a scalar, but could be a vector. Thinking about it, the concept could probably be satisfied by any collection that provides `operator[]`

include/alpaka/meta/CudaVectorArrayWrapper.hpp
sliwowitz: done

include/alpaka/meta/CudaVectorArrayWrapper.hpp
sliwowitz: done

sliwowitz:
> Please rebase this PR on the recent develop!

done

include/alpaka/meta/IsArrayOrVector.hpp
bernhardmgruber: I conjectured that you are requiring `operator[]` and thus your concept should be called something like `Subscriptable`. Then than you could use SFINAE to check it. Something like:
```c++
template <typename T, typename SFINAE = void>
struct Subscriptable : std::false_type {};

template <typename T>
struct Subscriptable<T, std::void_t<decltype(std::declval<T>()[0])> > : std::true_type {};
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
bernhardmgruber: Ahh, you might be right here. There was a time `constexpr` on a member function implied `const`, but I thought that changed from C++11 to C++14? Might be another nvcc bug.

sliwowitz: Yes. My guess would be a *nvcc* bug as well. But like I said, GNU STL's `std::array` does the same thing to their `operator[]` so I think it should be ok.

sliwowitz: Same in LLVM's `libcxx` https://github.com/llvm-mirror/libcxx/blob/78d6a7767ed57b50122a161b91f59f19c9bd0d19/include/array#L191

include/alpaka/meta/IsArrayOrVector.hpp
sliwowitz: I'm not sure about this because that would also accept a plain pointer like `float* x` and call the vector distribution on that which is probably not what was intended. I'm leaning to dropping this altogether, and instead requiring something like `T::size` because the distribution class needs to know the size of the vector anyway. This would not work on plain arrays though, but that's non-standard usage so let's say if someone really wants to put their custom array into our distributions, they have to wrap it in something like `std::array` first.

bernhardmgruber: You said you only need `operator[]`. Now you need a size as well :D

Anyway, then I let you figure out what you need.

bernhardmgruber (dismissed):

include/alpaka/meta/IsArrayOrVector.hpp
sliwowitz: I think this discussion would better fit the in the Philox PR where this stuff is actually used. I'll remove the two files form here, and solve it there after the CI doom subsides :-)

bernhardmgruber (approved):

Issue #1414, opened 2021-09-24
Some `alpaka::math::MathStdLib` functions trigger `[-Werror,-Wdouble-promotion]` when not called with `double`


jkelling:
Using `alpaka::math::MathStdLib` functions `sin` and `cos` with an aruments of a different precision than `float` (e.g. `float`), and probably other triggers the following clang (Main/14) warning:
```
~alpaka/include/alpaka/math/sin/Traits.hpp:37:32:
error: implicit conversion increases floating-point precision: 'const float' to 'double' [-Werror,-Wdouble-promotion]
                    return sin(arg);
                           ~~~ ^~~
```

This is because, e.g., `sin` is not overloaded for different precisions. For `float` `sinf` should be used. (`sinl` for `long double` although this may not be supported by alpaka anyway.)

psychocoderHPC:
Which backend is producing the error?

jkelling:
Currently none that is upstream.

The CI for #1308 showed float-to-doubke promotion warnings, which wanted to addressed them, by using the requested type consistently. This gets blocked by `alpaka::math` functions prompting float arguments to a double return value.

jkelling:
One could also say: "all backends implementing the math concept using `alpaka::math::MathStdLib`", though. This warning would be triggered in user code when the user wants to use float, and it would point to an actual problem.

jkelling:
This problem is not limited to clang. With [gcc a different warning is raised](https://github.com/alpaka-group/alpaka/pull/1308/checks?check_run_id=3777533235):
```
/home/runner/work/alpaka/alpaka/include/alpaka/rand/RandAlpaka.hpp:122:76: error: implicit conversion from 'float' to 'double' to match other operand of binary expression [-Werror=double-promotion]
  122 |                             const T mag = sigma * sqrt(static_cast<T>(-2.) * log(u1));
      |                                                        ~~~~~~~~~~~~~~~~~~~~^~~~~~~~~
/home/runner/work/alpaka/alpaka/include/alpaka/rand/RandAlpaka.hpp:122:49: error: implicit conversion from 'float' to 'double' to match other operand of binary expression [-Werror=double-promotion]
  122 |                             const T mag = sigma * sqrt(static_cast<T>(-2.) * log(u1));
      |                                           ~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #1436, opened 2021-10-29
Disable float-equal warning

bernhardmgruber:
Disable the warning on floating point comparisons in the unit tests because it is too noisy.

### DIFF
--- a/test/common/devCompileOptions.cmake
+++ b/test/common/devCompileOptions.cmake
@@ -63,7 +63,8 @@ ELSE()
         # Higher levels (max is 5) produce some strange warnings
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wstrict-overflow=2")
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wtrampolines")
-        LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wfloat-equal")
+        # Too much noise
+        #LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wfloat-equal")
         LIST(APPEND ALPAKA_DEV_COMPILE_OPTIONS "-Wundef")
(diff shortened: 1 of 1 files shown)


jkelling:
I would prefer to be careful with the float types instead. I will prepare a PR for https://github.com/alpaka-group/alpaka/issues/1414 , which may reduce your problem a bit.

bernhardmgruber:
Well, this PR is motivated by the many warnings we need to disable in various tests, because there it is actually fine to just compare two floats. See e.g.: https://github.com/alpaka-group/alpaka/pull/1431/files. I think the warning is too noisy and the additional costs outweight the benefits. Furthermore, since alpaka does not employ much numeric code, chances are also small that we will run into missing an epsilon comparison that this warning is suggesting to do.

jkelling:
Thanks for posting the link to the diff.

Disabling them here looks reasonable, as there is no complex computation going on. Both sides of the comparison are using the same operations ...  on the face of it ... , so one could expect them to be always exactly equal.

The only way this could go wrong is if some hardware uses different FP units for computations in registers and atomics (special FP close to the memory, like GPUs have for textures). If the close-to-memory units are not IEEE compliant these test could fail. ...  I hoping this is completely hypothetical... .

BenjaminW3:
Can't we find a middle ground and disable the warning for the whole `AtomicTest.cpp` file instead of for each comparison?

bernhardmgruber:
> Can't we find a middle ground and disable the warning for the whole `AtomicTest.cpp` file instead of for each comparison?

Sure. But @sliwowitz also ran into this issue with new tests on his work on RNGs. So I figured the warning is probably not worth the additional hassle. But I gave my reasons. If most people agree on keeping it, we can leave the warning.

bernhardmgruber:
We decided in the alpaka meeting now to keep the warning. In cases where the warning is unjustified, we should add the corresponding pragmas or find a different work around.

PR #1440, opened 2021-11-02
Add function to bypass compiler warnings for exact float comparisons

sliwowitz:
This replaces the discontinued PR #1436 with a function that can be explicitly called when exact float comparison is needed. It should be used sparingly (this is repeatedly mentioned in the comments), but in case it is needed, it can help reduce source length (e.g. in PR #1431) by moving the necessary pragmas to one place.

### DIFF
--- a/include/alpaka/math/FloatEqualExact.hpp
+++ b/include/alpaka/math/FloatEqualExact.hpp
@@ -1,0 +1,57 @@
+/* Copyright 2021 Jiri Vyskocil
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/Common.hpp>
+
+#include <type_traits>
+
+namespace alpaka
+{
+    namespace math
+    {
+        /** Compare two floating point numbers for exact equivalence. Use only when necessary, and be aware of the
+         * implications. Most codes should not use this function and instead implement a correct epsilon-based
+         * comparison. If you are unfamiliar with the topic, check out
+         * https://www.geeksforgeeks.org/problem-in-comparing-floating-point-numbers-and-how-to-compare-them-correctly/
+         * or Goldberg 1991: "What every computer scientist should know about floating-point arithmetic",
+         * https://dl.acm.org/doi/10.1145/103162.103163
+         *
+         * This function calls the == operator for floating point types, but disables the warning issued by the
+         * compiler when compiling with the float equality warning checks enabled. This warning is valid an valuable in
+         * most codes and should be generally enabled, but there are specific instances where a piece of code might
+         * need to do an exact comparison (e.g. @a CudaVectorArrayWrapperTest.cpp). The verbose name for the function
+         * is intentional as it should raise a red flag if used while not absolutely needed. Users are advised to add a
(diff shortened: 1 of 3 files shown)


include/alpaka/math/FloatEqualExact.hpp
```diff
@@ -0,0 +1,54 @@
...
+         *
+         * @tparam T both operands have to be the same type and conform to std::is_floating_point
+         * @param a first operand
+         * @param b second operand
+         * @return a == b
+         */
+        template<typename T>
+        ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC typename std::enable_if_t<std::is_floating_point<T>::value, bool>
+        floatEqualExactNoWarning(T a, T b)
+        {
```
bernhardmgruber: Why the `enable_if` instead of a `static_assert` in the body? I cannot imagine where you would need this function to be used in a SFINAE context where you would want to deselect a template because you cannot call e.g. `floatEqualExactNoWarning` on an `int`.

test/unit/meta/src/CudaVectorArrayWrapperTest.cpp
```diff
@@ -21,6 +22,43 @@
...
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals<float, int>(float a, int b)
+{
+    return alpaka::math::floatEqualExactNoWarning(a, static_cast<float>(b));
+}
+
+template<>
+ALPAKA_FN_INLINE ALPAKA_FN_HOST_ACC bool equals<double, int>(double a, int b)
+{
+    return alpaka::math::floatEqualExactNoWarning(a, static_cast<double>(b));
+}
```
bernhardmgruber: Isn't that a bit overengineered? You could just compare against FP literals in the code below or just cast to the FP type.

test/unit/meta/src/CudaVectorArrayWrapperTest.cpp
```diff
@@ -37,46 +75,25 @@ class CudaVectorArrayWrapperTestKernel
...
         static_assert(std::is_same<decltype(t1[0]), T&>::value, "CudaVectorArrayWrapper in-kernel type test failed!");
-#    ifdef __GNUC__
-#        pragma GCC diagnostic push
-#        pragma GCC diagnostic ignored "-Wfloat-equal"
-#    endif
-        ALPAKA_CHECK(*success, t1[0] == 0);
-#    ifdef __GNUC__
-#        pragma GCC diagnostic pop
-#    endif
+        ALPAKA_CHECK(*success, equals(t1[0], 0));
```
bernhardmgruber: Why not:
```suggestion
        ALPAKA_CHECK(*success, equals(t1[0], T{0}));
```
?

include/alpaka/math/FloatEqualExact.hpp
sliwowitz: You're right, thanks. No particular reason, probably just the force of habit.

test/unit/meta/src/CudaVectorArrayWrapperTest.cpp
sliwowitz: Right again. I forgot about T{0} which renders this contraption useless

test/unit/meta/src/CudaVectorArrayWrapperTest.cpp
sliwowitz: Yes. I don't know how I forgot about this. I must have been mentally stuck somewhere in C++11 for some reason.

bernhardmgruber (approved):

Issue #1441, opened 2021-11-02
Box-Muller states for the Philox PRNG


sliwowitz:
Box-Muller method is a fast way to generate normally distributed random numbers - especially on the GPU. This is mentioned in the [documentation](https://www.thesalmons.org/john/random123/releases/1.11.2pre/docs/) of the reference implemetation of the Philox algorithm. The reference implementation also shows how to include Box-Muller states in the Philox state variable.

PR #1444, opened 2021-11-03
Add RandAlpaka

jkelling:
### Features Added:
* Portable implementations of distributions (incl. Box-Muller for Gaussian)
* Default to TinyMersenneTwister, which is plain C and thus callable in
  many device contexts albeit maybe not most efficient on GPU

### Relations
* Required by https://github.com/alpaka-group/alpaka/pull/1308 to provide a generic PRNG that can be compiled for accelerator
* Needs to be taken into account by https://github.com/alpaka-group/alpaka/pull/1319 in so far as to not break https://github.com/alpaka-group/alpaka/pull/1308 .
* This may help to address https://github.com/alpaka-group/alpaka/issues/1441 by adding a simple Box-Muller transform.

# ToDo

- [x] merge after #1446
- [x] clean-up branch

### DIFF
--- a/include/alpaka/rand/RandDefault.hpp
+++ b/include/alpaka/rand/RandDefault.hpp
@@ -1,0 +1,235 @@
+/* Copyright 2021 Jeffrey Kelling
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/Common.hpp>
+#include <alpaka/core/Unused.hpp>
+#include <alpaka/math/cos/Traits.hpp>
+#include <alpaka/math/isnan/Traits.hpp>
+#include <alpaka/math/log/Traits.hpp>
+#include <alpaka/math/sin/Traits.hpp>
+#include <alpaka/math/sqrt/Traits.hpp>
+#include <alpaka/rand/RandPhilox.hpp>
+#include <alpaka/rand/Traits.hpp>
+
+#include <algorithm>
+#include <limits>
+#include <type_traits>
+
+namespace alpaka
+{
+    namespace rand
+    {
+        class RandDefault : public concepts::Implements<ConceptRand, RandDefault>
+        {
+        };
+
+        namespace distribution
+        {
+            namespace gpu
+            {
+                namespace detail
+                {
+                    template<typename TFloat>
+                    struct BitsType;
+
+                    template<>
+                    struct BitsType<float>
+                    {
+                        using type = std::uint32_t;
+                    };
+                    template<>
+                    struct BitsType<double>
+                    {
+                        using type = std::uint64_t;
+                    };
+                } // namespace detail
+
+                //! The GPU random number normal distribution.
+                template<typename T>
+                class UniformUint
+                {
+                    static_assert(std::is_integral<T>::value, "Return type of UniformUint must be integral.");
+
+                public:
+                    UniformUint() = default;
+
+                    template<typename TEngine>
+                    ALPAKA_FN_HOST_ACC auto operator()(TEngine& engine) -> T
+                    {
+                        using BitsT = typename TEngine::result_type;
+                        T ret = 0;
+                        constexpr auto N = sizeof(T) / sizeof(BitsT);
+                        for(unsigned int a = 0; a < N; ++a)
+                        {
+                            ret
+                                ^= (static_cast<T>(engine())
+                                    << (sizeof(BitsT) * std::numeric_limits<unsigned char>::digits * a));
+                        }
+                        return ret;
+                    }
+                };
+
+                //! The GPU random number uniform distribution.
+                template<typename T>
+                class UniformReal
+                {
+                    static_assert(
+                        std::is_floating_point<T>::value,
+                        "Return type of UniformReal must be floating point.");
+
+                    using BitsT = typename detail::BitsType<T>::type;
+
+                public:
+                    UniformReal() = default;
+
+                    template<typename TEngine>
+                    ALPAKA_FN_HOST_ACC auto operator()(TEngine& engine) -> T
+                    {
+                        constexpr BitsT limit = static_cast<BitsT>(1) << std::numeric_limits<T>::digits;
+                        const BitsT b = UniformUint<BitsT>()(engine);
+                        const auto ret = static_cast<T>(b & (limit - 1)) / limit;
+                        return ret;
+                    }
+                };
+
+                /*! The GPU random number normal distribution.
+                 *
+                 * \note
+                 * This type contains state and is not thread-safe: To be used
+                 * per thread, not shared.
+                 *
+                 * \note When reproducibility is a concern, each instance of
+                 * this class should be used with only on random engine
+                 * instance, or two consecutive number should be generated with
+                 * each engine used. This is due to the implicit caching of one
+                 * Gaussian random number.
+                 */
+                template<typename Acc, typename T>
+                class NormalReal
+                {
+                    static_assert(
+                        std::is_floating_point<T>::value,
+                        "Return type of NormalReal must be floating point.");
+
+                    const Acc& m_acc;
+                    T m_cache = std::numeric_limits<T>::quiet_NaN();
+
+                public:
+                    /*! \warning Retains a reference to \p acc, thus must not
+                     * outlive it.
+                     */
+                    NormalReal(const Acc& acc) : m_acc(acc)
+                    {
+                    }
+
+                    NormalReal(const NormalReal& o) = delete;
+
+                    //! The move ctor clears `m_cache` of source.
+                    //! \todo This is to be deleted when moving to C++17
+                    NormalReal(NormalReal&& o) : m_acc(o.m_acc), m_cache(o.m_cache)
+                    {
+                        o.m_cache = std::numeric_limits<T>::quiet_NaN();
+                    }
+
+                    template<typename TEngine>
+                    ALPAKA_FN_HOST_ACC auto operator()(TEngine& engine) -> T
+                    {
+                        constexpr T sigma = 1., mu = 0.;
+                        if(math::isnan(m_acc, m_cache))
+                        {
+                            UniformReal<T> uni;
+
+                            T u1, u2;
+                            do
+                            {
+                                u1 = uni(engine);
+                                u2 = uni(engine);
+                            } while(u1 <= std::numeric_limits<T>::epsilon());
+
+                            // compute z0 and z1
+                            const T mag = sigma * math::sqrt(m_acc, static_cast<T>(-2.) * math::log(m_acc, u1));
+                            constexpr T twoPi = static_cast<T>(2. * M_PI);
+                            // getting two normal number out of this, store one for later
+                            m_cache = mag * static_cast<T>(math::cos(m_acc, twoPi * u2)) + mu;
+
+                            return mag * static_cast<T>(math::sin(m_acc, twoPi * u2)) + mu;
+                        }
+                        else
+                        {
+                            const T ret = m_cache;
+                            m_cache = std::numeric_limits<T>::quiet_NaN();
+                            return ret;
+                        }
+                    }
+                };
+            } // namespace gpu
+        } // namespace distribution
+
+        namespace distribution
+        {
+            namespace traits
+            {
+                //! The GPU device random number float normal distribution get trait specialization.
+                template<typename T>
+                struct CreateNormalReal<RandDefault, T, std::enable_if_t<std::is_floating_point<T>::value>>
+                {
+                    template<typename TAcc>
+                    ALPAKA_FN_HOST_ACC static auto createNormalReal(TAcc const& acc)
+                    {
+                        return rand::distribution::gpu::NormalReal<TAcc, T>(acc);
+                    }
+                };
+                //! The GPU device random number float uniform distribution get trait specialization.
+                template<typename T>
+                struct CreateUniformReal<RandDefault, T, std::enable_if_t<std::is_floating_point<T>::value>>
+                {
+                    ALPAKA_FN_HOST_ACC static auto createUniformReal(RandDefault const& rand)
+                    {
+                        alpaka::ignore_unused(rand);
+                        return rand::distribution::gpu::UniformReal<T>();
+                    }
+                };
+                //! The GPU device random number integer uniform distribution get trait specialization.
+                template<typename T>
+                struct CreateUniformUint<RandDefault, T, std::enable_if_t<std::is_integral<T>::value>>
+                {
+                    ALPAKA_FN_HOST_ACC static auto createUniformUint(RandDefault const& rand)
+                    {
+                        alpaka::ignore_unused(rand);
+                        return rand::distribution::gpu::UniformUint<T>();
+                    }
+                };
+            } // namespace traits
+        } // namespace distribution
+        namespace engine
+        {
+            namespace traits
+            {
+                //! The GPU device random number default generator get trait specialization.
+                template<>
+                struct CreateDefault<RandDefault>
+                {
+                    template<typename TAcc>
+                    ALPAKA_FN_HOST_ACC static auto createDefault(
+                        TAcc const& acc,
+                        std::uint32_t const& seed,
+                        std::uint32_t const& subsequence,
+                        std::uint32_t const& offset) -> rand::Philox4x32x10<TAcc>
+                    {
+                        alpaka::ignore_unused(acc);
+                        return rand::Philox4x32x10<TAcc>(seed, subsequence, offset);
+                    }
+                };
+
+            } // namespace traits
+        } // namespace engine
+    } // namespace rand
+} // namespace alpaka
--- a/include/alpaka/acc/AccOacc.hpp
+++ b/include/alpaka/acc/AccOacc.hpp
@@ -21,7 +21,7 @@
 #    include <alpaka/intrinsic/IntrinsicFallback.hpp>
 #    include <alpaka/math/MathStdLib.hpp>
 #    include <alpaka/mem/fence/MemFenceOacc.hpp>
-#    include <alpaka/rand/RandStdLib.hpp>
+#    include <alpaka/rand/RandDefault.hpp>
 #    include <alpaka/time/TimeStdLib.hpp>
 #    include <alpaka/warp/WarpSingleThread.hpp>
 
@@ -66,7 +66,7 @@ namespace alpaka
         : public bt::IdxBtLinear<TDim, TIdx>
         , public math::MathStdLib
         , public MemFenceOacc
-        , public rand::RandStdLib
+        , public rand::RandDefault
         , public TimeStdLib
         , public warp::WarpSingleThread
         ,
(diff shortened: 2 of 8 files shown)


include/alpaka/rand/RandAlpaka.hpp
```diff
@@ -0,0 +1,200 @@
...
+
+namespace alpaka
+{
+    namespace rand
+    {
+        //! "Tiny" state mersenne twister implementation
+        class TinyMersenneTwisterGpu : public concepts::Implements<ConceptRand, TinyMersenneTwisterGpu>
+        {
+        };
+        using RandAlpaka = TinyMersenneTwisterGpu;
```
bernhardmgruber: I don't like the name `RandAlpaka`. Please either call it `RandTinyMersenneTwister` to indicate what implemenatation it is, or if that does not matter we could call it `RandDefault`. But I prefer the first one.

This would also affect the header names unfortunately.

jkelling: As #1319 is merged now: @sliwowitz do you think it would be simple to use the new Philox instead of the TinyMT?

We can rename to RandDefault.

sliwowitz: I think it should be possible. I'll give it a try and report back here.

bernhardmgruber:
Please rebase!

jkelling:
I would like to look into using the Philox instead of TinyMT before this is merged.The rebase will happen in this process.

jkelling:
* Renamed to `RandDefault`
* Uses Philox now
* Fix two signedness issues in Philox (`-Werror,-Wsign-compare,-Wsign-conversion`)

jkelling:
Can be merged after CI is through.

sliwowitz:
This PR removed  the STL generators - switching to Philox also removes TinyMT from the alpaka::rand API doesn't it?

jkelling:
> This PR removed the STL generators - switching to Philox also removes TinyMT from the alpaka::rand API doesn't it?

No. It only adds `RandDefault` which is a handle for the portable fallback RNG and random distributions. It does not touch `RandStdLib` or the TinyMT. The first version was using TinMT (the CPU version) as the fallback RNG, the change is to use Philox instead.

sliwowitz:
Oh. I thought you replaced the RandStdLib file, but now I see you're adding RandDefault as a default instead.

sliwowitz:
OK. A switch to Philox as seems like a good idea because then we have identical implementation for all `Acc`s by default. I'd like to check my implementations performance before we do a release, but that's no blocker for this PR.

sliwowitz (dismissed):

jkelling:
> OK. A switch to Philox as seems like a good idea because then we have identical implementation for all `Acc`s by default. I'd like to check my implementations performance before we do a release, but that's no blocker for this PR.

This is not across all `Acc`s. It is currently only used by `AccOacc` and `AccOmp5`, because these require the portability. Other have backend-specific RNG, like something based onf curan/hiprand for CUDA/HIP and `RandStdLib` for the CPU backends.

Having all backend use `RandDefault` as their default is an `alpaka::rand` redesign issue, which is outside the scope of this PR.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,198 @@
...
+            {
+                namespace detail
+                {
+                    template<typename TFloat>
+                    struct BitsType;
+
+                    template<>
+                    struct BitsType<float>
+                    {
+                        typedef std::uint32_t type;
```
bernhardmgruber:
```suggestion
                        using type = std::uint32_t;
```
In more places as well.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+                    template<typename TEngine>
+                    ALPAKA_FN_HOST_ACC auto operator()(TEngine& engine) -> T
+                    {
+                        using BitsT = typename TEngine::result_type;
+                        T ret = 0;
+                        constexpr auto N = sizeof(T) / sizeof(BitsT);
+                        for(unsigned int a = 0; a < N; ++a)
+                        {
+                            ret ^= (static_cast<T>(engine()) << (sizeof(BitsT) * 8 * a));
```
bernhardmgruber:
```suggestion
                            ret ^= (static_cast<T>(engine()) << (sizeof(BitsT) * CHAR_BIT * a));
```

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,198 @@
...
+                    struct BitsType<double>
+                    {
+                        typedef std::uint64_t type;
+                    };
+                } // namespace detail
+
+                //! The GPU random number normal distribution.
+                template<typename T>
+                class UniformUint
+                {
```
bernhardmgruber, line 58: Since further down in the code you assume that `T` is an integral type, consider adding a `static_assert` for that.

bernhardmgruber: Same for `UniformReal`.

include/alpaka/rand/RandAlpaka.hpp
psychocoderHPC: Please check the register-footprint of `Philox` before making it the default. As I know from @sliwowitz `Philox` is generating 4 numbers within one call so it could be generating rng number that could have a larger footprint.
A large memory footprint means mostly slower kernel execution.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,198 @@
...
+                public:
+                    NormalReal(const Acc& acc) : m_acc(acc)
+                    {
+                    }
+
+                    template<typename TEngine>
+                    ALPAKA_FN_HOST_ACC auto operator()(TEngine& engine) -> T
+                    {
+                        constexpr T sigma = 1., mu = 0.;
+                        if(std::isnan(m_buf))
```
bernhardmgruber: Can you call `std::isnan` in device code? I am pretty sure you must call an alpaka math routine here. And if it does not exist, create it.

include/alpaka/rand/RandAlpaka.hpp
jkelling: It is a fallback, not yet supposed to be the default for all `Acc`s. Using it as default for other backends should be discussed separately.

include/alpaka/rand/RandDefault.hpp
jkelling: Yes. This works with NVHPC and also OpenMP traget compilers aim to allow math functions form the `std` namespace. It also work in CUDA.

But your are right, that we may not want to rely on that. Adding this to `alpaka::math` should also come with it friends like `isfinite`, `isinf`, ... (six in total). I should put this in front in a separate PR, then.

include/alpaka/math/MathStdLib.hpp
```diff
@@ -65,6 +68,9 @@ namespace alpaka
             , public SqrtStdLib
             , public TanStdLib
             , public TruncStdLib
+            , public IsnanStdLib
```
psychocoderHPC: Please provide the first commit with new math functions within a separate PR.
This improvement is very useful but sneaked in via an independent RNG PR. It is not possible to write a changelog if this commit stays in this PR.

psychocoderHPC (changes requested):

include/alpaka/rand/RandDefault.hpp
psychocoderHPC: Yes please move this into a separate PR :heart:

include/alpaka/math/MathStdLib.hpp
jkelling: https://github.com/alpaka-group/alpaka/pull/1446

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,198 @@
...
+                        const auto ret = static_cast<T>(b & (limit - 1)) / limit;
+                        return ret;
+                    }
+                };
+
+                //! The GPU random number normal distribution.
+                template<typename Acc, typename T>
+                class NormalReal
+                {
+                    const Acc& m_acc;
```
psychocoderHPC, line 122: IMO the copy operator of `NormalReal` should be deleted to avoid that the user is storing this object in global data. Maybe adding a comment to oxygen of this class could also help.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,198 @@
...
+                            T u1, u2;
+                            do
+                            {
+                                u1 = uni(engine);
+                                u2 = uni(engine);
+                            } while(u1 <= std::numeric_limits<T>::epsilon());
+
+                            // compute z0 and z1
+                            const T mag = sigma * math::sqrt(m_acc, static_cast<T>(-2.) * math::log(m_acc, u1));
+                            constexpr T TwoPi = static_cast<T>(2. * M_PI);
```
psychocoderHPC: Why is `TwoPi` starting with upper case latter and all other variables with lower?

psychocoderHPC (changes requested):

include/alpaka/rand/RandDefault.hpp
bernhardmgruber: Technically, copy and move are already implicitly deleted because `NormalReal` contains a reference.

jkelling: @bernhardmgruber I do not think this is the case, at least with gcc 11:
```c++
#include <iostream>

struct A
{
	int& a;

	A(int &a) : a(a) {}
};

int main()
{
	int i = 23;
	A a(i);
	A b(a);
	a.a = 42;
	std::cout << b.a << std::endl;
}
```
Compiles without warnings `g++ --std=c++17 -Wall -pedantic` and prints what one would expect.

The compiler just turns the reference into a pointer. Using a reference there instead of pointer is just syntactic sugar, if the compiler does not optimize everything out.

psychocoderHPC: Should we explicitly delete this copy and/or move constructor?

jkelling: I am not sure how this really keeps the user form storing it in global memory and sharing it (good point about the comment).

Deleting copy would avoid someone copying the cache and possibly getting two times the same value. Not other distribution has deleted copies, so this would make this one different. If the copy is not deleted, it should not be trivial and clear the cache of the copy, but this would add overhead which is not there for other distributions. As it stands deleting copy could break generic code, so I would avoid this unless all distributions' copies are deleted (same for move.)

bernhardmgruber: Ok, let's be precise: the copy and move assignment operators are deleted. The constructors should not be impacted.

> The compiler just turns the reference into a pointer.

This happens in the codegen phase. But syntactically, you still have a reference which restricts the surrounding class.

jkelling: Deleting both move and copy breaks the tests unless we move to C++17 where copy elision is mandatory for return values.

bernhardmgruber: > Should we explicitly delete this copy and/or move constructor?

Even if you delete them, you could still placement new a `NormalReal` into a buffer. And IMO there is nothing wrong with copying/moving that class around. So I would even change the reference member to a pointer member, so the class can be copied/moved. But we should document that this class keeps a reference to the passed accelerator and that an instance should not outlive a kernel invocation.

jkelling: @bernhardmgruber was referring to the copy/move assignment operators there, those are implicitly deleted.

include/alpaka/rand/RandDefault.hpp
jkelling: Fixed.

include/alpaka/rand/RandDefault.hpp
jkelling: Pushed https://github.com/alpaka-group/alpaka/pull/1444/commits/a8b261fe42734112e484216522c5a436fdb27191 which cleanly implements copy and move ctors.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,226 @@
...
+
+                public:
+                    NormalReal(const Acc& acc) : m_acc(acc)
+                    {
+                    }
+
+                    //! The copy ctor does not copy `m_buf`.
+                    NormalReal(const NormalReal& o) : m_acc(o.m_acc)
+                    {
+                    }
```
bernhardmgruber: I think this behavior is surprising. If you take a copy of a `NormalReal` and then ask for the next number on the old and new instance, you will get different results.

I would prefer to have the default copy ctor here.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,226 @@
...
+                    //! The move ctor clears `m_buf` of source.
+                    NormalReal(NormalReal&& o) : m_acc(o.m_acc), m_buf(o.m_buf)
+                    {
+                        o.m_buf = std::numeric_limits<T>::quiet_NaN();
+                    }
```
bernhardmgruber: I don't think there is actually a need to define move operations for this class. What should the semantic be of moving a distribution? And why should that influence the moved from distribution? What is the need to destroy the moved from object? We don't get any efficiency by enabling move operations.

include/alpaka/rand/RandDefault.hpp
bernhardmgruber: > Deleting both move and copy breaks the tests unless we move to C++17 where copy elision is mandatory for return values.

Please also turn the stored accelerator object into a pointer, so the class can be freely copied. That should make the tests work.

jkelling: > And IMO there is nothing wrong with copying/moving that class around.
> So I would even change the reference member to a pointer member, so the class can be copied/moved.

There is no harm as such, but for performance it would be best to pass a reference to it around. Both copying and creating multiple instance only wastes registers on GPU. Letting an instance go out of scope and creating a new one when needed again also saves registers and is cheaper than a copy. I do not think copy and assignment of this type, or any alpaka distribution is useful, as they do not contain parameters (like min/max/sigma/mean).


> But we should document that this class keeps a reference to the passed accelerator and that an instance should not outlive a kernel invocation.

Ok. This is formally true for any object created from an `Acc`, i.e. all distributions and also random engines, this is just the only one that uses its `Acc` for something.

include/alpaka/rand/RandDefault.hpp
jkelling: This is only the distribution, the engine determines which numbers are generated. The expected behavior would be:
```c++
NormalReal<Acc> a(acc), b(acc);
Engine<Acc> e(acc);

inGeneral( a(e) != b(e) ) ; // because of two engine invocations

b = a;

inGeneral( a(e) != b(e) ) ; // because of two engine invocations
```

If a value was cached in `a` before `a = b` and the cache is not cleared, you break this behavior, because neither `a` nor `b` invoke the engine to ask for a new random number. This would not happen with a distribution that does not cache numbers. Caching is just an optimization, it must not interfere with the randomness.

include/alpaka/rand/RandDefault.hpp
jkelling: We can delete move. Though if it s there, is must be like this, for the same reason the copy should not copy the cache.

include/alpaka/rand/RandDefault.hpp
jkelling: Github issues do not appear to work when we write simultaneously. Your comments only appear to me somewhere in the middle long after I wrote something else, so I miss them. Then I get an email about a comment which turns out to be in the middle of the conversation... .

> > The compiler just turns the reference into a pointer.
> 
> This happens in the codegen phase. But syntactically, you still have a reference which restricts the surrounding class.

Yes, this is the intention: You cannot assign to a reference. `m_acc` should not change during the lifetime of the object, anything else would be a bug, hence the reference.

We could actually add assignments that do not touch the reference. See https://github.com/alpaka-group/alpaka/pull/1444/commits/31d18e754579a9ec05dd6ad71df443d293a2abb3 . This is the semantically clean way to do the assignments, though it is admittedly pointless. Note, that these object always exist withing a kernel, where there is only one `Acc` handle, i.e. you can never assign an object with a different `m_acc`.

include/alpaka/rand/RandDefault.hpp
```diff
@@ -0,0 +1,243 @@
...
+
+                    //! The copy must not obtain a copy of `m_buf`.
+                    NormalReal& operator=(const NormalReal& o)
+                    {
+                        ALPAKA_ASSERT_OFFLOAD(&m_acc == &o.m_acc);
+                    }
```
jkelling: This does nothing. Basically copy should be deleted because we cannot have two distributions with the same cached value around.

include/alpaka/rand/RandDefault.hpp
bernhardmgruber: Ok, that's right! I could still make the case tough that you could copy the engine as well and expect the results to be the same, but maybe that is not a relevant use case. Anyway, I think it's fine then.

include/alpaka/rand/RandDefault.hpp
bernhardmgruber: > Both copying and creating multiple instance only wastes registers on GPU. Letting an instance go out of scope and creating a new one when needed again also saves registers and is cheaper than a copy.

I strongly trust the optimizer to inline everything and remove all redundancies. Passing references around also costs registers and AFAIK CUDA cannot even form pointers/references into registers, so the optimizer must eliminate those anyway. So in the end it should not matter whether you copy distribution objects or pass them by references. And since the object is light-weight, I would prefer making it as regular as possible. But after this discussion, I don't care so much anymore. You are free to choose a behavior.

include/alpaka/rand/RandDefault.hpp
jkelling: You can either copy the cache or not, both have drawbacks, but the property I stated above is essential. If we give weight to your second case, the caching optimization not be possible. Copying the engine it a thing that might actually happen. One would expect a any distribution instance to give the same numbers on both engines, which also speaks against caching. However removing the cache and only using one of the two random numbers seems to sacrifice performance to support some very rare case. The way I see to minimize this problem is to disallow copy and move as far as possible (we only need the move ctor currently and can delete all when moving to C++17) and document the interaction of cache and engines. I pushed https://github.com/alpaka-group/alpaka/pull/1444/commits/004ae6cdd8f4a3321ecc0c0e22bf96cb9074204e along those lines.

jkelling:
The class `NormalReal` now has a deleted copy ctor and the move ctor is marked to be deleted when alpaka moves to C++17 by a doxgen `\todo`. The move ctor clears the cache of the moved-from instance.

@bernhardmgruber @psychocoderHPC  if you approve, I will clean-up the branch before merge.

include/alpaka/rand/RandDefault.hpp
jkelling: The CUDA compiler will inline everything and remove reference function arguments in the process, as they are just the same variable. If you copy, in the case for this `NormalReal` is not redundant because it contains the cache variable which changes independently for original and copy in their respective `operator()` calls, thus the copy cannot be optimized out (I agree, that this will be different for all other distribution, because they do not have any data members). This behavior does not depend on how exactly the copy ctor or assignment behave.

Going with deleting as much as possible, to try keep the user from getting unexpected results at runtime from this class.

psychocoderHPC (dismissed):

jkelling:
@psychocoderHPC rebased and ready for merge after CI.

bernhardmgruber (approved):

PR #1450, opened 2021-11-08
Isolate parallel region for nested parallelism with tbb backend

### LINKED ISSUE #1449: Inner-loop parallelism is not isolated when using tbb backend
When using alpaka tbb backend from within a parallel section of a code that is already parallelized using tbb, the alpaka kernel  becomes the innermost parallel loop of a nested parallel section.
As tbb treats all the tasks the same, whether they come from the outermost or the innermost loop and a thread might steal a task. If not protected, this might lead to deadlocks or unexpected results.

More details in: 
https://link.springer.com/chapter/10.1007/978-1-4842-4398-5_12

felicepantaleo:
This PR solves the issue https://github.com/alpaka-group/alpaka/issues/1449

Some benchmarks will follow.

### DIFF
--- a/include/alpaka/kernel/TaskKernelCpuTbbBlocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuTbbBlocks.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Benjamin Worpitz, Erik Zenker, René Widera
+/* Copyright 2019 Benjamin Worpitz, Erik Zenker, René Widera, Felice Pantaleo
  *
  * This file is part of alpaka.
  *
--- a/include/alpaka/kernel/TaskKernelCpuTbbBlocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuTbbBlocks.hpp
@@ -98,23 +98,25 @@ namespace alpaka
                 throw std::runtime_error("A block for the TBB accelerator can only ever have one single thread!");
             }
 
-            tbb::this_task_arena::isolate([&] {
-                tbb::parallel_for(
-                    static_cast<TIdx>(0),
-                    static_cast<TIdx>(numBlocksInGrid),
-                    [&](TIdx i)
-                    {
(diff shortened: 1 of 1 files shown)


psychocoderHPC (dismissed):

psychocoderHPC:
@felicepantaleo Thanks for your first alpaka pull-request.

sbastrakov (dismissed):
Why didn't you add yourself to the copyright note?

bernhardmgruber (dismissed):

bernhardmgruber:
This PR seems to fail compilation because of missing TBB features. I am assuming you have a fairly modern version of TBB (i.e. oneTBB)? I created a PR to update to oneTBB here: #1456.

fwyzard:
While in our production environment (CMSSW) we are indeed using v2021.4, in the `pixeltrack-standalone` test bench we are still using TBB v2020.3, _i.e._ the last tag before the breaking changes introduced in oneTBB.

Moving to oneTBB v2021.x is on the todo list...

bernhardmgruber:
> Moving to oneTBB v2021.x is on the todo list...

I ran into troubles with oneTBB 2021.1 in #1456 so I use 2021.2. Would that work for you or do you require support for an older version of oneTBB? Could we even use a newer version of oneTBB as minimum?

fwyzard:
If you switch to oneTBB 2021, I would suggest the latest one, currently v2021.4

bernhardmgruber:
We switched to oneTBB 2021.4 now: #1456. Please rebase your PR.

bernhardmgruber (approved):

psychocoderHPC (approved):

bernhardmgruber:
> Some benchmarks will follow.

@felicepantaleo I trust you that your changeset improves the TBB backend, since I cannot verify this easily. If you ever do another benchmark, please post it. I would be curious!

psychocoderHPC:
@felicepantaleo thanks for the PR, @bernhardmgruber thanks for updating the CI to support TBB versions which allow using the task arena feature.

Issue #1246, opened 2021-01-26
Introduce accelerator tags


j-stephan:
In today's meeting @psychocoderHPC brought up the idea of having special Accelerator tags that are always visible - the backends that are actually activated do not matter. This enables the user to easier specialize traits for accelerators without having to guard them with `#ifdef`.

We would like to implement this for alpaka 0.7.0. Due to time constraints this has been marked as an optional feature.

j-stephan:
I think this may also be related to #966.

bernhardmgruber:
I would like to add my proposition during the last alpaka VC, that we should just declare all backend accelerator templates upfront. This way, users can specialize their traits based on the incomplete accelerator types and no extra tag types are needed. This has the drawback, that we cannot refer to members of incomplete accelerators. Still, I would like this solution to be explored.

Another solution would be a global enum of all accelerators. Each accelerator has a getter for his enumerator and traits on the user side are specialized on these enumerators.

sbastrakov:
I think tag types are almost always same or better than enum for these purposes. Tags can  be added from anywhere, including user code, while enum is one and centralized.

psychocoderHPC:
> Another solution would be a global enum of all accelerators. Each accelerator has a getter for his enumerator and traits on the user side are specialized on these enumerators.

What @sbastrakov said. Enums can not be extended from the user code. 

I try to write a draft with my thoughts, soon.

frobnitzem:
Is this related to the enumeration of accelerator types in: https://github.com/alpaka-group/alpaka/blob/8eea24392234dc5a913f125b531ea6545c161558/example/vectorAdd/src/vectorAdd.cpp#L85

As a new alpaka user, I'm held up by the difficulty of storing `auto const devAcc = alpaka::getDevByIdx<Acc>(0u);` in my own classes.  All the alpaka documentation creates this this as "const auto".  How do I name it's storage type (`const alpaka::DevUniformCudaHipRt`) when declaring classes meant to be specialized to one device? Maybe I should just template it.

sbastrakov:
Yes, it's sometimes awkward due to alpaka operating with conditionally defined accelerator types, e.g. `AccGpuCudaRt<...>` being defined when cuda was configured to be used and found. These types (or types depending on them) are propagated basically everywhere as template parameters. Unconditionally defined tags would make it more simple and less ifdef-guardy.

[...]

PR #1453, opened 2021-11-08
Avoid locking for atomic operation on simple types

### LINKED ISSUE #1452: Mutex in atomic operation increases contention and reduces performance
When running our pixeltrack-standalone benchmark with the tbb backend we noticed that the tbb backend performance does not scale with the number of threads:

image

By profiling we realized that most of the time is spent waiting for this lock to be released:
https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/atomic/AtomicStdLibLock.hpp#L89-L90

This mutex could be replaced by atomic operations when working with simple types.

felicepantaleo:
This pull request is an attempt to solve the issue https://github.com/alpaka-group/alpaka/issues/1452 

For Trivially Copyable types, aligned with `boost::atomic_ref<T>::required_alignment`, I specialized atomic operations to avoid locking. In the case in which atomics are often used to increase a counter, this may be beneficial.

update:
image


include/alpaka/atomic/AtomicStdLibLock.hpp
```diff
@@ -81,22 +83,40 @@ namespace alpaka
...
+                if constexpr(IsLockFree)
+                {
+                    return TOp()(addr, value);
+                }
+                else
+                {
+                    std::lock_guard<std::mutex> lock(atomic.getMutex(addr));
+                    return TOp()(addr, value);
+                }
             }
```
bernhardmgruber: I wonder if we could always use `boost::atomic_ref` as implementation. I believe the cases where `boost::atomic_ref`  would not work, are probably also not going to work in CUDA.

include/alpaka/atomic/Op.hpp
```diff
@@ -24,17 +26,28 @@ namespace alpaka
...
 #    pragma GCC diagnostic ignored "-Wconversion"
 #endif
-            ref += value;
-            return old;
+                ref += value;
+                return old;
 #if BOOST_COMP_GNUC
 #    pragma GCC diagnostic pop
 #endif
+            }
```
bernhardmgruber: So this means now that if the serial backend is enabled, it will still use atomic operations, which it did not before, which is a potential pessimization. This might need some discussion, or a general architectural refactoring. It is weird to me that the `AtomicStdLibLock` implementation uses `operator()` of the passes operators, whereas the CUDA/HIP implementation just uses the passed operators for overload resolution but then provides an independent implementation. I would rather have the second architecture for all backends. But I am open for suggestions.

bernhardmgruber:
Thank you for the contribution! I think we absolutely want that patch!

For me personally, I have some concerns regarding the current architecture of the atomics implementation and how this patch affects it. I am discussing this with the alpaka team.

include/alpaka/atomic/Op.hpp
bernhardmgruber: Looking through what kind of atomics are used at which level, I am surprised to find that even the serial backend uses the lock based atomic implementation for the grid atomics.

bernhardmgruber: Still, the `AtomicNoOp` implementation calls the `operator()` of the passed `Op` which now gets an atomic implementation for the `T`s satisfying `IsLockFree`.

include/alpaka/atomic/AtomicStdLibLock.hpp
jkelling: The CUDA backend has its own atomic implementation. This one is used by `AccCpuSerial`, `AccCpuTbbBlocks`, `AccCpuFibers`, `AccCpuThreads` and at between grids by `AccCpuOmp2Blocks` and `AccCpuOmp2Threads`. it does not affect other backends.

include/alpaka/atomic/Op.hpp
jkelling: These implementations should be specializations for `AtomicStdLibLock` or better for a new atomic backend `AtomicAtomicRef`. I think this file should not be modified.

include/alpaka/atomic/AtomicStdLibLock.hpp
bernhardmgruber: Sure! But what I meant was that the implemented `if constexpr` here and inside the functors avoid `boost::atomic_ref` for `T`s which are not copyable or badly aligned. And I argued that if `T` is something that is not copyable or badly aligned, that will likely also not work with the CUDA atomics, so we could just use the boost implementation for all `T`s on the backends you listed.

fwyzard: According to the [CUDA documentation](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions)
> An atomic function performs a read-modify-write atomic operation on one 32-bit or 64-bit word residing in global or shared memory.

According to the [Boost documentation](https://www.boost.org/doc/libs/1_77_0/doc/html/atomic/interface.html#atomic.interface.interface_atomic_ref)
> The referenced object must have alignment not less than indicated by the `boost::atomic_ref<T>::required_alignment` constant. That constant may be larger than the natural alignment of type `T`.

In our test I did not find any cases where this check would fail, but I don't know what are the internals of the Boost implementation.

include/alpaka/atomic/AtomicStdLibLock.hpp
```diff
@@ -81,22 +83,40 @@ namespace alpaka
         template<typename TOp, typename T, typename THierarchy, size_t THashTableSize>
         struct AtomicOp<TOp, AtomicStdLibLock<THashTableSize>, T, THierarchy>
         {
+            static constexpr bool IsLockFree
+                = std::is_trivially_copyable_v<T> && boost::atomic_ref<T>::required_alignment <= alignof(T)
```
jkelling: This compile-time check will never fail, because `boost::atomic_ref<T>::required_alignment` is probably just `alignof(T)`.

`boost::atomic_ref<T>::required_alignment` is a requirement on `addr`, i.e. would need to be checked at runtime.

fwyzard: @jkelling wrote:
> This compile-time check will never fail, because `boost::atomic_ref<T>::required_alignment` is probably just `alignof(T)`.

According to the [Boost documentation](https://www.boost.org/doc/libs/1_77_0/doc/html/atomic/interface.html#atomic.interface.interface_atomic_ref) it is not necessarily the case:

> The referenced object must have alignment not less than indicated by the `boost::atomic_ref<T>::required_alignment` constant. That constant may be larger than the natural alignment of type `T`.

There are 9 files with over 9k lines of code inside Boost that specialise the atomic traits depending on the architecture, CPU capabilities, compiler, and actual type size - so I would not just assume that this is simply the same as `alignof(T)`.


@jkelling wrote:
> `boost::atomic_ref<T>::required_alignment` is a requirement on `addr`, i.e. would need to be checked at runtime.

We can assume that a variable of type `T` is aligned at least as much as `alignof(T)`, so if the compile time check succeeds, we don't need a runtime check.

What one could do is to add a second branch with a runtime check for the case where the compile time check fails (_i.e._ `boost::atomic_ref<T>::required_alignment` is larger than `alignof(T)`) but the actual variable is still suitably aligned (`addr` is a multiple of `boost::atomic_ref<T>::required_alignment`).

That could indeed be a further optimisation if one encounters cases where the compile time check fails.

include/alpaka/atomic/Op.hpp
psychocoderHPC: > Looking through what kind of atomics are used at which level, I am surprised to find that even the serial backend uses the lock based atomic implementation for the grid atomics.

This is required because you can have two kernel running in different async queues. In that case you must guarantee that the operations are atomic. Grid means that operations are atomic between grids.

fwyzard: @bernhardmgruber wrote:
> Looking through what kind of atomics are used at which level, I am surprised to find that even the serial backend uses the lock based atomic implementation for the grid atomics.

I understood that "grid atomics" means "atomic with respect to operations running in different grids".
Then, yes, it makes sense that the serial backend uses some atomic operations for them, if the "kernel" asks for it - because there can be multiple "grids" running concurrently in multiple host threads.

fwyzard: @jkelling wrote:
> I think this file should not be modified.

Yes, you are right, the changes should not be in `Op.hpp`.

bernhardmgruber: > This is required because you can have two kernel running in different async queues. In that case you must guarantee that the operations are atomic. Grid means that operations are atomic between grids.

Thank you, that makes perfect sense!

include/alpaka/atomic/AtomicStdLibLock.hpp
jkelling: The implementation checks the alignment at runtime using a `BOOST_ASSERT` (So only in debug mode, I think?).

While it is good to check as much as possible at compile time. This check for alignment is not reliable, because at runtime a value may be aligned to higher or lower moduli than `alignof()` (the latter should only happen if someone is doing something bad, so this is not that important). Maybe it is best to just let the backend, i.e. boost, do the check, even though this would remove the fallback to `AtomicStdLibLock` and thus just fail in these corner cases.

The boost doc says `required_alignment` may be larger than the native alignment for lock-free atomics, for types on which atomics are not lock-free it is always equal to the native alignment. It looks like there is no fallback to a locking atomic is `alignof(T) <= alignment(value) < required_alignment`. This would probably be more efficient, than using Alpaka's `AtomicStdLibLock` in this case, because there the lock is shared between all operations at any address at level of atomicity.

fwyzard: OK, we will rework this PR and write a complete replacement for `AtomicStdLibLock` (_e.g._ `AtomicBoostAtomicRef`) instead the current intrusive changes.

I would still prefer to keep
```c++
static_assert(std::is_trivially_copyable_v<T>, "Atomic operations can only be performed on variables of a trivially-copyable type.");
```
and
```c++
static_assert(boost::atomic_ref<T>::required_alignment <= alignof(T), "Atomic operations can only be performed variables of a correctly aligned type (see the documentation of boost.atomic_ref).");
```
On the other hand
  - if the `boost::atomic_ref<T>::required_alignment` does happen to be larger than `alignof(T)`, I cannot think of any workarounds, since I don't think it is possible to implement a properly-aligned type.
  - in my tests,  `boost::atomic_ref<T>::required_alignment`  was always the same as `alignof(T)`

So I'd be OK with keeping only the `trivially_copyable` check, and deferring the alignment check at runtime.

jkelling: > OK, we will rework this PR and write a complete replacement for `AtomicStdLibLock` (_e.g._ `AtomicBoostAtomicRef`) instead the current intrusive 

I left the `Boost` part out intentionally, so we can migrate it to STL without affecting anything using this implementation when alpaka moves to C++20 . Calling it `Std` would currently be wrong, hence just `AtomicAtomicRef`. How do others feel about this?

> I would still prefer to keep [... compile time alignment check]

Including `AtomicStdLibLock` as a fallback for that `alignof(T) < required_alignment<T>` case would entail keeping around that hash table of mutexes. This looks like too much cost for a case which you did encounter in 9k lines on your probably quite standard CPUs. Maybe this case does not occur for any alpaka users.

I am neutral on the compile-time check though, its negative impact is necessarily as rare as the benefit of the fallback and it avoids some, equally rare, runtime errors.

fwyzard: I did not suggest keeping the `AtomicStdLibLock` fallback, but only having a `static_assert()` on the condition.

This would still have the benefit of catching the (rare, as you imply) occurrences where `alignof(T) < required_alignment<T>`, failing with a compile time error instead of a runtime error, without any extra cost.
No fallback would be provided (apart from the user changing the atomic implementation back to `AtomicStdLibLock` by hand).

> This looks like too much cost for a case which you did encounter in 9k lines 

I'm not sure if you actually read my comments ?
9k lines of code is what Boost needs to implement all the checks and traits used by `atomic_ref` for various architectures.
I certainly did not read through all of them - I simply trust the Boost documentation when it says that an explicit check is needed before using a certain feature, rather than assuming that it is probably just not needed or that maybe the problem does not occur.

> on your probably quite standard CPUs. Maybe this case does not occur for any alpaka users.

I have no objections to removing the check altogether, if you can guarantee that it can never happen on the architectures we are interested in: x86_64, aarch64, ppc64.

jkelling: > I did not suggest keeping the `AtomicStdLibLock` fallback, but only having a `static_assert()` on the condition.

It came up in both our comments, so I wanted to address it. After I wrote I realized, though, that this table is static, so the cost would be negligible when not used, so maybe one could do it after all. Either the way you did it already using `if consexpr` or using SFINAE to have `AtomicBoostAtomicRef` become `AtomicStdLibLock` in that case.

> This would still have the benefit of catching the (rare, as you imply) occurrences where `alignof(T) < required_alignment<T>`, failing with a compile time error instead of a runtime error, without any extra cost. No fallback would be provided (apart from the user changing the atomic implementation back to `AtomicStdLibLock` by hand).

That is what I wrote. I do not object to this option.

> > This looks like too much cost for a case which you did encounter in 9k lines
> 
> I'm not sure if you actually read my comments ? 9k lines of code is what Boost needs to implement all the checks and traits used by `atomic_ref` for various architectures. I certainly did not read through all of them - I simply trust the Boost documentation when it says that an explicit check is needed before using a certain feature, rather than assuming that it is probably just not needed or that maybe the problem does not occur.

I read them, but wrote a long time after reading. You mentioned, that you did not encounter it in your code and I thought your wrote something about the size of that, so I scrolled up and saw the 9k, sorry. The number does not actually matter though. I am also not saying the boost implementation is not needed. The have support for lots of platforms in there. The questions is just how to handle this in alpaka, because it does affect portability on the one had (i.e. when we do not check) but it could also break existing code if we check and do not have a fall back. The latter Could only be dismissed by assuming, that alpaka code is not running on any of those platforms and we do not care about them. Whether we want the fallback also depends on how much we still care.

> > on your probably quite standard CPUs. Maybe this case does not occur for any alpaka users.
> 
> I have no objections to removing the check altogether, if you can guarantee that it can never happen on the architectures we are interested in: x86_64, aarch64, ppc64.

Did you test this PR on all three already?

include/alpaka/atomic/AtomicStdLibLock.hpp
BenjaminW3: I do not not think that the fallback lock version hurts us very much so I would be fine if we keep it.

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,239 @@
...
+            {
+                static_assert(std::is_trivially_copyable_v<T> && boost::atomic_ref<T>::required_alignment <= alignof(T), "Type not supported by AtomicAtomicRef, please use AtomicStdLibLock for your accelerator.");
+                boost::atomic_ref<T> ref(*addr);
+                T old = ref;
+                T result = ((old >= value) ? 0 : static_cast<T>(old + 1));
+                while(!ref.compare_exchange_weak(old, result))
+                {
+                    result = ((old >= value) ? 0 : static_cast<T>(old + 1));
+                }
+                return old;
```
bernhardmgruber, line 138: Why can´t we just do this?
```suggestion
                return ref.fetch_add(T{1});
```

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,239 @@
...
```
bernhardmgruber: Copy paste error?
```suggestion
                boost::atomic_ref<T> ref(*addr);
                T old = ref;
                T result = ((old >= value) ? 0 : static_cast<T>(old - 1));
                while(!ref.compare_exchange_weak(old, result))
                {
                    result = ((old >= value) ? 0 : static_cast<T>(old - 1));
                }
                return old;
```
And maybe this suffices:
```suggestion
                return ref.fetch_sub(T{1});
```

felicepantaleo:
There is still a race condition somewhere, I'm investigating...

felicepantaleo (commented):
I switched to `AtomicAtomicRef` just to test the PR, it is not supposed to be swapped in the final version (or at least not like this)

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,239 @@
...
+#include <boost/atomic.hpp>
+#include <type_traits>
+#include <array>
+
+namespace alpaka
+{
+    //! The CPU threads accelerator atomic ops.
+    //
+    //  Atomics can be used in the grids, blocks and threads hierarchy levels.
+    //  Atomics are not guaranteed to be save between devices.
```
fwyzard:
```suggestion
    //  Atomics are not guaranteed to be safe between devices.
```
?
It's not clear what "across multiple devices mean" for the `DevCpu` device, since there exists always only one, but these should be safe also across different cores and different sockets.

felicepantaleo: thanks, of course this is safe across cores/sockets

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,231 @@
+/* Copyright 2021 Felice Pantaleo (CERN), Andrea Bocci (CERN)
```
j-stephan: No need for the affiliation here. We will correctly reference it in the 0.9 release which will be a citable publication with you as co-authors.

felicepantaleo:
I have now access to an ARM64 and Power, I will test it during the week and update the performance plots.

include/alpaka/atomic/AtomicAtomicRef.hpp
felicepantaleo: `atomicInc` is not equivalent to `atomicAdd(&a, 1)`. There is a rollover `value` above which the counter is reset to `0`.
See: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomicinc

include/alpaka/acc/AccCpuTbbBlocks.hpp
```diff
@@ -12,6 +12,7 @@
 #ifdef ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED
 
 // Base classes.
+#    include <alpaka/atomic/AtomicAtomicRef.hpp>
 #    include <alpaka/atomic/AtomicHierarchy.hpp>
 #    include <alpaka/atomic/AtomicNoOp.hpp>
 #    include <alpaka/atomic/AtomicStdLibLock.hpp>
```
j-stephan: Is this include still required here?

include/alpaka/alpaka.hpp
```diff
@@ -28,6 +28,7 @@
 #include <alpaka/acc/AccOmp5.hpp>
 #include <alpaka/acc/Traits.hpp>
 // atomic
+#include <alpaka/atomic/AtomicAtomicRef.hpp>
 #include <alpaka/atomic/AtomicNoOp.hpp>
 #include <alpaka/atomic/AtomicOmpBuiltIn.hpp>
 #include <alpaka/atomic/AtomicStdLibLock.hpp>
```
j-stephan: Also here.

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,231 @@
...
+#include <alpaka/core/BoostPredef.hpp>
+
+#include <boost/atomic.hpp>
+
+#include <array>
+#include <type_traits>
+
+namespace alpaka
+{
+    //! The CPU threads accelerator atomic ops based on atomic_ref.
```
j-stephan: Isn't this also for other accelerators (TBB)?

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,231 @@
...
+    {
+        //! The CPU threads accelerator AtomicAdd.
+        template<typename T, typename THierarchy>
+        struct AtomicOp<AtomicAdd, AtomicAtomicRef, T, THierarchy>
+        {
+            ALPAKA_FN_HOST static auto atomicOp(AtomicAtomicRef const& atomic, T* const addr, T const& value) -> T
+            {
+                static_assert(
+                    std::is_trivially_copyable_v<T> && boost::atomic_ref<T>::required_alignment <= alignof(T),
+                    "Type not supported by AtomicAtomicRef, please use AtomicStdLibLock for your accelerator.");
```
j-stephan: Are we able to use other atomics than specified in the atomic hierarchy? Naively I would assume you would need to switch to another accelerator entirely.

j-stephan (changes requested):

include/alpaka/atomic/AtomicAtomicRef.hpp
bernhardmgruber: Ah ok. I naively assumed a different behavior. I should check the docs next time. Thx!

include/alpaka/atomic/AtomicAtomicRef.hpp
fwyzard: In principle it can be for any CPU-based accelerator (threads, fibers, OpenMP 2); we have tested it only for the serial and TBB ones.
Maybe we should amend the comment to
```suggestion
    //! The CPU accelerator atomic ops based on atomic_ref.
```
(here and in `include/alpaka/atomic/AtomicStdLibLock.hpp`)

include/alpaka/atomic/AtomicAtomicRef.hpp
fwyzard: Not that we could find: the hierarchy is defined by the accelerator, and is not templated on the data type.
The comment is supposed to suggest that the user has run into a corner case not supported by `boost::atomic_ref`, and should indeed modify the accelerator to use `AtomicStdLibLock` instead of `AtomicAtomicRef`.

I think it could be done adding something like
```c++
        public AtomicHierarchy<
#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
            AtomicAtomicRef, // grid atomics
            AtomicAtomicRef, // block atomics
#else
            AtomicStdLibLock<16>, // grid atomics
            AtomicStdLibLock<16>, // block atomics
#endif  // ALPAKA_DISABLE_ATOMIC_ATOMICREF
            AtomicNoOp         // thread atomics
        >,
```
to all the affected accelerators, and change the comment here to suggest setting `-DALPAKA_DISABLE_ATOMIC_ATOMICREF` to fall back to `AtomicStdLibLock`.

Or, since that involves repeating the same `#ifdef` in many files and in multiple places per file, we could add it to a single place in a new file:
`include/alpaka/atomic/AtomicCpu.hpp`:
```c++
#pragma once

#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
#include <alpaka/atomic/AtomicAtomicRef.hpp>
#else
#include <alpaka/atomic/AtomicStdLibLock.hpp>
#endif  // ALPAKA_DISABLE_ATOMIC_ATOMICREF

namespace alpaka {

#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
  using AtomicCpu = AtomicAtomicRef;
#else
  using AtomicCpu = AtomicStdLibLock<16>;
#endif  // ALPAKA_DISABLE_ATOMIC_ATOMICREF

}  // namespace alpaka
```

What do you think ?

felicepantaleo: I agree with this latter solution if someone really needs fallbacks. This might be needed just for the atomicCAS of for some embedded software applications that deal with packed/non-aligned data structures. 
If you agree, I will integrate this proposed solution in the PR.

include/alpaka/atomic/AtomicAtomicRef.hpp
j-stephan: Yes, that would be more fitting.

include/alpaka/atomic/AtomicAtomicRef.hpp
j-stephan: I don't see problem with the latter approach (for now - in the future this would need to be solved by accelerator tags and policies. But this requires a major refactoring on accelerators). Feel free to give it a try :-)

psychocoderHPC: I am not sure if `#ifdef` is the right way. Control how an accelerator behaves should be managed via policies but to implement it somehow clean IMO [accelerator tags](https://github.com/alpaka-group/alpaka/issues/1246) are required.

I do not see that we will get a non preprocessor way ready soon therefore I agree with the second solution :-(

felicepantaleo:
I have rerun the tests with the latest version of this PR.
Serial means that we are running one copy of the workflow per thread. 
In tbb instead only one workflow is executed and parallelization happens only within the kernels.
image

include/alpaka/atomic/AtomicAtomicRef.hpp
j-stephan: This still needs to be renamed.

include/alpaka/atomic/AtomicCpu.hpp
```diff
@@ -0,0 +1,26 @@
...
+
+namespace alpaka
+{
+#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
+    using AtomicCpu = AtomicAtomicRef;
+#else
+    using AtomicCpu = AtomicStdLibLock<16>;
+#endif // ALPAKA_DISABLE_ATOMIC_ATOMICREF
+
+} // namespace alpaka
```
j-stephan: There is a missing newline here.

j-stephan (changes requested):
Thanks! Almost good now, two minor remarks below.

This PR certainly improves performance, but it looks like there is still room for improvement. Could you add a plot similar to the one in #1452?

felicepantaleo:
> Thanks! Almost good now, two minor remarks below.
> 
> This PR certainly improves performance, but it looks like there is still room for improvement. Could you add a plot similar to the one in #1452?

Here it is (on the right)
image

include/alpaka/atomic/AtomicCpu.hpp
felicepantaleo: would something like this work?
 `The atomic ops based on atomic_ref for CPU accelerators.`

include/alpaka/atomic/AtomicCpu.hpp
```diff
@@ -0,0 +1,26 @@
...
+#    include <alpaka/atomic/AtomicStdLibLock.hpp>
+#endif // ALPAKA_DISABLE_ATOMIC_ATOMICREF
+
+namespace alpaka
+{
+#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
+    using AtomicCpu = AtomicAtomicRef;
+#else
+    using AtomicCpu = AtomicStdLibLock<16>;
+#endif // ALPAKA_DISABLE_ATOMIC_ATOMICREF
```
bernhardmgruber, line 24: Out of curiosity: What kind of scenarios require defining `ALPAKA_DISABLE_ATOMIC_ATOMICREF`? When is this escape hatch needed? I thought that the `atomic_ref` based implemented could fully replace the lock based one.

Consider also adding this explanation as a comment.

bernhardmgruber (commented):
Please rebase on the latest develop branch since we just merged C++17 support a couple of days ago. This should get the CI working.

include/alpaka/atomic/AtomicCpu.hpp
fwyzard: As discussed earier in the thread with @jkelling , the [documentation of `boost::atomic_ref`](https://www.boost.org/doc/libs/1_77_0/doc/html/atomic/interface.html#atomic.interface.interface_atomic_ref) says

> The referenced object must have alignment not less than indicated by the `boost::atomic_ref<T>::required_alignment` constant. That constant may be larger than the natural alignment of type `T`. In **Boost.Atomic**, `required_alignment` indicates the alignment at which operations on the object are lock-free; otherwise, if lock-free operations are not possible, `required_alignment` shall not be less than the natural alignment of `T`. 

This is checked at compile time [in each atomic function](https://github.com/alpaka-group/alpaka/pull/1453/files#diff-dccee67d3667c58da43cdba0ebb326a42f05d6fb3b877c3709c6a5491917ebedR39-R42), with
```c++
                static_assert(
                    std::is_trivially_copyable_v<T> && boost::atomic_ref<T>::required_alignment <= alignof(T),
                    "Type not supported by AtomicAtomicRef, please recompile defining "
                    "ALPAKA_DISABLE_ATOMIC_ATOMICREF.");
```
and at runtime by `boost::atomic_ref` itself.

We have not encountered any situation or any POD or non-POD type for which the check failed, but we cannot exclude it could happen on some combination of compiler and architecture.

bernhardmgruber: > We have not encountered any situation or any POD or non-POD type for which the check failed

That's the information I was looking for. Given that alpaka atomics are only implemented/tested on the builtin integral and floating point types, I was wondering whether `boost::atomic_ref` fails for anyone of those. Hypothetically, alpaka could support atomics on any kind of type if the user specializes the right templates. So yes, the backend could fail to compile. Whether that is likely or a good idea at all is a different question.

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -0,0 +1,242 @@
...
+        //! The CPU accelerators AtomicAdd.
+        template<typename T, typename THierarchy>
+        struct AtomicOp<AtomicAdd, AtomicAtomicRef, T, THierarchy>
+        {
+            ALPAKA_FN_HOST static auto atomicOp(AtomicAtomicRef const& atomic, T* const addr, T const& value) -> T
+            {
+                static_assert(
+                    std::is_trivially_copyable_v<T> && boost::atomic_ref<T>::required_alignment <= alignof(T),
+                    "Type not supported by AtomicAtomicRef, please recompile defining "
+                    "ALPAKA_DISABLE_ATOMIC_ATOMICREF.");
```
bernhardmgruber: Could you please refactor that check into a separate function to eliminate some of the redundancy in the following functions?
```c++
template <typename T>
void isSupportedByAtomicAtomicRef() {
    static_assert(
        std::is_trivially_copyable_v<T> && boost::atomic_ref<T>::required_alignment <= alignof(T),
        "Type not supported by AtomicAtomicRef, please recompile defining "
        "ALPAKA_DISABLE_ATOMIC_ATOMICREF.");
}
```
```suggestion
                isSupportedByAtomicAtomicRef<T>();
```

fwyzard:
I think the CIs are failing because `boost::atomic_ref` requires Boost 1.73 (1.74 would be better as it includes some bugfixes to it), while the CI configuration requires Boost 1.65.1 .

Shall we make Boost 1.74 a new requirement ?
Or leave it optional, and fall back to `AtomicStdLibLock` if a recent enough version of Boost is not available ?

include/alpaka/atomic/AtomicCpu.hpp
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
```
fwyzard, line 11: I have not tested it, but this should work:
```suggestion

#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
#    if BOOST_VERSION < 107300
#        warning boost::atomic_ref requires Boost version 1.73 or higher. Please update your version of Boost, or disable the use of boost::atomic_ref with -DALPAKA_DISABLE_ATOMIC_ATOMICREF
#    endif
#    define ALPAKA_DISABLE_ATOMIC_ATOMICREF
#endif

```

You could make CMake set `-DALPAKA_DISABLE_ATOMIC_ATOMICREF` if the version of Boost is not high enough - but I have no idea how to do that :-/

psychocoderHPC:
> I think the CIs are failing because `boost::atomic_ref` requires Boost 1.73 (1.74 would be better as it includes some bugfixes to it), while the CI configuration requires Boost 1.65.1 .
> 
> Shall we make Boost 1.74 a new requirement ? Or leave it optional, and fall back to `AtomicStdLibLock` if a recent enough version of Boost is not available ?

Boost is released August 14th, 2020 05:02 GMT. @alpaka-group/alpaka-maintainers what do you think? The current latest stable boos release is 1.77.0. Maybe we should talk about this topic in the today's developer meeting.

bernhardmgruber:
> Shall we make Boost 1.74 a new requirement ?

@j-stephan suggested that to me a while back (cannot find the comment at the moment). I am generally in favor.

include/alpaka/atomic/AtomicCpu.hpp
j-stephan: Since we now require Boost 1.74 this change is no longer necessary.

j-stephan (approved):
Looks good to me.

j-stephan (changes requested):
Please rebase on the current `develop`. Let us hope this fixes the CI issues...

include/alpaka/atomic/AtomicCpu.hpp
fwyzard: OK, so nothing is needed on the CMake side.

But there is no check anywhere else in the code that `BOOST_VERSION` is 1.74 or greater, so I think it would be useful to include this check, at least for the people that do not use CMake.

fwyzard:
Rebased on top of the current `develop` branch in #1566.

j-stephan:
Since #1566 seems to be in a mergeable state: can this be closed?

fwyzard:
I would say so, but let's wait for @felicepantaleo  to be back.

felicepantaleo:
Sure, sorry for the delay.

Issue #716, opened 2018-12-27
Test multiple TBB versions in CI


BenjaminW3:
Add a new variable `ALPAKA_TBB_VERSION` into the `.travis.yml` file and use it in `install_tbb.sh`. This should be easy for windows because there it is already downloaded manually but might require some changes on linux because we are currently using the system tbb there.

PR #1456, opened 2021-11-09
Update to oneTBB 2021.4.0

### LINKED ISSUE #1424: TBB back-end needs to be migrated to oneTBB
Since TBB is now also under the oneAPI umbrella some interfaces have changed / were deprecated. We should update the back-end to be compatible with both versions until the old style TBB is no longer required.

https://docs.oneapi.io/versions/latest/onetbb/tbb_userguide/Migration_Guide.html

bernhardmgruber:
This PR updates to oneTBB 2021.4.0.


CMakeLists.txt
```diff
@@ -69,9 +69,6 @@ include(${_ALPAKA_ADD_EXECUTABLE_FILE})
 set(_ALPAKA_ADD_LIBRARY_FILE "${_ALPAKA_ROOT_DIR}/cmake/addLibrary.cmake")
 include(${_ALPAKA_ADD_LIBRARY_FILE})
 
-# Add module search path
-set(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} "${_ALPAKA_ROOT_DIR}/cmake/modules/")
-
 # parse environment variable `CMAKE_MODULE_PATH`
 if(DEFINED ENV{CMAKE_MODULE_PATH})
     string(REPLACE ":" ";" ENV_CMAKE_MODULE_PATH $ENV{CMAKE_MODULE_PATH})
```
bernhardmgruber: I don't know if we still need this. I think I saw a reference to HIP somewhere? Does HIP also need a CMake module?

script/install_tbb.sh
```diff
@@ -17,14 +17,22 @@ source ./script/set.sh
...
-    travis_retry sudo apt-get -y --quiet --allow-unauthenticated --no-install-recommends install libtbb-dev
+    travis_retry sudo apt-get -qqq update
+    travis_retry sudo apt-get install -y wget build-essential pkg-config cmake ca-certificates gnupg
+    travis_retry sudo wget https://apt.repos.intel.com/intel-gpg-keys/GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
+
+    sudo apt-key add GPG-PUB-KEY-INTEL-SW-PRODUCTS.PUB
+    travis_retry sudo add-apt-repository "deb https://apt.repos.intel.com/oneapi all main"
+
+    travis_retry sudo apt-get update
+    travis_retry sudo apt-get install -y intel-oneapi-tbb-devel-2021.2.0
```
j-stephan: The version number does not match the title of the PR.

j-stephan (changes requested):
Since you are touching the TBB installation already: Do you think you can fix #716 along the way?

script/install_tbb.sh
bernhardmgruber: Thx! I had to change it after creating the PR, because cmake could not find 2021.1.1 yet. I think the full cmake support came in 2021.2.0.

CMakeLists.txt
j-stephan: ROCm comes with its own `FindHIP.cmake` nowadays and a user should point to its location on the command line. IMO this can be removed, too.

bernhardmgruber: So users will never define an environment variable `CMAKE_MODULE_PATH` to inject `FindHIP.cmake`?

bernhardmgruber: Ok, removed.

j-stephan: To expand a bit on this: The user should provide the correct path by setting one of the following:

1. The `<PackageName>_ROOT` environment variable / CMake variable.
2. The `CMAKE_PREFIX_PATH` CMake variable.
3. The `<PackageName>_DIR` environment variable.
4. The user package registry - [Link](https://cmake.org/cmake/help/v3.18/manual/cmake-packages.7.html#user-package-registry).

The top overrides the bottom, i.e. `<PackageName>_ROOT` will override anything else, `CMAKE_PREFIX_PATH` will override `<PackageName>_DIR` and the package registry, and so on.

bernhardmgruber: thx for the explanation!

bernhardmgruber:
I really don't understand why the variable `ALPAKA_CI_TBB_VERSION` from the `ci.yml` does not arrive in the `install_tbb.sh` script. Does anyone have an idea?

.github/workflows/ci.yml
```diff
@@ -88,417 +89,10 @@ jobs:
...
-        - name: linux_gcc-10_release_oacc
-          os: ubuntu-latest
-          env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 10,       ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.74.0, ALPAKA_CI_CMAKE_VER: 3.20.0, OMP_NUM_THREADS: 2, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", CMAKE_CXX_FLAGS: "-foffload=disable", ALPAKA_ACC_ANY_BT_OACC_ENABLE: ON, ALPAKA_OFFLOAD_MAX_BLOCK_SIZE: 1, ACC_DEVICE_TYPE: "host", ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE: OFF, ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE: OFF}
         - name: linux_gcc-11_debug
           os: ubuntu-latest
           env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 11,       ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.76.0, ALPAKA_CI_CMAKE_VER: 3.21.1, OMP_NUM_THREADS: 2, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04"}
```
j-stephan: You probably want to set the TBB version here as well, otherwise the CI job won't know about it.

bernhardmgruber: Discussed offline, the job should inherit the environment form the matrix.

bernhardmgruber:
> I really don't understand why the variable `ALPAKA_CI_TBB_VERSION` from the `ci.yml` does not arrive in the `install_tbb.sh` script. Does anyone have an idea?

Found it. We need to specify all variables from the Github actions environment again for the docker image.

bernhardmgruber:
Turns out the Intel apt repository is unresponsive quite often, causing a lot of surious CI failures.

j-stephan:
I think I saw a guide somewhere which tells you how to cache binaries (and other stuff). Maybe we should look at that in the future.

bernhardmgruber:
> I think I saw a guide somewhere which tells you how to cache binaries (and other stuff). Maybe we should look at that in the future.

I already tried that for the CUDA installer once, thinking it would make the CI faster. But actually, restoring the cache took longer than downloading the CUDA installer again. So I am afraid that would slow down the CI even more. It might solve the unresponsiveness of the Intel apt repo though. Hmm.

bernhardmgruber:
It seems we cannot combine nvcc 11.1-11.4, g++-9 and oneTBB since nvcc does not support some gcc builtins that are somehow pulled in via avx512 intrinsics.

bernhardmgruber:
It turns out that the gcc builtin was removed in g++-10 and also libstdc++ from g++-10 is no longer using it. Could it be that nvcc in the failing builds uses a different host compiler than the CXX compiler and libstdc++?

bernhardmgruber:
It seems we somehow get conversion errors now in the OMP atomics:
```
/home/runner/work/alpaka/alpaka/include/alpaka/atomic/AtomicOmpBuiltIn.hpp:42:17: error: conversion from 'int' to 'short int' may change value [-Werror=conversion]
   42 | #        pragma omp atomic capture
      |                 ^~~
```

.github/workflows/ci.yml
```diff
@@ -190,25 +192,25 @@ jobs:
...
           os: ubuntu-latest
           env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 11,     ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.19.7, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", CMAKE_CXX_FLAGS: "-fopenmp=libomp -fopenmp-targets=x86_64-pc-linux-gnu -Wno-openmp-mapping", ALPAKA_ACC_ANY_BT_OMP5_ENABLE: ON, ALPAKA_OFFLOAD_MAX_BLOCK_SIZE: 1, CMAKE_EXE_LINKER_FLAGS: "-fopenmp"}
         - name: linux_clang-12_release
           os: ubuntu-latest
-          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 12,     ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.76.0, ALPAKA_CI_CMAKE_VER: 3.21.1, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", CMAKE_CXX_EXTENSIONS: OFF}
+          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 12,     ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.76.0, ALPAKA_CI_CMAKE_VER: 3.21.1, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF, CMAKE_CXX_EXTENSIONS: OFF}
```
bernhardmgruber, line 213: TBB is turned off here, because it is not compatible with libc++ (would need to compile TBB with libc++).

.github/workflows/ci.yml
```diff
@@ -125,7 +127,7 @@ jobs:
         ### Windows
         - name: windows_cl-2019_release
           os: windows-2019
-          env: {CXX: cl.exe,  CC: cl.exe, ALPAKA_CI_CL_VER: 2019,                                   CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.70.0, ALPAKA_CI_CMAKE_VER: 3.20.0, OMP_NUM_THREADS: 1,                                                                                                                                                                                           ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF}
+          env: {CXX: cl.exe,  CC: cl.exe, ALPAKA_CI_CL_VER: 2019,                                   CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.70.0, ALPAKA_CI_CMAKE_VER: 3.20.0, OMP_NUM_THREADS: 1,                                                                                                                                                                                           ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF}
```
bernhardmgruber, line 130: Enable TBB here, so we have a non-nvcc windows build.

.github/workflows/ci.yml
```diff
@@ -97,7 +99,7 @@ jobs:
           env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 8,        ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.66.0, ALPAKA_CI_CMAKE_VER: 3.20.0, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_CI_ANALYSIS: ON, ALPAKA_DEBUG: 2}
         - name: linux_clang-8_debug_analysis
           os: ubuntu-latest
-          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 8,      ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.20.0, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_CI_ANALYSIS: ON, ALPAKA_DEBUG: 2}
+          env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 8,      ALPAKA_CI_STDLIB: libc++,    CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.73.0, ALPAKA_CI_CMAKE_VER: 3.20.0, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_CI_ANALYSIS: ON, ALPAKA_DEBUG: 2, ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF}
```
bernhardmgruber, line 102: libc++ not compatible with TBB.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
```diff
@@ -42,7 +42,7 @@ namespace alpaka
 #        pragma omp atomic capture
                 {
                     old = ref;
-                    ref += value;
+                    ref = static_cast<T>(ref + value);
```
bernhardmgruber: For some reason, including TBB suddenly generated warnings here, which were turned into errors.

script/travis_retry.sh
```diff
@@ -28,18 +28,19 @@ travis_retry() {
...
     "$@"
     result=$?
     [ $result -eq 0 ] && break
     count=$((count + 1))
     sleep 1
   done
-  [ $count -gt 3 ] && {
-    echo -e "\n${ANSI_RED}The command \"$*\" failed 3 times.${ANSI_RESET}\n" >&2
+  [ $count -gt $max ] && {
+    echo -e "\n${ANSI_RED}The command \"$*\" failed $max times.${ANSI_RESET}\n" >&2
```
bernhardmgruber, line 43: I was surprised this actually worked. The Intel apt repository is sometimes not available and `sudo apt add-repository` fails after 3 attempts. But trying more often actually works and does not noticably slow down the CI. I am open to have a different `max` here, but `3` is definitely too little :)

.github/workflows/ci.yml
```diff
@@ -342,9 +344,9 @@ jobs:
...
-        - name: linux_nvcc-11.3_gcc-9_release
+        - name: linux_nvcc-11.3_gcc-9_debug
           os: ubuntu-latest
-          env: { CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 9,        ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.69.0, ALPAKA_CI_CMAKE_VER: 3.19.7,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CI_CUDA_VERSION: "11.3", CMAKE_CUDA_COMPILER: nvcc,                                              ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF }
+          env: { CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 9,        ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.69.0, ALPAKA_CI_CMAKE_VER: 3.19.7,                     ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:18.04", ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CI_CUDA_VERSION: "11.3", CMAKE_CUDA_COMPILER: nvcc,                                              ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF }
```
bernhardmgruber, line 349: All other nvcc/g++ builds are release, so I changed one to debug to also test that. I think I wanted to try something because of the failing AVX512 intrinsics in release mode. It is unrelated to adding TBB, but I would keep it.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
psychocoderHPC: I am not sure if a cast within the capture is allowed. It is not defined by the standard. IMO only a binary operator is allowed.
https://www.openmp.org/spec-html/5.0/openmpsu95.html

psychocoderHPC (changes requested):
I can not merge this change because I do not know if this and other changes of the OpenMP atomics are standard conform.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
psychocoderHPC: IMO only

```
{ v = x; x binop= expr; } 
{ x binop= expr; v = x; } 
{ v = x; x = x binop expr; } 
{ v = x; x = expr binop x; } 
{ x = x binop expr; v = x; } 
{ x = expr binop x; v = x; } 
{ v = x; x = expr; } 
{ v = x; x++; } 
{ v = x; ++x; } 
{ ++x; v = x; } 
{ x++; v = x; } 
{ v = x; x--; } 
{ v = x; --x; } 
{ --x; v = x; } 
{ x--; v = x; }
```

Is allowed.

script/travis_retry.sh
psychocoderHPC: An ideal magic number would be `42` ^^

bernhardmgruber:
> I can not merge this change because I do not know if this and other changes of the OpenMP atomics are standard conform.

I see. Then I have to disable the warnings locally.

include/alpaka/atomic/AtomicOmpBuiltIn.hpp
bernhardmgruber: Replaced by pragmas to disable the warnings.

j-stephan (approved):

j-stephan:
Merging is blocked until @psychocoderHPC gives his approval.

psychocoderHPC (approved):

PR #1457, opened 2021-11-09
Simplify math trait specializations using standard library

### LINKED ISSUE #1414: Some `alpaka::math::MathStdLib` functions trigger `[-Werror,-Wdouble-promotion]` when not called with `double`

sbastrakov:
Merge most of them with the general ADL trait implementations by adding `using std::[math_function];` there. Exceptions are `max`, `min`, `rsqrt` which had custom logic in specialization, it is unchanged. A custom specialization is still possible with the old way, just now mostly unnecessary. Implements [this suggestion](https://github.com/alpaka-group/alpaka/issues/1414#issuecomment-963962497).
It should help with complex number math implementations, and maybe also solve some issues with ambiguity leading to compile-time errors or unnecessary casts.

It should probably be possible to to the same for CUDA/HIP. That requires that `std::` versions are callable from device side, which should be the case for functions with `constexpr` versions. However this may require more checks and passes for CI. Also we would need to check that no unnecessary casts are done and correct precision is used. So this PR only does the CPU side.

### DIFF
--- a/include/alpaka/math/abs/AbsStdLib.hpp
+++ b/include/alpaka/math/abs/AbsStdLib.hpp
@@ -9,37 +9,15 @@
 
 #pragma once
 
-#include <alpaka/core/Unused.hpp>
 #include <alpaka/math/abs/Traits.hpp>
-
-#include <cmath>
-#include <cstdlib>
-#include <type_traits>
 
 namespace alpaka
 {
     namespace math
     {
-        //! The standard library abs.
+        //! The standard library abs, implementation covered by the general template.
         class AbsStdLib : public concepts::Implements<ConceptMathAbs, AbsStdLib>
         {
         };
-
-        namespace traits
-        {
-            //! The standard library abs trait specialization.
-            template<typename TArg>
-            struct Abs<
-                AbsStdLib,
-                TArg,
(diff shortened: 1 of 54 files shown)


bernhardmgruber (dismissed):

sbastrakov:
I will set it to draft for now, until i'm finished with complex number math unit tests based on this PR and my complex branch. This is to avoid changing something twice or reverting the change in next PR.

bernhardmgruber (approved):

sbastrakov:
The second commit should be a bugfix for previously missing ADL for `rsqrt` and `sincos`. However i am not fully sure why did the tests work for those. Are the tests also buggy there? The issue became apparent in my complex branch

psychocoderHPC (approved):

PR #1458, opened 2021-11-09
[on hold] Rewrite tests and examples to use accessors

bernhardmgruber:
This PR originated as part of #1249 and rewrites all alpaka examples and tests to use the new accessors facility.

### DIFF
--- a/test/integ/axpy/src/axpy.cpp
+++ b/test/integ/axpy/src/axpy.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Axel Huebl, Benjamin Worpitz, Matthias Werner
+/* Copyright 2019-2021 Axel Huebl, Benjamin Worpitz, Matthias Werner, Bernhard Manfred Gruber
  *
  * This file is part of alpaka.
  *
@@ -30,18 +30,17 @@ public:
     //! \tparam TAcc The type of the accelerator the kernel is executed on..
     //! \tparam TElem The matrix element type.
     //! \param acc The accelerator the kernel is executed on.
-    //! \param numElements Specifies the number of elements of the vectors X and Y.
     //! \param alpha Scalar the X vector is multiplied with.
     //! \param X Vector of at least n elements.
     //! \param Y Vector of at least n elements.
     ALPAKA_NO_HOST_ACC_WARNING
-    template<typename TAcc, typename TElem, typename TIdx>
+    template<typename TAcc, typename TMemoryHandleX, typename TMemoryHandleY, typename TElem, typename TIdx>
     ALPAKA_FN_ACC auto operator()(
         TAcc const& acc,
-        TIdx const& numElements,
         TElem const& alpha,
-        TElem const* const X,
-        TElem* const Y) const -> void
+        alpaka::experimental::Accessor<TMemoryHandleX, TElem, TIdx, 1, alpaka::experimental::ReadAccess> const X,
+        alpaka::experimental::Accessor<TMemoryHandleY, TElem, TIdx, 1, alpaka::experimental::ReadWriteAccess> const Y)
+        const -> void
     {
         static_assert(alpaka::Dim<TAcc>::value == 1, "The AxpyKernel expects 1-dimensional indices!");
 
@@ -49,6 +48,7 @@ public:
         auto const threadElemExtent = alpaka::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc)[0u];
         auto const threadFirstElemIdx = gridThreadIdx * threadElemExtent;
 
+        auto const numElements = X.extents[0];
         if(threadFirstElemIdx < numElements)
         {
             // Calculate the number of elements to compute in this thread.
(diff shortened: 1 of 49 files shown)


SimeonEhrig:
I see, some examples are not ported to accessors. I'm not sure, if this was intention or only missing, because some examples was added in the mean time. Nevertheless, I for it, that the example `randomCells2D` should not be ported, because it demonstrate the usage of pointers as kernel arguments and the usage of the pitch.

test/integ/axpy/src/axpy.cpp
```diff
@@ -166,10 +166,9 @@ TEMPLATE_LIST_TEST_CASE("axpy", "[axpy]", TestAccs)
     auto const taskKernel = alpaka::createTaskKernel<Acc>(
         workDiv,
         kernel,
-        numElements,
         alpha,
-        alpaka::getPtrNative(memBufAccX),
-        alpaka::getPtrNative(memBufAccY));
+        alpaka::experimental::readAccess(memBufAccX),
```
j-stephan, line 170: clang complains about this when in C++17 mode:

```
/usr/lib64/gcc/x86_64-pc-linux-gnu/11.1.0/../../../../include/c++/11.1.0/type_traits:2933:5: error: no type named 'type' in 'std::invoke_result<AxpyKernel, const alpaka::AccCpuSerial<std::integral_constant<unsigned long, 1>, unsigned long> &, const unsigned long &, const float &, const alpaka::experimental::Accessor<float *, float, unsigned long, 1, alpaka::experimental::ReadAccess> &, const alpaka::experimental::Accessor<float *, float, unsigned long, 1, alpaka::experimental::ReadWriteAccess> &>'
    using invoke_result_t = typename invoke_result<_Fn, _Args...>::type;
    ^~~~~
```

The accessors created here are passed by reference to the lower layers but the kernel (correctly) expects them to be passed by value.

bernhardmgruber: I can try to have a look later.

bernhardmgruber: I can build all tests and examples with clang-12 locally now.

j-stephan: DPCPP appears to be based on clang-13. Maybe something changed between clang versions?

bernhardmgruber: I guess we need to get #1485 in then.

j-stephan: I think I found the culprit:

https://github.com/alpaka-group/alpaka/blob/f9b1b7f7b37954288ad08ad905fa577ee1776c42/include/alpaka/kernel/Traits.hpp#L208-L218

Note that all arguments are transformed into a `const` reference during the check.

No idea why this works for other compilers, though.

j-stephan: Never mind. It turns out that I'm stupid and forgot to delete the `numElements` parameter from the kernel instantiation call. Everything works as expected now.

test/integ/axpy/src/axpy.cpp
```diff
@@ -30,25 +30,25 @@ class AxpyKernel
...
+        alpaka::experimental::Accessor<TMemoryHandleY, TElem, TIdx, 1, alpaka::experimental::ReadWriteAccess> const Y)
+        const -> void
     {
         static_assert(alpaka::Dim<TAcc>::value == 1, "The AxpyKernel expects 1-dimensional indices!");
 
         auto const gridThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0u];
         auto const threadElemExtent = alpaka::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc)[0u];
         auto const threadFirstElemIdx = gridThreadIdx * threadElemExtent;
 
+        auto const numElements = X.extents[0];
```
j-stephan, line 51: The `X.extents[0]` isn't part of the `Accessor` interface (it only belongs to `ViewAccessor`).

j-stephan: I wonder whether this should be obtainable through `getExtent`, too.

bernhardmgruber: IIRC, we don't have any other accessor than a ViewAccessor at the moment. And yes, we could think about accessors implementing `getExtent`. For now I went with the easier solution.

j-stephan: Well, there are my SYCL accessors ;-)

bernhardmgruber:
I would really like to abandon this work in favor of `mdspan`. Let's see how far we get with the latter.

j-stephan:
Since we now have `mdspan` and have decided to abandon accessors we can close this PR, right?

bernhardmgruber:
> Since we now have `mdspan` and have decided to abandon accessors we can close this PR, right?

In principle, yes. I will just have a look at the PR again whether there is something else that's useful.

bernhardmgruber:
Ok, the fixes for the `bufferCopy` example where done with `mdspan` as well. Looks like we can close it :)

Issue #1331, opened 2021-06-01
Windows CI broken for CUDA builds.


j-stephan:
The GitHub actions script was updated to the most recent Visual Studio version 16.10. This breaks CUDA compilation:

https://github.com/actions/virtual-environments/issues/3485

Unfortunately I don't know enough about our CI setup to fix this. It doesn't seem simple since the old image appears to be no longer available. It would be great if one of the experts on this topic could have a look. Pinging @BenjaminW3 @SimeonEhrig @psychocoderHPC

BenjaminW3:
As a fast workaround, we could take those builds out of the list of required builds. This would unblock merging of PRs until this is fixed.

j-stephan:
The bug was reported to the Visual Studio team:

https://developercommunity.visualstudio.com/t/MSVC-C142930037-with-CUDA-112-occur/1437930

j-stephan:
I have also filed a bug on NVIDIA's (non-public) bug tracker with the internal ID 3318837.

bernhardmgruber:
I read through the MSVC bugreport and nvcc fails to compile a simple hello world. It hiccups inside the MS STL. Therefore I conclude that we cannot fix this on our side and need to disable the offending combination of compilers.

psychocoderHPC:
I disabled the requirement for Windows CUDA 11.0/11.1 and 11.3. 
CUDA 10.2 is working and 10.1 should be easy to fix (I will do it).

j-stephan:
NVIDIA responded. They are aware of the issue and will include a fix in the next major CUDA release.

Until then they recommend to switch to `-std=c++17` which apparently fixes the error.

psychocoderHPC:
> NVIDIA responded. They are aware of the issue and will include a fix in the next major CUDA release.
> 
> Until then they recommend to switch to `-std=c++17` which apparently fixes the error.

Thanks for the update, I will wait for which CI tests in #1332 crash and test setting all failing windows tests to `c++17` within the same PR.

psychocoderHPC:
#1332 is now removing the tests for MSVC+nvcc CUDA 11.0 and 11.1. 
Marking tests as not required does not help that maintainer can merge the PR, I realized such behavior already before.

j-stephan:
Since we are moving to C++17 with 0.9 I guess this will automatically solve itself.

bernhardmgruber:
We discovered in #1466, that also the switch to C++17 will not solve this issue.

[...]

PR #1466, opened 2021-11-15
Switch to C++17

### LINKED ISSUE #1011: Considerations for C++17-only release
CUDA 11 will add support for C++17. This would allow us to require C++17 as minimum version for alpaka which would allow us to simplify some internals.

As an intermediate step we should create a release that still supports c++14 but already adds support for CUDA 11 and c++17.

In a follow up release we could drop support for the following dependencies because they miss full c++17 support:
* CUDA 9.0, 9.1, 9.2, 10.0, 10.1, 10.2
* clang 4
* gcc 5 and 6

bernhardmgruber:
This PR switches to C++17, including:

* Require C++17 in CMake
* Remove nvcc version below 11 from GitHub CI
* Remove CUDA SDK installer below 9.2 from install scripts
* Update supported CUDA versions in README.md
* Eliminate obsolete C++ version checks
* Remove support for nvcc + clang-5
* Update documentation
* Fix warnings
* Merge GitLab CI CUDA job description files
* Add more GitLab CI tests with clang as CUDA compiler: clang-11/cuda11.0, clang-12/cuda11.0
* Add GitHub CI test with clang-12 + nvcc 11.4
* support passing CMAKE_CUDA_FLAGS variable through CI


cmake/alpakaCommon.cmake
```diff
@@ -415,11 +415,11 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
                 endif()
             endif()
 
-            # NOTE: Since CUDA 10.2 this option is also alternatively called '--extended-lambda'
+            # NOTE: Before CUDA 10.2 this option was also alternatively called '--expt-extended-lambda'
```
j-stephan: I don't think this note is still necessary.

.github/workflows/ci.yml
```diff
@@ -130,24 +130,6 @@ jobs:
...
-        - name: windows_nvcc-10.2_cl-2019_release
-          os: windows-2019
-          env: {CXX: cl.exe,  CC: cl.exe, ALPAKA_CI_CL_VER: 2019,                                   CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.68.0, ALPAKA_CI_CMAKE_VER: 3.19.7,                     ALPAKA_ACC_GPU_CUDA_ENABLE: ON, ALPAKA_CI_CUDA_VERSION: "10.2",                                                                                                                           ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_ANY_BT_OMP5_ENABLE: OFF, ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF}
-
         ## CUDA 11.0 and CUDA 11.1 with nvcc + MSVC will not be tested because of C++17 workaround for the bug
```
j-stephan, line 133: Since we now demand C++17 we should enable these tests again.

.github/workflows/ci.yml
```diff
@@ -233,87 +215,6 @@ jobs:
           os: ubuntu-latest
           env: {CXX: icpc,    CC: icc,                                 ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Release, ALPAKA_CI_BOOST_BRANCH: boost-1.75.0, ALPAKA_CI_CMAKE_VER: 3.19.2, OMP_NUM_THREADS: 2, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04",                                                                                                                                         ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: OFF, ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE: OFF}
 
-        ## CUDA 10.0
```
j-stephan, line 236: Removing these is so satisfying.

j-stephan (changes requested):
Thanks! A few remarks:

cmake/alpakaCommon.cmake
bernhardmgruber: ok, dropped.

.github/workflows/ci.yml
bernhardmgruber: Good point! Readded.

bernhardmgruber:
It looks like we are going to need to drop clang-5 as well:
```
cd /home/runner/work/alpaka/alpaka/build/example/bufferCopy && /usr/local/cuda-11.0/bin/nvcc -forward-unknown-to-host-compiler -ccbin=/usr/bin/clang++ -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLED -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLED -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=47 -DALPAKA_CI -DALPAKA_DEBUG=0 -DALPAKA_DEBUG_OFFLOAD_ASSUME_HOST -DALPAKA_OFFLOAD_MAX_BLOCK_SIZE=256 -I/home/runner/work/alpaka/alpaka/include -isystem=/usr/local/cuda-11.0/targets/x86_64-linux/include -isystem=/home/runner/work/alpaka/alpaka/boost -isystem=/usr/local/cuda-11.0/include -g --generate-code=arch=compute_80,code=[compute_80,sm_80] --extended-lambda --expt-relaxed-constexpr -g -lineinfo -Xcompiler=-fopenmp -Xcudafe=--display_error_number -Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored -Xcompiler -pthread -std=c++17 -MD -MT example/bufferCopy/CMakeFiles/bufferCopy.dir/src/bufferCopy.cpp.o -MF CMakeFiles/bufferCopy.dir/src/bufferCopy.cpp.o.d -x cu -c /home/runner/work/alpaka/alpaka/example/bufferCopy/src/bufferCopy.cpp -o CMakeFiles/bufferCopy.dir/src/bufferCopy.cpp.o
nvcc warning : The -std=c++17 flag is not supported with the configured host compiler. Flag will be ignored.
In file included from /home/runner/work/alpaka/alpaka/example/bufferCopy/src/bufferCopy.cpp:18:
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/alpaka.hpp:18:
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/acc/AccCpuOmp2Blocks.hpp:19:
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/atomic/AtomicHierarchy.hpp:12:
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/atomic/Traits.hpp:14:
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/core/Concepts.hpp:12:
In file included from /usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/type_traits:35:
/usr/bin/../lib/gcc/x86_64-linux-gnu/7.5.0/../../../../include/c++/7.5.0/bits/c++0x_warning.h:32:2: error: This file requires compiler and library support for the ISO C++ 2011 standard. This support must be enabled with the -std=c++11 or -std=gnu++11 compiler options.
#error This file requires compiler and library support \
```

j-stephan:
Weird, looks like nvcc doesn't know that clang 5 has C++17 support. Maybe we can add a `-Xcompiler -std=c++17` here.

bernhardmgruber:
> Weird, looks like nvcc doesn't know that clang 5 has C++17 support. Maybe we can add a `-Xcompiler -std=c++17` here.

We could. clang 5 should support it. But then there is no official way to query what host compiler is used via cmake. See also: https://gitlab.kitware.com/cmake/cmake/-/issues/20901.

psychocoderHPC:
IMO clang5 should be dropped instead of adding workarounds.

j-stephan:
> IMO clang5 should be dropped instead of adding workarounds.

Should we keep it as a normal C++ compiler and just disable it for CUDA?

j-stephan:
@bernhardmgruber Please also remember to update the support matrix in `README.md`.

bernhardmgruber:
> @bernhardmgruber Please also remember to update the support matrix in `README.md`.

Did I update the support matrix incorrectly?

j-stephan:
Ah, no, you didn't. I just somehow skipped over it during my last review. My bad.

j-stephan (dismissed):

bernhardmgruber:
Switching to C++17 does not fix compilation with nvcc 11.0/11.1 and MSVC. It seems we can therefore not fix #1331 with this PR.

j-stephan:
The issue is different, though. The last time it complained about `missing (` everywhere, now it is some template instantiation error. So I guess #1331 is *technically* fixed.

j-stephan:
I'm happy to ditch CUDA 11.0 and 11.1, though. @psychocoderHPC, what do you think?

psychocoderHPC:
> I'm happy to ditch CUDA 11.0 and 11.1, though

Removing the support for CUDA 11.0/11.1 on Windows is IMO no problem.
For linux I would not remove these versions!

j-stephan:
The first few lines of `README.md` (language badge and introductory text) also still refer to C++14.

bernhardmgruber:
During the alpaka VC we decided to not add nvcc 11.0/11.1 + MSVC support again and also drop support for nvcc with clang-5 as host compiler (clang-5 stays supported without nvcc).

README.md
```diff
@@ -79,7 +79,7 @@ This library uses C++14 (or newer when available).
...
 | Boost.Fiber |:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:x:|:white_check_mark:|
 |TBB|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:white_check_mark:|:x:|
-|CUDA (nvcc)|:white_check_mark: <br/> (CUDA 9.2-11.4) |:white_check_mark: <br/> (CUDA 10.1-11.4) |:white_check_mark: <br/> (CUDA 11.0-11.4)|:x:|:x:|:white_check_mark: <br/> (CUDA 10.1-11.4)|:white_check_mark: <br/> (CUDA 11.0-11.4)|:white_check_mark: <br/> (CUDA 11.1-11.4)|:white_check_mark: <br/> (CUDA 11.4)| - |:x:|:white_check_mark: <br/> (CUDA 10.1,10.2,11.2,11.3, 11.4)|
+|CUDA (nvcc)|:white_check_mark: <br/> (CUDA 11.0-11.4) |:white_check_mark: <br/> (CUDA 11.0-11.4) |:white_check_mark: <br/> (CUDA 11.0-11.4)|:x:|:x:|:white_check_mark: <br/> (CUDA 11.0-11.4)|:white_check_mark: <br/> (CUDA 11.0-11.4)|:white_check_mark: <br/> (CUDA 11.1-11.4)|:white_check_mark: <br/> (CUDA 11.4)| - |:x:|:white_check_mark: <br/> (CUDA 11.0-11.4)|
```
j-stephan: The MSVC+CUDA entry needs to be updated here.

j-stephan (changes requested):
For some reason I cannot leave a comment inline in `before_install.sh`. Since C++17 breaks Fibers + Boost 1.65 - maybe we should raise the Boost version limit, too?

README.md
bernhardmgruber: thx!

bernhardmgruber:
> For some reason I cannot leave a comment inline in `before_install.sh`. Since C++17 breaks Fibers + Boost 1.65 - maybe we should raise the Boost version limit, too?

I am fine with raising the minimum required Boost version to 1.67. This version was released in April 2018, so is at least 3.5 years old. Sounds reasonable to me.

j-stephan (dismissed):

j-stephan:
Personally I would also be fine with raising the limit to 1.71 (found in Ubuntu 20.04, the current LTS branch). @psychocoderHPC, do we need to support older boost versions than 1.71?

j-stephan (dismissed):

bernhardmgruber:
Trying to solve the failing gitlab pipeline, I increased `CMAKE_CUDA_ARCHITECTURES` to at least 35 and also fixed a warning in `UniformCudaHip.hpp`

psychocoderHPC:
> Personally I would also be fine with raising the limit to 1.71 (found in Ubuntu 20.04, the current LTS branch). @psychocoderHPC, do we need to support older boost versions than 1.71?

I suggest removing boost ^^, but since this is not realistic I do not see a reason why we should not switch to boost 1.7.1+ if this decreases the code maintenance effort.

bernhardmgruber:
It seems that with a minimum CUDA SDK of 11.0, we can no longer use clang < 12 as CUDA compiler in the CI because it fails to find the CUDA SDK: https://gitlab.com/hzdr/crp/alpaka/-/pipelines/410102179

bernhardmgruber:
> It seems that with a minimum CUDA SDK of 11.0, we can no longer use clang < 12 as CUDA compiler.

I tried several workarounds locally suggested by @psychocoderHPC, passing `--no-cuda-version-check` and `--cuda-gpu-arch=sm_50` via `CMAKE_CUDA_FLAGS` without success.

bernhardmgruber:
I can reproduce the issue locally as well. Just run:
`CXX=clang++-9 cmake -DCMAKE_CUDA_COMPILER=clang++-9 -DALPAKA_ACC_GPU_CUDA_ENABLE=ON '-DCMAKE_CUDA_ARCHITECTURES=50;70' -DCMAKE_CUDA_FLAGS=--no-cuda-version-check ..`
With CUDA 11.0 or higher installed, it will error with:
```
CMake Error at /usr/share/cmake-3.20/Modules/CMakeDetermineCompilerId.cmake:676 (message):
  Compiling the CUDA compiler identification source file
  "CMakeCUDACompilerId.cu" failed.

  Compiler: /usr/bin/clang++-9

  Build flags:

  Id flags:
  --cuda-path=/usr/local/cuda;--cuda-gpu-arch=sm_50;--cuda-gpu-arch=sm_70 -v

  

  The output was:

  1

  clang version 9.0.1-16.1ubuntu1

  Target: x86_64-pc-linux-gnu

  Thread model: posix

  InstalledDir: /usr/bin

  Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/10

  Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/11

  Found candidate GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/9

  Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/10

  Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/11

  Found candidate GCC installation: /usr/lib/gcc/x86_64-linux-gnu/9

  Selected GCC installation: /usr/bin/../lib/gcc/x86_64-linux-gnu/11

  Candidate multilib: .;@m64

  Selected multilib: .;@m64

  Found CUDA installation: /usr/local/cuda, version 7.0

  clang: error: cannot find libdevice for sm_50.  Provide path to different
  CUDA installation via --cuda-path, or pass -nocudalib to build without
  linking with libdevice.

  clang: error: cannot find libdevice for sm_70.  Provide path to different
  CUDA installation via --cuda-path, or pass -nocudalib to build without
  linking with libdevice.

  clang: error: GPU arch sm_70 is supported by CUDA versions between 9.0 and
  10.1 (inclusive), but installation at /usr/local/cuda is 7.0.  Use
  --cuda-path to specify a different CUDA install, pass a different GPU arch
  with --cuda-gpu-arch, or pass --no-cuda-version-check.

  

  

  /usr/share/cmake-3.20/Modules/CMakeDetermineCompilerId.cmake:6 (CMAKE_DETERMINE_COMPILER_ID_BUILD)
  /usr/share/cmake-3.20/Modules/CMakeDetermineCompilerId.cmake:48 (__determine_compiler_id_test)
  /usr/share/cmake-3.20/Modules/CMakeDetermineCUDACompiler.cmake:278 (CMAKE_DETERMINE_COMPILER_ID)
  cmake/alpakaCommon.cmake:348 (enable_language)
  CMakeLists.txt:87 (include)


```

bernhardmgruber:
It seems we can also not test clang-6 with nvcc 11 in the Gitlab CI, because libomp-6-dev is not available on either Ubuntu 18.04 or 20.04. I upgrade the job to clang-7 therefore.

bernhardmgruber:
After feedback from @psychocoderHPC, I removed clang as CUDA compiler runs with clang < 11. See comment here: https://github.com/alpaka-group/alpaka/issues/1011#issuecomment-972795489.

bernhardmgruber:
@psychocoderHPC suggested to me to keep CUDA < 11 for clang as CUDA compiler, since the objective is *just* to move to C++17. I therefore readded the runs with clang and older CUDA versions.

bernhardmgruber:
OK, CI  passing, seems we are good! :)

.github/workflows/ci.yml
```diff
@@ -16,8 +16,8 @@ on: [push, pull_request]
 # NOTE: Testing the full matrix is not practical.
 # Therefore we aim to have each value been set in at lest one job.
 # CXX                                           : {g++, clang++}
-#   [g++] ALPAKA_CI_GCC_VER                     : {7, 8, 9, 10}
-#   [clang++] ALPAKA_CI_CLANG_VER               : {5.0, 6.0, 7, 8, 9, 10, 11}
+#   [g++] ALPAKA_CI_GCC_VER                     : {7, 8, 9, 10, 11}
+#   [clang++] ALPAKA_CI_CLANG_VER               : {5.0, 6.0, 7, 8, 9, 10, 11, 12}
```
j-stephan, line 20: Why not add clang 13, too?

.github/workflows/ci.yml
```diff
@@ -41,7 +41,7 @@ on: [push, pull_request]
 # ALPAKA_ACC_ANY_BT_OMP5_ENABLE                 : {ON, OFF}
 #   [ON] OMP_NUM_THREADS                        : {1, 2, 3, 4}
 # ALPAKA_ACC_GPU_CUDA_ENABLE                    : {ON, OFF}
-#   [ON] ALPAKA_CI_CUDA_VERSION                 : {9.2, 10.0, 10.1, 10.2, 11.0, 11.1, 11.2, 11.3, 11.4}
+#   [ON] ALPAKA_CI_CUDA_VERSION                 : {11.0, 11.1, 11.2, 11.3, 11.4}
```
j-stephan, line 44: Why not add CUDA 11.5, too?

README.md
```diff
@@ -92,10 +92,10 @@ Dependencies
 The **alpaka** library itself just requires header-only libraries.
 However some of the accelerator back-end implementations require different boost libraries to be built.
 
-When an accelerator back-end using *Boost.Fiber* is enabled, `boost-fiber` and all of its dependencies are required to be built in C++14 mode `./b2 cxxflags="-std=c++14"`.
-When *Boost.Fiber* is enabled and alpaka is built in C++17 mode with clang and libstc++, Boost >= 1.67.0 is required.
+When an accelerator back-end using *Boost.Fiber* is enabled, `boost-fiber` and all of its dependencies are required to be built in C++17 mode `./b2 cxxflags="-std=c++17"`.
+When *Boost.Fiber* is enabled and alpaka is built with clang and libstc++, Boost >= 1.67.0 is required.
```
j-stephan, line 96: Didn't we decide to raise the Boost version to 1.71 anyway?

j-stephan (commented):
Looking good, thanks! I'll remove the deleted CI jobs from the branch protection rules. Three questions:

jkelling:
Should todos that have been marked for the C++17 move be addressed in this PR?

There is at least this one: https://github.com/jkelling/alpaka/blob/develop/include/alpaka/rand/RandDefault.hpp#L135-L140

j-stephan:
IMHO it shouldn't. This PR just enables C++17. We can (and should) do the cleanup of the code base in a follow up PR. This is already large enough as-is.

.github/workflows/ci.yml
bernhardmgruber: I guess we should put that into a separate PR: #1475

README.md
bernhardmgruber: We discussed that, yes. But it is not necessary for moving to C++17.

.github/workflows/ci.yml
bernhardmgruber: Since I was messing with these descriptions anyway, I just updated the version here to match the jobs we already had. We should add clang-13 in a separate PR. See:  #1476.

j-stephan (approved):

README.md
bernhardmgruber: If you want to upgrade Boost as well, please create a separate issue and start a discussion.

PR #1468, opened 2021-11-15
Update version number to 0.9.0-develop

j-stephan:


include/alpaka/version.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2019 Benjamin Worpitz, Erik Zenker
+/* Copyright 2021 Benjamin Worpitz, Erik Zenker, Jan Stephan
```
sbastrakov, line 1: Why not `2019-2021` ?

j-stephan: Just following the usual pattern of the rest of alpaka here:

https://github.com/alpaka-group/alpaka/blob/cc270069756370330a6651df1876f2d4b5cdedc5/include/alpaka/acc/AccCpuTbbBlocks.hpp#L1

https://github.com/alpaka-group/alpaka/blob/cc270069756370330a6651df1876f2d4b5cdedc5/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp#L1

Etc. Although the year doesn't really matter I suppose, it is purely cosmetical.

bernhardmgruber: If a copyright header matters, then the last year we made updates to the work counts. So for me the chance is fine.

bernhardmgruber (approved):

include/alpaka/version.hpp
sbastrakov: Ah, but there are also examples of a different pattern, e.g. [here](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/alpaka.hpp)

sbastrakov: I do not have any preference, just i thought that way with a - was our standard

j-stephan: In theory, the year is no longer needed since the US changed their copyright laws a few years ago. The copyright in both Europe and the US starts from the year where a file was first published - it doesn't matter if it was changed later. So we could either remove the year completely in the future or just do some research and put back the original years. Updating the copyright year is meaningless.

Maybe we should talk about this in the next alpaka meeting.

bernhardmgruber: What really? But if I modify a source file, I create new original work, for which copyright should start from the beginning again.
That's also why it is IMO pointless to have copyright headers and just rely on your VCS to track when changes have been done.

j-stephan: See [this](https://stackoverflow.com/a/2391555) and [this](https://stackoverflow.com/a/20911485) Stack Overflow answers.

bernhardmgruber: Ok, but when we update a copyright header, git still keeps a copy of the old header around. So technically, we still have a copy of the old work with the old copyright year :) Anyway, thx for the links!

PR #1470, opened 2021-11-16
Small example and test fixes

j-stephan:
This PR fixes some minor issues I noticed while upgrading the SYCL branch:

* The Catch2 version printed out now matches the version shipped internally.
* The CMake `project()` commands for the examples and test cases now only enable `CXX` as language. The previous approach would implicitly enable `C`, too, which breaks the Intel DPC++ toolchain which comes without a C compiler.
* All subprojects now require the same CMake version as alpaka itself.

Needs to be backported to 0.8.

### DIFF
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-# Copyright 2015-2020 Benjamin Worpitz, Jan Stephan
+# Copyright 2021 Benjamin Worpitz, Jan Stephan
 #
 # This file exemplifies usage of alpaka.
 #
@@ -22,7 +22,7 @@
 
 cmake_minimum_required(VERSION 3.18)
 
-project("alpakaExamples")
+project("alpakaExamples" LANGUAGES CXX)
 
 ################################################################################
 # Add subdirectories.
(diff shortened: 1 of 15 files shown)


example/CMakeLists.txt
```diff
@@ -22,7 +22,8 @@
 
 cmake_minimum_required(VERSION 3.18)
 
-project("alpakaExamples")
+project("alpakaExamples"
+        LANGUAGES CXX)
```
bernhardmgruber: Any specific reason for the newline here? Why not:
```suggestion
project("alpakaExamples" LANGUAGES CXX)
```

bernhardmgruber (dismissed):

example/CMakeLists.txt
j-stephan: No, this has no deeper meaning. I can change it if you want.

bernhardmgruber: I prefer the single line, but I don't want to burn the CI time for it. Do as you wish.

j-stephan: Fixed. This is also more consistent with the top-level `CMakeLists.txt`.

bernhardmgruber (approved):

PR #1479, opened 2021-11-22
Add the device type trait specialization for Events

### LINKED ISSUE #1428: Add DevType<Event> trait
Hi,
for simplifying some generic code we are writing, we would benefit for traits to extract the `Device` type from the `Event` type, something that returns `TDev` for `EventGenericThreads<TDev>` and `DevUniformCudaHipRt` for `EventUniformCudaHip`.

Of course we can add such  trait in our code, but it could come in useful for other projects as well.

fwyzard:


fwyzard:
Fixes #1428 .

include/alpaka/event/EventUniformCudaHipRt.hpp
```diff
@@ -123,6 +123,12 @@ namespace alpaka
     };
     namespace traits
     {
+        //! The CUDA/HIP RT device event device type trait specialization.
+        template <>
```
BenjaminW3: clang-format says:
```suggestion
        template<>
```

BenjaminW3 (dismissed):

include/alpaka/event/EventUniformCudaHipRt.hpp
fwyzard: what version of clang do you use ?
`clang-format` from clang 11 complains about unknown options in the format file :-(

bernhardmgruber: We use clang-format 12 (specifically 12.0.1), as documented [here](https://github.com/alpaka-group/alpaka/blob/develop/CONTRIBUTING.md).

fwyzard: Fixed, thanks.

bernhardmgruber (approved):

PR #1480, opened 2021-11-23
RFC: Implement queue-ordered memory buffers

fwyzard:
Add a new type of buffer that uses non-blocking, queue-ordered memory allocations and deallocations.

So far only the `DevCpu` and `DevCudaRt` devices are supported.
Compilation with HIP raises an error, since it does not support the the underlying stream-ordered memory allocation functions.

The `DevCpu` implementation
  - allocates memory immediately;
  - frees memory when the queue associated to the buffer reaches the point in time where the destructor of the buffer is called.

The `DevCudaRt` implementation
  - allocates memory with `cudaMallocAsync`;
  - frees memory with `cudaFreeAsync`.

### DIFF
--- a/include/alpaka/alpaka.hpp
+++ b/include/alpaka/alpaka.hpp
@@ -118,6 +118,8 @@
 #include <alpaka/mem/alloc/AllocCpuAligned.hpp>
 #include <alpaka/mem/alloc/AllocCpuNew.hpp>
 #include <alpaka/mem/alloc/Traits.hpp>
+#include <alpaka/mem/buf/AsyncBufCpu.hpp>
+#include <alpaka/mem/buf/AsyncBufUniformCudaHipRt.hpp>
 #include <alpaka/mem/buf/BufCpu.hpp>
 #include <alpaka/mem/buf/BufOacc.hpp>
 #include <alpaka/mem/buf/BufOmp5.hpp>
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -196,6 +196,9 @@ namespace alpaka
     template<typename TElem, typename TDim, typename TIdx>
     class BufCpu;
 
+    template<typename TElem, typename TDim, typename TIdx>
+    class AsyncBufCpu;
+
     namespace traits
     {
         //! The CPU device memory buffer type trait specialization.
@@ -203,6 +206,12 @@ namespace alpaka
         struct BufType<DevCpu, TElem, TDim, TIdx>
         {
             using type = BufCpu<TElem, TDim, TIdx>;
+        };
+        //! The CPU device memory buffer type trait specialization.
+        template<typename TElem, typename TDim, typename TIdx>
+        struct AsyncBufType<DevCpu, TElem, TDim, TIdx>
+        {
+            using type = AsyncBufCpu<TElem, TDim, TIdx>;
         };
 
         //! The CPU device platform type trait specialization.
--- a/include/alpaka/dev/DevUniformCudaHipRt.hpp
+++ b/include/alpaka/dev/DevUniformCudaHipRt.hpp
@@ -169,6 +169,9 @@ namespace alpaka
     template<typename TElem, typename TDim, typename TIdx>
     class BufUniformCudaHipRt;
 
+    template<typename TElem, typename TDim, typename TIdx>
+    class AsyncBufUniformCudaHipRt;
+
     namespace traits
     {
         //! The CUDA/HIP RT device memory buffer type trait specialization.
@@ -176,6 +179,12 @@ namespace alpaka
         struct BufType<DevUniformCudaHipRt, TElem, TDim, TIdx>
         {
             using type = BufUniformCudaHipRt<TElem, TDim, TIdx>;
+        };
+        //! The CUDA/HIP RT stream-ordered device memory buffer type trait specialization.
+        template<typename TElem, typename TDim, typename TIdx>
+        struct AsyncBufType<DevUniformCudaHipRt, TElem, TDim, TIdx>
+        {
+            using type = AsyncBufUniformCudaHipRt<TElem, TDim, TIdx>;
         };
 
         //! The CUDA/HIP RT device platform type trait specialization.
--- a/include/alpaka/mem/buf/AsyncBufCpu.hpp
+++ b/include/alpaka/mem/buf/AsyncBufCpu.hpp
@@ -1,0 +1,293 @@
+/* Copyright 2019 Alexander Matthes, Axel Huebl, Benjamin Worpitz
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/Unused.hpp>
+#include <alpaka/core/Vectorize.hpp>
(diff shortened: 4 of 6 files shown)


fwyzard:
This PR is a draft, with two purposes:
  - to illustrate a possible development and ask for feedback about the idea;
  - to provide a concrete implementation and ask for feedback about that.

From the CMS point of view, this is an intermediate step; the final solution will likely also implement a full caching memory allocator for device and pinned host memory, based either on the original implementation in CUB (that we have forked and further developed), or on the Rapids Memory Manager (with which we don't have direct experience yet, but might be better maintained in the future).

bussmann:
I would love to see this, but as an independent package well interfaced with Alpaka. This keeps memory handling flexible and open to future changes while scope focussed. It is a strength, not a weakness of Alpaka that it is as agnostic of both memory layout and allocation strategy as possible.

See first attempts with LLAMA for memory layout and MallocMC for fast allocation. Both things are orthogonal to what Alpaka does and of value as stand alone libraries, yet interact extremely well when brought together, adding value when combined.

bussmann:
@bernhardmgruber

felicepantaleo:
@bussmann, thanks for your comment. 
Could you clarify whether you are suggesting that this memory buffer goes in an external package or that the future "full caching memory allocator" should go in an external package.
In case you mean the latter, I agree with your choice. 
In case you mean the former, it is not clear to me how you draw the line between which memory buffer goes in alpaka and which does not, and the reason why there is any memory buffer at all in alpaka.

bussmann:
Dear Felice, the role of memory buffers in Alpaka is a difficult one. At best, I would keep assumptions on memory as small as possible, same with functionality for handling memory buffers. This of course limits everyday practicality. For now, I would support a pragmatic approach and implement practical things that help.

BUT, memory is more than allocation & layout. Clearly understanding orthogonal design of task execution (Alpaka) and memory management will come over time to the project and come from learning from wrong design decisions.

My guide to decide on these things is to create categories of functionality one needs and see if they would make sense as a standalone library.

For now, this RFC sounds sensible, but I would strongly advocate to carefully design the interface for a future possibility to put this somewhere else.

For Alpaka I imagine to move out Information on memory and devices into a separate place that can then be shared with other libraries, each with a very specific, bounded focus and as little clutter as possible, thus minimizing unwamted dependencies and maintenance. This comes at the cost of needing to design and maintain interfaces. It is a shift of complexity, not its Elimination.

Hope this helps a bit clarifying my thoughts on what Alpaka should and should not do and be.

fwyzard:
About:
>  Continuous Integration / linux_clang-9_cuda-9.2_debug_analysis (pull_request) Failing after 5m — linux_clang-9_cuda-9.2_debug_analysis 

The likely reason is that `cudaMallocAsync()`/`cudaFreeAsync()` require CUDA 11.2 .

bernhardmgruber:
> The DevCpu implementation
> * allocates memory immediately;
> * frees memory when the queue associated to the buffer reaches the point in time where the destructor of the buffer is called.

I wonder whether it would be faster to just immediately free as well. And in this scenario, we could just make `BufCpuAsync` an alias of the normal `BufCpu`.

Furthermore, I wonder whether we need a strong distinction between a normal and an async buffer. It seems a lot of functionality is duplicated now, like many of the trait implementations. I think we could also implement async buffers by providing a new allocation function, but make this function just return a plain alpaka buffer. The buffer could internally keep track of whether it is async or not. We could also offer a function to ask a buffer whether it is async or not. If this information needs to be known at compile time, we could also manifest it in the template argument list of the normal buffers and dispatch inside the implementation with `if constexpr`.
I also taked to @psychocoderHPC about that briefly and he suggested that maybe an allocation policy template parameter might be a good idea.

Partly replying to @bussmann, I think creating asynchronous buffers fits well into alpaka, since alpaka is about providing abstractions from compute APIs for kernel development. Everything on top, like the proposed caching allocator, might indeed be a good fit for a library built on top of alpaka. But I have no strong opinion on this. We also have implementations of random number generators in alpaka, so, given this caching allocator is generic enough, I am fine with including it. But I guess this discussion can wait until after this PR.

fwyzard:
> I wonder whether it would be faster to just immediately free as well. And in this scenario, we could just make `BufCpuAsync` an alias of the normal `BufCpu`.

Not freeing immediately is the whole point of the new kind of buffer. The guarantee that the "device" memory stays alive until it's no longer used by the operations that have been queued allows to use the same both GPU-based and CPU-based accelerators.

> I think we could also implement async buffers by providing a new allocation function, but make this function just return a plain alpaka buffer. The buffer could internally keep track of whether it is async or not. 

Yes, this is a good point.
My first implementation required many more differences, with the `AsyncBuf` being templated on the `Queue` type.
After some iterations I managed to make only the constructor dependent on the `Queue`, making most of the implementation the same. If we can limit the differences to the constructor or allocation functions, we may not need the additional buffer types.

fwyzard:
About having the async buffers or the caching allocator in Alpaka or in a separate library, we don't mind it either way.

My only concern is that having things that rely on the "internals" of Alpaka in a separate library may be a maintenance problem - but if we do decide to go with a separate library, we can certainly find ways around it.

bussmann:
As I said, let's be pragmatic and add what is needed and later see if it can be pulled out of Alpaka with added value later. Just be careful not entangling things that should be orthogonal.

bernhardmgruber:
> Not freeing immediately is the whole point of the new kind of buffer. The guarantee that the "device" memory stays alive until it's no longer used by the operations that have been queued allows to use the same both GPU-based and CPU-based accelerators.

Ah, I get it! I thought the main point was to delay the allocation to make it faster on the host thread. But your scenario totally makes sense and probably enables some nice usage patterns where the host thread does no longer need to keep resources alive  for asynchronous kernel launches.

fwyzard:
> My first implementation required many more differences, with the `AsyncBuf` being templated on the `Queue` type.
After some iterations I managed to make only the constructor dependent on the `Queue`, making most of the implementation the same. If we can limit the differences to the constructor or allocation functions, we may not need the additional buffer types.

Thinking some more about it, I can think of two ways that would significantly reduce code duplication:
  1. implement different constructors for the `Buf` and `AsyncBuf`; this could be done kind of _implicitly_ by adding a second constructor that takes `Queue` instead of a `Device` as its first argument, or _explicitly_ by adding a "tag" parameter to the signature, or some attributes, _etc._
  2. make the constructors of the various `Buf` classes dumber, and use the external factory functions to construct them as "sync" or "async" variants

Looking the buffer implementations it's hard to say which one would be better, because different types use a different approach:
  - the constructor of `BufCpu` allocates the memory (like `std::make_shared<T>()`) by wrapping an object that inherits from `AllocCpuAligned`, that in turn handles the details of N-dimensional allocations;
  - the constructor of `BufUniformCudaHipRt` takes ownership of (device) memory that has already been allocated in `BufAlloc` (like `std::shared_ptr<T>(T*)`), and it's `BufAlloc` that takes care of the N-dimensional allocations;
  - from a quick look it seems that `BufOacc` and `BufOmp5` follow the latter approach.

If we were to change `BufCpu` to follow this latter approach (allocation are handled by `BufAlloc` and deallocation by the destructor of the `m_spBufImpl`), it would then be simple to unify the various `Buf` and `AsyncBuf` by having all differences only in the `BufAlloc` functors.

If you think this is a reasonable approach, I can make a second PR to see what it would look like.

bussmann:
I like option 2 better as it makes the buffers dumber. This way, one can add even more classes of buffers via the factory later if one likes to. Option 1 is actually two options where I would ALWAYS root for the explicit one.

fwyzard:
See #1481 for an implementation that follows the second option.

PR #1481, opened 2021-11-24
Implement queue-ordered memory buffers

### LINKED ISSUE #1448: stream-ordered memory allocations, and a caching memory allocator
Many of the algorithms in CMSSW and the split `acquire()`/`produce()` approach rely on _stream-ordered_ memory allocations:
  - device memory is allocated (from the host) in association with a CUDA stream / Alpaka queue
  - device memory is used by asynchronous operations enqueued to the same stream / queue
  - device memory is marked for freeing (from the host) while it may still be used (on the device): the "stream ordering" guarantees that it is not freed and reused until the "event" of the freeing is reached by the stream associated to the memory

This effectively means that host-side `malloc` and `free` operations on device memory are asynchronous, or
_stream-ordered_, similar to the asynchronous `memcpy`, `memset` and kernel submissions.

This approach is available in CUDA since version 11.2 via [`cudaMallocAsync`/`cudaFreeAsync`/_etc._](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html), but has been available from CUB as part of its _caching allocator_ since quite some time, and that version has been integrated and extended for CMSSW.
Surprisingly enough, the CUB-derived approach is still significantly faster than `cudaMallocAsync`/`cudaFreeAsync`.

While some work to re-implement the _caching allocator_ on top of Alpaka has started, it's not clear if that is the best
approach, or if we should aim for a different solution. In fact, while the _caching allocator_ make sense only when
using _stream-ordered_ memory operations (a large part of the cost comes from the global synchronisation when freeing
device memory), _stream-ordered_ memory operations could be implemented independently of it.

Some possible options are:
  - implement the _caching allocator_ and _stream-ordered_ memory operations on top of Alpaka buffers (current attempt)
      - only for the CUDA backend, or for all backends ?
      - the _stream-ordered_  part is needed for correctness for all non-blocking backends

  - implement the _stream-ordered_ memory operations in Alpaka
      - on top of the current Buffers ?
      - as a new AsyncBuffer type ?
      - orthogonal to the _caching allocator_ part

  - implement the _caching allocator_ only for the CUDA backend, using `cudaMalloc`/`cudaFree` directly

  - implement the _caching allocator_ and _stream-ordered_ memory operations within Alpaka, for all backends

fwyzard:
Move the implementation of the allocation and deallocation strategy from the `BufCpuImpl` and `BufUniformCudaHipRt` classes to the `AllocBuf<DevCpu>` and `AllocBuf<DevUniformCudaHipRt>` factory functors. This allows the implementation of different strategies in the `allocBuf` or other functions, without affecting the buffer classes.

Add the `allocAsyncBuf` function to allocate buffers using non-blocking, queue-ordered memory allocations and deallocations.

These functions take as argument a queue instead of a device; the memory associated with the buffer is guaranteed to be "alive" according to the lifetime of the buffer as seen by the queue: from the moment the queue reaches the buffer's constructor to the moment it reaches the destructor.

So far only the `DevCpu` and `DevCudaRt` devices are supported.
Attempting to use this with HIP raises an error, since HIP does not support the underlying stream-ordered memory allocation functions.

The `DevCpu` implementation
  - allocates memory immediately;
  - frees memory when the queue associated to the buffer reaches the point where the destructor of the buffer is called.

The `DevCudaRt` implementation
  - allocates memory with `cudaMallocAsync`, so only 1D buffers are supported;
  - frees memory with `cudaFreeAsync`.

### DIFF
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -240,6 +240,32 @@ namespace alpaka
                 return BufCpu<TElem, TDim, TIdx>(dev, memPtr, std::move(deleter), extent);
             }
         };
+        //! The BufCpu stream-ordered memory allocation trait specialization.
+        template<typename TElem, typename TDim, typename TIdx>
+        struct AsyncBufAlloc<TElem, TDim, TIdx, DevCpu>
+        {
+            template<typename TQueue, typename TExtent>
+            ALPAKA_FN_HOST static auto allocAsyncBuf(TQueue queue, TExtent const& extent) -> BufCpu<TElem, TDim, TIdx>
+            {
+                ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
+
+                static_assert(
+                    std::is_same<Dev<TQueue>, DevCpu>::value,
+                    "The BufCpu buffer can only be used with a queue on a DevCpu device!");
+                DevCpu const& dev = getDev(queue);
+
+                // alpaka::AllocCpuAligned is stateless
+                using Allocator
+                    = AllocCpuAligned<std::integral_constant<std::size_t, core::vectorization::defaultAlignment>>;
+                static_assert(std::is_empty_v<Allocator>, "AllocCpuAligned is expected to be stateless");
+                TElem* memPtr
+                    = alpaka::malloc<TElem>(Allocator{}, static_cast<std::size_t>(extent::getExtentProduct(extent)));
+                auto deleter = [queue = std::move(queue)](TElem* ptr) mutable
+                { alpaka::enqueue(queue, [ptr]() { alpaka::free(Allocator{}, ptr); }); };
+
+                return BufCpu<TElem, TDim, TIdx>(dev, memPtr, std::move(deleter), extent);
+            }
+        };
         //! The BufCpu memory mapping trait specialization.
         template<typename TElem, typename TDim, typename TIdx>
         struct Map<BufCpu<TElem, TDim, TIdx>, DevCpu>
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Alexander Matthes, Axel Huebl, Benjamin Worpitz
+/* Copyright 2021 Alexander Matthes, Axel Huebl, Benjamin Worpitz, Andrea Bocci
  *
  * This file is part of alpaka.
  *
@@ -26,6 +26,7 @@
 #include <alpaka/mem/alloc/AllocCpuAligned.hpp>
 #include <alpaka/meta/DependentFalseType.hpp>
 
+#include <functional>
 #include <memory>
 #include <type_traits>
 
@@ -36,7 +37,6 @@ namespace alpaka
         //! The CPU memory buffer.
         template<typename TElem, typename TDim, typename TIdx>
         class BufCpuImpl final
-            : public AllocCpuAligned<std::integral_constant<std::size_t, core::vectorization::defaultAlignment>>
         {
             static_assert(
                 !std::is_const<TElem>::value,
@@ -46,11 +46,15 @@ namespace alpaka
 
         public:
             template<typename TExtent>
-            ALPAKA_FN_HOST BufCpuImpl(DevCpu const& dev, TExtent const& extent)
-                : AllocCpuAligned<std::integral_constant<std::size_t, core::vectorization::defaultAlignment>>()
-                , m_dev(dev)
+            ALPAKA_FN_HOST BufCpuImpl(
+                DevCpu const& dev,
+                TElem* pMem,
+                std::function<void(TElem*)> deleter,
+                TExtent const& extent)
+                : m_dev(dev)
                 , m_extentElements(extent::getExtentVecEnd<TDim>(extent))
-                , m_pMem(alpaka::malloc<TElem>(*this, static_cast<std::size_t>(computeElementCount(extent))))
+                , m_pMem(pMem)
+                , m_deleter(std::move(deleter))
                 , m_pitchBytes(static_cast<TIdx>(extent::getWidth(extent) * static_cast<TIdx>(sizeof(TElem))))
 #if(defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_LANG_CUDA) || (defined(ALPAKA_ACC_GPU_HIP_ENABLED) && BOOST_LANG_HIP)
                 , m_bPinned(false)
@@ -82,23 +86,14 @@ namespace alpaka
                 unpin(*this);
 #endif
                 // NOTE: m_pMem is allowed to be a nullptr here.
-                alpaka::free(*this, m_pMem);
-            }
-
-        private:
-            //! \return The number of elements to allocate.
-            template<typename TExtent>
-            ALPAKA_FN_HOST static auto computeElementCount(TExtent const& extent) -> TIdx
-            {
-                auto const extentElementCount = extent::getExtentProduct(extent);
-
-                return extentElementCount;
+                m_deleter(m_pMem);
             }
 
         public:
             DevCpu const m_dev;
             Vec<TDim, TIdx> const m_extentElements;
             TElem* const m_pMem;
+            std::function<void(TElem*)> m_deleter;
             TIdx const m_pitchBytes;
 #if(defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && BOOST_LANG_CUDA) || (defined(ALPAKA_ACC_GPU_HIP_ENABLED) && BOOST_LANG_HIP)
             bool m_bPinned;
@@ -111,9 +106,10 @@ namespace alpaka
     class BufCpu
     {
     public:
-        template<typename TExtent>
-        ALPAKA_FN_HOST BufCpu(DevCpu const& dev, TExtent const& extent)
-            : m_spBufCpuImpl(std::make_shared<detail::BufCpuImpl<TElem, TDim, TIdx>>(dev, extent))
+        template<typename TExtent, typename Deleter>
+        ALPAKA_FN_HOST BufCpu(DevCpu const& dev, TElem* pMem, Deleter deleter, TExtent const& extent)
+            : m_spBufCpuImpl{
+                std::make_shared<detail::BufCpuImpl<TElem, TDim, TIdx>>(dev, pMem, std::move(deleter), extent)}
         {
         }
 
@@ -233,7 +229,15 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
-                return BufCpu<TElem, TDim, TIdx>(dev, extent);
+                // alpaka::AllocCpuAligned is stateless
+                using Allocator
+                    = AllocCpuAligned<std::integral_constant<std::size_t, core::vectorization::defaultAlignment>>;
+                static_assert(std::is_empty_v<Allocator>, "AllocCpuAligned is expected to be stateless");
+                TElem* memPtr
+                    = alpaka::malloc<TElem>(Allocator{}, static_cast<std::size_t>(extent::getExtentProduct(extent)));
+                auto deleter = [](TElem* ptr) { alpaka::free(Allocator{}, ptr); };
+
+                return BufCpu<TElem, TDim, TIdx>(dev, memPtr, std::move(deleter), extent);
             }
         };
         //! The BufCpu memory mapping trait specialization.
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -33,6 +33,7 @@
 #    include <alpaka/dev/Traits.hpp>
 #    include <alpaka/dim/DimIntegralConst.hpp>
 #    include <alpaka/mem/buf/Traits.hpp>
+#    include <alpaka/meta/DependentFalseType.hpp>
 #    include <alpaka/vec/Vec.hpp>
 
 #    include <functional>
@@ -338,6 +339,70 @@ namespace alpaka
                     extent);
             }
         };
+
+        //! The CUDA/HIP stream-ordered 1D memory allocation trait specialization.
+        template<typename TElem, typename TIdx>
+        struct AsyncBufAlloc<TElem, DimInt<1u>, TIdx, DevUniformCudaHipRt>
+        {
+#    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && (BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(11, 2, 0))
+            static_assert(
+                meta::DependentFalseType<TElem>::value,
+                "Support for stream-ordered memory buffers require CUDA 11.2 or higher.");
+#    endif
+#    if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+            static_assert(
+                meta::DependentFalseType<TElem>::value,
+                "HIP devices do not support stream-ordered memory buffers.");
+#    endif
+
+            template<typename TQueue, typename TExtent>
+            ALPAKA_FN_HOST static auto allocAsyncBuf(TQueue queue, TExtent const& extent)
+                -> BufUniformCudaHipRt<TElem, DimInt<1u>, TIdx>
+            {
+                ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
+
+                DevUniformCudaHipRt const& dev = getDev(queue);
+
+                auto const width = extent::getWidth(extent);
+                auto const widthBytes = width * static_cast<TIdx>(sizeof(TElem));
+
+                // Set the current device.
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(dev.m_iDevice));
+                // Allocate the buffer on this device.
+                void* memPtr;
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(MallocAsync)(
+                    &memPtr,
+                    static_cast<std::size_t>(widthBytes),
+                    queue.m_spQueueImpl->m_UniformCudaHipQueue));
+                // Prepare a deleter for this device
+                auto deleter = [queue = std::move(queue)](TElem* ptr)
+                {
+                    // Set the current device.
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(getDev(queue).m_iDevice));
+                    // Free the buffer.
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(
+                        FreeAsync)(reinterpret_cast<void*>(ptr), queue.m_spQueueImpl->m_UniformCudaHipQueue));
+                };
+#    if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
+                std::cout << __func__ << " ew: " << width << " ewb: " << widthBytes << " ptr: " << memPtr << std::endl;
+#    endif
+                return BufUniformCudaHipRt<TElem, DimInt<1u>, TIdx>(
+                    dev,
+                    reinterpret_cast<TElem*>(memPtr),
+                    std::move(deleter),
+                    static_cast<TIdx>(widthBytes),
+                    extent);
+            }
+        };
+        //! The CUDA/HIP stream-ordered ND memory allocation trait specialization.
+        template<typename TElem, typename TDim, typename TIdx>
+        struct AsyncBufAlloc<TElem, TDim, TIdx, DevUniformCudaHipRt>
+        {
+            static_assert(
+                meta::DependentFalseType<TElem>::value,
+                "HIP/CUDA devices support only one-dimensional stream-ordered memory buffers.");
+        };
+
         //! The BufUniformCudaHipRt CUDA/HIP device memory mapping trait specialization.
         template<typename TElem, typename TDim, typename TIdx>
         struct Map<BufUniformCudaHipRt<TElem, TDim, TIdx>, DevUniformCudaHipRt>
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera
+/* Copyright 2021 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci
  *
  * This file is part of alpaka.
  *
@@ -62,20 +62,16 @@ namespace alpaka
 
     public:
         //! Constructor
-        template<typename TExtent>
+        template<typename TExtent, typename Deleter>
         ALPAKA_FN_HOST BufUniformCudaHipRt(
             DevUniformCudaHipRt const& dev,
             TElem* const pMem,
+            Deleter deleter,
             TIdx const& pitchBytes,
             TExtent const& extent)
             : m_dev(dev)
             , m_extentElements(extent::getExtentVecEnd<TDim>(extent))
-            , m_spMem(
-                  pMem,
-                  // NOTE: Because the BufUniformCudaHipRt object can be copied and the original object could have been
-                  // destroyed, a std::ref(m_dev) or a this pointer can not be bound to the callback because they are
-                  // not always valid at time of destruction.
-                  std::bind(&BufUniformCudaHipRt::freeBuffer, std::placeholders::_1, m_dev))
+            , m_spMem(pMem, std::move(deleter))
             , m_pitchBytes(pitchBytes)
         {
             ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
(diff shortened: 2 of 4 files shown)


fwyzard:
This is a reimplementation of #1480 following the discussion in https://github.com/alpaka-group/alpaka/pull/1480#issuecomment-977975403 and later.

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -69,6 +77,25 @@ namespace alpaka
         return traits::BufAlloc<TElem, Dim<TExtent>, TIdx, TDev>::allocBuf(dev, extent);
     }
 
+    //! The stream-ordered memory buffer type trait alias template to remove the ::type.
+    template<typename TDev, typename TElem, typename TDim, typename TIdx>
+    using AsyncBuf = typename traits::AsyncBufType<alpaka::Dev<TDev>, TElem, TDim, TIdx>::type;
```
fwyzard: I'm not convinced that this is needed with this approach, since the `AsyncBufType` should always be the same as the `BufType` when it is defined.
On the other hand, since currently this is only defined for the CPU and CUDA devices, it could be used to check if the async buffer functionality is actually implemented for a device ?

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -89,21 +85,8 @@ namespace alpaka
...
-
-            // Set the current device.
-            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(dev.m_iDevice));
-            // Free the buffer.
-            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(Free)(reinterpret_cast<void*>(memPtr)));
-        }
-
     public:
-        DevUniformCudaHipRt m_dev; // NOTE: The device has to be destructed after the memory pointer because it is
-                                   // required for destruction.
```
fwyzard, line 106: The comment is no longer relevant, because the deleter captures the device by value.

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -81,24 +79,12 @@ namespace alpaka
...
-                auto const extentElementCount = extent::getExtentProduct(extent);
-
-                return extentElementCount;
             }
 
         public:
             DevCpu const m_dev;
             Vec<TDim, TIdx> const m_extentElements;
-            TElem* const m_pMem;
+            std::shared_ptr<TElem> const m_spMem;
```
fwyzard: This approach does add a small overhead, replacing a `TElem*` with a `std::shared_ptr<TElem>`.
This could be avoided by storing the deleter directly in an `std::function` instead of using the `shared_ptr` for that.

An other possibility would be to follow the same approach used by `BufUniformCudaHipRt`:
  - move the metadata (device, extent, pitch) from `BufCpuImpl` directly into `BufCpu`
  - replace the `shared_ptr<BufCpuImpl>` with a `shared_ptr<TElem>`, and set the destructor there

However, this would trade the one less indirection with more expensive copies for the `BufCpu` objects, since the metadata would have to be copied as well.

fwyzard: Storing the deleter explicitly proved to be simple enough, and should add a smaller overhead than using an extra `shared_ptr`.

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -82,23 +82,14 @@ namespace alpaka
...
-
-                return extentElementCount;
+                m_fDeleter(m_pMem);
             }
 
         public:
             DevCpu const m_dev;
             Vec<TDim, TIdx> const m_extentElements;
             TElem* const m_pMem;
+            std::function<void(TElem*)> m_fDeleter;
```
fwyzard: is `m_deleter` a better choice of name ?

fwyzard:
I don't understand this error [from the CI](https://github.com/fwyzard/alpaka/runs/4321763635?check_suite_focus=true#step:5:5135):
```
 /home/runner/work/alpaka/alpaka/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp:396:23: error: no member named 'DependentFalseType' in namespace 'alpaka::meta'
                meta::DependentFalseType<TElem>::value,
                ~~~~~~^
/home/runner/work/alpaka/alpaka/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp:396:42: error: 'TElem' does not refer to a value
                meta::DependentFalseType<TElem>::value,
                                         ^
/home/runner/work/alpaka/alpaka/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp:392:27: note: declared here
        template<typename TElem, typename TDim, typename TIdx>
                          ^
/home/runner/work/alpaka/alpaka/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp:396:50: error: no member named 'value' in the global namespace
                meta::DependentFalseType<TElem>::value,
                                               ~~^
3 errors generated when compiling for sm_30.
```

bernhardmgruber:
Are you missing an `#include <alpaka/meta/DependentFalseType.hpp>` in `BufUniformCudaHipRt.hpp`?

fwyzard:
Ah, thanks, good catch !

include/alpaka/mem/buf/BufCpu.hpp
bernhardmgruber: Yes please. The `f` reminds me too much of Hungarian notation.

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -45,12 +45,12 @@ namespace alpaka
             static_assert(!std::is_const<TIdx>::value, "The idx type of the buffer can not be const!");
 
         public:
-            template<typename TExtent>
-            ALPAKA_FN_HOST BufCpuImpl(DevCpu const& dev, TExtent const& extent)
-                : AllocCpuAligned<std::integral_constant<std::size_t, core::vectorization::defaultAlignment>>()
-                , m_dev(dev)
+            template<typename TExtent, typename Deleter>
+            ALPAKA_FN_HOST BufCpuImpl(DevCpu const& dev, TElem* pMem, Deleter deleter, TExtent const& extent)
```
bernhardmgruber: Since you are stuffing the deleter into a `std::function` later anyway, I would just use it in the constructor interface as well. But that's just personal taste.
```suggestion
            template<typename TExtent>
            ALPAKA_FN_HOST BufCpuImpl(DevCpu const& dev, TElem* pMem, std::function<void(TElem*)> deleter, TExtent const& extent)
```

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -111,9 +102,10 @@ namespace alpaka
     class BufCpu
     {
     public:
-        template<typename TExtent>
-        ALPAKA_FN_HOST BufCpu(DevCpu const& dev, TExtent const& extent)
-            : m_spBufCpuImpl(std::make_shared<detail::BufCpuImpl<TElem, TDim, TIdx>>(dev, extent))
+        template<typename TExtent, typename Deleter>
+        ALPAKA_FN_HOST BufCpu(DevCpu const& dev, TElem* pMem, Deleter deleter, TExtent const& extent)
+            : m_spBufCpuImpl{
+                std::make_shared<detail::BufCpuImpl<TElem, TDim, TIdx>>(dev, pMem, std::move(deleter), extent)}
```
bernhardmgruber, line 112:
```suggestion
        template<typename TExtent>
        ALPAKA_FN_HOST BufCpu(DevCpu const& dev, TElem* pMem, std::function<void(TElem*)> deleter, TExtent const& extent)
            : m_spBufCpuImpl{
                std::make_shared<detail::BufCpuImpl<TElem, TDim, TIdx>>(dev, pMem, std::move(deleter), extent)}
```

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -233,7 +225,39 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
-                return BufCpu<TElem, TDim, TIdx>(dev, extent);
+                // alpaka::AllocCpuAligned is stateless
+                using Allocator
+                    = AllocCpuAligned<std::integral_constant<std::size_t, core::vectorization::defaultAlignment>>;
```
bernhardmgruber, line 234: Should we assert this? Something like:
```c++
struct Test : Allocator { int i; };
static_assert(sizeof(Test) == sizeof(int));
```

include/alpaka/mem/buf/Traits.hpp
bernhardmgruber: So far it seems we don't need it. I would remove it.

bernhardmgruber (commented):
Looks mostly good! Here are some more comments.

include/alpaka/mem/buf/BufCpu.hpp
fwyzard: Sure, I can make the change in the `BufCpuImpl` constructor.

I would not make the change for the `BufUniformCudaHipRt` constructor, since the latter passes the deleter to a `shared_ptr`, and I don't know how that's handled internally.

And I would not make the change for the `BufCpu` constructor, to keep the interface as close as possible among the different buffers.

include/alpaka/mem/buf/BufCpu.hpp
fwyzard: I see, but I don't think it would be useful, because a similar `assert()` would still pass for an `Allocator` that holds _e.g._ a pointer to an internal buffer, etc.

fwyzard: Ah, no, sorry, I misread the example.

fwyzard: What about
```c++
static_assert(std::is_empty_v<Allocator>, "AllocCpuAligned is expected to be stateless");
```
?

include/alpaka/mem/buf/Traits.hpp
fwyzard: OK.

include/alpaka/mem/buf/BufCpu.hpp
bernhardmgruber: OK, fine for me!

include/alpaka/mem/buf/Traits.hpp
fwyzard: Maybe add a default implementation for `AsyncBufAlloc` that fails with a `static_assert` and informs the user that the given buffer does not support stream-ordered memory allocation ?

fwyzard: Or better to leave it undefined, so it can be checked with SFINAE ?

include/alpaka/mem/buf/BufCpu.hpp
bernhardmgruber: Perfect! I did not know about that trait :) Thx!

include/alpaka/mem/buf/Traits.hpp
bernhardmgruber: Wouldn't that eliminate the purpose of alpaka, to give you a best effort implementation of a particular behavior across all accelerators? Your code would no longer be portable. And we can workaround the limitations on all backends by emplacing a host task into the queue, that frees the buffer.

In case you could query though, whether the underlying backend supports async alloc/free, what would you do with the information?

fwyzard: > And we can workaround the limitations on all backends by emplacing a host task into the queue, that frees the buffer.

I agree with you.
Sorry, I didn't mean that one should _not_ implement the stream-ordered behaviour all accelerator types.
But this PR currently does not, and I thought it would be useful so make it explicit somehow.

I'm happy to remove the `AsyncBuf` trait and _not_ add any checks, with the goal to eventually support stream-ordered allocations on every accelerator.

> In case you could query though, whether the underlying backend supports async alloc/free, what would you do with the information?

I guess the user code could fall back to some private implementation, or adapt the algorithms to use the standard allocations with some extra synchronisations.

bernhardmgruber: IIUC only the HIP, OpenAcc and OpenMP5 backends are missing? Would it be too hard to provide a quick implementation for them? With a callback, similar to the CPU implementation? If it is, I could open an issue and implement it later.

bernhardmgruber: > I guess the user code could fall back to some private implementation, or adapt the algorithms to use the standard allocations with some extra synchronisations.

Hmm. I asked the question because I wondered if the user would then resort to anyway implement the async alloc/free themselves by placing a callback into the queue. Phrased differently: can the user actually implement something better than our fallback implementation of placing the callback. Maybe we should provide a query than. But we can also wait until someone actually needs this feature.

fwyzard: I don't have experience with OpenACC and OpenMP5, so I'm not sure about those :-(

For HIP I don't think there is a CPU thread managed by Alpaka associated with the Queue, so we cannot use the same approach used for the PU buffer.
And, assuming it has the same limitations as CUDA that a callback cannot call a CUDA/HIP function, a callback cannot be used either.
The other implementation I can think of is a full-fledged caching memory allocator; that's my long term goal anyway, but not something I can implement quickly. Also, so far it's not clear if it would be desirable to have it inside Alpaka or built on top or on the side.

bernhardmgruber: > I don't have experience with OpenACC and OpenMP5, so I'm not sure about those :-(

Let's open an issue then when this PR is merged. @jkelling is experienced with those backends, so he might have a look later.

> And, assuming it has the same limitations as CUDA that a callback cannot call a CUDA/HIP function, a callback cannot be used either.

I think we have a workaround for that by keeping a background thread around. But I guess we can also implemented that later.

bernhardmgruber (changes requested):
Please apply the changed discussed in the various comments. I would also like to have at least 1 unit test for the proposed feature. Otherwise the PR looks good to me.

fwyzard:
Can you suggest one existing unit test to copy from and modify ?

bernhardmgruber:
> Can you suggest one existing unit test to copy from and modify ?

You can start by copying `test/unit/mem/buf/src/BufTest.cpp:22` and `...:44` and replacing `allocBuf` with `allocBufAsync`.

fwyzard:
I think I have addressed all the comments.
I'll be looking at adding a unit test next.

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -41,6 +41,35 @@ static auto testBufferMutable(alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>>
...
+
+    // synchronous operations must wait for the memory to be available
+    alpaka::wait(queue);
+    auto const offset = alpaka::Vec<Dim, Idx>::zeros();
+    alpaka::test::testViewImmutable<Elem>(buf, dev, extent, offset);
+
+    // memory will be deallocated once the queue reaches this point
+    alpaka::wait(queue);
+}
+
```
fwyzard, line 110: @bernhardmgruber did you have in mind something like this, or something else ?

Given that queue-ordered memory allocations are available only for CUDA and CPU buffers, I expect that the tests will fail for the other accelerators.
What would be a good way to make the new test run conditionally on the accelerator type ?

Also, is there a way to run the tests locally ?
I couldn't find any instructions for it.

bernhardmgruber: Something of that size looks fine to me for starters, thank you! I think you could add more tests on the lifetime aspect of these async buffers, like enqueuing a kernel, then destroying the buffer instance, then waiting for the kernel that still uses that buffer.

> What would be a good way to make the new test run conditionally on the accelerator type ?

Template specialization or overloading. I would pick the latter, add the `Acc` as unused first argument, then add empty overloads of `testAsyncBufferMutable` for the specific accelerators where the test does not work yet.

> Also, is there a way to run the tests locally ?

alpaka uses cmake's standard testing facilities. Set `BUILD_TESTING` to `ON` in cmake and build all the test targets (or just all targets since the examples should compile as well, in case you turned them on). Run `ctest` in the build dir to execute all tests.

fwyzard:
OK, the conditions in the test should now be complete.

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -18,6 +18,91 @@
...
+template<typename TAcc>
+[[maybe_unused]] static auto printAsyncBufferUnsupported() -> void
+{
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevCudaRt>)
+    {
+#    if BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(11, 2, 0)
+        std::cout << "Support for stream-ordered memory buffers require CUDA 11.2 or higher. "
+                  << "The test will be skipped." << std::endl;
```
bernhardmgruber:
```suggestion
        std::cout << "Support for stream-ordered memory buffers requires CUDA 11.2 or higher. "
                  << "The test will be skipped." << std::endl;
```

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -64,6 +178,40 @@ TEMPLATE_LIST_TEST_CASE("memBufZeroSizeTest", "[memBuf]", alpaka::test::TestAccs
...
+
+    if constexpr(isAsyncBufferSupported<Acc>())
+    {
+        auto const extent
+            = alpaka::createVecFromIndexedFn<Dim, alpaka::test::CreateVecWithIdx<Idx>::template ForExtentBuf>();
+        testAsyncBufferMutable<Acc>(extent);
+    }
+    else
+    {
+        printAsyncBufferUnsupported<Acc>();
```
bernhardmgruber:
```suggestion
        INFO("Stream-ordered memory buffers not supported");
```
And delete `printAsyncBufferUnsupported`. I think this is good enough. But do as you like.

Two more occurences below.

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -18,6 +18,91 @@
...
+    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevCudaRt>)
+    {
+#    if BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(11, 2, 0)
+        return false;
+#    else
+        if constexpr(alpaka::Dim<TAcc>::value > 1)
+        {
+            return false;
+        }
+#    endif // BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(11, 2, 0)
```
bernhardmgruber:
```suggestion
        return (BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)) && alpaka::Dim<TAcc>::value == 1;
```

bernhardmgruber (dismissed):

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -18,6 +18,91 @@
...
+[[maybe_unused]] static auto printAsyncBufferUnsupported() -> void
+{
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevCudaRt>)
+    {
+#    if BOOST_LANG_CUDA < BOOST_VERSION_NUMBER(11, 2, 0)
+        std::cout << "Support for stream-ordered memory buffers require CUDA 11.2 or higher. "
+                  << "The test will be skipped." << std::endl;
+#    else
+        if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevCudaRt> and alpaka::Dim<TAcc>::value > 1)
```
bernhardmgruber:
```suggestion
        if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevCudaRt> && alpaka::Dim<TAcc>::value > 1)
```
The alternative operator keywords are only availabel in MSVC when compiling in standard conformance mode (`/permissive-`). Unfortunately, CMake chooses the non-conforming mode by default. Please avoid them for the sake of broader MSVC support.

fwyzard:
OK, making progress with the tests, now there are only two failures left.

---
LeakSanitizer complains about
```
28: Direct leak of 31680 byte(s) in 1 object(s) allocated from:
28:     #0 0x4e12f8  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x4e12f8)
28:     #1 0x875676  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x875676)
28:     #2 0x61683d  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x61683d)
28:     #3 0x8f01e9  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x8f01e9)
28:     #4 0x8d7b29  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x8d7b29)
28:     #5 0x8d0e2d  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x8d0e2d)
28:     #6 0x8ce7b7  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x8ce7b7)
28:     #7 0x8e2a98  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x8e2a98)
28:     #8 0x8e01b3  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x8e01b3)
28:     #9 0x93065d  (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0x93065d)
28:     #10 0x7fbb31ae8bf6  (/lib/x86_64-linux-gnu/libc.so.6+0x21bf6)
...
```
which I could use some help to investigate; in particular:
  - can you make the sanitizer produce a more meaningful stack trace ?
  - can you suggest how to reproduce it locally ?

---
All NVCC builds on Windows fail with
```
D:/a/alpaka/alpaka/test/unit/mem/buf/src/BufTest.cpp(72): error : expected a ")" [D:\a\alpaka\alpaka\build\test\unit\mem\buf\memBufTest.vcxproj]
             detected during:
               instantiation of "auto printAsyncBufferUnsupported<TAcc>()->void [with TAcc=alpaka::test::detail::AccGpuUniformCudaHipRtIfAvailableElseInt<alpaka::DimInt<2ULL>, uint64_t>]"
   (195): here
               instantiation of "void ____C_A_T_C_H____T_E_M_P_L_A_T_E____T_E_S_T____F_U_N_C____7<TestType>() [with TestType=alpaka::test::detail::AccGpuUniformCudaHipRtIfAvailableElseInt<alpaka::DimInt<2ULL>,
   (181): here
               instantiation of "void <unnamed>::ns_____C_A_T_C_H____T_E_M_P_L_A_T_E____T_E_S_T____6::____C_A_T_C_H____T_E_M_P_L_A_T_E____T_E_S_T____6<Types...>::reg_tests() [with Types=<alpaka::test::detail::
   (181): here
```
That code looks fine to me, and GCC and clang do not complain about it... is there some limitation with Visual Studio 2019 in C++ 17 mode ?

test/unit/mem/buf/src/BufTest.cpp
fwyzard: sigh ...

test/unit/mem/buf/src/BufTest.cpp
fwyzard: OK for me.

test/unit/mem/buf/src/BufTest.cpp
fwyzard: Thanks.

fwyzard:
OK, all checks finally pass.

The last problem was interesting, as it highlights a detail of how the stream-ordered buffers should be used: since the buffer will queue the call to free the memory when it goes out of scope, to be able to wait for it to happen the lifetime of the queue should be longer than the buffer's own scope.
This is unlikely to be an issue in a larger program, where the queue will likely be used across many functions (otherwise it makes little sense to use an async buffer), but it needed special attention in the unit test.

bernhardmgruber:
> since the buffer will queue the call to free the memory when it goes out of scope, to be able to wait for it to happen the lifetime of the queue should be longer than the buffer's own scope.

I wonder whether we could check that. Or even block in the queue's destructor. Or extent the queue's lifetime. Since we have some `shared_ptr`s in the backend, we could keep `shared_ptr`s or possibly `weak_ptr`s to other alpaka objects around. Like having a shared pointer to the queue an async buffer was created on inside the buffer, so the queue is kept alive, etc. Or the queue having a vector of weak pointes to buffers, to verify that they are all gone and block otherwise. But this is definitely out of the scope for this PR.

bernhardmgruber (approved):

fwyzard:
The queue itself should already live at least as long as the buffer, since the buffer's destructor holds a copy of it - but it goes away when the call to free the memory is enqueued.

We could keep the queue alive until all pending memory operations have completed by passing a copy of it to the enqueued function itself:
```patch
diff --git a/include/alpaka/mem/buf/BufCpu.hpp b/include/alpaka/mem/buf/BufCpu.hpp
index 1d7b72e44b84..52b28400f1c6 100644
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -261,7 +261,17 @@ namespace alpaka
                 TElem* memPtr
                     = alpaka::malloc<TElem>(Allocator{}, static_cast<std::size_t>(extent::getExtentProduct(extent)));
                 auto deleter = [queue = std::move(queue)](TElem* ptr) mutable
-                { alpaka::enqueue(queue, [ptr]() { alpaka::free(Allocator{}, ptr); }); };
+                {
+                    alpaka::enqueue(
+                        queue,
+                        [ptr, queue]()
+                        {
+                            // free the memory
+                            alpaka::free(Allocator{}, ptr);
+                            // keep the queue alive until all memory operations have cimpleted
+                            alpaka::ignore_unused(queue);
+                        });
+                };
 
                 return BufCpu<TElem, TDim, TIdx>(dev, memPtr, std::move(deleter), extent);
             }
```

I haven't considered if this may cause any circular references or not, or other issues.


Alternatively, making the queue's destructor wait on all pending operations would solve the issue, as well.

PR #1493, opened 2021-11-30
Switch to C++17 std::apply

### LINKED ISSUE #912: ApplyTuple: Returned Value Lifetime
https://lgtm.com/projects/g/ComputationalRadiationPhysics/alpaka/?mode=list

The `invoke_impl` implementation detail of `meta::ApplyTuple` creates a warning on LGTM of the form:
```
May return stack-allocated memory from pmd.
```

What does this mean:
- https://lgtm.com/rules/1507923806246/
- occurrence: [`include/alpaka/meta/ApplyTuple.hpp` line 40](https://github.com/ComputationalRadiationPhysics/alpaka/blob/9811f23a306a731342bed038b5af7e5494310f20/include/alpaka/meta/ApplyTuple.hpp#L40)

### LINKED ISSUE #1494: Use C++17 metaprogramming facilities.
In the `meta` directory there are several files referencing their C++17 counterparts:

* `Metafunctions.hpp` (`std::conjunction`, `std::disjunction` and `std::negation`)
* `Void.hpp` (`std::void_t`)
* `ApplyTuple.hpp` (`std::invoke` and `std::apply`)

Now that we have made the switch to C++17 we should address this.

sbastrakov:
Remove old `meta::apply` that was mimicking it and tests for it. Remove `meta::invoke` that now became unused (and `std::invoke` is now available anyways).

I do not understand why did we have `core::apply` for the "third party" tuple. I assume this tuple should also be compatible with `std::apply`? Maybe the tests will prove me wrong. That change is a separate commit so can remove it if needed.


j-stephan:
I think you can also safely remove the `Metafunctions.hpp` as part of this PR. AFAIK they are only used for our handrolled `apply` and `invoke`.

sbastrakov:
@j-stephan no, those were actually not used for our `apply`. I see that it is used somewhere else tho.

j-stephan:
You're right, I confused this with `meta/Unique.hpp`.

bernhardmgruber (dismissed):

include/alpaka/core/Tuple.hpp
```diff
@@ -11,31 +11,10 @@
...
 #include <alpaka/core/STLTuple/STLTuple.hpp>
 
-#include <cstddef>
-#include <utility>
-
 namespace alpaka
 {
     namespace core
     {
         using namespace ::utility::tuple;
```
j-stephan: This is the only meaningful line remaining in this file. Is this still required?

sbastrakov: I think my change to this file may have been wrong and second commit of the PR has to be removed. But anyhow the highlighted part makes sense to keep imo

bernhardmgruber:
Please rebase!

j-stephan:
Please rebase!

jkelling:
@sbastrakov :
> I do not understand why did we have core::apply for the "third party" tuple. I assume this tuple should also be compatible with std::apply? Maybe the tests will prove me wrong. That change is a separate commit so can remove it if needed.

The third party tuple does not compile with `std::apply`. I just checked: At least one issue is, that it does not implement the `tuple_size` trait, which `std::apply` employs. So with your changes to the OpenMP5 and  OpenACC backends, the related CI jobs should fail.

sbastrakov:
Thanks for clarification, I retracted the changes to `core::apply`.

j-stephan (approved):

PR #1484, opened 2021-11-30
Switch all CI images to ubuntu 20.04

bernhardmgruber:
Also updates the links to CUDA installers.
Fixes #1471.

PR #1502, opened 2021-11-30
Use C++17 mandatory copy elision

### LINKED ISSUE #1497: Use C++17 mandatory copy elision for CPU mem fences.
In `mem/fence/MemFenceCpu.hpp` we find the following TODO:

```c++
/* TODO: Once we have C++17 enabled by default the following line should be turned into the usual
 * `auto x = ...` notation.
 */
static std::atomic<int> dummy{42};
```

This can now be done.

j-stephan:
Fixes #1497.

### DIFF
--- a/include/alpaka/mem/fence/MemFenceCpu.hpp
+++ b/include/alpaka/mem/fence/MemFenceCpu.hpp
@@ -59,10 +59,7 @@ namespace alpaka
                  *   a == 10 && b == 20
                  */
 
-                /* TODO: Once we have C++17 enabled by default the following line should be turned into the usual
-                 * `auto x = ...` notation.
-                 */
-                static std::atomic<int> dummy{42};
+                static auto dummy = std::atomic<int>{42};
 
                 /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
                  * these dummy operations to ensure this.*/


include/alpaka/mem/fence/MemFenceCpu.hpp
```diff
@@ -59,10 +59,7 @@ namespace alpaka
                  *   a == 10 && b == 20
                  */
 
-                /* TODO: Once we have C++17 enabled by default the following line should be turned into the usual
-                 * `auto x = ...` notation.
-                 */
-                static std::atomic<int> dummy{42};
+                static auto dummy = std::atomic<int>{42};
```
bernhardmgruber, line 62: Does this make the code better? Is this change mandated by a style guide? I would be fine with the original version and just remove the comment.

sbastrakov: I think it follows the AAA aka almost always auto (which in C++17 becomrs always auto) which we have as a recommendation but not a requirement

j-stephan: Looks like older clang versions don't (fully) support this feature anyway :-(

j-stephan: Wait, this might be solved by the Ubuntu upgrade in #1484. The failing checks are still using the old gcc stdlib from Ubuntu 18.04.

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #1488, opened 2021-11-30
Test installation in CI

bernhardmgruber:
Tests installation as part of the CI.

PR #1507, opened 2021-12-02
Fix installation

### LINKED ISSUE #1487: Installation is broken
The PR #1456 removed the directory "cmake/modules", which is still referenced in `CMakeLists.txt` for installation alpaka. This makes the installation fail. See e.g. LLAMA CI: https://github.com/alpaka-group/llama/runs/4365043305?check_suite_focus=true

bernhardmgruber:
I split this off from #1488, because I need this fix fast so the LLAMA CI will work again.


j-stephan (approved):

psychocoderHPC (approved):

psychocoderHPC:
The broken CI test will be fixed with #1510

PR #1511, opened 2021-12-03
Simplify CPU kernel launch

### LINKED ISSUE #1492: Create perfectly forwarding function objects for CPU back-ends
The `TaskKernel` implementations for the TBB, OpenMP 2.x and Serial back-ends feature the following comment:

```c++
// TODO: With C++14 we could create a perfectly argument forwarding function object within the constructor.
```

We should address this as part of the C++17 migration.

j-stephan:
Fixes #1492 (with an even simpler solution).

### DIFF
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -827,12 +827,6 @@ namespace alpaka
             std::cout << __func__ << " blockSharedMemDynSizeBytes: " << blockSharedMemDynSizeBytes << " B"
                       << std::endl;
 #    endif
-            // Bind all arguments except the accelerator.
-            // TODO: With C++14 we could create a perfectly argument forwarding function object within the constructor.
-            auto const boundKernelFnObj = meta::apply(
-                [this](ALPAKA_DECAY_T(TArgs) const&... args)
-                { return std::bind(std::ref(m_kernelFnObj), std::placeholders::_1, std::ref(args)...); },
-                m_args);
 
             // The number of blocks in the grid.
             TIdx const numBlocksInGrid(gridBlockExtent.prod());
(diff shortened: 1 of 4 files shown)


include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -910,7 +903,7 @@ namespace alpaka
 #    endif
                 acc.m_gridBlockIdx = mapIdx<TDim::value>(index, gridBlockExtent);
 
-                boundKernelFnObj(acc);
+                std::apply(m_kernelFnObj, std::tuple_cat(std::make_tuple(acc), m_args));
```
bernhardmgruber: I wonder whether that changes the semantic in a problematic way. Your new solution makes N threads access `this` to fetch `m_kernelFnObj` and `m_args` in parallel. All kernel threads also share the functor and argument instances. This is definitely a problem if the functor mutates itself. A user could also now get a reference to one of `m_args`, if they use a mutable reference in the kernel function interface and produce a race condition by writing to it.

I think I would prefer giving each thread an independent copy of the functor and args.

j-stephan: > This is definitely a problem if the functor mutates itself.

I don't think we allow this as it should be marked `const`.

psychocoderHPC: Yes arguments are const, a non-const reference in the user kernel signature should not work. An open question for OpenMP is if it is better to use variables `shared` or an instance per thread. I would say it does not matter if all arguments are const but I am not fully sure.

bernhardmgruber: Ok. That solves the issue for `m_args`. What if the `operator()` on the kernel functor is not const? Is that allowed? If not, we are fine with the changes.

j-stephan: I don't think non-`const` function objects are allowed. Every instance of the kernel `operator()` I've seen so far has been marked as `const`.

j-stephan: Our docs explicitly say that the `operator()` must be `const`. See the code snippet at https://alpaka.readthedocs.io/en/latest/basic/library.html#kernel-definition

bernhardmgruber: Then your changes are fine!

bernhardmgruber (dismissed):

psychocoderHPC (approved):

PR #1515, opened 2021-12-07
Avoid setting the CUDA/HIP device when not necessary

### LINKED ISSUE #1477: alpaka API call overhead can be high
Alpaka is calling with the HIP/CUDA backend for each memcopy, kernel call, ... `hip/cudaSetDavice` and `hip/cudaGetLastError`. Those calls can increase the latency for API calls a lot.
In cases where the user can guarantee that only one device per process is used `hip/cudaSetDevice` is a useless call. For HIP calling `hipGetLastError` introduces 11us overhead. In an application with 1k API calls per second, this could result in 11ms overhead if the calls can not be hidden by calculations.
`hip/cudaGetLast` error can be handled easily not ask for each API call the last error, IMO this should be only used for debug mode.
https://github.com/alpaka-group/alpaka/blob/0ef767dc1ef097bea9a35865cbcbd92b74d85f3f/include/alpaka/core/UniformCudaHip.hpp#L122-L125

HIP trace of PIConGPU:
GetLastError
Note: The pure HIP-API call is only 2us but additional overhead is coming from handling the error and in this case maybe cupla too.

`hip/cudaSetDavice` is more challenging, maybe we should allow the user to set a flag for a device that this is the only used device for the application if you know that he is not using multiple devices.

fwyzard:
Do not set the current CUDA/HIP device when it's not necessary:
  - `cudaMemsetAsync` variants can be called with a different device
  - `cudaFree` and `cudaFreeAsync` can be called with a different device
  - `cudaStreamSynchronize` and `cudaStreamDestroy` can be called with a different device
  - `cudaEventQuery` and `cudaEventDestroy` can be called with a different device

`cudaMemset` requires setting the current device, but is not used in Alpaka.

`cudaMemcpy` and `cudaMemcpy` variants working on `cudaMalloc`'ed memory can be called with a different device.
Unfortunately, the correct device must be set when working on memory allocated by `cudaMallocAsync` (see https://github.com/fwyzard/nvidia_bug_3446335).

### DIFF
--- a/include/alpaka/event/EventUniformCudaHipRt.hpp
+++ b/include/alpaka/event/EventUniformCudaHipRt.hpp
@@ -79,10 +79,6 @@ namespace alpaka
                 {
                     ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
-                    // Set the current device. \TODO: Is setting the current device before cuda/hip-EventDestroy
-                    // required?
-
-                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(m_dev.m_iDevice));
                     // In case event has been recorded but has not yet been completed when cuda/hip-EventDestroy() is
                     // called, the function will return immediately and the resources associated with event will be
                     // released automatically once the device has completed event.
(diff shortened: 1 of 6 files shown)


fwyzard:
See [this gist](https://gist.github.com/fwyzard/1aac6214c8303b4a74abcf170c61b7b9) for a working test program.

fwyzard:
This was only tested with CUDA 11.5 on a system with two T4 GPUs.
Before these changes could be merged more testing is definitely needed, on different systems, different CUDA versions, and with AMD hardware using HIP/ROCm.

psychocoderHPC:
This PR is partially solving #1477

j-stephan (approved):
Thanks!

PR #1518, opened 2021-12-07
Add changelog for 0.8

j-stephan:


CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+  - GCC 11 #1383
+  - HIP-clang #1338
+  - HIP-nvcc #1337
+  - Xcode 12.5.1 #1385
+  - Xcode 13 #1421
+- Platform support removed:
+  - clang < 5.0 #1385
+  - CUDA < 9.2 #1385
+  - GCC < 7.0 #1385
+  - `hipcc` #1338
```
bernhardmgruber:
```suggestion
  - hipcc #1338
```
all other compiler names are typeset differently

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+  - MSVC < 2019 #1385
```
bernhardmgruber: Either
```suggestion
  - Visual Studio < 2019 #1385
```
or
```suggestion
  - MSVC < 14.2 #1385
```

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+  - Ubuntu 16.04 #1352
+  - Xcode 11.x < 11.3.1 #1385
+  - Xcode 12.x < 12.4 #1385
+ 
+### Bug Fixes:
+- Added missing `#include <limits>` in a few places which would lead to compilation errors for CPU back-ends #1327
+- Added missing `std::` to fixed-width integers where necessary #1327
+- fixed behavior of `assert` and `printf` for OpenMP offloading targets #1351
+- `ALPAKA_STATIC_ACC_MEM_CONSTANT` now works correctly for clang-CUDA as well as HIP #1386
+- The OpenMP 5 and OpenACC back-ends now correctly pass the parameters as a `is_trivially_copyable` type #1387
```
bernhardmgruber:
```suggestion
- The OpenMP 5 and OpenACC back-ends now correctly pass the parameters as an `is_trivially_copyable` type #1387
```

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+  - `hipcc` #1338
+  - MSVC < 2019 #1385
+  - Ubuntu 16.04 #1352
+  - Xcode 11.x < 11.3.1 #1385
+  - Xcode 12.x < 12.4 #1385
+ 
+### Bug Fixes:
+- Added missing `#include <limits>` in a few places which would lead to compilation errors for CPU back-ends #1327
+- Added missing `std::` to fixed-width integers where necessary #1327
+- fixed behavior of `assert` and `printf` for OpenMP offloading targets #1351
```
bernhardmgruber:
```suggestion
- Fixed behavior of `assert` and `printf` for OpenMP offloading targets #1351
```

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+- `ALPAKA_STATIC_ACC_MEM_CONSTANT` now works correctly for clang-CUDA as well as HIP #1386
+- The OpenMP 5 and OpenACC back-ends now correctly pass the parameters as a `is_trivially_copyable` type #1387
+- Fixed `alpaka_compiler_option` checking for the wrong variable name #1392
+- Fixed a bug in the HIP back-end's peer-to-peer `memcpy` implementation #1400
+- The CMake function `alpaka_compiler_option` has been turned into a macro which solves parameter scope issues #1401
+- Fixed compilation error occuring with CUDA >= 11.3 #1404
+- the `-pthread` flag is now correctly passed to the (host) compiler and linker #1420
```
bernhardmgruber: Please either start all lines consistently with an upper case or lower case letter.

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+- alpaka now correctly sets the CUDA host compiler #1423
+- alpaka's headers are now treated as CMake `SYSTEM` headers so that internal warnings no longer annoy users #1451
+- Projects using alpaka can now set `ALPAKA_CXX_STANDARD` without alpaka ignoring this #1463
```
bernhardmgruber: They could already do that. I think you mean that it can be set inside a cmake script before including alpaka via `find_package`. Consider wording this more precise.

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+### New Features:
+- alpaka now supports the Philox random number generator #1319
+- alpaka's kernel language now supports memory fences (a.k.a thread fences) #1379
+- `alpaka::Vec` now supports structured bindings #1393
+- The OpenMP 5 and OpenACC back-ends now support statically mapped memory #1394
+- alpaka now has factory methods for creating memory views #1398
+- If OpenMP >= 5.1 is supported the back-end makes use of `atomic capture compare` #1411
+- alpaka now experimentally supports accessors instead of pointers to access memory #1433
+- added wrapper for CUDA's native vector types so that they may be handled like arrays #1435
+- added function for **exact** floating point comparisons #1440
```
bernhardmgruber: maybe mention that this is testing functionality.

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+- math traits have been (internally) simplified #1457
+- Visual Studio project files are no longer tracked by git #1464
+- The alpaka CMake project now enables the `CXX` language by default everywhere (previously some test cases would enable `C`) #1470
+
+### Breaking Changes:
+- legacy TBB support is deprecated. alpaka will move to oneTBB in 0.9 #1329
+- HIP + nvcc is no longer supported #1337
+- `hipcc` is no longer supported as compiler for HIP. Use HIP's clang instead #1338
+- the behavior of `ALPAKA_STATIC_ACC_MEM_CONSTANT` and `ALPAKA_STATIC_ACC_MEM_GLOBAL` was changed #1386
+- `alpaka::rand::engine::createDefault()` now features an additional `offset` parameter #1434
```
bernhardmgruber, line 72: is that a breaking change?

bernhardmgruber (changes requested):

CHANGELOG.md
j-stephan: It changes the API, so IMO yes.

bernhardmgruber: Oh sorry, I read "now features an optional offset parameter". It's fine then!

CHANGELOG.md
```diff
@@ -3,6 +3,91 @@ All notable changes to this project will be documented in this file.
...
+- Removed unnecessary `explicit` specifiers #1419
+- simplified implementation of `ALPAKA_UNROLL` #1437
+- math traits have been (internally) simplified #1457
+- Visual Studio project files are no longer tracked by git #1464
+- The alpaka CMake project now enables the `CXX` language by default everywhere (previously some test cases would enable `C`) #1470
+
+### Breaking Changes:
+- legacy TBB support is deprecated. alpaka will move to oneTBB in 0.9 #1329
+- HIP + nvcc is no longer supported #1337
+- `hipcc` is no longer supported as compiler for HIP. Use HIP's clang instead #1338
```
psychocoderHPC: `hipcc` is still supported but `native clang can be used too.
The title must be refrased.

CHANGELOG.md
psychocoderHPC: this line must be removed. `hipcc` is still supported but not required anymore. It is now possible to use a modern clang too. `hipcc` is only wrapping the clang compiler and can therefore still be used.

psychocoderHPC (changes requested):

CHANGELOG.md
j-stephan: Note: We should think about deprecating hipcc. CMake 3.21 introduced native HIP support which only works with hip-clang (hipcc is explicitly not supported).

CHANGELOG.md
j-stephan: > I think you mean that it can be set inside a cmake script before including alpaka

What is the difference to "projects using alpaka"?

CHANGELOG.md
j-stephan: Why would it be testing functionality? It lives inside `alpaka::math` and is properly documented and unit tested.

CHANGELOG.md
psychocoderHPC: Yes we can do it but not sure if it helps to deprecate something in the changelog only. Normally you add a cmake warning which can be suppressed if needed.

CHANGELOG.md
SimeonEhrig: It means a another project, which uses alpaka via `find_package` and the improvement is, that you can use a normal variable instead a cached variable, now:

```cmake
# before it was ignored and the default value was used
set(ALPAKA_CXX_STANDARD 17)
find_package(alpaka)
```

bernhardmgruber: I have a LLAMA  example that is a "project using alpaka". I can configure it with `cmake -DALPAKA_CXX_STANDARD=20` and this worked before the change merged here. The difference you want to highlight is that this can now be done inside the `CMakeLists.txt` of my LLAMA example.

CHANGELOG.md
bernhardmgruber: It is? I thought it was part of the unit test project. Ok, then leave it as it is!

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #1520, opened 2021-12-08
CI: run linux CUDA job on gitlab only

psychocoderHPC:
- remove linux CUDA jobs from github workflows
- add CUDA tests to gitlab CI
  - by default build for Quadro P5000 (sm_60)

Build only few tests for two architectures to reduce the compile overhead.

- [x] remove CUDA jobs from the workflow required list (tasks for an admin)
**NOTE** this PR is currently used to check why the CUDA CI jobs with matmul are failing.
- [x] merge after #1527 (includes as first commit, no rebase anymore required)


.gitlab-ci.yml
```diff
@@ -28,5 +28,14 @@ variables:
...
+  - local: '/script/gitlabci/job_cuda9.2.yml'
+  - local: '/script/gitlabci/job_cuda10.0.yml'
+  - local: '/script/gitlabci/job_cuda10.1.yml'
+  - local: '/script/gitlabci/job_cuda10.2.yml'
+  - local: '/script/gitlabci/job_cuda11.0.yml'
+  - local: '/script/gitlabci/job_cuda11.1.yml'
+  - local: '/script/gitlabci/job_cuda11.2.yml'
+  - local: '/script/gitlabci/job_cuda11.3.yml'
+  - local: '/script/gitlabci/job_cuda11.4.yml'
+  - local: '/script/gitlabci/job_cuda11.5.yml'
```
bernhardmgruber, line 40: Why is this necessary? Is there any advantage?

bernhardmgruber (approved):
I cannot review that the new jobs exactly represent the old ones, so I trust you on these.

.gitlab-ci.yml
psychocoderHPC: - The files get long and it is hard to keep an overview.
- It is much easier to compare to setups for different CUDA versions because I can render both side by side.

bernhardmgruber: I see. So there is no: multiple files can run in parallel etc.? I thought there might be an advantage like that.

psychocoderHPC: All jobs running in parallel. The problem is that the CI has only 4 slots for NVIDIA jobs. This is limiting parallelism.

Please do not ask me why we have only 4 slots else I start :sob:

psychocoderHPC:
Note: The gitlab CI is this week running with fewer resources because there is maintenance and all CPU jobs running on AMD and NVIDIA nodes.

bernhardmgruber:
> Note: The gitlab CI is this week running with fewer resources because there is maintenance and all CPU jobs running on AMD and NVIDIA nodes.

I was already worried that the move to the GitLab CI would generally slow down the CI :S Let's see then.

bernhardmgruber (dismissed):

psychocoderHPC:
> > Note: The gitlab CI is this week running with fewer resources because there is maintenance and all CPU jobs running on AMD and NVIDIA nodes.
> 
> I was already worried that the move to the GitLab CI would generally slow down the CI :S Let's see then.

A slowdown is possible but the gitlab CI is running runtime tests we were not able to perform on github actions. "Correctness goes over performance!"

I think we have an open issue for better CI performance and we need more clever combinations to solve the CI performance issue.

psychocoderHPC:
For NVIDIA only the test randomStrategies is failing, this test will be fixed with #1527

bernhardmgruber (approved):

PR #1525, opened 2021-12-09
Revert "Avoid setting the CUDA/HIP device when not necessary"

j-stephan:
Reverts alpaka-group/alpaka#1515

Opening this draft PR because of some weird CI issues we have been experiencing with CUDA and HIP lately. This draft is checking whether #1515 is  causing these.


psychocoderHPC:
I will close this PR, the CI issue is fixed with "fix Philox RNG" #1527 and additionally the CI node HIP driver installation was broken. Both issues are solved now.

PR #1527, opened 2021-12-10
fix Philox RNG

psychocoderHPC:
- randomStrategies test:
  - fix out of memory access
  - always fill all cells with rng numbers
- fix randomCells2D out of memory access
- PhiloxSingle: fix error `'cudaErrorLaunchOutOfResources': 'too many
resources requested for launch'`
  - remove runtime dependent access into a static array hold in register
(avoid register spilling)
  - fix that the first random number of 4 is always skipped

The changes reducing the register footprint from 90 registers to 50 for
sm_70.

@j-stephan This bug affects the release 0.8.0!

### DIFF
--- a/example/randomStrategies/src/randomStrategies.cpp
+++ b/example/randomStrategies/src/randomStrategies.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2021 Jiri Vyskocil
+/* Copyright 2021 Jiri Vyskocil, Rene Widera
  *
  * This file is part of alpaka.
  *
@@ -112,8 +112,11 @@ struct InitRandomKernel<Strategy::seed>
     ) const -> void
     {
         auto const idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0]; ///< index of the current thread
-        TRandEngine engine(idx, 0, 0); // Initialize the engine
-        states[idx] = engine; // Save the initial state
+        if(idx < extent[0])
+        {
+            TRandEngine engine(idx, 0, 0); // Initialize the engine
+            states[idx] = engine; // Save the initial state
+        }
     }
 };
 
@@ -130,8 +133,11 @@ struct InitRandomKernel<Strategy::subsequence>
     {
         alpaka::ignore_unused(skipLength);
         auto const idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0]; ///< index of the current thread
-        TRandEngine engine(0, idx, 0); // Initialize the engine
-        states[idx] = engine; // Save the initial state
+        if(idx < extent[0])
+        {
+            TRandEngine engine(0, idx, 0); // Initialize the engine
+            states[idx] = engine; // Save the initial state
+        }
     }
 };
 
@@ -148,8 +154,11 @@ struct InitRandomKernel<Strategy::offset>
     {
         alpaka::ignore_unused(skipLength);
         auto const idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0]; ///< index of the current thread
-        TRandEngine engine(0, 0, idx * skipLength); // Initialize the engine
-        states[idx] = engine; // Save the initial state
+        if(idx < extent[0])
+        {
+            TRandEngine engine(0, 0, idx * skipLength); // Initialize the engine
+            states[idx] = engine; // Save the initial state
+        }
     }
 };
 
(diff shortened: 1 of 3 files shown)


include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -84,12 +84,23 @@ namespace alpaka
                  */
                 ALPAKA_FN_HOST_ACC auto nextNumber()
                 {
+                    // Element zero will always contain the next valid random number.
+                    auto result = state.result[0];
                     state.position++;
```
psychocoderHPC, line 89: @sliwowitz During the first access this line increments the position from `0` to `1` and `return state.result[state.position];` will therefore always skip the value in the array on pos index 0.

psychocoderHPC:
@sliwowitz Can we have on Monday a VC on to take care that I have not broken anything?

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -167,16 +176,21 @@ struct FillKernel
...
-
-        RandomEngine<TAcc> engine(states[idx]); // Setup the PRNG using the saved state for this thread.
-        alpaka::rand::UniformReal<float> dist; // Setup the random number distribution
-        for(unsigned i = 0; i < length; ++i)
+        auto const numGridThreads = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc)[0];
+        if(idx < NUM_POINTS)
         {
-            cells[length * idx + i] = dist(engine); // Roll the dice!
+            // each worker is handling one random state
+            auto const worker
```
psychocoderHPC: schould be `workers` or `numWorkers`

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -98,6 +109,15 @@ namespace alpaka
...
                     offset += state.position < 4 ? 0 : 4;
                     state.position -= state.position < 4 ? 0 : 4u;
+                    while(state.position > 0)
+                    {
+                        // Shift state results to allow hard coded access to element zero.
+                        // This will avoid high register usage on NVIDIA devices.
+                        state.result[0] = state.result[1];
+                        state.result[1] = state.result[2];
+                        state.result[2] = state.result[3];
+                        --state.position;
```
psychocoderHPC: This is wrong, we need to remember the old position and decrease the value for the loop only.

include/alpaka/rand/Philox/PhiloxSingle.hpp
```diff
@@ -98,6 +109,15 @@ namespace alpaka
                     state.position = static_cast<decltype(state.position)>(state.position + (offset & 3));
                     offset += state.position < 4 ? 0 : 4;
                     state.position -= state.position < 4 ? 0 : 4u;
```
psychocoderHPC, line 115: static assert missing: this is only valid for `TParams::counterSize == 4`

psychocoderHPC:
The bug 
```
Device::callbackQueue aborting with error : HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception. code: 0x1016
```

we currently see was also seen in CMS patatrack https://github.com/cms-patatrack/pixeltrack-standalone/issues/178.
From what I see in the issue it could be something with shared memory.

include/alpaka/rand/Philox/PhiloxSingle.hpp
psychocoderHPC: All elements will be used because the access to the rng number is after the advance function where the position is set back to zero.

psychocoderHPC:
> The bug
> 
> ```
> Device::callbackQueue aborting with error : HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception. code: 0x1016
> ```
> 
> we currently see was also seen in CMS patatrack [cms-patatrack/pixeltrack-standalone#178](https://github.com/cms-patatrack/pixeltrack-standalone/issues/178). From what I see in the issue it could be something with shared memory.

Is fixed by this PR. We avoid runtime access into a compile time array within registers.

psychocoderHPC:
note: @sliwowitz and I reviewed this PR today together in a VC. We found that  randomCells2D produced out of memory access too.

j-stephan (approved):

sliwowitz:
As we discussed with @psychocoderHPC yesterday, the shifting of array states, instead of runtime indexing, saves GPU register resulting in better CUDA performance. In case it turns out to be a performance issue on the CPU, the affected functions can be moved to `PhiloxBaseCudaArray.hpp` and `PhiloxBaseStdArray.hpp` respectively to provide optimized implementations for each backend.

PR #1529, opened 2021-12-11
fix buffer slicing test 

psychocoderHPC:
- add missing synchronization

- [ ] remove second commit (reduce test matrix)


test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -161,6 +166,9 @@ TEMPLATE_LIST_TEST_CASE("memBufSlicingTest", "[memBuf]", TestAccWithDataTypes)
         ptrNative[i] = static_cast<Data>(mappedTo1D);
     }
 
+    // Finish device tasks before comparing results.
+    slicingTest.synchronizeQueue();
```
psychocoderHPC, line 170: This is not required because we use always a blocking queue.

psychocoderHPC: Never the less I will keep this PR alive to test if our CI working. We are currently updating the CI node.

psychocoderHPC:
fix is not required because I missed that the test is using the blocking queue.

PR #1532, opened 2021-12-14
CI: gilab parallelism

psychocoderHPC:
Adjust the maximum parallelism the CI job is allowed to use based on the CI node environment variable `$CI_MAX_PARALLELISM`.

- [x] merge after #1541 (this branch is rebased against #1541 to speedup all checks), please review only the last commit!


script/gitlabci/job_base.yml
```diff
@@ -4,6 +4,8 @@
     ALPAKA_CI_BOOST_BRANCH: "boost-${ALPAKA_BOOST_VERSION}"
     BOOST_LIBRARYDIR: "/opt/boost/${ALPAKA_BOOST_VERSION}/lib"
     ALPAKA_CI_OS_NAME: "Linux"
+    ALPAKA_CI_BUILD_JOBS: "$CI_MAX_PARALLELISM"
+    OMP_NUM_THREADS: "$CI_MAX_PARALLELISM"
```
bernhardmgruber: Where are those variables coming from?

psychocoderHPC: Thats the right question! I copied it from PIConGPU and there I have the comment that this variable is set in the CI Web interface. I used the wrong variable name `CI_CPUS` should be used because this is a variable set by the HZDR CI instances.

psychocoderHPC: Updated an I added a short string above the usage where this variable is coming from.

script/gitlabci/job_base.yml
```diff
@@ -4,6 +4,9 @@
     ALPAKA_CI_BOOST_BRANCH: "boost-${ALPAKA_BOOST_VERSION}"
     BOOST_LIBRARYDIR: "/opt/boost/${ALPAKA_BOOST_VERSION}/lib"
     ALPAKA_CI_OS_NAME: "Linux"
+    # $CI_CPUS is defined by the HZDR gitlab CI
+    ALPAKA_CI_BUILD_JOBS: "$CI_CPUS"
+    OMP_NUM_THREADS: "$CI_CPUS"
```
bernhardmgruber, line 9: So `CI_CPUS` is the number of virtual cores? Funnily enough, GitHub claims they also have 2 cores on each runner, but LLAMA builds notably faster with 4 build jobs than 2. So I wondered if GitHub gives you 2 physical cores with hyperthreading. Anyhow, nothing wrong with your code here!

script/gitlabci/job_base.yml
```diff
@@ -108,4 +103,3 @@
   tags:
     - amd
     - rocm
-
```
bernhardmgruber, line 111: Unnecessary whitespace change?

bernhardmgruber (dismissed):

script/gitlabci/job_base.yml
psychocoderHPC: I removed a useless empty line, not required for the correctness but two empty lines at EOF look bad^^

bernhardmgruber: ok fine!

script/gitlabci/job_base.yml
psychocoderHPC: > So CI_CPUS is the number of virtual cores?

Yes this is the number of virtual cores the CI instance can use. I have not checked if this is a suggestion to not interrupt other instances or if all other cores are hidden. I think this depends on the node. We have OpenStack virtualized nodes for CPU and native nodes for CUDA and HIP.

bernhardmgruber (approved):

psychocoderHPC:
The runtime of the HIP 4.2 debug CI test is going down from 9:30 minutes for #1541 to 5:30 minutes.
Before #1541 the runtime was 13 minutes, so we will have soon `2.X` times faster CI tests.

j-stephan (approved):

PR #1536, opened 2021-12-14
Support zero-dimensional (scalar) buffers

### LINKED ISSUE #1535: feature request: support for 0-dimensional buffers
Many of the memory allocations and copies that we deal with in the CMS software concern individual objects, or _scalars_, instead of N-dimensional arrays.

While a _scalar_ can always be represented as an N-dimensional (usually, 1-dimensional) buffer with a single element, this requires some extra care when copying or converting between the native C++ point-of-view (where a `T` and `T[1]` are different things) and the buffer point-of-view (where they are both mapped to a 1-dimensional buffer with a single element).

fwyzard:
Implement support for zero-dimensional (_i.e._, scalar) extents, vectors, buffers, and allocations.

Specialise the `memset` and `memcpy` operations for scalar buffers for all back-ends: host CPU, CUDA/HIP, OpenMP, OpenACC.

Add a new `zeroDimBufferTest` to check that all operations are supported and behave correctly: allocations, host-device, device-device, and device-host copies, and memset.


fwyzard:
First attempt to implement support for zero-dimensional buffers.

include/alpaka/extent/Traits.hpp
```diff
@@ -83,8 +83,7 @@ namespace alpaka
                 std::index_sequence<TIndices...> const& indices) -> Idx<TExtent>
             {
                 alpaka::ignore_unused(indices);
-
-                return meta::foldr(std::multiplies<Idx<TExtent>>(), getExtent<TIndices>(extent)...);
+                return meta::foldr(std::multiplies<Idx<TExtent>>(), Idx<TExtent>(1u), getExtent<TIndices>(extent)...);
```
bernhardmgruber:
```suggestion
                return (getExtent<TIndices>(extent) * ... * Idx<TExtent>(1u));
```

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -199,6 +199,44 @@ namespace alpaka
...
+                ALPAKA_ASSERT(Scalar{} == extent);
+                auto const width = TIdx(1u);
+                auto const widthBytes = width * static_cast<TIdx>(sizeof(TElem));
+
+                // Set the current device.
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(dev.m_iDevice));
+                // Allocate the buffer on this device.
+                void* memPtr;
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
+                    ALPAKA_API_PREFIX(Malloc)(&memPtr, static_cast<std::size_t>(widthBytes)));
```
bernhardmgruber:
```suggestion
                // Set the current device.
                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(SetDevice)(dev.m_iDevice));
                // Allocate the buffer on this device.
                void* memPtr;
                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
                    ALPAKA_API_PREFIX(Malloc)(&memPtr, sizeof(TElem)));
```

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -199,6 +199,44 @@ namespace alpaka
...
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(Free)(reinterpret_cast<void*>(ptr)));
+                };
+#    if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
+                std::cout << __func__ << " ew: " << width << " ewb: " << widthBytes << " ptr: " << memPtr << std::endl;
+#    endif
+                return BufUniformCudaHipRt<TElem, DimInt<0u>, TIdx>(
+                    dev,
+                    reinterpret_cast<TElem*>(memPtr),
+                    std::move(deleter),
+                    static_cast<TIdx>(widthBytes),
```
bernhardmgruber:
```suggestion
                    static_cast<TIdx>(sizeof(TElem)),
```

bernhardmgruber (changes requested):
I would recommend a few simplifications, but otherwise the proposed change looks fine!

Now I would really like to see one or two unit tests :)

fwyzard:
> Now I would really like to see one or two unit tests :)

I wrote a test, but I'm not sure how to integrate in the existing unit tests, because - as far as I can tell - `test/unit/mem/buf/src/BufTest.cpp` gets the dimensions to be tested from outside.

Adding `alpaka::DimInt<0u>` to `include/alpaka/test/dim/TestDims.hpp` does not seem like the best idea, because I don't think it will make sense for the other tests, _e.g._ it doesn't make much sense to instantiate a 0-dimensional kernel.

bernhardmgruber:
> I wrote a test, but I'm not sure how to integrate in the existing unit tests, because - as far as I can tell - `test/unit/mem/buf/src/BufTest.cpp` gets the dimensions to be tested from outside.

You don't need to fit it into the existing testing infrastructure, which is quite cohesive. You can just define a new test case that is independent from the rest of the common infrastructure. Maybe someone else has a better suggestions, @j-stephan?

> Adding `alpaka::DimInt<0u>` to `include/alpaka/test/dim/TestDims.hpp` does not seem like the best idea, because I don't think it will make sense for the other tests, _e.g._ it doesn't make much sense to instantiate a 0-dimensional kernel.

If you cannot make all tests work with zero dimensions then don't do it. I could imagine a use case of a 0-dimensional kernel though for FGPAs or just to run a single task. At least, I don't see a reason why we should forbid it. But all of that can be handled later. Here is a ticket: #1538.

fwyzard:
Here is the test program that finally works:
```c++
#include <cstdint>
#include <iostream>

#define ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
#define ALPAKA_ACC_GPU_CUDA_ENABLED
#include <alpaka/alpaka.hpp>

constexpr int memset_value(int c) {
    int v = c & 0xFF;
    return (v | v << 8 | v << 16 | v << 24);
}

int main()
{
    using Idx = std::size_t;
    using Dim0D = alpaka::DimInt<0u>;
    using Dim1D = alpaka::DimInt<1u>;
    using Scalar = alpaka::Vec<Dim0D, Idx>;
    using Vec1D = alpaka::Vec<Dim1D, Idx>;

    // check that a Scalar extent has exactly 1 element
    Scalar scalar;
    std::cout << "Scalar size: " << scalar.prod() << '\n' << std::endl;

    // CPU host
    using Host = alpaka::DevCpu;
    auto const host = alpaka::getDevByIdx<Host>(0u);
    using HostAcc = alpaka::AccCpuSerial<Dim1D, Idx>;
    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<HostAcc>() << std::endl;
    using HostQueue = alpaka::Queue<HostAcc, alpaka::Blocking>;
    HostQueue hostQueue(host);

    // host buffer
    auto h_buffer = alpaka::allocBuf<int, Idx>(host, Scalar{});
    std::cout << "host buffer allocated at " << alpaka::getPtrNative(h_buffer) << " with " << alpaka::extent::getExtentProduct(h_buffer) << " elements" << std::endl;

    // async host buffer
    auto h_async = alpaka::allocAsyncBuf<int, Idx>(hostQueue, Scalar{});
    std::cout << "async host buffer allocated at " << alpaka::getPtrNative(h_async) << " with " << alpaka::extent::getExtentProduct(h_async) << " elements" << std::endl;

    // host-side buffer memset
    const int value1 = 42;
    std::cout << "host-side buffer memset" << std::endl;
    alpaka::memset(hostQueue, h_buffer, value1, Scalar{});
    alpaka::wait(hostQueue);
    std::cout << "check: " << *alpaka::getPtrNative(h_buffer) << " == " << memset_value(value1) << std::endl;
    assert(memset_value(value1) == *alpaka::getPtrNative(h_buffer));

    // host-side async buffer memset
    const int value2 = 99;
    std::cout << "host-side async buffer memset" << std::endl;
    alpaka::memset(hostQueue, h_async, value2, Scalar{});
    alpaka::wait(hostQueue);
    std::cout << "check: " << *alpaka::getPtrNative(h_async) << " == " << memset_value(value2) << std::endl;
    assert(memset_value(value2) == *alpaka::getPtrNative(h_async));

    // host-host copies
    std::cout << "buffer host-host copies" << std::endl;
    alpaka::memcpy(hostQueue, h_async, h_buffer, Scalar{});
    alpaka::wait(hostQueue);
    std::cout << "check: " << *alpaka::getPtrNative(h_async) << " == " << memset_value(value1) << std::endl;
    assert(memset_value(value1) == *alpaka::getPtrNative(h_async));
    alpaka::memcpy(hostQueue, h_buffer, h_async, Scalar{});
    alpaka::wait(hostQueue);
    std::cout << "check: " << *alpaka::getPtrNative(h_buffer) << " == " << memset_value(value1) << std::endl;
    assert(memset_value(value1) == *alpaka::getPtrNative(h_buffer));

    std::cout << std::endl;

    // GPU device
    using Device = alpaka::DevCudaRt;
    auto const device = alpaka::getDevByIdx<Device>(0u);
    using DeviceAcc = alpaka::AccGpuCudaRt<Dim1D, Idx>;
    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<DeviceAcc>() << std::endl;
    using DeviceQueue = alpaka::Queue<DeviceAcc, alpaka::NonBlocking>;
    DeviceQueue deviceQueue(device);

    // device buffer
    auto d_buffer = alpaka::allocBuf<int, Idx>(device, Scalar{});
    std::cout << "device buffer allocated at " << alpaka::getPtrNative(d_buffer) << " with " << alpaka::extent::getExtentProduct(d_buffer) << " elements" << std::endl;

    // async device buffer
    auto d_async = alpaka::allocAsyncBuf<int, Idx>(deviceQueue, Scalar{});
    std::cout << "async device buffer allocated at " << alpaka::getPtrNative(d_async) << " with " << alpaka::extent::getExtentProduct(d_async) << " elements" << std::endl;

    // host-device copies
    std::cout << "host-device copies" << std::endl;
    alpaka::memcpy(deviceQueue, d_buffer, h_buffer, Scalar{});
    alpaka::memcpy(deviceQueue, d_async, h_async, Scalar{});

    // device-device copies
    std::cout << "device-device copies" << std::endl;
    alpaka::memcpy(deviceQueue, d_buffer, d_async, Scalar{});
    alpaka::memcpy(deviceQueue, d_async, d_buffer, Scalar{});

    // device-side buffer memset
    std::cout << "device-side buffer memset" << std::endl;
    alpaka::memset(deviceQueue, d_buffer, value1, Scalar{});
    alpaka::memset(deviceQueue, d_async, value2, Scalar{});

    // device-host copies
    std::cout << "device-host copies" << std::endl;
    alpaka::memcpy(deviceQueue, h_buffer, d_buffer, Scalar{});
    alpaka::memcpy(deviceQueue, h_async, d_async, Scalar{});

    alpaka::wait(deviceQueue);
    std::cout << "check: " << *alpaka::getPtrNative(h_buffer) << " == " << memset_value(value1) << std::endl;
    std::cout << "check: " << *alpaka::getPtrNative(h_async) << " == " << memset_value(value2) << std::endl;
    assert(memset_value(value1) == *alpaka::getPtrNative(h_buffer));
    assert(memset_value(value2) == *alpaka::getPtrNative(h_async));

    return EXIT_SUCCESS;
}
```

Here is a local outcome:
```
Scalar size: 1

Using alpaka accelerator: AccCpuSerial<1,m>
host buffer allocated at 0x55f7c222e000 with 1 elements
async host buffer allocated at 0x55f7c2230000 with 1 elements
host-side buffer memset
check: 707406378 == 707406378
host-side async buffer memset
check: 1667457891 == 1667457891
buffer host-host copies
check: 707406378 == 707406378
check: 707406378 == 707406378

Using alpaka accelerator: AccGpuCudaRt<1,m>
device buffer allocated at 0x7fb2b5200000 with 1 elements
async device buffer allocated at 0x302000000 with 1 elements
host-device copies
device-device copies
device-side buffer memset
device-host copies
check: 707406378 == 707406378
check: 1667457891 == 1667457891
```

Can anyone suggest how to make it into a unit test ?

include/alpaka/mem/buf/cpu/Copy.hpp
```diff
@@ -177,6 +176,81 @@ namespace alpaka
...
+                printDebug();
+#endif
+                std::memcpy(
+                    reinterpret_cast<void*>(m_dstMemNative),
+                    reinterpret_cast<void const*>(m_srcMemNative),
+                    sizeof(Elem));
+            }
+
+            std::uint8_t* const m_dstMemNative;
+            std::uint8_t const* const m_srcMemNative;
```
bernhardmgruber, line 251: I would prefer `std::byte` here, but I guess alpaka uses `std::uint8_t` in many other places as well. I opened #1545 to refactor this at some point.

include/alpaka/mem/buf/uniformCudaHip/Set.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2019 Benjamin Worpitz, Erik Zenker, Matthias Werner, René Widera
+/* Copyright 2021 Benjamin Worpitz, Erik Zenker, Matthias Werner, René Widera,
+ *                Andrea Bocci
```
bernhardmgruber: Why the line break here? Did clang-format do this?

bernhardmgruber:
@fwyzard Since it is a rather composite test, you could add it as a new integration test. E.g. in alpaka/test/integ/zeroDimBuffer. Just copy one of the other integration test CMakeLists.txt. Proper integration and extraction of more fine grained unit tests can then be handled by our professional support.

include/alpaka/mem/buf/uniformCudaHip/Set.hpp
fwyzard: I copied the line break length from `include/alpaka/mem/buf/uniformCudaHip/Copy.hpp`.
I can check if `clang-format` is happy with having all names on the same line and use that.
Which one do you prefer ?

bernhardmgruber: I would keep it on the same line.

include/alpaka/mem/buf/cpu/Copy.hpp
fwyzard: I've had mixed experience with `std::byte`, but yes, I agree it would be the type whose intent best matches the use case.
However, I have not seen it used anywhere in Alpaka so far.

include/alpaka/mem/buf/uniformCudaHip/Set.hpp
fwyzard: `clang-format` is happy to have all names on one line.

Do you prefer 
```suggestion
/* Copyright 2021 Benjamin Worpitz, Erik Zenker, Matthias Werner, René Widera, Andrea Bocci
```
?

fwyzard: I've put all names back on one line.

fwyzard:
The new integration test works for me locally... let's see how it goes through the CI

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,137 @@
...
+    using Device = alpaka::Dev<DeviceAcc>;
+    using DeviceQueue = alpaka::Queue<DeviceAcc, alpaka::NonBlocking>;
+
+    using HostAcc = alpaka::AccCpuSerial<Dim1D, Idx>;
+    using Host = alpaka::DevCpu;
+    using HostQueue = alpaka::Queue<HostAcc, alpaka::Blocking>;
+
+    // check that a Scalar extent has exactly 1 element
+    Scalar scalar;
+    INFO("Scalar size: " << scalar.prod())
```
bernhardmgruber, line 106: Do we need such text output? I would rather replace it by a `CHECK(scalar.prod() == 1)`.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include <catch2/catch.hpp>
+
+#include <cstdint>
+#include <iostream>
+
+constexpr int memset_value(int c)
+{
+    int v = c & 0xFF;
+    return (v | v << 8 | v << 16 | v << 24);
+}
```
bernhardmgruber: This code assumes that `int` is 32bit. Can you please add a `static_assert` for that? Or write the code generically.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,137 @@
...
+    INFO(
+        "async host buffer allocated at " << alpaka::getPtrNative(h_async) << " with "
+                                          << alpaka::extent::getExtentProduct(h_async) << " elements")
+
+    // host-side buffer memset
+    const int value1 = 42;
+    INFO("host-side buffer memset")
+    alpaka::memset(hostQueue, h_buffer, value1, Scalar{});
+    alpaka::wait(hostQueue);
+    INFO("check: " << *alpaka::getPtrNative(h_buffer) << " == " << memset_value(value1))
```
bernhardmgruber: I would just remove that line. The `CHECK` is good enough.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,137 @@
...
+    REQUIRE(memset_value(value1) == *alpaka::getPtrNative(h_buffer));
```
bernhardmgruber:
```suggestion
    CHECK(memset_value(value1) == *alpaka::getPtrNative(h_buffer));
```
Unless you really need the test to abort when this is wrong, use `CHECK` please.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Yes, I prefer to have it.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Fell free to open a PR after this one has been merged.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: `int` is 32 bits on any single architecture I have access to.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Personally, I don't need the test at all. I have been asked to write one, without any indications or instructions as to how to do it, and this is what I could came up with.

I **strongly** suggest you start working on some documentation on how to add tests and run the existing ones, if you are interested in have any contributions to Alpaka.

fwyzard: Anyway, sure, I can replace everywhere
```c++
    INFO("check: " << *alpaka::getPtrNative(h_buffer) << " == " << memset_value(value1))
    REQUIRE(memset_value(value1) == *alpaka::getPtrNative(h_buffer));
```
with
```c++
    CHECK(memset_value(value1) == *alpaka::getPtrNative(h_buffer));
```

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Changed to support any arbitrary trivial type.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Change to `CHECK`, assuming people don't look at the messages anyway.

bernhardmgruber (dismissed):

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,175 @@
...
+    HostQueue hostQueue(host);
+
+    // host buffer
+    auto h_buffer1 = alpaka::allocBuf<int, Idx>(host, Scalar{});
+    INFO(
+        "host buffer allocated at " << alpaka::getPtrNative(h_buffer1) << " with "
+                                    << alpaka::extent::getExtentProduct(h_buffer1) << " element(s)")
+
+    // async host buffer
+    auto h_buffer2 = alpaka::allocAsyncBuf<int, Idx>(hostQueue, Scalar{});
```
psychocoderHPC: This is currently triggering a static assert within alpaka see [here](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1897665119#L3637) because this feature needs CUDA 11.2.

The solution from line 142 should solve the issue:

```
auto d_buffer2 = isAsyncBufferSupported<DeviceAcc>() ? alpaka::allocAsyncBuf<int, Idx>(deviceQueue, Scalar{})
                                                         : alpaka::allocBuf<int, Idx>(device, Scalar{});
```

psychocoderHPC (changes requested):

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Mhm, this is unexpected.
`hostQueue` refers to the CPU serial backend, for which `allocAsyncBuf` should always be defined.

psychocoderHPC: I think the specialization for zero-dimensional host allocation is missing and therefore is falling back to the device specialization.

psychocoderHPC: Mhh there should be no need for the specialization.

psychocoderHPC: I think I know the problem:
https://github.com/alpaka-group/alpaka/blob/5e008e48cf079c81593788713f12ed114291beaa/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp#L342-L344

The compiler is allowed to evaluate the static assert within the class/struct at any time, not only if an instance of this object is created. The check must be deferred.

```
            static_assert(
                sizeof(TElem) != 0 && meta::DependentFalseType<TElem>::value,
                "Support for stream-ordered memory buffers requires CUDA 11.2 or higher.");
```

Checking the size of a template type will defer the check. The check for HIP few lines should be deferred too but maybe clang is not evaluating the class before the instanziation.

psychocoderHPC: Could you add this check as PR within your PR. I can on Monday open a separate PR we can to fix the error.

psychocoderHPC: Ok I missed that the assert is using a dependent fals, mmhh.

fwyzard: I've setup a local test area with CUDA 11.0, and I can reproduce the error... investigating...

fwyzard: I suspect it happens because both branches of the ternary `condition ? if_true : if_false` operator are compiled.

fwyzard:
The latest change should fix [the warning](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1898187253#L2360) from `linux_nvhpc-21.9_cuda-11.4_oacc`:
```
"/builds/hzdr/crp/alpaka/test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp", line 75: warning: statement is unreachable
      return alpaka::allocBuf<TElem, TIdx>(alpaka::getDev(queue), extent);
```

psychocoderHPC:
@fwyzard Please ignore the timeout of the HIP runtime tests. The GPU stucks again :-(, a support ticket is opened.

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,185 @@
...
+#endif // ALPAKA_ACC_ANY_BT_OACC_ENABLED
+
+#ifdef ALPAKA_ACC_ANY_BT_OMP5_ENABLED
+    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevOmp5>)
+    {
+        return false;
+    }
+#endif // ALPAKA_ACC_ANY_BT_OMP5_ENABLED
+
+    return true;
```
psychocoderHPC: The CI is saying that this statement is not reachable if `OpenAcc` is enabled.

psychocoderHPC (changes requested):

fwyzard:
(rebased on top of the HEAD of the develop branch)

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: Well, yes:
  - `#ifdef ALPAKA_ACC_ANY_BT_OACC_ENABLED` will be true, so the corresponding `if constexpr` is checked;
  - `if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevOacc>)` will be true, so the corresponding `return false` is executed;
  - the `return true` at the end of the function is not reachable.

I would say that this is really by construction. How does one silence the corresponding warning or error ?

fwyzard:
The error in the CI's [linux_hip4.2_clang_debug_hip_only](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1903893851) tests seem unrelated to this PR:
```
26: /builds/hzdr/crp/alpaka/test/unit/kernel/src/KernelGenericLambda.cpp:19: FAILED:
26: due to a fatal error condition:
26:   SIGABRT - Abort (abnormal termination) signal
26: 
26: ===============================================================================
26: test cases: 2 | 1 passed | 1 failed
26: assertions: 2 | 1 passed | 1 failed
```

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
psychocoderHPC: Maye, not nice but the only way which coming up to my mind at the moment.

```C++
bool result = true;
# ifdef ACC1
if constexpr ...
  result = false;
#endif
# ifdef ACC2
if constexpr ...
  result = true;
#endif
return result;
```

psychocoderHPC:
> The error in the CI's [linux_hip4.2_clang_debug_hip_only](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1903893851) tests seem unrelated to this PR:
> 
> ```
> 26: /builds/hzdr/crp/alpaka/test/unit/kernel/src/KernelGenericLambda.cpp:19: FAILED:
> 26: due to a fatal error condition:
> 26:   SIGABRT - Abort (abnormal termination) signal
> 26: 
> 26: ===============================================================================
> 26: test cases: 2 | 1 passed | 1 failed
> 26: assertions: 2 | 1 passed | 1 failed
> ```

I will review the failing test tomorrow, maybe this is a hidden bug (I assume independent of this PR) or a very instable AMD driver:

```
26: -------------------------------------------------------------------------------
26: genericLambdaKernelIsWorking - alpaka::test::TestAccs - 1
26: -------------------------------------------------------------------------------
26: /builds/hzdr/crp/alpaka/test/unit/kernel/src/KernelGenericLambda.cpp:19
26: ...............................................................................
26: 
26: /builds/hzdr/crp/alpaka/test/unit/kernel/src/KernelGenericLambda.cpp:19: FAILED:
26: due to a fatal error condition:
26:   SIGABRT - Abort (abnormal termination) signal
26: 
26: ===============================================================================
26: test cases: 2 | 1 passed | 1 failed
26: assertions: 2 | 1 passed | 1 failed
26: 
26/39 Test #26: kernelTest .......................Subprocess aborted***Exception:   0.55 sec
```

fwyzard:
All CI test failure are due to missing Boost 1.74: 
> ```
> CMake Error at C:/Program Files/CMake/share/cmake-3.20/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
>   Could NOT find Boost (missing: Boost_INCLUDE_DIR) (Required is at least
>   version "1.74.0")
> ```

except for one that failed to `git clone`:
> ```
> error: RPC failed; curl 56 OpenSSL SSL_read: Connection was reset, errno 10054
> error: 2326 bytes of body are still expected
> fetch-pack: unexpected disconnect while reading sideband packet
> Error: fatal: early EOF
> Error: fatal: fetch-pack: invalid index-pack output
> Error: Git fetch failed with exit code: 128
> Error: Exit code 1 returned from process: file name 'c:\runners\2.285.1\bin\Runner.PluginHost.exe', arguments 'action "GitHub.Runner.Plugins.Repository.v1_0.CheckoutTask, Runner.Plugins"'.
> ```

AFAICT, all tests that did run were successful.

fwyzard:
Looking deeper, it seems that the failures due to missing boost 1.74 can be traced back to a similar `got clone` problem:
```
2021-12-26T22:31:44.6996464Z +++ git clone -b boost-1.74.0 --quiet --recursive --single-branch --depth 1 https://github.com/boostorg/boost.git 'D:\a\alpaka\alpaka/boost'
...
2021-12-26T22:31:45.2906413Z 
2021-12-26T22:39:43.2730802Z error: RPC failed; curl 56 OpenSSL SSL_read: Connection was reset, errno 10054
2021-12-26T22:39:43.2732950Z error: 6436 bytes of body are still expected
2021-12-26T22:39:43.2734110Z fetch-pack: unexpected disconnect while reading sideband packet
2021-12-26T22:39:43.2735089Z fatal: early EOF
2021-12-26T22:39:43.2735887Z fatal: fetch-pack: invalid index-pack output
2021-12-26T22:39:43.2737360Z fatal: clone of 'https://github.com/boostorg/accumulators.git' into submodule path 'D:/a/alpaka/alpaka/boost/libs/accumulators' failed
```

fwyzard:
(no changes, force-pushed just to re-trigger the tests)

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
psychocoderHPC: @fwyzard How do you solved the issue with the `unreachable code`? 
Somehow you solved it because the CI is not passing but I can not see why, is it because of line 12?

psychocoderHPC (dismissed):

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
fwyzard: I don't remember having made changes here - but in the meantime I did fix the OpenMP 5 and OpenAcc `memcpy` and `memset` implementations for zero-dimensional buffers.

psychocoderHPC:
I restarted github actions in the hope that the windows boost 1.74.0 issue is gone.

fwyzard:
:+1:

test/integ/zeroDimBuffer/src/zeroDimBuffer.cpp
```diff
@@ -0,0 +1,187 @@
...
```
j-stephan: To silence the unreachable code warning one would have to make the following messy change:

```suggestion
#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevCudaRt>)
    {
        return (BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)) && (alpaka::Dim<TAcc>::value == 1);
    }
    else
#endif // ALPAKA_ACC_GPU_CUDA_ENABLED

#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevHipRt>)
    {
        return false;
    }
    else
#endif // ALPAKA_ACC_GPU_HIP_ENABLED

#ifdef ALPAKA_ACC_ANY_BT_OACC_ENABLED
    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevOacc>)
    {
        return false;
    }
    else
#endif // ALPAKA_ACC_ANY_BT_OACC_ENABLED

#ifdef ALPAKA_ACC_ANY_BT_OMP5_ENABLED
    if constexpr(std::is_same_v<alpaka::Dev<TAcc>, alpaka::DevOmp5>)
    {
        return false;
    }
    else
#endif // ALPAKA_ACC_ANY_BT_OMP5_ENABLED

   return true;
```

This might be a rare case of `if constexpr` being uglier than just having template specializations.

j-stephan (changes requested):
See my proposal above. Afterwards we should be good to go.

fwyzard:
Back from the winter break... I've rebased on top of `develop` and applied your suggestions, let's see if all compilers are happy with it.

fwyzard:
There seems to be [a problem](https://gitlab.com/hzdr/crp/alpaka/-/pipelines/441728450) with the CUDA GitLab runners:
```
12: /builds/hzdr/crp/alpaka/test/integ/axpy/src/axpy.cpp:69: FAILED:
12: due to unexpected exception with message:
12:   Unable to return device handle for device 0. There are only 0 devices!
```

j-stephan:
Yes, we've been observing this since yesterday evening. I've opened a support ticket and am waiting for someone to pick it up...

j-stephan:
The support rebooted the failing CI runner. Hopefully things will work now.

j-stephan (approved):

j-stephan:
Thanks!

Issue #1517, opened 2021-12-07
Improve CI efficiency


j-stephan:
Discussed offline with @bernhardmgruber: The current way of running the CI every time when a PR is merged is very inefficient. We just tested the code base against the PR, why test it again? This unnecessarily blocks the CI for a few hours.

Instead it might be more useful to run the `develop` CI on a regular basis. Maybe once per week or once every other day. Thoughts?

psychocoderHPC:
The CI is running for the dev branch after a merge because a PR can be based on an old development branch. 
If you merge two PRs in a row we typically rebase these PRs via github to the latest development branch this could introduce issues even if there is no merge conflict. If so you know which PR is the root o the issue. It is hard to say which PR is breaking the development branch if you test the code only once per week.

psychocoderHPC:
To reduce the load I suggest testing for PRs and merging development fewer combinations per compiler and performing a full matrix test once per week. This should catch most issues when the PR is opened.

psychocoderHPC:
Another way to reduce the CI load is by staging the tests. This means you test first a handful of combinations and only on successes perform more tests.
For example, this is what we do for PIConGPU: https://gitlab.com/hzdr/crp/picongpu/-/pipelines/424201783

j-stephan:
Also a solution to #1439 might help.

psychocoderHPC:
for the record: PIConGPU is using a CI job generator to avoid that a combination being tested twice. https://github.com/ComputationalRadiationPhysics/picongpu/blob/dev/share/ci/n_wise_generator.py
The generator is spaghetti phython but is only running valid combinations e.g. host compiler + CUDA, .... 
With a generator is is easy to generate different complex matrices, e.g if we do not want to run a full matrix per PR.

psychocoderHPC:
To reduce the CI job runtime we should maybe run the header include check called `headerCheck` in a separate job and only once for each compiler e.g. gcc, clang, hipp, nvcc, intel. There is a very small possibility we miss something if we have a workaround within the code that is only activated for a specific compiler version.
IMO this risk is very tiny but the saved runtime is gigantic.
We could run scheduled tests once per week/month where we run the header check within each test. In a case where we missed an include the fix is very easy and can be done without deep knowledge about the PR where the missing include was introduced.

bernhardmgruber:
The header checks are already only run as part of the `analysis` CI runs. All other CI runs should not run header checks.

psychocoderHPC:
[...]

PR #1537, opened 2021-12-15
CI: group tests in stages

psychocoderHPC:
- Group tests in stages to reduce the CI pressure. In cases where a stage is failing it makes no sense to run all others.
- Additionally, the second commit set the environment variable `ALPAKA_CI` to avoid running the `headerCheck` if not explicitly activated.
- Disable `randomCells2D` for the OpenACC backend because of invalid memory access. The reason for the issue should be investigated independently to avoid blocking other PRs.

Per YAML file I incremented the stage id, I have not balanced the stages by hand because this would make it hard to keep an overview.

Writing everything by hand should be only a workaround we need asap a generator https://github.com/alpaka-group/alpaka/issues/1517#issuecomment-988044352


bernhardmgruber:
I see `nvhpc` in the last 8th stage of the pipeline and wonder whether we should have it earlier to avoid running 7 stages to then fail in the 8th because of a very unique compiler.

psychocoderHPC:
> I see `nvhpc` in the last 8th stage of the pipeline and wonder whether we should have it earlier to avoid running 7 stages to then fail in the 8th because of a very unique compiler.

I also thought about it, if so we should reorder the yaml file stay with simple indexing within the file. We could do it in a separate PR but should finish first our open PRs.

example/CMakeLists.txt
```diff
@@ -36,6 +36,9 @@ add_subdirectory("kernelSpecialization/")
 add_subdirectory("monteCarloIntegration/")
 add_subdirectory("openMPSchedule/")
 add_subdirectory("randomStrategies/")
-add_subdirectory("randomCells2D/")
+# This example is producing invalid memory access during CI tests with the OpenACC backend.
+if(NOT ALPAKA_CI OR NOT ALPAKA_ACC_ANY_BT_OACC_ENABLE)
+    add_subdirectory("randomCells2D/")
+endif()
```
bernhardmgruber, line 42: This needs to be fixed at some point (discussed offline): #1540.

.gitlab-ci.yml
```diff
@@ -1,3 +1,14 @@
+stages:
+  - stage0
+  - stage1
+  - stage2
+  - stage3
+  - stage4
+  - stage5
+  - stage6
+  - stage7
+  - stage8
```
bernhardmgruber, line 10: I am worried that the pipeline is too deep and CI builds take too long. But we will see.

bernhardmgruber (approved):

example/CMakeLists.txt
psychocoderHPC: Thanks for opening the issue. It was to late yesterday night to do it.

PR #1539, opened 2021-12-15
Add cmake option ALPAKA_CHECK_HEADERS

bernhardmgruber:
OFF by default. Enabled by CI when ALPAKA_CI_ANALYSIS is ON.


bernhardmgruber:
Checking a few CI logs, it seems the behavior is correct and the header check is only run in the analysis jobs.

psychocoderHPC:
I stopped the gitlab CI for this PR to prefer #1537.

psychocoderHPC (approved):

PR #1543, opened 2021-12-17
Format tables in README.md

bernhardmgruber:
This PR formats the tables in README.md. This increases readability of the markdown code.


psychocoderHPC:
This PR contains only changes in the README, we do not have a way to skip compile tests in such cases, therefore I will merge this PR already before the CI is ready.

PR #1544, opened 2021-12-17
Fix Boost version docs and remove redundant check in Boost installation

bernhardmgruber:
I just saw that we forgot to update the minimum required Boost verison in the README.md and looked for checks we no longer need to do. This lead to a drive-by fix in the Boost installation script.

### DIFF
--- a/script/install_boost.sh
+++ b/script/install_boost.sh
@@ -99,7 +99,6 @@ then
         ALPAKA_BOOST_B2+=" --layout=tagged --toolset=${CC}"
     fi
 
-    # TODO: Win32: adress-model=32
     ALPAKA_BOOST_B2+=" architecture=x86 address-model=64 link=static threading=multi runtime-link=shared"
 
     if [ "$ALPAKA_CI_OS_NAME" = "Windows" ]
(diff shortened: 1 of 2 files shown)


script/install_boost.sh
```diff
@@ -120,16 +119,12 @@ then
     then
         ALPAKA_BOOST_B2_CXXFLAGS+=" -Wunused-private-field -Wno-unused-local-typedef -Wno-c99-extensions -Wno-variadic-macros"
     fi
-    # Select the libraries required.
-    # If the variable is not set, the backend will most probably be used by default so we install it.
-    if [ "${ALPAKA_CI_INSTALL_FIBERS}" == "ON" ]
```
psychocoderHPC, line 125: Why do you remove this check?

In cases where fiber is enabled we perform a real complex installation of boost. We do not want to have fiber installed in each check. This saves us a lot of CI testing time.

psychocoderHPC (changes requested):

script/install_boost.sh
psychocoderHPC: IN cases we have not fiber enabled we do not need boost atomics, threads, ...

bernhardmgruber: Because the same check is already on line 69 and thus has no effect.

psychocoderHPC (approved):

PR #1549, opened 2021-12-21
Fix pitch access calculation in randomCells2D example.

### LINKED ISSUE #1540: Fix randomCells2D example for OpenACC
The randomCells2D example fails during the CI for the OpenACC backend due to a memory related bug. The example has been disabled in CI as part of #1537 and needs to be fixed and reenabled.

sliwowitz:
As reported in issue #1540, the randomCells2D example would fail if the pitch were not a multiple of the size of the RNG state. This PR should fix that. Additionally, it introduces a ballpark test for convergence, returning non-zero and warning the user if the result is way off (over 10 times the expected 1/sqrt(N) error).


sliwowitz:
1. Windows `nvcc` found a wrong `__device__` instead of `__host__ __device__` signature of the helper function (first I had it in the kernels only, then I wanted to use the helper in the host code for consistency and forgot to change this). Weird that on Linux, `nvcc` didn't complain
2. [linux_nvhpc-21.9_cuda-11.4_oacc](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1907538789) fails on gitlab CI with `call to cuStreamSynchronize returned error 700: Illegal address during kernel execution`. I suppose this is still the same error as in the original report

sliwowitz:
I'll set up the failing environment and try to debug the issue locally.

psychocoderHPC:
> 2\. [linux_nvhpc-21.9_cuda-11.4_oacc](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1907538789) fails on gitlab CI with `call to cuStreamSynchronize returned error 700: Illegal address during kernel execution`. I suppose this is still the same error as in the original report

I opened issue #1550 for the OpenAcc issue we saw this error already in other PRs.

sliwowitz:
I didn't manage to locally recreate the exact environment as used by the Gitlab CI, but at least I managed to install `nvhpc` version 22.1 (newer than the 21.9 that we have on gitlab). Instead of `call to cuStreamSynchronize returned error 700: Illegal address during kernel execution` we've seen here, the example finished but with a wrong result (something like 0.000002 instead of approx. 0.5). Before getting down to debug this, I rebased this PR on current develop. Suddenly, the example started to produce the correct result. I pushed the rebased update here - maybe it will pass the CI too.

j-stephan (approved):

PR #1554, opened 2022-01-03
Use generic container interfaces

### LINKED ISSUE #1498: Use C++17 `std::data()`.
In `mem/view/Traits.hpp` we find the following section:

```c++
//! Creates a view to a device std::vector
//!
//! \param dev Device from where vec can be accessed.
//! \param vec std::vector container. The data hold by the container the must be accessible from the given device.
//! \return A view to device memory.
template<typename TDev, typename TElem, typename TAllocator>
auto createView(TDev const& dev, std::vector<TElem, TAllocator>& vec)
{
    //! \todo With C++17 use std::data() and add support for all contiguous container
    return createView(dev, vec.data(), extent::getExtent(vec));
}
```

This can now be addressed.

j-stephan:
Fixes #1498.

This PR replaces all occurrences of `Container.{begin,end,data,size}` with `std::{begin,end,data,size}`. In some instances it was even possible to replace a `begin`-`end` pair with a ranged `for` loop.

The `createView` functions are now a bit more generic as they now accept any container which has a `data()` member function and for which the `GetExtent` trait is defined.

### DIFF
--- a/include/alpaka/mem/view/Traits.hpp
+++ b/include/alpaka/mem/view/Traits.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2021 Axel Huebl, Benjamin Worpitz, Matthias Werner, Andrea Bocci
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan
  *
  * This file is part of alpaka.
  *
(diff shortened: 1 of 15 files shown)


include/alpaka/mem/view/Traits.hpp
```diff
@@ -435,55 +435,30 @@ namespace alpaka
...
+    //! \param extent Number of elements held by the container. Using a multi-dimensional extent will result in a
+    //!               multi-dimensional view to the memory represented by the container.
     //! \return A view to device memory.
-    template<typename TDev, typename TElem, std::size_t Tsize, typename TExtent>
-    auto createView(TDev const& dev, std::array<TElem, Tsize>& array, TExtent const& extent)
+    template<typename TDev, typename TContainer, typename TExtent>
+    auto createView(TDev const& dev, TContainer& con, TExtent const& extent)
     {
-        return createView(dev, array.data(), extent);
+        return createView(dev, std::data(con), extent);
```
sbastrakov, line 461: For more generic use, probably ADL is needed here?
```suggestion
        using std::data;
        return createView(dev, data(con), extent);
```

j-stephan: I'm not sure. Do we want to support containers that require their own (freestanding) `data()` call instead of being compatible with `std::data`?

sbastrakov: Ah, you are right. It's fine as is.

sbastrakov (approved):

psychocoderHPC (approved):

Issue #1297, opened 2021-04-20
compile time issues with nvcc and GCC 10.3


psychocoderHPC:
The CI is showing compile time issues with nvcc and GCC 10.3

```
/usr/include/c++/10/chrono: In substitution of 'template<class _Rep, class _Period> template<class _Period2> using __is_harmonic = std::__bool_constant<(std::ratio<((_Period2::num / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::num, _Period::num)) * (_Period::den / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::den, _Period::den))), ((_Period2::den / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::den, _Period::den)) * (_Period::num / std::chrono::duration<_Rep, _Period>::_S_gcd(_Period2::num, _Period::num)))>::den == 1)> [with _Period2 = _Period2; _Rep = _Rep; _Period = _Period]':
/usr/include/c++/10/chrono:473:154:   required from here
/usr/include/c++/10/chrono:428:27: internal compiler error: Segmentation fault
  428 |  _S_gcd(intmax_t __m, intmax_t __n) noexcept
      |                           ^~~~~~
Please submit a full bug report,
with preprocessed source if appropriate.
See <file:///usr/share/doc/gcc-10/README.Bugs> for instructions.
```

see:
 - https://github.com/alpaka-group/alpaka/pull/1294/checks?check_run_id=2384249915#step:5:3859
 - https://github.com/alpaka-group/alpaka/pull/1294/checks?check_run_id=2384249448#step:5:3876

A similar issue was reported:
- https://github.com/NVIDIA/nccl/issues/494
- https://www.mail-archive.com/search?l=gcc@gcc.gnu.org&q=subject:%22ICE%22&o=newest&f=1

GCC bug reports:

https://github.com/alpaka-group/alpaka/issues/1297#issuecomment-829441210

I am not sure but the error could be fixed with https://gcc.gnu.org/bugzilla/show_bug.cgi?id=100101

**ATTENTION** Currently two tests in the CI are disabled: #1302

SimeonEhrig:
CUDA 11.3 does only supports up to GCC 9.x: https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html
I had also some problems with GCC 10 and CUDA.

psychocoderHPC:
[...]

PR #1556, opened 2022-01-03
Replace `enable_if` with `if constexpr`

### LINKED ISSUE #1500: Use `if constexpr`
Since we heavily rely on template metaprogramming we should investigate where we can (sensibly) use `if constexpr` instead of `std::enable_if` and friends.

j-stephan:
Fixes #1500.

This PR replaces a number of `std::enable_if` instances with their `if constexpr` equivalents where it makes sense.

### DIFF
--- a/include/alpaka/core/Assert.hpp
+++ b/include/alpaka/core/Assert.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Axel Huebl, Benjamin Worpitz, Matthias Werner
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan
  *
  * This file is part of alpaka.
  *
@@ -37,29 +37,20 @@ namespace alpaka
     {
         namespace detail
         {
-            template<typename TArg, typename TSfinae = void>
-            struct AssertValueUnsigned;
             template<typename TArg>
-            struct AssertValueUnsigned<TArg, std::enable_if_t<!std::is_unsigned<TArg>::value>>
+            struct AssertValueUnsigned
             {
-                ALPAKA_NO_HOST_ACC_WARNING
-                ALPAKA_FN_HOST_ACC static auto assertValueUnsigned(TArg const& arg) -> void
+                ALPAKA_NO_HOST_ACC_WARNING ALPAKA_FN_HOST_ACC static auto assertValueUnsigned(TArg const& arg)
                 {
                     alpaka::ignore_unused(arg);
-                    ALPAKA_ASSERT_OFFLOAD(arg >= 0);
-                }
-            };
-            template<typename TArg>
-            struct AssertValueUnsigned<TArg, std::enable_if_t<std::is_unsigned<TArg>::value>>
-            {
-                ALPAKA_NO_HOST_ACC_WARNING
-                ALPAKA_FN_HOST_ACC static auto assertValueUnsigned(TArg const& arg) -> void
-                {
-                    alpaka::ignore_unused(arg);
+                    if constexpr(std::is_signed_v<TArg>)
+                        ALPAKA_ASSERT_OFFLOAD(arg >= 0);
+
                     // Nothing to do for unsigned types.
                 }
             };
         } // namespace detail
+
         //! This method checks integral values if they are greater or equal zero.
         //! The implementation prevents warnings for checking this for unsigned types.
         ALPAKA_NO_HOST_ACC_WARNING
@@ -71,41 +62,26 @@ namespace alpaka
 
         namespace detail
         {
-            template<typename TLhs, typename TRhs, typename TSfinae = void>
-            struct AssertGreaterThan;
             template<typename TLhs, typename TRhs>
-            struct AssertGreaterThan<
-                TLhs,
-                TRhs,
-                std::enable_if_t<!std::is_unsigned<TRhs>::value || (TLhs::value != 0u)>>
+            struct AssertGreaterThan
             {
-                ALPAKA_NO_HOST_ACC_WARNING
-                ALPAKA_FN_HOST_ACC static auto assertGreaterThan(TRhs const& lhs) -> void
+                ALPAKA_NO_HOST_ACC_WARNING ALPAKA_FN_HOST_ACC static auto assertGreaterThan(TRhs const& rhs)
                 {
-                    alpaka::ignore_unused(lhs);
-                    ALPAKA_ASSERT_OFFLOAD(TLhs::value > lhs);
-                }
-            };
-            template<typename TLhs, typename TRhs>
-            struct AssertGreaterThan<
-                TLhs,
-                TRhs,
-                std::enable_if_t<std::is_unsigned<TRhs>::value && (TLhs::value == 0u)>>
-            {
-                ALPAKA_NO_HOST_ACC_WARNING
-                ALPAKA_FN_HOST_ACC static auto assertGreaterThan(TRhs const& lhs) -> void
-                {
-                    alpaka::ignore_unused(lhs);
-                    // Nothing to do for unsigned types camparing to zero.
+                    alpaka::ignore_unused(rhs);
+                    if constexpr(std::is_signed_v<TRhs> || (TLhs::value != 0u))
+                        ALPAKA_ASSERT_OFFLOAD(TLhs::value > rhs);
+
+                    // Nothing to do for unsigned types comparing to zero.
                 }
             };
         } // namespace detail
-        //! This method asserts that the integral value TArg is less than Tidx.
+
+        //! This function asserts that the integral value TLhs is greater than TRhs.
         ALPAKA_NO_HOST_ACC_WARNING
         template<typename TLhs, typename TRhs>
-        ALPAKA_FN_HOST_ACC auto assertGreaterThan(TRhs const& lhs) -> void
+        ALPAKA_FN_HOST_ACC auto assertGreaterThan(TRhs const& rhs) -> void
         {
-            detail::AssertGreaterThan<TLhs, TRhs>::assertGreaterThan(lhs);
+            detail::AssertGreaterThan<TLhs, TRhs>::assertGreaterThan(rhs);
         }
     } // namespace core
 } // namespace alpaka
--- a/include/alpaka/core/Decay.hpp
+++ b/include/alpaka/core/Decay.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2020 Sergei Bastrakov
+/* Copyright 2022 Sergei Bastrakov, Jan Stephan
  *
  * This file is part of alpaka.
  *
@@ -29,3 +29,19 @@
 #else
 #    define ALPAKA_DECAY_T(Type) std::decay_t<Type>
 #endif
+
+namespace alpaka
+{
+    //! Provides a decaying wrapper around std::is_same. Example: is_decayed_v<volatile float, float> returns true.
+    /* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because these
+    /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+    template<typename T, typename U>
+    inline constexpr auto is_decayed_v = std::is_same_v<ALPAKA_DECAY_T(T), ALPAKA_DECAY_T(U)>;
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
+} // namespace alpaka
--- a/include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Axel Huebl, Benjamin Worpitz, Bert Wesarg
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Bert Wesarg, Jan Stephan
  *
  * This file is part of alpaka.
  *
@@ -12,6 +12,7 @@
 #if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
 
 #    include <alpaka/core/CudaHipMath.hpp>
+#    include <alpaka/core/Decay.hpp>
 #    include <alpaka/core/Unused.hpp>
 #    include <alpaka/math/abs/Traits.hpp>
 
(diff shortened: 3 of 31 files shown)


include/alpaka/math/abs/AbsUniformCudaHipBuiltIn.hpp
```diff
@@ -30,32 +30,22 @@ namespace alpaka
...
+
+                    if constexpr(std::is_same_v<TArg, float>)
+                        return ::fabsf(arg);
+                    else if constexpr(std::is_same_v<TArg, double>)
+                        return ::fabs(arg);
+                    else if constexpr(std::is_same_v<TArg, int>)
+                        return ::abs(arg);
+                    else if constexpr(std::is_same_v<TArg, long int>)
+                        return ::labs(arg);
+                    else
```
psychocoderHPC, line 50: IMO we should use `else if constexpr(std::is_same_v<TArg, long long int>)` and static assert at the end of the function to avoid any side effects, e.g. if `is_siogned` is defined for an data type but it is not `long long int`

include/alpaka/math/acos/AcosUniformCudaHipBuiltIn.hpp
```diff
@@ -30,22 +30,16 @@ namespace alpaka
...
-            struct Acos<AcosUniformCudaHipBuiltIn, float>
-            {
-                __device__ auto operator()(AcosUniformCudaHipBuiltIn const& acos_ctx, float const& arg) -> float
-                {
-                    alpaka::ignore_unused(acos_ctx);
-                    return ::acosf(arg);
+                    if constexpr(std::is_same_v<TArg, float>)
+                        return ::acosf(arg);
+                    else
+                        return ::acos(arg);
```
psychocoderHPC, line 43: Same as above. IMO we should always be explicit and assert for other data types. This would avoid implicit casts e.g. if the user is using `long double`

psychocoderHPC (changes requested):
IMO we should be always explicit and not ise 

```
if condition
  ...
else
  ...
```
else I am fine with this PR.

include/alpaka/math/acos/AcosUniformCudaHipBuiltIn.hpp
j-stephan: `long double` is treated as `double` in CUDA device code (nvcc warns about this). Otherwise all `is_floating_point` cases should be handled, right?

psychocoderHPC: We do not have only one CUDA compiler but the same issue is happening with `half`, ...

include/alpaka/math/max/MaxStdLib.hpp
```diff
@@ -27,32 +27,21 @@ namespace alpaka
...
                 {
                     alpaka::ignore_unused(max_ctx);
+
                     using std::fmax;
-                    return fmax(x, y);
+                    using std::max;
+
+                    if constexpr(std::is_integral_v<Tx> && std::is_integral_v<Ty>)
+                        return max(x, y);
+                    else // Floating point or mixed integral-floating point
```
j-stephan: @psychocoderHPC How do we want to handle the standard library math functions? Should we impose the same restrictions as for CUDA devices (no `long double`)?

include/alpaka/math/rsqrt/RsqrtUniformCudaHipBuiltIn.hpp
```diff
@@ -30,22 +30,16 @@ namespace alpaka
         {
             //! The CUDA rsqrt trait specialization.
             template<typename TArg>
-            struct Rsqrt<RsqrtUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_arithmetic<TArg>::value>>
+            struct Rsqrt<RsqrtUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_arithmetic_v<TArg>>>
```
j-stephan, line 34: @psychocoderHPC 

The root functions (`cbrt`, `sqrt`, `rsqrt`) are the only floating point operations for which we allow integral parameters, too. Is this intended or an accident?

include/alpaka/math/acos/AcosUniformCudaHipBuiltIn.hpp
```diff
@@ -30,22 +30,16 @@ namespace alpaka
...
-            };
 
-            template<>
-            struct Acos<AcosUniformCudaHipBuiltIn, float>
-            {
-                __device__ auto operator()(AcosUniformCudaHipBuiltIn const& acos_ctx, float const& arg) -> float
-                {
-                    alpaka::ignore_unused(acos_ctx);
-                    return ::acosf(arg);
+                    if constexpr(std::is_same_v<TArg, float>)
```
sbastrakov: This is dangerous. What if due to changed / weird template instantiation `TArg` becomes `volatile float` - your check will be false then. I think a proper way is to check that decayed type is float, probably via a helper trait.

sbastrakov: See example: https://godbolt.org/z/re61eWzr4

Edit: changed link for a more illustrative example

sbastrakov: Btw in case our unit tests don't catch such a case, perhaps they should be extended.

include/alpaka/math/rsqrt/RsqrtUniformCudaHipBuiltIn.hpp
sbastrakov: I think `sqrt` should allow integers, as so does the standard library version. For the other two I think makes sense to do same as `sqrt`.

include/alpaka/math/acos/AcosUniformCudaHipBuiltIn.hpp
j-stephan: I've added a decaying check. Please re-review :-)

include/alpaka/math/max/MaxStdLib.hpp
psychocoderHPC: IMO we should forbid `long double` until it became important for one of our users. IMO the current scope of alpaka is portability and this is easier if `long double` can not be used.

psychocoderHPC: INdependent of what you choose but please remove the `else` as in all other cases, that we are always explicit.

include/alpaka/math/min/MinStdLib.hpp
```diff
@@ -27,32 +27,21 @@ namespace alpaka
...
                 {
                     alpaka::ignore_unused(min_ctx);
+
                     using std::fmin;
-                    return fmin(x, y);
+                    using std::min;
+
+                    if constexpr(std::is_integral_v<Tx> && std::is_integral_v<Ty>)
+                        return min(x, y);
+                    else // Floating point or mixed integral-floating point
```
psychocoderHPC: This part should be explicit too and the else branch should contain a static assert.

psychocoderHPC (dismissed):
min/max requires the static assert in the els branch, the rest looks good!

include/alpaka/core/Decay.hpp
```diff
@@ -29,3 +29,10 @@
 #else
 #    define ALPAKA_DECAY_T(Type) std::decay_t<Type>
 #endif
+
+namespace alpaka
+{
+    //! Provides a decaying wrapper around std::is_same. Example: is_decayed_v<volatile float, float> returns true.
+    template<typename T, typename U>
+    inline constexpr auto is_decayed_v = std::is_same_v<ALPAKA_DECAY_T(T), U> || std::is_same_v<T, ALPAKA_DECAY_T(U)>;
```
j-stephan: It appears that clang 5 and 6 are not treating `inline` variables correctly here.

include/alpaka/core/Assert.hpp
```diff
@@ -71,32 +61,16 @@ namespace alpaka
...
-                TRhs,
-                std::enable_if_t<std::is_unsigned<TRhs>::value && (TLhs::value == 0u)>>
-            {
-                ALPAKA_NO_HOST_ACC_WARNING
-                ALPAKA_FN_HOST_ACC static auto assertGreaterThan(TRhs const& lhs) -> void
-                {
-                    alpaka::ignore_unused(lhs);
-                    // Nothing to do for unsigned types camparing to zero.
+                    if constexpr(std::is_signed_v<TRhs> || (TLhs::value != 0u))
+                        ALPAKA_ASSERT_OFFLOAD(TLhs::value > lhs);
```
sbastrakov: I think `lhs` should be called `rhs`?

j-stephan: You're right. Seems like an error in the original code that I didn't notice during refactoring.

sbastrakov: Yes, unrelated to your change, but made it unnecessarily difficult to understand this piece.

include/alpaka/core/Decay.hpp
```diff
@@ -29,3 +29,19 @@
...
+{
+    //! Provides a decaying wrapper around std::is_same. Example: is_decayed_v<volatile float, float> returns true.
+    /* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because these
+    /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+    template<typename T, typename U>
+    inline constexpr auto is_decayed_v = std::is_same_v<ALPAKA_DECAY_T(T), U> || std::is_same_v<T, ALPAKA_DECAY_T(U)>;
```
sbastrakov: Why not `std::is_same_v<ALPAKA_DECAY_T(T), ALPAKA_DECAY_T(U)>` ?

include/alpaka/core/Assert.hpp
```diff
@@ -71,41 +62,26 @@ namespace alpaka
...
-                TRhs,
-                std::enable_if_t<std::is_unsigned<TRhs>::value && (TLhs::value == 0u)>>
-            {
-                ALPAKA_NO_HOST_ACC_WARNING
-                ALPAKA_FN_HOST_ACC static auto assertGreaterThan(TRhs const& lhs) -> void
-                {
-                    alpaka::ignore_unused(lhs);
-                    // Nothing to do for unsigned types camparing to zero.
+                    alpaka::ignore_unused(rhs);
+                    if constexpr(std::is_signed_v<TRhs> || (TLhs::value != 0u))
```
j-stephan, line 71: I wonder if `TLhs` here should also be `TRhs`.

include/alpaka/core/Assert.hpp
j-stephan: Okay, should be a bit clearer now.

include/alpaka/math/cbrt/CbrtUniformCudaHipBuiltIn.hpp
```diff
@@ -30,22 +31,18 @@ namespace alpaka
...
-            struct Cbrt<CbrtUniformCudaHipBuiltIn, float>
-            {
-                __device__ auto operator()(CbrtUniformCudaHipBuiltIn const& cbrt_ctx, float const& arg) -> float
-                {
-                    alpaka::ignore_unused(cbrt_ctx);
-                    return ::cbrtf(arg);
+                    if constexpr(is_decayed_v<TArg, float>)
+                        return ::cbrtf(arg);
+                    else if constexpr(is_decayed_v<TArg, double> || std::is_integral_v<TArg>)
+                        return ::cbrt(arg);
```
sbastrakov, line 43: I am not sure, doesn't integer part cause a conversion warning? If so, should be separated with explicit conversion there

include/alpaka/math/fmod/FmodUniformCudaHipBuiltIn.hpp
```diff
@@ -34,23 +35,18 @@ namespace alpaka
...
-                {
-                    alpaka::ignore_unused(fmod_ctx);
-                    return ::fmodf(x, y);
+
+                    if constexpr(is_decayed_v<Tx, float> && is_decayed_v<Ty, float>)
+                        return ::fmodf(x, y);
+                    else if constexpr(is_decayed_v<Tx, double> || is_decayed_v<Ty, double>)
+                        return ::fmod(x, y);
+                    else
+                        static_assert(!sizeof(Tx), "Unsupported data type");
```
sbastrakov, line 49: I find this a bit illogical. If implementation only supports arguments of the same type, why have them as `Tx`, `Ty` and not make this partial template specialization only for matching types? We still have a general version for all rest. Same applies for other binary operation changes in this PR

sbastrakov (commented):
Some more suggestions

include/alpaka/math/fmod/FmodUniformCudaHipBuiltIn.hpp
sbastrakov: Oh, I now see it actually supports mixed precision

include/alpaka/math/cbrt/CbrtUniformCudaHipBuiltIn.hpp
j-stephan: So far the CI hasn't complained about this part, so I guess the conversion is fine.

include/alpaka/core/Assert.hpp
j-stephan: Nope, `TLhs` is fine here. The check is only used inside `Vec.hpp` and ensures that the user doesn't request a vector element outside of the vector dimensionality.

sbastrakov: I think the logic is to avoid comparison `0 > unsigned_type_value`. So the condition is the opposite of this case

include/alpaka/math/cbrt/CbrtUniformCudaHipBuiltIn.hpp
sbastrakov: I don't think we check it in unit tests. We should at least try it manually and see it compiles

include/alpaka/core/Assert.hpp
j-stephan: Well, the struct is called `AssertGreaterThan`. But since a `Vec<0>` class defines `Lhs = 0` the check is not applicable because it would fail for every possible vector index including `0`. Looks a bit like an oversight to me, we should probably ensure that users don't call `Vec<0>[1]`.

include/alpaka/math/cbrt/CbrtUniformCudaHipBuiltIn.hpp
j-stephan: Wow, you're right. We only check `float` and `double`. That sounds dangerous...

include/alpaka/core/Assert.hpp
sbastrakov: True, so probably was copy-pasted from somewhere else and not properly adapted.

include/alpaka/math/cbrt/CbrtUniformCudaHipBuiltIn.hpp
j-stephan: I added the following check to one of our test cases:

```c++
auto c = alpaka::math::cbrt(acc, 49);
ALPAKA_CHECK(*success, c >= 7.0);
```

Ignore the check, I just wanted to make sure the compiler doesn't optimize away the unused `c`. It doesn't complain about a conversion, though.

sbastrakov: I think the previous CUDA/HIP specializations did not support floating-point and integral as a mix, since they were partially specializd either specifically for `float` and `double` or floating-point types.

Thanks for checking

sbastrakov (approved):

fwyzard:
With GCC 10.3 and CUDA 11.2, configured with
```bash
cmake /home/fwyzard/src/alpaka-group/alpaka \
  -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON \
  -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE=OFF \
  -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF \
  -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE=OFF \
  -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=OFF \
  -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=OFF \
  -DALPAKA_ACC_ANY_BT_OMP5_ENABLE=OFF \
  -DALPAKA_ACC_ANY_BT_OACC_ENABLE=OFF \
  -DALPAKA_ACC_GPU_CUDA_ENABLE=ON \
  -DALPAKA_ACC_GPU_CUDA_ONLY_MODE=OFF \
  -DBUILD_TESTING=ON \
  -Dalpaka_BUILD_EXAMPLES=ON \
  -DCMAKE_CXX_COMPILER=g++-10 \
  -DCMAKE_CXX_FLAGS='' \
  -DCMAKE_EXE_LINKER_FLAGS='' \
  -DOpenMP_CXX_VERSION=5 \
  -DCMAKE_CUDA_COMPILER=/usr/local/cuda-11.2/bin/nvcc \
  -DCMAKE_CUDA_ARCHITECTURES=86 \
  -DCMAKE_VERBOSE_MAKEFILE=ON
```
this is generating a lot of warnings like
```
.../alpaka-group/alpaka/include/alpaka/vec/Vec.hpp(471): warning #940-D: missing return statement at end of non-void function "alpaka::traits::SubVecFromIndices<alpaka::Vec<TDim, TVal>, std::integer_sequence<std::size_t, TIndices...>, void>::subVecFromIndices [with TDim=alpaka::DimInt<3UL>, TVal=uint32_t, TIndices=<0UL, 1UL, 2UL>]"
```

j-stephan:
That is weird. Our CUDA + g++-10 CI is not issuing this warning. And there is indeed a return statement at the end of the function:

https://github.com/alpaka-group/alpaka/blob/826c71a9bdef8aa0f7f5aabe0392f3aeaf2e0f7d/include/alpaka/vec/Vec.hpp#L451-L472

We have faced issues with older CUDA 11.x and GCC 10 in the past (see #1297 for example) which is why we don't officially support anything below 11.5 with GCC 10 (see support matrix). Looks like this is another case of incompatibilities between those two.

fwyzard:
The function has
```c++
if constexpr ...
  return ...
else
  return ...
```

Looks like NVCC does not realise that both branches have a `return` statement, and no other option is possible 🤷

j-stephan:
It turns out that this is a `nvcc` bug that was only fixed in CUDA 11.5. I'll submit a workaround as part of #1563.

PR #1559, opened 2022-01-05
[RFC] Support the CUDA/HIP host API without a CUDA/HIP compiler (option 1)

fwyzard:
Require a CUDA or HIP compiler only for compiling device code or kernel launches.

This is a proof-of-concept implementation of the first option discussed in the proposal #1560.

While I've tried to minimize (hopefully down to zero) the changes visible to the code that includes `alpaka/alpaka.hpp`, I would like to gather some feedback regarding some of the details.

###  Header structure

This PR moves all headers with `__device__` code to a `.../device/` subdirectory, but other approaches are possible, for example:
  - keep the device code in the existing header, and create a new host-only header with a different name or directory
  - keep the device code in the same directory, with a different header name
  - other ideas ?

By the way, `include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp` was **not** moved to a `device` subdirectory, despite it being compilable only with a device compiler.
Should I move it to `include/alpaka/kernel/device/TaskKernelGpuUniformCudaHipRt.hpp` ?

Also, suggestions for a better name for `alpakaHostAPI.hpp` are welcome :-)

### New header for the CUDA/HIP compiler check

I've moved all the CUDA/HIP compiler checks of the form
```c++
#    include <alpaka/core/BoostPredef.hpp>

#    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && !BOOST_LANG_CUDA
#        error If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!
#    endif

#    if defined(ALPAKA_ACC_GPU_HIP_ENABLED) && !BOOST_LANG_HIP
#        error If ALPAKA_ACC_GPU_HIP_ENABLED is set, the compiler has to support HIP!
#    endif
```
to a dedicated header, and replaced them everywhere with
```c++
#    include <alpaka/core/DeviceOnly.hpp>
```

Does this approach look reasonable ?
If people prefer to have the checks explicitly, I can move them back.

I also considered wrapping them in a macro instead - but one cannot wrap an `#error` directive in a macro.
GCC and clang could use `_Pragma ("GCC error \"If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!\"")` - but I don't know if other compilers provide a similar pragma.


j-stephan:
Since this is a RFC I have cancelled the CI run. Please notify me if you want this PR checked, I will then restart the CI.

fwyzard:
> Since this is a RFC I have cancelled the CI run. Please notify me if you want this PR checked, I will then restart the CI.

Please, do let the CI run: I've tested this locally on a couple combinations of compilers and accelerators, but not as extensive as what the CI does.

While the main purpose of the PR is to collect people's feedback, I'd like it to cover all the corner cases that the CI can test.

SimeonEhrig:
> Finally, I'm not sure how to add a unit test for this: as far as I understand, CMake will always use nvcc (or clang --cuda) to build all code if ALPAKA_ACC_GPU_CUDA_ENABLED is defined.

Only, if you use `alpaka_add_executable`. If you use `add_executable`, it uses the configured C++ compiler. For example, in application which I port from CUDA to alpaka, most of the targets are compiled with the gcc/clang and only the targets, which contains cuda/alpaka code uses the nvcc.

Nevertheless, we need also think about, how we support the feature with CMake, because it is the recommend build system for alpaka.

By the way. I see also a benefit for the hip compiler ins this PR. The hip compiler is incredible slow :-(

fwyzard:
(rebased)

fwyzard:
I think we can move the overall discussion to #1560, and keep here only the part relevant to this approach and implementation.

fwyzard:
Looks like option 2 (#1561) or option 4 (#1565) are better than this one.

Issue #1560, opened 2022-01-06
[RFC] Support the CUDA/HIP host API without a CUDA/HIP compiler


fwyzard:
This RFC proposes to require a CUDA or HIP compiler only for compiling device code or kernel launches.
The rest of the CUDA or HIP host API (device queries, memory operations, etc.) would be usable with a standard compiler, as long the required libraries are available.

This proposal aims make available in Alpaka the division of host and device compilers used in CMS, where we try to minimise the code that needs to be compiled by `nvcc`. The main reasons are
  - `nvcc` tends to lag behind `gcc` and `clang` in the adoption of new C++ features
  - `nvcc` had problems handling the CMSSW framework and data formats (including the ROOT serialisation/deserialization part)

Using native CUDA, we can compile most of the CUDA-related API (device queries, memory operations, _etc._) with `g++`, and we need to use `nvcc` only for device code (obviously) and kernel launches (due to the `kernel<<<...>>>(...)` syntax and the use of templated `__global__` functions).

The same modularity can be achieved with Alpaka with minimal changes to the headers structure or the preprocessor symbols, depending on the "user interface" that is decided to expose.
Below few different approaches are described, along with a proof-of-concept implementation for some of them.

If and how to expose them host-only API to CMake is not discussed yet.

## Option 1: separate user-facing headers

The first approach changes the headers structure splitting them into host-only and host-and-device headers, both at the implementation level, into `some/feature.hpp` and `some/device/feature.hpp`, and at the user-facing level, into `alpaka.hpp` and `alpakaHostAPI.hpp`:
  - the usual user-facing header, `#include <alpaka/alpaka.hpp>`, includes all the alpaka headers as before, and requires a CUDA/HIP compiler if `ALPAKA_ACC_GPU_CUDA_ENABLED` is set, _etc._
  - a new user-facing header, `#include <alpaka/alpakaHostAPI.hpp>`, includes only the headers that can be compiled without a CUDA/HIP compiler; this does not allow writing or enqueuing kernels and device code, but allows device queries, memory operations, _etc._

Code that includes `alpaka/alpaka.hpp` should not observe any changes in behaviour; new host-only code can be written using the new `alpaka/alpakaHostAPI.hpp` header.

A proof-of-concept implementation is available in #1559.

## Option 2: autodetect a device compiler

The second approach keep the existing header structure. Instead, the code that requires a device compiler is guarded by preprocessor macros:
```c++
[...]

PR #1561, opened 2022-01-06
[RFC] Support the CUDA/HIP host API without a CUDA/HIP compiler (option 2)

fwyzard:
Require a CUDA or HIP compiler only for compiling device code or kernel launches.

This is a proof-of-concept implementation of the second option discussed in the proposal #1560, including a new integration test for the host-only API.


fwyzard:
The [only error](https://github.com/alpaka-group/alpaka/runs/4761923848?check_suite_focus=true) (so far) seems unrelated to the PR itself:
```
2022-01-10T14:54:27.7903639Z Chocolatey v0.11.3
2022-01-10T14:54:27.9458335Z Installing the following packages:
2022-01-10T14:54:27.9466179Z cmake.install
2022-01-10T14:54:27.9471473Z By installing, you accept licenses for the packages.
2022-01-10T14:55:26.6245309Z cmake.install not installed. The package was not found with the source(s) listed.
2022-01-10T14:55:26.6246562Z  Source(s): 'https://community.chocolatey.org/api/v2/'
```

fwyzard:
We have decided that the preferred approach is option 4, _i.e._ #1565.

PR #1563, opened 2022-01-07
Remove `alpaka::ignore_unused`

### LINKED ISSUE #1555: `ignore_unused()` should be replaced with `[[maybe_unused]]` or removed entirely
With the switch to C++17 we have the `[[maybe_unused]]` attribute available. This should replace `alpaka::ignore_unused()` in all instances where a variable / parameter is only used in some code branches. Example:

```c++
 __device__ static auto atomicOp(
    AtomicUniformCudaHipBuiltIn const&,
    unsigned long int* const addr,
    unsigned long int const& value) -> unsigned long int
{
#   if UINT_MAX == ULONG_MAX // LLP64
    return ::atomicSub(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
#   else // ULONG_MAX == ULLONG_MAX LP64
    alpaka::ignore_unused(addr);
    alpaka::ignore_unused(value);
    static_assert(
        meta::DependentFalseType<THierarchy>::value,
        "atomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported when "
        "sizeof(unsigned long int) == 4");
#    endif
}
```

can be transformed into

```c++
 __device__ static auto atomicOp(
    AtomicUniformCudaHipBuiltIn const&,
    [[maybe_unused]] unsigned long int* const addr,
    [[maybe_unused]] unsigned long int const& value) -> unsigned long int
{
#   if UINT_MAX == ULONG_MAX // LLP64
    return ::atomicSub(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
#   else // ULONG_MAX == ULLONG_MAX LP64
    static_assert(
        meta::DependentFalseType<THierarchy>::value,
        "atomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported when "
        "sizeof(unsigned long int) == 4");
#    endif
}
```

(Those macro checks should probably also be a `if constexpr` now.)

In other cases `ignore_unused` can be removed entirely because it serves no purpose. Example:

```c++
__device__ auto operator()(FmodUniformCudaHipBuiltIn const& fmod_ctx, float const& x, float const& y) -> float
{
    alpaka::ignore_unused(fmod_ctx);
    return ::fmodf(x, y);
}
```

This can be transformed into:

```c++
__device__ auto operator()(FmodUniformCudaHipBuiltIn const&, float const& x, float const& y) -> float
{
    return ::fmodf(x, y);
}
```

j-stephan:
This PR removes all instances of `alpaka::ignore_unused` according to the following rules:

* User-facing API: `int foo(float bar)` is transformed into `int foo([[maybe_unused]] float bar)`
* Internal API: `int foo(float bar)` is transformed into
  * `int foo(float /* bar */)` if `bar` is never used
  * `int foo([[maybe_unused]] float bar)` if `bar` is used in compile-time code branches (`if constexpr`, `#ifdef` etc.)
* Discarded return values: `ignore_unused(foo(42.f))` is transformed into `std::ignore = foo(42.f);` according to the C++ core guidelines
* Conditionally unused variables are prefixed with `[[maybe_unused]]`

### Additional changes

* Deleted `mem/view/ViewCompileTimeArray.hpp`. Everything inside was commented out and nobody seemed to miss this functionality
* Removed commented out test case in `unit/kernel/src/KernelWithAdditionalParam.cpp`
* Refactored some `#if` / `#else` preprocessor branches into `if constexpr` branches. This occured in places I needed to touch for removing `ignoring_unused`.


j-stephan:
@jkelling Could you have a look at the parts affecting the OpenACC back-end? I have no idea why the one failing GitLab test behaves as it does.

jkelling:
> @jkelling Could you have a look at the parts affecting the OpenACC back-end? I have no idea why the one failing GitLab test behaves as it does.

I do know this failure mode for `matMulTest` with OpenACC and NVHPC: When you compile with `-On` this type of memory collapse happens. Note, that for some reason 79 values are non-zero---1/64 of all entries---and their value is 64*23 (minus some loss due to data races). So groups of 64 blocks are writing to the same element, while they should be spread put. Clearly the index calculation goes wrong there if a `-On` flag is passed (note, that the issue is not there only with `-O` (with no `n`) or no such flag). This PR does not change flags and the CI output does not show any. Maybe removing unused parameter names somewhere made the optimizer do something that is does anyway at a different level (both higher *and lower* (`-O` is `-O1` + some other optimizations, but less than `-O2` according to `--help`)). I will look into this.

jkelling:
This patch https://github.com/j-stephan/alpaka/pull/1 works around the issue. It is a problem with NVHPC's loop optimization which I reported before but is apparently still present in 22.1 (current latest version). I have no idea though, which of you changes activated this behavior in the compiler, but it does not really matter.

On the plus side: with this workaround we can run `matMulTest` even at `-O2` (and maybe other opt levels), this does not hold for other tests, though.

j-stephan:
Great, thanks for investigating this! I've updated this PR with your workaround: https://github.com/alpaka-group/alpaka/pull/1563/files#diff-c15c991219ddfec26bfa494bff144f2e6c9c041efdae5e0219008bad59347e18

Hopefully the CI will pass now.

jkelling (commented):
I think you should have left my workaround in a separate commit because it does not so much fix your changes but works around an other unrelated issue...

include/alpaka/math/pow/PowUniformCudaHipBuiltIn.hpp
```diff
@@ -37,16 +37,21 @@ namespace alpaka
...
-
                     if constexpr(is_decayed_v<TBase, float> && is_decayed_v<TExp, float>)
                         return ::powf(base, exp);
                     else if constexpr(is_decayed_v<TBase, double> || is_decayed_v<TExp, double>)
                         return ::pow(base, exp);
                     else
                         static_assert(!sizeof(TBase), "Unsupported data type");
+
+                    using Ret [[maybe_unused]]
+                    = std::conditional_t<is_decayed_v<TBase, float> && is_decayed_v<TExp, float>, float, double>;
```
jkelling, line 53: Evidently this passed clang-format, but is this the indentation style we want for using declarations?

include/alpaka/math/remainder/RemainderUniformCudaHipBuiltIn.hpp
```diff
@@ -39,18 +39,20 @@ namespace alpaka
...
-
                     if constexpr(is_decayed_v<Tx, float> && is_decayed_v<Ty, float>)
                         return ::remainderf(x, y);
                     else if constexpr(is_decayed_v<Tx, double> || is_decayed_v<Ty, double>)
                         return ::remainder(x, y);
                     else
                         static_assert(!sizeof(Tx), "Unsupported data type");
+
+                    using Ret [[maybe_unused]]
+                    = std::conditional_t<is_decayed_v<Tx, float> && is_decayed_v<Ty, float>, float, double>;
```
jkelling, line 54: Same here: weird indentation.

jkelling (dismissed):

include/alpaka/core/Unreachable.hpp
```diff
@@ -0,0 +1,30 @@
...
+#    else
+#        define ALPAKA_UNREACHABLE(...) return __VA_ARGS__
+#    endif
+#elif BOOST_COMP_MSVC
+#    define ALPAKA_UNREACHABLE(...) __assume(false)
+#elif BOOST_COMP_GNUC || BOOST_COMP_CLANG || BOOST_COMP_INTEL || BOOST_COMP_PGI
+#    define ALPAKA_UNREACHABLE(...) __builtin_unreachable()
+#else
+#    define ALPAKA_UNREACHABLE(...)
+#endif
```
bernhardmgruber: I don't like the name for the macro, because the one case where it turns into a return statement is surprising. Maybe rename it into `ALPAKA_UNREACHABLE_RETURN`. This way, we can still define a portable wrapper for just unreachable.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -423,7 +422,7 @@ namespace alpaka
                 // Wait for the kernel execution to finish but do not check error return of this call.
                 // Do not use the alpaka::wait method because it checks the error itself but we want to give a custom
                 // error message.
-                ignore_unused(ALPAKA_API_PREFIX(StreamSynchronize)(queue.m_spQueueImpl->m_UniformCudaHipQueue));
+                std::ignore = ALPAKA_API_PREFIX(StreamSynchronize)(queue.m_spQueueImpl->m_UniformCudaHipQueue);
```
bernhardmgruber, line 425: When I read this, I hear Marshall Clow's voice telling me to not do it. But it is indeed a nifty trick.

include/alpaka/math/atan/AtanUniformCudaHipBuiltIn.hpp
```diff
@@ -33,16 +33,16 @@ namespace alpaka
             template<typename TArg>
             struct Atan<AtanUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_floating_point_v<TArg>>>
             {
-                __device__ auto operator()(AtanUniformCudaHipBuiltIn const& atan_ctx, TArg const& arg)
+                __device__ auto operator()(AtanUniformCudaHipBuiltIn const& /* atan_ctx */, TArg const& arg)
```
bernhardmgruber, line 36: Why is the name of the parameter even important here? Why not just drop it?

```suggestion
                __device__ auto operator()(AtanUniformCudaHipBuiltIn const&, TArg const& arg)
```

include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
```diff
@@ -52,6 +50,12 @@ namespace alpaka
                         return ::fmax(x, y);
                     else
                         static_assert(!sizeof(Tx), "Unsupported data type");
+
+                    using Ret [[maybe_unused]] = std::conditional_t<
+                        std::is_integral_v<Tx> && std::is_integral_v<Ty>,
+                        decltype(::max(x, y)),
+                        std::conditional_t<is_decayed_v<Tx, float> && is_decayed_v<Ty, float>, float, double>>;
+                    ALPAKA_UNREACHABLE(Ret{});
```
bernhardmgruber, line 58: That seems like a lot of magic to just supress a warning. Is there no pragma we can wrap around the entire alpaka headers?

include/alpaka/math/pow/PowUniformCudaHipBuiltIn.hpp
bernhardmgruber: I think this is fixed in newer clang-format versions (quick&dirty try in godbolt.org).

include/alpaka/math/max/MaxUniformCudaHipBuiltIn.hpp
jkelling: Warning pragmas are compier specific, no? So this is more portable.

include/alpaka/core/Unreachable.hpp
```diff
@@ -0,0 +1,30 @@
...
+#if(BOOST_COMP_NVCC && BOOST_ARCH_PTX)
+#    if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 3, 0)
+#        define ALPAKA_UNREACHABLE(...) __builtin_unreachable()
+#    else
+#        define ALPAKA_UNREACHABLE(...) return __VA_ARGS__
+#    endif
+#elif BOOST_COMP_MSVC
+#    define ALPAKA_UNREACHABLE(...) __assume(false)
+#elif BOOST_COMP_GNUC || BOOST_COMP_CLANG || BOOST_COMP_INTEL || BOOST_COMP_PGI
+#    define ALPAKA_UNREACHABLE(...) __builtin_unreachable()
```
jkelling: This does not exactly seem to work for NVHPC/PGI, [e.g.](https://github.com/alpaka-group/alpaka/runs/4790090659?check_suite_focus=true#step:5:3141).

include/alpaka/core/Unreachable.hpp
j-stephan: The `return` doesn't have any effect, it is simply a dummy which we unfortunately have to use for older CUDA variants. The purpose of the macro is to mark a code region as unreachable.

include/alpaka/math/atan/AtanUniformCudaHipBuiltIn.hpp
j-stephan: This has been requested by @psychocoderhpc for documentation purposes.

include/alpaka/core/Unreachable.hpp
j-stephan: So NVHPC warns about `__builtin_unreachable()` being unreachable... *sigh*. I give up, I guess we have to live with that warning until we can remove CUDA < 11.5 or NVHPC is fixed.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
j-stephan: This is basically coming  straight from Mount Sinai: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-casts

> Never cast to `(void)` to ignore a `[[nodiscard]]` return value. If you deliberately want to discard such a result, first think hard about whether that is really a good idea (there is usually a good reason the author of the function or of the return type used `[[nodiscard]]` in the first place). If you still think it’s appropriate and your code reviewer agrees, use `std::ignore =` to turn off the warning which is simple, portable, and easy to grep.

include/alpaka/core/Unreachable.hpp
jkelling: If the current version removes the warning for CUDA < 11, then keep it. With NVHPC we do not have to support old versions, i.e. it just moves up with time dropping the old ones.

j-stephan: Sorry, there was a typo in my comment (I edited it to say "< CUDA 11.5"). But I'm still hopeful that NVHPC will be fixed sooner than us dropping everything except the current latest CUDA version.

jkelling: It looks like `__builtin_unreachable()` is just not supported by PGI/NVHPC and may never be as it is a GNU buiilt-in.

So, remove PGI here:
```suggestion
#elif BOOST_COMP_GNUC || BOOST_COMP_CLANG || BOOST_COMP_INTEL
#    define ALPAKA_UNREACHABLE(...) __builtin_unreachable()
```
and instead add `--diag_suppress 111 ` [here](https://github.com/alpaka-group/alpaka/blob/develop/cmake/alpakaCommon.cmake#L146) to disable this warning.

jkelling:
I guess NVHPC has some [sense of humor](https://github.com/alpaka-group/alpaka/runs/4790090659?check_suite_focus=true#step:5:3260):
```
"/home/runner/work/alpaka/alpaka/include/alpaka/idx/MapIdx.hpp", line 65: warning: pointless comparison of unsigned integer with zero
                  for(std::size_t a(0u); a < lastIdx - 1; ++a)
                                           ^
```
... it is warning about the workaround for its loop bug ...

include/alpaka/idx/MapIdx.hpp
```diff
@@ -64,8 +61,15 @@ namespace alpaka
 
                 // in-between
                 TElem hyperPlanesBefore = extent[lastIdx];
+#if BOOST_COMP_PGI && (defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) || defined(ALPAKA_ACC_ANY_BT_OACC_ENABLED))
```
jkelling:
```suggestion
#if BOOST_COMP_PGI && (defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) || defined(ALPAKA_ACC_ANY_BT_OACC_ENABLED)) && !defined(TPR30645)
```

This adds the internal bug number. Once the bug is fixed this macro will be defined and the workaround disabled.

include/alpaka/math/pow/PowUniformCudaHipBuiltIn.hpp
j-stephan: Attributes cause an amazing amount of troubles across the C++ ecosystem. My syntax highlighting in the editor also breaks here...

sbastrakov (approved):

Issue #1552, opened 2021-12-23
OpenMP 5 backend fails in integration tests with Alpaka 0.8.0


fwyzard:
Hi,
I'm trying to run the integration tests with Alpaka 0.8.0, and I'm getting some segmentation fault errors.

I'm working on Ubuntu 21.10 with GCC 10.3.0, and I'm setting up the test with
```bash
git clone -b 0.8.0 git@github.com:alpaka-group/alpaka.git

mkdir build
cd build

cmake ../alpaka \
  -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON \
  -DALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE=ON \
  -DALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE=OFF \
  -DALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE=ON \
  -DALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE=ON  \
  -DALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE=ON  \
  -DALPAKA_ACC_ANY_BT_OMP5_ENABLE=ON \
  -DALPAKA_ACC_ANY_BT_OACC_ENABLE=OFF \
  -DBUILD_TESTING=ON \
  -DCMAKE_CXX_COMPILER=g++-10 \
  -DCMAKE_CXX_FLAGS=-foffload=disable \
  -DCMAKE_VERBOSE_MAKEFILE=ON

make -j4
```

When I run the tests with
```bash
make test
```
I get
```
$ make test
Running tests...
/usr/bin/ctest --force-new-ctest-process 
Test project /home/fwyzard/src/alpaka-group/x/build
      Start  1: headerCheckTest
 1/28 Test  #1: headerCheckTest ..................   Passed    0.00 sec
      Start  2: axpyTest
 2/28 Test  #2: axpyTest .........................***Exception: SegFault  1.85 sec
      Start  3: mandelbrotTest
 3/28 Test  #3: mandelbrotTest ...................***Exception: SegFault 32.98 sec
      Start  4: matMulTest
 4/28 Test  #4: matMulTest .......................***Exception: SegFault  0.30 sec
      Start  5: sharedMemTest
 5/28 Test  #5: sharedMemTest ....................***Exception: SegFault  2.61 sec
      Start  6: accTest
[...]

PR #1564, opened 2022-01-08
Cleanup For OpenACC (and some OpenMP 5 target)

### LINKED ISSUE #1550: Ci: openAcc test show illegal memory access
As mentioned https://github.com/alpaka-group/alpaka/pull/1549#issuecomment-998748452 the openAcc runtime test is showing illegal memory access, the reason is not clear:

[linux_nvhpc-21.9_cuda-11.4_oacc](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1907538789) fails on gitlab CI with call to cuStreamSynchronize returned error 700: Illegal address during kernel execution. I suppose this is still the same error as in the original report
```
      Start  9: randomCells2D
9: Test command: /builds/hzdr/crp/alpaka/build/example/randomCells2D/randomCells2D
9: Test timeout computed to be: 1500
9: Failing in Thread:1
9: call to cuStreamSynchronize returned error 700: Illegal address during kernel execution
9: 
 9/37 Test  #9: randomCells2D ....................***Failed    1.14 sec
test 10
```

CC-ing: @jkelling could you have a look into the issue next year, please.

jkelling:
This PR contains a collection of smaller fixes to the OpenACC backend:

* Fix a bug when selection a device other than device 0: https://github.com/alpaka-group/alpaka/commit/c01eb0b5b51946e85298b7fedbd231bf4312f5f9
* Make `getDevById()` return the same device handle if called multiple times with the same index. This must be done because the device handle allocates some memory to implement a mutex. https://github.com/alpaka-group/alpaka/commit/2189d98291ead6e64c1eaa22e52f3e5d7dc2d866
* Add support to peer-to-peer copy (there was only assert prohibiting copies between device before) https://github.com/alpaka-group/alpaka/commit/dd8ec93c75d2869bd7592fd4f1c86992fc8ffe04
* Move CI to gcc 11 (also for OMP5) and NVHPC to 21.11 (does not actually change CI outcome)

Some other Fixes:
* Make the OpenACC backend allow more threads in CI ( 74f9bc69ae1652a323c67f562fba7f8b508576d8 ), which also works around https://github.com/alpaka-group/alpaka/issues/1550 .
* Also run tests with GCC for OMP5 (not only analysis, https://github.com/alpaka-group/alpaka/commit/74f9bc69ae1652a323c67f562fba7f8b508576d8) and add some documentation about how running on host with GCC works (https://github.com/alpaka-group/alpaka/commit/91914d87517b52a69e56e925a23f27572f09cf45) , courtesy of this discussion: https://github.com/alpaka-group/alpaka/issues/1552 .

Please merge before https://github.com/alpaka-group/alpaka/pull/1549 and rebase.

Closes #1550


test/unit/kernel/src/KernelWithOmpSchedule.cpp
```diff
@@ -122,6 +122,10 @@ void testKernel()
     REQUIRE(fixture(kernelWithTrait));
 }
 
+// Disaling these tests for GCC + OMP5 & OACC because GCC does not like static
```
j-stephan:
```suggestion
// Disabling these tests for GCC + OMP5 & OACC because GCC does not like static
```

include/alpaka/dev/DevOacc.hpp
```diff
@@ -306,6 +326,36 @@ namespace alpaka
...
+        //! The OpenACC platform device get trait specialization.
+        template<>
+        struct GetDevByIdx<PltfOacc>
+        {
+            //! \param devIdx device id, less than GetDevCount
+            ALPAKA_FN_HOST static auto getDevByIdx(std::size_t devIdx) -> DevOacc
+            {
+                ALPAKA_DEBUG_FULL_LOG_SCOPE;
+
+                std::size_t const devCount(getDevCount<PltfOacc>());
```
j-stephan, line 348: Why not make this `static`? Can the device count change during runtime?

script/gitlabci/job_cuda11.5.yml
```diff
@@ -100,3 +100,15 @@ linux_nvcc-11.5_clang-12_release:
...
   stage: stageRun1
+
+# nvhpc
+linux_nvhpc-21.11_cuda-11.5_oacc:
+  extends: .base_oacc_nvhpc
+  variables:
+    ALPAKA_CI_UBUNTU_VER: "20.04"
+    ALPAKA_CI_CUDA_VERSION: "11.5"
+    ALPAKA_CI_NVHPC_VER: "21.11"
+    CMAKE_BUILD_TYPE: ""
```
j-stephan, line 111: Is there a reason we're not setting the `CMAKE_BUILD_TYPE`?

j-stephan (changes requested):
Thanks! A few minor things:

include/alpaka/dev/DevOacc.hpp
jkelling: 1. [Other](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/pltf/PltfUniformCudaHipRt.hpp#L84) backends do not do this either. Although it sounds like a good idea.
2. As far as alpaka is concerned it should not change. OpenACC has a concept of a device type though. The [current implementation](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/pltf/PltfOacc.hpp#L51) just returns the currently set device type, as Alpaka does not have any concept for this. In theory, if one would set a different device type between calls to this function, the returned count could change.

script/gitlabci/job_cuda11.5.yml
jkelling: We are actually explicitly setting it to no build type. This is because all cmake build types add some `-O` and/or `-g` flags. Passing `-g` or any `-O` (other than `-O1`, I think) to nvc++ breaks the generated code for some tests.

Note, that this backend is highly not production ready. The compiler flags set in the Ci (should) work. If you change any, your have a high change of running into some compiler bug.

include/alpaka/dev/DevOacc.hpp
jkelling: 1. [Other](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/pltf/PltfUniformCudaHipRt.hpp#L84) backends do not do this either. Although it sounds like a good idea.
2. As far as alpaka is concerned it should not change. OpenACC has a concept of a device type though. The [current implementation](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/pltf/PltfOacc.hpp#L51) just returns the currently set device type, as Alpaka does not have any concept for this. In theory, if one would set a different device type between calls to this function, the returned count could change.

script/gitlabci/job_cuda11.5.yml
jkelling: We are actually explicitly setting it to no build type. This is because all cmake build types add some `-O` and/or `-g` flags. Passing `-g` or any `-O` (other than `-O1`, I think) to nvc++ breaks the generated code for some tests.

Note, that this backend is highly not production ready. The compiler flags set in the Ci (should) work. If you change any, your have a high change of running into some compiler bug.

jkelling:
CI: Windows jobs fail with:
```
Failures
 - cmake.install - cmake.install not installed. The package was not found with the source(s) listed.
 ```
Looks like a CI hickup. I did not touch anything these jobs should care about, apart from `KernelWithOmpSchedule.cpp`. @bernhardmgruber  any thoughts?

**Comment by jkelling** [comment]
Due to renaming/deleting/adding of CI jobs the list of required jobs needs to be changed.

File: include/alpaka/dev/DevOacc.hpp (hunk 2 above)
Comment (jkelling) [reply include/alpaka/dev/DevOacc.hpp:348]: Btw.: USB-C GPU cases are a thing. These should be hot-plugable. What happens to, e.g., `cudaGetDeviceCount()` if you plug one in or remove one while the cuda runtime is initialized?

**Comment by jkelling** [comment]
I need to rebase and squash two commits when this is approved, before merge.

**Comment by bernhardmgruber** [comment]
> CI: Windows jobs fail with:
> ...
> Looks like a CI hickup.

CMake in the Windows CI is downloaded via chocolatey, which timed out. I restarted the jobs and they passed now.

File: .github/workflows/ci.yml [hunk 4]
```diff
@@ -94,9 +94,6 @@ jobs:
       matrix:
         include:
         ### Analysis builds
-        - name: linux_gcc-10_debug_analysis_omp5
-          os: ubuntu-latest
-          env: {CXX: g++,     CC: gcc,    ALPAKA_CI_GCC_VER: 10,       ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_CI_BOOST_BRANCH: boost-1.74.0, ALPAKA_CI_CMAKE_VER: 3.18.6, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:20.04", ALPAKA_CI_ANALYSIS: ON, ALPAKA_DEBUG: 2, ALPAKA_ACC_ANY_BT_OMP5_ENABLE: ON, CMAKE_CXX_FLAGS: "-foffload=disable"}
```
bernhardmgruber, line 99: Why did you remove the analysis build?

j-stephan (dismissed):

include/alpaka/dev/DevOacc.hpp
```diff
@@ -138,15 +135,38 @@ namespace alpaka
...
+        {
+            static std::vector<std::shared_ptr<oacc::detail::DevOaccImpl>> devices;
+            static std::mutex mutex;
+
+            if(!devices.size())
+            {
+                std::lock_guard<std::mutex> lock(mutex);
+                if(!devices.size())
+                    devices.resize(getDevCount<PltfOacc>());
+            }
```
bernhardmgruber: I am not sure if you can use double-checked locking if the locking condition does a non-atomic read. But either way, you can avoid it alltogether by just providing an initializer for the static variable:

```suggestion
            static auto devices = []{
                std::vector<std::shared_ptr<oacc::detail::DevOaccImpl>> d;
                d.resize(getDevCount<PltfOacc>());
                return d;
            }();
            static std::mutex mutex;
```

j-stephan:
I have changed the required checks. Please notify me when you're ready to merge.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -335,29 +335,39 @@ namespace alpaka
...
-                        // , but we want the tests to compile anyway
-                        [](void* dst, void* src, std::size_t size)
-                        {
-                            void* buf = core::alignedAlloc(core::vectorization::defaultAlignment, size);
-                            acc_memcpy_from_device(buf, src, size);
-                            acc_memcpy_to_device(dst, buf, size);
-                            core::alignedFree(buf);
-                        }
+                // acc_memcpy_device is only available since OpenACC2.5
+                // , but we want the tests to compile anyway
```
bernhardmgruber: Weird line break. I would put the comma into the preceding sentence, or just have the comment in one line and let clang-format break it.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -335,29 +335,39 @@ namespace alpaka
...
+                            [devSrc = getDev(viewSrc),
+                             devDst = getDev(viewDst)](void* dst, void* src, std::size_t size)
+                            {
+                                void* buf = core::alignedAlloc(core::vectorization::defaultAlignment, size);
+                                devSrc.makeCurrent();
+                                acc_memcpy_from_device(buf, src, size);
+                                devDst.makeCurrent();
+                                acc_memcpy_to_device(dst, buf, size);
+                                core::alignedFree(buf);
+                            });
```
bernhardmgruber: Can any of these operations throw an exception? If so, we could leak `buf`. `buf` should be managed by a scope guard. Maybe wrap it in a `std::unique_ptr` with a custom deleter.

bernhardmgruber (changes requested):

.github/workflows/ci.yml
jkelling: It only was an analysis build because running the tests did not work. It was rather changed to be a full test.

include/alpaka/dev/DevOacc.hpp
jkelling: > I am not sure if you can use double-checked locking if the locking condition does a non-atomic read.

The non-atomicity of the check+lock is the reason for the second check. So, the first check does not need to be atomic. The second check require cache coherence though to work. Is this what you are considering?

> But either way, you can avoid it alltogether by just providing an initializer for the static variable:

I read, that some compilers do not properly implement the thread-safe initialization of static variables, so I decided to keep the initializer as simple as possible. Anyway, if we go with a complex initializer, why not this:
```suggestion
            static std::vector<std::shared_ptr<oacc::detail::DevOaccImpl>> devices(getDevCount<PltfOacc>());
            static std::mutex mutex;
```
?

include/alpaka/mem/buf/oacc/Copy.hpp
jkelling: Clang-format does not break it. (The comment is not new, only its indentation changed in this patch.)

include/alpaka/mem/buf/oacc/Copy.hpp
jkelling: They cannot. The OpenACC API is C, it will (should) call `abort()` on error. Wrapping it looks cleaner though, I also do not like the `free` at the end.

.github/workflows/ci.yml
bernhardmgruber: The analysis build runs checks (like the header-check) that are not done in non-analysis builds. So by removing this one, you loose those checks on the OpenACC related headers. I would rather keep the test, but I have no strong opinion.

include/alpaka/dev/DevOacc.hpp
bernhardmgruber: > The non-atomicity of the check+lock is the reason for the second check. So, the first check does not need to be atomic. The second check require cache coherence though to work. Is this what you are considering?

I know the basics of double-checked locking, but I think you are still missing a memory barrier: https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/. Also you cannot read/write to a non-atomic variable outside locks, and the first read of `device.size()` is definitely outside a lock.

bernhardmgruber: > I read, that some compilers do not properly implement the thread-safe initialization of static variables, so I decided to keep the initializer as simple as possible.

Then the compiler is not C++11 conforming. This was the case with old VS versions and those are long gone. I would rely on this behavior.

`static std::vector<std::shared_ptr<oacc::detail::DevOaccImpl>> devices(getDevCount<PltfOacc>());` is btw an even better solution!

.github/workflows/ci.yml
jkelling: This is new in a4ab9eee9e5 (last month), I was not aware of this change. Originally, the header check was also performed in all the other tests and the analysis build did not run the tests.

My intent, and the original intent of the existing `_omp5` and `_oacc` tests is to doe the header check and run the tests. Only the GCC 10 test originally was supposed to only compiler but not run, because running did not work.

Thanks you for pointing this out. I will set `ALPAKA_CHECK_HEADERS=ON` then for all tests.

bernhardmgruber: The header check was there for a long time, we just restricted it to the analysis builds to speedup the CI. Furthermore, the header check is no longer enabled by default, so when you build locally, it is not run.

jkelling: > The header check was there for a long time

Indeed. It just did not notice, it was not build by-default anymore. I have to adjust to this.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -335,29 +335,40 @@ namespace alpaka
...
+                            viewSrc,
+                            extent,
+                            getDev(viewDst),
+                            [devSrc = getDev(viewSrc),
+                             devDst = getDev(viewDst)](void* dst, void* src, std::size_t size)
+                            {
+                                static const auto alignedDeleter = [](void* p) { core::alignedFree(p); };
+                                std::unique_ptr<void, decltype(alignedDeleter)> buf(
+                                    core::alignedAlloc(core::vectorization::defaultAlignment, size),
+                                    alignedDeleter);
```
bernhardmgruber: Can you not just pass `core::alignedFree` directly?

```suggestion
                                std::unique_ptr<void, decltype(&core::alignedFree)> buf(
                                    core::alignedAlloc(core::vectorization::defaultAlignment, size),
                                    &core::alignedFree);
```

include/alpaka/dev/DevOacc.hpp
jkelling: >  but I think you are still missing a memory barrier
> [...]
> Also you cannot read/write to a non-atomic variable outside locks, and the first read of `device.size()` is definitely outside a lock.

The initialization of `devices`, and therefore the variable behind `devices.size()`, happens inside a lock. The first read is not directly sequenced with the "write" in line 155. In the case a thread B reads old data after thread A has already written, then it will proceed to lock the mutex, checking again at the second read which is sequenced and it will get the correct data then. Regarding the initialization, the first read being a data race does not matter, because that is what the mutex is for.

There is another possible issue though, which I do not know if it can be solved without having every thread lock the mutex unconditionally on the first encounter: `devices.resize()` is as far from atomic as it can be. If it writes `devices.size()` before actually allocating the memory and storing the `devices.data()` pointer (and I think the compiler could just reorder these operations) then that thread B could run into trouble, if it reads the updated value of `devices.size()` and then proceeds to dereference a `nullptr`. I guess `std::call_once` takes care of this issue.

> Then the compiler is not C++11 conforming. This was the case with old VS versions and those are long gone. I would rely on this behavior.

I recall something about VS in this context. Good to know, that this is fixed.

include/alpaka/mem/buf/oacc/Copy.hpp
jkelling: Sure. I  messed up using `decltype` to get function *pointer* type m( .

.github/workflows/ci.yml
jkelling: I added the header check for the `_oacc` and `_omp5` jobs (eafe84f409). There is only one per (compiler,target), so having a separate analysis job for each would add more setup time than it would save in sequential build time.

include/alpaka/dev/DevOacc.hpp
```diff
@@ -138,15 +135,32 @@ namespace alpaka
...
+
+            auto& d = devices.at(static_cast<unsigned>(iDevice));
+            if(!d)
+            {
+                std::lock_guard<std::mutex> lock(mutex);
+                if(!d)
+                    return devices[static_cast<unsigned>(iDevice)]
+                        = std::make_shared<oacc::detail::DevOaccImpl>(iDevice);
+            }
+            return d;
```
bernhardmgruber: I am still concerned that this code has a race condition. You are writing to the memory location `d` points to under a mutex, but read unguarded from it (on a different thread inside this function or after the function returns).

Example:
Thread A checks `d`, which is `nullptr`, acquires the lock, checks again and starts creating the `shared_ptr`. The `shared_ptr` is already constructed and thus not `nullptr`, but the `acc::detail::DevOaccImpl` is not yet constructed.
Now thread B checks `d`. It is not `nullptr` and returns it. The caller than accesses an unconstructed `acc::detail::DevOaccImpl`.

Googling this problem I find: https://stackoverflow.com/a/12546724/2406044, which also uses atomic loads and stores.

.github/workflows/ci.yml
jkelling: Gitlab does not like what I did to the [`gitlab-ci.yml`](https://github.com/alpaka-group/alpaka/pull/1564/commits/eafe84f4099e29e04e5e1b9074947a65a725a560#diff-72d96b76899af358610624393ce5344832f5d04aa6bbd5481871029b91faedec) in eafe84f . I do not have access to to the ci-lint there. Can someone with access please tell me what is wrong? Are end-of-line comments not allowed? Or does the `ON` need to be quoted?

include/alpaka/dev/DevOacc.hpp
jkelling: I do not think, that an access to an uninitialized access to `acc::detail::DevOaccImpl` can happen the way your describe, because the `=` is a sequence point: The right-hand side, i.e. the construction of the `DevOaccImp` and that of the `std::shared_ptr` is executed first the assignment happens afterwards.

The issue you linked is about the assignment of `std::shared_ptr`not being atomic ... this may be an issue especially because it actually holds two pointers (payload, control block).

This whole thing could also be implemented using `std::unique_ptr`, `shared_ptr` is not really required. Normal `alpaka::Dev` objects are implicitly shared. This one is actually not: it is just a handle for some global instance. The ownership will always be with `devices` which could thus be a `std::vector<std::unique_ptr<oacc::detail::DevOaccImpl>>`. Its lifetime encompasses those of all `DecOacc` instances, thus no refcounting is required and `DevOacc` can just hold a plain pointer or a reference with no ownership.

jkelling: 2189d98 changes it to the `unique_ptr` implementation. I can still look into adding an atomic store for the assignment for platforms where pointers are written in multiple pieces if you think this is a potential issue.

include/alpaka/dev/DevOacc.hpp
```diff
@@ -138,66 +135,82 @@ namespace alpaka
...
         ALPAKA_FN_HOST auto registerQueue(std::shared_ptr<IGenericThreadsQueue<DevOacc>> spQueue) const -> void
         {
-            m_spDevOaccImpl->registerQueue(spQueue);
+            m_devOaccImpl.registerQueue(spQueue);
         }
 
-    public:
-        std::shared_ptr<oacc::detail::DevOaccImpl> m_spDevOaccImpl;
+    private:
+        oacc::detail::DevOaccImpl& m_devOaccImpl;
```
bernhardmgruber: Hmm. Now `DevOacc` is no longer copy and moveable. Is this also the case for other backends? It seems like an unnecessary restriction. Consider using a pointer instead of a reference here.

include/alpaka/dev/DevOacc.hpp
bernhardmgruber: First of, there are no sequence points any more since C++11. You are right that there is a happens before relationship between the construction of the `unique_ptr` and the assignment of the address into the `vector`. However, that sequencing concerns visible side effects within a single thread. And the optimizer can reorder the construction and assignment, because it does not matter within a single thread.

In a multithreaded scenario you also need synchronization operations such as mutex locks/unlocks or atomic loads/stores with acquire/release semantic. This is necessary because also the CPU can freely reorder the instruction stream. The unlock of your mutex is a release operation, so your changes should become visible. However, any other thread reading `d` misses an acquire operation., because release operations only synchronize with acquire operations and not ordinary reads. So theoretically, there is a data race.
See also this blog post: https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

I am tempted to believe this code can work in practice because the initial value of `d` is `nullptr`, visible to all threads and if a thread missreads an already initialized `d` as `nullptr` (e.g. from cache) it would then acquire the lock, which is an acquire operation synchronizing with a potential release operation from a different thread, and then have a valid read.

Anyway, I have argued enough and I'll let you proceed with what you think is best.

bernhardmgruber (dismissed):

include/alpaka/dev/DevOacc.hpp
jkelling: You mean copy- and move- assignable ... the respective constructors are still there.

It looks like assignment of `Dev` types is tested nowhere, but I guess there is not harm in keeping it.

bernhardmgruber: I opened #1569 for someone to have a closer look at this. You can leave your code for this PR as you wish.

jkelling: I will change it to pointer, because you are right, that the limitation is not necessary. I just always default to using a reference in situations like this because I like that "immutable type" pattern.

include/alpaka/dev/DevOacc.hpp
jkelling: > I am tempted to believe this code can work in practice because the initial value of d is nullptr, [...] synchronizing with a potential release operation from a different thread, and then have a valid read.

This is what I am arguing, yes.

> [...] However, any other thread reading d misses an acquire operation., because release operations only synchronize with acquire operations and not ordinary reads. So theoretically, there is a data race.

This is what we need to talk about, not the atomicity of the reads. Sorry I missed that. Using atomic load/store on the pointers is not so simple, however, as `std::unique_ptr` does not specialize atomic operations, so we would have to have a lock for each thread, every time. I just pushed c890ef61a8506a3ae63490836550cf848143fea5 which uses `std::atomic<T*>` which should do what you want, though in a (inevitably) long way.

> See also this blog post: https://preshing.com/20130930/double-checked-locking-is-fixed-in-cpp11/

This one suggests `std::atomic_thread_fence`, before giving the pure atomic solution as more elegant, which it is not in this case. So using `std::atomic_thread_fence` with `std::unique_ptr` and no atomics should be simplest solution. This would be fine, except if pointers are written piecemeal, which would require atomics to be solved. The post for some (maybe this) reason used the fence `and` atomics.

So.. what do you think? If you vote for  c890ef61a8506a3ae63490836550cf848143fea5 then that `ManageAtomicPtrVector` would need to got into a separate header....

jkelling: @bernhardmgruber I guess an alternative to this would be to have a `std::vector<std::pair<std::once_flag, std::unique_ptr<..>>>` and use `std::call_once` for the initialization to let the STL handle the locking. Instead of the pair one could also have a small type that handles the call to `call_once` and only returns references to ready objects.

jkelling: I replaced it with the simpler `std::call_once`-verison now. There is a nested but generic helper class. As I do not see any other use cases for this in Alpaka at the moment, I suggest to keep it nested for now and merge as is.

@bernhardmgruber can you please re-review now, it is ready to merge from my end.

jkelling: ... ready to be merged after rebase and squashing, that is..

bernhardmgruber (approved):

PR #1565, opened 2022-01-10
[RFC] Support the CUDA/HIP host API without a CUDA/HIP compiler (option 4)

fwyzard:
Introduce a new preprocessor symbol: `ALPAKA_HOST_API`.
    
If `ALPAKA_HOST_API` is defined, a CUDA or HIP compiler is required only for compiling device code or kernel launches.

This is a proof-of-concept implementation of the [fourth option](https://github.com/alpaka-group/alpaka/issues/1560#issuecomment-1008674258) discussed in the proposal #1560, including a new integration test for the host-only API.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -368,12 +368,12 @@ if(ALPAKA_ACC_GPU_CUDA_ENABLE)
             # libstdc++ since version 7 when GNU extensions are enabled (e.g. -std=gnu++11)
             # uses `__CUDACC__` to avoid defining overloads using non-standard `__float128`.
             # This is fixed in clang-11: https://github.com/llvm/llvm-project/commit/8e20516540444618ad32dd11e835c05804053697
-            if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0)
-                target_compile_definitions(alpaka INTERFACE "__CUDACC__")
+            if(CMAKE_CUDA_COMPILER_VERSION VERSION_LESS 11.0)
+                target_compile_definitions(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:__CUDACC__>)
             endif()
 
-            if(CMAKE_CXX_COMPILER_VERSION GREATER_EQUAL 11.0)
-                target_compile_options(alpaka INTERFACE "-Wno-unknown-cuda-version")
+            if(CMAKE_CUDA_COMPILER_VERSION GREATER_EQUAL 11.0)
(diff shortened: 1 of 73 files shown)


fwyzard:
I think the failures of the `hostOnlyAPI` test with Clang 9 are due to the Alpaka CMake rules defining `__CUDACC__` explicitly:
```cmake
            # libstdc++ since version 7 when GNU extensions are enabled (e.g. -std=gnu++11)
            # uses `__CUDACC__` to avoid defining overloads using non-standard `__float128`.
            # This is fixed in clang-11: https://github.com/llvm/llvm-project/commit/8e20516540444618ad32dd11e835c05804053697
            if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0)
                target_compile_definitions(alpaka INTERFACE "__CUDACC__")
            endif()
```
This breaks the check to see if a CUDA-aware compiler is being used, because `BOOST_LANG_CUDA` will be non-zero even if clang is _not_ being used in CUDA mode.

j-stephan:
Yes, this should probably be a generator expression:

```cmake
            # libstdc++ since version 7 when GNU extensions are enabled (e.g. -std=gnu++11)
            # uses `__CUDACC__` to avoid defining overloads using non-standard `__float128`.
            # This is fixed in clang-11: https://github.com/llvm/llvm-project/commit/8e20516540444618ad32dd11e835c05804053697
            if(CMAKE_CXX_COMPILER_VERSION VERSION_LESS 11.0)
                target_compile_definitions(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:__CUDACC__>)
            endif()
```

j-stephan:
Or, better yet, check for the `CMAKE_CUDA_COMPILER_VERSION` instead of `CXX`. This seems to be a remnant from the time when we didn't have CMake's native CUDA support available.

fwyzard:
> Or, better yet, check for the `CMAKE_CUDA_COMPILER_VERSION` instead of `CXX`. This seems to be a remnant from the time when we didn't have CMake's native CUDA support available.

This did not work, because CMake is still configured with `clang++-9` as a CUDA compiler, even when building the host-only application.

fwyzard:
> ```cmake
> target_compile_definitions(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:__CUDACC__>)
> ```

this worked, thanks !

fwyzard:
See #1567 for the final implementation, that uses the `ALPAKA_HOST_ONLY` symbol, and does not include the changes to the existing integration and unit tests.

PR #1567, opened 2022-01-11
Support the CUDA/HIP host API without a CUDA/HIP compiler

### LINKED ISSUE #908: CUDA backend requires nvcc also for library calls
Using Alpaka with the CUDA back-end requires compiling much more code with `nvcc` than using nativa CUDA would.

In general, when writing CUDA code, one has to compile with `nvcc` the files with device code (`__device__` functions and `__global__` kernels); while host code that calls the CUDA API runtime functions (e.g. `cudaMalloc()`, `cudaMemcpy()`, `cudaLaunchKernel()`, ...) can be compiled with `gcc`.
A simple example can be find in [simple-kernel/cuda](https://github.com/fwyzard/simple-kernel/tree/master/cuda). Note that while `kernel.cu` is compiled by `nvcc`, `test.cc` is compiled by `gcc`.

Instead, writing Alpaka code for the CUDA backed requires building also the part that deals with allocating memory buffers and launching kernels with `nvcc`.
The corresponding example can be found in [simple-kernel/alpaka](https://github.com/fwyzard/simple-kernel/tree/master/alpaka). Note how both `kernel.cc` and `test.cc` need to be compiled by `nvcc`.

Given that `nvcc` lags behid in terms of support for C++ standards and features, this adds severe limitations to what code can interoperate with Alpaka.

What kind of changes would be required inside Alpaka to add similar partial support of the CUDA back-end for the host compiler ?

### LINKED ISSUE #1560: [RFC] Support the CUDA/HIP host API without a CUDA/HIP compiler

fwyzard:
Introduce a new preprocessor symbol: `ALPAKA_HOST_ONLY`.
    
If `ALPAKA_HOST_ONLY` is defined, a CUDA or HIP compiler is required only for compiling device code or kernel launches.

Includes a new integration test for the host-only API.

See #1560 for the proposal behind this change, along with  discussion of alternative approaches.

### DIFF
--- a/include/alpaka/math/max/MaxStdLib.hpp
+++ b/include/alpaka/math/max/MaxStdLib.hpp
@@ -45,6 +45,8 @@ namespace alpaka
                         return fmax(x, y);
                     else
                         static_assert(!sizeof(Tx), "Unsupported data type");
+
+                    ALPAKA_UNREACHABLE(std::common_type_t<Tx, Ty>{});
                 }
             };
         } // namespace traits
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -186,8 +186,9 @@ namespace alpaka
                         return false;
                     }
                 }
+                return true;
             }
-            return true;
+            ALPAKA_UNREACHABLE(bool{});
         }
         ALPAKA_NO_HOST_ACC_WARNING
         ALPAKA_FN_HOST_ACC auto operator!=(Vec const& rhs) const -> bool
--- a/test/integ/hostOnlyAPI/CMakeLists.txt
+++ b/test/integ/hostOnlyAPI/CMakeLists.txt
@@ -1,0 +1,28 @@
+#
+# Copyright 2022 Andrea Bocci
+#
+# This file is part of alpaka.
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+#
+
+set(_TARGET_NAME "hostOnlyAPITest")
+
+append_recursive_files_add_to_src_group("src/" "src/" "cpp" _FILES_SOURCE)
+
+add_executable(
+    ${_TARGET_NAME}
+    ${_FILES_SOURCE})
+target_link_libraries(
+    ${_TARGET_NAME}
+    PRIVATE common)
+
+target_compile_definitions(${_TARGET_NAME} PRIVATE ALPAKA_HOST_ONLY)
+
+set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER "test/integ")
+
+if(ALPAKA_ACC_GPU_CUDA_ENABLE OR ALPAKA_ACC_GPU_HIP_ENABLE)
+    add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} ${_ALPAKA_TEST_OPTIONS})
+endif()
--- a/test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
+++ b/test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
@@ -1,0 +1,180 @@
+/* Copyright 2022 Andrea Bocci
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <alpaka/alpaka.hpp>
+#include <alpaka/test/acc/TestAccs.hpp>
+
+#include <catch2/catch.hpp>
+
+#include <cstdint>
+#include <cstring>
+#include <iostream>
+
+// fill an trivial type with std::memset
+template<typename T>
+constexpr T memset_value(int c)
+{
+    T t;
+    std::memset(&t, c, sizeof(T));
+    return t;
+}
+
+//! check if asynchronous (queue-ordered) memory buffers are supported by the given Accelerator
+template<typename TDev, typename TDim>
+static constexpr auto isAsyncBufferSupported() -> bool
+{
+#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
+    if constexpr(std::is_same_v<TDev, alpaka::DevCudaRt>)
+    {
+        return (CUDA_VERSION >= 11020) && (TDim::value == 1);
+    }
+    else
+#endif // ALPAKA_ACC_GPU_CUDA_ENABLED
+
+#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+        if constexpr(std::is_same_v<TDev, alpaka::DevHipRt>)
+    {
+        return false;
+    }
+    else
+#endif // ALPAKA_ACC_GPU_HIP_ENABLED
+
+#ifdef ALPAKA_ACC_ANY_BT_OACC_ENABLED
+        if constexpr(std::is_same_v<TDev, alpaka::DevOacc>)
+    {
+        return false;
+    }
+    else
+#endif // ALPAKA_ACC_ANY_BT_OACC_ENABLED
+
+#ifdef ALPAKA_ACC_ANY_BT_OMP5_ENABLED
+        if constexpr(std::is_same_v<TDev, alpaka::DevOmp5>)
+    {
+        return false;
+    }
+    else
+#endif // ALPAKA_ACC_ANY_BT_OMP5_ENABLED
+
+        return true;
+}
+
+template<typename TElem, typename TQueue, typename TExtent>
+auto allocAsyncBufIfSupported(TQueue const& queue, TExtent const& extent)
+    -> alpaka::Buf<alpaka::Dev<TQueue>, TElem, alpaka::Dim<TExtent>, alpaka::Idx<TExtent>>
+{
+    using Idx = alpaka::Idx<TExtent>;
+    if constexpr(isAsyncBufferSupported<alpaka::Dev<TQueue>, alpaka::Dim<TExtent>>())
+    {
+        return alpaka::allocAsyncBuf<TElem, Idx>(queue, extent);
+    }
+    else
+    {
+        return alpaka::allocBuf<TElem, Idx>(alpaka::getDev(queue), extent);
+    }
+}
+
+// 0- and 1- dimensional space
+using Idx = std::size_t;
+using Dim1D = alpaka::DimInt<1u>;
+using Vec1D = alpaka::Vec<Dim1D, Idx>;
+
+// enabled accelerators with 1-dimensional kernel space
+using TestAccs = alpaka::test::EnabledAccs<Dim1D, Idx>;
+
+TEMPLATE_LIST_TEST_CASE("hostOnlyAPI", "[hostOnlyAPI]", TestAccs)
+{
+    using Device = alpaka::Dev<TestType>;
+    using DeviceQueue = alpaka::Queue<Device, alpaka::NonBlocking>;
+
+    using Host = alpaka::DevCpu;
+    using HostQueue = alpaka::Queue<Host, alpaka::Blocking>;
+
+    // CPU host
+    auto const host = alpaka::getDevByIdx<Host>(0u);
+    INFO("using alpaka device: " << alpaka::getName(host))
+    HostQueue hostQueue(host);
+
+    // host buffer
+    auto h_buffer1 = alpaka::allocBuf<int, Idx>(host, Vec1D{Idx{42}});
+    INFO(
+        "host buffer allocated at " << alpaka::getPtrNative(h_buffer1) << " with "
+                                    << alpaka::extent::getExtentProduct(h_buffer1) << " element(s)")
+
+    // async host buffer
+    auto h_buffer2 = allocAsyncBufIfSupported<int>(hostQueue, Vec1D{Idx{42}});
+    INFO(
+        "second host buffer allocated at " << alpaka::getPtrNative(h_buffer2) << " with "
+                                           << alpaka::extent::getExtentProduct(h_buffer2) << " element(s)")
+
+    // host-side memset
+    const int value1 = 42;
+    const int expected1 = memset_value<int>(value1);
+    INFO("host-side memset")
+    alpaka::memset(hostQueue, h_buffer1, value1);
+    alpaka::wait(hostQueue);
+    CHECK(expected1 == *alpaka::getPtrNative(h_buffer1));
+
+    // host-side async memset
+    const int value2 = 99;
+    const int expected2 = memset_value<int>(value2);
+    INFO("host-side async memset")
+    alpaka::memset(hostQueue, h_buffer2, value2);
+    alpaka::wait(hostQueue);
+    CHECK(expected2 == *alpaka::getPtrNative(h_buffer2));
+
+    // host-host copies
+    INFO("buffer host-host copies")
+    alpaka::memcpy(hostQueue, h_buffer2, h_buffer1);
+    alpaka::wait(hostQueue);
+    CHECK(expected1 == *alpaka::getPtrNative(h_buffer2));
+    alpaka::memcpy(hostQueue, h_buffer1, h_buffer2);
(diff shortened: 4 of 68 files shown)


test/integ/hostOnlyAPI/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+append_recursive_files_add_to_src_group("src/" "src/" "cpp" _FILES_SOURCE)
+
+add_executable(
+    ${_TARGET_NAME}
+    ${_FILES_SOURCE})
+target_link_libraries(
+    ${_TARGET_NAME}
+    PRIVATE common)
+
+add_compile_definitions(ALPAKA_HOST_ONLY)
```
j-stephan: It is recommended to add these per target to avoid polluting the public definitions:

```suggestion
target_compile_definitions(${_TARGET_NAME} PRIVATE ALPAKA_HOST_ONLY)
```

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
```diff
@@ -0,0 +1,185 @@
+/* Copyright 2022 Andrea Bocci
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+// make sure the CPU_B_SEQ_T_SEQ backend is always available
```
j-stephan: Is this for your CMS build system? CMake should enable the sequential back-end by default if the test cases are enabled.

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
```diff
@@ -0,0 +1,185 @@
...
+constexpr T memset_value(int c)
+{
+    T t;
+    std::memset(&t, c, sizeof(T));
+    return t;
+}
+
+//! check if asynchronous (queue-ordered) memory buffers are supported by the given Accelerator
+template<typename TAcc>
+static constexpr auto isAsyncBufferSupported() -> bool
```
j-stephan, line 30: I've seen this in a few test cases now. Maybe it would make sense to refactor and put this in a shared header file?

include/alpaka/vec/Vec.hpp
```diff
@@ -186,8 +186,8 @@ namespace alpaka
                         return false;
                     }
                 }
+                return true;
```
j-stephan, line 189: What is the reason for this change? I believe this will cause `unreachable code` warnings with some known buggy compilers.

j-stephan (changes requested):
Overall, this looks like a good addition. I have a few remarks (see below).

This change might be a good opportunity to introduce linkable CMake targets for host and device code (think `alpaka::host` and `alpaka::device`) to make this separation also visible in our default build system. Right now this has to be explicitly enabled by passing `-DALPAKA_HOST_ONLY` to the compiler, right? So the default behaviour is unchanged?

fwyzard:
> Right now this has to be explicitly enabled by passing `-DALPAKA_HOST_ONLY` to the compiler, right? So the default behaviour is unchanged?

Yes, correct.

test/integ/hostOnlyAPI/CMakeLists.txt
fwyzard: OK.

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
fwyzard: Mhm, no, we should build the tests with the native CMake.

I guess I misunderstood something I read, that `ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED` was not guaranteed to always be available. I'll remove the definition.

j-stephan: By default it isn't, but we force-enable it if the examples and/or tests are switched on through CMake.

include/alpaka/vec/Vec.hpp
fwyzard: Actually it should be the opposite.

The old code
```c++
            if constexpr(TDim::value == 0)
            {
                return true;
            }
            else
            {
                ...
            }
            return true;
```
has an unreachable `return true` at the end, if `TDim::value == 0`.

Moving it into the `else` clause should fix it:
```c++
            if constexpr(TDim::value == 0)
            {
                return true;
            }
            else
            {
                ...
                return true;
            }
```

j-stephan: Ah, I phrased that wrong. The old code would indeed cause an unreachable code warning (and rightly so). The new code will cause the buggy compilers to complain though. `nvcc` (before CUDA 11.5) will believe that there is now a missing return statement at the end of the function because it is unable to correctly identify `if constexpr` code paths. That is why we recently introduced `ALPAKA_UNREACHABLE` to address this:

```suggestion
                return true; // no problem for conforming compilers
            }
            ALPAKA_UNREACHABLE(bool{}); // silence nvcc < 11.5 and other buggy compilers
        }
```

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
fwyzard: OK - should I make it into a trait ?

However, I would prefer to make this change in a separate PR, cleaning up all instances.

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
fwyzard: OK.

include/alpaka/vec/Vec.hpp
fwyzard: Ah, I see.

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
fwyzard: Mhm, it doesn't seem to always be the case, see https://github.com/alpaka-group/alpaka/runs/4853400473?check_suite_focus=true#step:4:2982 :
```
D:\a\alpaka\alpaka\test\integ\hostOnlyAPI\src\hostOnlyAPI.cpp(95,29): error C2039: 'AccCpuSerial': is not a member of 'alpaka' [D:\a\alpaka\alpaka\build\test\integ\hostOnlyAPI\hostOnlyAPITest.vcxproj]
```

fwyzard: Is it because this is compiled as a standard (host-only) application, instead of an "alpaka" application ?

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
j-stephan: A function or a trait are fine. A separate PR would probably be better because this doesn't really belong to this PR.

fwyzard:
> ERROR: Job failed: execution took longer than 3h0m0s seconds

could you kick the HIP runners ?

psychocoderHPC:
> > ERROR: Job failed: execution took longer than 3h0m0s seconds
> 
> could you kick the HIP runners ?

I restarted the jobs.

fwyzard:
> > > ERROR: Job failed: execution took longer than 3h0m0s seconds
> > 
> > 
> > could you kick the HIP runners ?
> 
> I restarted the jobs.

Thank you @psychocoderHPC ... could you restart them again ?

The tests themselves _should_ be fine, they ran successfully as part of https://github.com/alpaka-group/alpaka/pull/1568 which is a superset of this PR, but one of the HIP tests went again into timeout.

j-stephan:
Restarted. Apparently one of the AMD GPUs crashed and thus the node needed to be restarted by hand...

include/alpaka/math/max/MaxStdLib.hpp
```diff
@@ -45,6 +45,8 @@ namespace alpaka
                         return fmax(x, y);
                     else
                         static_assert(!sizeof(Tx), "Unsupported data type");
+
+                    ALPAKA_UNREACHABLE(std::common_type_t<Tx, Ty>{});
```
j-stephan, line 49: Neat!

test/integ/hostOnlyAPI/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+append_recursive_files_add_to_src_group("src/" "src/" "cpp" _FILES_SOURCE)
+
+add_executable(
+    ${_TARGET_NAME}
+    ${_FILES_SOURCE})
+target_link_libraries(
+    ${_TARGET_NAME}
+    PRIVATE common)
+
+target_compile_definitions(${_TARGET_NAME} PRIVATE ALPAKA_HOST_ONLY ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
```
j-stephan: The `ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED` shouldn't be necessary here, it is automatically added by the top-level `CMakeLists.txt`.

```suggestion
target_compile_definitions(${_TARGET_NAME} PRIVATE ALPAKA_HOST_ONLY)
```

j-stephan (changes requested):

test/integ/hostOnlyAPI/CMakeLists.txt
fwyzard: I did try without it, but it failed in the CI on the Windows NVCC targets.

j-stephan: Hm, it seems that the serial back-end is explicitly disabled for the Windows-Release runners. I'll look into this.

fwyzard: I've opened #1577 to run the tests without this change.

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
```diff
@@ -0,0 +1,181 @@
...
+    using DeviceQueue = alpaka::Queue<DeviceAcc, alpaka::NonBlocking>;
+
+    using HostAcc = alpaka::AccCpuSerial<Dim1D, Idx>;
+    using Host = alpaka::DevCpu;
+    using HostQueue = alpaka::Queue<HostAcc, alpaka::Blocking>;
+
+    // CPU host
+    auto const host = alpaka::getDevByIdx<Host>(0u);
+    INFO("Using alpaka accelerator: " << alpaka::getAccName<HostAcc>())
+    HostQueue hostQueue(host);
```
j-stephan: You can work around the serial back-end by instantiating the CPU types directly (we do this in other test cases, too, to avoid testing the serial back-end all the time). This is probably easier than enabling the serial back-end everywhere.

```suggestion
    using PltfHost = alpaka::PltfCpu;
    using Host = alpaka::DevCpu;
    using HostQueue = alpaka::Queue<Host, alpaka::Blocking>;

    // CPU host
    auto const host = alpaka::getDevByIdx<PltfHost>(0u);
    INFO("Using alpaka accelerator: " << alpaka::getName(host))
    HostQueue hostQueue(host);
```

fwyzard: Ah, I see... `AccCpuSerial` accelerator is available only if  `ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED` is defined, but the `PltfCpu` platform and `DevCpu` device are always available.

j-stephan: Exactly :-)

test/integ/hostOnlyAPI/CMakeLists.txt
fwyzard: OK, as expected #1577 failed few of the tests:
  - [linux_nvcc-11.4_clang-11_release](https://gitlab.com/hzdr/crp/alpaka/-/jobs/1991403986)
  - [windows_nvcc-11.2_cl-2019_release_cuda-only_c++17](https://github.com/alpaka-group/alpaka/runs/4867897549?check_suite_focus=true) and similar

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
fwyzard: OK, let's see if the simplified version passes all tests.

fwyzard:
OK, this implementation of the test looks actually cleaner, and passing the CI tests... thanks for the suggestions.

test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
fwyzard: Can you check #1578 for a possible implementation ?

j-stephan (approved):

fwyzard:
:+1:

PR #1570, opened 2022-01-13
Add Accessor-like interface to memory buffers and views

### LINKED ISSUE #1558: Accessors for 0-sized buffers and views
As a follow up to #1536 and #1433, I would like to add a specialisation for accessors for 0-dimensional (scalar) buffers and views.

Since for a scalar the `[]` operator does not make much sense, what interface do you prefer ?
  - add an `operator T&()` or `operator T const&() const` conversion operator to the underlying scalar type[^1]
  - add the `*` and `->` operators  
  - something else

[^1]: I'm not sure if this could support write-only access, though

fwyzard:
Memory buffers and views with different dimensions implement different operators:
  - 0-dimensional (scalar) types implement the indirection operator `*` and dereference operator `->`;
  - 1-dimensional type implement the array subscript operator `[]` with an integer argument;
  - all N-dimensional types implement the array subscript operator `[]` with an N-dimensional vector argument.

Update the buffer unit test and zero-dimensional buffer integration test to test the new interface.

### DIFF
--- a/test/unit/mem/buf/src/BufTest.cpp
+++ b/test/unit/mem/buf/src/BufTest.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2021 Axel Huebl, Benjamin Worpitz, Andrea Bocci
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Andrea Bocci
  *
  * This file is part of alpaka.
  *
@@ -241,4 +241,66 @@ TEMPLATE_LIST_TEST_CASE("memBufAsyncConstTest", "[memBuf]", alpaka::test::TestAc
     {
         INFO("Stream-ordered memory buffers are not supported in this configuration.")
     }
+}
+
+template<typename TAcc>
+static auto testBufferAccessorAdaptor(
+    alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> const& extent,
+    alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> const& index) -> void
+{
+    using Dev = alpaka::Dev<TAcc>;
+    using Pltf = alpaka::Pltf<Dev>;
+
+    using Elem = float;
+    using Dim = alpaka::Dim<TAcc>;
+    using Idx = alpaka::Idx<TAcc>;
+
+    // assume dimensionality up to 4
+    CHECK(Dim::value <= 4);
+
+    Dev const dev = alpaka::getDevByIdx<Pltf>(0u);
+
+    // alpaka::malloc
+    auto buf = alpaka::allocBuf<Elem, Idx>(dev, extent);
+
+    // check that the array subscript operator access the correct element
+    auto const& pitch = alpaka::getPitchBytesVec(buf);
+    INFO("buffer extent: " << extent << " elements")
+    INFO("buffer pitch: " << pitch << " bytes")
+    CHECK((index < extent).foldrAll(std::logical_and<bool>()));
+
+    uintptr_t base = reinterpret_cast<uintptr_t>(std::data(buf));
+    uintptr_t expected = base;
+    if constexpr(Dim::value > 1)
+    {
+        expected += static_cast<uintptr_t>(pitch[1] * index[0]);
+    }
+    if constexpr(Dim::value > 2)
+    {
+        expected += static_cast<uintptr_t>(pitch[2] * index[1]);
+    }
+    if constexpr(Dim::value > 3)
+    {
+        expected += static_cast<uintptr_t>(pitch[3] * index[2]);
+    }
+    expected += sizeof(Elem) * static_cast<uintptr_t>(index[Dim::value - 1]);
+
+    INFO("element " << index << " expected at offset " << expected - base)
+    INFO("element " << index << " returned at offset " << reinterpret_cast<uintptr_t>(&buf[index]) - base)
+    CHECK(reinterpret_cast<Elem*>(expected) == &buf[index]);
+
+    // check that an out-of-bound access is detected
+    CHECK_THROWS_AS((void) buf.at(extent), std::out_of_range);
+}
+
+TEMPLATE_LIST_TEST_CASE("memBufAccessorAdaptorTest", "[memBuf]", alpaka::test::TestAccs)
+{
+    using Acc = TestType;
+    using Dim = alpaka::Dim<Acc>;
+    using Idx = alpaka::Idx<Acc>;
+
+    auto extent = alpaka::createVecFromIndexedFn<Dim, alpaka::test::CreateVecWithIdx<Idx>::template ForExtentBuf>();
+    auto index = alpaka::createVecFromIndexedFn<Dim, alpaka::test::CreateVecWithIdx<Idx>::template ForOffset>();
+
+    testBufferAccessorAdaptor<Acc>(extent, index);
 }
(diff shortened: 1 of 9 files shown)


test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -242,3 +242,76 @@ TEMPLATE_LIST_TEST_CASE("memBufAsyncConstTest", "[memBuf]", alpaka::test::TestAc
...
+    {
+        alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> out_of_bounds = extent;
+        out_of_bounds[0] += 1;
+        (void) buf.at(out_of_bounds);
+    }
+    catch(std::out_of_range&)
+    {
+        out_of_bounds_detected = true;
+    }
+    CHECK(out_of_bounds_detected == true);
```
bernhardmgruber: Catch2 should have a native construct for such tests: `CHECK_THROWS_AS`, although I never used that facility. See also: https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#exceptions

Maybe something like this works:
```suggestion
    bool out_of_bounds_detected = false;
        alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> out_of_bounds = extent;
        out_of_bounds[0] += 1;
        CHECK_THROWS_AS(buf.at(out_of_bounds), std::out_of_range);
```

include/alpaka/mem/view/AccessorAdaptor.hpp
```diff
@@ -0,0 +1,169 @@
...
+            using value_type = Elem<TView>;
+            using pointer = value_type*;
+            using const_pointer = value_type const*;
+            using reference = value_type&;
+            using const_reference = value_type const&;
+
+        public:
+            AccessorAdaptor()
+            {
+                static_assert(experimental::traits::internal::IsView<TView>::value, "");
```
bernhardmgruber: We have C++17 now:
```suggestion
                static_assert(experimental::traits::internal::IsView<TView>::value);
```

include/alpaka/mem/view/AccessorAdaptor.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+#include <sstream>
+#include <type_traits>
+
+namespace alpaka
+{
+    namespace internal
+    {
+        template<typename TView>
+        struct AccessorAdaptor
```
bernhardmgruber: I think the type's name is misleading. First, it has very little to do with `alpaka::experimental::Accessor`, which is also already a bad name. I would try to avoid the name "accessor" because it has a very broad meaning. Also the term "accessor" in C++23 will have a different meaning (see `std::mdspan`). Second, the type does not strictly implement an adaptor, but a mixin.

Since this CRTP mixin is intended for views, you could for example name it `ViewAccessOps`.

include/alpaka/mem/view/AccessorAdaptor.hpp
```diff
@@ -0,0 +1,169 @@
...
+                return getPtrNative(*static_cast<TView const*>(this));
+            }
+
+            ALPAKA_FN_HOST auto operator*() -> reference
+            {
+                static_assert(
+                    alpaka::Dim<TView>::value == 0,
+                    "operator* is only valid for Buffers and Views of dimension 0");
+                return *data();
+            }
```
bernhardmgruber: I wonder whether we should SFINAE the operator away instead of static asserting. Opinions? Applies to all member functions with `static_assert`s on the dimensionality.

include/alpaka/mem/view/AccessorAdaptor.hpp
```diff
@@ -0,0 +1,169 @@
...
+            template<std::size_t TDim, typename TIdx>
+            ALPAKA_FN_HOST auto ptr_at(Vec<DimInt<TDim>, TIdx> index) const -> const_pointer
+            {
+                using Idx = alpaka::Idx<TView>;
+                static_assert(
+                    alpaka::Dim<TView>::value == TDim,
+                    "the index type must have the same dimensionality as the Buffer or View");
+                static_assert(
+                    std::is_convertible_v<TIdx, Idx>,
+                    "the index type must be convertible to the index of the Buffer or View");
```
bernhardmgruber: You could just use the `TView`'s dimensionality and avoid the check:

```suggestion
            template<typename TIdx>
            ALPAKA_FN_HOST auto ptr_at(Vec<alpaka::Dim<TView>, TIdx> index) const -> const_pointer
            {
                using Idx = alpaka::Idx<TView>;
                static_assert(
                    std::is_convertible_v<TIdx, Idx>,
                    "the index type must be convertible to the index of the Buffer or View");
```

bernhardmgruber (changes requested):
Generally, looks good to me.

include/alpaka/mem/view/AccessorAdaptor.hpp
fwyzard: OK, but just FYI I copied this approach from [include/alpaka/mem/view/ViewAccessor.hpp](https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/mem/view/ViewAccessor.hpp#L238)

include/alpaka/mem/view/AccessorAdaptor.hpp
fwyzard: I tried to do that, but I couldn't find a way to make it work.
The problem sees to that, due to the CRTP mixin approach, the type of `TView` is still incomplete at the time of the class definition, so it cannot be checked.
If you have suggestions for making it work, I can try to use the SFINAE approach instead of `static_assert`.

include/alpaka/mem/view/AccessorAdaptor.hpp
fwyzard: Same problem as above: `alpaka::Dim<TView>` does not work during the class definition, so it cannot be used as part of the method signature.

include/alpaka/mem/view/AccessorAdaptor.hpp
fwyzard: > it has very little to do with `alpaka::experimental::Accessor`

well, it implements the same interface, basically allowing to use Buffer or View directly in place of an Accessor

> ... which is also already a bad name

this is probably outside the scope of his PR.





> Since this CRTP mixin is intended for views, you could for example name it `ViewAccessOps`.

Other suggestions for a better name ?

include/alpaka/mem/view/AccessorAdaptor.hpp
bernhardmgruber: Sure, that feature was added before we had C++17. @sliwowitz has some refactorings for such things in the pipeline.

bernhardmgruber: Sure, that feature was added before we had C++17. @sliwowitz has some refactorings for such things in the pipeline.

include/alpaka/mem/view/AccessorAdaptor.hpp
bernhardmgruber: You're right, you cannot get the dimensionality from `TView`. In LLAMA I had the same issue and went for an extra integral template parameter.

bernhardmgruber (approved):

include/alpaka/mem/view/AccessorAdaptor.hpp
bernhardmgruber: Ok let's leave it as is for now.

include/alpaka/mem/view/AccessorAdaptor.hpp
bernhardmgruber: Ok let's leave it as is for now.

PR #1404, opened 2021-08-30
fix CUDA 11.3+ compile

psychocoderHPC:
Fix error shown with CUDA 11.3+

```
/builds/hzdr/crp/alpaka/example/reduce/src/reduce.cpp:93:38: error: '__T14' was not declared in this scope
     auto const taskKernelReduceMain(alpaka::createTaskKernel<Acc>(
                                 ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Some kind of error was already fixed with #1293, it is not clear why this piece of code was not producing errors before.
@SimeonEhrig said maybe the bug is coming because of the fix in  #1392, before #1392 maybe some compile options were not set correctly.

Issue #1293, opened 2021-04-19
CUDA 11.3 support


psychocoderHPC:
CUDA 11.3. is released and should be added.

There are some issues with CUDA 11.3

```
/alpaka/event/EventGenericThreads.hpp:280:19: error: ‘__T30’ was not declared in this scope
                 auto vQueues(dev.getAllQueues())
```

This can be solved by changing:

```C++
auto vQueues(dev.getAllQueues());
```
to
```C++
auto vQueues = dev.getAllQueues();
```

psychocoderHPC:
I am currently work on adding CUDA 11.3 and fixing the issues.

sbastrakov:
Do we need to backport it to 0.6.1? The 0.6.0 release didn't support CUDA 11.3.

j-stephan:
0.6 is older than CUDA 11.3, so I'd say no. We are backporting specific features because of users who use older CUDA versions. I doubt they're planning to adopt CUDA 11.3 and stick with alpaka 0.6.x at the same time.

psychocoderHPC:
> Do we need to backport it to 0.6.1? The 0.6.0 release didn't support CUDA 11.3.

Yes I would like to port it back. The reason is that we have already an issue in PIConGPU: https://github.com/ComputationalRadiationPhysics/picongpu/issues/3591 and I assume we will get soon more.

I assume that a backport is easy and should provide a low amount of conflicts.

Never the less we can discuss it tomorrow in the developer meeting.

PR #1574, opened 2022-01-18
OpenMP 5: Add shared mem Implementation based on `omp allocate`

jkelling:
This PR adds:
*  `BlockSharedMemStOmp5BuiltIn` which is based on the OpenMP 5.0 `allocate` directive with the `omp_pteam_mem_alloc` allocator.
* `BlockSharedMemDynOmp5BuiltIn` which splits into two strategies (subclasses):
  * `BlockSharedMemDynOmp5BuiltInFixed`: Mostly a simplification of `BlockSharedMemDynMember` to be used with the `St` built-in.
  * `BlockSharedMemDynOmp5BuiltInOmpAlloc`: An experiment based on OpenMP's `omp_alloc()` API call, which according to the standard should also be allowed in target regions.
* CMake option `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM` to enable OMP5 built-in smem and select the DYN strategy. The ability to en-/disable the built-in is required because some compilers lie about their support level and implementations are currently incomplete/buggy.
* A test (`blockSharedSharingTest`) which check if shared memory is shared correctly and produces results which can be used to determine which way sharing my be wrong.

These shared memory implementations (`BlockSharedMemStOmp5BuiltIn` + `BlockSharedMemDynOmp5BuiltInOmpAlloc`) could also be use with the `CPU_B_SEQ_T_OMP2` backend but this should wait until OpenMP 5.0 implementations mature.

See changes to [README_OMP5.md](https://github.com/alpaka-group/alpaka/compare/develop...jkelling:omp5?expand=1#diff-4d29ff3e097efc17e546301c3e67474a96c0a4e4f3f395f5dbc2df90943d9e66) for details.


README_OMP5.md
```diff
@@ -49,6 +49,61 @@ environment requirements. Add flags to set the required compiler and linker flag
     -DCMAKE_CXX_FLAGS=""
   ```
 
+## Block-shared Memory
+
+Shared memory if implemented using a small object allocator in
```
j-stephan: This sentence sounds a bit weird. Could you rephrase it?

test/unit/block/sharedSharing/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright 2022 Benjamin Worpitz, Axel Huebl, Jan Stephan, Jeffrey Kelling
```
j-stephan, line 2: You don't need to copy the authors of other test cases here if this is your own test case and only you touched this particular file.

include/alpaka/block/shared/st/BlockSharedMemStOmp5BuiltIn.hpp
```diff
@@ -0,0 +1,49 @@
...
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#if(defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) || defined(ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENBLED))                           \
```
j-stephan: Typo:

```suggestion
#if(defined(ALPAKA_ACC_ANY_BT_OMP5_ENABLED) || defined(ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLED))                           \
```

Also, why is the OpenMP 2 back-end tested here?

include/alpaka/block/shared/dyn/BlockSharedMemDynOmp5BuiltIn.hpp
```diff
@@ -0,0 +1,107 @@
...
+
+    private:
+    };
+#    endif
+
+    //! The OpenMP 5.0 block dynamic shared memory allocator with fixed amount of smem
+    class alignas(core::vectorization::defaultAlignment) BlockSharedMemDynOmp5BuiltInFixed
+        : public BlockSharedMemDynOmp5BuiltIn
+    {
+        std::array<std::byte, BlockSharedDynMemberAllocKiB << 10u> m_fixed;
```
j-stephan: The white space surrounding the `10u` looks a bit weird here.

README_OMP5.md
```diff
@@ -49,6 +49,61 @@ environment requirements. Add flags to set the required compiler and linker flag
...
+* `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM=DYN_FIXED`: Use `omp allocate`, use a
+  fixed size team-shared array for dynamic shared mem (fixed size is
+  `ALPAKA_BLOCK_SHARED_DYN_MEMBER_KIB`).
+* `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM=DYN_ALLOC`: Use `omp allocate`, use a
+  `omp_alloc()` API call in target region to allocate dynamic shared memory. The
+  standard appears to allow for this, but is not useful for some reasons:
+  * In the best case, this would lead to an on-device `malloc` on GPU, which has
+    bad performance and does not use on-chip memory.
+  * At least in clang GPU targets (nvptx64, hsa), the symbols `omp_alloc` and `omp_free`
+    are undefined (linker error, code compiles).
```
j-stephan:
```suggestion
OpenMP 5 offers the directive `omp allocate allocator(omp_pteam_mem_alloc)`
(used by `BlockSharedMemStOmp5BuiltIn`) which can in theory be used for *static*
shared memory variables. There is no useful built-in support for dynamic
block-shared memory to go with that. Usage of the built-in can be configured
using the `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM` flag:
* `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM=OFF`: Do not use `omp allocate` (default,
  only available behavior with OpenMP < 5).
* `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM=DYN_FIXED`: Use `omp allocate`, use a
  fixed size team-shared array for dynamic shared mem (fixed size is
  `ALPAKA_BLOCK_SHARED_DYN_MEMBER_KIB`).
* `ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM=DYN_ALLOC`: Use `omp allocate`, use a
  `omp_alloc()` API call in the target region to allocate dynamic shared memory. The
  standard appears to allow this, but is not useful for some reasons:
  * In the best case, this would lead to an on-device `malloc` on GPU, which has
    bad performance and does not use on-chip memory.
  * At least in clang GPU targets (nvptx64, hsa), the symbols `omp_alloc` and `omp_free`
    are undefined (linker error, code compiles).
```

Also, the `DYN_ALLOC` case sounds like something that no one would want to use. Did I understand that correctly?

README_OMP5.md
```diff
@@ -49,6 +49,61 @@ environment requirements. Add flags to set the required compiler and linker flag
...
+
+### Compiler support
+
+[blockSharedSharingTest](test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp) tests correct sharing.
+
+| compiler | target | `OFF` | `DYN_FIXED` | `DYN_ALLOC` |
+| --- | --- | --- | --- | --- |
+| clang 14 (1.) | x86 | S | S (2.) | S (2.) |
```
j-stephan: Would it be possible to use some emojis here, like in the compiler support matrix of our `README.md`?

cmake/alpakaCommon.cmake
```diff
@@ -156,6 +156,9 @@ set(ALPAKA_OFFLOAD_MAX_BLOCK_SIZE "256" CACHE STRING "Maximum number threads per
 option(ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST "Allow host-only contructs like assert in offload code in debug mode." ON)
 set(ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB "47" CACHE STRING "Kibibytes (1024B) of memory to allocate for block shared memory for backends requiring static allocation (includes CPU_B_OMP2_T_SEQ, CPU_B_TBB_T_SEQ, CPU_B_SEQ_T_SEQ)")
 
+set(ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM "OFF" CACHE STRING "Whether to use OMP5 built-in directive for block-shared memory and how to treat dynamic shared memory.")
```
```suggestion
set(ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM "OFF" CACHE STRING "Whether to use OMP5 built-in directives for block-shared memory and how to treat dynamic shared memory.")
```

include/alpaka/block/shared/OffloadUseBuiltInSharedMem.hpp
```diff
@@ -0,0 +1,26 @@
...
+    enum class OffloadUseBuiltInSharedMem : char
+    {
+        OFF, //! Do not use built-in shared memory facilites, use BlockSharedMemDynMember
+        DYN_FIXED, //! Use built-in shared memory, allocate fixed space for dyn smem
+        DYN_ALLOC //! Use built-in shared memory, use runtime allocation API
+    };
+
+    constexpr OffloadUseBuiltInSharedMem OFFLOAD_USE_BUILTIN_SHARED_MEM
+        = OffloadUseBuiltInSharedMem::ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM;
+} // namespace alpaka
```
j-stephan: How about something like this?

```suggestion
namespace alpaka
{
    struct OffloadBuiltInSharedMemOff {};
    struct OffloadBuiltInSharedMemFixed {};
    struct OffloadBuiltInSharedMemAlloc {};
#ifdef ALPAKA_OFFLOAD_BUILTIN_SHARED_MEM_FIXED
    using OffloadBuiltInSharedMem = OffloadBuiltInSharedMemFixed;
#elif defined(ALPAKA_OFFLOAD_BUILTIN_SHARED_MEM_ALLOC)
    using OffloadBuiltInSharedMem = OffloadBuiltInSharedMemAlloc;
#else
    using OffloadBuiltInSharedMem = OffloadBuiltInSharedMemOff;
#endif
} // namespace alpaka
```

This would interact with the other suggested changes below and allow for user extensibility.

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -50,13 +53,56 @@
...
+            : public BlockSharedMemDynOmp5BuiltInOmpAlloc
+            , public BlockSharedMemStOmp5BuiltIn
+        {
+            AccOmp5BlockSharedMem(std::size_t sizeBytes)
+                : BlockSharedMemDynOmp5BuiltInOmpAlloc(sizeBytes)
+                , BlockSharedMemStOmp5BuiltIn()
+            {
+            }
+        };
+#    endif
```
j-stephan: With the changes from above this would like this:

```suggestion
        template<typename TBuiltinSharedMem = OffloadBuiltInSharedMemOff>
        struct AccOmp5BlockSharedMem
            : public BlockSharedMemDynMember<>
            , public BlockSharedMemStOmp5
        {
            AccOmp5BlockSharedMem(std::size_t sizeBytes)
                : BlockSharedMemDynMember<>(sizeBytes)
                , BlockSharedMemStOmp5(staticMemBegin(), staticMemCapacity())

            {
            }
        };

#    if _OPENMP >= 201811
        template<>
        struct AccOmp5BlockSharedMem<OffloadBuiltInSharedMemFixed>
            : public BlockSharedMemDynOmp5BuiltInFixed
            , public BlockSharedMemStOmp5BuiltIn
        {
            AccOmp5BlockSharedMem(std::size_t sizeBytes)
                : BlockSharedMemDynOmp5BuiltInFixed(sizeBytes)
                , BlockSharedMemStOmp5BuiltIn()
            {
            }
        };

        template<>
        struct AccOmp5BlockSharedMem<OffloadBuiltInSharedMemAlloc>
            : public BlockSharedMemDynOmp5BuiltInOmpAlloc
            , public BlockSharedMemStOmp5BuiltIn
        {
            AccOmp5BlockSharedMem(std::size_t sizeBytes)
                : BlockSharedMemDynOmp5BuiltInOmpAlloc(sizeBytes)
                , BlockSharedMemStOmp5BuiltIn()
            {
            }
        };
#    endif
```

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -73,8 +119,12 @@ namespace alpaka
             AtomicOmpBuiltIn     // thread atomics
         >,
         public math::MathStdLib,
-        public BlockSharedMemDynMember<>,
-        public BlockSharedMemStOmp5,
+#    if _OPENMP >= 201811
+        public detail::AccOmp5BlockSharedMem<OFFLOAD_USE_BUILTIN_SHARED_MEM>,
```
j-stephan:
```suggestion
        public detail::AccOmp5BlockSharedMem<OffloadBuiltInSharedMem>, // default value determined by CMake flag
```

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -73,8 +119,12 @@ namespace alpaka
...
+#    else
+          public detail::AccOmp5BlockSharedMem<OffloadUseBuiltInSharedMem::OFF>
```
j-stephan:
```suggestion
          public detail::AccOmp5BlockSharedMem<OffloadBuiltInSharedMemOff>
```

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -109,10 +159,11 @@ namespace alpaka
                   AtomicOmpBuiltIn // atomics between threads
                   >()
             , math::MathStdLib()
-            , BlockSharedMemDynMember<>(blockSharedMemDynSizeBytes)
-            ,
-            //! \TODO can with some TMP determine the amount of statically alloced smem from the kernelFuncObj?
-            BlockSharedMemStOmp5(staticMemBegin(), staticMemCapacity())
+#    if _OPENMP >= 201811
+            , detail::AccOmp5BlockSharedMem<OFFLOAD_USE_BUILTIN_SHARED_MEM>(blockSharedMemDynSizeBytes)
```
j-stephan:
```suggestion
            , detail::AccOmp5BlockSharedMem<OffloadBuiltInSharedMem>(blockSharedMemDynSizeBytes)
```

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -109,10 +159,11 @@ namespace alpaka
...
+#    else
+            , detail::AccOmp5BlockSharedMem<OffloadUseBuiltInSharedMem::OFF>(blockSharedMemDynSizeBytes)
```
j-stephan:
```suggestion
            , detail::AccOmp5BlockSharedMem<OffloadBuiltInSharedMemOff>(blockSharedMemDynSizeBytes)
```

cmake/alpakaCommon.cmake
```diff
@@ -605,6 +608,7 @@ if(ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST)
 endif()
 target_compile_definitions(alpaka INTERFACE "ALPAKA_OFFLOAD_MAX_BLOCK_SIZE=${ALPAKA_OFFLOAD_MAX_BLOCK_SIZE}")
 target_compile_definitions(alpaka INTERFACE "ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=${ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB}")
+target_compile_definitions(alpaka INTERFACE "ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM=${ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM}")
```
```suggestion
if(ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM STREQUAL "DYN_FIXED")
    target_compile_definitions(alpaka INTERFACE ALPAKA_OFFLOAD_BUILTIN_SHARED_MEM_FIXED)
elseif(ALPAKA_OFFLOAD_USE_BUILTIN_SHARED_MEM STREQUAL "DYN_ALLOC")
    target_compile_definitions(alpaka INTERFACE ALPAKA_OFFLOAD_BUILTIN_SHARED_MEM_ALLOC)
endif()
```

test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <alpaka/test/acc/TestAccs.hpp>
+#include <alpaka/test/queue/Queue.hpp>
+#include <alpaka/workdiv/Traits.hpp>
+#include <alpaka/workdiv/WorkDivMembers.hpp>
+
+#include <catch2/catch.hpp>
+
+//! This tests checks if block-shared memory is shared correctly: only between all threads in a block.
+//!
+//! The check is done each thread atomically adding 1 to a variable `shared`. Possible outcomes are:
```
j-stephan:
```suggestion
//! The check is done by each thread atomically adding 1 to a variable `shared`. Possible outcomes are:
```

test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp
```diff
@@ -0,0 +1,154 @@
...
+{
+    auto const blockThreadIdx(alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc)[0u]);
+    auto const gridBlockIdx(alpaka::getIdx<alpaka::Grid, alpaka::Blocks>(acc)[0u]);
+
+    if(blockThreadIdx == 0u)
+        shared = 0u;
+
+    alpaka::syncBlockThreads(acc);
+
+    // In the expected case We need only `hierarchy::Blocks` here, but in
```
j-stephan:
```suggestion
    // In the expected case we only need `hierarchy::Blocks` here, but in
```

j-stephan (changes requested):
Thanks! As discussed offline a few suggestions:

README_OMP5.md
jkelling:
```suggestion
Shared memory is implemented using a small object allocator in
```

include/alpaka/block/shared/st/BlockSharedMemStOmp5BuiltIn.hpp
jkelling: ~~Initially I also considered also testing these smem implementations with this backend. This change should not be in this PR, though: Originally I placed my test case in this file and then these the existing test cases triggered compiler bugs on their own. With the new test case being separate this work-around is not necessary.~~
From the conversations list, I thought this was in a different place.

include/alpaka/block/shared/dyn/BlockSharedMemDynOmp5BuiltIn.hpp
jkelling: Indeed. This is how clang-format want it though. For example, if I place a space after `10u`, clang-format will remove it. I can do this:
```suggestion
        std::array<std::byte, (BlockSharedDynMemberAllocKiB << 10u)> m_fixed;
```

README_OMP5.md
jkelling: If my assumptions about how this would need to be implemented on GPU, then no one would want to use it there. On CPU this could be viable.

include/alpaka/block/shared/OffloadUseBuiltInSharedMem.hpp
jkelling: Looks good.

I see on issue, though: While this allows the user to add new variants on which they can also specialize `AccOmp5BlockSharedMem<>` below, they cannot make `AccOmp5` use any custom strategy, because the complie-time switch above always defines `alpaka::OffloadBuiltInSharedMem`. I guess, to fully allow extensibility, one would need to add a flag like `ALPAKA_OFFLOAD_BUILTIN_SHARED_MEM_USER` which would leave `OffloadBuiltInSharedMem` undefined, allowing the user to define it, before including `AccOmp5.hpp`.

include/alpaka/block/shared/st/BlockSharedMemStOmp5BuiltIn.hpp
jkelling: These implementations could in theory be used for `Omp2Threads` and may be better then `BlockScharedMem*Member`, depending on what the compiler and runtime are doing, so I prefer to enable them for Omp2Threads, too.

jkelling:
Cuda 11.3 and 11.4 do not seem to [like this code](https://gitlab.com/hzdr/crp/alpaka/-/pipelines/450530610). Is anyone familiar with this type of error:
```
test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp:35:25: error: use of undeclared identifier '__T12'
const auto gridBlockIdx(__T12(alpaka::getIdx< alpaka::origin::Grid, alpaka::unit::Blocks> (acc)[0U])); 
                        ^
```

or this one:

```
test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp:121:1: error: no matching function for call to 'blockSharedMemSharingTestKernelHelper'
blockSharedMemSharingTestKernelHelper(acc, sums, shared); 
^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]
/builds/hzdr/crp/alpaka/test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp:32:1: note: candidate template ignored: substitution failure [with TAcc = alpaka::AccCpuSerial<std::integral_constant<unsigned long, 1>, unsigned int>]
blockSharedMemSharingTestKernelHelper(const TAcc &acc, uint32_t *sums, uint32_t &shared) 
^
```

@psychocoderHPC are these known bugs in 11.3 and 11.4 (there are no such issues in 11.2 and 11.5)

README_OMP5.md
jkelling: Sure, the ones used there are a bit binary though... . I updated the `README_OMP5` only with those two.

test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp
```diff
@@ -0,0 +1,154 @@
...
+//!
+//! The check is done by each thread atomically adding 1 to a variable `shared`. Possible outcomes are:
+//! * `shared < blockThreadCount`: memory is not shared between all threads of a block; usually `shared == 1`: memory
+//!   is thread-private (fail)
+//! * `shared > blockThreadCount`: memory is shared between blocks (fail)
+//! * `shared == blockThreadCount`: memory is shared correctly (pass)
+template<typename TAcc>
+ALPAKA_FN_ACC void blockSharedMemSharingTestKernelHelper(TAcc const& acc, std::uint32_t* sums, std::uint32_t& shared)
+{
+    auto const blockThreadIdx(alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc)[0u]);
```
psychocoderHPC: This must be 

```C++
auto const blockThreadIdx = alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc)[0u];
auto const gridBlockIdx = alpaka::getIdx<alpaka::Grid, alpaka::Blocks>(acc)[0u];
```

seee: https://github.com/alpaka-group/alpaka/pull/1404 and #1293

psychocoderHPC (dismissed):

psychocoderHPC:
> Cuda 11.3 and 11.4 do not seem to [like this code](https://gitlab.com/hzdr/crp/alpaka/-/pipelines/450530610). Is anyone familiar with this type of error:
> 
> ```
> test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp:35:25: error: use of undeclared identifier '__T12'
> const auto gridBlockIdx(__T12(alpaka::getIdx< alpaka::origin::Grid, alpaka::unit::Blocks> (acc)[0U])); 
>                         ^
> ```
> 
> or this one:
> 
> ```
> test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp:121:1: error: no matching function for call to 'blockSharedMemSharingTestKernelHelper'
> blockSharedMemSharingTestKernelHelper(acc, sums, shared); 
> ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> [...]
> /builds/hzdr/crp/alpaka/test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp:32:1: note: candidate template ignored: substitution failure [with TAcc = alpaka::AccCpuSerial<std::integral_constant<unsigned long, 1>, unsigned int>]
> blockSharedMemSharingTestKernelHelper(const TAcc &acc, uint32_t *sums, uint32_t &shared) 
> ^
> ```
> 
> @psychocoderHPC are these known bugs in 11.3 and 11.4 (there are no such issues in 11.2 and 11.5)

This is a bug in CUDA 11.3 and 11.4 and should be fixed in CUDA with 11.5 or 11.6. 
The the provided link in https://github.com/alpaka-group/alpaka/pull/1574#pullrequestreview-856374151

test/unit/block/sharedSharing/src/BlockSharedMemSharing.cpp
jkelling: m( copied this from the `vectorAdd` but replaced `TIdx` by auto, not noticing, that this looses the type. ... not sure if I want to know why this even works in most compilers ... must be interpreted as one of the ~20 ways in C++ one should rather not use to initialize things... . Thanks.

include/alpaka/acc/AccOmp5.hpp
```diff
@@ -73,8 +119,12 @@ namespace alpaka
             AtomicOmpBuiltIn     // thread atomics
         >,
         public math::MathStdLib,
-        public BlockSharedMemDynMember<>,
-        public BlockSharedMemStOmp5,
+#    if _OPENMP >= 201811
+        public detail::AccOmp5BlockSharedMem<OffloadBuiltInSharedMem>, // default value determined by CMake flag
+#    else
+          public detail::AccOmp5BlockSharedMem<OffloadBuiltInSharedMemOff>
+        ,
```
j-stephan: That is an interesting formatting choice `clang-format` has made here...

README_OMP5.md
```diff
@@ -49,6 +49,61 @@ environment requirements. Add flags to set the required compiler and linker flag
...
+| clang 14 (1.) | nvptx | :white_check_mark: | :white_check_mark: | E (3.) |
+| clang 14 (1.) | hsa | C | C | E (3.) |
+| gcc 11 | x86 | :white_check_mark: | N | N |
+| gcc 11 | nvptx | :white_check_mark:/:x: (4.) | N | N |
+| nvhpc 22.1 | x86 | :white_check_mark: | N (5.) | N (5.) |
+
+Keys:
+* :white_check_mark:: Test Passes.
+* :x:l: Test fails, shared mem not shared.
+* :x:g: Test fails, shared mem gloal/shared too widely.
```
j-stephan, line 92: What do the `l` and `g` do here? In the markdown-aware file view they don't seem to do anything (other than exist).

j-stephan (changes requested):
One small thing, then we're good to go IMO.

README_OMP5.md
jkelling: they are for "local" and "global". Before there were "Fl" and "Fg" next to "F", which I replaced by ":x:".

include/alpaka/acc/AccOmp5.hpp
jkelling: I will change the style of the base-class list to have leading commas instead of trailing, like here:
https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/acc/AccOacc.hpp#L64-L82

This way clang-format makes more sensible choices...

I also like this style better, because this way end-of-line comments make more sense and you only need to change one line when appending/removing at the end.

jkelling:
I am still trying to convince clang in the CI to run more than one thread per team for one to work around https://github.com/alpaka-group/alpaka/issues/1576 and also because it would make the test better.

jkelling:
@j-stephan The PR is ready for merge now (I think there is not much need for squashing commits here).

The CI passed. I renamed/replaced one job, so the list of required jobs needs to be updated: `
linux_clang-11_debug_omp5 ->  linux_clang-13_debug_omp5`. This updated is just for good measure while I touched this job anyway. The underlying reason for the change to the job config remain with clang 13.

j-stephan (approved):

PR #1578, opened 2022-01-19
Add hasAsyncBufSupport trait, and update tests to use it

fwyzard:
Implement a new trait, `hasAsyncBufSupport<Dev, Dim>`, to check if the given device supports allocating stream-ordered buffers of the given dimensionality.

### DIFF
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -468,6 +468,18 @@ namespace alpaka
                 meta::DependentFalseType<TElem>::value,
                 "HIP/CUDA devices support only one-dimensional stream-ordered memory buffers.");
         };
+#    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && (CUDA_VERSION >= 11020)
+        //! The CUDA/HIP stream-ordered scalar memory allocation capability trait specialization.
+        template<>
+        struct HasAsyncBufSupport<DimInt<0u>, DevUniformCudaHipRt> : public std::true_type
+        {
+        };
+        //! The CUDA/HIP stream-ordered 1D memory allocation capability trait specialization.
+        template<>
+        struct HasAsyncBufSupport<DimInt<1u>, DevUniformCudaHipRt> : public std::true_type
+        {
+        };
+#    endif
 
         //! The BufUniformCudaHipRt CUDA/HIP device memory mapping trait specialization.
         template<typename TElem, typename TDim, typename TIdx>
--- a/include/alpaka/mem/buf/Traits.hpp
+++ b/include/alpaka/mem/buf/Traits.hpp
@@ -98,8 +98,17 @@ namespace alpaka
     //!
     //! TDev is the type of device to allocate the buffer on.
     //! TDim is the dimensionality of the buffer to allocate.
+    /* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because these
+    /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
     template<typename TDev, typename TDim>
     constexpr inline bool hasAsyncBufSupport = traits::HasAsyncBufSupport<TDim, TDev>::value;
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
 
     //! Maps the buffer into the memory of the given device.
     //!
--- a/include/alpaka/mem/buf/Traits.hpp
+++ b/include/alpaka/mem/buf/Traits.hpp
@@ -29,6 +29,12 @@ namespace alpaka
         template<typename TElem, typename TDim, typename TIdx, typename TDev, typename TSfinae = void>
         struct AsyncBufAlloc;
 
+        //! The stream-ordered memory allocator capability trait.
+        template<typename TDim, typename TDev>
+        struct HasAsyncBufSupport : public std::false_type
+        {
+        };
+
         //! The memory mapping trait.
         template<typename TBuf, typename TDev, typename TSfinae = void>
         struct Map;
@@ -87,6 +93,13 @@ namespace alpaka
     {
         return traits::AsyncBufAlloc<TElem, Dim<TExtent>, TIdx, alpaka::Dev<TQueue>>::allocAsyncBuf(queue, extent);
     }
+
+    //! Check if the given device can allocate a stream-ordered memory buffer of the given dimensionality.
+    //!
+    //! TDev is the type of device to allocate the buffer on.
+    //! TDim is the dimensionality of the buffer to allocate.
+    template<typename TDev, typename TDim>
+    constexpr inline bool hasAsyncBufSupport = traits::HasAsyncBufSupport<TDim, TDev>::value;
 
     //! Maps the buffer into the memory of the given device.
     //!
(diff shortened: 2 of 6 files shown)


fwyzard:
This PR is implemented on top of #1567, and can be rebased once that is merged.

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -88,6 +94,13 @@ namespace alpaka
         return traits::AsyncBufAlloc<TElem, Dim<TExtent>, TIdx, alpaka::Dev<TQueue>>::allocAsyncBuf(queue, extent);
     }
 
+    //! Check if the given device can allocate a stream-ordered memory buffer of the given dimensionality.
+    //!
+    //! TDev is the type of device to allocate the buffer on.
+    //! TDim is the dimensionality of the buffer to allocate.
```
fwyzard, line 100: `\tparam` is not supported for template _variables_

fwyzard:
Rebased to fix the conflicts.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -468,6 +468,18 @@ namespace alpaka
...
         };
+#    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && (CUDA_VERSION >= 11020)
+        //! The CUDA/HIP stream-ordered scalar memory allocation capability trait specialization.
+        template<>
+        struct IsAsyncBufferSupported<DimInt<0u>, DevUniformCudaHipRt> : public std::true_type
+        {
+        };
+        //! The CUDA/HIP stream-ordered 1D memory allocation capability trait specialization.
+        template<>
+        struct IsAsyncBufferSupported<DimInt<1u>, DevUniformCudaHipRt> : public std::true_type
```
j-stephan: This doesn't work for higher dimensions?

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -29,6 +29,12 @@ namespace alpaka
         template<typename TElem, typename TDim, typename TIdx, typename TDev, typename TSfinae = void>
         struct AsyncBufAlloc;
 
+        //! The stream-ordered memory allocator capability trait.
+        template<typename TDim, typename TDev>
+        struct IsAsyncBufferSupported : public std::false_type
```
j-stephan: A bit conciser:

```suggestion
        struct HasAsyncBufSupport : public std::false_type
```

j-stephan (changes requested):
Thanks, this looks useful. Two small things:

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
fwyzard: Currently no, because the implementation is based on `cudaMallocAsync`, and there are no 2D/3D counterparts.
If there is interest, the 2D/3D/ND versions could be implemented "by hand" on top of the linear memory returned by `cudaMallocAsync` and making some conservative assumptions about what `pitch` to use, kind of like it's done in the CPU case.

include/alpaka/mem/buf/Traits.hpp
fwyzard: OK, I can make the change.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
j-stephan: No, the default behaviour should be fine for now. 1D it is then.

bernhardmgruber (dismissed):

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -88,6 +94,13 @@ namespace alpaka
         return traits::AsyncBufAlloc<TElem, Dim<TExtent>, TIdx, alpaka::Dev<TQueue>>::allocAsyncBuf(queue, extent);
     }
 
+    //! Check if the given device can allocate a stream-ordered memory buffer of the given dimensionality.
+    //!
+    //! TDev is the type of device to allocate the buffer on.
+    //! TDim is the dimensionality of the buffer to allocate.
+    template<typename TDev, typename TDim>
+    constexpr bool hasAsyncBufferSupport = traits::HasAsyncBufferSupport<TDim, TDev>::value;
```
j-stephan: Sorry, I only saw this now:

```suggestion
    constexpr inline bool hasAsyncBufferSupport = traits::HasAsyncBufferSupport<TDim, TDev>::value;
```

Otherwise the compiler/linker might complain about symbol duplication if this is used in multiple translation units.

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -29,6 +29,12 @@ namespace alpaka
         template<typename TElem, typename TDim, typename TIdx, typename TDev, typename TSfinae = void>
         struct AsyncBufAlloc;
 
+        //! The stream-ordered memory allocator capability trait.
+        template<typename TDim, typename TDev>
+        struct HasAsyncBufferSupport : public std::false_type
```
j-stephan: Why not

```suggestion
        struct HasAsyncBufSupport : public std::false_type
```

? This would keep the trait consistent with the rest of the `Buf` API.

j-stephan (changes requested):

j-stephan:
Please rebase as well to pull in the latest CI changes.

include/alpaka/mem/buf/Traits.hpp
fwyzard: Sure.

include/alpaka/mem/buf/Traits.hpp
fwyzard: Good point - I'm not sure why I didn't get any errors in the tests or in our code base, but adding the `inline` is definitely better.

j-stephan (dismissed):

fwyzard:
Rebased and updated to address the last comments.

include/alpaka/mem/buf/Traits.hpp
fwyzard: Actually, looks like at least some versions of clang do not like the change:
```
/home/runner/work/alpaka/alpaka/include/alpaka/mem/buf/Traits.hpp:102:27: error: no previous extern declaration for non-static variable 'hasAsyncBufSupport' [-Werror,-Wmissing-variable-declarations]
    constexpr inline bool hasAsyncBufSupport = traits::HasAsyncBufSupport<TDim, TDev>::value;
                          ^
```

j-stephan: Ah yes, I've encountered this in the past, too. clang 5 and 6 are buggy. Guard the line with this:

```c++
/* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because these /  clang versions incorrectly warn about a missing 'extern'. */
#if BOOST_COMP_CLANG
#    pragma clang diagnostic push
#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
#endif

/* Your code */

#if BOOST_COMP_CLANG
#    pragma clang diagnostic pop
#endif
```

fwyzard: Looking at the clang archives, this is a bug [reported for clang 5 and 6](https://github.com/llvm/llvm-project/issues/35210) and [fixed](https://code.ornl.gov/ecpcitest/proteas/llvm-project/-/commit/742e3215bf9e9aa657d13201ccb138ab2d0090fa) in later releases.

fwyzard: Ah, I see that you encountered the same problem in https://github.com/alpaka-group/alpaka/blob/4096982a4befcbd39ee52659699aaaaa704f9048/include/alpaka/core/Decay.hpp#L36-L46 , so I'll copy the same workaround.

fwyzard: > Ah yes, I've encountered this in the past, too. clang 5 and 6 are buggy. Guard the line with this: ...

thanks @j-stephan, for some reason I wasn't getting notified about these comments :-/

fwyzard:
Re-pushed to retrigger the CI workflows.

j-stephan (approved):

PR #1579, opened 2022-01-21
add `getNativeHandle()` to alpaka devices

### LINKED ISSUE #1427: add iDevice() to DevUniformCudaHipRt and DevCpu
Would it be possible to add an `iDevice()` method to `DevUniformCudaHipRt` and `DevCpu` ?

Currently
  - `DevOacc` and `DevOmp5` have a private `m_iDevice` member, and a public `iDevice()` method
  - `DevUniformCudaHipRt` has a public `m_iDevice` member
  - `DevCpu` has neither (can be assumed to be always `0`)

For writing generic code, it would be nice if all devices exposed the same interface (directly or via traits / accessors).

tonydp03:
This PR adds a getNativeHandle method to DevUniformCudaHipRt as requested by issue #1427 and will add the same for DevCpu. As for now, Alpaka assumes that there is a single CPU device, which will be set to some number (i.e. 0) and returned by an iDevice public method, but things can change in case we want to distinguish nodes/sockets.

### DIFF
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -120,6 +120,11 @@ namespace alpaka
             m_spDevCpuImpl->registerQueue(spQueue);
         }
 
+        int getNativeHandle()
+        {
+            return 0;
+        }
+
     public:
         std::shared_ptr<cpu::detail::DevCpuImpl> m_spDevCpuImpl;
     };
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -1,4 +1,5 @@
-/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber,
+ * Antonio Di Pilato
  *
  * This file is part of alpaka.
  *
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -188,8 +188,7 @@ namespace alpaka
         template<>
         struct NativeHandle<DevCpu>
         {
-            using type = int;
-            static auto getNativeHandle(DevCpu const& dev) -> type
+            static auto getNativeHandle(DevCpu const& dev)
             {
                 return dev.getNativeHandle();
             }
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -19,6 +19,7 @@
 #include <alpaka/queue/QueueGenericThreadsNonBlocking.hpp>
 #include <alpaka/queue/Traits.hpp>
 #include <alpaka/queue/cpu/IGenericThreadsQueue.hpp>
+#include <alpaka/traits/Traits.hpp>
 #include <alpaka/wait/Traits.hpp>
 
 #include <algorithm>
@@ -121,7 +122,7 @@ namespace alpaka
             m_spDevCpuImpl->registerQueue(spQueue);
         }
 
-        auto getNativeDeviceHandle() const noexcept
+        auto getNativeHandle() const noexcept
         {
             return 0;
         }
@@ -185,9 +186,13 @@ namespace alpaka
 
         //! The CPU device native handle type trait specialization.
         template<>
-        struct NativeHandleDeviceType<DevCpu>
+        struct NativeHandle<DevCpu>
         {
             using type = int;
+            static auto getNativeHandle(DevCpu const& dev)
+            {
+                return dev.getNativeHandle();
+            }
         };
     } // namespace traits
 
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -121,7 +121,7 @@ namespace alpaka
             m_spDevCpuImpl->registerQueue(spQueue);
         }
 
-        int getNativeHandle() const noexcept
+        auto getNativeDeviceHandle() const noexcept
         {
             return 0;
         }
@@ -181,6 +181,13 @@ namespace alpaka
                 ALPAKA_DEBUG_FULL_LOG_SCOPE;
                 // The CPU does nothing on reset.
             }
+        };
+
+        //! The CPU device native handle type trait specialization.
+        template<>
+        struct NativeHandleDeviceType<DevCpu>
+        {
+            using type = int;
         };
     } // namespace traits
 
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -120,7 +120,7 @@ namespace alpaka
             m_spDevCpuImpl->registerQueue(spQueue);
         }
 
-        int getNativeHandle()
+        int getNativeHandle() const noexcept
         {
             return 0;
         }
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -189,7 +189,7 @@ namespace alpaka
         struct NativeHandle<DevCpu>
         {
             using type = int;
-            static auto getNativeHandle(DevCpu const& dev)
+            static auto getNativeHandle(DevCpu const& dev) -> type
             {
                 return dev.getNativeHandle();
             }
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -278,8 +278,7 @@ namespace alpaka
         template<>
         struct NativeHandle<DevOacc>
         {
-            using type = int;
-            static auto getNativeHandle(DevOacc const& dev) -> type
+            static auto getNativeHandle(DevOacc const& dev)
             {
                 return dev.getNativeHandle();
             }
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -279,7 +279,7 @@ namespace alpaka
         struct NativeHandle<DevOacc>
         {
             using type = int;
-            static auto getNativeHandle(DevOacc const& dev)
+            static auto getNativeHandle(DevOacc const& dev) -> type
             {
                 return dev.getNativeHandle();
             }
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -24,6 +24,7 @@
 #    include <alpaka/queue/QueueGenericThreadsNonBlocking.hpp>
 #    include <alpaka/queue/Traits.hpp>
 #    include <alpaka/queue/cpu/IGenericThreadsQueue.hpp>
+#    include <alpaka/traits/Traits.hpp>
 #    include <alpaka/wait/Traits.hpp>
 
 #    include <openacc.h>
@@ -174,15 +175,15 @@ namespace alpaka
     public:
         ALPAKA_FN_HOST auto operator==(DevOacc const& rhs) const -> bool
         {
-            return m_devOaccImpl->getNativeDeviceHandle() == rhs.m_devOaccImpl->getNativeDeviceHandle();
+            return m_devOaccImpl->getNativeHandle() == rhs.m_devOaccImpl->getNativeHandle();
         }
         ALPAKA_FN_HOST auto operator!=(DevOacc const& rhs) const -> bool
         {
             return !((*this) == rhs);
         }
-        auto getNativeDeviceHandle() const noexcept
+        auto getNativeHandle() const noexcept
         {
-            return m_devOaccImpl->getNativeDeviceHandle();
+            return m_devOaccImpl->getNativeHandle();
         }
         acc_device_t deviceType() const
         {
@@ -239,7 +240,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getMemBytes(DevOacc const& dev) -> std::size_t
             {
-                return acc_get_property(dev.getNativeDeviceHandle(), dev.deviceType(), acc_property_memory);
+                return acc_get_property(dev.getNativeHandle(), dev.deviceType(), acc_property_memory);
             }
         };
 
@@ -249,7 +250,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getFreeMemBytes(DevOacc const& dev) -> std::size_t
             {
-                return acc_get_property(dev.getNativeDeviceHandle(), dev.deviceType(), acc_property_free_memory);
+                return acc_get_property(dev.getNativeHandle(), dev.deviceType(), acc_property_free_memory);
             }
         };
 
@@ -273,11 +274,15 @@ namespace alpaka
             }
         };
 
-        //! The OpenACC device native handle type trait specialization.
+        //! The OpenACC device native handle trait specialization.
         template<>
-        struct NativeHandleDeviceType<DevOacc>
+        struct NativeHandle<DevOacc>
         {
             using type = int;
+            static auto getNativeHandle(DevOacc const& dev)
+            {
+                return dev.getNativeHandle();
+            }
         };
     } // namespace traits
 
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -87,7 +87,7 @@ namespace alpaka
                 m_queues.push_back(std::move(spQueue));
             }
 
-            int getNativeHandle() const
+            int getNativeHandle() const noexcept
             {
                 return m_iDevice;
             }
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Benjamin Worpitz, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, Jan Stephan, Bernhard Manfred Gruber, Antonio Di Pilato
  *
  * This file is part of Alpaka.
  *
@@ -180,7 +180,7 @@ namespace alpaka
         {
             return !((*this) == rhs);
         }
-        int getNativeHandle() const
+        int getNativeHandle() const noexcept
         {
             return m_devOaccImpl->getNativeHandle();
         }
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -87,7 +87,7 @@ namespace alpaka
                 m_queues.push_back(std::move(spQueue));
             }
 
-            int iDevice() const
+            int getNativeHandle() const
             {
                 return m_iDevice;
             }
@@ -99,9 +99,9 @@ namespace alpaka
             void makeCurrent() const
             {
 #    if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
-                std::cout << "acc_set_device_num( " << iDevice() << ", [type] )" << std::endl;
+                std::cout << "acc_set_device_num( " << getNativeHandle() << ", [type] )" << std::endl;
 #    endif
-                acc_set_device_num(iDevice(), deviceType());
+                acc_set_device_num(getNativeHandle(), deviceType());
             }
 
             std::uint32_t* gridsLock() const
@@ -174,15 +174,15 @@ namespace alpaka
     public:
         ALPAKA_FN_HOST auto operator==(DevOacc const& rhs) const -> bool
         {
-            return m_devOaccImpl->iDevice() == rhs.m_devOaccImpl->iDevice();
+            return m_devOaccImpl->getNativeHandle() == rhs.m_devOaccImpl->getNativeHandle();
         }
         ALPAKA_FN_HOST auto operator!=(DevOacc const& rhs) const -> bool
         {
             return !((*this) == rhs);
         }
-        int iDevice() const
+        int getNativeHandle() const
         {
-            return m_devOaccImpl->iDevice();
+            return m_devOaccImpl->getNativeHandle();
         }
         acc_device_t deviceType() const
         {
@@ -239,7 +239,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getMemBytes(DevOacc const& dev) -> std::size_t
             {
-                return acc_get_property(dev.iDevice(), dev.deviceType(), acc_property_memory);
+                return acc_get_property(dev.getNativeHandle(), dev.deviceType(), acc_property_memory);
             }
         };
 
@@ -249,7 +249,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getFreeMemBytes(DevOacc const& dev) -> std::size_t
             {
-                return acc_get_property(dev.iDevice(), dev.deviceType(), acc_property_free_memory);
+                return acc_get_property(dev.getNativeHandle(), dev.deviceType(), acc_property_free_memory);
             }
         };
 
--- a/include/alpaka/dev/DevOacc.hpp
+++ b/include/alpaka/dev/DevOacc.hpp
@@ -174,15 +174,15 @@ namespace alpaka
     public:
         ALPAKA_FN_HOST auto operator==(DevOacc const& rhs) const -> bool
         {
-            return m_devOaccImpl->getNativeHandle() == rhs.m_devOaccImpl->getNativeHandle();
+            return m_devOaccImpl->getNativeDeviceHandle() == rhs.m_devOaccImpl->getNativeDeviceHandle();
         }
         ALPAKA_FN_HOST auto operator!=(DevOacc const& rhs) const -> bool
         {
             return !((*this) == rhs);
         }
-        int getNativeHandle() const noexcept
+        auto getNativeDeviceHandle() const noexcept
         {
-            return m_devOaccImpl->getNativeHandle();
+            return m_devOaccImpl->getNativeDeviceHandle();
         }
         acc_device_t deviceType() const
         {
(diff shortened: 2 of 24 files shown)


include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -65,8 +65,16 @@ namespace alpaka
         {
             return !((*this) == rhs);
         }
+        int iDevice() const
+        {
+            return m_iDevice;
+        }
+        int setDevice(int devIdx) const
```
fwyzard: I think `setDevice` should _not_ be `const`.

tonydp03: Correct, updated the PR.

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -65,8 +65,16 @@ namespace alpaka
         {
             return !((*this) == rhs);
         }
+        int iDevice() const
+        {
+            return m_iDevice;
+        }
+        int setDevice(int devIdx)
```
BenjaminW3: I would prefer if we would hold to common practice and have the methods be called `getXxx` and `setXxx` where `Xxx` might be `DeviceIndex`, `DeviceIdx` or `Device`.

tonydp03: We can, but in this case, according to the issue opened by Andrea, I would also rename also the iDevice() method for OpenAcc and OpenMp5 to have the same "interface".

fwyzard:
For what is worth, I would prefer `getNativeId` to `getNativeHandle`, since the values being used are consecutive numerical ids starting at 0 (for the non-host devices).

"handle" gives more the idea of an opaque data type whose value should no be used.

tonydp03:
> 

In today's meeting we agreed to change the name of the function `iDevice` to `getNativeHandle`, but I think `getNativeId` should work as well. Tomorrow I'll also update DevOpenAcc and DevOpenMp5 to have the same name. @j-stephan @jkelling what do you think?

jkelling:
> In today's meeting we agreed to change the name of the function `iDevice` to `getNativeHandle`, but I think `getNativeId` should work as well.

The reason we agreed on `getNativeHandle` for a general interface is, that not all backends have a numerical ID to return (see SYCL), thus will return some other type, and not all of the rest are guaranteed to return a positive number (see OpenMP 5).

While I am ok with the change, as it introduces a fixed naming scheme for a thing that already exists internally in different forms to all backends, I am still on the fence about the use case on the user side. A code that relies on there being a positive integer ID will not be portable because it will not compile with some backends and my crash with others. The use case we discussed in the meeting may be better served by making `Dev` types comparable, so that the user can use these handles as key in an `std::map`. (But this is a discussion for another place.)

fwyzard:
> The reason we agreed on `getNativeHandle` for a general interface is, that not all backends have a numerical ID to return (see SYCL), thus will return some other type, and not all of the rest are guaranteed to return a positive number (see OpenMP 5).

I see, thanks for clarifying the reasons. I considered the OpenMP 5 case at the limit, since IIUC a negative number would only be used for the "host" case. But the use of a "handle" does make sense if SYCL does not have a numerical id at all.

With such architecture, how does the user get hold of a device ?
The only function I know is `alpaka::getDevByIdx(id)`, but that makes sense only with a numerical id.

Do you plan (or is there already) something like `alpaka::enumerateDevices(platform)` that returns a `vector` of all the devices on a given platform ?

> While I am ok with the change, as it introduces a fixed naming scheme for a thing that already exists internally in different forms to all backends, I am still on the fence about the use case on the user side. A code that relies on there being a positive integer ID will not be portable because it will not compile with some backends and my crash with others.

Indeed, I agree that if you foresee backends that do not support a numerical id, an opaque handle makes more sense.
 
> The use case we discussed in the meeting may be better served by making `Dev` types comparable, so that the user can use these handles as key in an `std::map`. (But this is a discussion for another place.)

In our code we "enumerate" all the devices in a given platform and store them in a `vector` - then we can use the position in the vector as a kind of private numerical id.

jkelling:
> In our code we "enumerate" all the devices in a given platform and store them in a `vector` - then we can use the position in the vector as a kind of private numerical id.

This could in principle be done with any type of handle: One could fill that map and then enumerate it. Maybe it is worth considering to move this behavior into alpaka: Define an interface `getNumericalDevID` (that would be a trait for `Dev` types, not a member function) which is defined by the device and it has to take care of mapping whatever the backend offers to a positive integer.

@j-stephan knows about SYCL. He said, that SYCL does indeed return a vector of device at some point and `getDevByIdx()` then takes the element at that index from the vector. It is just, that the `DevSycl` does not know the index afterwards. Connecting the `DevSycl` to that index would incur a small runtime overhead, I guess. Maybe @j-stephan elaborate on that.

tonydp03:
I think that for now we can keep things as they are in this PR (I'm adding one last thing) and add to the "todo list" the interface adapted to the SYCL case (as soon as it will be merged into develop - according to Jan it will happen in the next few weeks/months) and maybe we could also consider how to redesign things in case of multiple-CPU machines, as already discussed before.

j-stephan:
> He said, that SYCL does indeed return a vector of device at some point and getDevByIdx() then takes the element at that index from the vector. It is just, that the DevSycl does not know the index afterwards. Connecting the DevSycl to that index would incur a small runtime overhead, I guess. Maybe @j-stephan elaborate on that.

SYCL has a `sycl::device` class that acts as a handle to the physical device (similar to what an alpaka device does). `sycl::device`s are equality comparable:

```c++
sycl::device a = /* ... */;
sycl::device b = /* ... */;
if(a == b) // This is allowed
{
}
```

`sycl::device` is not able to provide a numerical index, though. It is possible to obtain a `std::vector<sycl::device>` through a static `sycl::device::get_devices()` method and then index into that. This is similar to what I currently do for the SYCL back-end.

The number of devices cannot change during runtime (so we need not to worry about hotpluggable USB GPUs @jkelling ;-)) but I'm not certain that repeated calls to `sycl::device::get_devices()` will always return the vector in the same order.

> according to Jan it will happen in the next few weeks/months

Weeks hopefully. I'm basically just waiting for our other back-end experts to return from vacation & sick leave so we can discuss the implementation details. Afterwards it shouldn't take too long to merge.

include/alpaka/dev/DevCpu.hpp
```diff
@@ -123,6 +123,11 @@ namespace alpaka
             m_spDevCpuImpl->registerQueue(spQueue);
         }
 
+        int getNativeHandle()
```
j-stephan:
```suggestion
        int getNativeHandle() const noexcept
```

include/alpaka/dev/DevOacc.hpp
```diff
@@ -89,7 +89,7 @@ namespace alpaka
                     m_queues.push_back(std::move(spQueue));
                 }
 
-                int iDevice() const
+                int getNativeHandle() const
```
j-stephan:
```suggestion
                int getNativeHandle() const noexcept
```

include/alpaka/dev/DevOacc.hpp
```diff
@@ -177,15 +177,15 @@ namespace alpaka
...
         {
-            return m_devOaccImpl->iDevice() == rhs.m_devOaccImpl->iDevice();
+            return m_devOaccImpl->getNativeHandle() == rhs.m_devOaccImpl->getNativeHandle();
         }
         ALPAKA_FN_HOST auto operator!=(DevOacc const& rhs) const -> bool
         {
             return !((*this) == rhs);
         }
-        int iDevice() const
+        int getNativeHandle() const
```
j-stephan:
```suggestion
        int getNativeHandle() const noexcept
```

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -91,7 +91,7 @@ namespace alpaka
                     m_queues.push_back(spQueue);
                 }
 
-                int iDevice() const
+                int getNativeHandle() const
```
j-stephan:
```suggestion
                int getNativeHandle() const noexcept
```

include/alpaka/dev/DevOmp5.hpp
```diff
@@ -149,15 +149,15 @@ namespace alpaka
...
         {
-            return m_spDevOmp5Impl->iDevice() == rhs.m_spDevOmp5Impl->iDevice();
+            return m_spDevOmp5Impl->getNativeHandle() == rhs.m_spDevOmp5Impl->getNativeHandle();
         }
         ALPAKA_FN_HOST auto operator!=(DevOmp5 const& rhs) const -> bool
         {
             return !((*this) == rhs);
         }
-        int iDevice() const
+        int getNativeHandle() const
```
j-stephan:
```suggestion
        int getNativeHandle() const noexcept
```

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -65,8 +69,12 @@ namespace alpaka
         {
             return !((*this) == rhs);
         }
+        int getNativeHandle() const
```
j-stephan:
```suggestion
        int getNativeHandle() const noexcept
```

j-stephan (changes requested):
Let us provide some extra guarantees to the compiler. Otherwise this looks good.

j-stephan (changes requested):
Okay, functionally this looks good now. Could you rebase onto the current `develop` branch and also update the copyright notice in the first line of each file you touched? Please add your name and update the year to 2022 if necessary.

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -57,6 +57,10 @@ namespace alpaka
         DevUniformCudaHipRt() = default;
 
     public:
+        DevUniformCudaHipRt(int iDevice) : m_iDevice(iDevice)
```
fwyzard: This allows the creation of a `DevUniformCudaHipRt` with an arbitrary device id, even an invalid one.
_cf._ [`PltfUniformCudaHipRt::GetDevByIdx::getDevByIdx()`](https://github.com/alpaka-group/alpaka/blob/2d57d41f076735520d17d077d22469b8fe6842f7/include/alpaka/pltf/PltfUniformCudaHipRt.hpp#L69-L110)

fwyzard:
If going with `getNativeHandle()`, could you add a `typedef` or trait for what the type of the "native handle" should be ?

Either as a member type:
```c++
//! The CUDA/HIP RT device handle.
class DevUniformCudaHipRt : ...
{
public:
    using NativeHandle = int;
...
}
```
or like for other trait-based types in alpaka:
```c++
namespace traits
{

    template<typename TDev>
    struct NativeHandleType;

} // namespace traits

template<typename TDev>
using NativeHandle = NativeHandleType<TDev>::type;

...

namespace traits
{

    template<>
    struct NativeHandleType<DevUniformCudaHipRt> {
        using type = int;
    }

} // namespace traits
```
or more simply (does it work ?):
```c++
template <typename TDev>
using NativeHandle;

...

template <>
using NativeHandle<DevUniformCudaHipRt> = int;
```

j-stephan:
Adding additional traits is the right way. I don't think the the last version with the undefined `using` will work.

include/alpaka/dev/DevUniformCudaHipRt.hpp
tonydp03: This choice was made after discussing with Jan about the fact that it might not be convenient adding a `setNativeHandle` function just for [this](https://github.com/tonydp03/alpaka/blob/0c098883ac12bcfa4e38f3e3a8c0c222d3f35a7b/include/alpaka/pltf/PltfUniformCudaHipRt.hpp#L95) (it's the only part of the code where it would be used, but maybe it would be safer). Any suggestion?

fwyzard: you could make this constructor `private` and `PltfUniformCudaHipRt::GetDevByIdx::getDevByIdx()` a `friend`

include/alpaka/dev/DevCpu.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber, Antonio Di
+ * Pilato
```
j-stephan:
```suggestion
/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber,
 * Antonio Di Pilato
```

include/alpaka/dev/DevOacc.hpp
j-stephan: `noexcept` is still missing here.

include/alpaka/dev/DevOmp5.hpp
j-stephan: `noexcept` is still missing here.

include/alpaka/test/event/EventHostManualTrigger.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2022 Benjamin Worpitz, Matthias Werner, Jan Stephan, Andrea Bocci, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, Matthias Werner, Jan Stephan, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di
+ * Pilato
```
j-stephan:
```suggestion
/* Copyright 2022 Benjamin Worpitz, Matthias Werner, Jan Stephan, Andrea Bocci, Bernhard Manfred Gruber,
 * Antonio Di Pilato
```

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2022 Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di
+ * Pilato
```
j-stephan:
```suggestion
/* Copyright 2022 Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci, Bernhard Manfred Gruber, 
 * Antonio Di Pilato
```

j-stephan (changes requested):

tonydp03:
The required changes have been implemented. Also, as discussed during the meeting, the method `getNativeHandle` was renamed as `getNativeDeviceHandle` as in future there might be different `NativeHandle`s (i.e. for queues, etc.) and we don't have, at the moment, a `Traits.hpp` file where to store common traits for devices and other objects at the same time.

fwyzard:
> in future there might be different `NativeHandle`s (i.e. for queues, etc.)

Sounds like a good idea, just yesterday I had to write some code to extract the `cudaStream_t` from a `QueueUniformCudaHipRtNonBlocking` object.


> the method `getNativeHandle` was renamed as `getNativeDeviceHandle`

This does not seem like a good idea, as IMHO it make the API more cumbersome without being necessary:
  - `DevCpu::getNativeHandle()` can return the device id
  - `QueueUniformCudaHipRtNonBlocking::getNativeHandle()` could return the stream handle,
  - _etc._

The method always knows the type it is being called on, doesn't it ?

fwyzard:
Sorry if I keep insisting on this, but `device.getNativeDeviceHandle()` is simply the wrong choice for an API.
Please change it back to `getNativeHandle().`

tonydp03:
True, this was actually a part of discussion during the meeting on Tuesday and at the end we agreed with @psychocoderHPC to rename the function, but we could actually discuss it here if keeping the generic handle might be a better idea.

j-stephan:
I had a short discussion with @psychocoderHPC about this. Please create the following API:

```c++
// getNativeHandle() should be a free function that calls the corresponding trait underneath
auto handle = alpaka::getNativeHandle(devAcc);
```

Later this can be expanded to also work with queues, buffers, etc. If a back-end does not support a native handle it should trigger a `static_assert`.

tonydp03:
Now `traits/Traits.hpp` has the common traits for future implementation of the NativeHandle method for queues and other alpaka objects.

tonydp03:
Rebased on develop after the namespace changes merged in #1587.

include/alpaka/dev/DevCpu.hpp
```diff
@@ -176,6 +183,17 @@ namespace alpaka
                 // The CPU does nothing on reset.
             }
         };
+
+        //! The CPU device native handle type trait specialization.
+        template<>
+        struct NativeHandle<DevCpu>
+        {
+            using type = int;
```
bernhardmgruber: I don't see where this `type` is used. Can we remove it?

include/alpaka/dev/DevCpu.hpp
```diff
@@ -176,6 +183,17 @@ namespace alpaka
...
+            static auto getNativeHandle(DevCpu const& dev)
```
bernhardmgruber, line 191:
```suggestion
            static auto getNativeHandle(DevCpu const& dev) -> type
```
If you expose the type the native handle should have, you should also make sure your functions commit to it.

tonydp03: Done for all the devices.

include/alpaka/dev/Traits.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2020 Benjamin Worpitz, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, Bernhard Manfred Gruber, Antonio Di Pilato
```
j-stephan: Did you change something in this file?

include/alpaka/dev/DevCpu.hpp
j-stephan: @tonydp03 Does CMS need this?

include/alpaka/mem/buf/BufOmp5.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2021 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Bernhard Manfred Gruber
+/* Copyright 2021 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Bernhard Manfred Gruber, Antonio
```
j-stephan:
```suggestion
/* Copyright 2022 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Bernhard Manfred Gruber, Antonio
```

include/alpaka/mem/buf/BufOacc.hpp
```diff
@@ -1,5 +1,5 @@
 /* Copyright 2021 Jeffrey Kelling, Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Bernhard Manfred
```
j-stephan:
```suggestion
/* Copyright 2022 Jeffrey Kelling, Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Bernhard Manfred
```

include/alpaka/kernel/TaskKernelOmp5.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2021 Benjamin Worpitz, René Widera, Bernhard Manfred Gruber
+/* Copyright 2021 Benjamin Worpitz, René Widera, Bernhard Manfred Gruber, Antonio Di Pilato
```
j-stephan:
```suggestion
/* Copyright 2022 Benjamin Worpitz, René Widera, Bernhard Manfred Gruber, Antonio Di Pilato
```

j-stephan (changes requested):
Almost there...

include/alpaka/dev/Traits.hpp
tonydp03: Well, there was the native handle here but now was removed, so I'll revert back to 2020 and without my name.

include/alpaka/dev/DevCpu.hpp
tonydp03: Well, for now the function just returns 0, so we don't really needed. But in a previous discussion, conveyed that it's useful to have the same interface for all the devices (also we were discussing about handling multi-CPU machines in a different way in future). Could it be useful to keep it for now or we'll modify it at the right time?

j-stephan: I would remove it for now since the user usually doesn't query the trait directly but uses the free function instead. If we decide later that this is required it is not too hard to put in.

fwyzard: > @tonydp03 Does CMS need this?

Yes, to use the common interface to get the "native handle" of a device, we also need a common way to get its type.

j-stephan: Alright, I don't really have a strong opinion on this. @bernhardmgruber, what do you think?

bernhardmgruber: @fwyzard, in case you need the type of the handle: `decltype(alpaka::getNativeHandle(std::declval<Dev>()))`. Where is this not enough?
I also think the use cases to need the handle type are rare, because why would you need the native handle in generic code, when you don't know what backend is in use? And if you are in non-generic code and know the device type, then you also probably know the handle type.

Furthermore, from the design perspective, alpaka has a trait for each piece of information the user wants to query. ` HandleType` would now be the first trait to actually do two things. Get the handle type and get the handle value.

fwyzard: I strongly disagree with the suggestion that `decltype(alpaka::getNativeHandle(std::declval<Dev>()))` is as user friendly as `alpaka::NativeHandle<Dev>` .

fwyzard: > Furthermore, from the design perspective, alpaka has a trait for each piece of information the user wants to query. ` HandleType` would now be the first trait to actually do two things. Get the handle type and get the handle value.

Ehm, no ?
`alpaka::NativeHandle<Dev>` gives the type.
`alpaka::getNativeHandle(dev)` returns the value.

bernhardmgruber: I did not claim my suggestion is user friendly, but I explained why this is not a primary concern.

Then, `typename alpaka::traits::NativeHandle<Dev>::type` would give you the type. But now the user has to grab into the `alpaka::traits` which is our namespace for customization points. This is not what we want the users to do.

If you would really like to have that alias, then how about adding:
```c++
namespace alpaka {
  template <typename Dev>
  using NativeHandle = decltype(getNativeHandle(std::declval<Dev>()));
}
```
and still removing the nested `type` alias in the `alpaka::traits::NativeHandle` specializations?

fwyzard: I do not mind how the `alpaka::NativaHandle<Dev>` type is made available.

From what I can see, everything else in alpaka is implemented with a nested template type or function in `alpaka::traits`, and exposed with an alias in the top level `alpaka` namespace. If this one case is better implemented differently, sure, why not.

include/alpaka/traits/Traits.hpp
```diff
@@ -0,0 +1,38 @@
...
+    } // namespace traits
+
+    //! Get the native handle of the alpaka object.
+    //! It will return the alpaka object handle if there is any, otherwise it generates a compile time error.
+    template<typename TImpl>
+    ALPAKA_FN_HOST auto getNativeHandle(TImpl const& impl)
+    {
+        return traits::NativeHandle<TImpl>::getNativeHandle(impl);
+    }
+} // namespace alpaka
```
bernhardmgruber, line 42: Please add this alias, as requested by @fwyzard.
```suggestion
    template <typename TImpl>
    using NativeHandle = decltype(getNativeHandle(std::declval<TImpl>()));
} // namespace alpaka
```
Please also add a unit test for that alias.

bernhardmgruber (changes requested):

include/alpaka/dev/DevCpu.hpp
bernhardmgruber: @tonydp03 please remove the alias.

tonydp03:
Added the alias as requested and a unit test.

bernhardmgruber (dismissed):

test/unit/traits/src/NativeHandleTest.cpp
```diff
@@ -0,0 +1,26 @@
...
+
+TEMPLATE_LIST_TEST_CASE("NativeHandle", "[handle]", alpaka::test::TestAccs)
+{
+    using Dev = alpaka::Dev<TestType>;
+    using Pltf = alpaka::Pltf<Dev>;
+    Dev const dev(alpaka::getDevByIdx<Pltf>(0u));
+    auto handle = alpaka::getNativeHandle(dev);
+
+    STATIC_REQUIRE(std::is_same_v<alpaka::NativeHandle<Dev>, decltype(handle)>);
+    STATIC_REQUIRE(std::is_same_v<alpaka::NativeHandle<Dev>, int>); // It won't work for SYCL backend
```
j-stephan, line 24: What is this for?

tonydp03: Testing the consistency of the alias Native Handle (same type of the handle returned by the getNativeHandle function) and that the NativeHandle of a device is an integer (but this will be changed with Sycl backend)

bernhardmgruber: First line tests `getNativeHandle` and third line tests consistency of the returned type with the trait.

The last line is to check that the native type is `int` which is true for all our backends at the moment. We will need to adapt this when the SYCL branch comes in, since it probably has a different native handle. I was not sure whether I want this last test initially, but this way we can enforce that we do not break the native handle type accidentally (e.g. making it suddently unsigned somewhere).

j-stephan (approved):

bernhardmgruber (approved):

PR #1584, opened 2022-01-31
Run clang-tidy modernize checks

bernhardmgruber:
This PR just blindly runs `run-clang-tidy-12.py -checks="-*,modernize-*" -header-filter=.* -fix`. Manually revered changes in tinymt32.h and STLTuple.h. Manually revert some refactorings resulting in compile errors.

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -37,11 +37,9 @@ namespace alpaka
             class ThreadSafeQueue : private std::queue<T>
             {
             public:
-                ThreadSafeQueue()
-                {
-                }
+                ThreadSafeQueue() = default;
                 //! \return If the queue is empty.
-                auto empty() const -> bool
+                [[nodiscard]] auto empty() const -> bool
                 {
                     return std::queue<T>::empty();
                 }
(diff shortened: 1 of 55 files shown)


include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -174,7 +172,7 @@ namespace alpaka
 
             private:
                 //! The execution function.
-                virtual auto run() -> void final
```
j-stephan, line 177: Why is the `virtual` removed here?

j-stephan (dismissed):

include/alpaka/core/ConcurrentExecPool.hpp
fwyzard: both `override` and `final` imply `virtual`, so `clang-tidy` removes it

j-stephan: Ah, I see. Thanks!

j-stephan:
Some clang versions don't like these changes:

https://gitlab.com/hzdr/crp/alpaka/-/pipelines/460211464

bernhardmgruber:
> Some clang versions don't like these changes:
> 
> https://gitlab.com/hzdr/crp/alpaka/-/pipelines/460211464

I already fixed these today, but the GitLab CI is not restarting.

j-stephan:
Hm, it could be that you need to force-push again. I think because of today's GitLab issues the pipeline wasn't notified about anything you committed after yesterday.

j-stephan (approved):

PR #1586, opened 2022-02-02
Update BoostPredef.hpp

### LINKED ISSUE #611: BOOST_LANG_CUDA relies on cmake's __CUDACC__ workaround
Currently, [we have](https://github.com/ComputationalRadiationPhysics/alpaka/pull/606):
```C
#if defined(__CUDA__) || defined(__CUDACC__)
      #include <cuda.h>
      #define BOOST_LANG_CUDA BOOST_PREDEF_MAKE_10_VVRRP(CUDA_VERSION)
  #else
      #define BOOST_LANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
  #endif
```
As example we take a common check:
```C
#if !BOOST_LANG_CUDA
    #error If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!
#endif
```
One consequence is, that this would fail, if cmake would not have a workaround for `__CUDACC__`.
When I use `make VERBOSE=1 rand` and take the commands without that workaround, I get an error (I actually had that issue with HIP(nvcc), where there is no workaround by cmake, and `__HIPCC__` which is not supposed to be defined by compiler):
> /alpaka/include/alpaka/pltf/PltfCudaRt.hpp:29:6: error: #error If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!
     #error If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!

cmake adds a definition of `__CUDACC__` and `NVCC` to nvcc (=> `nvcc -M -D__CUDACC__ ... -DNVCC ...`), so the host compiler called by nvcc also sees `__CUDACC__`. 
CMake source: See [cmake's run_nvcc.cmake here](https://github.com/Kitware/CMake/blob/master/Modules/FindCUDA/run_nvcc.cmake#L192) and [here](https://github.com/Kitware/CMake/blob/master/Modules/FindCUDA/run_nvcc.cmake#L206):

```cmake
# nvcc doesn't define __CUDACC__ for some reason when generating dependency files.  This
# can cause incorrect dependencies when #including files based on this macro which is
# defined in the generating passes of nvcc invocation.  We will go ahead and manually
# define this for now until a future version fixes this bug.
set(CUDACC_DEFINE -D__CUDACC__)
```

We probably could place an own define (even `BOOST_LANG_CUDA` or `BOOST_LANG_HIP`?) into the `NVCC_FLAGS` and `HIPCC_FLAGS`, to detect that the host compiler has been called by nvcc or hipcc. Or maybe there is a more portable solution, and I don't see it at the moment.

j-stephan:
Since we have updated to Boost 1.74 recently we can remove some of our handmade solutions for earlier Boost versions. ~~I also simplified some of the preprocessor checks in a few places.~~ Nevermind, the simplifications just broke things.

Fixes #611.

### DIFF
--- a/include/alpaka/core/BoostPredef.hpp
+++ b/include/alpaka/core/BoostPredef.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2019 Benjamin Worpitz, Matthias Werner
+/* Copyright 2022 Benjamin Worpitz, Matthias Werner, Jan Stephan
  *
  * This file is part of alpaka.
  *
@@ -10,12 +10,6 @@
 #pragma once
 
 #include <boost/predef.h>
-
-// In boost since 1.68.0
-// BOOST_PREDEF_MAKE_10_VVRRP(V)
-#if !defined(BOOST_PREDEF_MAKE_10_VVRRP)
-#    define BOOST_PREDEF_MAKE_10_VVRRP(V) BOOST_VERSION_NUMBER(((V) / 1000) % 100, ((V) / 10) % 100, (V) % 10)
-#endif
 
 //---------------------------------------HIP-----------------------------------
 // __HIPCC__ is defined by hipcc (if either __CUDACC__ is defined)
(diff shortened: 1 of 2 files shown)


include/alpaka/core/Common.hpp
```diff
@@ -90,8 +90,7 @@
 //!     alpaka::memcpy(queue, viewGlobalMemUninitialized, bufHost, extent);
 //! }
 //! \endcode
-#if((BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX)              \
-    || BOOST_LANG_HIP)
+#if((BOOST_COMP_CLANG_CUDA) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX) || BOOST_LANG_HIP)
```
j-stephan: I wonder if this can be simplified to

```suggestion
#if(BOOST_LANG_CUDA || BOOST_LANG_HIP)
```

@psychocoderHPC Do you know?

fwyzard:
Just a suggestion, you could add a check on the boost version, to make it easy to catch an old boost version when building without CMake.

j-stephan:
Mh, I'm not a big fan of handling dependencies inside the code itself. This is something that should be handled by the build system.

fwyzard:
> Mh, I'm not a big fan of handling dependencies inside the code itself. This is something that should be handled by the build system.

Aren't all the
```c++
#        if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && !BOOST_LANG_CUDA
#            error If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!
#        endif
```
checks doing just that ?

j-stephan:
To be honest I'm not certain those are still required. IIRC those come from the time when we didn't have native CMake support for CUDA. Now CMake can check if the compiler actually supports the CUDA language so this is kind of moot.

fwyzard:
So far I understood that building with CMake is what you actively support, but building without CMake is supposed to work as well.

If building with CMake becomes a _requirement_, even just to get meaningful errors or a working compilation, then we won't be able to use alpaka any more.

j-stephan:
> So far I understood that building with CMake is what you actively support, but building without CMake is supposed to work as well.

True, and we have no plans to change that.

What I am trying to convey here is that I am against putting *more* dependency management into the source code than already present. Library versions are something that should be handled by the build system, not source code. I don't know the details of your own build system, but shouldn't it be possible to add a version check there?

fwyzard:
> I don't know the details of your own build system, but shouldn't it be possible to add a version check there?

We don't have checks on individual library versions - we build and ship basically all the libraries that we depend on (minus system ones like glibc or libssl).

I guess I'm simply looking at this from a different point of view.

When a function has some preconditions, it's good policy to check for them with `assert`, `static_assert`, _etc._ (even in production code, or at least in "debug mode" if the runtime cost is otherwise too high).
Then it's the responsibility of the caller to make sure those preconditions are met.

Similarly, as different parts of alpaka have different requirements on external libraries (CUDA > xxx, boost > yyy, etc.) it would be nice if those parts had a compile-time check that what they rely upon is available.
Then it's the responsibility of the build system to make sure that those requirements are met.

fwyzard:
Anyway - of course _not_ having the check won't break any working setup.

j-stephan:
I'll put this on the agenda for next week's meeting @fwyzard.

include/alpaka/core/Common.hpp
j-stephan: It cannot.

fwyzard:
> I'll put this on the agenda for next week's meeting @fwyzard.

I'm on vacation next week... but you don't need me to discuss, I've written here my reasoning :-)

bernhardmgruber (dismissed):

bernhardmgruber (dismissed):

bernhardmgruber (approved):

j-stephan:
@fwyzard We have agreed to introduce a new header that checks for requirements, including a Boost version check. This will be part of a separate PR.

bernhardmgruber (approved):

PR #1591, opened 2022-02-15
Refactor alpaka/rand

### LINKED ISSUE #1496: `delete` `rand::distribution::gpu::NormalReal`'s move constructor
In `rand/RandDefault.hpp` we find the following move constructor for the `NormalReal` class:

```c++
//! The move ctor clears `m_cache` of source.
//! \todo This is to be deleted when moving to C++17
NormalReal(NormalReal&& o) : m_acc(o.m_acc), m_cache(o.m_cache)
{
    o.m_cache = std::numeric_limits<T>::quiet_NaN();
}
```

Time to delete this. @jkelling @sliwowitz Any volunteers?

bernhardmgruber:
Refactors the code in alpaka/rand:

* collapse nested namespaces
* make some state members private

Then, it makes `alpaka::rand::distribution::gpu::NormReal` copyable, since other distributions are also copyable. This makes #1496  obsolete.


sliwowitz (approved):
Looks good. I'd say your solution of using friends to handle the CUDA distributions is definitely better than keeping the state public.

j-stephan (approved):

PR #1593, opened 2022-02-16
Remove namespace alpaka::extent

bernhardmgruber:
This PR removes the subnamespace `alpaka::extent` and inlines its content into namespace `alpaka`. See also: #1034.

- [x] Please merge first #1592


j-stephan:
Is there a reason for not removing these in the original namespace removal PR?

bernhardmgruber:
Well, we had a PR for almost each removed namespace. I think I stopped back then because I was not sure whether it could create confusion to have `getWidth` inside `alpaka` instead of `alpaka::extent`. But I gave it some more thought now and I think that it should be fine since it applies to many objects across alpaka like buffers, views, `Vec`, etc.

j-stephan (approved):

Issue #1597, opened 2022-02-17
[RFC] Remove pitched allocations


bernhardmgruber:
Some backends in alpaka insert padding after the rows for a 2D or higher dimensional buffer (notable CUDA/HIP with cudaMallocPitch and cudaMalloc3D). This additional padding is hardware dependent and selected by the implementation with the hope to provide faster memory access to the buffer inside a kernel. That concerns coalescing, the memory controller, number of memory transactions etc.

However, this additional padding, often also called row pitch, or just pitch, is a frequent source of bugs, even in alpaka. See e.g. #1549 or #1240. Furthermore, it is also a blocker for new abstractions like `std::mdspan`, see: #1482.

So I browsed the internet, trying to find what the actual performance gain of a pitched allocation is compared to a non-pitched one. Here are some results:
* [This presentation](http://www.quest-itn.org/events/1st-quest-workshop/1st-quest-workshop/programme/presentations/tuesday/GPU_tutorial.pdf) claims that a pitched allocations can give 75% (Tesla C1060) or 30-40% (GTX 480) performance improvements. Note that the listed GPUs are from 2009 and 2010.
* [This NVIDIA forum answer from 2018](https://forums.developer.nvidia.com/t/what-is-the-stream-ordered-equivalent-of-cudamallocpitch/189574/2) states that "pitched allocations should have been noticeably important in cc1.x GPUs. Those GPUs died out circa 2016.". The question was why there is no pitched allocation API for managed memory.
* [This Stackoverflow comment](https://stackoverflow.com/questions/29694627/cudamallocpitch-for-managed-memory#comment47699641_29694627) states that "GPU caches, in place since cc 2.0 devices appeared, have to some extent mitigated the performance impact of not using pitched allocations on the GPU."
* [This NVIDIA forum answer from 2017](https://forums.developer.nvidia.com/t/bad-performance-using-mallocpitch-and-memcpy2d/50203) states that "the pitched method may not give any better performance than the unpitched method. Pitched allocations were especially useful on early GPUs, but are of less significance on modern GPUs. Depending on your GPU, its possible that the overhead associated with pitch calculations in the kernel (especially for such a simple kernel) may outweigh any benefit from pitched access."
* [This example on Stackoverflow](https://stackoverflow.com/questions/30903380/cudamalloc-vs-cudamalloc3d-performance-for-a-2d-array) shows that cudaMalloc3D is 10% slower than cudaMalloc because of additional instruction overhead in the kernel.

[...]

PR #1599, opened 2022-02-18
Measure copy in matMul

bernhardmgruber:
This PR adds a new function `alpaka::test::integ::measureRunTimeMs` and then uses this function to measure and print the copy times in the matMul integration test.

### DIFF
--- a/test/integ/matMul/src/matMul.cpp
+++ b/test/integ/matMul/src/matMul.cpp
@@ -227,10 +227,21 @@ TEMPLATE_LIST_TEST_CASE("matMul", "[matMul]", TestAccs)
     auto bufBAcc = alpaka::allocBuf<Val, Idx>(devAcc, extentB);
     auto bufCAcc = alpaka::allocBuf<Val, Idx>(devAcc, extentC);
 
-    // Copy Host -> Acc.
-    alpaka::memcpy(queueAcc, bufAAcc, bufAHost);
-    alpaka::memcpy(queueAcc, bufBAcc, bufBHost);
-    alpaka::wait(queueHost);
+    // Copy inputs Host -> Acc.
+    std::cout << "Input1 copy time: "
+              << alpaka::test::integ::measureRunTimeMs(
+                     [&]
+                     {
+                         alpaka::memcpy(queueAcc, bufAAcc, bufAHost);
+                         alpaka::memcpy(queueAcc, bufBAcc, bufBHost);
+                         alpaka::wait(queueAcc);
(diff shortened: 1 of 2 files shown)


test/integ/matMul/src/matMul.cpp
```diff
@@ -228,10 +228,16 @@ TEMPLATE_LIST_TEST_CASE("matMul", "[matMul]", TestAccs)
...
-    alpaka::wait(queueHost);
-    alpaka::memcpy(queueAcc, bufCAcc, bufCHost);
+    std::cout << "Input copy time:  "
+              << alpaka::test::integ::measureRunTimeMs(
+                     [&]
+                     {
+                         alpaka::memcpy(queueAcc, bufAAcc, bufAHost);
+                         alpaka::memcpy(queueAcc, bufBAcc, bufBHost);
+                         alpaka::memcpy(queueAcc, bufCAcc, bufCHost);
+                         alpaka::wait(queueHost);
```
bernhardmgruber: I moved the wait to the last command to have the total copy time of all 3 buffers.

bernhardmgruber: Ok, that was a bug :D The `wait` ensures that the `memset` on `bufCHost` finished before copying from it.

bernhardmgruber:
Given the jittering of the copy measurement (see #1597), I wonder if there are better ways in alpaka for measuring the time of an `alpaka::memcpy`?

bussmann:
@j-stephan should have a library for this...

bernhardmgruber:
> @j-stephan should have a library for this...

I think a fully fledged instrumentation library might be an overkill for taking a measurement of three buffer transfers in an integration test :S But I would love to hear about how @j-stephan would measure the transfer time.

bussmann:
That was my point 😉

j-stephan:
This is basically a trace of single alpaka call. Using bactria for this use case is certainly possible but introduces some programming overhead since it isn't really intended for something this simple.

In the future (I'm aiming for alpaka 0.10 here) we will hopefully have bactria integrated into alpaka which will enable us to automatically gather performance data for each alpaka call. We'd need to discuss on how to integrate this into the CI, though. Do we want to print everything to `stdout`? Do we want to save a `*.csv` file somewhere? Do we want to write into a database? Do we want to do some weird blockchain stuff like @psychocoderHPC once suggested? And someone would need to write a bactria output plugin for this.

j-stephan (approved):

PR #1608, opened 2022-02-21
Refactor BufCpu and BufUniformCudaHipRt

bernhardmgruber:
This PR removes printing the pitch of `BufCpu`, which is always unpitched. It also simplifies the implementation of `BufUniformCudaHipRt`.

This refactoring falls out of a branch where I test alpaka without any pitched allocations in the CUDA/HIP backend.

### DIFF
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -56,7 +56,6 @@ namespace alpaka
                 , m_extentElements(getExtentVecEnd<TDim>(extent))
                 , m_pMem(pMem)
                 , m_deleter(std::move(deleter))
-                , m_pitchBytes(static_cast<TIdx>(getWidth(extent) * static_cast<TIdx>(sizeof(TElem))))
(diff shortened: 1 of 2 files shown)


fwyzard:
Just an opinion: if the CUDA/HIP buffers keep the pitch, I would prefer to
  - have the possibility of querying it
  - have a functional pitch also for the CPU buffers, so that one can use the same pitch for CPU and CUDA/HIP, making the copies a bit faster

Unless the N-dimensional allocations and the pitch just go away, of course.

bussmann:
One of the virtues of LLAMA is to have memory layout aware copy, so you just copy without having to manually care for the pitch.

fwyzard:
> One of the virtues of LLAMA is to have memory layout aware copy, so you just copy without having to manually care for the pitch.

`cudaMemcpy2D[Async]` and `cudaMemcpy3D[Async]` also take care of the pitch, but it doesn't mean that it's free in terms of performance.
Though, if anyone has any evidence that `cudaMemcpy2D` does not incur any performance drop with respect to `cudaMemcpy`, it would be nice to hear :-)

j-stephan (approved):

PR #1609, opened 2022-02-21
Documentation for writing unit tests

### LINKED ISSUE #1513: Improve build system documentation
We have received some questions regarding the usage of the CMake-based build system:

* How are back-ends enabled?
* How are test cases enabled?

We should document this better.

tonydp03:
This PR adds a draft documentation for writing unit tests commenting a simple example line by line and describing a sample workflow. It can be improved by adding comments to more complex examples.

### DIFF
--- a/docs/source/dev/test.rst
+++ b/docs/source/dev/test.rst
@@ -1,0 +1,92 @@
+.. highlight:: cpp
+
+Writing a unit test
+===================
+
+After implementing a new functionality in Alpaka, it is recommended to test it. Indeed, having no compile time errors
+when building alpaka with any backend doesn't mean that such functionality is well implemented and behaves as expected.
+
+Unit tests are written and integrated with `Catch2 <https://github.com/catchorg/Catch2>`_ and they are standalone executables located in the ``test/unit`` and
+``test/integ`` paths inside the ``build`` folder. They can be built all at once by setting the ``CMake``
+argument ``-DBUILD_TESTING=ON`` and running the command ``cmake --build .`` in the ``build`` folder, and then run with the
+``ctest`` command as described in the :doc:`Installation section </basic/install>`.
+
+In alternative, a target test can be built by adding the flag ``--target <testname>`` to the ``cmake --build`` command and
+then executed as a normal binary file (``./<testname>``) in the corresponding folder mentioned above.
+
+Sample workflow
+---------------
+
+As a sample workflow, consider the ``getWarpSize`` functionality that, given an ``alpaka device``, returns the number
+of threads that constitute a warp.
+
+First of all, the unit test should be located inside a folder under the ``test/unit`` path (namely ``dev``). Such a folder
+will contain a ``CMakeLists.txt`` file and a ``src`` folder, in which the ``.cpp`` file with the test itself will be located.
+Furthermore, it is necessary to add the folder in which the test is located to ``test/unit/CMakeLists.txt``:
+
+.. code-block::
+    add_subdirectory("dev/")
+
+to make sure that ``CMake`` will build the test written in the ``.cpp`` file under that directory.
+
+After these steps, it all comes to writing the *actual* test. First of all, the necessary header libraries should be
+included the ``.cpp`` file:
+
+.. code-block:: c++
+    #include <alpaka/dev/Traits.hpp>
+    #include <alpaka/test/acc/TestAccs.hpp>
+    #include <catch2/catch.hpp>
+
+The first ``include`` statement targets specifically the header in which the functionality is included; with the second one
+only accelerators corresponding to enabled backends are selected for testing, while the third one is necessary to integrate
+Catch2.
+
+Test cases using an alpaka accelerator are then typically introduced with the ``TEMPLATE_LIST_TEST_CASE`` macro (as described
+`here <https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#type-parametrised-test-cases>`_).
+It takes three arguments:
+* a free form test name (must be unique)
+* a tag
+* the accelerator(s) that must run the test case(s) (i.e. ``alpaka::test::TestAccs`` targets all the accelereators selected
+by the ``TestAccs`` header).
+
+Some aliases might be useful:
+
+.. code-block:: c++
+    using Dev = alpaka::Dev<TestType>;
+    using Pltf = alpaka::Pltf<Dev>;
+
+The template parameter for ``alpaka::Dev`` is ``TestType``, which basically target the specific accelerator that corresponds
+to the backend which is executing the test.
+
+Since the function that must be tested has a device object as argument, it is necessary to define one:
+
+.. code-block:: c++
+    Dev const dev(alpaka::getDevByIdx<Pltf>(0u));
+
+Having all the necessary objects well defined, the function can be called and the result stored in an object:
+
+.. code-block:: c++
+    auto const warpExtent = alpaka::getWarpSize(dev);
+
+Finally, the condition under which the test succeeds should be defined with a Catch2 macro.
+The most used macros are: ``REQUIRE`` and ``CHECK`` family.
+From the `Catch2 documentation <https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md>`_:
+
+*The REQUIRE family of macros tests an expression and aborts the test case if it fails. The CHECK family are
+equivalent but execution continues in the same test case even if the assertion fails.*
+
+For example:
+
+.. code-block:: c++
+    REQUIRE(warpExtent > 0);
+
+evaluates the expression between the round brackets and if an exception is thrown, it is caught, reported, and counted
+as a failure. Such expression depends on what must be actually tested (i.e. the size of the warp can never be equal or
+less than 0).
+Furthermore, additional, compile-time expressions can be evaluated as well:
+
+.. code-block:: c++
+    STATIC_REQUIRE(std::is_same_v<decltype(warpExtent), int>);
+
+which checks that the type of the warp size is integer (cannot be float). Differently from the ``REQUIRE`` macro,
+if the expression is false, the ``STATIC_REQUIRE`` will throw an error at compile time, without running the application. 
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -65,6 +65,7 @@ Individual chapters are based on the information of the chapters before.
    dev/backends.rst
    dev/details.rst
    dev/style
+   dev/test.srt
    dev/sphinx
    dev/ci
    API Reference <https://alpaka-group.github.io/alpaka>


docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+Writing a unit test
+==================
+
+After implementing a new functionality in Alpaka, it is recommended to test it. Indeed, having no compile time errors
+when building alpaka with any backend doesn't mean that such functionality is well implemented and behaves as expected.
+
+Unit tests are written and integrated with `Catch2`_ and they are standalone executables located in the ``test/unit`` and
+``test/integ`` paths inside the ``build`` folder. They can be built all at once by setting the ``CMake``
+argument ``-DBUILD_TESTING=ON`` and running the command ``cmake --build .``, and then run with the ``ctest`` command as
+described in the :doc:`Installation section </basic/install>`.
```
bernhardmgruber: Consider pointing out that the commands are supposed to be run in the build folder.

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+
+.. _Catch2: https://github.com/catchorg/Catch2
```
bernhardmgruber: Maybe just put the link behind the first mentioning of Catch2 in the preceding paragraph.

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+    alpaka_add_executable(
+        ${_TARGET_NAME}
+        ${_FILES_SOURCE})
+    target_link_libraries(
+        ${_TARGET_NAME}
+        PRIVATE common)
+
+    set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER "test/unit")
+
+    add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} ${_ALPAKA_TEST_OPTIONS})
```
bernhardmgruber: I am hesitant to show the content of a `CMakeLists.txt` because it does not add a lot of value and might change at some point. And users are free to implement their `CMakeLists.txt` differently.

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+
+where the first line sets the name of the test that ``CMake`` will build for execution. Furthermore, it is necessary to add
+the folder in which the test is located to the ``CMakeLists.txt`` file under the ``test/unit`` path:
```
bernhardmgruber:
```suggestion
the folder in which the test is located to ``test/unit/CMakeLists.txt``:
```

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+.. code-block:: c++
+    #include <alpaka/dev/Traits.hpp>
+    #include <alpaka/test/acc/TestAccs.hpp>
+    #include <catch2/catch.hpp>
+
+The first ``include`` statement targets specifically the header in which the functionality is included; with the second one
+only accelerators corresponding to enabled backends are selected for testing, while the third one is necessary to integrate
+Catch2.
+
+One or more test cases are then introduced with the ``TEMPLATE_LIST_TEST_CASE`` macro. It takes three arguments:
```
bernhardmgruber: There might be other tests where you would want to use a different test declaration. Maybe rephrase:
```suggestion
Test cases using an alpaka accelerator are then typically introduced with the ``TEMPLATE_LIST_TEST_CASE`` macro. It takes three arguments:
```

bernhardmgruber: Maybe also point to the Catch2 documentation here: https://github.com/catchorg/Catch2/blob/devel/docs/test-cases-and-sections.md#type-parametrised-test-cases

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Having all the necessary objects well defined, the function can be called and the result stored in an object:
+
+.. code-block:: c++
+    auto const warpExtent = alpaka::getWarpSize(dev);
+
+Finally, the condition under which the test succeed should be defined with the ``REQUIRE`` macro:
+
+.. code-block:: c++
+    REQUIRE(warpExtent > 0);
```
bernhardmgruber: There is also the `CHECK` macro, which I usually prefer because it does not abort the test when the check fails. Maybe just point to the Catch2 documentation here on how to do checks. There is a lot more: https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+
+Such condition depends on what must be actually tested (i.e. the size of the warp can never be equal or less than 0).
+Furthermore, other additional conditions can be set as well:
+
+.. code-block:: c++
+    REQUIRE(std::is_same_v<decltype(warpExtent), int>);
```
bernhardmgruber: IMO that should be a `STATIC_REQUIRE`.

docs/source/dev/test.rst
```diff
@@ -0,0 +1,107 @@
...
+
+which checks that the type of the warp size is integer (cannot be float).
+
+.. hint::
+    An alternative to the ``REQUIRE`` macro is the ``STATIC_REQUIRE`` macro. The difference between them is that if the
+    test fails, in the first case the failure will be notified at run time, while in the second case it will result in
+    a compile time error.
```
bernhardmgruber: I would omit this and point to the Catch2 documentation. There is so much more, like checking for exceptions, handling floating points etc.

bernhardmgruber (commented):
Thank you for adding this! Still, I think the alpaka docs are not the right place to explain too much about Catch2. They have their own (great) documentation. We should just explain how alpaka uses Catch2 and this is mostly about the `TEMPLATE_LIST_TEST_CASE` and `alpaka::test::TestAccs`, which you already did. I think you can cut a few bits and add a few pointers to the Catch2 documentation. Still, great work and thank you for doing this!

tonydp03:
Thanks for the feedbacks @bernhardmgruber , they are extremely useful :D I'll do some changes asap; I used a very simple example as there are many more complicated ones - but I guess that any external contributor would need help to write anything similar. Also pointing to several parts of the Catch2 documentation might be a very good idea.

docs/source/dev/test.rst
tonydp03: About this, I think that it might be useful to discuss it on VC with the others as well, since all the tests that have been written have e REQUIRE rather than STATIC_REQUIRE. Maybe it's worth to change it? (STATIC_REQUIRE is actually useful imho)

docs/source/dev/test.rst
```diff
@@ -0,0 +1,85 @@
...
+
+Writing a unit test
+==================
+
+After implementing a new functionality in Alpaka, it is recommended to test it. Indeed, having no compile time errors
+when building alpaka with any backend doesn't mean that such functionality is well implemented and behaves as expected.
+
+.. _Catch2: https://github.com/catchorg/Catch2
+
+Unit tests are written and integrated with `Catch2`_ and they are standalone executables located in the ``test/unit`` and
```
bernhardmgruber:
```suggestion
Unit tests are written and integrated with `Catch2 <https://github.com/catchorg/Catch2>`_ and they are standalone executables located in the ``test/unit`` and
```

docs/source/dev/test.rst
```diff
@@ -0,0 +1,85 @@
...
+``test/integ`` paths inside the ``build`` folder. They can be built all at once by setting the ``CMake``
+argument ``-DBUILD_TESTING=ON`` and running the command ``cmake --build .`` in the ``build`` folder, and then run with the
+``ctest`` command as described in the :doc:`Installation section </basic/install>`.
+
+In alternative, a target test can be built by adding the flag ``--target <testname>`` to the ``cmake --build`` command and
+then executed as a normal binary file (``./<testname>``) in the corresponding folder mentioned above.
+
+Sample workflow
+-------
```
bernhardmgruber:
```suggestion
Sample workflow
-------------------
```

bernhardmgruber (changes requested):
Looks almost good. Please also squash all your commits together into one.

docs/source/dev/test.rst
bernhardmgruber: Since you have an example of `REQUIRE` (runtime test) a few lines above, I would really like to see `STATIC_REQUIRE` here as an example for a compile time test.

fwyzard: For someone like me that does not have prior experience with Catch2, it would be useful to have a very quick introduction to the most common checks to use (`CHECK`, `REQUIRE`, `STATIC_REQUIRE`, _etc._)

docs/source/dev/test.rst
```diff
@@ -0,0 +1,92 @@
+.. highlight:: cpp
+
+Writing a unit test
+==================
```
bernhardmgruber:
```suggestion
Writing a unit test
===================
```

docs/source/dev/test.rst
```diff
@@ -0,0 +1,92 @@
...
+Unit tests are written and integrated with `Catch2 <https://github.com/catchorg/Catch2>`_ and they are standalone executables located in the ``test/unit`` and
+``test/integ`` paths inside the ``build`` folder. They can be built all at once by setting the ``CMake``
+argument ``-DBUILD_TESTING=ON`` and running the command ``cmake --build .`` in the ``build`` folder, and then run with the
+``ctest`` command as described in the :doc:`Installation section </basic/install>`.
+
+In alternative, a target test can be built by adding the flag ``--target <testname>`` to the ``cmake --build`` command and
+then executed as a normal binary file (``./<testname>``) in the corresponding folder mentioned above.
+
+Sample workflow
+-------------------
```
bernhardmgruber:
```suggestion
Sample workflow
---------------
```
Ok, maybe I messed that up previously. The goal here is to make the line as long as the heading's text above it.

docs/source/dev/test.rst
```diff
@@ -0,0 +1,92 @@
...
+
+For example:
+
+.. code-block:: c++
+    REQUIRE(warpExtent > 0);
+
+evaluates the expression between the round brackets and if an exception is thrown, it is caught, reported, and counted
+as a failure. Such expression depends on what must be actually tested (i.e. the size of the warp can never be equal or
+less than 0).
+Furthermore, other additional expressions can be evaluated as well:
```
bernhardmgruber:
```suggestion
Furthermore, additional, compile-time expressions can be evaluated as well:
```
I kind of prefer stating what the macro primarily does before showing it.

bernhardmgruber (commented):
Final nits and we are good!

docs/source/dev/test.rst
bernhardmgruber: Sorry about that one!

bernhardmgruber (approved):

PR #1610, opened 2022-02-22
Add a deduction guide for Vec

bernhardmgruber:
This PR adds a deduction guide to `alpaka::Vec` to make its construction less verbose.

- [x] Please merge #1620 first.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -98,6 +98,20 @@ namespace alpaka
         //! Value constructor.
         //! This constructor is only available if the number of parameters matches the vector idx.
         ALPAKA_NO_HOST_ACC_WARNING
+#if BOOST_COMP_NVCC && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(11, 3, 0)                                              \
+    && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 4, 0)
+        // This constructor tries to avoid SFINAE, which crashes nvcc 11.3. We also need to have a first
+        // argument, so an unconstrained ctor with forwarding references does not hijack the compiler provided
+        // copy-ctor.
+        template<typename... TArgs>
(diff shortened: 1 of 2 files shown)


j-stephan:
What is the state here?

bernhardmgruber:
> What is the state here?

The state is:
`2022-03-10T15:28:23.0101765Z D:\a\alpaka\alpaka\include\alpaka/vec/Vec.hpp(102): internal error : assertion failed at: "cp_gen_be.c", line 4745 in gen_class_qualifier [D:\a\alpaka\alpaka\build\test\unit\vec\vecTest.vcxproj]` on nvcc 11.3 (Linux and Windows). So there is nothing I can do here for now.

j-stephan:
> So there is nothing I can do here for now.

Okay. What do you plan to do in the mid- to long term? It might take a while until we can phase out CUDA 11.3.

bernhardmgruber:
> Okay. What do you plan to do in the mid- to long term? It might take a while until we can phase out CUDA 11.3.

Well, maybe a minor update fixes it at some point. Otherwise, i would just let the PR wait here.

include/alpaka/vec/Vec.hpp
```diff
@@ -98,13 +98,28 @@ namespace alpaka
         //! Value constructor.
         //! This constructor is only available if the number of parameters matches the vector idx.
         ALPAKA_NO_HOST_ACC_WARNING
+#if BOOST_COMP_NVCC && BOOST_COMP_NVCC >= BOOST_VERSION_NUMBER(11, 3, 0)                                              \
+    && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 4, 0)
+        // This constructor tries needs to avoid SFINAE, which crashes nvcc 11.3. We also need to have a first
```
bernhardmgruber:
```suggestion
        // This constructor tries to avoid SFINAE, which crashes nvcc 11.3. We also need to have a first
```

j-stephan (approved):

PR #1611, opened 2022-02-22
Unit test move ctor/assign of Buf

bernhardmgruber:
This PR adds a unit test for the move ctor/assign operator of buffers. It additionally deletes move operations on buffer implementations, since those are not necessary and may even cause bugs. It also renames a few template and function parameter names to be more meaningful.

- [x] Merge #1614 first

### DIFF
--- a/test/unit/mem/buf/src/BufTest.cpp
+++ b/test/unit/mem/buf/src/BufTest.cpp
@@ -264,3 +264,49 @@ TEMPLATE_LIST_TEST_CASE("memBufAccessorAdaptorTest", "[memBuf]", alpaka::test::T
 
     testBufferAccessorAdaptor<Acc>(extent, index);
 }
+
+TEMPLATE_LIST_TEST_CASE("memBufMove", "[memBuf]", alpaka::test::TestAccs)
+{
+    using Acc = TestType;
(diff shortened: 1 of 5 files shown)


test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -264,3 +264,34 @@ TEMPLATE_LIST_TEST_CASE("memBufAccessorAdaptorTest", "[memBuf]", alpaka::test::T
...
+    using Elem = std::size_t;
+
+    auto const dev = alpaka::getDevByIdx<Pltf>(0u);
+    auto const extent = alpaka::Vec<alpaka::DimInt<0>, Idx>{};
+
+    // move constructor
+    {
+        auto buf1 = alpaka::allocBuf<Elem, Idx>(dev, extent);
+        *buf1 = 1;
+        auto buf2{std::move(buf1)};
```
fwyzard, line 297: at this point `buf1` is invalid, right ?

bernhardmgruber: yes. It's a moved from object. But it should destruct fine.

j-stephan (approved):

PR #1619, opened 2022-03-01
Also run unit tests with zero dimensions

### LINKED ISSUE #1538: Test and extend zero-dimensional usage
Lots of components in alpaka are instantiated for some kind of dimensionality. Vectors for sizes and extents, buffers, accelerators, views, etc. Zero-dimensional versions of these might be handy in some situation, like zero-dimensional buffers: #1535. Let's revisit all alpaka constructs and check whether they allow zero-dimensional usage and extend the unit tests to cover this.

bernhardmgruber:
Fixes: #1538

### DIFF
--- a/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
@@ -96,12 +96,16 @@ namespace alpaka
             template<typename TDim, typename TIdx>
             ALPAKA_FN_HOST auto checkVecOnly3Dim(Vec<TDim, TIdx> const& vec) -> void
             {
-                for(auto i = std::min(static_cast<typename TDim::value_type>(3), TDim::value); i < TDim::value; ++i)
+                if constexpr(TDim::value > 0)
                 {
-                    if(vec[TDim::value - 1u - i] != 1)
+                    for(auto i = std::min(typename TDim::value_type{3}, TDim::value); i < TDim::value; ++i)
                     {
-                        throw std::runtime_error("The CUDA/HIP accelerator supports a maximum of 3 dimensions. All "
-                                                 "work division extents of the dimensions higher 3 have to be 1!");
+                        if(vec[TDim::value - 1u - i] != 1)
+                        {
+                            throw std::runtime_error(
+                                "The CUDA/HIP accelerator supports a maximum of 3 dimensions. All "
+                                "work division extents of the dimensions higher 3 have to be 1!");
+                        }
                     }
                 }
             }
@@ -110,12 +114,12 @@ namespace alpaka
             ALPAKA_FN_HOST auto convertVecToUniformCudaHipDim(Vec<TDim, TIdx> const& vec) -> dim3
             {
                 dim3 dim(1, 1, 1);
-                for(auto i = static_cast<typename TDim::value_type>(0);
-                    i < std::min(static_cast<typename TDim::value_type>(3), TDim::value);
-                    ++i)
-                {
-                    reinterpret_cast<unsigned int*>(&dim)[i] = static_cast<unsigned int>(vec[TDim::value - 1u - i]);
-                }
+                if constexpr(TDim::value >= 1)
+                    dim.x = static_cast<unsigned>(vec[TDim::value - 1u]);
+                if constexpr(TDim::value >= 2)
+                    dim.y = static_cast<unsigned>(vec[TDim::value - 2u]);
+                if constexpr(TDim::value >= 3)
+                    dim.z = static_cast<unsigned>(vec[TDim::value - 3u]);
                 checkVecOnly3Dim(vec);
                 return dim;
             }
--- a/include/alpaka/mem/buf/oacc/Set.hpp
+++ b/include/alpaka/mem/buf/oacc/Set.hpp
@@ -36,85 +36,82 @@ namespace alpaka
     class DevOacc;
 }
 
-namespace alpaka
+namespace alpaka::trait
 {
-    namespace trait
+    //! The OpenACC device memory set trait specialization.
+    template<typename TDim>
+    struct CreateTaskMemset<TDim, DevOacc>
     {
-        //! The OpenACC device memory set trait specialization.
-        template<typename TDim>
-        struct CreateTaskMemset<TDim, DevOacc>
+        template<typename TExtent, typename TView>
+        ALPAKA_FN_HOST static auto createTaskMemset(TView& view, std::uint8_t const& byte, TExtent const& extent)
         {
-            template<typename TExtent, typename TView>
-            ALPAKA_FN_HOST static auto createTaskMemset(TView& view, std::uint8_t const& byte, TExtent const& extent)
-            {
-                using Idx = typename trait::IdxType<TExtent>::type;
-                auto pitch = getPitchBytesVec(view);
-                auto byteExtent = getExtentVec(extent);
-                byteExtent[TDim::value - 1] *= static_cast<Idx>(sizeof(Elem<TView>));
-                constexpr auto lastDim = TDim::value - 1;
+            using Idx = typename trait::IdxType<TExtent>::type;
+            auto pitch = getPitchBytesVec(view);
+            auto byteExtent = getExtentVec(extent);
+            byteExtent[TDim::value - 1] *= static_cast<Idx>(sizeof(Elem<TView>));
+            constexpr auto lastDim = TDim::value - 1;
 
-                if(pitch[0] <= 0)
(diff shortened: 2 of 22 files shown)


bernhardmgruber:
There are some problems around `alpaka::Vec`, so I am waiting for #1620 to easen debugging.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -96,26 +96,24 @@ namespace alpaka
...
-                        throw std::runtime_error("The CUDA/HIP accelerator supports a maximum of 3 dimensions. All "
-                                                 "work division extents of the dimensions higher 3 have to be 1!");
-                    }
-                }
+                if constexpr(TDim::value > 0)
+                    for(auto i = std::min(typename TDim::value_type{3}, TDim::value); i < TDim::value; ++i)
+                        if(vec[TDim::value - 1u - i] != 1)
+                            throw std::runtime_error(
+                                "The CUDA/HIP accelerator supports a maximum of 3 dimensions. All "
+                                "work division extents of the dimensions higher 3 have to be 1!");
```
j-stephan: Please add a few `{}` here.

include/alpaka/vec/Vec.hpp
```diff
@@ -275,8 +276,9 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC friend constexpr auto operator+(Vec const& p, Vec const& q) -> Vec
         {
             Vec r;
-            for(typename TDim::value_type i = 0; i < TDim::value; ++i)
-                r[i] = p[i] + q[i];
+            if constexpr(TDim::value > 0)
+                for(typename TDim::value_type i = 0; i < TDim::value; ++i)
+                    r[i] = p[i] + q[i];
```
j-stephan:
```suggestion
            if constexpr(TDim::value > 0)
            {
                for(typename TDim::value_type i = 0; i < TDim::value; ++i)
                    r[i] = p[i] + q[i];
            }
```

Also for the other instances.

include/alpaka/vec/Vec.hpp
```diff
@@ -285,8 +287,21 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC friend constexpr auto operator-(Vec const& p, Vec const& q) -> Vec
         {
             Vec r;
-            for(typename TDim::value_type i = 0; i < TDim::value; ++i)
-                r[i] = p[i] - q[i];
+#if BOOST_COMP_NVCC && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 3, 0)
+            if(TDim::value > 0)
```
j-stephan, line 293: Are older nvcc versions unable to consume `if constexpr`?

j-stephan (changes requested):
A few nitpicks:

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
bernhardmgruber: Do we have a style guide recommendation for that? I can only find it for `if` statements, where parentheses should be omitted if the body is a single statement: https://github.com/ComputationalRadiationPhysics/contributing/blob/master/codingGuideLines/cpp.md#12-if-statements

Personally, I also try to avoid all braces so I have less lines wasted with the closing parentheses. I did add them in a few places though, because of nvcc bugs with `if constexpr`.

include/alpaka/vec/Vec.hpp
bernhardmgruber: Not in this specific instance, where you will get a syntax error with nvcc 11.2 and below. Generally, `if constexpr` is supported.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
j-stephan: > parentheses should be omitted if the body is a single statement:

Parentheses should be omitted if the body is a *one liner*, not a single statement. I don't think the construction above counts as one line...

bernhardmgruber: I find that fragile. That would mean if I would refactor a variable name and clang-format then decides to break, that I would need to add braces as well.

bernhardmgruber:
I added a few more unnecessary braces for you @j-stephan.

j-stephan:
They are necessary... from a certain point of view.

include/alpaka/mem/buf/omp5/Set.hpp
```diff
@@ -96,20 +97,23 @@ namespace alpaka
...
-                using Vec1D = Vec<Dim1D, Idx>;
-
                 static_assert(Dim<TView>::value == 0u, "The view is required to have dimensionality 0!");
                 static_assert(Dim<TExtent>::value == 0u, "The extent is required to have dimensionality 0!");
 
-                return createTaskKernel<AccOmp5<Dim1D, Idx>>(
-                    WorkDivMembers<Dim1D, Idx>(Vec1D::zeros(), Vec1D::zeros(), Vec1D::zeros()),
-                    MemSetKernel(),
+                using ExtIdx = Idx<TExtent>;
+                using Dim0 = DimInt<0u>;
```
psychocoderHPC: Why do you change this kernel? This kernel must be a one-dimensional kernel to emulate the memset function on the device.

include/alpaka/mem/buf/omp5/Set.hpp
```diff
@@ -96,20 +97,23 @@ namespace alpaka
...
+                using Acc = AccOmp5<Dim0, ExtIdx>;
+                auto kernel = [] ALPAKA_FN_ACC(const Acc&, std::uint8_t b, std::uint8_t* mem)
+                {
+                    for(auto i = 0u; i < sizeof(Elem<TView>); i++)
```
psychocoderHPC: Please call the original kernel `MemSetKernel` because your for loop is not covering pitched memory.

include/alpaka/mem/buf/oacc/Set.hpp
```diff
@@ -98,20 +98,23 @@ namespace alpaka
...
                 static_assert(Dim<TView>::value == 0u, "The view is required to have dimensionality 0!");
                 static_assert(Dim<TExtent>::value == 0u, "The extent is required to have dimensionality 0!");
 
-                return createTaskKernel<AccOacc<DimInt<1u>, Idx>>(
-                    WorkDivMembers<DimInt<1u>, Idx>(Vec1D::zeros(), Vec1D::zeros(), Vec1D::zeros()),
-                    MemSetKernel(),
+                using ExtIdx = Idx<TExtent>;
+                using Dim0 = DimInt<0u>;
+                using Acc = AccOacc<Dim0, ExtIdx>;
+                auto kernel = [] ALPAKA_FN_ACC(const Acc&, std::uint8_t b, std::uint8_t* mem)
```
psychocoderHPC: Same as in the Omp5 `Set` function. Please keep the old implementation.

psychocoderHPC (changes requested):

include/alpaka/mem/buf/omp5/Set.hpp
bernhardmgruber: AFAIK I only changed the implementation of the zero dimensional Set implementation, where there are no pitches and we only set `sizeof(TElem)` bytes. I can add a comment to point that out more explicitely.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -96,12 +96,16 @@ namespace alpaka
...
+                    for(auto i = std::min(typename TDim::value_type{3}, TDim::value); i < TDim::value; ++i)
                     {
-                        throw std::runtime_error("The CUDA/HIP accelerator supports a maximum of 3 dimensions. All "
-                                                 "work division extents of the dimensions higher 3 have to be 1!");
+                        if(vec[TDim::value - 1u - i] != 1)
+                        {
+                            throw std::runtime_error(
+                                "The CUDA/HIP accelerator supports a maximum of 3 dimensions. All "
+                                "work division extents of the dimensions higher 3 have to be 1!");
+                        }
```
j-stephan, line 108: It is perfect! Perfect in every way!

j-stephan (approved):

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
bernhardmgruber: :*

psychocoderHPC (approved):

PR #1620, opened 2022-03-02
Refactor Vec

### LINKED ISSUE #1015: Integer underflow when creating Vec
Given the following code
```
auto v = alpaka::extent::getExtentVecEnd<alpaka::dim::DimInt<2>>(size_t{1});
```
nvcc yields this warning: `warning #68-D: integer conversion resulted in a change of sign`.

The reason is that, `getExtentVecEnd` will eventually expand to:
```
alpaka::vec::Vec<
        alpaka::dim::DimInt<2>,
        decltype(alpaka::extent::detail::CreateExtent<0>::create(size_t{1}))>(
        (alpaka::extent::detail::CreateExtent<-1>::create(size_t{1})),
        alpaka::extent::detail::CreateExtent<0>::create(size_t{1}));
```
The substitution `alpaka::extent::detail::CreateExtent<-1>` is triggering the warning, because the non-type template parameter is `size_t` and -1 is passed.

This whole situation arises, because I create a workdiv for an alpaka accelerator passed in by the user, of which I do not know the dimensionality, but I want to enqueue a 1D kernel:
```
template<typename AlpakaAcc>
void f(AlpakaAcc & acc)
{
    using Dim = typename alpaka::dim::traits::DimType<AlpakaAcc>::type;
    using Idx = typename alpaka::idx::traits::IdxType<AlpakaAcc>::type;
    const auto workDiv
        = alpaka::workdiv::WorkDivMembers<Dim, Idx>{Idx{1}, Idx{256}, Idx{1}};
}
```
The code works at runtime, because `alpaka::extent::detail::CreateExtent<-1>::create()` yields 0. Still, the warning is annoying and I feel like you have not thought about this case.

I suggest either changing the `size_t` non-type template parameters to a signed data type or preventing the whole thing from compiling.

### LINKED ISSUE #1389: alpaka::Vec ctor too restrictive
The ctor of `alpaka::Vec` taking a pack of arguments is too restrictive IMO:
```c++
using Vec = alpaka::Vec<alpaka::DimInt<2>, std::size_t>;
Vec v(1, 2); // compilation error, because decltype(1) != std::size_t
```
The workaround is to cast the arguments:
```c++
Vec v(std::size_t{1}, std::size_t{2});
```
I think this requirement is too restrictive and the ctor requirement `std::enable_if_t<(sizeof...(TArgs) + 1 == TDim::value) && (std::is_same<TVal, std::decay_t<TArg0>>::value)>>` should be weakened into `std::enable_if_t<(sizeof...(TArgs) + 1 == TDim::value) && (std::is_convertible<std::decay_t<TArg0>, TVal>::value)>>`.

bernhardmgruber:
I had to debug around `Vec` and here is a quick simplification.

The big change is that `Vec` is now default constructible, which simplifies other implementations a lot. Furthermore, all functions are now `constexpr`. And I added `begin/end`.

Also fixes #1015 and #1389.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -98,13 +98,10 @@ namespace alpaka
         //! This constructor is only available if the number of parameters matches the vector idx.
         ALPAKA_NO_HOST_ACC_WARNING
         template<
-            typename TArg0,
             typename... TArgs,
             typename = std::enable_if_t<
-                // There have to be dim arguments.
-                (sizeof...(TArgs) + 1 == TDim::value) && (std::is_same_v<TVal, std::decay_t<TArg0>>)>>
-        ALPAKA_FN_HOST_ACC constexpr Vec(TArg0&& arg0, TArgs&&... args)
-            : m_data{std::forward<TArg0>(arg0), std::forward<TArgs>(args)...}
+                sizeof...(TArgs) == TDim::value && (std::is_convertible_v<std::decay_t<TArgs>, TVal> && ...)>>
+        ALPAKA_FN_HOST_ACC constexpr Vec(TArgs&&... args) : m_data{static_cast<TVal>(std::forward<TArgs>(args))...}
         {
         }
 
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -532,18 +532,19 @@ namespace alpaka
     //! \return The extent vector.
     ALPAKA_NO_HOST_ACC_WARNING
     template<typename TExtent>
-    ALPAKA_FN_HOST_ACC auto constexpr getExtentVec(TExtent const& extent = TExtent())
-        -> Vec<Dim<TExtent>, Idx<TExtent>>
+    ALPAKA_FN_HOST_ACC auto constexpr getExtentVec(TExtent const& extent = {}) -> Vec<Dim<TExtent>, Idx<TExtent>>
     {
         return createVecFromIndexedFn<Dim<TExtent>, detail::CreateExtent>(extent);
     }
 
     //! \tparam TExtent has to specialize GetExtent.
-    //! \return The extent but only the last N elements.
+    //! \return The extent but only the last TDim elements.
     ALPAKA_NO_HOST_ACC_WARNING
     template<typename TDim, typename TExtent>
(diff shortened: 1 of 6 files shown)


j-stephan:
You have created two issues with regard to `Vec` in the past: #1015, #1389. Is any of these fixed by this PR?

bernhardmgruber:
> You have created two issues with regard to `Vec` in the past: #1015, #1389. Is any of these fixed by this PR?

No.

j-stephan:
Any chance you can fix them as part of this PR? :-)

bernhardmgruber:
> Any chance you can fix them as part of this PR? :-)

Fine.

include/alpaka/vec/Vec.hpp
```diff
@@ -446,141 +407,91 @@ namespace alpaka
...
+                Vec<TDimL, TVal> const& vecL,
+                Vec<TDimR, TVal> const& vecR) -> Vec<DimInt<TDimL::value + TDimR::value>, TVal>
             {
-                return createVecFromIndexedFn<DimInt<TDimL::value + TDimR::value>, alpaka::detail::CreateConcat>(
-                    vecL,
-                    vecR);
+                Vec<DimInt<TDimL::value + TDimR::value>, TVal> r;
+#if BOOST_COMP_NVCC
+#    pragma push
+#    pragma nv_diag_suppress = unsigned_compare_with_zero
```
fwyzard: does nvcc still generate the code for the loop, if `TDimL::value` is 0 ?

bernhardmgruber: it at least complains about it in the frontend.

fwyzard: mhm, IIRC what I've done elsewhere was to protect the loop with something like `if constexpr (TDimL::value > 0) { ... }`

bernhardmgruber: sure, that would work as well. but it should not be necessary. a loop like that:
```c++
for (unsigned int i = 0; i < 0; i++) { .. }
```
is fine and will just not execute its body. that's why I chose to suppress the warning rather than adding the conditional, which might maybe cause another warning.

bernhardmgruber:
This is so annoying. I cannot reproduce the compilation failure with g++ 7.5 and nvcc 11.0.194 on my machine. The CI runs with nvcc 11.0.221.

bernhardmgruber:
I added a few workarounds for nvcc 11.0/11.1.

j-stephan (approved):

PR #1622, opened 2022-03-04
Make alpaka_BUILD_EXAMPLES and alpaka_INSTALL_TEST_HEADER uppercase

bernhardmgruber:
All other alpaka cmake options are uppercase too.


j-stephan (approved):

psychocoderHPC:
Note: IMO it should be the other way around. `alpaka` should always be small letters https://github.com/alpaka-group/alpaka/pull/1650#issuecomment-1070383988

PR #1627, opened 2022-03-07
Remove incompatible clang + CUDA combinations

### LINKED ISSUE #1625: clang + libstdc++ 9.4 + nvcc broken
Today our ubuntu containers updated from libstdc++ 9.3 to 9.4, which broke compilation with nvcc + clang as host compiler.

See e.g. this CI run from today at 11:00: https://gitlab.com/hzdr/crp/alpaka/-/pipelines/485977443
And this one from 13:00: https://gitlab.com/hzdr/crp/alpaka/-/pipelines/486110519

The error message starts with:
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(70): error: qualified name is not allowed
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(72): error: expected a "("
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(72): error: expected a type specifier
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(72): error: expected a ")"
/usr/bin/../lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/ext/numeric_traits.h(76): error: expected a "("
...
```
The problem is also referenced in this NVIDIA forum post: https://forums.developer.nvidia.com/t/clang12-is-not-able-to-compile-a-simple-test-program-with-cuda11-4-3-on-ubuntu20-04-3-lts-on-wsl/200000.

j-stephan:
Attempt to fix #1625. Only a few CUDA versions seem to be affected so let us try to single them out and disable them.


bernhardmgruber:
In principle I strongly support resolving the issue, but the PR seems a bit drastic to me. I think the root problem is libstdc++ 9.4? Maybe we can use a different libstdc++. Not necessarily downgrading to 9.3, but maybe installing 10?

bernhardmgruber:
I wonder if/how the landscape would change when ubuntu 22.04 becomes available. I guess it will come with a newer libstdc++. Until then, the easiest solution might really be to disable the offending runs. We should create a ticket though to re-enable these runs in the future.

bernhardmgruber (dismissed):

bernhardmgruber (approved):

bernhardmgruber:
@j-stephan please open an issue that we should try the removed runs again with ubuntu 22.04!

PR #1631, opened 2022-03-09
Add ROCm 5.0 support

### LINKED ISSUE #1600: Add ROCm 5.0 to CI
ROCm 5.0 was released a few days ago: https://www.phoronix.com/scan.php?page=news_item&px=Radeon-ROCm-5.0

It should be added to the CI.

j-stephan:
Fixes #1600. Also balances out the GitLab CI pipelines a bit.


cmake/alpakaCommon.cmake
```diff
@@ -505,9 +505,14 @@ endif()
 # Find HIP.
 if(ALPAKA_ACC_GPU_HIP_ENABLE)
 
-    # minimal supported HIP version
+    # supported HIP version range
     set(_ALPAKA_HIP_MIN_VER 4.0)
-    find_package(hip "${_ALPAKA_HIP_MIN_VER}")
+    set(_ALPAKA_HIP_MAX_VER 5.0)
+    find_package(hip "${_ALPAKA_HIP_MAX_VER}")
```
j-stephan, line 511: CMake will not accept ROCm 5.0 if we are searching for 4.0. So we have to search for both versions here.

fwyzard: does CMake care about the minor versions, _e.g._ 5.0.2 vs 5.0 ?

j-stephan: It is not actually CMake itself complaining but the CMake configuration script which is part of the ROCm installation. In this case the script (`hip-config-version.cmake`) only cares about major versions, so every 5.x version should work.

README.md
```diff
@@ -83,6 +83,7 @@ This library uses C++17 (or newer when available).
...
+| [HIP-5.0](https://alpaka.readthedocs.io/en/latest/install/HIP.html) (clang)| :x:                                       | :x:                                       | :x:                                       | :x:                                  | :x:                    | :x:                     | :x:                                       | :x:                                       | :x:                                       | :x:                                       | :white_check_mark:                   | -                                                     | -                                         | -                                    |
```
bernhardmgruber, line 86: I think we should merge the table rows similar to how we do it with CUDA.

j-stephan: If the CI passes I'll do that in a follow-up PR. We should have a talk about supported HIP versions anyway...

Edit: And if does not pass I'll do it in this PR along with whatever needs a fix.

bernhardmgruber (approved):

cmake/alpakaCommon.cmake
psychocoderHPC: @steindev observed the same issue with ROCm 5.1 on crusher (ORNL) . I was on the way to open an issue and now saw that we solved the issue already :tada:

PR #1632, opened 2022-03-09
Do not throw in destructors

### LINKED ISSUE #468: Destructors should not throw
Colleagues from the DLR using Alpaka run into a problem while using Alpaka together with a singleton pattern, in which a single global class handles the program environment.

Independently whether one should use this pattern or not, CUDA has problems with global variables as the may destruct after the cuda driver is already shut down: https://stackoverflow.com/questions/40979060/cudaerrorcudartunloading-error-29-due-to-driver-shutting-down

So in the end `ALPAKA_CUDA_RT_CHECK` fails with an unhandled exception: https://github.com/ComputationalRadiationPhysics/alpaka/blob/develop/include/alpaka/stream/StreamCudaRtSync.hpp#L103

As it is not possible to catch these exceptions destructors should normally not throw, but the class should have explicit finishinig / closing methods which can be checked for errors (if wanted), see https://stackoverflow.com/questions/130117/throwing-exceptions-out-of-a-destructor .

So I suggest to make every destructor of Alpaka non-throwing, e.g. by catching all exceptions inside the destructor so that error messages from `ALPAKA_CUDA_RT_CHECK` still appear.

j-stephan:
Fixes #468.

This PR adds a new CUDA/HIP error checking macro which just prints the error message instead of throwing a `std::runtime_error` (which is useful in destructors).

### DIFF
--- a/include/alpaka/core/UniformCudaHip.hpp
+++ b/include/alpaka/core/UniformCudaHip.hpp
@@ -34,35 +34,44 @@ namespace alpaka::uniform_cuda_hip::detail
 #    else
     using Error_t = hipError_t;
 #    endif
+
     //! CUDA/HIP runtime API error checking with log and exception, ignoring specific error values
-    ALPAKA_FN_HOST inline auto rtCheck(Error_t const& error, char const* desc, char const* file, int const& line)
-        -> void
+    template<bool TThrow>
+    ALPAKA_FN_HOST inline void rtCheck(
+        Error_t const& error,
+        char const* desc,
+        char const* file,
+        int const& line) noexcept(!TThrow)
     {
         if(error != ALPAKA_API_PREFIX(Success))
         {
-            std::string const sError(
+            auto const sError = std::string{
                 std::string(file) + "(" + std::to_string(line) + ") " + std::string(desc) + " : '"
                 + ALPAKA_API_PREFIX(GetErrorName)(error) + "': '"
-                + std::string(ALPAKA_API_PREFIX(GetErrorString)(error)) + "'!");
-#    if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
-            std::cerr << sError << std::endl;
-#    endif
+                + std::string(ALPAKA_API_PREFIX(GetErrorString)(error)) + "'!"};
+
+            if constexpr(!TThrow || ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL)
+                std::cerr << sError << std::endl;
+
             ALPAKA_DEBUG_BREAK;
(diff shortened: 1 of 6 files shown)


include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -34,29 +34,35 @@ namespace alpaka::uniform_cuda_hip::detail
 #    else
     using Error_t = hipError_t;
 #    endif
+
     //! CUDA/HIP runtime API error checking with log and exception, ignoring specific error values
+    template<bool TThrow>
     ALPAKA_FN_HOST inline auto rtCheck(Error_t const& error, char const* desc, char const* file, int const& line)
         -> void
```
bernhardmgruber:
```suggestion
    template<bool TThrow>
    ALPAKA_FN_HOST inline void rtCheck(Error_t const& error, char const* desc, char const* file, int const& line) noexcept(!TThrow)
```

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -82,40 +88,68 @@ namespace alpaka::uniform_cuda_hip::detail
             }
         }
     }
-    //! CUDA runtime API last error checking with log and exception.
+    //! CUDA/HIP runtime API last error checking with log and exception.
+    template<bool TThrow>
     ALPAKA_FN_HOST inline auto rtCheckLastError(char const* desc, char const* file, int const& line) -> void
```
bernhardmgruber:
```suggestion
    template<bool TThrow>
    ALPAKA_FN_HOST inline void rtCheckLastError(char const* desc, char const* file, int const& line) noexcept(!TThrow)
```

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -34,29 +34,35 @@ namespace alpaka::uniform_cuda_hip::detail
...
+                throw std::runtime_error(sError);
         }
     }
-    //! CUDA/Hip runtime API error checking with log and exception, ignoring specific error values
+
+    //! CUDA/HIP runtime API error checking with log and exception, ignoring specific error values
     // NOTE: All ignored errors have to be convertible to Error_t.
-    template<typename... TErrors>
+    template<bool TThrow, typename... TErrors>
     ALPAKA_FN_HOST auto rtCheckIgnore(
```
bernhardmgruber: Also add `noexcept(!TThrow)` here.

bernhardmgruber (dismissed):

bernhardmgruber (dismissed):

bernhardmgruber (approved):

Issue #1634, opened 2022-03-09
[RFC] Remove support for `std::function` kernels.


j-stephan:
I propose to remove support for `std::function` kernels because of the following reasons:

1. Using a `std::function` as a kernel task is currently only supported for CPU back-ends. For CUDA and HIP this is explicitly disabled and with SYCL it is untested.
2. Since all back-ends support trivially copyable kernel arguments we should enforce that all arguments passed to the underlying back-end kernel are in fact trivially copyable, including the user-defined kernel task. 
3. I've never seen a `std::function` based alpaka kernel outside of our test cases.

bernhardmgruber:
I also don´t see a big use case for passing kernels as `std::function`s. You have my support!

psychocoderHPC:
I would avoid supporting std::function. The reason is that writing heterogenious code would be much harder if there are ways to spawn kernel which will not be supported by all backends.

PR #1635, opened 2022-03-09
Enforce trivially copyable kernel arguments

### LINKED ISSUE #703: Investigate usage of std::is_trivially_copyable

### LINKED ISSUE #1384: kernel parameter should be trivially copy-able
During the CUDA HIP refactoring, important checks were removed

https://github.com/alpaka-group/alpaka/blob/ab0b8a460f398fef7e475892ff5af98f947ce128/include/alpaka/kernel/TaskKernelGpuCudaRt.hpp#L145-L154

These checks are important to check compatibility and must be reintroduced.

j-stephan:
Fixes #1384. This PR enforces all kernel arguments to be trivially copyable.

The check for the kernel function object being trivially copyable is deliberately left out until a consensus on #1634 has been reached.

### DIFF
--- a/include/alpaka/kernel/Traits.hpp
+++ b/include/alpaka/kernel/Traits.hpp
@@ -223,6 +223,9 @@ namespace alpaka
         detail::CheckFnReturnType<TAcc>{}(kernelFnObj, args...);
 
         static_assert(
+            (std::is_trivially_copyable_v<std::decay_t<TArgs>> && ...),
+            "Kernel arguments must be trivially copyable!");
+        static_assert(
             Dim<std::decay_t<TWorkDiv>>::value == Dim<TAcc>::value,
             "The dimensions of TAcc and TWorkDiv have to be identical!");
         static_assert(
--- a/include/alpaka/test/mem/view/Iterator.hpp
+++ b/include/alpaka/test/mem/view/Iterator.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Benjamin Worpitz, Erik Zenker, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, Erik Zenker, Bernhard Manfred Gruber, Jan Stephan
  *
  * This file is part of alpaka.
  *
@@ -122,10 +122,10 @@ namespace alpaka::test
             }
 
         private:
-            Elem* const m_nativePtr;
+            Elem* m_nativePtr;
             Idx m_currentIdx;
-            Vec<Dim, Idx> const m_extents;
-            Vec<Dim, Idx> const m_pitchBytes;
+            Vec<Dim, Idx> m_extents;
+            Vec<Dim, Idx> m_pitchBytes;
         };
 #if BOOST_COMP_GNUC
 #    pragma GCC diagnostic pop
--- a/test/unit/math/src/Functor.hpp
+++ b/test/unit/math/src/Functor.hpp
@@ -36,7 +36,7 @@ namespace alpaka
         /* ranges is not a constexpr, so that it's accessible via for loop*/                                          \
         static constexpr Arity arity = ARITY;                                                                         \
         static constexpr size_t arity_nr = static_cast<size_t>(ARITY);                                                \
-        const Range ranges[arity_nr] = {__VA_ARGS__};                                                                 \
+        Range ranges[arity_nr] = {__VA_ARGS__};                                                                       \
                                                                                                                       \
         ALPAKA_NO_HOST_ACC_WARNING                                                                                    \
         template<                                                                                                     \


include/alpaka/kernel/Traits.hpp
```diff
@@ -222,6 +222,9 @@ namespace alpaka
         // check for void return type
         detail::CheckFnReturnType<TAcc>{}(kernelFnObj, args...);
 
+        static_assert(
+            std::conjunction_v<std::is_trivially_copyable<std::decay_t<TArgs>>...>,
+            "Kernel arguments must be trivially copyable!");
```
bernhardmgruber:
```suggestion
        static_assert(
            (std::is_trivially_copyable_v<std::decay_t<TArgs>> && ... && true),
            "Kernel arguments must be trivially copyable!");
```
I prefer a fold, but up to you!

bernhardmgruber (dismissed):

include/alpaka/kernel/Traits.hpp
j-stephan: From what I understand a fold instantiates all parameters while `std::conjunction` will look at them individually (and fail at the first `false`).

fwyzard: By the way, the final `&& true` is not necessary.

bernhardmgruber: Maybe, but it should not matter. Anyway, do what you prefer! :)

bernhardmgruber: > By the way, the final && true is not necessary.

That would not work for kernels without arguments, which are probably rare or even nonsensical. But why prevent it?

fwyzard: Mhm, I think it should work also with an empty argument list: https://godbolt.org/z/TP99M1nfh . 

[Cppreference](https://en.cppreference.com/w/cpp/language/fold#Explanation) says
> When a unary fold is used with a pack expansion of length zero, only the following operators are allowed:
> 1) Logical AND (`&&`). The value for the empty pack is `true`

fwyzard: > From what I understand a fold instantiates all parameters while `std::conjunction` will look at them individually (and fail at the first `false`).

I think the fold expression itself does follow short-circuit evaluation: https://godbolt.org/z/e3jbbh4d3 .
The values after the first `false` are not evaluated.

j-stephan:
Interestingly clang and gcc disagree on what is a `trivially_copyable` type: https://godbolt.org/z/svYv4Ydxx

include/alpaka/kernel/Traits.hpp
j-stephan: I was following the info from this link: https://stackoverflow.com/questions/55161235/is-there-a-reason-to-use-stdconjunction-stddisjunction-instead-of-a-fold-exp#comment97065030_55161485

fwyzard:
Interesting...

The do agree on the individual data members (they are all trivially copyable), but not on the aggregate structure 🤔

include/alpaka/kernel/Traits.hpp
j-stephan: But anyway, I don't think this will hurt our compilation performance much. And having a language feature instead of a library feature is also nice, so I'll change to a fold expression.

fwyzard:
If I add an explicit default constructor to the example
```c++
struct X
{
    X() = default;
    
    Vec const ext;
    float* ptr;
    std::integral_constant<unsigned long, 1> idx;
};
```

GCC is happy, while clang complains with
```
<source>:9:5: warning: explicitly defaulted default constructor is implicitly deleted [-Wdefaulted-function-deleted]
    X() = default;
    ^
<source>:11:15: note: default constructor of 'X' is implicitly deleted because field 'ext' of const-qualified type 'const Vec' would not be initialized
    Vec const ext;
```

So, clang does not consider `X` to be `trivially_constructible`, while GCC does.

j-stephan:
Yeah... I'll just remove the `const`ness of `IteratorView`'s members, it doesn't serve a purpose anyway.

fwyzard:
There seem to be an open defect report about this since quite some time now: http://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#253 .

include/alpaka/kernel/Traits.hpp
fwyzard: Ah, I see. Yes, you are right, the arguments are instantiated, even if they are not evaluated: https://godbolt.org/z/osjbc98an .

bernhardmgruber (approved):

PR #1637, opened 2022-03-10
Merge HIP rows in `README.md`

j-stephan:
Merges the three HIP rows in the compiler support matrix into one.


bernhardmgruber (approved):
Such beauty!

PR #1640, opened 2022-03-10
Retry downloading boost

bernhardmgruber:


fwyzard:
In case it helps, I found the Sourforge mirror of boost to be a bit more reliable than the official one.

bernhardmgruber:
> In case it helps, I found the Sourforge mirror of boost to be a bit more reliable than the official one.

Thx for the tip! If the problem persists, we can try this as well!

j-stephan (approved):

PR #1641, opened 2022-03-11
Add a grid-wise memory fence

fwyzard:
A grid-wise memory fence is a no-op in the CPU serial case, running with a single thread; and is implemented as a per-device fence in the CUDA/HIP, OpenMP 2, OpenMP 5 and SYCL backends, that do not have a native per-grid fence.

As usual there are no fences for the OpenAcc backend.

Documentation and tests have been updated accordingly.

### DIFF
--- a/docs/source/basic/library.rst
+++ b/docs/source/basic/library.rst
@@ -161,9 +161,10 @@ guarantees the following **for the local thread**  and regardless of global or s
 **Note**: ``alpaka::mem_fence`` does not guarantee that there will be no *LoadStore* reordering. Depending on the
 back-end, loads occurring before the fence may still be reordered with stores occurring after the fence.
 
-Memory fences can be issued on the block level (``alpaka::memory_scope::Block``) and the device level
-(``alpaka::memory_scope::Device``). Depending on the memory scope, the *StoreStore* order will be visible to other
-threads in the same block or the whole device.
+Memory fences can be issued on the block level (``alpaka::memory_scope::Block``), grid level
+(``alpaka::memory_scope::Grid``) and the device level (``alpaka::memory_scope::Device``).
+Depending on the memory scope, the *StoreStore* order will be visible to other threads in the same block, in the same grid
+(_i.e._ within the same kernel launch), or on the whole device (_i.e._ across concurrent kernel launches).
 
 Some accelerators (like GPUs) follow weaker cache coherency rules than x86 CPUs. In order to avoid storing to (or loading
 from) a cache or register it is necessary to prefix all observed buffers with `ALPAKA_DEVICE_VOLATILE`. This enforces
--- a/test/unit/mem/fence/src/FenceTest.cpp
+++ b/test/unit/mem/fence/src/FenceTest.cpp
@@ -20,8 +20,17 @@ struct IsSingleThreaded : public std::false_type
 {
 };
 
+/* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because these
+/  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
 template<typename TAcc>
 inline constexpr bool isSingleThreaded = IsSingleThreaded<TAcc>::value;
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
 
 #ifdef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
 template<typename TDim, typename TIdx>
--- a/test/unit/mem/fence/src/FenceTest.cpp
+++ b/test/unit/mem/fence/src/FenceTest.cpp
@@ -14,7 +14,77 @@
 
 #include <catch2/catch.hpp>
 
-class DeviceFenceTestKernel
+// Trait to detect whether an accelerator supports or not multiple threads per block
+template<typename TAcc>
+struct IsSingleThreaded : public std::false_type
+{
+};
+
+template<typename TAcc>
+inline constexpr bool isSingleThreaded = IsSingleThreaded<TAcc>::value;
+
+#ifdef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
+template<typename TDim, typename TIdx>
+struct IsSingleThreaded<alpaka::AccCpuSerial<TDim, TIdx>> : public std::true_type
+{
+};
+#endif // ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
+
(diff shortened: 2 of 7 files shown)


fwyzard:
As usual there are no fences for the OpenAcc backend.

fwyzard:
I did not try to implement a grid-wise fence for the experimental SYCL backend.

fwyzard:
@tonydp03 @j-stephan FYI

j-stephan:
> I did not try to implement a grid-wise fence for the experimental SYCL backend.

There is actually no need for that. The SYCL back-end has a special case for the block-level fence and will resort to the device-wide fence otherwise.

j-stephan (changes requested):
Looks solid to me. Could you update the copyright headers, too?

fwyzard:
> Could you update the copyright headers, too?

Done.

j-stephan (dismissed):

fwyzard:
> The SYCL back-end has a special case for the block-level fence and will resort to the device-wide fence otherwise.

I see, thanks !

fwyzard:
As expected, using this new  level of fences our application achieved the same performance as using the original implementation: https://github.com/cms-patatrack/pixeltrack-standalone/pull/320#issuecomment-1065378932 .

bernhardmgruber (approved):

bernhardmgruber (changes requested):
Thank you for this PR! The implementaiton LGTM.

I would also like to see this new feature being called in one of the unit tests. And the cheatsheet mentions memory fences, so the new grid level option should be mentioned there as well.

fwyzard:
> I would also like to see this new feature being called in one of the unit tests.

The current test based on  `DeviceFenceTestKernel` is actually testing for grid-wise fences, since it involves a single kernel.


> And the cheatsheet mentions memory fences, so the new grid level option should be mentioned there as well.

OK, I've added it along with the block and device cases.

fwyzard:
> I would also like to see this new feature being called in one of the unit tests.

Done: I've repurposed the original device-wise fence test using a single kernel as a test for the grid-wise fences, ~~and added a dedicated test using two potentially concurrent kernels for the device-wise fences.~~

fwyzard:
OK, the grid-wise fence works.
I'll try to see what was the problem with the device-wise test I wrote in a separate PR.

fwyzard:
See #1642 for the new test.

j-stephan (dismissed):

j-stephan:
~~@bernhardmgruber Please don't merge before #1642 has passed the CI.~~

docs/source/basic/library.rst
```diff
@@ -161,9 +161,9 @@ guarantees the following **for the local thread**  and regardless of global or s
 **Note**: ``alpaka::mem_fence`` does not guarantee that there will be no *LoadStore* reordering. Depending on the
 back-end, loads occurring before the fence may still be reordered with stores occurring after the fence.
 
-Memory fences can be issued on the block level (``alpaka::memory_scope::Block``) and the device level
-(``alpaka::memory_scope::Device``). Depending on the memory scope, the *StoreStore* order will be visible to other
-threads in the same block or the whole device.
+Memory fences can be issued on the block level (``alpaka::memory_scope::Block``), grid level
+(``alpaka::memory_scope::Grid``) and the device level (``alpaka::memory_scope::Device``).
+Depending on the memory scope, the *StoreStore* order will be visible to other threads in the same block or the whole device.
```
psychocoderHPC: Currently, the new option `Grid` is not explained in the documentation string.

should be `Depending on the memory scope, the *StoreStore* order will be visible to other threads in the same block, other blocks in a grid or the whole device (grids between grids).`

psychocoderHPC (changes requested):

fwyzard:
> @bernhardmgruber Please don't merge before #1642 has passed the CI.

Why not ?
The new code I added is being tested successfully.

j-stephan:
> Why not ?

My bad, I wrote this before looking at #1642. I assumed you would test the new feature there.

docs/source/basic/library.rst
fwyzard: Done.

psychocoderHPC (dismissed):

j-stephan (dismissed):

fwyzard:
By the way, is the old test actually testing the grid-wise fence ?
Isn't `fixture` running with a single block, single thread ?

fwyzard:
I've come to the conclusion that both the test and the check are wrong :-/

The comments in the code with the implementations of the fences say
> The following are all valid outcomes:
>    `a == 1 && b == 2`
>    `a == 10 && b == 2`
>    `a == 10 && b == 20`

So the check should be `! (a == 1 && b == 20)`, or `a != 1 || b != 20`.

Instead the check requires `a != 1 && b == 20`.

so far the tests have all passed because they run a single kernel with a single block and a single thread -- so they effectively run serially the whole thing:
```c++
vars[0] = 10;
vars[1] = 20;
auto const b = vars[1];
auto const a = vars[0];
```
which is equivalent to setting `a = 10` and `b = 20`.

fwyzard:
And I've added the usual workaround for the buggy `-Wmissing-variable-declarations` warning in clang 5 and 6.

test/unit/mem/fence/src/FenceTest.cpp
```diff
@@ -14,7 +14,86 @@
 
 #include <catch2/catch.hpp>
 
-class DeviceFenceTestKernel
+// Trait to detect whether an accelerator supports or not multiple threads per block
+template<typename TAcc>
+struct IsSingleThreaded : public std::false_type
+{
+};
```
bernhardmgruber, line 21: I would just make this a variable template directly and then specialize it later:
```suggestion
template<typename TAcc>
inline constexpr bool isSingleThreaded = false;
```

bernhardmgruber (approved):

test/unit/mem/fence/src/FenceTest.cpp
fwyzard: I see... I would prefer to make this change globally, also in the other places that use template variables.

fwyzard:
Looks like the failure is due to a runner getting stuck ?

> ERROR: Job failed: execution took longer than 3h0m0s seconds

fwyzard:
> Looks like the failure is due to a runner getting stuck ?

Should I just re-trigger the tests ?
Any way to do it other than force-pushing to the branch ?

j-stephan:
Sorry for the late response, we just had our alpaka VC. This looks like the runner GPU crashed. We informed our IT department and are now waiting for the node to be rebooted. Afterwards we can restart the job, there is no need to do a force-push.

psychocoderHPC:
> > Looks like the failure is due to a runner getting stuck ?
> 
> Should I just re-trigger the tests ? Any way to do it other than force-pushing to the branch ?

The runner is back in action and I restarted the failing test.

j-stephan (approved):

fwyzard:
:+1:

PR #1642, opened 2022-03-14
Add a test for the device-wise fence

fwyzard:
Since the original device-wise fence test using a single kernel has been repurposed as a test for the grid-wise fences, this PR adds a new test for the device-wise fences using two concurrent kernels.

### DIFF
--- a/test/unit/mem/fence/src/FenceTest.cpp
+++ b/test/unit/mem/fence/src/FenceTest.cpp
@@ -14,7 +14,7 @@
 
 #include <catch2/catch.hpp>
 
-class DeviceFenceTestKernel
+class GridFenceTestKernel
 {
 public:
     template<typename TAcc>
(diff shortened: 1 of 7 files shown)


test/unit/mem/fence/src/FenceTest.cpp
```diff
@@ -113,17 +152,26 @@ TEMPLATE_LIST_TEST_CASE("FenceTest", "[fence]", TestAccs)
...
+    alpaka::WorkDivMembers<Dim, Idx> workDiv(
+        alpaka::Vec<Dim, Idx>::ones(),
+        alpaka::Vec<Dim, Idx>::ones(),
+        alpaka::Vec<Dim, Idx>::ones());
+    alpaka::exec<Acc>(queue, workDiv, deviceKernelWriter, vars_dev.data());
+    REQUIRE(fixture(deviceKernelReader, vars_dev.data()));
+    alpaka::wait(queue);
+
+    GridFenceTestKernel gridKernel;
+    REQUIRE(fixture(gridKernel, vars_dev.data()));
```
j-stephan, line 177: I think you need to reset `vars_dev` to `{1, 2}` before caling `GridFenceTestKernel`. Otherwise it will contain `{10, 20}` from the start (because of the previous `DeviceFenceTestKernelWriter`) which makes the check meaningless.

j-stephan (changes requested):

test/unit/mem/fence/src/FenceTest.cpp
fwyzard: Ah, yes, you are right.

fwyzard:
Superseded by the updates in #1641 .

PR #1644, opened 2022-03-15
Refactor `getWarpSize` to `getWarpSizes` and return `std::vector<std::size_t>`

### LINKED ISSUE #1595: `getWarpSize()` should return a vector
We currently assume that each device has exactly one supported warp size. This is true for CUDA GPUs and (to an extent) for AMD GPUs. On CPUs this is already questionable if we take vectorization into account but the current approach seems to work reasonably well.

As the SYCL back-end is nearing completion we can observe that this is no longer true: My laptop GPU (Ice Lake) reports that it can operate in sub-groups (the SYCL equivalent of warps) of size 8, 16 and 32. The exact size is chosen by the compiler depending on some heuristics but can also be enforced by the user through kernel attributes in C++ code (example: `[[sycl::reqd_sub_group_size(16)]]`). The SYCL runtime will claim all three are supported when queried.

`getWarpSize` should therefore be renamed to `getWarpSizes` and return a vector of supported sizes.

j-stephan:
Fixes #1595. This PR renames all instances of `getWarpSize` to `getWarpSizes` and always returns a `std::vector<std::size_t>` instead of a single `std::size_t`.

### DIFF
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -23,9 +23,11 @@
 #include <alpaka/wait/Traits.hpp>
 
 #include <algorithm>
+#include <cstddef>
 #include <map>
 #include <memory>
 #include <mutex>
+#include <string>
 #include <vector>
 
 namespace alpaka
@@ -127,7 +129,7 @@ namespace alpaka
             return 0;
         }
 
-    public:
+    private:
         std::shared_ptr<cpu::detail::DevCpuImpl> m_spDevCpuImpl;
     };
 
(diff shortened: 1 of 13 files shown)


fwyzard:
Mhm, I'm wondering, with the old interface did we have the possibility of using the warp size in a `constexpr` context ?

psychocoderHPC:
> Mhm, I'm wondering, with the old interface did we have the possibility of using the warp size in a `constexpr` context ?

Mostly not, warp size is for many backends only known at runtime, in cases you mix e.g. consumer AMD and professional AMD GPUs the warp size can differ for any of the GPUs too. 

A general issue with the alpaka interface to get the warp size is that we need a device. Therefore the device must be created. Alpaka in the backend is creating a stream to be sure that e.g. CUDA is creating the device context, in cases where the GPU is running in process exclusive mode the application can not select a device based on there supported warp size because in this example only one process can get the device and query the warp size. I will open a separate issue for this behavior.

sbastrakov (approved):

sbastrakov:
@fwyzard The interface being changed in this PR is for host-side query, so I doubt `constexpr` was very valuable there. For in-kernel query there is a different interface (not touched by this PR) `alpaka::warp::getSize(acc)`. That one is also not `constexpr`, and I do not remember why, maybe it could become compile-time. So in spirit host-side is like CUDA/HIP get device properties and query warp size from it, and the other is like a buildin variable `warpSize` (no idea about SYCL).

bernhardmgruber (approved):

PR #1650, opened 2022-03-16
Complete renaming: `alpaka_BUILD_EXAMPLES` -> `ALPAKA_BUILD_EXAMPLES`

j-stephan:
#1622 missed a few instances. This caused CMake warnings and (more importantly) the examples not being built during the CI runs.


psychocoderHPC:
Normally `alpaka_*` is correct and `ALPAKA_*`  is wrong. The project name is all small letters and therefore following the typical CMake naming convention alpaka should be small. see e.g. https://cmake.org/cmake/help/latest/module/FindBoost.html

psychocoderHPC:
I checked a few other projects in CMake. There is a wild mix but checking most modern one I would say the keep always the form of the project name + `_XXXX`

```
``Python3_LIBRARIES``
  The Python 3 libraries.
``Python3_LIBRARY_DIRS``
  The Python 3 library directories.
``Python3_RUNTIME_LIBRARY_DIRS``
  The Python 3 runtime library directories.
``Python3_VERSION``
  Python 3 version.
```

I do not say that @ax3l is doing it correctly but `openPMD-api` is doing it too https://github.com/openPMD/openPMD-api/blob/509cdede7bb1c31539a3428b9a5ed9f423058536/CMakeLists.txt#L88

The original upper case variable names are from the beginning of the alpaka project where we used CMake 2.X and the variable name style is IMO outdated.

bernhardmgruber:
I don't care how we name the cmake variables, but I would like them to be consistent, so they appear in one places in `ccmake` and are grouped correctly by `cmake-gui`.

bernhardmgruber (approved):

j-stephan:
I'll submit a follow-up PR which changes all CMake variables to the `alpaka_XXX` pattern.

PR #1651, opened 2022-03-16
Renamed "traits" namespace to "trait"

### LINKED ISSUE #179: rename traits to trait
All namespaces within alpaka are singular to shorten the names and make the code clearer. The only exception violating the coding guide lines is the `traits` namespace. Therefore I suggest to rename it to `trait` and remove thousands of `s` from the code base.
This would make the trait call sites look more natural because it would be e.g. `stream::trait::Enqueue` instead of `stream::traits::Enqueue`.

tonydp03:
Fixes #179. The Alpaka namespace "traits" was renamed to "trait" in all its occurrencies. However, are we sure we want to keep the namespace and not removing it entirely to simplify the code, as @bernhardmgruber suggested?

### DIFF
--- a/docs/source/basic/cheatsheet.rst
+++ b/docs/source/basic/cheatsheet.rst
@@ -263,7 +263,7 @@ Allocate static shared memory variable
 Get dynamic shared memory pool, requires the kernel to specialize
   .. code-block:: c++
 
-     traits::BlockSharedMemDynSizeBytes
+     trait::BlockSharedMemDynSizeBytes
        Type * dynamicSharedMemoryPool = getDynSharedMem<Type>(acc);
 
 Synchronize threads of the same block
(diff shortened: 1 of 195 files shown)


psychocoderHPC:
I am against removing the `trait` namespace. The namespace is separating the free functions the user is touching from the implementations alpaka developers or experts are providing. see https://alpaka-group.github.io/alpaka/namespaces.html

bernhardmgruber (dismissed):
I did not check every file, but I think trusting the CI is fine in this case.

bernhardmgruber:
The commit was made with @tonydp03 as author. @psychocoderHPC do you insist on commiting under the Tools author? And @j-stephan do you want @tonydp03 to update all the copyright headers? I don't care much about both of them.

psychocoderHPC (changes requested):
@bernhardmgruber mentioned it please commit your changes with

```
GIT_AUTHOR_NAME="Tools" GIT_AUTHOR_EMAIL="alpaka@hzdr.de" git commit ...
```

This PR is mostly processed with search and replace. We try to submit automated changes under an generic user.

j-stephan:
If we have the tools author I don't think we need to update the copyright header.

tonydp03:
I'll change the headers, rebase, and squash the new commit with the previous one under the Tools author.

bernhardmgruber (approved):

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #1652, opened 2022-03-17
Fix the host-only compilation for the HIP backend

### LINKED ISSUE #1647: Minor changes for building with HIP in host-only mode
Over the last week I've been working to make the alpaka port of our [`pixeltrack-standalone`](pixeltrack-standalone) application buildable with the HIP backend.

I've identified just a few changes that would be needed in alpaka -- but as I'm very not familiar with HIP and ROCm, I thought I'd ask for opinions before submitting a PR.

```diff
diff --git a/include/alpaka/core/CudaHipMath.hpp b/include/alpaka/core/CudaHipMath.hpp
index 68af040608b..9984410e370 100644
--- a/include/alpaka/core/CudaHipMath.hpp
+++ b/include/alpaka/core/CudaHipMath.hpp
@@ -15,6 +15,6 @@
 #    include <cuda_runtime.h>
 #endif
 
-#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED) && ! defined(ALPAKA_HOST_ONLY)
 #    include <hip/math_functions.h>
 #endif
```

This is required because `hip/math_functions.h` cannot be parsed by the host compiler (_e.g._ gcc).
The first error I get is about missing support for `_Float16` - but there might be more.
Would this be the right place for the check, or should I try to move up (down?) the include hierarchy, closer to the user's code ?

```diff
diff --git a/include/alpaka/core/Hip.hpp b/include/alpaka/core/Hip.hpp
index 9684994c07c..e60d46c4a09 100644
--- a/include/alpaka/core/Hip.hpp
+++ b/include/alpaka/core/Hip.hpp
@@ -11,13 +11,13 @@
 
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
 
-#    if !BOOST_LANG_HIP
+#    if !BOOST_LANG_HIP && ! defined(ALPAKA_HOST_ONLY)
 #        error If ALPAKA_ACC_GPU_HIP_ENABLED is set, the compiler has to support HIP!
 #    endif
 
 #    include <alpaka/core/CudaHipCommon.hpp>
 
-#    if BOOST_COMP_HIP
+#    ifndef HIPRT_CB
 #        define HIPRT_CB
 #    endif
 
```

The first change is required to make the compilation proceed with the host compiler.
The other possibility is to simply remove the whole check.
Any preference ?

The second change is needed to `#define` `HIPRT_CB` when using the host compiler.
It seems harmless enough... any comments, though ?

fwyzard:


fwyzard:
Fixes #1647

include/alpaka/core/CudaHipMath.hpp
```diff
@@ -15,6 +15,6 @@
 #    include <cuda_runtime.h>
 #endif
 
-#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED) && !defined(ALPAKA_HOST_ONLY)
```
psychocoderHPC: IMO instead of `!defined(ALPAKA_HOST_ONLY)`  it should be `defined(__HIP_DEVICE_COMPILE__)` to include hip math functions only within the device path.

In cupla https://github.com/alpaka-group/cupla/blob/287deacef75aac0833307e3d8aedabf42e4591ec/include/cupla/defines.hpp#L130  detecting the device compile is using an additional variable but maybe it is working with `__HIP_EVICE_COMPILE__` only.
@fwyzard could you try if this is working in your case?

psychocoderHPC (changes requested):

include/alpaka/core/CudaHipMath.hpp
fwyzard: Sure, I'll check if that works too.

fwyzard: Yes, using `defined(__HIP_DEVICE_COMPILE__)` instead of `!defined(ALPAKA_HOST_ONLY)` works.

psychocoderHPC (approved):

include/alpaka/core/CudaHipMath.hpp
```diff
@@ -15,6 +15,6 @@
 #    include <cuda_runtime.h>
 #endif
 
-#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#if defined(ALPAKA_ACC_GPU_HIP_ENABLED) && defined(__HIP_EVICE_COMPILE__)
```
j-stephan, line 18: I just noticed this typo while preparing the release notes:

`__HIP_EVICE_COMPILE__`

PR #1654, opened 2022-03-18
Enforce trivially copyable kernel function objects

### LINKED ISSUE #1384: kernel parameter should be trivially copy-able

### LINKED ISSUE #1634: [RFC] Remove support for `std::function` kernels.

j-stephan:
Fixes #1384 and #1634. Followup to #1635.

This PR enforces trivially copyable kernel function objects and therefore addresses the second half of #1384. It also removes support for `std::function` kernels (which was only partial anyway).


test/unit/kernel/src/KernelLambda.cpp
```diff
@@ -73,17 +73,10 @@ struct TestTemplateCapture
 
         alpaka::test::KernelExecutionFixture<TAcc> fixture(alpaka::Vec<Dim, Idx>::ones());
 
-        std::uint32_t const arg = 42u;
+        std::uint32_t arg = 42u;
```
bernhardmgruber, line 76: Why is removing `const` necessary?

j-stephan: To silence the clang warning that was previously guarded with macros. If you leave it `const` clang will complain about an unnecessary capture (because of some language legalese).

bernhardmgruber: I am surprised that this worked. I also don't think it is an elegant solution, because the `const` is rightfully there. But then, I also don't care so much, since it has no effect on the test :)

bernhardmgruber (approved):

PR #1658, opened 2022-03-22
Added CUDA/HIP support for missing types of atomic operations

### LINKED ISSUE #1355: On CUDA/HIP only AtomicAdd and AtomicExch support floating-point types
I know CUDA only supports those, but can alpaka implement e.g. `atomicSub` for these backends as calling the native `atomicAdd` with the negative value? For now it is a bit unintuitive, and also not portable between backends, as other atomic implementations seem to work for all types.

Was discovered when I looked at a PIConGPU compile issue with @BrianMarre .

tonydp03:
Resolves #1355 adding support for missing types of atomic operations for CUDA/HIP backend.

### DIFF
--- a/include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Benjamin Worpitz, René Widera, Jan Stephan, Andrea Bocci, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, René Widera, Jan Stephan, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di Pilato
  *
  * This file is part of alpaka.
  *
@@ -177,7 +177,7 @@ namespace alpaka
         {
             //
             __device__ static auto atomicOp(
-                AtomicUniformCudaHipBuiltIn const&,
+                AtomicUniformCudaHipBuiltIn const& ctx,
                 [[maybe_unused]] unsigned long int* const addr,
                 [[maybe_unused]] unsigned long int const& value) -> unsigned long int
             {
@@ -185,12 +185,45 @@ namespace alpaka
                 if constexpr(std::numeric_limits<unsigned int>::max() == std::numeric_limits<unsigned long>::max())
                     return ::atomicSub(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
                 else // LP64
-                    static_assert(
-                        !sizeof(THierarchy),
-                        "atomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported when "
-                        "sizeof(unsigned long int) == 4");
+                    return alpaka::atomicAdd(ctx, addr, -value);
 
                 ALPAKA_UNREACHABLE(0ul);
+            }
+        };
+        //! The GPU CUDA/HIP accelerator atomic operation.
+        template<typename THierarchy>
+        struct AtomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, unsigned long long int, THierarchy>
+        {
+            //
+            __device__ static auto atomicOp(
+                AtomicUniformCudaHipBuiltIn const&,
+                unsigned long long int* const addr,
+                unsigned long long int const& value) -> unsigned long long int
+            {
+                return ::atomicAdd(addr, -value);
+            }
+        };
+        //! The GPU CUDA/HIP accelerator atomic operation.
+        template<typename THierarchy>
+        struct AtomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, float, THierarchy>
+        {
+            //
+            __device__ static auto atomicOp(AtomicUniformCudaHipBuiltIn const&, float* const addr, float const& value)
+                -> float
+            {
+                return ::atomicAdd(addr, -value);
+            }
+        };
+        //! The GPU CUDA/HIP accelerator atomic operation.
+        template<typename THierarchy>
+        struct AtomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, double, THierarchy>
+        {
+            __device__ static auto atomicOp(
+                AtomicUniformCudaHipBuiltIn const& ctx,
+                double* const addr,
+                double const& value) -> double
+            {
+                return alpaka::atomicAdd(ctx, addr, -value);
             }
         };
 
@@ -241,8 +274,8 @@ namespace alpaka
 #        else
                 static_assert(
                     !sizeof(THierarchy),
-                    "atomicOp<AtomicMin, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported on sm >= "
-                    "3.5");
+                    "atomicOp<AtomicMin, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported on sm "
+                    ">= 3.5");
                 return 0ul;
 #        endif
             }
@@ -265,6 +298,48 @@ namespace alpaka
                     ">= 3.5");
                 return 0ull;
 #        endif
+            }
+        };
+        //! The GPU CUDA/HIPaccelerator atomic operation.
+        template<typename THierarchy>
+        struct AtomicOp<AtomicMin, AtomicUniformCudaHipBuiltIn, float, THierarchy>
+        {
+            __device__ static auto atomicOp(AtomicUniformCudaHipBuiltIn const&, float* const addr, float const& value)
+                -> float
+            {
+                int* address_as_i(reinterpret_cast<int*>(addr));
+                int old(*address_as_i);
+                int assumed;
+                do
+                {
+                    assumed = old;
+                    old = ::atomicCAS(address_as_i, assumed, __float_as_int(fminf(value, __int_as_float(assumed))));
+                } while(assumed != old);
+                return __int_as_float(old);
+            }
+        };
+        //! The GPU CUDA/HIPaccelerator atomic operation.
+        template<typename THierarchy>
+        struct AtomicOp<AtomicMin, AtomicUniformCudaHipBuiltIn, double, THierarchy>
+        {
+            __device__ static auto atomicOp(
+                AtomicUniformCudaHipBuiltIn const&,
+                double* const addr,
+                double const& value) -> double
+            {
+                unsigned long long int* address_as_ull(reinterpret_cast<unsigned long long int*>(addr));
+                unsigned long long int old(*address_as_ull);
+                unsigned long long int assumed;
+                do
+                {
+                    assumed = old;
+                    old = ::atomicCAS(
+                        address_as_ull,
+                        assumed,
+                        static_cast<unsigned long long int>(__double_as_longlong(
+                            fmin(value, __longlong_as_double(static_cast<long long int>(assumed))))));
+                } while(assumed != old);
+                return __longlong_as_double(static_cast<long long int>(old));
             }
         };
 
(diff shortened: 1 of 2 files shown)


tonydp03:
Should we also support other atomic operations for missing types through atomicCAS or any other implementation?

j-stephan:
> Should we also support other atomic operations for missing types through atomicCAS or any other implementation?

Yes, otherwise #1355 wouldn't be resolved :-) It just mentions `atomicSub` as an example.

tonydp03:
> > Should we also support other atomic operations for missing types through atomicCAS or any other implementation?
> 
> Yes, otherwise #1355 wouldn't be resolved :-) It just mentions `atomicSub` as an example.

Perfect, I wanted to be sure we agreed on implementing support for the other atomics as well :)

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -185,14 +185,70 @@ namespace alpaka
...
+                    assumed = old;
+                    old = ::atomicCAS(
+                        address_as_ull,
+                        assumed,
+                        static_cast<unsigned long long>(
+                            __double_as_longlong(-value + __longlong_as_double(static_cast<long long>(assumed)))));
+                    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
+                } while(assumed != old);
+                return __longlong_as_double(static_cast<long long>(old));
+#        endif
```
bernhardmgruber:
```suggestion
            return ::atomicAdd(addr, -value);
```
I don't think we need to duplicate the CAS workaround here. `atomicAdd` can do that.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -185,14 +185,70 @@ namespace alpaka
...
                     return ::atomicSub(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
                 else // LP64
-                    static_assert(
-                        !sizeof(THierarchy),
-                        "atomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported when "
-                        "sizeof(unsigned long int) == 4");
+                {
+                    return ::atomicAdd(
+                        reinterpret_cast<unsigned long long int*>(addr),
+                        static_cast<unsigned long long int>(-value));
```
bernhardmgruber:
```suggestion
                    return ::atomicAdd(addr, -value);
```
Why not stay with the same datatype when forwarding to `atomicAdd`?

bernhardmgruber (changes requested):

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
tonydp03: I tried but it gives me an error:
``` 
/data/user/adipilat/alpaka/include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp(230): error: no instance of overloaded function "atomicAdd" matches the argument list
            argument types are: (double *const, double)
```

I also tried to to that for atomicCAS when implementing support for `unsigned long int` (the native atomicCAS does not support the type, but there's an alpaka implementation that does it) instead of recasting types.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
tonydp03: I think I simply emulated the same behaviour as of atomic add, recasting depending on LLP64 or not. Indeed, as above, I got the same error...

bernhardmgruber: Ok, but if `atomicAdd` already does that for you, then just delegate to it :)

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
bernhardmgruber: Oh wait, it needs to be `atomicAdd`, not `::atomicAdd`. Or alternatively:
`AtomicOp<AtomicAdd, AtomicUniformCudaHipBuiltIn, double, THierarchy>::atomicOp(..., addr, -value);`

You would call the alpaka one, not the CUDA one directly.

tonydp03: `atomicAdd` gives the same error, although it's clearly the alpaka one being called 
```
/data/user/adipilat/alpaka/include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp(233): error: no instance of function template "alpaka::atomicAdd" matches the argument list
            argument types are: (double *const, double)
```

bernhardmgruber: Because you are missing the first argument that alpaka's version needs. Here is the complete version:
```c++
        template<typename THierarchy>
        struct AtomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, double, THierarchy>
        {
            __device__ static auto atomicOp(
                AtomicUniformCudaHipBuiltIn const& ctx,
                double* const addr,
                double const& value) -> double
            {
                atomicAdd(ctx, addr, -value);
            }
        };
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -177,22 +177,51 @@ namespace alpaka
...
             {
-                // LLP64
-                if constexpr(std::numeric_limits<unsigned int>::max() == std::numeric_limits<unsigned long>::max())
-                    return ::atomicSub(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
-                else // LP64
-                    static_assert(
-                        !sizeof(THierarchy),
-                        "atomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, unsigned long int> is only supported when "
-                        "sizeof(unsigned long int) == 4");
+                return atomicAdd(ctx, addr, -value); // Calls alpaka::atomicAdd function
```
bernhardmgruber:
```suggestion
                return alpaka::atomicAdd(ctx, addr, -value);
```
If you want to insist on which function is called, you can also spell it out then :)

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -177,22 +177,51 @@ namespace alpaka
...
+        //! The GPU CUDA/HIP accelerator atomic operation.
+        template<typename THierarchy>
+        struct AtomicOp<AtomicSub, AtomicUniformCudaHipBuiltIn, double, THierarchy>
+        {
+            __device__ static auto atomicOp(
+                AtomicUniformCudaHipBuiltIn const& ctx,
+                double* const addr,
+                double const& value) -> double
+            {
+                return atomicAdd(ctx, addr, -value); // Calls alpaka::atomicAdd function
```
bernhardmgruber:
```suggestion
                return alpaka::atomicAdd(ctx, addr, -value);
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -409,9 +526,52 @@ namespace alpaka
...
+                    assumed = old;
+                    old = ::atomicCAS(
+                        address_as_ull,
+                        assumed,
+                        static_cast<unsigned long long>(__double_as_longlong(value)));
+                    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
+                } while(assumed != old);
+                return __longlong_as_double(static_cast<long long>(old));
+            }
+        };
```
bernhardmgruber: Atomic exchange is rather simple. Why do you need a CAS loop for that?

```suggestion
        template<typename THierarchy>
        struct AtomicOp<AtomicExch, AtomicUniformCudaHipBuiltIn, double, THierarchy>
        {
            __device__ static auto atomicOp(
                AtomicUniformCudaHipBuiltIn const&,
                double* const addr,
                double const& value) -> double
            {
                return __longlong_as_double(static_cast<long long>(
                    ::atomicExch(reinterpret_cast<unsigned long long*>(addr),
                    static_cast<unsigned long long>(__double_as_longlong(value))));
            }
        };
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -409,9 +526,52 @@ namespace alpaka
...
+                -> int
+            {
+                int old(*addr);
+                int assumed, eval;
+                do
+                {
+                    assumed = old;
+                    eval = assumed >= value ? 0 : assumed + 1;
+                    old = ::atomicCAS(addr, assumed, eval);
+                    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
bernhardmgruber:
```suggestion
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -438,17 +598,66 @@ namespace alpaka
...
+                {
+                    unsigned long long int* address_as_u = reinterpret_cast<unsigned long long int*>(addr);
+                    unsigned long long int old(*address_as_u);
+                    unsigned long long int assumed, eval;
+                    do
+                    {
+                        assumed = old;
+                        eval = assumed >= static_cast<unsigned long long int>(value) ? 0 : assumed + 1;
+                        old = ::atomicCAS(address_as_u, assumed, eval);
+                        // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
bernhardmgruber:
```suggestion
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -438,17 +598,66 @@ namespace alpaka
...
+                unsigned long long int const& value) -> unsigned long long int
+            {
+                unsigned long long int old(*addr);
+                unsigned long long int assumed, eval;
+                do
+                {
+                    assumed = old;
+                    eval = assumed >= value ? 0 : assumed + 1;
+                    old = ::atomicCAS(addr, assumed, eval);
+                    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
bernhardmgruber:
```suggestion
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -438,17 +598,66 @@ namespace alpaka
...
+                -> int
+            {
+                int old(*addr);
+                int assumed, eval;
+                do
+                {
+                    assumed = old;
+                    eval = assumed == 0 || assumed > value ? value : assumed - 1;
+                    old = ::atomicCAS(addr, assumed, eval);
+                    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
bernhardmgruber:
```suggestion
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -475,14 +684,46 @@ namespace alpaka
...
+                    unsigned long long int old(*address_as_u);
+                    unsigned long long int assumed, eval;
+                    do
+                    {
+                        assumed = old;
+                        eval = assumed == 0 || assumed > static_cast<unsigned long long int>(value)
+                            ? static_cast<unsigned long long int>(value)
+                            : assumed - 1;
+                        old = ::atomicCAS(address_as_u, assumed, eval);
+                        // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
bernhardmgruber:
```suggestion
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -475,14 +684,46 @@ namespace alpaka
...
+                unsigned long long int const& value) -> unsigned long long int
+            {
+                unsigned long long int old(*addr);
+                unsigned long long int assumed, eval;
+                do
+                {
+                    assumed = old;
+                    eval = assumed == 0 || assumed > value ? value : assumed - 1;
+                    old = ::atomicCAS(addr, assumed, eval);
+                    // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
bernhardmgruber:
```suggestion
```

bernhardmgruber (changes requested):

bernhardmgruber (dismissed):

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -177,22 +177,51 @@ namespace alpaka
...
             //
             __device__ static auto atomicOp(
-                AtomicUniformCudaHipBuiltIn const&,
+                AtomicUniformCudaHipBuiltIn const& ctx,
                 [[maybe_unused]] unsigned long int* const addr,
                 [[maybe_unused]] unsigned long int const& value) -> unsigned long int
             {
-                // LLP64
-                if constexpr(std::numeric_limits<unsigned int>::max() == std::numeric_limits<unsigned long>::max())
-                    return ::atomicSub(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
```
j-stephan: I don't think we should remove the `LLP64` case when it is available natively.

j-stephan (changes requested):

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
tonydp03: So just calling `alpaka::atomicAdd` for the non `LLP64` case and leave `atomicSub` for `LLP64`?

j-stephan: Yes, that is better IMO.

bernhardmgruber (approved):

j-stephan (approved):

PR #1662, opened 2022-03-24
Implement separate types for the CUDA and HIP/ROCm backends

fwyzard:
The `UniformCudaHip` types are moved to a nested namespace, which is defined as either `cuda` or `hip` according to the backend being compiled.

Users should use the explicit backend types like `DevHipRt`, `AccGpuHipRt`, ... and `DevCudaRt`, `AccGpuCudaRt`, ... from the `alpaka` namespace.


fwyzard:
#1665 seems like a better approach

PR #1665, opened 2022-03-25
Implement separate types for the CUDA and HIP/ROCm backends

fwyzard:
The `UniformCudaHip` types are now templated on a new traits-like `struct` that encapsulates the CUDA or HIP API, depending on the backend being used.

Users should use the explicit backend types like `DevHipRt`, `AccGpuHipRt`, ... and `DevCudaRt`, `AccGpuCudaRt`, ...


fwyzard:
The header files and namespaces could probably be arranged better.
For example:
  - move all `*UniformCudaHipRt*.hpp` files to a `.../cuda_hip/` subdirectory
  - move all `*UniformCudaHipRt*` types to the `uniform_cuda_hip` or `detail` namespace

fwyzard:
The random number generators are not yet ported to the new interface -- hence many of the failures.

include/alpaka/acc/AccGpuCudaRt.hpp
```diff
@@ -11,79 +11,13 @@
...
-            {
-                return TaskKernelGpuUniformCudaHipRt<AccGpuCudaRt<TDim, TIdx>, TDim, TIdx, TKernelFnObj, TArgs...>(
-                    workDiv,
-                    kernelFnObj,
-                    std::forward<TArgs>(args)...);
-            }
-        };
-    } // namespace trait
-} // namespace alpaka
+    using AccGpuCudaRt = AccGpuUniformCudaHipRt<ApiCudaRt, TDim, TIdx>;
```
psychocoderHPC, line 20: We started to discuss this line [here](https://github.com/fwyzard/alpaka/commit/34f251aec7be8565a081aecc872467d92e0182af#r69328577) already.

summery: removing the struct would give the wrong naming via `GetAccName` and type via `AccType`

psychocoderHPC: @fwyzard I suggest to keep the implementation with the struct instead of the alias to keep the separation of the HIP and CUDA accelerator.

psychocoderHPC (changes requested):

include/alpaka/acc/AccGpuCudaRt.hpp
fwyzard: > removing the struct would give the wrong naming via `GetAccName` and type via `AccType`

Actually it doesn't, because `GetAccName` returns a name based on `API::name`:
```c++
        template<typename API, typename TDim, typename TIdx>
        struct GetAccName<AccGpuUniformCudaHipRt<API, TDim, TIdx>>
        {
            {
                return "AccGpu" + API::name + "Rt<" + std::to_string(TDim::value) + "," + typeid(TIdx).name() + ">";
            }
        };
 ```

And `AccType` should be `AccGpuUniformCudaHipRt<API, ...>` which is the same type as either `AccGpuCudaRt<...>` or `AccGpuHipRt<...>`.

**Comment by fwyzard** [comment]
Rebased, and fixed all tests failing locally with gcc and cuda.

**Comment by fwyzard** [comment]
@psychocoderHPC @bernhardmgruber I'm not sure how to handle this warning treated as error:
```
In file included from /home/runner/work/alpaka/alpaka/build/test/analysis/headerCheck/src/acc/AccGpuCudaRt.hpp.cpp:1:
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/acc/AccGpuCudaRt.hpp:15:
/home/runner/work/alpaka/alpaka/include/alpaka/core/ApiCudaRt.hpp:16:41: error: declaration requires an exit-time destructor [-Werror,-Wexit-time-destructors]
        static inline const std::string name = "Cuda";
                                        ^
```

GCC does not complain abut it... if it is an issue I can certainly work around it, but I'd like to hear your opinion about it.

**Comment by fwyzard** [comment]
OK, I'll just change the `std::string` to a `const char[]` buffer.

**Comment by bernhardmgruber** [comment]
> OK, I'll just change the `std::string` to a `const char[]` buffer.

`std::string_view` should also work.

**Comment by fwyzard** [comment]
As discussed this morning, a test for this functionality is available at https://github.com/fwyzard/test-rocm-cuda-alpaka .

With the current `develop` branch (actually with #1678 to make the explicit types available), the test fails with
```
ROCm runtime initialised
CUDA runtime initialised
CPU-only serial runtime initialised

device: Radeon Pro WX 9100
accelerator: AccGpuHipRt<1,j>
result: 10
ROCm kernel ran successfully

device: NVIDIA GeForce GTX 1080 Ti
accelerator: AccGpuCudaRt<1,j>
result: 10
  what():  alpaka/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp(249) 'hipFree(ptr)' returned error  : 'hipErrorInvalidValue': 'hipErrorInvalidValue'!
Aborted (core dumped)
```

With this #1665 (this PR), the test passes with
```
ROCm runtime initialised
CUDA runtime initialised
CPU-only serial runtime initialised

device: Radeon Pro WX 9100
accelerator: AccGpuHipRt<1,j>
result: 10
ROCm kernel ran successfully

device: NVIDIA GeForce GTX 1080 Ti
accelerator: AccGpuCudaRt<1,j>
result: 10
CUDA kernel ran successfully

device: AMD Ryzen 9 5900X 12-Core Processor            
accelerator: AccCpuSerial<1,j>
result: 10
CPU-only serial kernel ran successfully
```

File: include/alpaka/acc/AccGpuUniformCudaHipRt.hpp [hunk 2]
```diff
@@ -44,13 +44,14 @@
...
 {
-    template<typename TAcc, typename TDim, typename TIdx, typename TKernelFnObj, typename... TArgs>
+    template<typename API, typename TAcc, typename TDim, typename TIdx, typename TKernelFnObj, typename... TArgs>
     class TaskKernelGpuUniformCudaHipRt;
 
     //! The GPU CUDA accelerator.
     //!
     //! This accelerator allows parallel kernel execution on devices supporting CUDA.
     template<
+        typename API,
```
bernhardmgruber: Should we keep the name `API` or name it `Api` or `TApi`?

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,357 @@
...
+        using Stream_t = ::cudaStream_t;
+        using Pos_t = ::cudaPos;
+        using Extent_t = ::cudaExtent;
+        using FuncAttributes_t = ::cudaFuncAttributes;
+        using MemcpyKind_t = ::cudaMemcpyKind;
+        using Memcpy3DParms_t = ::cudaMemcpy3DParms;
+        using PitchedPtr_t = ::cudaPitchedPtr;
+
+        // constants
+        static constexpr auto success = ::cudaSuccess;
```
bernhardmgruber:
```suggestion
        inline static constexpr auto success = ::cudaSuccess;
```
This only makes a difference if the constant is ODR-used anywhere (e.g. its address is taken). I would add it though in case someone would like to do that. Applies to all constants in this struct and in `ApiHipRt`.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,357 @@
...
+        {
+            return ::cudaDeviceGetAttribute(value, attr, device);
+        }
+
+        static inline cudaError_t deviceGetLimit(size_t* pValue, enum cudaLimit limit)
+        {
+            return ::cudaDeviceGetLimit(pValue, limit);
+        }
+
+        static inline cudaError_t deviceReset(void)
```
bernhardmgruber:
```suggestion
        static inline cudaError_t deviceReset()
```
I personally dislike this relic from C. But this is not blocking this PR.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,357 @@
...
+        {
+            return ::cudaDeviceReset();
+        }
+
+        static inline cudaError_t deviceSetLimit(enum cudaLimit limit, size_t value)
```
bernhardmgruber:
```suggestion
        static inline cudaError_t deviceSetLimit(cudaLimit limit, size_t value)
```

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,357 @@
...
+#    else
+            // Unused.
+            static_cast<void>(devPtr);
+            static_cast<void>(hStream);
+            // Not implemented.
+            return ::cudaErrorUnknown;
+#    endif
+        }
+
+        static inline cudaError_t funcGetAttributes(struct cudaFuncAttributes* attr, const void* func)
```
bernhardmgruber:
```suggestion
        static inline cudaError_t funcGetAttributes(cudaFuncAttributes* attr, const void* func)
```
We don't need the elaborated type specifier here. Applies to other places as well.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,357 @@
...
+            return ::cudaStreamSynchronize(stream);
+        }
+
+        static inline cudaError_t streamWaitEvent(cudaStream_t stream, cudaEvent_t event, unsigned int flags)
+        {
+            return ::cudaStreamWaitEvent(stream, event, flags);
+        }
+
+        // device API
+        static inline struct cudaPitchedPtr makePitchedPtr(void* d, size_t p, size_t xsz, size_t ysz)
```
bernhardmgruber: If this is a device function, don't we need `__device__`?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -289,39 +298,40 @@ namespace alpaka
...
 #        endif
                 auto const gridBlockExtent = getWorkDiv<Grid, Blocks>(task);
                 auto const blockThreadExtent = getWorkDiv<Block, Threads>(task);
                 auto const threadElemExtent = getWorkDiv<Thread, Elems>(task);
 
-#        if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+#        if 1
                 dim3 const gridDim = uniform_cuda_hip::detail::convertVecToUniformCudaHipDim(gridBlockExtent);
                 dim3 const blockDim = uniform_cuda_hip::detail::convertVecToUniformCudaHipDim(blockThreadExtent);
 #        else
```
bernhardmgruber: Do we even need the #else branch anymore then?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -364,16 +374,18 @@ namespace alpaka
...
+                    TIdx,
+                    TKernelFnObj,
+                    remove_restrict_t<std::decay_t<TArgs>>...>;
 #        if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
-                // hipFuncAttributes not ported from HIP to HIP.
-                // TODO why this is currently not possible
-                //
+#            if 1
                 // Log the function attributes.
-#            if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
```
bernhardmgruber: Can we remove the preprocessor directives completely?

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -15,8 +15,10 @@
 // Backend specific includes.
 #    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
 #        include <alpaka/core/Cuda.hpp>
+struct ApiCudaRt;
 #    else
 #        include <alpaka/core/Hip.hpp>
+struct ApiHipRt;
 #    endif
```
bernhardmgruber: But if both backends are enabled, you would need both includes?
```suggestion
#        include <alpaka/core/Cuda.hpp>
#        include <alpaka/core/Hip.hpp>
struct ApiCudaRt;
struct ApiHipRt;
```

include/alpaka/mem/buf/uniformCudaHip/Set.hpp
```diff
@@ -84,20 +85,25 @@ namespace alpaka
                     Dim<TView>::value == Dim<TExtent>::value,
                     "The destination buffer and the extent are required to have the same dimensionality!");
 
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(ALPAKA_API_PREFIX(MemsetAsync)(
-                    getPtrNative(this->m_view),
+                auto& view = this->m_view;
+                ALPAKA_ASSERT(1 <= getWidth(view));
```
bernhardmgruber: Why did you readd this check? The `view` is zero dimensional, it must have a width of 1.

include/alpaka/pltf/PltfUniformCudaHipRt.hpp
```diff
@@ -18,8 +18,11 @@
 // Backend specific includes.
 #    if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
 #        include <alpaka/core/Cuda.hpp>
-#    else
+struct ApiCudaRt;
+#    endif
+#    if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
 #        include <alpaka/core/Hip.hpp>
+struct ApiHipRt;
 #    endif
```
bernhardmgruber: Just include both headers. No need for the `#if`.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -10,6 +10,7 @@
 
 #pragma once
 
+
```
bernhardmgruber:
```suggestion
```

include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
fwyzard: Anything works for me.

include/alpaka/core/ApiCudaRt.hpp
fwyzard: Right... I copied the signatures from the CUDA header, but you are right they can be simplified for the C++-only case (like Alpaka).

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: Yes, that's definitely a leftover from some test !

include/alpaka/core/ApiCudaRt.hpp
fwyzard: According to [cppreference.com](https://en.cppreference.com/w/cpp/language/constexpr)
> A `constexpr` specifier used in a function or `static` data member (since C++17) declaration implies `inline`.

so the explicit `inline` should not be necessary.

I can still add it if you prefer, though.

include/alpaka/core/ApiCudaRt.hpp
fwyzard: No, they are actually host functions - I don't remember where I copied the (wrong) comment from.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: Nope.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: Yes.

include/alpaka/mem/buf/uniformCudaHip/Set.hpp
fwyzard: I had to rebase my changes on top of the latest developments, it may have slipped in while resolving the conflicts.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
fwyzard: Yes.
Also, the forward declarations should be in the `alpaka` namespace - I'll fix them.

fwyzard: Well, it's actually not possible to enable both backends at the same time, because of the conflicts between `cuda_runtime.h` and `hip/hip_runtime.h` :-/

include/alpaka/core/ApiCudaRt.hpp
bernhardmgruber: Cool! I learned sth! You can omit the `inline` then.

fwyzard:
@bernhardmgruber I've rebased the branch and addressed all simple comments.

The only one left is whether to use `API` or `Api` or `TApi` as the template parameter name (I'm fine with any option, just let me know what you all prefer).

bernhardmgruber:
> @bernhardmgruber I've rebased the branch and addressed all simple comments.
> 
> The only one left is whether to use `API` or `Api` or `TApi` as the template parameter name (I'm fine with any option, just let me know what you all prefer).

OK, let's go with `TApi` then.
The rest of the PR LGTM.

fwyzard:
Latest set of changes:
  - renamed `API` to `TApi` everywhere
  - changed the two "policy" structs to use the generic types (_e.g._ `Stream_t`) instead of the CUDA/HIP ones (_e.g._ `cudaStream_t`/`hipStream_t`) or just `auto`
  - rebased on top of daa516b5fcfd110bf1ba97a4bc3a4d2b9af1ce3b

fwyzard:
8d0d179a6af8d258eb31d075965d6879a13a7869 is tricker to reimplement... ideally I would like to wrap the difference between `cudaLaunchHostFunc` and `cudaStreamAddCallback` directly in `ApiCudaRt.hpp` -- but I'm not yet sure how to do it.

fwyzard:
The last commit ec07d47d1512ea780042893167adbaaf7fdcb373 reimplements the changes from 8d0d179a6af8d258eb31d075965d6879a13a7869 with a somewhat different approach.

All differences between CUDA versions and between CUDA and HIP are encapsulated in `ApiCudaRt.hpp` and `ApiHipRt.hpp`:
  - CUDA versions < 10.0 use `cudaStreamAddCallback`
  - CUDA versions >= 10.0 use `cudaLaunchHostFunc`
  - all HIP versions so far use `hipStreamAddCallback`
  - CUDA uses the `__stdcall` calling convention on Windows (CUDART_CB)
  - HIP uses the standard calling convention on all platforms

The interface exposed by both APIs is similar to that of `cudaLaunchHostFunc`, without the use of a special calling convention.

fwyzard:
@j-stephan during the VC yesterday we agreed that this PR should target 0.10.0/1.0.0, as it is too invasive for 0.9.0 .

bernhardmgruber (dismissed):

j-stephan (dismissed):
Thanks for the PR. I want to push out the RC for 0.9.0 this afternoon so I'm locking the merge button until then.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,377 @@
...
+        {
+#    if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)
+            return ::cudaFreeAsync(devPtr, stream);
+#    else
+            // Unused.
+            static_cast<void>(devPtr);
+            static_cast<void>(stream);
+            // Not implemented.
+            return errorUnknown;
+#    endif
```
j-stephan:
```suggestion
        static inline Error_t freeAsync([[maybe_unused]] void* devPtr, [[maybe_unused]] Stream_t stream)
        {
#    if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)
            return ::cudaFreeAsync(devPtr, stream);
#    else
            // Not implemented.
            return errorUnknown;
#    endif
```

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,377 @@
...
+            return ::cudaMallocAsync(devPtr, size, stream);
+#    else
+            // Unused.
+            static_cast<void>(devPtr);
+            static_cast<void>(size);
+            static_cast<void>(stream);
+            // Not implemented.
+            return errorUnknown;
+#    endif
+        }
```
j-stephan:
```suggestion
        static inline Error_t mallocAsync([[maybe_unused]] void** devPtr, [[maybe_unused]] size_t size, [[maybe_unused]] Stream_t stream)
        {
#    if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)
            return ::cudaMallocAsync(devPtr, size, stream);
#    else
            // Not implemented.
            return errorUnknown;
#    endif
        }
```

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -0,0 +1,373 @@
...
+        }
+
+        static inline Error_t deviceSetLimit(Limit_t limit, size_t value)
+        {
+            // Unused.
+            static_cast<void>(limit);
+            static_cast<void>(value);
+            // Not implemented.
+            return errorUnknown;
+        }
```
j-stephan:
```suggestion
        static inline Error_t deviceSetLimit(Limit_t, size_t)
        {
            // Not implemented.
            return errorUnknown;
        }
```

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -0,0 +1,373 @@
...
+
+        static inline Error_t freeAsync(void* devPtr, Stream_t stream)
+        {
+            // Unused.
+            static_cast<void>(devPtr);
+            static_cast<void>(stream);
+            // Not implemented.
+            return errorUnknown;
+        }
```
j-stephan:
```suggestion
        static inline Error_t freeAsync(void*, Stream_t)
        {
            // Not implemented.
            return errorUnknown;
        }
```

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -0,0 +1,373 @@
...
+
+        static inline Error_t mallocAsync(void** devPtr, size_t size, Stream_t stream)
+        {
+            // Unused.
+            static_cast<void>(devPtr);
+            static_cast<void>(size);
+            static_cast<void>(stream);
+            // Not implemented.
+            return errorUnknown;
+        }
```
j-stephan:
```suggestion
        static inline Error_t mallocAsync(void**, size_t, Stream_t)
        {
            // Not implemented.
            return errorUnknown;
        }
```

j-stephan (changes requested):
Thanks! A few minor suggestions:

include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
j-stephan: `TApi` is the way.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,377 @@
...
+#    if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)
+            return ::cudaFreeAsync(devPtr, stream);
+#    else
+            // Unused.
+            static_cast<void>(devPtr);
+            static_cast<void>(stream);
+            // Not implemented.
+            return errorUnknown;
+#    endif
+        }
```
fwyzard: I see... how would you see this approach instead ?
```suggestion
#    if BOOST_LANG_CUDA >= BOOST_VERSION_NUMBER(11, 2, 0)
        static inline Error_t freeAsync(void* devPtr, Stream_t stream)
        {
            return ::cudaFreeAsync(devPtr, stream);
        }
#    else
        static inline Error_t freeAsync(void*, Stream_t)
        {
            // Not implemented.
            return errorUnknown;
        }
#    endif
```

fwyzard: @j-stephan

j-stephan: Yes, that would work. Is there a reason you don't want to use `[[maybe_unused]]`?

fwyzard: In general I'm fine with using `[[maybe_unused]]`, for example when the conditions when a parameter is used or not are not obvious.

In this specific case we know when the parameters are used or not, so I though this approach would make more sense.

But, all in all, I don't mind much, and I'm fine using whatever you prefer.

j-stephan: > In this specific case we know when the parameters are used or not, so I though this approach would make more sense.

Sure, but `[[maybe_unused]]` is a compiler hint ;-) We also know this when using `if constexpr`, for example. For that reason I prefer `[[maybe_unused]]`.

fwyzard: OK.

fwyzard:
Rebased on top of `develop`, and addressed the last round of comments.

include/alpaka/core/ApiHipRt.hpp
fwyzard: I've left the parameter names in `/* comments */`, to keep a reminder of what they are supposed to be, when the functions will eventually be supported by HIP.

j-stephan (dismissed):

bernhardmgruber (dismissed):

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -0,0 +1,374 @@
+#pragma once
```
psychocoderHPC, line 10: Please add the license header to this file:

```
/* Copyright 2022 Andrea Bocci
 *
 * This file is part of alpaka.
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 */
```

We add the license to each header, in cases someone is coping one file out of alpaka the license is still clear.

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -0,0 +1,372 @@
+#pragma once
```
psychocoderHPC, line 10: Please add the license header to this file:

```
/* Copyright 2022 Andrea Bocci
 *
 * This file is part of alpaka.
 *
 * This Source Code Form is subject to the terms of the Mozilla Public
 * License, v. 2.0. If a copy of the MPL was not distributed with this
 */
```

psychocoderHPC (changes requested):
I found only two files without a license. All other code looks fine.

Thanks for this nice PR!

include/alpaka/core/ApiHipRt.hpp
fwyzard: :+1:

include/alpaka/core/ApiCudaRt.hpp
fwyzard: :+1:

bernhardmgruber (approved):

psychocoderHPC (approved):

fwyzard:
:+1:

PR #1667, opened 2022-03-26
Refactor CUDA/HIP queues to reduce code duplication

fwyzard:


psychocoderHPC:
#1669 must be merged to fix the documentation build issue.

j-stephan (dismissed):

fwyzard:
can you force just the [linux_hip4.2_clang_debug_hip_only](https://gitlab.com/hzdr/crp/alpaka/-/jobs/2258662047) runner to re-run, or should I force-push the branch to re-run all tests ?

fwyzard:
or is there a real problem with these changes with HIP 4.2 ?

j-stephan:
No, unfortunately we encounter these timeouts quite often. According to our IT department the AMD GPU driver doesn't take too kindly to memory access violations of *other* programs on the same GPU. Since we have to share these resources with the rest of our institute this is an unfortunate but regular occurrence. In the best case it is enough to restart the failing job (which I just did), in the worst case the whole node has to be rebooted.

fwyzard:
> No, unfortunately we encounter these timeouts quite often. According to our IT department the AMD GPU driver doesn't take too kindly to memory access violations of _other_ programs on the same GPU.

I see... and thanks for the explanation, that's a useful piece of information if in the future we find some AMD GPUs to run on.

include/alpaka/queue/QueueUniformCudaHipRtBlocking.hpp
```diff
@@ -11,174 +12,19 @@
...
-        ALPAKA_FN_HOST auto operator==(QueueUniformCudaHipRtBlocking const& rhs) const -> bool
-        {
-            return (m_spQueueImpl == rhs.m_spQueueImpl);
-        }
-        ALPAKA_FN_HOST auto operator!=(QueueUniformCudaHipRtBlocking const& rhs) const -> bool
-        {
-            return !((*this) == rhs);
-        }
-    };
+    using QueueUniformCudaHipRtBlocking = uniform_cuda_hip::detail::QueueUniformCudaHipRtBase<true>;
```
bernhardmgruber: Do we need this file anyway now? We could just move that one line into `QueueUniformCudaHipRtBase.hpp` and delete the file. Same for the non-blocking version.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
```diff
@@ -25,111 +27,132 @@
...
-                , public concepts::Implements<ConceptGetDev, QueueUniformCudaHipRtBase>
+        //! The CUDA/HIP RT queue.
+        template<bool TBlocking>
+        class QueueUniformCudaHipRtBase final
+            : public concepts::Implements<ConceptCurrentThreadWaitFor, QueueUniformCudaHipRtBase<TBlocking>>
+            , public concepts::Implements<ConceptQueue, QueueUniformCudaHipRtBase<TBlocking>>
+            , public concepts::Implements<ConceptGetDev, QueueUniformCudaHipRtBase<TBlocking>>
+        {
+        public:
+            ALPAKA_FN_HOST QueueUniformCudaHipRtBase<TBlocking>(DevUniformCudaHipRt const& dev)
```
bernhardmgruber:
```suggestion
            ALPAKA_FN_HOST QueueUniformCudaHipRtBase(DevUniformCudaHipRt const& dev)
```
Providing template arguments should not be necessary on constructors. Same for the equality operators.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
```diff
@@ -25,111 +27,132 @@
...
+        };
 
-            //! The CUDA RT blocking queue.
-            class QueueUniformCudaHipRtBase
-                : public concepts::Implements<ConceptCurrentThreadWaitFor, QueueUniformCudaHipRtBase>
-                , public concepts::Implements<ConceptQueue, QueueUniformCudaHipRtBase>
-                , public concepts::Implements<ConceptGetDev, QueueUniformCudaHipRtBase>
+        //! The CUDA/HIP RT queue.
+        template<bool TBlocking>
+        class QueueUniformCudaHipRtBase final
```
bernhardmgruber:
```suggestion
        class QueueUniformCudaHipRt final
```
Why do we need the `Base` in the name? It's not a base class.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
```diff
@@ -142,22 +165,133 @@ namespace alpaka
...
+                enqueued,
+                notified,
+                finished,
+            };
+
+            struct CallbackSynchronizationData : public std::enable_shared_from_this<CallbackSynchronizationData>
+            {
+                std::mutex m_mutex;
+                std::condition_variable m_event;
+                CallbackState state = CallbackState::enqueued;
```
bernhardmgruber: Please either use the `m_` prefix or don't use it. I prefer to not have it here.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
```diff
@@ -142,22 +165,133 @@ namespace alpaka
...
+                    [pCallbackSynchronizationData, task]()
+                    {
+                        // If the callback has not yet been called, we wait for it.
+                        {
+                            std::unique_lock<std::mutex> lock(pCallbackSynchronizationData->m_mutex);
+                            if(pCallbackSynchronizationData->state != CallbackState::notified)
+                            {
+                                pCallbackSynchronizationData->m_event.wait(
+                                    lock,
+                                    [pCallbackSynchronizationData]()
```
bernhardmgruber:
```suggestion
                                    [&pCallbackSynchronizationData]
```

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
```diff
@@ -142,22 +165,133 @@ namespace alpaka
...
+                }
+                pCallbackSynchronizationData->m_event.notify_one();
+
+                // Wait for the executing thread to finish the task if it has not already finished.
+                std::unique_lock<std::mutex> lock(pCallbackSynchronizationData->m_mutex);
+                if(pCallbackSynchronizationData->state != CallbackState::finished)
+                {
+                    pCallbackSynchronizationData->m_event.wait(
+                        lock,
+                        [pCallbackSynchronizationData]()
```
bernhardmgruber:
```suggestion
                        [&pCallbackSynchronizationData]
```

bernhardmgruber (commented):
Generally LGTM. Here are some comments:

include/alpaka/queue/QueueUniformCudaHipRtBlocking.hpp
fwyzard: Since it defines the user-facing type, I thought people might be using the file explicitly.
If that's not supposed to happen then, sure, I would move everything into a single file, and maybe rename it simply `QueueUniformCudaHipRt.hpp`.

j-stephan: FWIW, we also follow the one-file-per-user-facing-typedef approach in the SYCL back-end. So this is nothing new.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
fwyzard: I didn't change it from what it was like before...
Anyway, what is the preferred naming scheme ?
I'd rather not have simply `std::mutex mutex` right above.

j-stephan: We haven't yet defined this in our style guide, but our work-in-progress clang-tidy file mandates the `m_` prefix.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
fwyzard: I didn't write this - but `pCallbackSynchronizationData` is a pointer, so capturing by value seems reasonable.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
fwyzard: Right, will fix.

include/alpaka/queue/QueueUniformCudaHipRtBlocking.hpp
fwyzard: OK, I'll keep the files then.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
bernhardmgruber: According to our [.clang-tidy](https://github.com/ComputationalRadiationPhysics/contributing/blob/master/formatting-tools/.clang-tidy) we have a leading `m_`. But I also remember a discussion that we don't need to for aggregates. @j-stephan?

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
fwyzard: Good point, I'll rename it.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
fwyzard: `m_state` it is, then.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
fwyzard: Same as above.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
bernhardmgruber: IIUC, it is a shared pointer here, so we can safe the extra copy.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRtBase.hpp
bernhardmgruber: It is a raw pointer here, so we can indeed stay with a by-value capture.

j-stephan:
To reduce CI pressure I have cancelled the CI run for the intermediate push @fwyzard. Don't worry if you receive a cancellation notice per mail.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -0,0 +1,298 @@
...
+                pCallbackSynchronizationData->m_event.notify_one();
+
+                // Wait for the executing thread to finish the task if it has not already finished.
+                std::unique_lock<std::mutex> lock(pCallbackSynchronizationData->m_mutex);
+                if(pCallbackSynchronizationData->m_state != CallbackState::finished)
+                {
+                    pCallbackSynchronizationData->m_event.wait(
+                        lock,
+                        [pCallbackSynchronizationData]()
+                        { return pCallbackSynchronizationData->m_state == CallbackState::finished; });
```
bernhardmgruber, line 232:
```suggestion
                        [pCallbackSynchronizationData]
                        { return pCallbackSynchronizationData->m_state == CallbackState::finished; });
```

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -0,0 +1,298 @@
...
+                    {
+                        // If the callback has not yet been called, we wait for it.
+                        {
+                            std::unique_lock<std::mutex> lock(pCallbackSynchronizationData->m_mutex);
+                            if(pCallbackSynchronizationData->m_state != CallbackState::notified)
+                            {
+                                pCallbackSynchronizationData->m_event.wait(
+                                    lock,
+                                    [pCallbackSynchronizationData]()
+                                    { return pCallbackSynchronizationData->m_state == CallbackState::notified; });
```
bernhardmgruber, line 263:
```suggestion
                                    [&pCallbackSynchronizationData]
                                    { return pCallbackSynchronizationData->m_state == CallbackState::notified; });
```
please capture the shared pointer by reference here.

bernhardmgruber (changes requested):

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
fwyzard: No, I am not removing the parentheses.

SimeonEhrig: I didn't know, that is allowed. I think, it is not well known, that is allowed. Therefore I would also keep the parentheses.

fwyzard: Yes, it is allowed, but IMHO it makes the (already confusing) lambda syntax even more confusing.

bernhardmgruber: > No, I am not removing the parentheses.

I don't insist on it. I would be interested in your reasoning though. Why keep them?

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
fwyzard: I'm not going to, for two reasons:
1. I think it's a bad idea to squeeze unrelated changes in an existing PR/patch set
2. I would need to go through the whole function and lambda hierarchy to understand the lifetime of the various objects and convince myself that this would be a safe change, without any risk that the lifetime of the capturing lambda exceeds that of the captured object; given that these lambdas are being used in different threads, I don't feel confident in doing that

fwyzard: On the other hand, @bernhardmgruber if you have done the analysis and convinced yourself that it is a safe, I would suggest you to make this change in a follow up PR.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
fwyzard: As I wrote above, IMHO the lambda syntax is more confusing without them.

SimeonEhrig: Because, if you are not familiar with the syntax, how do find out, what the syntax mean? If you didn't had the idea of a lambda, try to ask a search engine, what `[]{}` means in c++.

I have similar problems in Python.

bernhardmgruber: Ok, you convinced me. Keep the parentheses.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
bernhardmgruber: Fine, I'll send a follow-up PR.

bernhardmgruber (approved):

PR #1669, opened 2022-03-28
fix Sphinx documentation

psychocoderHPC:
It is not possible to build the documentation due to missing dependencies.

### DIFF
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -5,6 +5,8 @@ breathe==4.16.0
 sphinxcontrib.programoutput
 #sphinxcontrib-napoleon>=0.7
 pygments
+Jinja2<3.0
+markupsafe<2.0.0
 # generate plots
 #matplotlib
 #scipy


docs/requirements.txt
```diff
@@ -5,6 +5,8 @@ breathe==4.16.0
 sphinxcontrib.programoutput
 #sphinxcontrib-napoleon>=0.7
 pygments
+Jinja2<3.0
```
j-stephan, line 8: Is this future proof? Do we need a version update in the future?

j-stephan (approved):

fwyzard:
Could you kick the one runner that seems stuck ?

docs/requirements.txt
psychocoderHPC: I do not know this is fu..ing python^^. I tested it on my local machine and if I used something newer I got version conflicts with other libs.

Issue #1636, opened 2022-03-10
Standalone mode: Add compatibility check header


j-stephan:
User request: For the standalone mode there should be a header that does some basic requirements checking (like the required boost version).

PR #1671, opened 2022-03-28
Smarter `atomic_ref` selection

### LINKED ISSUE #1572: Which CPU atomic implementation should be enabled by default?
As a follow up to #1566:

We need to decide which CPU atomics implementation should be enabled by default.

### `boost::atomic_ref`

+ Pro: better performance
+ Con: hard **runtime** dependency on Boost.Atomic

### lock-based

+ Pro: no runtime dependency, alpaka is still header-only
+ Con: worse performance

@alpaka-group/alpaka-maintainers Please discuss (once you are back...)

j-stephan:
Fixes #1572. This PR changes the way the CPU atomic implementation is selected according to the following rules:

1. Check if we're in C++20 mode. If so, check for C++20 `atomic_ref` using a custom CMake feature check. If successful, use C++20 `std::atomic_ref<T>`. If unsuccessful, go to 2.
2. If Boost.Atomic has been found and the user did not disable `boost::atomic_ref<T>` explicitly, use `boost::atomic_ref<T>`. Otherwise, go to 3.
3. Use the legacy lock-based atomics.

Note: This change mostly affects the CMake part of alpaka. Standalone users need to supply the `-DALPAKA_HAS_STD_ATOMIC_REF` flag themselves when using a conformant C++ standard library (currently only libstdc++ and MSVC STL. libc++ does not support this.)

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -141,6 +141,7 @@ if(NOT TARGET alpaka)
     add_library(alpaka INTERFACE)
 
     target_compile_features(alpaka INTERFACE cxx_std_${alpaka_CXX_STANDARD})
+
     if (CMAKE_CXX_COMPILER_ID STREQUAL "NVHPC" OR CMAKE_CXX_COMPILER_ID STREQUAL "PGI")
         # Workaround for STL atomic issue: https://forums.developer.nvidia.com/t/support-for-atomic-in-libstdc-missing/135403/2
         # still appears in NVHPC 20.7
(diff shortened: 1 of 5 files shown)


include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -12,13 +12,27 @@
 #include <alpaka/atomic/Traits.hpp>
 #include <alpaka/core/BoostPredef.hpp>
 
-#include <boost/atomic.hpp>
+#if defined(ALPAKA_HAS_STD_ATOMIC_REF)
```
fwyzard, line 15: why not simply
```suggestion
#if defined(__cpp_lib_atomic_ref)
```
?

j-stephan: I want to avoid unnecessary includes. `__cpp_lib_atomic_ref` is only defined **after** including `<atomic>`.

fwyzard: you could `#include <version>`

j-stephan: That would work. However, it would mean that this:

```c++
#if defined(ALPAKA_HAS_STD_ATOMIC_REF)
#    include <atomic>
```

turns into this:

```c++
#if __cplusplus >= 202002L
#    include <version>
#endif

#if defined(__cpp_lib_atomic_ref)
#    include <atomic>
```

which I wanted to avoid. And since I needed to change the way CMake picks up the dependency anyway (we don't want to link to Boost.Atomic if that is unnecessary) I introduced the macro `ALPAKA_HAS_STD_ATOMIC_REF`.

Once we have a separate requirement header (see #1636) we can move the checks into there. And that would look pretty similar to the second approach.

cmake/alpakaCommon.cmake
```diff
@@ -141,6 +141,10 @@ if(NOT TARGET alpaka)
     add_library(alpaka INTERFACE)
 
     target_compile_features(alpaka INTERFACE cxx_std_${alpaka_CXX_STANDARD})
+
+    # Force MSVC to set a correct version macro: https://devblogs.microsoft.com/cppblog/msvc-now-correctly-reports-__cplusplus/
+    target_compile_options(alpaka INTERFACE $<$<CXX_COMPILER_ID:MSVC>:"/Zc:__cplusplus">)
+
```
bernhardmgruber: Why is this necessary?

bernhardmgruber (commented):
I find the feature test a bit involved and would prefer a leaner solution, but as you said, you want to decide in cmake as well whether to link against Boost.Atomic or not. I was just reading that Boost.Atomic is header-only when we do not need the lock-based fallbacks. Do you know whether this is the case for alpaka? If yes, then the implementation here could be simplified similar to @fwyzard suggested.

cmake/alpakaCommon.cmake
j-stephan: It is a remnant from when I originally approached this in the way described above (i.e., checking for `__cplusplus` etc.). I'll remove this since it seems to break MSVC for some weird reason.

j-stephan:
> I was just reading that Boost.Atomic is header-only when we do not need the lock-based fallbacks. Do you know whether this is the case for alpaka?

According to the [documentation](https://www.boost.org/doc/libs/1_74_0/doc/html/atomic/interface.html#atomic.interface.feature_macros) this depends on the way Boost is built for the target system. So we have no way of knowing for certain. Right now we always assume that we need to link to something when using Boost.Atomic but we could probably add another `try_compile` to figure out whether Boost.Atomic is actually lock-free on the target system.

bernhardmgruber (approved):

PR #1677, opened 2022-04-05
Change the syntax of saving result of ::omp_get_num_threads()

sbastrakov:
The previously used syntax was valid but sometimes caused internal error for Intel compiler in CI:
```
/home/runner/work/alpaka/alpaka/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp(134): internal error: assertion failed: cd_new_hash_table: malloc failed (cdhashtbl.c, line 134 in cd_new_hash_table)

                                  int const numThreads(::omp_get_num_threads());
```

The changed one suggested by @bernhardmgruber  appears to work around the issue.

I faced this issue sometimes with #1336 and after making the fix in that branch it no longer happened, so I believe it indeed works around the issue.

### DIFF
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
@@ -131,7 +131,7 @@ namespace alpaka
                                         "The OpenMP 2.0 runtime did not create a parallel region!");
                                 }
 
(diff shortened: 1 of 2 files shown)


bernhardmgruber (approved):

sbastrakov:
Ready to be merged.

PR #1685, opened 2022-04-12
Add a new API to allocate pinned host memory

fwyzard:
Allocates pinned/mapped memory on host, accessible by the given device:
  - for the CPU device, uses standard memory allocations;
  - for CUDA/HIP devices, uses the `cudaHostAlloc`/`hipHostMalloc` API.

### DIFF
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -29,7 +29,7 @@ namespace alpaka
         using Error_t = ::cudaError_t;
         using Event_t = ::cudaEvent_t;
         using Extent_t = ::cudaExtent;
-        using Flag_t = int;
+        using Flag_t = unsigned int;
         using FuncAttributes_t = ::cudaFuncAttributes;
         using HostFn_t = void (*)(void* data); // same as cudaHostFn_t, without the CUDART_CB calling convention
         using Limit_t = ::cudaLimit;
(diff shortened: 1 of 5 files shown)


fwyzard:
One of the reasons for adding this is for consistency with a pointer-based buffer implementation for OneAPI: OneAPI does not support "pinning" or "mapping" and existing memory allocation (à la `cudaHostRegister`) but it does support allocating host memory visible from the device (via `malloc_host`).

fwyzard:
Note: I put the allocation of host memory accessible by CUDA/HIP devices in `include/alpaka/mem/buf/BufUniformCudaHipRt.hpp`, but it could just as well go in `include/alpaka/mem/buf/BufCpu.hpp`...

Do let me know if there are different preferences.

fwyzard:
@psychocoderHPC do you think this is OK, or any changes are needed ?

sbastrakov:
@fwyzard just FYI @psychocoderHPC is on vacation this week

fwyzard:
ok :-)

psychocoderHPC:
@fwyzard sry for the delay, I try to check your PR tomorrow.

psychocoderHPC (approved):

fwyzard:
Thanks :+1:

@psychocoderHPC @j-stephan do let me know if you would like me to make other changes related to the memory handling or host / device buffers.

PR #1686, opened 2022-04-12
Add ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT macro

fwyzard:
If `ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT` is defined, use it as the default alignment for host memory allocations.
A compile-time check ensures that it is positive and a power of 2.

Otherwise, the alignment is chosen to enable optimal performance dependant on the target architecture.

### DIFF
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -221,9 +221,20 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
+                // If ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT is defined, positive, and a power of 2, use it as the
+                // default alignment for host memory allocations. Otherwise, the alignment is chosen to enable optimal
+                // performance dependant on the target architecture.
+#if defined(ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT)
+                static_assert(
+                    ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT > 0
+                        && ((ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT & (ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT - 1)) == 0),
+                    "If defined, ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT must be a power of 2.");
+                constexpr std::size_t alignment = static_cast<std::size_t>(ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT);
+#else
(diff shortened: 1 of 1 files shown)


fwyzard:
Here is a simple patch to allow users to specify the minimum alignment for Alpaka host buffers at compile time.

fwyzard:
@bernhardmgruber this should be "good enough" for my test case (globally using an alignment of 128 bytes for all host buffers).
Of course it's not a general approach, since it does not allow specifying the alignment on a per-buffer case.

fwyzard:
Fixed a missing pair of parentheses :-/

bernhardmgruber:
Thank you for the PR! I can understand your motivation for this change. However, I think it is not in the right place. The macro `ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT` you added allows to change the value of `alpaka::core::vectorization::defaultAlignment`. Your macro wants to influence the alignment of host memory and is different from the alignment required for vectorization. The two aspects are connected in `BufCpu.hpp` in the specialization of `BufAlloc` and `BufAllocAsync`. I propose to add your change there.

Furthermore, I think it is not required to check the validity of `ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT`. If it is present, it should be used directly. You may add a `static_assert` if you want.

Finally, I wonder whether that should be added to cmake and documented. But then, the feature might be superseded by a proper integration: #1687. What do other people think? @j-stephan and @psychocoderHPC.

fwyzard:
> The two aspects are connected in `BufCpu.hpp` in the specialization of `BufAlloc` and `BufAllocAsync`. I propose to add your change there.

Sure.

> Furthermore, I think it is not required to check the validity of `ALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT`. If it is present, it should be used directly. You may add a `static_assert` if you want.

OK, I can change it into a `static_assert`.

> [...] the feature might be superseded by a proper integration: #1687.

If you prefer to hold this pending a more dynamic implementation, that's also OK for me.

fwyzard:
The GitLab CI claims to have completed all tests: https://gitlab.com/hzdr/crp/alpaka/-/pipelines/514984072 , but maybe it failed to report back to GitHub ?
Is there a way to kick it, or should I just force-push the branch to trigger a new run ?

bernhardmgruber (approved):

j-stephan (approved):

j-stephan:
> Finally, I wonder whether that should be added to cmake and documented.

Yes and yes. Please open an issue for this.

bernhardmgruber:
> > Finally, I wonder whether that should be added to cmake and documented.
> 
> Yes and yes. Please open an issue for this.

Done: #1691.

PR #1689, opened 2022-04-12
Use aligned new/delete instead of OS APIs

bernhardmgruber:


include/alpaka/core/AlignedAlloc.hpp
```diff
@@ -9,54 +9,22 @@
 
 #pragma once
 
-#include <alpaka/core/BoostPredef.hpp>
```
j-stephan, line 12: Why are you removing this?

bernhardmgruber: Because it was no longer necessary, until I had to readd the special handling for Intel. Good catch, I will readd it!

include/alpaka/core/AlignedAlloc.hpp
```diff
@@ -12,51 +12,20 @@
 #include <alpaka/core/BoostPredef.hpp>
 #include <alpaka/core/Common.hpp>
 
-#if BOOST_COMP_MSVC
-#    include <malloc.h>
-#else
-#    include <cstdlib>
+#include <new>
+#if BOOST_COMP_INTEL
+#    include <aligned_new>
```
j-stephan, line 17: Do we need to remove this in one of your icpc PRs?

bernhardmgruber: Yes. Can be removed when icpc is dropped.

j-stephan (approved):

PR #1692, opened 2022-04-13
Fix CUDA/HIP math function annotations and cover it by unit tests

### LINKED ISSUE #1688: alpaka math functions does not work in lambdas with alpaka 0.9 and CUDA backend
# Problem

I want to pass a lambda to a kernel, which is called and use a `alpaka::math` function. Therefore the lambda needs an `acc` object as argument. The CUDA backend does not support generic lambdas, therefore it needs to be an explicit type. This causes a compiler error. 

```bash
FAILED: CMakeFiles/alpaka_lambda.dir/main.cpp.o 
/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/cuda-11.0.2-mtxiuzc35ykbmjk4fnpmhwnjbgquvcv5/bin/nvcc -forward-unknown-to-host-compiler -ccbin=/usr/bin/g++ -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=47 -DALPAKA_DEBUG=0 -DALPAKA_DEBUG_OFFLOAD_ASSUME_HOST -DALPAKA_OFFLOAD_MAX_BLOCK_SIZE=256 -DBOOST_ALL_NO_LIB -isystem=/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/cuda-11.0.2-mtxiuzc35ykbmjk4fnpmhwnjbgquvcv5/targets/x86_64-linux/include -isystem=/home/sehrig/application/miniconda3/envs/vikunja-dev/include -isystem=/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/boost-1.74.0-lqyr7ou7iafiwvq5fb6ipp62o6l2gqlf/include -isystem=/opt/spack-modules/linux-ubuntu20.04-skylake_avx512/gcc-10.2.0/cuda-11.0.2-mtxiuzc35ykbmjk4fnpmhwnjbgquvcv5/include --generate-code=arch=compute_52,code=[compute_52,sm_52] -Xcompiler=-fPIE --extended-lambda --expt-relaxed-constexpr -Xcudafe=--display_error_number -Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored -Xcompiler -pthread -std=c++17 -MD -MT CMakeFiles/alpaka_lambda.dir/main.cpp.o -MF CMakeFiles/alpaka_lambda.dir/main.cpp.o.d -x cu -c ../main.cpp -o CMakeFiles/alpaka_lambda.dir/main.cpp.o
/home/sehrig/application/miniconda3/envs/vikunja-dev/include/alpaka/math/Traits.hpp(850): error #3050: calling a __device__ function("operator()") from a __host__ __device__ function("max") is not allowed
          detected during instantiation of "auto alpaka::math::max(const T &, const Tx &, const Ty &) [with T=alpaka::AccGpuCudaRt<alpaka::DimInt<1UL>, std::size_t>, Tx=float, Ty=float]" 
../main.cpp(40): here

/home/sehrig/application/miniconda3/envs/vikunja-dev/include/alpaka/math/Traits.hpp(850): error #3050: calling a __device__ function("operator()") from a __host__ __device__ function("max") is not allowed
          detected during instantiation of "auto alpaka::math::max(const T &, const Tx &, const Ty &) [with T=alpaka::AccGpuCudaRt<alpaka::DimInt<1UL>, std::size_t>, Tx=float, Ty=float]" 
../main.cpp(40): here

2 errors detected in the compilation of "../main.cpp".
ninja: build stopped: subcommand failed.
```

The same code is working with the serial backend, OpenMP backend and also the Hip backend. Maybe it also happen with specific CUDA versions. CUDA 11.0 failed and CUDA 11.2 was fine.

# Code to reproduce

Use the alpaka version of commit [ebc1171](https://github.com/alpaka-group/alpaka/commit/ebc1171feac21f1e21c49bcd9f053e7b01b584d0) (branch `0.9.0-rc1`).

CMakeLists.txt
```cmake

project(alpaka_lambda)
set(CMAKE_POSITION_INDEPENDENT_CODE ON)
set(alpaka_CUDA_EXPT_EXTENDED_LAMBDA ON)

option(alpaka_ACC_GPU_CUDA_ENABLE "enable alpaka CUDA accelerator" ON)

find_package(alpaka)

alpaka_add_executable(${PROJECT_NAME} main.cpp)
target_link_libraries(${PROJECT_NAME} PRIVATE alpaka::alpaka)
```

main.cpp
```c++
#include <alpaka/alpaka.hpp>


struct GlobalKernel
{
    template<typename TAcc, typename TFunctor>
    ALPAKA_FN_ACC auto operator()(TAcc const& acc, TFunctor const& functor) const -> void
    {
        float f = functor(acc, 42.f);
        printf("%f\n", f);
    }
};

int main()
{
    using Dim = alpaka::DimInt<1>;
    using Idx = std::size_t;

    using Acc = alpaka::AccGpuCudaRt<Dim, Idx>;
    using Queue = alpaka::Queue<Acc, alpaka::Blocking>;

    auto const devAcc = alpaka::getDevByIdx<Acc>(0u);
    Queue queue(devAcc);

    Vec const elementsPerThread(Vec::all(static_cast<Idx>(1)));
    Vec const threadsPerGrid(Vec::all(static_cast<Idx>(1)));
    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
    WorkDiv const workDiv = alpaka::getValidWorkDiv<Acc>(
        devAcc,
        threadsPerGrid,
        elementsPerThread,
        false,
        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);

    // does not work
    auto functor = [] ALPAKA_FN_HOST_ACC(alpaka::AccGpuCudaRt<Dim, Idx> const& acc, float const i) -> float
    {
        // return i + 1.f;
        return alpaka::math::max(acc, 41.f, i) + 0.5f;
    };
    GlobalKernel kernel;

    alpaka::exec<Acc>(queue, workDiv, kernel, functor);
    alpaka::wait(queue);

    return 0;
}
```

# Possible solution

@psychocoderHPC points me, that is required that the CUDA implementation of `max` needs an explicit return type.
I fixed the type of this function to `float` and looks like, that it works. 
https://github.com/alpaka-group/alpaka/blob/b074b0df68a96321dc73261ab2b9d3d41180f18c/include/alpaka/math/MathUniformCudaHipBuiltIn.hpp#L642

sbastrakov:
First commit is the tests, second is potential fix by @SimeonEhrig @psychocoderHPC .

~~The tests are done in a bit of quick-and-dirty fashion, wanted to make a quick PR to pass it through CI.~~ Ready for  review

TODO:

- [x] Beautify the tests while CI is running
- [x] I messed up authorship of the second commit, will fix.

### DIFF
--- a/include/alpaka/math/MathUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/math/MathUniformCudaHipBuiltIn.hpp
@@ -229,7 +229,7 @@ namespace alpaka::math
         template<typename TArg>
         struct Abs<AbsUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_signed_v<TArg>>>
         {
-            __device__ auto operator()(AbsUniformCudaHipBuiltIn const& /* abs_ctx */, TArg const& arg)
+            __host__ __device__ auto operator()(AbsUniformCudaHipBuiltIn const& /* abs_ctx */, TArg const& arg)
             {
                 if constexpr(is_decayed_v<TArg, float>)
                     return ::fabsf(arg);
@@ -254,7 +254,7 @@ namespace alpaka::math
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            __device__ auto operator()(TCtx const& ctx, Complex<T> const& arg)
+            __host__ __device__ auto operator()(TCtx const& ctx, Complex<T> const& arg)
             {
                 return sqrt(ctx, arg.real() * arg.real() + arg.imag() * arg.imag());
             }
@@ -264,7 +264,7 @@ namespace alpaka::math
         template<typename TArg>
         struct Acos<AcosUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_floating_point_v<TArg>>>
         {
-            __device__ auto operator()(AcosUniformCudaHipBuiltIn const& /* acos_ctx */, TArg const& arg)
+            __host__ __device__ auto operator()(AcosUniformCudaHipBuiltIn const& /* acos_ctx */, TArg const& arg)
             {
                 if constexpr(is_decayed_v<TArg, float>)
                     return ::acosf(arg);
@@ -283,7 +283,7 @@ namespace alpaka::math
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            __device__ auto operator()(TCtx const& ctx, Complex<T> const& arg)
+            __host__ __device__ auto operator()(TCtx const& ctx, Complex<T> const& arg)
             {
                 // This holds everywhere, including the branch cuts: acos(z) = -i * ln(z + i * sqrt(1 - z^2))
                 return Complex<T>{0.0, -1.0} * log(ctx, arg + Complex<T>{0.0, 1.0} * sqrt(ctx, T(1.0) - arg * arg));
@@ -296,7 +296,7 @@ namespace alpaka::math
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            __device__ auto operator()(TCtx const& ctx, TArgument const& argument)
+            __host__ __device__ auto operator()(TCtx const& ctx, TArgument const& argument)
             {
                 // Fall back to atan2 so that boundary cases are resolved consistently
                 return atan2(ctx, TArgument{0.0}, argument);
@@ -309,7 +309,7 @@ namespace alpaka::math
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            __device__ auto operator()(TCtx const& ctx, Complex<T> const& argument)
+            __host__ __device__ auto operator()(TCtx const& ctx, Complex<T> const& argument)
             {
                 return atan2(ctx, argument.imag(), argument.real());
             }
@@ -319,7 +319,7 @@ namespace alpaka::math
         template<typename TArg>
         struct Asin<AsinUniformCudaHipBuiltIn, TArg, std::enable_if_t<std::is_floating_point_v<TArg>>>
         {
-            __device__ auto operator()(AsinUniformCudaHipBuiltIn const& /* asin_ctx */, TArg const& arg)
+            __host__ __device__ auto operator()(AsinUniformCudaHipBuiltIn const& /* asin_ctx */, TArg const& arg)
             {
                 if constexpr(is_decayed_v<TArg, float>)
                     return ::asinf(arg);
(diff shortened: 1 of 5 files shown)


include/alpaka/math/MathUniformCudaHipBuiltIn.hpp
```diff
@@ -18,6 +18,7 @@
 #include <alpaka/math/Complex.hpp>
 #include <alpaka/math/Traits.hpp>
 
+#include <algorithm>
```
j-stephan: Why is this required?

sbastrakov: Oh idk. @SimeonEhrig do you?

SimeonEhrig: That is a mistake. I forgot to remove it.

sbastrakov:
I now put this lambda to existing base math unit test fixture, so that it always applies to all tests for math functions that also checks the value. An unfortunate side product of that change is that on the dev system now it compiles even without the fix! In that case, it fails runtime with `invalid device function`. I guess if we want a hard compilation error, I can also add a single test like in my first attempt, but for a fixed math function. That would be a relatively small test then (also in terms of future maintenance).

sbastrakov:
Oh no, some CI jobs failed. I will have a look tomorrow morning.

sbastrakov:
The current CI issue for clang-cuda seems to be caused by complex number math on CPU. I am now working on that, hopefully fix it soon by adding similar extra function annotations.

(The previously applied fix solved it for real numbers and tests now cover that).

j-stephan (approved):
LGTM.

sbastrakov:
Unfortunately, yet another fix is required as one more thing appeared. Working on it. I've also realized that this PR was opened to 0.9.0-rc1 branch, and not to `dev`. So should I close this PR and open a new one to `develop` with the fixed version?

j-stephan:
No, the release is more important right now. We'll cherry-pick afterwards.

sbastrakov:
A new attempt is up. Along the way I also found another issue with `float` (real-number) tests using `double`. I accidentally made this mistake when introducing tests for complex numbers (that required to change a lot of generic math unit tests things). I made the fix as a separate commit in this PR.

sbastrakov:
@j-stephan ready for re-review. CI finally passed.

j-stephan (approved):

PR #1695, opened 2022-04-14
Add documentation about setting the stdlib with the Clang compiler

SimeonEhrig:
During my work on the CI, I recognized that the automatic selection of the standard library of Clang can cause new problems, depending of the environment. To fix this problems, I spent some time to understand how Clang is selecting and using the different standard libraries. I think this is also interesting for other developer and users. Therefore, I wrote it down in the alpaka documentation.

### DIFF
--- a/docs/source/advanced/compiler.rst
+++ b/docs/source/advanced/compiler.rst
@@ -1,0 +1,52 @@
+Compiler Specifics
+==================
+
+Alpaka supports a large number of different compilers. Each of the compilers has its own special features. This page explains some of these specifics.
+
+Choosing the correct Standard Library in Clang
+++++++++++++++++++++++++++++++++++++++++++++++
+
+Clang supports both, ``libstdc++`` shipped with the GNU GCC toolchain and ``libc++``, LLVM's own implementation of the C++ standard library. By default, clang and all clang-based compilers, such as the ``hipcc``, use ``libstdc++`` (GNU GCC). If more than one ``GCC`` version is installed, it is not entirely clear which version of ``libstdc++`` is selected. The following code can be used to check which standard library and version clang is using by default with the current setup.
+
+.. code-block:: c++
+    
+    #include <iostream>
+
+    int main(){
+        #ifdef _GLIBCXX_RELEASE
+        std::cout << "use libstdc++ (GNU GCC's standard library implementation)" << std::endl;
+        std::cout << "version: " <<   _GLIBCXX_RELEASE << std::endl;
+        #endif
+
+        #ifdef _LIBCPP_VERSION
+        std::cout << "use libc++ (LLVM's standard library implementation)" << std::endl;
+        std::cout << "version: " <<   _LIBCPP_VERSION << std::endl;
+        #endif
+    }
+
+The command ``clang -v ...`` shows the include paths and also gives information about the standard library used.
+
+Choose a specific libstdc++ version
+-----------------------------------
+
+Clang provides the argument ``--gcc-toolchain=<path>`` which allows you to select the path of a GCC installation. For example, if you built the ``GCC`` compiler from source, you can select the installation prefix, which is the base folder with the subfolders ``include``, ``lib`` and so on.
+
+If you are using CMake, you can set the ``--gcc-toolchain`` flag via the following CMake command line argument:
+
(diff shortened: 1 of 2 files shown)


docs/source/advanced/compiler.rst
```diff
@@ -0,0 +1,52 @@
+Compiler Specifics
+==================
+
+Alpaka supports a large number of different compilers. Each of the compilers has its own special features. This page explains some of these specifics.
```
bernhardmgruber, line 4: I usually prefer to add a linebreak after each sentence. Makes for better diffs when editing.

docs/source/advanced/compiler.rst
```diff
@@ -0,0 +1,52 @@
...
+
+Choosing the correct Standard Library in Clang
+++++++++++++++++++++++++++++++++++++++++++++++
+
+Clang/LLVM supports both the ``libstdc++`` shipped with the GNU GCC toolchain and the ``libc++``, llvm's own implementation of the standard library. By default, Clang and all Clang-based compilers, such as the ``hipcc``, use the ``libstdc++`` (GNU GCC). If more than one ``GCC`` version is installed, it is not entirely clear which version of ``libstdc++`` is selected. The following code can be used to check which default library and version Clang is using with the current setup.
```
bernhardmgruber:
```suggestion
Clang supports both, ``libstdc++`` shipped with the GNU GCC toolchain and ``libc++``, LLVM's own implementation of the C++ standard library. By default, clang and all clang-based compilers, such as the ``hipcc``, use ``libstdc++`` (GNU GCC). If more than one ``GCC`` version is installed, it is not entirely clear which version of ``libstdc++`` is selected. The following code can be used to check which standard library and version clang is using by default with the current setup.
```

docs/source/advanced/compiler.rst
```diff
@@ -0,0 +1,52 @@
...
+* ``-DCMAKE_CXX_FLAGS="--gcc-toolchain=<path>"`` if you use Clang as compiler for CPU backends or the HIP backend.
+* ``-DCMAKE_CUDA_FLAGS="--gcc-toolchain=<path>"`` if you use Clang as CUDA compiler.
+
+.. hint:: If you are using Ubuntu and install a new gcc version via apt, it is not possible to select a specific gcc version because apt installs all headers and shared libraries in subfolders of ``/usr/include`` and ``/usr/lib``. Therefore, you can only use the ``/usr`` base path and Clang will automatically select one of the installed libstdc++ versions.
+
+.. hint:: If you installed Clang/LLVM with spack and a gcc compiler, the Clang compiler will use the ``libstdc++`` of the compiler used to build Clang/LLVM.
+
+
+Select the libc++
+-----------------
```
bernhardmgruber:
```suggestion
Selecting libc++
----------------
```

docs/source/advanced/compiler.rst
```diff
@@ -0,0 +1,52 @@
...
+
+The ``libc++`` can be used if you set the compiler flag ``-stdlib=libc++``.
```
bernhardmgruber:
```suggestion
``libc++`` can be used if you set the compiler flag ``-stdlib=libc++``.
```

docs/source/advanced/compiler.rst
```diff
@@ -0,0 +1,52 @@
...
+
+If you are using CMake, you can select the ``libc++`` via the following CMake command line argument:
```
bernhardmgruber:
```suggestion
If you are using CMake, you can select ``libc++`` via the following CMake command line argument:
```

docs/source/advanced/compiler.rst
sbastrakov: I also prefer that style

SimeonEhrig: I would avoid manual line breaks. Otherwise the renderer gets trouble to produce correct new lines on the website or in the pdf.

bernhardmgruber: I have never seen issues with line breaks. But it is also not a deal-breaker for me if you keep it like this.

bernhardmgruber (approved):

docs/source/advanced/compiler.rst
SimeonEhrig: To be fare, the website is nearly static and you check the rendering manually. Therefore it is more a theoretical problem.

PR #1696, opened 2022-04-14
Updated documentation for 0.9.0

tonydp03:
Updated documentation for `alpaka::memcpy` and `alpaka::memset`, which now take `queue` as first argument, both for synchronous and asynchronous copies. The extent is not necessary in case one wants to copy the full buffer and not partially, but it might be worth leaving it in the documentation.


j-stephan (approved):
Thanks for catching this!

PR #1700, opened 2022-04-26
Add icpx to CI

### LINKED ISSUE #1238: Support for ICX/ICPX in CI
@ax3l suggested adding new Intel ICX/ICPX compiler to our CI.

see: https://github.com/alpaka-group/alpaka/issues/995#issuecomment-758120292

bernhardmgruber:
Fixes: #1238


psychocoderHPC (dismissed):

bernhardmgruber:
```
/home/runner/work/alpaka/alpaka/include/alpaka/block/shared/st/BlockSharedMemStOmp5BuiltIn.hpp:34:26: error: 'omp_pteam_mem_alloc' is equivalent to 'omp_default_mem_alloc’ for variables with static storage duration, consider using a local variable instead [-Werror,-Wopenmp-allocate]
#    pragma omp allocate(shMem) allocator(omp_pteam_mem_alloc)
                         ^
```
How should I fix/workaround this warning? @jkelling?

test/common/devCompileOptions.cmake
```diff
@@ -132,8 +132,8 @@ ELSE()
             LIST(APPEND alpaka_DEV_COMPILE_OPTIONS "-Wcast-align=strict")
         ENDIF()
 
-    # Clang or AppleClang
-    ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+    # Clang, AppleClang, ICPX
+    ELSEIF(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR ${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
```
j-stephan: `IntelLLVM` is not known before CMake 3.20. It will probably still work with CMake 3.18 (the alpaka minimum) I guess.

test/common/devCompileOptions.cmake
```diff
@@ -157,7 +157,13 @@ ELSE()
         IF(alpaka_ACC_GPU_HIP_ENABLE)
             LIST(APPEND alpaka_DEV_COMPILE_OPTIONS "-Wno-unused-command-line-argument")
         ENDIF()
-    # ICC
+
+        # ICPX
+        if (${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
+            # fast math is turned on by default, which breaks our unit tests
+            LIST(APPEND alpaka_DEV_COMPILE_OPTIONS "-fp-model=precise")
```
j-stephan: This, on the other hand, might break. Before CMake 3.20 `icpx` is identified as `Clang`.

j-stephan (changes requested):

jkelling:
> ```
> /home/runner/work/alpaka/alpaka/include/alpaka/block/shared/st/BlockSharedMemStOmp5BuiltIn.hpp:34:26: error: 'omp_pteam_mem_alloc' is equivalent to 'omp_default_mem_alloc’ for variables with static storage duration, consider using a local variable instead [-Werror,-Wopenmp-allocate]
> #    pragma omp allocate(shMem) allocator(omp_pteam_mem_alloc)
>                          ^
> ```
> 
> How should I fix/workaround this warning? @jkelling?

[OpenMP5,1, section 2.21.1.2](https://www.openmp.org/spec-html/5.1/openmpsu113.html#x148-1620002.21.1.2)
```
 Variables with static storage duration that are declared in called routines in the region are shared. 
````
I guess their interpretation of the word "shared" here is "shared among all threads executing the region, but no further". In this case, the `static` in the line following this would suffice and we could just remove that directive, as the warning suggests.

However, the meaning of "shared" in that bit of the standard is not that clear to me. [Section 2.14.7](https://www.openmp.org/spec-html/5.1/openmpsu70.html#x98-1080002.14.7) (`declare target`) states:
```
If a variable with static storage duration is declared in a device routine then the named variable is treated as if it had appeared in a to clause on a declare target directive. 
```
`#pragma omp declare target (to: var)` would initialize the value of var once in side the device environment and, unless there is a `threadprivate(var)` somewhere, it would be shared on the whole device, i.e. at least between all teams of a target region, I think even between target regions in different tasks (device-global). At least inside a target region, this would demand `omp_pteam_mem_alloc != omp_default_mem_alloc`.

So, either the compiler is wrong or the Standard is inconsistent. Although, since this is occurring inside a target region, the compiler would be wrong either way.

As a work around, we could remove the directive only for icpx, as it claims it would do what we want without the directive. On the other hand, I am not confident, that the compiler will stick to this promise once you actually target an accelerator. Adding `-Wno-openmp-allocate` for icpx would be another option or disabling this warning around that line.

File: test/common/devCompileOptions.cmake (hunk 1 above)
Comment (bernhardmgruber) [reply test/common/devCompileOptions.cmake]: Since the code is the same for clang and icpx, it should not matter.

File: test/common/devCompileOptions.cmake (hunk 2 above)
Comment (bernhardmgruber) [reply test/common/devCompileOptions.cmake]: I changed it to clang's `-ffp-model=precise` and add it for both, clang and icpx. Let's see what the CI says.

**Comment by bernhardmgruber** [comment]
@jkelling thx for your explanation! I disabled the warning now and linked to your post.

File: test/common/devCompileOptions.cmake [hunk 3]
```diff
@@ -157,7 +157,12 @@ ELSE()
         IF(alpaka_ACC_GPU_HIP_ENABLE)
             LIST(APPEND alpaka_DEV_COMPILE_OPTIONS "-Wno-unused-command-line-argument")
         ENDIF()
-    # ICC
+        IF(CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM" OR
+            (CMAKE_VERSION VERSION_LESS 20 AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10))
```
j-stephan: Shouldn't this line look more like this:

```suggestion
            (CMAKE_VERSION VERSION_LESS 3.20 AND CMAKE_CXX_COMPILER_ID STREQUAL "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 10))
```

bernhardmgruber: Yes, it should :)

test/common/devCompileOptions.cmake
```diff
@@ -157,7 +157,24 @@ ELSE()
...
+                #ifndef __INTEL_LLVM_COMPILER\
+                #    error\
+                #endif
+                int main() {}"
+                IS_ICPX)
+        ENDIF()
+        IF (IS_ICPX)
+            # fast math is turned on by default with ICPX, which breaks our unit tests
+            LIST(APPEND alpaka_DEV_COMPILE_OPTIONS "-ffp-model=precise")
+        ENDIF()
```
bernhardmgruber, line 176: I eventually settled with distinguishing between clang and icpx and only enable fp model precise for icpx.

test/common/devCompileOptions.cmake
bernhardmgruber: I only added it for icpx now, see below.

README.md
```diff
@@ -70,18 +70,18 @@ Supported Compilers
...
+| OpenMP 5.0 (CPU)                                                               | :x:                                       | :x:                                       | :x:                                       | :x:                                  | :x:                    | :x:                     | :x:                                             | :x:                                              | :white_check_mark:                       | :white_check_mark:                        | -                                    | :x:                | :x:                                                   | :x:                                       | :x:                                  |
```
j-stephan, line 78: `icpx` not supporting OpenMP 5 and TBB seems weird to me. Have you just not tested this or are you positive it doesn't work?

bernhardmgruber: I did not test it. I will create an issue and send a subsequent PR.

j-stephan (dismissed):
Please rebase!

README.md
bernhardmgruber: It's here: #1705

j-stephan (approved):

PR #1702, opened 2022-04-26
Drop support for icpc

### LINKED ISSUE #62: OpenMP 5 accelerator does not compile with ICC
The `ExecCpuOmp4` / `AccCpuOmp4` does not compile with ICC because: `... called in "omp target" region must appear in prior "omp declare target" pragma`.
This affects at least `core::mapIdx`, `block::sync::syncBlockThreads` and all the accelerator functions that can be called within a kernel.
GCC compiles this correctly and does not even issue a warning (vioalting the standard). This is reasonable because it currently only supports CPU targets which does not require to build for additional back-ends.

The problem is, that the `omp declare target` syntax is so verbose that it can not be represented by a simple annotation in front of the functions but requires `#pragma omp declare target` and `#pragma omp end declare target` after the functions.

### LINKED ISSUE #1288: Intel compiler version 20 bug
As discussed on the VC, making an issue to document and so that we could reference it in the code briefly. While working on #1279 I've faced an issue with internal compiler error when using Intel compiler version 20. It seems to not appear for earlier versions. The behavior we observed exactly matches [this bug already reported to Intel](https://community.intel.com/t5/Intel-C-Compiler/compiler-2020-1-217-internal-error-assertion-failed-find-assoc/m-p/1227009#M37871).

@psychocoderHPC found a workaround for that piece of code and shared it in that discussion:
```cpp
template < typename F>
void foo(const F& f)
{
  #pragma omp parallel for schedule(dynamic) collapse(2) 
  for (unsigned int v = 0; v < 0; ++v){
   for (unsigned int t = 0; t < 0; ++t){
      auto x = [&f]( const auto & i){ f(i);};
      x(0);
   }
  }
}

void f() {
 foo( [](const auto &indices_2) {} );
}
```

### LINKED ISSUE #1701: Drop support for icpc
The Intel Classic compiler (icpc) was deprecated by Intel a while ago and the Intel LLVM compiler (icpx) is an established alternative by now. In order to reduce maintenance work, I propose to drop support for the Intel Classic compiler.

bernhardmgruber:
Fixes: #62, #1288, #1701


cmake/alpakaCommon.cmake
```diff
@@ -179,6 +179,10 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_
     message(FATAL_ERROR "Clang versions < 4.0 are not supported!")
 endif()
 
+if(CMAKE_CXX_COMPILER_ID MATCHES "Intel ") # match Intel xxxx.x.x but not IntelLLVM xxxx.x.x
+    message(FATAL_ERROR "Support for the Intel Classic compiler (icpc) has stopped. Please upgrade to the Intel LLVM compiler (ipcx)!")
```
```suggestion
    message(FATAL_ERROR "The Intel Classic compiler (icpc) is no longer supported. Please upgrade to the Intel LLVM compiler (ipcx)!")
```

Also, shouldn't we make this a warning instead? If a user wants to use `icpc` anyway they can try - we shouldn't block them from doing so.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -11,6 +11,11 @@
 
 #include <boost/predef.h>
 
+#ifdef __INTEL_COMPILER
+#    error                                                                                                            \
+        "Support for the Intel Classic compiler (icpc) has stopped. Please upgrade to the Intel LLVM compiler (ipcx)."
```
j-stephan:
```suggestion
        "The Intel Classic compiler (icpc) is no longer supported. Please upgrade to the Intel LLVM compiler (ipcx)."
```

Also again: warning or error?

include/alpaka/core/BoostPredef.hpp
```diff
@@ -57,17 +62,6 @@
 #    define BOOST_COMP_CLANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
 #endif
 
-// Intel compiler detection
-// As of Boost 1.74, Boost.Predef's compiler detection is a bit weird. Recent Intel compilers will be identified as
-// BOOST_COMP_INTEL_EMULATED. Boost.Predef has lackluster front-end support and mistakes the EDG front-end
-// for an actual compiler.
-// TODO: Whenever you look at this code please check whether https://github.com/boostorg/predef/issues/28 and
-// https://github.com/boostorg/predef/issues/51 have been resolved.
```
j-stephan: The explanation here needs to be put in front of the PGI compiler detection now.

j-stephan (changes requested):

cmake/alpakaCommon.cmake
bernhardmgruber: ok!

include/alpaka/core/BoostPredef.hpp
bernhardmgruber: warning sounds good.

include/alpaka/core/BoostPredef.hpp
bernhardmgruber: done.

j-stephan (approved):

j-stephan:
Please rebase on `develop` and remove the ICPC-related changes introduced in #1689.

j-stephan (changes requested):
See above.

bernhardmgruber:
> Please rebase on develop and remove the ICPC-related changes introduced in https://github.com/alpaka-group/alpaka/pull/1689.

Done.

j-stephan (dismissed):

j-stephan (approved):

PR #1704, opened 2022-05-02
Add braces around aggregate initializer

fwyzard:


fwyzard:
Without this, compiling with `-Werror=missing-braces` results in
```
.../external/gcc/10.3.0-84898dea653199466402e67d73657f10/include/c++/10.3.0/bits/shared_ptr_base.h:558:14:   required from here
.../external/alpaka/develop-879b95ffce2/include/alpaka/core/UniformCudaHip.hpp:73:74: error: missing braces around initializer for 'std::__array_traits<cudaError, 1>::_Type' {aka 'cudaError [1]'} [-Werror=missing-braces]
```

sbastrakov (approved):

bernhardmgruber:
It's interesting that there is a warning about this, because eliding these braces is allowed by the standard: https://en.cppreference.com/w/cpp/language/aggregate_initialization

> The braces around the nested initializer lists may be elided (omitted), in which case as many initializer clauses as necessary are used to initialize every member or element of the corresponding subaggregate, and the subsequent initializer clauses are used to initialize the following members of the object.

Anyway, since the warning exists we can try to conform.

bernhardmgruber (approved):

j-stephan (approved):

j-stephan:
Looks like GitLab und GitHub don't like each other anymore. I'm force-merging this now but we need to look into this ASAP.

fwyzard:
> It's interesting that there is a warning about this, because eliding these braces is allowed by the standard

To give more context, this warning (treated as error) seems to happen only when compiling with `nvcc`.
The original code is
```c++
std::array<typename TApi::Error_t, sizeof...(ignoredErrorCodes)> const aIgnoredErrorCodes{ignoredErrorCodes...};
```
but after processing by `nvcc`, `g++` actually receives
```c++
const std::array<typename TApi::Error_t, sizeof...(ignoredErrorCodes)> aIgnoredErrorCodes{(ignoredErrorCodes)...};
```
with an extra set of parenthesis around `ignoredErrorCodes`.

Could that be the reason for the warning ?

Issue #391, opened 2017-08-31
do not create new std::thread for each registered CUDA stream callback


BenjaminW3:
Currently, for each registered CUDA stream callback a new std::thread is created. This can be very time/resource demanding when many callbacks are registered.
The most flexible solution would be a thread pool with a queue of ready callbacks. If there is only one thread, the latency would be highest but the resource usage would be least. If there are multiple threads, the latency/resource tradeoff could be adapted per use case.

psychocoderHPC:
Note: It is also possible to use a self growing thread pool which always create a new thread if there is no free thread available but never shrinks (remove threads). Than we will reach a maximum number of created threads depending of the concurrency within the program.

j-stephan:
@fwyzard @felicepantaleo @waredjeb @tonydp03 I suppose this is what you encountered in the alpaka port of pixeltrack-standalone. We'll discuss this in our next internal meeting.

j-stephan:
I think we can solve this through a new policy / trait. By default use the current behaviour, but allow users to override it with an alpaka-provided thread pool policy (or their own).

j-stephan:
As @bernhardmgruber kindly pointed out, `cudaStreamAddCallback` is deprecated and will be removed in a future CUDA release. We should take a look at this sooner rather than later.

fwyzard:
The [latest documentation](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1g74aa9f4b1c2f12d994bf13876a5a2498) says that

> This function is slated for _eventual deprecation and removal_. If you do not require the callback to execute in case of a device error, consider using `cudaLaunchHostFunc`. Additionally, this function is not supported with `cudaStreamBeginCapture` and `cudaStreamEndCapture`, unlike `cudaLaunchHostFunc`.

PR #1707, opened 2022-05-03
Plugged ThreadPool into alpaka and refactored DevUniformCudaHipRt

tonydp03:
Solves #391. This PR implements the first version of the threadpool to handle CUDA/HIP callbacks instead of creating an `std::thread` each time (cuda/hip)`LaunchHostFunc` is called. In order to do so, refactoring `DevUniformCudaHipRt` was necessary; now it as a separate implementation as the other alpaka devices.

Currently the threadpool starts with `std::hardware_concurrency()` number of threads, but the next step is to start with only one thread and create a new one whenever no threads are available/ready to execute the function.

### DIFF
--- a/include/alpaka/core/ThreadPool.hpp
+++ b/include/alpaka/core/ThreadPool.hpp
@@ -1,0 +1,79 @@
+/* Copyright 2022 Antonio Di Pilato
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <condition_variable>
+#include <functional>
+#include <future>
+#include <mutex>
+#include <queue>
+#include <thread>
+#include <vector>
+
+namespace alpaka::core
+{
+    class ThreadPool
+    {
+        using taskType = std::packaged_task<void()>;
+
+    public:
+        ThreadPool()
+        {
+            for(unsigned int i = 0u; i < std::thread::hardware_concurrency(); ++i)
+                //! All the threads run the lambda function in loop
+                m_threads.emplace_back(std::thread(
+                    [this]
+                    {
+                        taskType task;
+                        while(true)
+                        {
+                            {
+                                std::unique_lock<std::mutex> lock{m_mutex};
+                                m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
+
+                                if(m_stop && m_tasks.empty())
+                                    return;
+
+                                task = std::move(m_tasks.front());
+                                m_tasks.pop();
+                            }
+                            task();
+                        }
+                    }));
+        }
+        ~ThreadPool()
+        {
+            {
+                std::unique_lock<std::mutex> lock{m_mutex};
+                m_stop = true;
+            }
+            m_cond.notify_all();
+            for(auto& thread : m_threads)
+                thread.join();
+        }
+        auto submit(taskType&& newTask) -> std::future<void>
+        {
+            auto f = newTask.get_future();
+            {
+                std::unique_lock<std::mutex> lock{m_mutex};
+                m_tasks.emplace(std::move(newTask));
+            }
+            m_cond.notify_one();
+            return f;
+        }
+
+    private:
+        std::vector<std::thread> m_threads;
+        std::condition_variable m_cond;
+        std::mutex m_mutex;
+        bool m_stop = false;
+        std::queue<taskType> m_tasks;
+    };
+} // namespace alpaka::core
--- a/include/alpaka/core/ThreadPool.hpp
+++ b/include/alpaka/core/ThreadPool.hpp
@@ -9,6 +9,7 @@
 
 #pragma once
 
+#include <atomic>
 #include <condition_variable>
 #include <functional>
 #include <future>
@@ -24,30 +25,6 @@ namespace alpaka::core
         using taskType = std::packaged_task<void()>;
 
     public:
-        ThreadPool()
-        {
-            for(unsigned int i = 0u; i < std::thread::hardware_concurrency(); ++i)
-                //! All the threads run the lambda function in loop
-                m_threads.emplace_back(std::thread(
-                    [this]
-                    {
-                        taskType task;
-                        while(true)
-                        {
-                            {
-                                std::unique_lock<std::mutex> lock{m_mutex};
-                                m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
-
-                                if(m_stop && m_tasks.empty())
-                                    return;
-
-                                task = std::move(m_tasks.front());
-                                m_tasks.pop();
-                            }
-                            task();
-                        }
-                    }));
-        }
         ~ThreadPool()
         {
             {
@@ -65,15 +42,45 @@ namespace alpaka::core
                 std::unique_lock<std::mutex> lock{m_mutex};
                 m_tasks.emplace(std::move(newTask));
             }
+            {
+                std::unique_lock<std::mutex> lock_c{c_mutex};
+                if(m_busyThreads == m_threads.size())
+                {
+                    m_threads.emplace_back(std::thread(
+                        [this]
+                        {
+                            taskType task;
+                            while(true)
+                            {
+                                {
+                                    std::unique_lock<std::mutex> lock{m_mutex};
+                                    m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
+
+                                    if(m_stop && m_tasks.empty())
+                                        break;
+
+                                    task = std::move(m_tasks.front());
+                                    m_tasks.pop();
+                                }
+
+                                task();
+                                m_busyThreads--;
+                            }
+                        }));
+                }
+            }
             m_cond.notify_one();
+            m_busyThreads++;
+
             return f;
         }
 
     private:
         std::vector<std::thread> m_threads;
         std::condition_variable m_cond;
-        std::mutex m_mutex;
-        bool m_stop = false;
+        std::mutex m_mutex, c_mutex;
+        bool m_stop{false};
         std::queue<taskType> m_tasks;
+        std::atomic<unsigned int> m_busyThreads{0};
     };
 } // namespace alpaka::core
--- a/include/alpaka/dev/DevUniformCudaHipRt.hpp
+++ b/include/alpaka/dev/DevUniformCudaHipRt.hpp
@@ -12,6 +12,7 @@
 #if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
 
 #    include <alpaka/core/Concepts.hpp>
+#    include <alpaka/core/ThreadPool.hpp>
 #    include <alpaka/dev/Traits.hpp>
 #    include <alpaka/mem/buf/Traits.hpp>
 #    include <alpaka/pltf/Traits.hpp>
@@ -28,6 +29,7 @@
 #    endif
 
 #    include <cstddef>
+#    include <future>
 #    include <string>
 #    include <vector>
 
@@ -43,7 +45,20 @@ namespace alpaka
     {
         template<typename TApi, bool TBlocking>
         class QueueUniformCudaHipRt;
-    }
+
+        inline auto getThreadPool() -> core::ThreadPool&
+        {
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+            static core::ThreadPool pool;
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic pop
+#    endif
+            return pool;
+        }
+    } // namespace uniform_cuda_hip::detail
 
     template<typename TApi>
     using QueueUniformCudaHipRtBlocking = uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, true>;
@@ -71,7 +86,7 @@ namespace alpaka
     public:
         ALPAKA_FN_HOST auto operator==(DevUniformCudaHipRt const& rhs) const -> bool
         {
-            return m_iDevice == rhs.m_iDevice;
+            return getNativeHandle() == rhs.getNativeHandle();
         }
         ALPAKA_FN_HOST auto operator!=(DevUniformCudaHipRt const& rhs) const -> bool
         {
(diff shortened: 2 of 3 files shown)


include/alpaka/core/ThreadPool.hpp
```diff
@@ -0,0 +1,79 @@
...
+        ThreadPool() : num_threads(std::thread::hardware_concurrency())
+        {
+            for(int i = 0u; i < num_threads; ++i)
+                //! All the threads run the lambda function in loop
+                m_threads.emplace_back(std::thread(
+                    [this]
+                    {
+                        while(true)
+                        {
+                            taskType task;
```
fwyzard: is it necessary to create a new `task` object for every iteration ?

include/alpaka/core/ThreadPool.hpp
```diff
@@ -0,0 +1,79 @@
...
+                            {
+                                std::unique_lock<std::mutex> lock{m_mutex};
+                                m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
+
+                                if(m_stop && m_tasks.empty())
+                                    return;
+
+                                task = std::move(m_tasks.front());
+                                m_tasks.pop();
```
fwyzard: is this safe ?
can it happen that
  - thread 1 moves a task from `m_tasks.front()`
  - thread 2 moves a task from `m_tasks.front()`
  - thread 1 pops
  - thread 2 pops
 
 ?
 
 Then "thread 2" would get a moved-from task, and the second task would never be run.

fwyzard: Ah... sorry, I had not seen the `lock`.

fwyzard:
@tonydp03 a few suggestion to improve this implementation:

This approach creates a new thread pool for each CUDA/HIP device, and populates it with N threads.
With these changes, multiple calls to `GetDevByIdx` with the same device index will return different objects each with their own thread pool.
It would seem better to create a single thread pool, and reuse it for all CUDA/HIP devices.

When creating the pool, populating it with N threads (or even just 1 thread) may not be necessary: a job that never `enqueue`s a host function should not need to create any host threads at all.
Once new threads can be created on demand, it would seem better to start with an empty pool.

Just as an example, on a machine like our `patatrack02` (which has 40 CPU cores and 5 GPUs) a job that enumerates all GPUs and just creates an Alpaka device for each of them will end up with 200 idle threads !

tonydp03:
> @tonydp03 a few suggestion to improve this implementation:
> 
> This approach creates a new thread pool for each CUDA/HIP device, and populates it with N threads. With these changes, multiple calls to `GetDevByIdx` with the same device index will return different objects each with their own thread pool. It would seem better to create a single thread pool, and reuse it for all CUDA/HIP devices.

I see the point. At some point in the discussion there was also the idea to have the thread pool as class member of the Platform rather than the device, but I'm not sure how to implement it and if it would be the correct solution. 

> When creating the pool, populating it with N threads (or even just 1 thread) may not be necessary: a job that never `enqueue`s a host function should not need to create any host threads at all. Once new threads can be created on demand, it would seem better to start with an empty pool.
> 
> Just as an example, on a machine like our `patatrack02` (which has 40 CPU cores and 5 GPUs) a job that enumerates all GPUs and just creates an Alpaka device for each of them will end up with 200 idle threads !

I agree, I used `std::hardware_concurrency` just because I was testing a standalone locally on my computer, then I left that number. But with the self-growing pool, we should be able to start also with no threads at (rather than one).

fwyzard:
> At some point in the discussion there was also the idea to have the thread pool as class member of the Platform rather than the device, but I'm not sure how to implement it and if it would be the correct solution.

I think currently the Platform classes cannot be instantiated, only used as a type.
If they are changed into singleton-style classes, they could hold the thread pool... then we would have one pool for CUDA and one pool for HIP/ROCm -- and out of my head I don't know if having two pools would be better or worse than having a single one.

include/alpaka/core/ThreadPool.hpp
bernhardmgruber: so I guess you suggest moving the variable atop the while loop?

include/alpaka/core/ThreadPool.hpp
```diff
@@ -0,0 +1,79 @@
...
+                std::unique_lock<std::mutex> lock{m_mutex};
+                m_tasks.emplace(std::move(newTask));
+            }
+            m_cond.notify_one();
+            return f;
+        }
+
+    private:
+        std::vector<std::thread> m_threads;
+        int num_threads;
```
bernhardmgruber:
```suggestion
```
This data member is actually not needed. It is the same as `m_threads.size()`.

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -43,7 +44,30 @@ namespace alpaka
...
+            }
+            auto sendTaskToThePool(std::packaged_task<void()>&& task) -> std::future<void>
+            {
+                return callbackPool.submit(std::move(task));
+            }
+
+        private:
+            int m_iDevice;
+            core::ThreadPool callbackPool;
+        };
```
bernhardmgruber: As @fwyzard suggested, maybe just use a single threadpool for entire alpaka:

```suggestion
auto& getThreadPool() -> core::ThreadPool {
    static core::ThreadPool pool;
    return pool;
}
```

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -80,14 +101,20 @@ namespace alpaka
 
         [[nodiscard]] auto getNativeHandle() const noexcept -> int
         {
-            return m_iDevice;
+            return m_spDevUniformCudaHipRtImpl->getNativeHandle();
+        }
+        auto sendTaskToThePool(std::packaged_task<void()>&& task) -> std::future<void>
+        {
+            return m_spDevUniformCudaHipRtImpl->sendTaskToThePool(std::move(task));
```
bernhardmgruber:
```suggestion
            return detail::getThreadPool().submit(std::move(task));
```

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -261,15 +261,12 @@ namespace alpaka
                             spCallbackSynchronizationData->m_state = CallbackState::finished;
                         }
                         spCallbackSynchronizationData->m_event.notify_one();
-                    });
+                    }));
 
+                // If the queue is Blocking, wait for the result.
```
bernhardmgruber: I don't think this comment is necessary given the simplicity of the code below.
```suggestion
```

include/alpaka/core/ThreadPool.hpp
fwyzard: possibly - but I did not look into how heavy an object `std::packaged_task` is, and whether it is considered good practice to reuse them or not.

tonydp03: It should be 64 bytes (as returned by `sizeof(std::packaged_task<void()>)`. I guess it would be better to reuse it. I tested it both in standalone and in alpaka (tests passed).

include/alpaka/core/ThreadPool.hpp
tonydp03: Removed the variable and put the number of threads (`std::thread::hardware_concurrency()` directly in the `for` loop of the constructor.

include/alpaka/dev/DevUniformCudaHipRt.hpp
tonydp03: Should I revert back to the Device only (without implementation) or do we want to keep the refactoring?

bernhardmgruber: What is the benefit of the `DevUniformCudaHipRtImpl` class? I would rather remove it, if there is no compelling reason to keep it.

tonydp03: `auto&` with trailing return type (rather than plain `auto`) apparently makes the compiler angry. So I just went for `core::ThreadPool&` in the declaration

bernhardmgruber: Sorry, error on my side. Either `auto& getThreadPool() {...}` or `auto getThreadPool() -> core::ThreadPool& {...}` are correct. I prefer the latter.

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -43,7 +45,13 @@ namespace alpaka
     {
         template<typename TApi, bool TBlocking>
         class QueueUniformCudaHipRt;
-    }
+
+        static core::ThreadPool& getThreadPool()
```
bernhardmgruber:
```suggestion
        inline core::ThreadPool& getThreadPool()
```

bernhardmgruber (changes requested):

tonydp03:
I pushed a new commit so you can check differences with the previous version. In the dynamic thread pool, no threads are spawned in the constructor of the class, and the first thread is created when submitting the first task to the pool. Then an `atomic` variable is used to check if all the threads are busy (executing a task) or if any of them are waiting for new tasks through the `condition_variable`.

I also used a second `mutex` to ensure that the `emplace_back` function that fills the vector with new threads is thread-safe, but I'm not entirely sure this is really necessary (from some simple tests, it seems not) - maybe I'm missing something which makes it thread-safe anyway (as of my knowledge, the `if` statement on an `atomic` variable is not enough).

include/alpaka/core/ThreadPool.hpp
```diff
@@ -65,15 +42,45 @@ namespace alpaka::core
...
+
             return f;
         }
 
     private:
         std::vector<std::thread> m_threads;
         std::condition_variable m_cond;
-        std::mutex m_mutex;
-        bool m_stop = false;
+        std::mutex m_mutex, c_mutex;
```
bernhardmgruber, line 81: The naming of `c_mutex` is confusing. Please rename the mutexes into something like ` m_tasksMutex` and `m_threadsMutex`.

include/alpaka/core/ThreadPool.hpp
```diff
@@ -65,15 +42,45 @@ namespace alpaka::core
...
+        bool m_stop{false};
         std::queue<taskType> m_tasks;
+        std::atomic<unsigned int> m_busyThreads{0};
```
bernhardmgruber, line 84: Since this value can become negative when a newly created thread already finishes a task before you reach `m_busyThreads++`, I am inclined to want a signed integral here.

bernhardmgruber (commented):
Please also rephrase your commit messages to be written in imperative from, not past tense.

include/alpaka/core/ThreadPool.hpp
```diff
@@ -0,0 +1,86 @@
...
+#include <mutex>
+#include <queue>
+#include <thread>
+#include <vector>
+
+namespace alpaka::core
+{
+    class ThreadPool
+    {
+        using taskType = std::packaged_task<void()>;
```
fwyzard, line 25: I'm not sure if there are strong naming conventions in alpaka... looking through the headers, _most_ of the types and type aliases are upper case, so maybe
```suggestion
        using Task = std::packaged_task<void()>;
```
?

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -43,7 +45,20 @@ namespace alpaka
...
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+            static core::ThreadPool pool;
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic pop
+#    endif
+            return pool;
+        }
+    } // namespace uniform_cuda_hip::detail
```
fwyzard, line 61: After re-reading the changes I've been thinking... there is no particular reason why the thread pool should be bound to the CUDA/HIP device type, is it ?

Why not move this `getThreadPool()` as a `static` method of the `ThreadPool` class itself ?

And then use it directly from `enqueue()`, instead of making `enqueue()` call `sendTaskToThePool()` and `sendTaskToThePool()` use the thread pool ?

psychocoderHPC: This would introduce for the first time in alpaka a state which is not controlled by the user. 
If possible we should avoid such global state.

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -43,7 +45,20 @@ namespace alpaka
...
         class QueueUniformCudaHipRt;
-    }
+
+        inline auto getThreadPool() -> core::ThreadPool&
+        {
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+            static core::ThreadPool pool;
```
psychocoderHPC, line 55: I am confused, I thought we discussed that the thread pool will be part of the device and not unbound to any object.
Take care in cases where a device is destroyed you must check that the thread pool has no outstanding tasks else it could be you executing on invalid devices or memory.

IMO the pool should be a member of `DevUniformCudaHipRt`

psychocoderHPC (changes requested):

include/alpaka/dev/DevUniformCudaHipRt.hpp
fwyzard: I'm not sure I understand the difference ?

The current implementation uses a `static` data member in the CUDA or HIP device class.
My suggestion would use a `static` data member in the thread pool class.

And in both cases the pool is created only the first time it is needed.

include/alpaka/dev/DevUniformCudaHipRt.hpp
fwyzard: Mhm... is there any advantage from having multiple pools ? One per GPU, or even one per copy of the `DevUniformCudaHipRt` object that manages the same GPU ?

include/alpaka/dev/DevUniformCudaHipRt.hpp
psychocoderHPC: I was not aware that the current implementation is using a static variable to create the thread pool. 
IMO we should avoid using static states and applied for a change https://github.com/alpaka-group/alpaka/pull/1707/files#r867801735 but we should discuss it tomorrow in the developer meeting.

include/alpaka/dev/DevUniformCudaHipRt.hpp
psychocoderHPC: It keeps alpaka global state free and avoids issue if a device got [reset](https://github.com/alpaka-group/alpaka/blob/0819e45098d38dc3c62a3fcee1e513e0492658d9/include/alpaka/dev/DevUniformCudaHipRt.hpp#L161) by the user.

During think about the issue, I checked the code and realized that alpaka is not currently freeing the device.

j-stephan:
Superseded by #1719.

PR #1710, opened 2022-05-09
Introduce `alpaka::math::constants`

### LINKED ISSUE #1553: Avoid M_PI
`alpaka/rand/RandDefault.hpp` uses `M_PI` which is a non-standard extension. Please replace the constant by e.g. `boost::math::constants::pi`.

j-stephan:
Fixes #1553.

This PR introduces the `alpaka::math::constants` namespace which includes several predefined math constants. These are the intersection of the [C++20 math constants](https://en.cppreference.com/w/cpp/header/numbers) and the non-standard constants defined by [libc](https://www.gnu.org/software/libc/manual/html_node/Mathematical-Constants.html) and [CRT](https://docs.microsoft.com/en-us/cpp/c-runtime-library/math-constants?view=msvc-170).

C++20 constants are now preferred. If they are unavailable, `M_PI` and friends will be used. In this case `_USE_MATH_DEFINES` will be added to the MSVC (or nvcc + MSVC) compiler command line.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -325,6 +325,25 @@ if(${alpaka_DEBUG} GREATER 1)
 endif()
 
 #-------------------------------------------------------------------------------
+# If available, use C++20 math constants. Otherwise, fall back to M_PI etc.
+if(${alpaka_CXX_STANDARD} VERSION_LESS "20")
+    set(alpaka_HAS_STD_MATH_CONSTANTS FALSE)
+else()
+    try_compile(alpaka_HAS_STD_MATH_CONSTANTS # Result stored here
+                "${PROJECT_BINARY_DIR}/alpakaFeatureTests" # Binary directory for output file
+                SOURCES "${_alpaka_FEATURE_TESTS_DIR}/MathConstants.cpp" # Source file
+                CXX_STANDARD 20
+                CXX_STANDARD_REQUIRED TRUE
+                CXX_EXTENSIONS FALSE)
+endif()
+
+if(NOT alpaka_HAS_STD_MATH_CONSTANTS)
+    message(STATUS "C++20 math constants not found. Falling back to non-standard constants.")
+    # Enable non-standard constants for MSVC.
+    target_compile_definitions(alpaka INTERFACE "$<$<OR:$<CXX_COMPILER_ID:MSVC>,$<AND:$<COMPILE_LANGUAGE:CUDA>,$<PLATFORM_ID:Windows>>>:_USE_MATH_DEFINES>")
+endif()
+
+#-------------------------------------------------------------------------------
 # Find TBB.
 if(alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE)
     find_package(TBB 2021.4.0.0 REQUIRED)
--- a/include/alpaka/math/Traits.hpp
+++ b/include/alpaka/math/Traits.hpp
@@ -14,9 +14,134 @@
 
 #include <cmath>
 #include <complex>
+#if __has_include(<numbers>)
+#    include <numbers>
+#endif
 
 namespace alpaka::math
 {
+    namespace constants
+    {
+        /* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because
+        these /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+#ifdef __cpp_lib_math_constants
+        inline constexpr double e = std::numbers::e;
+        inline constexpr double log2e = std::numbers::log2e;
+        inline constexpr double log10e = std::numbers::log10e;
+        inline constexpr double pi = std::numbers::pi;
+        inline constexpr double inv_pi = std::numbers::inv_pi;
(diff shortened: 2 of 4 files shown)


sbastrakov:
Nice! Why didn't you also add variable template versions (start of the list [here](https://en.cppreference.com/w/cpp/numeric/constants))?

j-stephan:
> Nice! Why didn't you also add variable template versions (start of the list [here](https://en.cppreference.com/w/cpp/numeric/constants))?

Because the non-standard counterparts are not available on all platforms. libc supports `M_PIf` (single precision) but CRT does not - they only have `double`.

sbastrakov:
But you could cast from a double version as a backup?

j-stephan:
Sure, that would work. I just wasn't sure about possible precision issues so I decided to leave this out. There is also the case of `long double` which is supported by C++20 and libc, but not CRT. Casting is likely not what we want here. Then again, I'm not sure we even care about `long double`...

bernhardmgruber:
Why all the work, if we can just use `boost::math::constants`? See: https://www.boost.org/doc/libs/1_79_0/libs/math/doc/html/math_toolkit/constants.html

sbastrakov:
But then it's a `Boost::Math` dependency, which we don't want i guess?

j-stephan:
We've been over this... no more Boost dependencies unless they have no alternative.

bernhardmgruber:
> But then it's a `Boost::Math` dependency, which we don't want i guess?

> We've been over this... no more Boost dependencies unless they have no alternative.

It's another Boost dependency to the ones we already have, but this time only for C++17. The dependency is gone in C++20.

j-stephan:
Each additional Boost dependency is a potential point of failure. I've recently been through the pain of getting Boost to work on a HPC cluster again (funnily enough, Boost.Math was broken). If we can avoid Boost (even "only" for C++17 which is our standard for probably the next five years) we should.

bernhardmgruber:
> Each additional Boost dependency is a potential point of failure.

So is each line of code we add ourselves :)

Anyway, go ahead with your PR!

include/alpaka/math/Traits.hpp
```diff
@@ -12,11 +12,54 @@
...
+#        include <cmath>
+#        include <numbers>
+#        define ALPAKA_HAS_STANDARD_CONSTANTS
+#    else
+// We are in C++20 mode but the stdlib does not support the math constants. Fall back to non-standard.
+#        include <cmath>
+#    endif
+#else
+#    include <cmath>
+#endif
```
bernhardmgruber:
```suggestion
#include <cmath>
#if __has_include(<numbers>)
#    include <numbers>
#endif
```

Please avoid checking the `__cplusplus` constant, as it is raised only after a standard is fully implemented. Individual C++20 features are usually available much sooner. Also, MSVC defines `__cplusplus` to `199711L` to not break compatibility, unless you compile with `/Zc:__cplusplus`.

include/alpaka/math/Traits.hpp
```diff
@@ -12,11 +12,54 @@
...
 {
+    namespace constants
+    {
+        /* TODO: Remove the following pragmas once support for clang 5 and 6 is removed. They are necessary because
+        these /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+#ifdef ALPAKA_HAS_STANDARD_CONSTANTS
```
bernhardmgruber:
```suggestion
#ifdef __cpp_lib_math_constants
```

cmake/alpakaCommon.cmake
```diff
@@ -324,6 +324,25 @@ if(${alpaka_DEBUG} GREATER 1)
...
+                CXX_STANDARD_REQUIRED TRUE
+                CXX_EXTENSIONS FALSE)
+endif()
+
+if(NOT alpaka_HAS_STD_MATH_CONSTANTS)
+    message(STATUS "C++20 math constants not found. Falling back to non-standard constants.")
+    # Enable non-standard constants for MSVC
+    target_compile_definitions(alpaka INTERFACE "$<$<OR:$<CXX_COMPILER_ID:MSVC>,$<AND:$<COMPILE_LANGUAGE:CUDA>,$<PLATFORM_ID:Windows>>>:_USE_MATH_DEFINES>")
+endif()
+
```
bernhardmgruber: How about just unconditionally providing `_USE_MATH_DEFINES` to MSVC and skip the entire feature detection? That would simplify cmake a lot here.

Also, we would need to run the `try_compile` only for MSVC or on Windows. No need to make the cmake configure slower on Linux.

sbastrakov: I don't think a library should enforce `_USE_MATH_DEFINES ` to a user's app or any other things like that unless they are required to use the library at all.

j-stephan: Yes, MSVC should only have this conditionally. Also I think it is still useful as status output for macOS / Linux users.

>  No need to make the cmake configure slower on Linux.

Compiling the test takes about 0.2 seconds with both gcc and clang. I think Linux users will survive this ;-)

bernhardmgruber: But my WSL is slow ... fine, go ahead! ;)

j-stephan:
I also added the variable template versions now @sbastrakov.

bernhardmgruber (approved):

PR #1712, opened 2022-05-10
Add example illustrating typical data-parallel patterns with alpaka

sbastrakov:
The example is not ready yet, but here is the current version. I feel for now the memory part is awkward and I have to somehow reformulate it.

The goals of the example are like what i described on the call:

- Demonstrate some standard techniques for new alpaka users, expecially not familiar with GPU programming
- Show our general idea of decoupling thread index space, user data index space, and memory index space. So far I didn't manage the memory part, but hope to with futher work.

### DIFF
--- a/example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
+++ b/example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
@@ -1,0 +1,453 @@
+/* Copyright 2022 Sergei Bastrakov
+ *
+ * This file exemplifies usage of alpaka.
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED “AS IS” AND ISC DISCLAIMS ALL WARRANTIES WITH
+ * REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
+ * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
+ * IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <alpaka/alpaka.hpp>
+#include <alpaka/example/ExampleDefaultAcc.hpp>
+
+#include <iostream>
+#include <typeinfo>
+
+//! Functor to process the given index in the user data domain.
+//! This example demonstrates patterns for data-parallel loops, so the function itself is not a focus.
+//! So we simply return some index-dependant value.
+//!
+//! Note that generally here we operate in the data domain, not in a thread index domain.
+//! Some thread will eventually be mapped to calculate this data domain point.
+//! This mapping and work distribution is defined by a combination of a kernel and its launch configuration.
+//! However, this function is generally independent of that.
+//! That is an important distinction that we underline with how the kernels in this examples are written.
+//!
+//! \param idx The index in the data domain to operate on.
+ALPAKA_FN_HOST_ACC float process(uint32_t idx)
+{
+    return static_cast<float>(idx + 3);
+}
+
+//! Test if the given buffer has a correct result.
+//!
+//! \tparam TQueue The queue type for work to be submitted to.
+//! \tparam TBufAcc The device buffer type.
+//! \param queue The queue object for work to be submitted to.
+//! \param bufAcc The device buffer.
+template<typename TQueue, typename TBufAcc>
+void testResult(TQueue& queue, TBufAcc& bufAcc)
+{
+    // Wait for kernel to finish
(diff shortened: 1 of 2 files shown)


sbastrakov:
Sorry it took longer than I thought. I have now pushed the version I think is final, hopefully CI will pass. I intend to finish this example next week if there are issues or review suggestions.

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
```diff
@@ -0,0 +1,452 @@
...
+//! \tparam TBufAcc The device buffer type.
+//! \param queue The queue object for work to be submitted to.
+//! \param bufAcc The device buffer.
+template<typename TAcc, typename TQueue, typename TBufAcc>
+void naiveOpenMPStyle(TQueue& queue, TBufAcc& bufAcc)
+{
+    auto const n = alpaka::getExtentProduct(bufAcc);
+    auto const deviceProperties = alpaka::getAccDevProps<TAcc>(alpaka::getDevByIdx<TAcc>(0u));
+    auto const maxThreadsPerBlock = deviceProperties.m_blockThreadExtentMax[0];
+    auto const numCores = 16u; // should be taken from hardware properties, alpaka currently does not expose it
```
sbastrakov: It is not great, but alpaka offers no way of doing it. I think we discussed that another library may do that. Note that it is not possible to take number of multiprocessors, as for CPU accelerators that would always be 1.

fwyzard: What about
```c++
auto const numCores = std::max(std::thread::hardware_concurrency(), 1u);
```
?

Otherwise if you want more information, I'd suggest `hwloc` as a third part library (but only for convenience, since it is already an optional dependency of TBB, OpenMPI, _etc_).

sbastrakov: Thanks, I will try your std:: suggestion! For hwloc I wouldn't like to make a dependency in an example

bernhardmgruber:
I would really like to see this being merged as an alpaka example. Furthermore, I suspect vikunja should use such implementations for the `for_each` primitive.

SimeonEhrig:
> I would really like to see this being merged as an alpaka example. Furthermore, I suspect vikunja should use such implementations for the `for_each` primitive.

vikunja is already using the [grid strided loop pattern](https://github.com/alpaka-group/vikunja/blob/d6cbf89d783cde91fe785eec4192142244c4e0e8/include/vikunja/access/BlockStrategy.hpp#L158) for tranform and reduce. And I thinking about, how I can use it in multi dimensional memory.

SimeonEhrig:
I also like the example really much. Just a idea, can we add some images to visualize the access pattern, like @bernhardmgruber  is it doing with llama?

bernhardmgruber:
> I also like the example really much. Just a idea, can we add some images to visualize the access pattern, like @bernhardmgruber is it doing with llama?

Here is your first student project! :)

bernhardmgruber:
Great, the failing nvc++ job's log is truncated after 4MB, and does not show an error :/

fwyzard:
> vikunja is already using the [grid strided loop pattern](https://github.com/alpaka-group/vikunja/blob/d6cbf89d783cde91fe785eec4192142244c4e0e8/include/vikunja/access/BlockStrategy.hpp#L158) for tranform and reduce. And I thinking about, how I can use it in multi dimensional memory.

I'll try to have a look (and maybe even understand...) what is used in vikunja !

In the meantime FYI here is what I coded for CMS, to be used like `for (auto i : elements_with_stride(acc, size))`:
  - [`elements_with_stride`](https://github.com/cms-sw/cmssw/blob/master/HeterogeneousCore/AlpakaInterface/interface/workdivision.h#L79-L163) (1-dimension using scalar indices)
  - [`elements_with_stride_nd`](https://github.com/cms-sw/cmssw/blob/master/HeterogeneousCore/AlpakaInterface/interface/workdivision.h#L165-L262) (N-dimension using `Vec` indices)

bernhardmgruber:
> In the meantime FYI here is what I coded for CMS, to be used like `for (auto i : elements_with_stride(acc, size))`:
> 
> * [`elements_with_stride`](https://github.com/cms-sw/cmssw/blob/master/HeterogeneousCore/AlpakaInterface/interface/workdivision.h#L79-L163) (1-dimension using scalar indices)
> * [`elements_with_stride_nd`](https://github.com/cms-sw/cmssw/blob/master/HeterogeneousCore/AlpakaInterface/interface/workdivision.h#L165-L262) (N-dimension using `Vec` indices)

I like those and I have similar ones in LLAMA. We could consider adding those to alpaka as well!

The usual drawback in my experience is though that for the ND case with N > 1, the manually written loops perform better and are easier to vectorize for CPU targets, because the iterators have to store state and need extra branches, e.g.: whether the ND index reached the end of one dimension and another dimension should be incremented.

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
bernhardmgruber: I included this suggestion in the PR.

bernhardmgruber (approved):

SimeonEhrig:
@bernhardmgruber It was on my todo list to review it for today :sweat_smile: But you was faster.

bernhardmgruber:
> @bernhardmgruber It was on my todo list to review it for today 😅 But you was faster.

You can still review it and we can send a follow-up PR, if you have any concerns! I just really wanted to have this PR in, since I needed it again yesterday as an example of how to write a grid-strided loop with alpaka.

@sbastrakov if you read this, thank you again for creating this example! It helped me again yesterday!

SimeonEhrig:
No, it's fine. I had a quick look again and if you use it, there should be no big mistake inside it.

PR #1716, opened 2022-05-10
Refactor and fix CPU intrinsics

### LINKED ISSUE #1715: `ffs` CPU intrinsic broken in C++20 mode
In C++20 mode our implementation of the `ffs` intrinsics breaks for CPU back-ends:

```
In file included from /home/runner/work/alpaka/alpaka/include/alpaka/acc/AccCpuFibers.hpp:23,
                 from /home/runner/work/alpaka/alpaka/include/alpaka/alpaka.hpp:18,
                 from /home/runner/work/alpaka/alpaka/example/complex/src/complex.cpp:18:
/home/runner/work/alpaka/alpaka/include/alpaka/intrinsic/IntrinsicCpu.hpp: In static member function 'static int32_t alpaka::trait::Ffs<alpaka::IntrinsicCpu>::ffs(const alpaka::IntrinsicCpu&, int32_t)':
/home/runner/work/alpaka/alpaka/include/alpaka/intrinsic/IntrinsicCpu.hpp:72:46: error: no matching function for call to 'countr_zero(int32_t&)'
   72 |                 return std::countr_zero(value);
      |                                         
```

We are using a `int32_t` parameter here but `std::countr_zero` is only defined for unsigned types: https://en.cppreference.com/w/cpp/numeric/countr_zero

bernhardmgruber:
This PR fixes the `ffs` implementation in C++20 and refactors the code around it.

Fixes: #1715

### DIFF
--- a/include/alpaka/intrinsic/IntrinsicCpu.hpp
+++ b/include/alpaka/intrinsic/IntrinsicCpu.hpp
@@ -14,6 +14,7 @@
 #include <alpaka/intrinsic/Traits.hpp>
 
 #include <bitset>
+#include <climits>
 #if __has_include(<bit>)
 #    include <bit>
 #endif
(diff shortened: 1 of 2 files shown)


include/alpaka/intrinsic/IntrinsicCpu.hpp
```diff
@@ -34,71 +35,49 @@ namespace alpaka
...
         template<>
         struct Ffs<IntrinsicCpu>
         {
-            static auto ffs(IntrinsicCpu const& /*intrinsic*/, std::int32_t value) -> std::int32_t
+            template<typename Integral>
+            static auto ffs(IntrinsicCpu const& /*intrinsic*/, Integral value) -> std::int32_t
             {
 #ifdef __cpp_lib_bitops
-                return std::countr_zero(value);
+                return std::countr_zero(static_cast<std::make_unsigned_t<T>>(value)) + 1;
```
j-stephan: What is `T` referring to here? Looks like it should be `Integral` instead.

j-stephan: I also think that the semantics are wrong here.  `ffs(0)` should return 0. `std::countr_zero(0) + 1` would return `33` instead when applied to `int32_t{0}`.

j-stephan (changes requested):

include/alpaka/intrinsic/IntrinsicCpu.hpp
bernhardmgruber: Good catch! We definitely need your C++20 tests as part of the CI. I just compiled locally with C++20 enabled and now see the error in the unit tests.

j-stephan (approved):

PR #1718, opened 2022-05-11
Drop Boost.fiber back-end

### LINKED ISSUE #22: parallelize fiber block execution
Enhance the fibers implementation by parallelizing the execution of the blocks.

### LINKED ISSUE #34: execute fibers row by row
By executing the fibers randomly we prevent memory prefetching.
Iterating X 1st, Y 2nd, Z 3rd (native C memory order) we would assist the prefetcher by using the expected default access pattern.

### LINKED ISSUE #1714: [RFC] Future of the Boost.fiber back-end
While working on #1713 I discovered that the Boost.fiber back-end is broken when enabling C++20. This was fixed in their repository a few days ago but all stable versions including the current one are unusable with C++20.

So we can either:
* raise the minimum Boost version to 1.80 (or whenever the fix lands in a stable release)
* remove the Boost.fiber back-end
* deactivate the Boost.fiber back-end for C++20 and Boost versions < (fixed version)

Before someone puts any work into this: Do we want to continue to support Boost.fiber as a back-end? What is the use case? Which benefits does this back-end have over our OpenMP and TBB back-ends?

j-stephan:
This PR removes the Boost.fiber back-end.

Closes #22. Closes #34. Fixes #1714.


script/install_boost.sh
```diff
@@ -145,7 +144,7 @@ then
     then
         ALPAKA_BOOST_B2_CXXFLAGS+=" -std=c++17"
     fi
-    ALPAKA_BOOST_B2+=" --with-fiber --with-context --with-thread --with-atomic --with-system --with-chrono --with-date_time"
+    ALPAKA_BOOST_B2+=" --with-context --with-thread --with-atomic --with-system --with-chrono --with-date_time"
```
bernhardmgruber:
```suggestion
    ALPAKA_BOOST_B2+="--with-atomic"
```
Looking at https://pdimov.github.io/boostdep-report/boost-1.70.0/atomic.html, no built dependencies should be needed for Boost.Atomic.

bernhardmgruber (dismissed):

bernhardmgruber (dismissed):

j-stephan:
Don't merge this just yet. We are waiting for @tonydp03 trying out the Fiber back-end with the CMS software.

psychocoderHPC:
I set the PR to draft to avoid accidentally merging.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -71,8 +71,7 @@ namespace alpaka::core
         };
 
         //! ITaskPkg.
-        // \NOTE: We can not use std::packaged_task as it forces the use of std::future
-        // but we additionally support boost::fibers::promise.
+        // \NOTE: We can not use std::packaged_task as it forces the use of std::future.
```
bernhardmgruber, line 74: This comment is no longer true, because without the fiber backend, we can actually use `std::packaged_task` and `std::future`. But no need to fix this now.

script/install_boost.sh
```diff
@@ -95,7 +94,7 @@ else
 fi
 
 # Only build boost if we need some of the non-header-only libraries
-if [ "${ALPAKA_CI_INSTALL_FIBERS}" == "ON" ] || [ "${ALPAKA_CI_INSTALL_ATOMIC}" == "ON" ]
+if [ "${ALPAKA_CI_INSTALL_ATOMIC}" == "ON" ]
```
bernhardmgruber, line 97: So for all C++20 builds we don't need to compile boost anymore? Nice!

bernhardmgruber (dismissed):

bernhardmgruber (approved):

PR #1719, opened 2022-05-11
Add CallbackThread to QueueUniformCudaHipRt

### LINKED ISSUE #391: do not create new std::thread for each registered CUDA stream callback

tonydp03:
Solves #391. The `CallbackThread` class has an `std::thread` that is only initialized if needed by a Cuda/Hip callback. An object of this class is a data member of  `QueueUniformCudaHipRtImpl` and accessed through a `QueueUniformCudaHipRt` object.

Previous PR #1707 can be closed (linked here for reference for future developments).

### DIFF
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -1,0 +1,78 @@
+/* Copyright 2022 Antonio Di Pilato
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <atomic>
+#include <condition_variable>
+#include <functional>
+#include <future>
+#include <mutex>
+#include <queue>
+#include <thread>
+
+namespace alpaka::core
+{
+    class CallbackThread
+    {
+        using Task = std::packaged_task<void()>;
+
+    public:
+        ~CallbackThread()
+        {
+            m_stop = true;
+            m_cond.notify_one();
+            if(m_thread.joinable())
+                m_thread.join();
+        }
+        auto submit(Task&& newTask) -> std::future<void>
+        {
+            auto f = newTask.get_future();
+            {
+                std::unique_lock<std::mutex> lock{m_mutex};
+                m_tasks.emplace(std::move(newTask));
+                if(!m_thread.joinable())
+                    startWorkerThread();
+            }
+            m_cond.notify_one();
+            return f;
+        }
+
+    private:
+        std::thread m_thread;
+        std::condition_variable m_cond;
+        std::mutex m_mutex;
+        std::atomic<bool> m_stop{false};
+        std::queue<Task> m_tasks;
+
+        auto startWorkerThread() -> void
+        {
+            m_thread = std::thread(
+                [this]
+                {
+                    Task task;
+                    while(true)
+                    {
+                        {
+                            std::unique_lock<std::mutex> lock{m_mutex};
+                            m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
+
+                            if(m_stop && m_tasks.empty())
+                                break;
+
+                            task = std::move(m_tasks.front());
+                            m_tasks.pop();
+                        }
+
+                        task();
+                    }
+                });
+        }
+    };
+} // namespace alpaka::core
--- a/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
+++ b/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
@@ -27,8 +27,11 @@
 #        include <alpaka/core/Hip.hpp>
 #    endif
 
+#    include <alpaka/core/CallbackThread.hpp>
+
 #    include <condition_variable>
 #    include <functional>
+#    include <future>
 #    include <memory>
 #    include <mutex>
 #    include <thread>
@@ -85,6 +88,8 @@ namespace alpaka
 
         public:
             DevUniformCudaHipRt<TApi> const m_dev; //!< The device this queue is bound to.
+            core::CallbackThread m_callbackThread;
+
         private:
             typename TApi::Stream_t m_UniformCudaHipQueue;
         };
@@ -113,6 +118,10 @@ namespace alpaka
             [[nodiscard]] auto getNativeHandle() const noexcept
             {
                 return m_spQueueImpl->getNativeHandle();
+            }
+            auto getCallbackThread() -> core::CallbackThread&
+            {
+                return m_spQueueImpl->m_callbackThread;
             }
 
         public:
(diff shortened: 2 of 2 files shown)


include/alpaka/core/CallbackThread.hpp
```diff
@@ -0,0 +1,77 @@
...
+
+                                    if(m_stop && m_tasks.empty())
+                                        break;
+
+                                    task = std::move(m_tasks.front());
+                                    m_tasks.pop();
+                                }
+
+                                task();
+                            }
```
fwyzard: Looks good to me.

Personally I would find the code a bit easier to understand it it were written without the lambda and the unique_lock block, like
```c++
    Task task;
    while(true)
    { 
        std::unique_lock<std::mutex> lock{m_mutex};
        while (not m_stop and m_tasks.empty()) {
            m_cond.wait(lock);
        } 
            
        if (m_stop and m_tasks.empty())
            break;
                
        task = std::move(m_tasks.front());
        m_tasks.pop();

        // run the task without holding the lock
        lock.unlock();
        task();
    }   
```
but I suspect it's just a matter of taste.
No need to change it.

include/alpaka/core/CallbackThread.hpp
```diff
@@ -0,0 +1,77 @@
...
+            }
+            m_cond.notify_one();
+            return f;
+        }
+
+    private:
+        std::thread m_thread;
+        std::condition_variable m_cond;
+        std::mutex m_mutex;
+        bool m_stop{false};
```
fwyzard: If you make `m_stop` an `std::atomic_bool` or an `std::atomic_flag` you can avoid the lock in the destructor.
Not sure it's worth the change, though.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -85,6 +88,8 @@ namespace alpaka
 
         public:
             DevUniformCudaHipRt<TApi> const m_dev; //!< The device this queue is bound to.
+            core::CallbackThread c_thread;
```
fwyzard:
```suggestion
            core::CallbackThread m_thread;
```

fwyzard: or something more descriptive, like `m_callbackThread`

fwyzard:
```
 /home/runner/work/alpaka/alpaka/include/alpaka/core/CallbackThread.hpp:51:66: error: declaration shadows a local variable [-Werror,-Wshadow-uncaptured-local]
                                    std::unique_lock<std::mutex> lock{m_mutex};
                                                                 ^
/home/runner/work/alpaka/alpaka/include/alpaka/core/CallbackThread.hpp:40:46: note: previous declaration is here
                std::unique_lock<std::mutex> lock{m_mutex};
```

Looks like some (older ?) compilers are getting confused about the local `lock` variable inside the lambda and the one outside it.

One easy fix is to rename one of the two variables.

An other possibility which would also make the overall code clearer is to move the function that runs inside the thread to a separate method. Something like
```c++
      ...
        if(!m_thread.joinable()) {
          startWorkerThread();
        }
      ...

private:
    startWorkerThread() {
          m_thread = std::thread([this]() {...});
    }
```

include/alpaka/core/CallbackThread.hpp
tonydp03: I followed the same style that is adopted somewhere else in alpaka, i.e. [here](https://github.com/alpaka-group/alpaka/blob/2fc0f04f45a7ea0a7e4ae7d3a5bfd1f031f56fb0/include/alpaka/core/ConcurrentExecPool.hpp#L479-L484), but I agree that the `unlock` method is easier to read.

psychocoderHPC: I prefer the implementation from @tonydp03 because it is moving the handling of false wakeups to the std implementation: 
https://en.cppreference.com/w/cpp/thread/condition_variable/wait is pointing out that `cond.wait(mutex,prediction) is equal to what @fwyzard wrote.

include/alpaka/core/CallbackThread.hpp
j-stephan: It is worthwhile. Using an atomic instead of a lock gives us more exception safety in the destructor. (Not that we could do anything sensible when one of these standard library calls fails, but it is nice to be more "legal".)

bernhardmgruber: However, changing it to an atomic makes the destructor faster, but pessimizes all loads to `m_stop` inside the thread loop. It is also an unusual design to me.

bernhardmgruber: > Using an atomic instead of a lock gives us more exception safety in the destructor.

How so? The ctor and dtor of `unique_lock` are `noexcept`.

include/alpaka/core/CallbackThread.hpp
bernhardmgruber: Please keep the lambda in `m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });`. It is necessary to protect against spurious wakeups.

I would prefer the explicit block to calling `unlock()` manually, but also a matter of taste.

include/alpaka/core/CallbackThread.hpp
j-stephan: No, the ctor is not `noexcept` when taking a mutex. See entry (3) [here](https://en.cppreference.com/w/cpp/thread/unique_lock/unique_lock)

bernhardmgruber: It seems I slipped one row below. Sorry for the fuss!

include/alpaka/core/CallbackThread.hpp
fwyzard: > It is necessary to protect against spurious wakeups.

The loop and the lambda are equivalent for that purpose (replacing the loop with an if would indeed be unsafe).
I personally find the loop more readable (one less `not` to keep in mind), but as I wrote it was just a mutter of understanding the code.

Does the fact that the lambda captures `this` (by value) make the code less optimal than the loop ?

bernhardmgruber (approved):

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
bernhardmgruber: @tonydp03 please address this! `c_` is not a good prefix, since it is easily confused with a constant in Hungarian notation.

bernhardmgruber (changes requested):

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
tonydp03: yes, I modified the name in the class instead of here. My bad, sorry :)

bernhardmgruber (approved):

j-stephan (approved):

PR #1721, opened 2022-05-12
Small fixes required for gcc-12

### LINKED ISSUE #1720: `alpaka::Complex<T>` and `std::complex<T>` are not layout compatible
gcc-12 landed in my distro today and I gave the alpaka test suite a try (in C++20 mode). This triggered the following error:

```
[ 22%] Building CXX object test/unit/math/CMakeFiles/mathTest.dir/src/mathComplexDouble.cpp.o
In file included from /home/stepha27/workspace/caravan/alpaka/test/unit/math/src/TestTemplate.hpp:19,
                 from /home/stepha27/workspace/caravan/alpaka/test/unit/math/src/mathComplexDouble.cpp:11:
/home/stepha27/workspace/caravan/alpaka/test/unit/math/src/mathComplexDouble.cpp: In instantiation of 'void C_A_T_C_H_T_E_M_P_L_A_T_E_T_E_S_T_F_U_N_C_7() [with TestType = alpaka::AccCpuSerial<std::integral_constant<long unsigned int, 1>, long unsigned int>]':
/home/stepha27/workspace/caravan/alpaka/test/unit/math/src/mathComplexDouble.cpp:41:1:   required from 'void {anonymous}::ns_C_A_T_C_H_T_E_M_P_L_A_T_E_T_E_S_T_8::C_A_T_C_H_T_E_M_P_L_A_T_E_T_E_S_T_8<Types>::reg_tests() [with Types = {alpaka::AccCpuSerial<std::integral_constant<long unsigned int, 1>, long unsigned int>, alpaka::AccCpuThreads<std::integral_constant<long unsigned int, 1>, long unsigned int>, alpaka::AccCpuTbbBlocks<std::integral_constant<long unsigned int, 1>, long unsigned int>, alpaka::AccCpuOmp2Blocks<std::integral_constant<long unsigned int, 1>, long unsigned int>, alpaka::AccCpuOmp2Threads<std::integral_constant<long unsigned int, 1>, long unsigned int>}]'
/home/stepha27/workspace/caravan/alpaka/test/unit/math/src/mathComplexDouble.cpp:41:1:   required from here
/home/stepha27/workspace/caravan/alpaka/test/unit/math/src/mathComplexDouble.cpp:43:25: error: static assertion failed: std::is_layout_compatible_v<alpaka::Complex<double>, std::complex<double>>
   43 |     STATIC_REQUIRE(std::is_layout_compatible_v<alpaka::Complex<double>, std::complex<double>>);
      |                    ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/stepha27/workspace/caravan/alpaka/test/unit/math/src/mathComplexDouble.cpp:43:25: note: 'std::is_layout_compatible_v<alpaka::Complex<double>, std::complex<double> >' evaluates to false
```

If I do a parallel build the same issue appears for `float`. @sbastrakov

j-stephan:
Fixes #1720.

1. Our layout tests for our complex implementation were too strict. They now check for the requirements without failing to compile.
2. gcc-12 complained about two constructors not being `noexcept`. Added `noexcept` to make gcc happy.

### DIFF
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -55,7 +55,7 @@ namespace alpaka
                 DevCpu dev,
                 TElem* pMem,
                 std::function<void(TElem*)> deleter,
-                TExtent const& extent)
+                TExtent const& extent) noexcept
                 : m_dev(std::move(dev))
                 , m_extentElements(getExtentVecEnd<TDim>(extent))
                 , m_pMem(pMem)
--- a/test/unit/math/src/mathComplexDouble.cpp
+++ b/test/unit/math/src/mathComplexDouble.cpp
@@ -1,4 +1,4 @@
-/** Copyright 2022 Jakob Krude, Benjamin Worpitz, Bernhard Manfred Gruber, Sergei Bastrakov
+/** Copyright 2022 Jakob Krude, Benjamin Worpitz, Bernhard Manfred Gruber, Sergei Bastrakov, Jan Stephan
  *
  * This file is part of alpaka.
  *
@@ -37,9 +37,24 @@ TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[math] [operator]", TestAccFunc
     testTemplate.template operator()<alpaka::Complex<double>>();
 }
 
-#ifdef __cpp_lib_is_layout_compatible
-TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[layout]", TestAccs)
+TEST_CASE("mathArrayOrientedComplexDouble", "[array-oriented]")
 {
-    STATIC_REQUIRE(std::is_layout_compatible_v<alpaka::Complex<double> std::complex<double>>);
+    // Ensure that our implementation matches the behaviour of std::complex with regard to array-oriented access.
+    // See https://en.cppreference.com/w/cpp/numeric/complex - Array-oriented access - for more information.
+    auto const complex_alpaka = alpaka::Complex<double>{42., 42.};
+    auto const complex_std = std::complex<double>{42., 42.};
+
+    auto const real_alpaka = reinterpret_cast<double const(&)[2]>(complex_alpaka)[0];
+    auto const real_std = reinterpret_cast<double const(&)[2]>(complex_std)[0];
+    REQUIRE(alpaka::math::floatEqualExactNoWarning(real_alpaka, real_std));
+
+    auto const imag_alpaka = reinterpret_cast<double const(&)[2]>(complex_alpaka)[1];
+    auto const imag_std = reinterpret_cast<double const(&)[2]>(complex_std)[1];
+    REQUIRE(alpaka::math::floatEqualExactNoWarning(imag_alpaka, imag_std));
 }
-#endif
+
+TEST_CASE("mathPaddingComplexDouble", "[padding]")
+{
+    // Ensure that we don't accidentally introduce padding
+    STATIC_REQUIRE(sizeof(alpaka::Complex<double>) == 2 * sizeof(double));
+}
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -241,26 +241,30 @@ if(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE OR
    alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR
    alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE)
 
-    # Check for C++20 std::atomic_ref first
-    if(${alpaka_CXX_STANDARD} VERSION_GREATER_EQUAL "20")
-        try_compile(alpaka_HAS_STD_ATOMIC_REF # Result stored here
-                    "${PROJECT_BINARY_DIR}/alpakaFeatureTests" # Binary directory for output file
-                    SOURCES "${_alpaka_FEATURE_TESTS_DIR}/StdAtomicRef.cpp" # Source file
-                    CXX_STANDARD 20
-                    CXX_STANDARD_REQUIRED TRUE
-                    CXX_EXTENSIONS FALSE)
-        if(alpaka_HAS_STD_ATOMIC_REF)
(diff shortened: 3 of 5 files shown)


include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -55,7 +55,7 @@ namespace alpaka
                 DevCpu dev,
                 TElem* pMem,
                 std::function<void(TElem*)> deleter,
-                TExtent const& extent)
+                TExtent const& extent) noexcept
```
bernhardmgruber, line 58: Technically, the constructor uses `std::cout << ...`, which is not `noexcept`. I don't expect it to throw though :) If that pleases g++-12 so be it.

test/unit/math/src/mathComplexDouble.cpp
```diff
@@ -37,9 +37,18 @@ TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[math] [operator]", TestAccFunc
     testTemplate.template operator()<alpaka::Complex<double>>();
 }
 
-#ifdef __cpp_lib_is_layout_compatible
-TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[layout]", TestAccs)
+TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[array-oriented]", TestAccs)
 {
-    STATIC_REQUIRE(std::is_layout_compatible_v<alpaka::Complex<double> std::complex<double>>);
```
bernhardmgruber, line 43: I think the check for `is_layout_compatible` should be correct. However, it fails with gcc and MSVC: https://godbolt.org/z/vdKeGenGr

Maybe a compiler bug in both cases? Can you please comment out the line instead of removing and add a comment that we should revisit this in the future?

test/unit/math/src/mathComplexDouble.cpp
```diff
@@ -37,9 +37,18 @@ TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[math] [operator]", TestAccFunc
...
+    auto const complex_alpaka = alpaka::Complex<double>{42., 42};
+    auto const complex_std = std::complex<double>{42., 42.};
+
+    auto const real_alpaka = reinterpret_cast<double const(&)[2]>(complex_alpaka)[0];
+    auto const real_std = reinterpret_cast<double const(&)[2]>(complex_std)[0];
+    REQUIRE(alpaka::math::floatEqualExactNoWarning(real_alpaka, real_std));
+
+    auto const imag_alpaka = reinterpret_cast<double const(&)[2]>(complex_alpaka)[1];
+    auto const imag_std = reinterpret_cast<double const(&)[2]>(complex_std)[1];
+    REQUIRE(alpaka::math::floatEqualExactNoWarning(imag_alpaka, imag_std));
```
bernhardmgruber:
```suggestion
    auto const complex_alpaka = alpaka::Complex<double>{42., 43};
    auto const complex_std = std::complex<double>{42., 43.};
    STATIC_REQUIRE(sizeof(complex_alpaka) == sizeof(double) * 2);
    STATIC_REQUIRE(sizeof(complex_alpaka) == sizeof(complex_std));
    REQUIRE(!std::memcmp(&complex_alpaka, &complex_std, sizeof(complex_std)));
```
I would rather just check their sizes and byte contents are the same. Also include the check that our complex is really just 2 doubles etc. So we know there is no padding anywhere.

test/unit/math/src/mathComplexDouble.cpp
j-stephan: See #1720 for the discussion about this.

include/alpaka/mem/buf/BufCpu.hpp
j-stephan: Well, we only `std::cout` stuff when we're in debug mode. When we're in debug mode and `cout` throws we probably have a much worse problem at hand ;-)

test/unit/math/src/mathComplexDouble.cpp
bernhardmgruber: Oh, thx! I guess `layout_compatible` might be too strict indeed.

test/unit/math/src/mathComplexDouble.cpp
j-stephan: Good point about the padding. Right now we're not ensuring that our complex is not padded. This might be a problem with `long double` which our complex supports.

bernhardmgruber: I think you can keep your checks since they are also listed on cppreference, as discussed in #1720. But consider adding the check `STATIC_REQUIRE(sizeof(complex_alpaka) == sizeof(double) * 2);` so we know that we did not introduce any additional padding.

j-stephan: Done.

test/unit/math/src/mathComplexDouble.cpp
```diff
@@ -37,9 +37,24 @@ TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[math] [operator]", TestAccFunc
...
+}
+
+TEMPLATE_LIST_TEST_CASE("mathOpsComplexDouble", "[padding]", TestAccs)
+{
+    // Ensure that we don't accidentally introduce padding
+    STATIC_REQUIRE(sizeof(alpaka::Complex<double>) == 2 * sizeof(double));
 }
```
bernhardmgruber:
```suggestion
TEST_CASE("mathOpsComplexDouble", "[padding]")
{
    // Ensure that we don't accidentally introduce padding
    STATIC_REQUIRE(sizeof(alpaka::Complex<double>) == 2 * sizeof(double));
}
```
No need to test for all accelerators :)

bernhardmgruber (dismissed):

test/unit/math/src/mathComplexDouble.cpp
j-stephan: Good catch!

j-stephan:
I also just noticed an issue with the lookup of `std::atomic_ref<T>` when in C++20 mode. I fixed this now.

bernhardmgruber (approved):

PR #1725, opened 2022-05-16
fix host memory alloc and pinning on OpenPower systems

psychocoderHPC:
The alignment requirements to pin host memory on OpenPower systems are not fulfilled.

We tried to fix this issue already in the past in #873. Unfortunately, the issue still exists and was reproducible with PIConGPU and cuda-memcheck. After applying this fix the error is solved.

~~This PR introduces a new boost dependency to the header only lib `boost::interprocess` :-(.
I thought about fixing it with macros guards but we do not know if we will run on ARM+CUDA again into this issue or if the system is running with huge pages we will definitely have this issue again, therefore I decided to use boost.~~

@j-stephan Can we discuss in the next developer meeting if we create with this fix a bug fix release 0.9.1, else we need to switch with PIConGPU to the current alpaka dev for 1.0.

### DIFF
--- a/include/alpaka/dev/cpu/SysInfo.hpp
+++ b/include/alpaka/dev/cpu/SysInfo.hpp
@@ -123,6 +123,26 @@ namespace alpaka
                 return std::string("unknown");
 #endif
             }
+
+            //! \return Pagesize in bytes used by the system.
+            inline size_t getPageSize()
+            {
+#if BOOST_OS_WINDOWS || BOOST_OS_CYGWIN
+                SYSTEM_INFO si;
+                GetSystemInfo(&si);
+                return si.dwPageSize;
+#elif BOOST_OS_UNIX || BOOST_OS_MACOS
+#    if defined(_SC_PAGESIZE)
+                return static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+#    else
+                return = static_cast<size_t>(getpagesize());
+#    endif
+#else
+#    error "getPageSize not implemented for this system!"
+                return 0;
+#endif
+            }
+
             //! \return The total number of bytes of global memory.
             //! Adapted from David Robert Nadeau:
             //! http://nadeausoftware.com/articles/2012/09/c_c_tip_how_get_physical_memory_size_system
@@ -156,21 +176,17 @@ namespace alpaka
                 std::uint64_t size(0);
                 std::size_t sizeLen{sizeof(size)};
                 if(sysctl(mib, 2, &size, &sizeLen, nullptr, 0) < 0)
-                {
                     throw std::logic_error("getTotalGlobalMemSizeBytes failed calling sysctl!");
-                }
                 return static_cast<std::size_t>(size);
 
 #    elif defined(_SC_AIX_REALMEM) // AIX.
                 return static_cast<std::size_t>(sysconf(_SC_AIX_REALMEM)) * static_cast<std::size_t>(1024);
 
-#    elif defined(_SC_PHYS_PAGES) && defined(_SC_PAGESIZE) // Linux, FreeBSD, OpenBSD, Solaris.
-                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES))
-                    * static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+#    elif defined(_SC_PHYS_PAGES) // Linux, FreeBSD, OpenBSD, Solaris.
+                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES)) * getPageSize();
 
-#    elif defined(_SC_PHYS_PAGES) && defined(_SC_PAGE_SIZE) // Legacy.
-                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES))
-                    * static_cast<std::size_t>(sysconf(_SC_PAGE_SIZE));
+#    elif defined(_SC_PHYS_PAGES) // Legacy.
+                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES)) * getPageSize();
 
 #    elif defined(CTL_HW)                                                                                             \
         && (defined(HW_PHYSMEM) || defined(HW_REALMEM)) // FreeBSD, DragonFly BSD, NetBSD, OpenBSD, and OSX.
@@ -185,9 +201,7 @@ namespace alpaka
             std::uint32_t size(0);
             std::size_t const sizeLen{sizeof(size)};
             if(sysctl(mib, 2, &size, &sizeLen, nullptr, 0) < 0)
-            {
                 throw std::logic_error("getTotalGlobalMemSizeBytes failed calling sysctl!");
-            }
             return static_cast<std::size_t>(size);
 #    endif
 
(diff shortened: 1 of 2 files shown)


include/alpaka/mem/alloc/AllocCpuAligned.hpp
```diff
@@ -37,16 +39,16 @@ namespace alpaka
...
                 // From testing with PIConGPU and cuda-memcheck we found out that the alignment is still required.
                 //
                 // For HIP the required alignment is the size of a cache line.
                 // https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674
-                // To avoid issues with HIP(cuda) the alignment will be set also for HIP(clang)
-                // to 4kib.
-                // @todo evaluate requirements when the HIP ecosystem is more stable
-                constexpr size_t minAlignement = std::max<size_t>(TAlignment::value, 4096);
+                // On most x86 systems the page size is 64KiB and on OpenPower 65KiB.
+                // Page size can be tested on the terminal with: `getconf PAGE_SIZE`
```
bernhardmgruber: It's still 4KiB on my machine :)
```
bgruber@graviola:~$ getconf PAGE_SIZE
4096
```
Is that a typo?

psychocoderHPC:
mhh looks like the boost part is not header only. :-(

j-stephan:
I'm not sure I like Boost.Interprocess. This seems like a small-enough change to implement this ourselves (through POSIX and Win32-API).

include/alpaka/dev/cpu/SysInfo.hpp
```diff
@@ -123,6 +123,26 @@ namespace alpaka
                 return std::string("unknown");
 #endif
             }
+
+            //! \return Pagesize in bytes used by the system.
+            inline size_t getPageSize()
+            {
+#if BOOST_OS_WINDOWS || BOOST_OS_CYGWIN
```
j-stephan, line 130: Wow, we support Cygwin?

psychocoderHPC:
@j-stephan I implemented the `getPageSize()` method and used it for memory alignment and for other functions where we already used the page size for getting free memory.

include/alpaka/dev/cpu/SysInfo.hpp
```diff
@@ -123,6 +123,26 @@ namespace alpaka
...
+                SYSTEM_INFO si;
+                GetSystemInfo(&si);
+                return si.dwPageSize;
+#elif BOOST_OS_UNIX || BOOST_OS_MACOS
+#    if defined(_SC_PAGESIZE)
+                return static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+#    else
+                return = static_cast<size_t>(getpagesize());
+#    endif
```
bernhardmgruber: Consider adding a comment that `getpagesize` is deprecated and that's why we have two versions here.

include/alpaka/dev/cpu/SysInfo.hpp
```diff
@@ -123,6 +123,26 @@ namespace alpaka
...
             }
+
+            //! \return Pagesize in bytes used by the system.
+            inline size_t getPageSize()
+            {
+#if BOOST_OS_WINDOWS || BOOST_OS_CYGWIN
+                SYSTEM_INFO si;
+                GetSystemInfo(&si);
+                return si.dwPageSize;
+#elif BOOST_OS_UNIX || BOOST_OS_MACOS
```
j-stephan, line 134: Shouldn't `BOOST_OS_UNIX` also work for macOS?

include/alpaka/dev/cpu/SysInfo.hpp
```diff
@@ -123,6 +123,26 @@ namespace alpaka
...
+#    if defined(_SC_PAGESIZE)
+                return static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+#    else
+                return = static_cast<size_t>(getpagesize());
```
j-stephan: Are there still systems around which don't use `_SC_PAGESIZE`? `getpagesize()` was deprecated [before 1997](https://pubs.opengroup.org/onlinepubs/7908799/xsh/getpagesize.html) (see the copyright header).

include/alpaka/dev/cpu/SysInfo.hpp
```diff
@@ -156,21 +176,17 @@ namespace alpaka
...
-                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES))
-                    * static_cast<std::size_t>(sysconf(_SC_PAGESIZE));
+#    elif defined(_SC_PHYS_PAGES) // Linux, FreeBSD, OpenBSD, Solaris.
+                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES)) * getPageSize();
 
-#    elif defined(_SC_PHYS_PAGES) && defined(_SC_PAGE_SIZE) // Legacy.
-                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES))
-                    * static_cast<std::size_t>(sysconf(_SC_PAGE_SIZE));
+#    elif defined(_SC_PHYS_PAGES) // Legacy.
+                return static_cast<std::size_t>(sysconf(_SC_PHYS_PAGES)) * getPageSize();
```
j-stephan: Now we have two `#elif defined(_SC_PHYS_PAGES)`. I think we can merge those.

include/alpaka/mem/alloc/AllocCpuAligned.hpp
```diff
@@ -37,16 +38,15 @@ namespace alpaka
...
                 // this was described in older programming guides but was removed later.
                 // From testing with PIConGPU and cuda-memcheck we found out that the alignment is still required.
                 //
                 // For HIP the required alignment is the size of a cache line.
                 // https://rocm-developer-tools.github.io/HIP/group__Memory.html#gab8258f051e1a1f7385f794a15300e674
-                // To avoid issues with HIP(cuda) the alignment will be set also for HIP(clang)
-                // to 4kib.
-                // @todo evaluate requirements when the HIP ecosystem is more stable
-                constexpr size_t minAlignement = std::max<size_t>(TAlignment::value, 4096);
+                // On most x86 systems the page size is 64KiB and on OpenPower 65KiB.
```
j-stephan: I agree with Bernhard here. On x86 it is probably 4KiB.

j-stephan (changes requested):
Thanks! A few nitpicks:

include/alpaka/mem/alloc/AllocCpuAligned.hpp
psychocoderHPC: yes it schould be 4kib

include/alpaka/dev/cpu/SysInfo.hpp
psychocoderHPC: not really but since it was in the original code pathes in I keeped it

include/alpaka/dev/cpu/SysInfo.hpp
psychocoderHPC: not sure the mange page in linux says that with sysconf  you are more portable but getpagesize is still available in linux.

include/alpaka/dev/cpu/SysInfo.hpp
psychocoderHPC: I do not know I have no MAC and other code snippeds in alpaka use the same combination.

psychocoderHPC:
I updated the PR.

include/alpaka/dev/cpu/SysInfo.hpp
j-stephan: @psychocoderHPC You missed this review comment. We have two branches doing the same thing here.

psychocoderHPC: you are right :-(

j-stephan (approved):

PR #1727, opened 2022-05-21
BufCpu::deleter: rm ignore_unused lambda for captured queue

### LINKED ISSUE #1726: Clang 15 `-Wunused-lambda-capture` in `BufCpu::deleter`
This line triggers `error: lambda capture 'queue' is not used [-Werror,-Wunused-lambda-capture]` in clang 15(/main):
https://github.com/alpaka-group/alpaka/blob/22374dbfab3596c1bdda38fe90b4b40b9f6f43a1/include/alpaka/mem/buf/BufCpu.hpp#L283-L284

While this `-Werror` is my immediate concern, I think there is something else wrong here, namely that this line is either superfluous or we have potential a use-after-free and this line is not helping:
https://github.com/alpaka-group/alpaka/blob/22374dbfab3596c1bdda38fe90b4b40b9f6f43a1/include/alpaka/mem/buf/BufCpu.hpp#L275-L286

`queue` is held by the closure in line 278 and will thus "stay alive" at until after the lambda finished executing, assuming the caller holds a an instance of the closure type and not a reference.
* If the caller of the lambda is only holding a reference and the lifetime of the actual instance is not ensured, we have potential use-after-free no matter what.
* If the call to `free` in line 282 is asynchronous, line 284 is doing nothing do keep `queue` in scope until after `free` completes, even if it were not capturing a reference (by capturing a reference it does not increment the ref count of the `shared_ptr`).

I think these two lines should just be removed, but maybe I am missing something, hence the issue instead of a PR...

jkelling:
Remove replacement of `alpaka_ignore` from `BufCPU::deleter`: no compiler covered by the CI is raising an unused there. Maybe this is because the notice, that the capture-by-copy by itself has an effect due to the non-trivial copy ctor.

Also : move and rephrase the comment about why the queue is being captured to make it more clear.

Closes #1726

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -503,12 +503,14 @@ namespace alpaka::core
                 // Checks whether pool is being destroyed, if so, stop running (lazy check without mutex).
                 while(!m_bShutdownFlag)
                 {
-                    auto currentTaskPackage = std::shared_ptr<ITaskPkg>{nullptr};
-
-                    // Use popTask so we only ever have one reference to the ITaskPkg
-                    if(popTask(currentTaskPackage))
                     {
-                        currentTaskPackage->runTask();
+                        auto currentTaskPackage = std::shared_ptr<ITaskPkg>{nullptr};
+
+                        // Use popTask so we only ever have one reference to the ITaskPkg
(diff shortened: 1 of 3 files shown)


jkelling:
Only used this provide a "red" state for #1728 . Closing again.

Issue #1676, opened 2022-03-31
`QueueGenericThreadsNonBlocking` technically illegal with OpenMP


jkelling:
I ran into an issue with the new `FenceTest` which occasionally crashes after succeeding with NVHPC for any OpenMP backend. This test runs kernels in parallel using separate non-blocking queues. All OpenMP backends use `QueueGenericThreadsNonBlocking`, only with different device type (aliased as `QueueCpuNonBlocking` and `QueueOmp5NonBlocking`.

The reason for this crash is probably a race condition in the OpenMP runtime. Which has only one set of global ICVs and is not thread save. Apparently, the standard allows this implementation ([section 1.3](https://www.openmp.org/spec-html/5.2/openmpse3.html#x16-150001.3))
```
An OpenMP program begins as a single thread of execution, called an initial thread. An initial thread executes sequentially, as if the code encountered is part of an implicit task region, called an initial task region, that is generated by the implicit parallel region surrounding the whole program. 
```

I did bring up the possibility of having a non-blocking queue for OpenMP based on OpenMP tasks. Maybe this is even necessary to be compliant.

j-stephan:
While we are talking about legality: Is it even legal (from OpenMP's point of view) to have an activated OpenMP back-end together with the CUDA back-end which might spawn a `std::thread`? AFAIK OpenMP considers this undefined behaviour.

jkelling:
Spawning a non-OpenMP thread that does not itself use any OpenMP (directives or runtime API functions) may still be UB (not where to look for an explicit statement about this, it is not stated in the section cited above), but there is much less or no risk of issues arising from this in the OpenMP runtime, because they change anything the runtime is observing (unless the OS has hard limit on the number of threads which the runtime is assuming to have full control over, or something like this).

Using OpenMP functionality inside multiple threads is a different story, because it can to race-condition in non-thread-safe OpenMP runtime code or confuse the runtime's bookkeeping when it is invoked inside a thread it has not in its book.

jkelling:
There is, btw. a big caveat to adding a queue bases on OpenMP tasks: An non-blockig queue would require OpenMP 5.0's tasks. OpenMP 4.0 already has `depend` which would be required to create a chain of tasks per queue, but in 4.x `taskwait` always waits for all (child) tasks (of the current thread), thus a `wait` would always synchronize all queues while in 5.0 `taskwait` may also have a`depend` clause.

PR #1728, opened 2022-05-23
Make QueueGenericThreadsNonBlocking complete tasks on destruction.

### LINKED ISSUE #1726: Clang 15 `-Wunused-lambda-capture` in `BufCpu::deleter`

jkelling:
The deleter task of BudCpu dioes not need to hold a `shared_ptr` to the queue, because the queue manages the task, not vice-versa.

Also: bundling the removal of an `std::move()`  from a `return` statement from `TaskPkg` (belonging to `ConcurrentExecPool`) in here, which I came across investigating the issue. (I think this is too small of a fix for its own PR, thugh: *do not squash this*)

Replaces https://github.com/alpaka-group/alpaka/pull/1727

Closes https://github.com/alpaka-group/alpaka/issues/1726

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -503,12 +503,15 @@ namespace alpaka::core
                 // Checks whether pool is being destroyed, if so, stop running (lazy check without mutex).
                 while(!m_bShutdownFlag)
                 {
-                    auto currentTaskPackage = std::shared_ptr<ITaskPkg>{nullptr};
-
-                    // Use popTask so we only ever have one reference to the ITaskPkg
-                    if(popTask(currentTaskPackage))
+                    // enter anonymous scope to limit lifetime of `currentTaskPackage`: destroy before entering wait
                     {
-                        currentTaskPackage->runTask();
+                        auto currentTaskPackage = std::shared_ptr<ITaskPkg>{nullptr};
+
+                        // Use popTask so we only ever have one reference to the ITaskPkg
+                        if(popTask(currentTaskPackage))
+                        {
+                            currentTaskPackage->runTask();
+                        }
                     }
                     {
                         std::unique_lock<TMutex> lock(m_mtxWakeup);
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -323,10 +323,18 @@ namespace alpaka::core
                 return m_numActiveTasks == 0u;
             }
 
-            void detach(std::unique_ptr<Type>&& self)
+            void detach(std::shared_ptr<Type>&& self)
             {
                 m_self = std::move(self);
                 m_bDetachedFlag = true;
+            }
+
+            std::shared_ptr<Type> takeDetachHandle()
+            {
+                if(m_bDetachedFlag.exchange(false))
+                    return std::move(m_self);
+                else
+                    return nullptr;
             }
 
         private:
@@ -345,10 +353,10 @@ namespace alpaka::core
                     }
                     else
                     {
-                        if(m_bDetachedFlag.exchange(false))
+                        auto self = takeDetachHandle();
+                        if(self)
                         {
-                            // Pool was detached and is idle, delete
-                            auto self = std::move(m_self);
+                            // Pool was detached and is idle, stop and delete
                             return;
                         }
                         TYield::yield();
@@ -383,7 +391,7 @@ namespace alpaka::core
             std::atomic<std::uint32_t> m_numActiveTasks = 0u;
             std::atomic<bool> m_bShutdownFlag = false;
             std::atomic<bool> m_bDetachedFlag = false;
-            std::unique_ptr<Type> m_self = nullptr;
+            std::shared_ptr<Type> m_self = nullptr;
         };
 
         //! ConcurrentExecPool using a condition variable to wait for new work.
@@ -512,7 +520,7 @@ namespace alpaka::core
                 return m_numActiveTasks == 0u;
             }
 
-            void detach(std::unique_ptr<Type>&& self)
+            void detach(std::shared_ptr<Type>&& self)
             {
                 m_self = std::move(self);
 
@@ -521,6 +529,14 @@ namespace alpaka::core
                     m_bDetachedFlag = true;
                 }
                 m_cvWakeup.notify_one();
+            }
+
+            std::shared_ptr<Type> takeDetachHandle()
+            {
+                if(m_bDetachedFlag.exchange(false))
+                    return std::move(m_self);
+                else
+                    return nullptr;
             }
 
         private:
@@ -544,10 +560,10 @@ namespace alpaka::core
                         std::unique_lock<TMutex> lock(m_mtxWakeup);
                         if(m_qTasks.empty())
                         {
-                            if(m_bDetachedFlag)
+                            auto self = takeDetachHandle();
+                            if(self)
                             {
-                                // Pool was detached and is idle, delete
-                                auto self = std::move(m_self);
+                                // Pool was detached and is idle, stop and delete
                                 lock.unlock();
                                 return;
                             }
@@ -597,7 +613,7 @@ namespace alpaka::core
             std::atomic<bool> m_bShutdownFlag = false;
 
             std::atomic<bool> m_bDetachedFlag = false;
-            std::unique_ptr<Type> m_self = nullptr;
+            std::shared_ptr<Type> m_self = nullptr;
         };
     } // namespace detail
 } // namespace alpaka::core
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Benjamin Worpitz, René Widera, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2022 Benjamin Worpitz, René Widera, Jan Stephan, Bernhard Manfred Gruber, Jeffrey Kelling
  *
  * This file is part of alpaka.
  *
@@ -15,6 +15,7 @@
 // Therefore, we can not even parse those parts when compiling device code.
 #include <alpaka/core/BoostPredef.hpp>
 #include <alpaka/core/Common.hpp>
+#include <alpaka/core/ThreadTraits.hpp>
 
 #include <atomic>
 #include <exception>
@@ -228,6 +229,8 @@ namespace alpaka::core
             bool TisYielding = true>
         class ConcurrentExecPool final
         {
+            using Type = ConcurrentExecPool<TIdx, TConcurrentExec, TPromise, TYield, TMutex, TCondVar, TisYielding>;
+
         public:
             //! Creates a concurrent executor pool with a specific number of concurrent executors and a maximum
             //! number of queued tasks.
@@ -235,11 +238,7 @@ namespace alpaka::core
             //! \param concurrentExecutionCount
             //!    The guaranteed number of concurrent executors used in the pool.
             //!    This is also the maximum number of tasks worked on concurrently.
-            ConcurrentExecPool(TIdx concurrentExecutionCount)
-                : m_vConcurrentExecs()
-                , m_qTasks()
-                , m_numActiveTasks(0u)
-                , m_bShutdownFlag(false)
+            ConcurrentExecPool(TIdx concurrentExecutionCount) : m_vConcurrentExecs(), m_qTasks()
             {
                 if(concurrentExecutionCount < 1)
                 {
@@ -324,6 +323,12 @@ namespace alpaka::core
                 return m_numActiveTasks == 0u;
             }
 
+            void detach(std::unique_ptr<Type>&& self)
+            {
+                m_self = std::move(self);
+                m_bDetachedFlag = true;
+            }
+
         private:
             //! The function the concurrent executors are executing.
             void concurrentExecFn()
@@ -340,6 +345,12 @@ namespace alpaka::core
                     }
                     else
                     {
+                        if(m_bDetachedFlag.exchange(false))
+                        {
+                            // Pool was detached and is idle, delete
+                            auto self = std::move(m_self);
+                            return;
+                        }
                         TYield::yield();
                     }
                 }
@@ -350,7 +361,10 @@ namespace alpaka::core
             {
                 for(auto&& concurrentExec : m_vConcurrentExecs)
                 {
-                    concurrentExec.join();
+                    if(!isThisThread(concurrentExec))
+                        concurrentExec.join();
+                    else
+                        concurrentExec.detach();
                 }
             }
             //! Pops a task from the queue.
@@ -366,8 +380,10 @@ namespace alpaka::core
         private:
             std::vector<TConcurrentExec> m_vConcurrentExecs;
             ThreadSafeQueue<std::shared_ptr<ITaskPkg>> m_qTasks;
-            std::atomic<std::uint32_t> m_numActiveTasks;
-            std::atomic<bool> m_bShutdownFlag;
+            std::atomic<std::uint32_t> m_numActiveTasks = 0u;
+            std::atomic<bool> m_bShutdownFlag = false;
+            std::atomic<bool> m_bDetachedFlag = false;
+            std::unique_ptr<Type> m_self = nullptr;
         };
 
         //! ConcurrentExecPool using a condition variable to wait for new work.
@@ -387,6 +403,8 @@ namespace alpaka::core
             typename TCondVar>
         class ConcurrentExecPool<TIdx, TConcurrentExec, TPromise, TYield, TMutex, TCondVar, false> final
         {
+            using Type = ConcurrentExecPool<TIdx, TConcurrentExec, TPromise, TYield, TMutex, TCondVar, false>;
+
         public:
             //! Creates a concurrent executors pool with a specific number of concurrent executors and a maximum
             //! number of queued tasks.
@@ -397,10 +415,8 @@ namespace alpaka::core
             ConcurrentExecPool(TIdx concurrentExecutionCount)
                 : m_vConcurrentExecs()
                 , m_qTasks()
-                , m_numActiveTasks(0u)
                 , m_mtxWakeup()
                 , m_cvWakeup()
-                , m_bShutdownFlag(false)
             {
                 if(concurrentExecutionCount < 1)
                 {
@@ -479,9 +495,9 @@ namespace alpaka::core
                 {
                     std::lock_guard<TMutex> lock(m_mtxWakeup);
                     m_qTasks.push(std::move(upTaskPackage));
-
-                    m_cvWakeup.notify_one();
                 }
+
+                m_cvWakeup.notify_one();
 
                 return future;
             }
@@ -494,6 +510,17 @@ namespace alpaka::core
             [[nodiscard]] auto isIdle() const -> bool
             {
                 return m_numActiveTasks == 0u;
+            }
+
+            void detach(std::unique_ptr<Type>&& self)
+            {
+                m_self = std::move(self);
+
+                {
+                    std::lock_guard<TMutex> lock(m_mtxWakeup);
+                    m_bDetachedFlag = true;
+                }
+                m_cvWakeup.notify_one();
             }
 
         private:
@@ -517,13 +544,23 @@ namespace alpaka::core
                         std::unique_lock<TMutex> lock(m_mtxWakeup);
                         if(m_qTasks.empty())
                         {
+                            if(m_bDetachedFlag)
+                            {
+                                // Pool was detached and is idle, delete
+                                auto self = std::move(m_self);
+                                lock.unlock();
+                                return;
+                            }
+
                             // If the shutdown flag has been set since the last check, return now.
                             if(m_bShutdownFlag)
                             {
                                 return;
                             }
 
-                            m_cvWakeup.wait(lock, [this]() { return ((!m_qTasks.empty()) || m_bShutdownFlag); });
+                            m_cvWakeup.wait(
+                                lock,
+                                [this]() { return ((!m_qTasks.empty()) || m_bShutdownFlag || m_bDetachedFlag); });
                         }
                     }
                 }
@@ -534,7 +571,10 @@ namespace alpaka::core
             {
                 for(auto&& concurrentExec : m_vConcurrentExecs)
                 {
-                    concurrentExec.join();
+                    if(!isThisThread(concurrentExec))
+                        concurrentExec.join();
+                    else
+                        concurrentExec.detach();
                 }
             }
             //! Pops a task from the queue.
@@ -550,11 +590,14 @@ namespace alpaka::core
         private:
             std::vector<TConcurrentExec> m_vConcurrentExecs;
             ThreadSafeQueue<std::shared_ptr<ITaskPkg>> m_qTasks;
-            std::atomic<std::uint32_t> m_numActiveTasks;
+            std::atomic<std::uint32_t> m_numActiveTasks = 0u;
 
             TMutex m_mtxWakeup;
             TCondVar m_cvWakeup;
-            std::atomic<bool> m_bShutdownFlag;
+            std::atomic<bool> m_bShutdownFlag = false;
+
+            std::atomic<bool> m_bDetachedFlag = false;
+            std::unique_ptr<Type> m_self = nullptr;
         };
     } // namespace detail
 } // namespace alpaka::core
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -59,7 +59,7 @@ namespace alpaka::core
                 }
                 else
                 {
-                    t = std::queue<T>::front();
+                    t = std::move(std::queue<T>::front());
                     std::queue<T>::pop();
                     return true;
                 }
@@ -200,7 +200,7 @@ namespace alpaka::core
             {
                 auto ret = fn0();
                 fn1();
-                return std::move(ret);
+                return ret;
             }
             else
             {
--- a/include/alpaka/dev/common/QueueRegistry.hpp
+++ b/include/alpaka/dev/common/QueueRegistry.hpp
@@ -1,0 +1,65 @@
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Jeffrey Kelling
+ *
+ * This file is part of Alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/Common.hpp>
+
+#include <memory>
+#include <mutex>
+#include <vector>
+
+namespace alpaka
+{
+    namespace detail
+    {
+        //! The CPU device implementation.
+        template<typename TQueue>
+        class QueueRegistry
+        {
+        public:
+            ALPAKA_FN_HOST auto getAllExistingQueues() const -> std::vector<std::shared_ptr<TQueue>>
+            {
+                std::vector<std::shared_ptr<TQueue>> vspQueues;
+
+                std::lock_guard<std::mutex> lk(m_Mutex);
+                vspQueues.reserve(std::size(m_queues));
+
+                for(auto it = std::begin(m_queues); it != std::end(m_queues);)
+                {
+                    auto spQueue(it->lock());
+                    if(spQueue)
+                    {
+                        vspQueues.emplace_back(std::move(spQueue));
+                        ++it;
+                    }
+                    else
+                    {
+                        it = m_queues.erase(it);
+                    }
+                }
+                return vspQueues;
+            }
+
+            //! Registers the given queue on this device.
+            //! NOTE: Every queue has to be registered for correct functionality of device wait operations!
+            ALPAKA_FN_HOST auto registerQueue(std::shared_ptr<TQueue> spQueue) const -> void
+            {
+                std::lock_guard<std::mutex> lk(m_Mutex);
+
+                // Register this queue on the device.
+                m_queues.push_back(spQueue);
+            }
+
+        private:
+            std::mutex mutable m_Mutex;
+            std::vector<std::weak_ptr<TQueue>> mutable m_queues;
+        };
+    } // namespace detail
+} // namespace alpaka
--- a/include/alpaka/dev/common/QueueRegistry.hpp
+++ b/include/alpaka/dev/common/QueueRegistry.hpp
@@ -11,9 +11,10 @@
 
 #include <alpaka/core/Common.hpp>
 
+#include <deque>
+#include <functional>
 #include <memory>
 #include <mutex>
-#include <vector>
 
 namespace alpaka
 {
@@ -33,7 +34,7 @@ namespace alpaka
 
                 for(auto it = std::begin(m_queues); it != std::end(m_queues);)
                 {
-                    auto spQueue(it->lock());
+                    auto spQueue = it->lock();
                     if(spQueue)
                     {
                         vspQueues.emplace_back(std::move(spQueue));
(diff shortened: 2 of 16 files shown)


jkelling:
@alpaka-group/alpaka-maintainers @fwyzard  It turns out, that an `alpaka::QueueGenericThreadsNonBlocking`, through `ConcurrentExecPool`, cancels  not yet running tasks in its destructor instead of synchronizing. I always thought, alpaka is modelled rather closely after CUDA, where a stream after destruction completes all tasks. This behavior is also inconsistent with that of `QueueUniformCudaHipRtNonBlocking`, which naturally inherits the behavior of CUDA streams.

I think queues should not behave differently based on the interface. user code must not be expected to decide how to work with a queue based on which backend is being used. Is this inconsistency (undocumented) policy or a bug?

jkelling:
https://github.com/alpaka-group/alpaka/pull/1728/commits/99198fae9174fbac884b78337105a83c412244dc changes the behavior to be consistent with `QueueUniformCudaHipRtNonBlocking` and makes this PR work as I would expect.

@alpaka-group/alpaka-maintainers  please review and discuss.

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
```diff
@@ -106,6 +106,15 @@ namespace alpaka
             return !((*this) == rhs);
         }
 
+        ~QueueGenericThreadsNonBlocking()
+        {
+            if(m_spQueueImpl.unique())
```
fwyzard: This approach doesn't seem thread-safe, and in fact `shared_ptr::unique()` has been deprecated in c++17 and removed in c++20.

jkelling: Yes.. In misread this at first as, `use_count()` being deprecated.

And yes, it is indeed not save because the devices hold `weak_ptr` to the `Impl`, which I also missed at first (otherwise it would effectively be safe because, when `use_count() == 1` in this d'tor no-one could make another copy, unless someone is holding a reference to `this` while it is being destructed ...).

This is more of a quick mockup to show how it should behave. A working way would be the canary between handle and impl mentioned [here](https://github.com/alpaka-group/alpaka/issues/1726#issuecomment-1133582736), but this requires more change, so I would like discuss what the preferred behavior is first.

One could also just has `~ConcurrentExecPool` wait for all tasks to complete instead of cancelling them, but then the non-blocking-queue's d'tor would block when the ref_count reaches zero, which is not desirable. Although, currently it would also block until the currently executing task is done....

fwyzard: Naively I would expect that tasks enqueued in a non-blocking queue will eventually run, even if the queue object is destroyed (_e.g._ because it goes out of scope).
As you point out this is the behaviour we get from CUDA stream: calling `cudaStreamDestroy()` will not destroy the stream until all queued up work has completed:
> In case the device is still doing work in the stream `stream` when [cudaStreamDestroy()](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1gfda584f1788ca983cb21c5f4d2033a62) is called, the function will return immediately and the resources associated with `stream` will be released automatically once the device has completed all work in `stream`.

Would it work for the destructor of `ConcurrentExecPool()` to queue one last task that will wait for all work to complete and only _then_ completes the destruction of the threads and other resources associated to the pool ?

jkelling: That is the basic idea, it is not so simple, though:
1. It the d'tor (be it that of `ConcurrentExecPool` or the queue) waits for the tasks to complete destruction would block, which would be like `cudaStreamDestroy()` implicitly calling `cudaStreamSynchronize()` instead of returning immediately. Thus, destruction of the queue impl (mainly `ConcurrentExecPool`) must be avoided and moved into another thread.
2. There is another problem, though: if we put a task that holds the last ref to the queue into the queue and it destructs, we get the following call stack:
   1. `~QueueImpl`
   2. `~ConcurrentExecPool`
   3. `thread::join()`
 
   where that last call is the thread joining itself, which should result in:
   ```
   terminate called after throwing an instance of 'std::system_error'
      what():  Resource deadlock avoided
   ``` 
   At the moment, I do not understand why this is not happening in the tests with the current develop version. I had a close at `memBufTest` and found, that with the queue being captured, the ref-count are actually always one higher, than I expect, so the deleter never destructs the queue, though if queue is not captured, the deleter is sometimes not called, meaning the queue's refcount went to zero before the deleter's turn in the queue.

jkelling: I created a simplified version of the queue, which crashes when joining itself:
https://gist.github.com/jkelling/792de4fc15e5b44dff4cd96c21ece45a

Is there a for why this does happen in alpaka? Beside the refcount accidentally being such, that the deleter task never destructs the queue?

I also do not see a way to solve this self-join problem with a generic task. The stopper task must be something the `ConcurrentExecPool` is aware of, so it can both destruct safely (avoid self-join) and the terminate the thread without accessing `this` again.

fwyzard: > 2\. where that last call is the thread joining itself, which should result in:

You can avoid this specific problem checking that `thread.get_id() != std::this_thread::get_id()` before calling `join()`.

jkelling: Yes, but then how do you tell the thread's main loop to terminate? It must not check the stop flag, because that was a member of the now-destructed `ConcurrentExecPool`. It *would* check it though, if do not tell it to just return. In the current implementation, only adding this check would make things more dangerous, because you replace an exception by undefined behavior (and not in the standard's sense where mostly nothing bad happens, but a read-after-free).

jkelling: Ok, I understand why this is not crashing:
https://github.com/alpaka-group/alpaka/blob/22374dbfab3596c1bdda38fe90b4b40b9f6f43a1/include/alpaka/core/ConcurrentExecPool.hpp#L504-L524
The task is held in the main loop in that shared pointer. If the queue is empty after the deleter task the executor keeps the pointer and goes into the wait, leaking the queue. In most cases in the tests. There is an "asan" test failing, when the capture is removed without any other change, because sometimes the delete does not get executed and the buffer is leaked. The same address sanitizer does not detect the leak of the queue, maybe because there is cycle there.

If I insert a `currentTaskPackage = nullptr;` after line 512 and force the deleter task to be the last thing holding the queue by adding tome sleeps in test, then I see the expected exception.

jkelling: In the old PR https://github.com/alpaka-group/alpaka/pull/1727 which would only fix the warning, I modified the test and `ConcurrentExecPool` to catch this problem, so the can call that version "red".

jkelling:
It appears we need to introduce a trait for `std::this_thread::get_id()` in the check `std::this_thread::get_id() != concurrentExec.get_id()`, because at the moment test are failing because this code break boost::fiber. On the other hand there is https://github.com/alpaka-group/alpaka/pull/1718 .

@j-stephan Should other thread type remain supported, or will this be merged soon?

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
jkelling: Apart from support for boost::fiber, this PR is ready.

@alpaka-group/alpaka-maintainers  please review.

jkelling:
I added a new trait `IsThisThread`, having this is more clean even with boost::fiber gone and the boost::fiber implementation of removes the noise of failing boost::fiber tests.

There are two failing runs on windows: [windows_cl-2019_release](https://github.com/alpaka-group/alpaka/runs/6591158601?check_suite_focus=true) and [windows_cl-2022_release](https://github.com/alpaka-group/alpaka/runs/6591158782?check_suite_focus=true) in both cases a similar set of test is failing and all failing test print with `***Exception: `, but no further information. The debug variants of these are succeeding. Could there an issue with using `std::this_thread::get_id()` in msvc? @bernhardmgruber do you have an idea?

jkelling:
Deleting a not-joined `std::thread` should throw an exception and there is no difference if a thread destructs it own handle. So, the thread that `isThisThread` and which cannot be joined must be detached instead. Curious, that I no test caught this, except those msvc release builds, at lest they are passing now with this in place.

I also added some code to make `ConcurrentExecPool::detach()` also work when there is more than one worker thread.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -367,7 +385,8 @@ namespace alpaka::core
             std::vector<TConcurrentExec> m_vConcurrentExecs;
             ThreadSafeQueue<std::shared_ptr<ITaskPkg>> m_qTasks;
             std::atomic<std::uint32_t> m_numActiveTasks;
-            std::atomic<bool> m_bShutdownFlag;
+            std::atomic<bool> m_bShutdownFlag, m_bDetachedFlag = false;
```
BenjaminW3: If NSDMI works the same way variable initialization works in other places, then this does not do what you think.
Only `m_bDetachedFlag` is set to `false` and `m_bShutdownFlag` is default initialized.
Please split this initialization in multiple lines.

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -551,7 +594,9 @@ namespace alpaka::core
 
             TMutex m_mtxWakeup;
             TCondVar m_cvWakeup;
-            std::atomic<bool> m_bShutdownFlag;
+            std::atomic<bool> m_bShutdownFlag, m_bDetachedFlag = false;
```
BenjaminW3: same here

BenjaminW3 (dismissed):

jkelling:
It does work the same way and intended to get this behaviour. So, the two-line version would be:
```c++
std::atomic<bool> m_bShutdownFlag;
std::atomic<bool> m_bDetachedFlag = false;
```
`m_bShutdownFlag` was there before and is initialized in the c'tor. I did not want to touch code I do not need to change but also did not want to use old-style initialization for the new member.

I can see how this inconsistency obfuscates the intent, so I guess it is best to update the trivial initializations of the pre-existing members.

On 26 May 2022 08:26:13 CEST, Benjamin Worpitz ***@***.***> wrote:
***@***.*** commented on this pull request.

include/alpaka/core/ConcurrentExecPool.hpp
jkelling: It does work the same way and intended to get this behaviour. So, the two-line version would be:
```c++
std::atomic<bool> m_bShutdownFlag;
std::atomic<bool> m_bDetachedFlag = false;
```
`m_bShutdownFlag` was there before and is initialized in the c'tor. I did not want to touch code I do not need to change but also did not want to use old-style initialization for the new member.

I can see how this inconsistency obfuscates the intent, so I guess it is best to update the trivial initializations of the pre-existing members.

jkelling:
Three CI runs for the new version containing only some style changes (https://github.com/alpaka-group/alpaka/pull/1728/commits/5f4dd68de50a060ef063be95ce7bcfee3dcf7ecc). 

In two cases gcc got killed:
*  [linux_gcc-8_debug ](https://github.com/alpaka-group/alpaka/runs/6607231273?check_suite_focus=true) (In this case the GUI view does not load the output, but the raw logs can be downloaded.)
   ```
   2022-05-26T10:15:38.9305827Z g++: fatal error: Killed signal terminated program cc1plus
   2022-05-26T10:15:38.9573880Z compilation terminated.
   2022-05-26T10:15:38.9596836Z make[2]: *** [test/unit/kernel/CMakeFiles/kernelTest.dir/build.make:147: test/unit/kernel/CMakeFiles/kernelTest.dir/src/KernelWithOmpSchedule.cpp.o] Error 1
   ```
* [linux_gcc-9_debug_c++17](https://github.com/alpaka-group/alpaka/runs/6615775997?check_suite_focus=true)
   ```
   g++: fatal error: Killed signal terminated program cc1plus
   compilation terminated.
   make[2]: *** [test/unit/kernel/CMakeFiles/kernelTest.dir/build.make:146: test/unit/kernel/CMakeFiles/kernelTest.dir   /src/KernelWithOmpSchedule.cpp.o] Error 1
   ```

Is it possible,that here the container went OoM?

One case is a bit concerning, because it is a SIGSEGV in `memViewTest` (at the end of the program, after all checks succeeded), but [linux_clang-5_debug_c++17](https://github.com/alpaka-group/alpaka/runs/6608821015?check_suite_focus=true#step:5:7584). I cannot reproduce this locally, nor in the CI it seems. This may be related to the changes in this PR, or may be something else... . Any suggestions?

jkelling:
> Three CI runs for the new version containing only some style changes ([5f4dd68](https://github.com/alpaka-group/alpaka/commit/5f4dd68de50a060ef063be95ce7bcfee3dcf7ecc)).

Btw.: by "three CI runs" I mean three different triggered actions, each of those had only one check fail, i.e. each failed check succeeded in the other two.

jkelling:
With the queues getting detached, there is a chance, that the worker thread of a queue runs past the end of `main()`, invoking UB. Such a thing may have caused the SIGSEGV in that clang check. (The current state of develop has the same problem, when a queue's worker thread gets leaked by the `BufCpu::deleter`)

If we want the non-blocking queue not to block in its c'tor, we need some other place to make sure the detached queue can be caught before program termination. The device may be a place to do it. The device's d'tor waiting for all task on the device to finish seems reasonable to me. The device is already connected to its queues, though the current list of `weak_ptr` would not do. A destructed queue would need to register a "waiter" that makes the d'tor wait for the queue (`ConcurrentlExecPool`) to actually destruct. This does add some complexity...

jkelling:
Two groups checks are failing now, but consistently:
1. clang <=8 with libc++: In some tests an exception is thrown that indicates, that an attempt made to lock a mutex which was already destructed, or not allocated. I wan reproduce this locally using clang 8 with libc++ and having an OMP2 backend enabled. The error requires these three things in combination to occur. Considering, that this requires older version of clang(libc++ (<= 8) and it does not occur with newer versions, neither with libstdc++, this is probably an old bug in libc++ which got fixed by now. As this affects an older compiler version, I prefer to not investigate further and either introduce a work around that make the queue wait in its c'tor or deprecate libc++ <= 8 . @alpaka-group/alpaka-maintainers do we have a defined support level for libc++?
2. With the non-cuda msvc debug builds, ~~the `fenceTest` is~~ some tests (not sure where I saw only the one) failing (says "segfault"), while tests pass in the release build, at least for the latest run, this may not be consistent. Does anyone with access to msvc have any idea what is going on there? This does look a bit like like the clang+libc++<=8-issue, only with a less specific exception.

test/unit/queue/src/QueueTest.cpp
```diff
@@ -171,3 +172,34 @@ TEMPLATE_LIST_TEST_CASE("queueShouldNotExecuteTasksInParallel", "[queue]", TestQ
...
     secondTaskFinishedFuture.get();
 }
+
+//! This task launches a long task in a non-blocking queue and destroys the
+//! queue before the task is finished. The run time of the task is a bit longer
+//! than all other tests here, so the task would run past program termination if
+//! not synchronized somewhere. This test has no assertion, because the tested
+//! behviour is outside of any function scope.
+//!
+//! A thread running past program termination invoked undefined behavior, which
```
fwyzard:
```suggestion
//! A thread running past program termination invokes undefined behavior, which
```

test/unit/queue/src/QueueTest.cpp
```diff
@@ -171,3 +172,34 @@ TEMPLATE_LIST_TEST_CASE("queueShouldNotExecuteTasksInParallel", "[queue]", TestQ
...
+//! occasionally leads to crashes after termination. I.e. this test may crash
+//! after catch2 reported success.
```
fwyzard: Would a catch2 check at the end of the long task be useful in marking the "undefined behaviour" as an explicit failure ?

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -162,27 +162,35 @@ TEMPLATE_LIST_TEST_CASE("memBufConstTest", "[memBuf]", alpaka::test::TestAccs)
...
+        alpaka::wait(queue);
+        auto const offset = alpaka::Vec<Dim, Idx>::zeros();
+        alpaka::test::testViewImmutable<Elem>(buf, dev, extent, offset);
+
+        // the buffer will queue the deallocation of the memory when it goes out of scope,
+        // and extend the lifetime of the queue until all memory operations have completed.
+        // delay the end end of the queue to push the buffer deletetion task
+        // after all local refs to queue have been dropped
+        alpaka::enqueue(queue, []() { std::this_thread::sleep_for(std::chrono::microseconds(1000)); });
+        [](auto) {}(std::move(queue));
```
fwyzard, line 190: What is the usefulness of this lambda ? It is executed immediately.

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -162,27 +162,35 @@ TEMPLATE_LIST_TEST_CASE("memBufConstTest", "[memBuf]", alpaka::test::TestAccs)
...
+    }
 
-    // the buffer will queue the deallocation of the memory when it goes out of scope,
-    // and extend the lifetime of the queue until all memory operations have completed.
+    std::this_thread::sleep_for(std::chrono::microseconds(1200));
```
fwyzard, line 196: Why is this needed ?

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
```diff
@@ -57,12 +57,30 @@ namespace alpaka
...
+                    m_dev.registerCleanup(
+                        [pool = std::weak_ptr<ThreadPool>(m_workerThread)]() noexcept
+                        {
+                            auto s = pool.lock();
+                            if(s)
+                            {
+                                s = s->takeDetachHandle();
+                            }
+                        });
+                    m_workerThread->detach(std::move(m_workerThread));
```
fwyzard, line 82: Is this legal ?
`detach()` is invoked on `m_workerThread` _after_ is has been moved from.

fwyzard (commented):
Could you make two separate PRs, one with only the changes to `include/alpaka/mem/buf/BufCpu.hpp`, and one with the semantic changes to `QueueGenericThreadsNonBlocking` ?

test/unit/queue/src/QueueTest.cpp
jkelling: No, because that check would just never happen in the failure scenario: The thread runs and either causes some trouble, because it touches memory or calls a function related to something already cleaned-up after `main()` or has no effect and at some point just gets killed by the OS when parent process terminates. So, the thread would not get to that check. If we were to wait for the check, then we eliminate the failure scenario, by effectively joining that thread.

It would be better if we could check for some defined behavior instead of checking for nothing undefined happening. For example, if in alpaka all queues devices were required to synchronize with and destroy their attached queues in their destructor, we could check that. But this is not a thing that alpaka devices do.

test/unit/mem/buf/src/BufTest.cpp
jkelling: The lambda destroys the `queue` handle in the local scope, by moving it into another scope, where it immediately goes out of scope. This there to remove the local `queue` handle before the buffer is destructed (opposite to the normal destruction order), to leave the last handle inside the buffer or its deleter.

test/unit/mem/buf/src/BufTest.cpp
jkelling: This makes sure, the test case does not end before the buffer's deleter ran, so any exceptions thrown during it will show up there. At this point the queue is detached (both on `develop` and in this PR), so it does not synchronize with the end of this scope.

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
jkelling: Yes.

All arguments for the function are obtained first (i.e. `this` for `detach()` and the rvalue smart pointer) in no specific order, but before the function `detach` is invoked. `std::move` only casts the reference to the object into an rvalue reference (basically it only affects overload resolution), it does not modify the object. The actual change to `m_workerThread`, which puts it into a moved-from state, happens inside `detach`, in for of the move-assignment.

jkelling:
> Could you make two separate PRs, one with only the changes to `include/alpaka/mem/buf/BufCpu.hpp`, and one with the semantic changes to `QueueGenericThreadsNonBlocking` ?

No, because the change to `BufCpu` only makes it broken in a different way: Currently the worker thread of the queue is leaked,. The queue itself remains referenced by the thread, so is technically not leaked, which is why the address sanitizer check passes. The change makes the buffer's memory occasionally leak, making that test fail.

jkelling:
Regarding these failing checks with clang + libc++ <=8 (and maybe likewise the msvc checks):

The exception from the mutex lock traces back into the worker thread of the queue. I.e. it looks like the mutex was destructed while the worker thread is still running. This should not be possible because the mutex is a member of `ConcurrentExecPool` and is just only destructed at the end of `~ConcurrentlExecPool`, which joins all worker threads, so does not reach the end while threads are running. The only exception to that is, when a worker thread destruct the `ConcurrentExecPool` in which case the thread is not joined but detached, so keeps running. However: the worker thread destructs the pool only in a [return statement](https://github.com/alpaka-group/alpaka/pull/1728/files#diff-1e7ed7f50e4605ed7f5e84260029cb16b4a8caf21e02e3831ebc116eef9e8448R558-R563), so it cannot be running afterwards. Ultimately, this could only be explained by a memory corruption in the code, if there is not weird compiler bug.

A memory corruption, and the fact, that an OpenMP backend must be active for this to happen, causes the bad feeling, that this might be related to https://github.com/alpaka-group/alpaka/issues/1676 : This PR adds some interesting ways of handling `std::threads` spawned inside an OpenMP environment, which is not required to be thread-safe and is allowed to assume, that all threads were spawned by the OpenMP runtime...

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
jkelling: Also note, that `detach()` is not invoked on`m_workerThread`, which is a `shared_ptr<>` (in the latest version), but on `*m_workerThread`, which is a `ConcurrentExecPool`.

jkelling: > in no specific order,

This is actually not correct: the `operator->` constitutes a sequence point, so `this` is obtained first, then the other function arguments are evaluated and assigned (those in no specific order, but there is only the one here). This does not really matter as `detach()` takes an rvalue reference and thus the rest of my post applies. If it would take the argument by value (which be wrong from a API-semantic point of vierw), then a move assignment would happen during function argument assignment, but then the sequence point would still make it behave in the same way.

test/unit/mem/buf/src/BufTest.cpp
fwyzard: I see.

But why do you need a lambda ?
Wouldn't moving from the queue in a nested scope be sufficient ?
```suggestion
        {
          // ...
          auto unused = std::move(queue);
        }
```

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
fwyzard: Thanks for the explanation, I tend to agree that the actual "moving from" happens inside the function body, after the function call.

Just on one point:

> the `operator->` constitutes a sequence point,

I cannot find any reference to this being the case, could you point to it ?

jkelling:
So, the rabbit hole is even deeper:

With the clang8 + libc++ + OMP2 setup, I found, that the invalid mutex lock occurred in a worker thread, that actually had destructed the Concurrent exec pool itself. A backtrace from that point of destruction lead through a `TaskPkg` holding an `EventThreadGeneric` which in turn owns a `Dev`. That `Dev` was apparently the last handle in this situation, so the destruction of the task package, which also destructed the event (which also must have been the last handle, so no one outside seemed to care about the event anymore), destructed the `ConcurentExecPool` (i.e. the involved queue must have been gone before that).

When I added the `m_cleanupList` in the device to catch detached worker `ConcurrentExecPool`s, I assumed, that devices would be sure to outlive their queues and would not get destructed inside queues ... alpaka Event seems to have a different opinion. I pushed a commit that makes the `cleanupList` a static variable inside the `Dev::registerCleanup()` function, so  it is no longer attached to a device instance. Most programs probably only hold one `Dev` per physical device anyway, so having a global mutex per type instead of one per instance, should not change anything performance-vise. An issue could be, that the list only grows when queues are destroyed but never shrinks.So if queue are used in a fire-and forget fashion, this can cause memory issues...

jkelling:
@alpaka-group/alpaka-maintainers Is there a reason for there being no hierarchy between devices an queues? Each queue seems to be attached to exactly one device over its life time and the `shared_ptr` connection, ensures, that the device's lifetime cover that of the queue. Why does the device not just own the queue? We could just have an `std::unique_ptr<IQueueType>` in the device and a plain pointer back the queue... .

test/unit/mem/buf/src/BufTest.cpp
jkelling: Might do (apart from the `}` there, but would this not raise an "unused" warning?

fwyzard: I would argue that it is equally unused in both cases... 
But if it causes a warning, I guess
```suggestion
        {
          auto unused [[maybe_unused]] = std::move(queue);
        }
```
should do :-/

jkelling: But in case of the lambda the missing parameter name tells the compiler, that it is expected to be unused.

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
jkelling: Hmm, maybe there is actually no sequence defined.

[Here](https://en.cppreference.com/w/cpp/language/eval_order) under "Rules", rules 2 or 14 could apply somehow, but 2 sequences only the value computations, so the side-effect of a move-assignment in the function-call expression could still go anywhere. I do not know, if, in 14, the "expression that names the function" includes the object and member access, or only anything that is where the function name goes, like some expression evaluation to a function pointer.

jkelling:
The last CI run went through without failures. I rebased and cleaned up the branch.

include/alpaka/dev/common/QueueRegistry.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <deque>
+#include <functional>
+#include <memory>
+#include <mutex>
+
+namespace alpaka
+{
+    namespace detail
+    {
+        //! The CPU device implementation.
```
fwyzard, line 23: is the comment still correct ?

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
fwyzard: **disclaimer**: *this is a contrived example and has nothing to do with the code in the PR*.

But I was curious, so I've tried to come up with a simpler example where the order of evaluation would actually have observable consequences: https://godbolt.org/z/j6c1GK8Tj .

I _think_ the order of evaluation in `p->print(move_from_and_return(p))` is the same as in `m_workerThread->detach(std::move(m_workerThread))`.

In the contrived example, the argument to the function _does_ "move from" the `shared_ptr`  _before_ the method is called, resulting in the call appling to a `nullptr`, and leading to a crash.

fwyzard:
@tonydp03 could you run the CPU-only benchmark of [pixeltrack-standalone](https://github.com/cms-patatrack/pixeltrack-standalone/pulls/fwyzard) with the `HEAD` of the alpaka `devel` branch, with and without this PR ?

include/alpaka/dev/common/QueueRegistry.hpp
jkelling: It is indeed not correct anymore.

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
jkelling: I also played with some example, but only with different ways of assigning the arguments (e.g. by-value instead of by-rvalue-ref), not with calling a function that has side-effects (`std::move` is only a cast). Intuitively, I think that the evaluation of the call to `move_from_and_return()` might even be guaranteed to happen before the outer parts of the expression, not just be called first depending on the compiler's whim.

I think we clarified, that `->` does not imply any sequence of evaluation here. I.e., if `detach` would take a value, than the compiler may do the move-assignment of the parameter or the dereference at `->` first as it chooses ... Or it maybe, since C++17, it is compelled to evaluate the part identifying the function first before assigning its argument, not sure.

jkelling: > Intuitively, I think that the evaluation of the call to `move_from_and_return()` might even be guaranteed to happen before the outer parts of the expression

Although, I do not see any of the rules implying this, so I guess the order in your example is probably undefined.

fwyzard:
> @tonydp03 could you run the CPU-only benchmark of [pixeltrack-standalone](https://github.com/cms-patatrack/pixeltrack-standalone/pulls/fwyzard) with the `HEAD` of the alpaka `devel` branch, with and without this PR ?

the results from @tonydp03 do not show any impact on the performance:

> image

psychocoderHPC:
I will not be able to check this PR before mid of the next week. sry

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -503,12 +503,14 @@ namespace alpaka::core
                 // Checks whether pool is being destroyed, if so, stop running (lazy check without mutex).
                 while(!m_bShutdownFlag)
                 {
-                    auto currentTaskPackage = std::shared_ptr<ITaskPkg>{nullptr};
-
-                    // Use popTask so we only ever have one reference to the ITaskPkg
-                    if(popTask(currentTaskPackage))
                     {
```
psychocoderHPC, line 550: please add a comment on why this code is in its own code block. I assume it has something to do with guaranteeing when ` currentTaskPackage` is destroyed. The comment should avoid that someone is removing the paranteeses.

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -162,27 +162,38 @@ TEMPLATE_LIST_TEST_CASE("memBufConstTest", "[memBuf]", alpaka::test::TestAccs)
...
+        auto const buf = alpaka::allocAsyncBuf<Elem, Idx>(queue, extent);
+
+        // synchronous operations must wait for the memory to be available
+        alpaka::wait(queue);
+        auto const offset = alpaka::Vec<Dim, Idx>::zeros();
+        alpaka::test::testViewImmutable<Elem>(buf, dev, extent, offset);
+
+        // the buffer will queue the deallocation of the memory when it goes out of scope,
+        // and extend the lifetime of the queue until all memory operations have completed.
+        // delay the end end of the queue to push the buffer deletetion task
```
psychocoderHPC: `end end`?

psychocoderHPC (changes requested):

jkelling:
Force-pushed nothing to restart pipeline `linux_gcc-9_debug_c++17 ` failed because `g++: fatal error: Killed signal terminated program cc1plus`.

jkelling:
The CI is showing an occasional timeout in `queueTest`. It seems to always happen in `QueueUniformCudaHipRt` (with either value for `TBlocking`). I am investigating.

jkelling:
This PR is ready now. The `queueTest` timeout is a pre-existing bug, reported in #1741 .

@psychocoderHPC please merge.

psychocoderHPC (approved):

Issue #1731, opened 2022-06-03
Missing pow(real, complex) and pow(complex, real) on GPUs


sbastrakov:
These two cases are not covered by existing unit tests (since their infrastructure is for same-type arguments of binary functions) and so it was missed that our CUDA/HIP math only implements `pow(real, real)` and `pow(complex, complex)`. For CPU implementations it is not a problem as we rely on `std::` implementations that have these versions.

I will provide a PR with tests to cover it and fix for CUDA/HIP shortly.

Issue #1732, opened 2022-06-03
Pow with mixed precision does not properly promote types


sbastrakov:
By not properly I mean not [what `std::pow` does](https://en.cppreference.com/w/cpp/numeric/complex/pow).
This concerns both on CPU and GPU. The issue is somewhat similar, but independent from #1731. Again, we have no existing unit tests for that. I will add a fix and a test

sbastrakov:
Oh a related issue also exists for real `pow` on CUDA/HIP as the newly added tests show - on some compilers there are warnings triggered by implicit casting in this case.

PR #1733, opened 2022-06-07
Fix mixed-type and mixed-precision pow() implementation

sbastrakov:
Fix issues #1731 and #1732, also add tests to cover these cases.

Edit: originally the PR concerned only implementation of `pow()` for `Complex<>`, but as it became clear with tests, there was also an issue for mixed `float`-`double` `pow` on CUDA/HIP. So the PR is now mostly, but not fully, for `Complex`.

### DIFF
--- a/include/alpaka/math/Complex.hpp
+++ b/include/alpaka/math/Complex.hpp
@@ -501,25 +501,28 @@ namespace alpaka
     //! Complex power of a complex number
     ALPAKA_NO_HOST_ACC_WARNING
     template<typename T, typename U>
-    constexpr ALPAKA_FN_HOST_ACC Complex<T> pow(Complex<T> const& x, Complex<U> const& y)
+    constexpr ALPAKA_FN_HOST_ACC auto pow(Complex<T> const& x, Complex<U> const& y)
     {
-        return std::pow(std::complex<T>(x), std::complex<U>(y));
+        // Use same type promotion as std::pow
+        auto const result = std::pow(std::complex<T>(x), std::complex<U>(y));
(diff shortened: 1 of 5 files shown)


bernhardmgruber (dismissed):

psychocoderHPC:
There is an implicit cast in the code:

```
/builds/hzdr/crp/alpaka/test/unit/math/src/powMixedTypes.cpp:47:39: error: implicit conversion increases floating-point precision: 'float' to 'const double' [-Werror,-Wdouble-promotion]
        return static_cast<TExpected>(0.0f);
               ~~~~~~~~~~~            ^~~~
/builds/hzdr/crp/alpaka/test/unit/math/src/powMixedTypes.cpp:29:66: note: in instantiation of function template specialization 'PowMixedTypesTestKernel<alpaka::Complex<double>>::toExpected<alpaka::Complex<float>>' requested here
        auto expected = alpaka::math::pow(acc, toExpected(arg1), toExpected(arg2));
                                                                 ^
/usr/lib/gcc/x86_64-linux-gnu/9/../../../../include/c++/9/bits/invoke.h:60:14: note: in instantiation of function template specialization 'PowMixedTypesTestKernel<alpaka::Complex<double>>::operator()<alpaka::AccCpuSerial<std::integral_constant<unsigned long, 1>, unsigned long>, double, alpaka::Complex<float>>' requested here
    { return std::forward<_Fn>(__f)(std::forward<_Args>(__args)...); }
```

psychocoderHPC (approved):

bernhardmgruber (approved):

PR #1734, opened 2022-06-07
Replace ConcurrentExecPool by CallbackThread in QueueGenericThreadsNonBlocking

bernhardmgruber:


bernhardmgruber:
As always in alpaka, if I think that something can be quickly changed in a few minutes, I end up in hours discovering and hunting edge cases.

This time I discovered that for the future returned from `CallbackThread::submit` it is not clear how long the task passed to it is going to live. The `std::packaged_task` used inside may choose to put the task into the shared state with the returned `std::future` or not. So when the task and any state it carries is destroyed is unspecified. This has implications for e.g. `EventGenericThreads`, which, when enqueued, copies itself into the task. If the task's lifetime is then bound to the future, which is stored into the event object, there is a cyclic shared pointer reference from event -> future -> task -> event and we leak memory. If the task is immediately destroyed upon completion and not bound to the future, this memory leak cannot happen.

However, if tasks are immediately destroyed when completed, we can run into the situation where the last queue handle may be passed into the task which is enqueued. Upon task destruction, the queue is destroyed, which destroyes and joins the callback thread. Since this destruction chain itself runs on the callback thread, the thread tries to join itself and deadlocks.

I encountered both situations today, depending on which changes I made to `CallbackThread`. For a reason yet obscure to me, `ConcurrentExecPool` does not suffer from both problems. Either it is clever enough to avoid both situations, or there is a bug which just has not occured yet.

jkelling:
> I encountered both situations today, depending on which changes I made to `CallbackThread`. For a reason yet obscure to me, `ConcurrentExecPool` does not suffer from both problems. Either it is clever enough to avoid both situations, or there is a bug which just has not occured yet.

I do not know if that circular dependence you described above can happen in `ConcurrentExecPool`. I do know that the `develop` version actually suffers from the self-join problem, as described in #1728 : It does not show it, because it never destroys the last task if that holds the last reference to the queue and instead leaks the thread and, with it, itself. [This change](https://github.com/alpaka-group/alpaka/pull/1728/commits/c4eae60310e2534954d3a6cc989132d68bc3278b#diff-1e7ed7f50e4605ed7f5e84260029cb16b4a8caf21e02e3831ebc116eef9e8448) turns the leak into a self-join with the appropriate exception being thrown.

bernhardmgruber:
I abandon this PR since there are more things to be solved with `CallbackThread`, including detecting self-joining upon destruction. This issue has been solved for `ConcurrentExecPool` on develop by @jkelling in #1728.

PR #1735, opened 2022-06-08
Diagnose CallbackThread joining itself

bernhardmgruber:
This can happen when the CallbackThread is moved into a task that is enqueued on itself.

### DIFF
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -13,6 +13,7 @@
 #include <condition_variable>
 #include <functional>
 #include <future>
+#include <iostream>
 #include <mutex>
 #include <queue>
 #include <thread>
@@ -29,7 +30,14 @@ namespace alpaka::core
             m_stop = true;
             m_cond.notify_one();
             if(m_thread.joinable())
+            {
+                if(std::this_thread::get_id() == m_thread.get_id())
+                {
+                    std::cerr << "ERROR in ~CallbackThread: thread joins itself" << std::endl;
+                    std::abort();
+                }
                 m_thread.join();
+            }
         }
         auto submit(Task&& newTask) -> std::future<void>
         {


psychocoderHPC:
I thought we have a separate thread for each queue that is handling the callbacks. Do I miss something?

include/alpaka/core/CallbackThread.hpp
```diff
@@ -29,7 +30,14 @@ namespace alpaka::core
             m_stop = true;
             m_cond.notify_one();
             if(m_thread.joinable())
+            {
+                if(std::this_thread::get_id() == m_thread.get_id())
+                {
+                    std::cerr << "ERROR in ~CallbackThread: thread joins itself" << std::endl;
+                    std::abort();
```
jkelling, line 37: Would it not be a bit more idiomatic to catch the `std::system_error` thrown without the check and report on this? ... or just let the exception fly...?

jkelling:
> I thought we have a separate thread for each queue that is handling the callbacks. Do I miss something?

I think there is.

The issue addressed here occurs when a task holds the last handle to the queue and thus destructs the queue inside the callback thread.

include/alpaka/core/CallbackThread.hpp
bernhardmgruber: Well, the `system_error` may occur due to other reasons as well (I don't know of any). So this way, I am sure about what situation caused the error and due to the `cerr` I can actually read it. The other day I ran into this problem during stack unwinding when another exception was already flying, so I only saw that `terminate` was called because a second exception was raised, but did not see the actual error.

jkelling: Ok.

Though I see, that we both are wrong, I think: On some branch, I added a [test before](https://github.com/alpaka-group/alpaka/pull/1740/commits/c379599377ad110ef437861ec6df65fd1ba2eb0e) that should trigger exactly this failure condition, but it did not trigger. So now, I wondered why. According to [cppreference](https://en.cppreference.com/w/cpp/thread/thread/joinable)
> [joinable] returns true if `get_id() != std::thread::id()`

Maybe this means it returns `false` otherwise, I.e. your check will never trigger and the thread just does not get joined both with and withou this PR.

bernhardmgruber: That is interesting! Because I can see the error message on Windows :)

jkelling: Maybe on windows the `std::thread::id()` returns `NAN` (SCNR).

bernhardmgruber: Ok I think we both misunterstood cppreference. `std::thread::id` is a [class](https://en.cppreference.com/w/cpp/thread/thread/id). So `get_id() != std::thread::id()` checks whether the thread id is not a default constructed `std::thread::id`. It is not the same as `get_id() != std::this_thread::get_id()`.

Ergo, my proposed check can trigger and diagnose a problem.

jkelling (approved):

jkelling:
@bernhardmgruber you need someone like @psychocoderHPC to review too, before this can be merged.

j-stephan (approved):

PR #1737, opened 2022-06-09
[Do not merge] Test timeout

jkelling:


bernhardmgruber:
@jkelling could you please retest whether there is still a timeout somewhere in the tests? And if there is not, please close the PR! Thx!

bernhardmgruber:
@jkelling since we think we squashed the queue bug now with #2026, can you please reevaluate whether you still need this PR? And if not, please close it. Thx!

PR #1738, opened 2022-06-09
mirror fix in version number

KseniaBastrakova:
there is a typo in the documentation - the version has not been updated to the latest


sbastrakov:
For the context, it is not just so that the latest alpaka release is used (although that by itself makes sense imo). The install instructions for the current `develop` do not work for the 0.8.0 release as is. So now strictly following documentation doesn't work, and this PR fixes it.

bernhardmgruber (approved):

PR #1739, opened 2022-06-09
Demangle type names

### LINKED ISSUE #1681: Proposal: demangle type names
There are a few places in the alpaka library and the test files where names and log messages use mangled type names:
https://github.com/alpaka-group/alpaka/blob/dacd3a2d955943b485b7fcc8eb3db7fd38a73df7/include/alpaka/acc/AccCpuSerial.hpp#L151

https://github.com/alpaka-group/alpaka/blob/dacd3a2d955943b485b7fcc8eb3db7fd38a73df7/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp#L293-L294

https://github.com/alpaka-group/alpaka/blob/dacd3a2d955943b485b7fcc8eb3db7fd38a73df7/test/unit/math/src/TestTemplate.hpp#L64-L66

_etc._

I think it would be a bit user-friendlier to demangle the type names.

It would be easy to do using Boost, just a matter of replacing `typeid(T).name()` with `boost::core::demangle(typeid(T).name())`.

We are using a similar approach in our code, but wrap it in a template constant to avoid repeated calls:
```c++
template <typename T>
inline const std::string demangle = boost::core::demangle(typeid(T).name());
```
and then use it simply as `demangle<T>`:
```c++
std::cout << demangle<T> << std::endl;
```

If the preference is to avoid the additional dependency on Boost by default, it could also be protected with an `#ifdef`:
```c++
#ifdef ALPAKA_DEMANGLE_TYPE_NAMES
template <typename T>
inline const std::string readableName = boost::core::demangle(typeid(T).name());
#else
template <typename T>
inline const std::string readableName = std::string(typeid(T).name());
#endif

...

std::cout << readableName<T> << std::endl;
```

tonydp03:
Resolves #1681.

### DIFF
--- a/include/alpaka/acc/AccCpuOmp2Blocks.hpp
+++ b/include/alpaka/acc/AccCpuOmp2Blocks.hpp
@@ -23,6 +23,7 @@
 #    include <alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp>
 #    include <alpaka/block/shared/st/BlockSharedMemStMember.hpp>
 #    include <alpaka/block/sync/BlockSyncNoOp.hpp>
+#    include <alpaka/core/DemangleTypeNames.hpp>
 #    include <alpaka/idx/bt/IdxBtZero.hpp>
 #    include <alpaka/idx/gb/IdxGbRef.hpp>
 #    include <alpaka/intrinsic/IntrinsicCpu.hpp>
@@ -154,7 +155,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getAccName() -> std::string
             {
-                return "AccCpuOmp2Blocks<" + std::to_string(TDim::value) + "," + typeid(TIdx).name() + ">";
+                return "AccCpuOmp2Blocks<" + std::to_string(TDim::value) + "," + core::demangled<TIdx> + ">";
             }
         };
 
--- a/include/alpaka/acc/AccCpuSerial.hpp
+++ b/include/alpaka/acc/AccCpuSerial.hpp
@@ -18,6 +18,7 @@
 #    include <alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp>
 #    include <alpaka/block/shared/st/BlockSharedMemStMember.hpp>
 #    include <alpaka/block/sync/BlockSyncNoOp.hpp>
+#    include <alpaka/core/DemangleTypeNames.hpp>
 #    include <alpaka/idx/bt/IdxBtZero.hpp>
 #    include <alpaka/idx/gb/IdxGbRef.hpp>
 #    include <alpaka/intrinsic/IntrinsicCpu.hpp>
@@ -148,7 +149,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getAccName() -> std::string
             {
-                return "AccCpuSerial<" + std::to_string(TDim::value) + "," + typeid(TIdx).name() + ">";
+                return "AccCpuSerial<" + std::to_string(TDim::value) + "," + core::demangled<TIdx> + ">";
             }
         };
 
--- a/include/alpaka/core/DemangleTypeNames.hpp
+++ b/include/alpaka/core/DemangleTypeNames.hpp
@@ -1,0 +1,28 @@
+/* Copyright 2022 Andrea Bocci, Antonio Di Pilato
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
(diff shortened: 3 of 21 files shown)


include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -154,7 +155,7 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getAccName() -> std::string
             {
-                return "AccCpuOmp2Blocks<" + std::to_string(TDim::value) + "," + typeid(TIdx).name() + ">";
+                return "AccCpuOmp2Blocks<" + std::to_string(TDim::value) + "," + core::demangle<TIdx> + ">";
```
fwyzard: would
```suggestion
                return core::demangle<AccCpuOmp2Blocks<TDim, TIdx>>;
```
work ?

include/alpaka/core/DemangleTypeNames.hpp
```diff
@@ -0,0 +1,18 @@
...
+ */
+
+#pragma once
+
+#include <boost/core/demangle.hpp>
+
+namespace alpaka::core
+{
+    template<typename T>
+    inline const std::string demangle = boost::core::demangle(typeid(T).name());
```
bernhardmgruber:
```suggestion
    template<typename T>
    inline const std::string demangled = boost::core::demangle(typeid(T).name());
```
I would call it `demangled`, because it is not a function so a verb seems inappropriate.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
bernhardmgruber: The question is how much control we want to have of the returned string. MSVC for example prepends `struct`/`class` to each class name. See e.g.: https://godbolt.org/z/6P3dfd1Mq.

But since `getAccName()` relied on `typeid` so far, we were already not giving stable names. For me, your proposed change is fine.

tonydp03: It works but the prints are, for example:
_before_: `AccCpuSerial<3,unsigned long>`
_after_: `alpaka::AccCpuSerial<std::integral_constant<unsigned long, 3ul>, unsigned long>`

fwyzard: Ah, I see, while Tony's approach would be more consistent across different architectures, right ?

bernhardmgruber: That is indeed a bit long. Maybe keep the shorter names then!

include/alpaka/core/DemangleTypeNames.hpp
bernhardmgruber: However, we could argue that `f<x>` is just another spelling for a function `f(x)` taking a compile time input (a meta function). So my argument is not a strong one.

include/alpaka/acc/AccCpuSerial.hpp
```diff
@@ -148,7 +149,8 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getAccName() -> std::string
             {
-                return "AccCpuSerial<" + std::to_string(TDim::value) + "," + typeid(TIdx).name() + ">";
+                return "AccCpuSerial<" + std::to_string(TDim::value) + "," + core::demangled<TIdx> + ">";
+                // return core::demangled<AccCpuSerial<TDim, TIdx>>;
```
bernhardmgruber:
```suggestion
```

bernhardmgruber (dismissed):
Please remove that one comment and we are good!

bernhardmgruber (dismissed):

tonydp03:
Still need to understand what's failing here though :)

include/alpaka/kernel/Traits.hpp
```diff
@@ -241,8 +242,8 @@ namespace alpaka
             "The idx type of TAcc and the idx type of TWorkDiv have to be identical!");
 
 #if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
-        std::cout << __func__ << " workDiv: " << workDiv << ", kernelFnObj: " << typeid(kernelFnObj).name()
-                  << std::endl;
+        std::cout << __func__ << " workDiv: " << workDiv
+                  << ", kernelFnObj: " << core::demangled<kernelFnObj> << std::endl;
```
bernhardmgruber: You need to pass a type here:
```suggestion
                  << ", kernelFnObj: " << core::demangled<decltype(kernelFnObj)> << std::endl;
```

bernhardmgruber (changes requested):

include/alpaka/kernel/Traits.hpp
tonydp03: Yeah it's kinda strange, as gcc was complaining in some other place (on my machine) but not there. Whatever :)

bernhardmgruber (approved):

Issue #642, opened 2018-09-17
Make ViewPlainPtr more convenient to use


krzikalla:
Following code:

```
alpaka::vec::Vec<alpaka::dim::DimInt<1>, size_t> size;
using TBuf = decltype(alpaka::mem::buf::alloc<T, size_t>(device, size));
auto buffer = alpaka::mem::buf::alloc<T, size_t>(device, size);
T* native_buf = static_cast<T*>(alpaka::mem::view::getPtrNative(buffer));
//...
ViewPlainPtr<???> view_buf(native_buf, device, size);
```

How can I figure out the template parameters of ViewPlainPtr (or the type altogether), given, that I can use TBuf at that code line? A function template would already help wrt. TDev and TElem. What would also be of a great help are some member typedefs in TBuf. Eventually I'd like to write

```
auto view_buf = GetViewPlainPtr(native_buf, device, size);
```

This one fails:
```
  template<typename TDev, typename TElem, typename TExtent>
  auto getViewPlainPtr(TElem* ptr, TDev const& dev, TExtent const & extent)
  {
    using namespace alpaka;
    return mem::view::ViewPlainPtr<TDev, TElem, dim::traits::DimType<TExtent>, size::traits::SizeType<TExtent>>(ptr, dev, extent);
  }
```
How do I figure out the dim type and the size type?

BenjaminW3:
Alpaka mostly uses the way of specialization of exernal trait templates instead of member typedefs.

>How do I figure out the dim type and the size type?

To get the `Dim` of a type you can use `using BufDim = alpaka::dim::Dim<TBuf>;` and the size is equivalent. (Note that the `Size` type has been renamed to `Idx` on the develop branch).

> How can I figure out the template parameters of ViewPlainPtr (or the type altogether), given, that I can use TBuf at that code line?

Final type of a `ViewPlainPtr` given any other view (buffer itself, ViewSubView, ViewPlainPtr, ...) should be something like:
```
template<
    typename TView>
using ViewPlainPtrFromView = alpaka::mem::view::ViewPlainPtr<
    alpaka::dev::Dev<TView>,
    alpaka::elem::Elem<TView>,
    alpaka::dim::Dim<TView>,
    alpaka::idx::Idx<TIdx>>;
```

It is identical for a `ViewSubView`:
```
template<
    typename TView>
using ViewSubViewFromView = alpaka::mem::view::ViewSubView<
    alpaka::dev::Dev<TView>,
[...]

PR #1743, opened 2022-06-13
Allow temporary destination views in memset/memcpy

### LINKED ISSUE #643: alpaka::mem::view::copy takes non-const references to views
Follow-up to #642 

Assuming I have a function template giving me a view I cannot write the following:

```
template<typename TDev, typename TElem, typename TExtent>
static auto getViewPlainPtr(TElem* ptr, TDev const& dev, TExtent const & extent)
{
  return mem::view::ViewPlainPtr<TDev, TElem, alpaka::dim::DimInt<1>, size_t>(ptr, dev, extent);
}

mem::view::copy(stream, getViewPlainPtr(...), buffer, size);
```

The problem is, that copy takes the destination view as a non-const reference. However the constness of the memory pointed to cannot be expressed by the constness of the view but must be encoded internally (STL solves this via iterator/const_iterator). A solution could be to handle Elem and const Elem template paramters appropriately.

### LINKED ISSUE #1742: Allow temporary destination views in `alpaka::memcpy`
Alpaka's memcpy is specified as:
```c++
template<typename TViewSrc, typename TViewDst, typename TQueue>
auto memcpy(TQueue& queue, TViewDst& viewDst, TViewSrc const& viewSrc) -> void;
```
Because the `viewDst` parameter is taken by reference, one cannot pass temporaries:
```c++
alpaka::memcpy(queue, alpaka::createView(dev, pointer, size), srcView); // compilation error
```
If we would take the destination parameter by forwarding reference instead, this would work, which is a nice usability improvement over:
```c++
auto dstView = alpaka::createView(dev, pointer, size);
alpaka::memcpy(queue, dstView, srcView); // works
```

bernhardmgruber:
This changeset allows temporary view objects to be passed to `alpaka::memset` and `alpaka::memcpy`. It also contains a refactoring of TaskCopyOacc, before applying the changes.

Fixes: #1742 and #642

### DIFF
--- a/include/alpaka/mem/buf/oacc/Copy.hpp
+++ b/include/alpaka/mem/buf/oacc/Copy.hpp
@@ -62,10 +62,10 @@ namespace alpaka
                 return TTask<TDim, TViewDst, TViewSrc, TExtent, TCopyPred>(viewDst, viewSrc, extent, dev, copyPred);
             }
 
-            //! The OpenACC device memory copy task base.
+            //! The OpenACC Nd device memory copy task.
             //!
             template<typename TDim, typename TViewDst, typename TViewSrc, typename TExtent, typename TCopyPred>
-            struct TaskCopyOaccBase
+            struct TaskCopyOacc
             {
                 using ExtentSize = alpaka::Idx<TExtent>;
                 using DstSize = alpaka::Idx<TViewDst>;
@@ -91,7 +91,7 @@ namespace alpaka
 
                 using Idx = alpaka::Idx<TExtent>;
 
-                ALPAKA_FN_HOST TaskCopyOaccBase(
+                ALPAKA_FN_HOST TaskCopyOacc(
                     TViewDst& viewDst,
                     TViewSrc const& viewSrc,
                     TExtent const& extent,
@@ -127,6 +127,54 @@ namespace alpaka
                               << std::endl;
                 }
 #    endif
+
+                ALPAKA_FN_HOST auto operator()() const -> void
+                {
+                    ALPAKA_DEBUG_FULL_LOG_SCOPE;
+
+#    if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
+                    printDebug();
+#    endif
+                    if constexpr(TDim::value == 1)
+                    {
+                        if(m_extent.prod() != 0)
+                        {
+                            m_dev.makeCurrent();
+                            m_copyPred(
+                                reinterpret_cast<void*>(m_dstMemNative),
+                                const_cast<void*>(reinterpret_cast<void const*>(m_srcMemNative)),
+                                static_cast<std::size_t>(m_extentWidthBytes));
+                        }
+                    }
+                    else
+                    {
+                        using DimMin1 = DimInt<TDim::value - 1u>;
+
+                        Vec<DimMin1, ExtentSize> const extentWithoutInnermost(subVecBegin<DimMin1>(m_extent));
+                        // [z, y, x] -> [y, x] because the z pitch (the full size of the buffer) is not required.
+                        Vec<DimMin1, DstSize> const dstPitchBytesWithoutOutmost(subVecEnd<DimMin1>(m_dstPitchBytes));
+                        Vec<DimMin1, SrcSize> const srcPitchBytesWithoutOutmost(subVecEnd<DimMin1>(m_srcPitchBytes));
+
+                        if(static_cast<std::size_t>(m_extent.prod()) != 0u)
+                        {
+                            m_dev.makeCurrent();
+                            meta::ndLoopIncIdx(
+                                extentWithoutInnermost,
+                                [&](Vec<DimMin1, ExtentSize> const& idx)
+                                {
+                                    m_copyPred(
+                                        reinterpret_cast<void*>(
+                                            m_dstMemNative
+                                            + (castVec<DstSize>(idx) * dstPitchBytesWithoutOutmost).sum()),
+                                        const_cast<void*>(reinterpret_cast<const void*>(
+                                            m_srcMemNative
+                                            + (castVec<SrcSize>(idx) * srcPitchBytesWithoutOutmost).sum())),
+                                        static_cast<std::size_t>(m_extentWidthBytes));
+                                });
+                        }
+                    }
+                }
+
                 const DevOacc m_dev;
                 Vec<TDim, ExtentSize> m_extent;
                 ExtentSize const m_extentWidthBytes;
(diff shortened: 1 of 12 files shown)


include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -48,24 +48,29 @@ namespace alpaka
...
             auto makeTaskCopyOacc(
-                TViewDst& viewDst,
+                TViewDstFwd&& viewDst,
                 TViewSrc const& viewSrc,
                 TExtent const& extent,
                 DevOacc const& dev,
                 TCopyPred copyPred)
             {
-                return TTask<TDim, TViewDst, TViewSrc, TExtent, TCopyPred>(viewDst, viewSrc, extent, dev, copyPred);
+                return TTask<TDim, std::remove_reference_t<TViewDstFwd>, TViewSrc, TExtent, TCopyPred>(
```
sbastrakov, line 62: Could `TViewDstFwd` possibly be a reference type here? If so, in which situation? (and there are a few similar cases in this PR, so i guess you made it for a reason, i just do not understand it yet)

include/alpaka/mem/view/Traits.hpp
```diff
@@ -165,31 +169,32 @@ namespace alpaka
     //! \param view The memory view to fill.
     //! \param byte Value to set for each element of the specified view.
     //! \param extent The extent of the view to fill.
-    template<typename TExtent, typename TView, typename TQueue>
-    ALPAKA_FN_HOST auto memset(TQueue& queue, TView& view, std::uint8_t const& byte, TExtent const& extent) -> void
+    template<typename TExtent, typename TViewFwd, typename TQueue>
+    ALPAKA_FN_HOST auto memset(TQueue& queue, TViewFwd&& view, std::uint8_t const& byte, TExtent const& extent) -> void
```
sbastrakov, line 173: I think we should clarify in the doxygen comment why are we taking `view` as a forwarding reference, so what a user should expect

Unrelated to this PR, imo this parameter should also be marked as `[out]`

(same applies to `memcopy` below)

include/alpaka/mem/buf/oacc/Copy.hpp
jkelling: If any calling code passes an lvalue `viewDst` to `makeTaskCopyOacc`, the type of `TViewDstFwd` will be an lvalue refrence.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -284,92 +270,89 @@ namespace alpaka
...
-                        viewSrc,
-                        extent,
-                        getDev(viewDst),
-                        acc_memcpy_to_device);
+                auto devDst = getDev(viewDst);
+                return alpaka::oacc::detail::makeTaskCopyOacc<alpaka::oacc::detail::TaskCopyOacc, TDim>(
+                    std::forward<TViewDstFwd>(viewDst),
+                    viewSrc,
+                    extent,
+                    std::move(devDst),
```
jkelling, line 286: Is constructing `devDst` first and then moving it a style choice? Constructing it in-place would do the same.

Although, the move is superfluous, because `makeTaskCopyOacc` takes `devDst` by const reference.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -284,92 +270,89 @@ namespace alpaka
...
-                            viewDst,
-                            viewSrc,
-                            extent,
-                            getDev(viewDst),
-                            acc_memcpy_device);
+                    return alpaka::oacc::detail::makeTaskCopyOacc<alpaka::oacc::detail::TaskCopyOacc, TDim>(
+                        std::forward<TViewDstFwd>(viewDst),
+                        viewSrc,
+                        extent,
+                        std::move(devDst),
```
jkelling, line 333: Same, as above: moving into a const ref.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -284,92 +270,89 @@ namespace alpaka
...
-                                devSrc.makeCurrent();
-                                acc_memcpy_from_device(buf.get(), src, size);
-                                devDst.makeCurrent();
-                                acc_memcpy_to_device(dst, buf.get(), size);
-                            });
+                    return alpaka::oacc::detail::makeTaskCopyOacc<alpaka::oacc::detail::TaskCopyOacc, TDim>(
+                        std::forward<TViewDstFwd>(viewDst),
+                        viewSrc,
+                        extent,
+                        devDst,
```
jkelling, line 343: This is inconsistent with the move above, in the other branch of the `#if _OPENACC >= 201510`.

jkelling (changes requested):

include/alpaka/mem/buf/oacc/Copy.hpp
bernhardmgruber: I suffixed all forwarding references with `Fwd`. So these template parameters can hold either a `T` or a `T&` (which reference collapses to `T&&` and `T&`).

include/alpaka/mem/buf/oacc/Copy.hpp
bernhardmgruber: I need to get the `devDst` from `viewDst` before the call to `makeTaskCopyOacc`, because `viewDst` is potentially moved by `std::forward`, so calling `getDev(viewDst)` inside the argument list might be a use-after-move.

include/alpaka/mem/buf/oacc/Copy.hpp
bernhardmgruber: I cannot move `devDst` here, because it is also captured by the lambda.

include/alpaka/mem/buf/oacc/Copy.hpp
sbastrakov: Thanks for explanations!

include/alpaka/mem/buf/oacc/Copy.hpp
bernhardmgruber: I changed `makeTaskCopyOacc` and the `TaskCopyOacc` to accept the device by value and move it into the class member.

include/alpaka/mem/view/Traits.hpp
bernhardmgruber: > Unrelated to this PR, imo this parameter should also be marked as [out]

Could you please clarify what you mean here? Would you like to see a comment `/* out */`?

sbastrakov: As doxygen `\param[out]`, or do we not use this style in alpaka? (if not, then all fine as-is)

bernhardmgruber: I added some documentation and marked the parameter as `[in, out]`, as the destination view needs to be correctly initialized (allocated, right extent) before passing it.

include/alpaka/mem/buf/oacc/Copy.hpp
```diff
@@ -48,24 +48,29 @@ namespace alpaka
...
+                TViewDstFwd&& viewDst,
                 TViewSrc const& viewSrc,
                 TExtent const& extent,
-                DevOacc const& dev,
+                DevOacc dev,
                 TCopyPred copyPred)
             {
-                return TTask<TDim, TViewDst, TViewSrc, TExtent, TCopyPred>(viewDst, viewSrc, extent, dev, copyPred);
+                return TTask<TDim, std::remove_reference_t<TViewDstFwd>, TViewSrc, TExtent, TCopyPred>(
+                    viewDst,
```
jkelling: I think this call would create a copy even is the initial call was made with a reference. Should this not be forwarded?
```suggestion
                    std::forward<TViewDst>(viewDst),
```

jkelling (changes requested):

include/alpaka/mem/buf/oacc/Copy.hpp
bernhardmgruber: Good catch!

jkelling (approved):

j-stephan (approved):

Issue #1729, opened 2022-05-30
alpaka buffers do not respect `const` correctness ? 


fwyzard:
As we try to use alpaka Buffer objects in the CMS software, we are facing a problem with `const` correctness due to the way the buffers are implemented.

Here is a sample program, compiled with `g++ -std=c++17 -O3 -Wall -I/home/fwyzard/src/alpaka-group/alpaka/include -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED test.cc -o test`:
```c++
#include <alpaka/alpaka.hpp>

template <typename TBuf>
void modify_const_buffer(TBuf buffer) {
  // buffer is not const, and can be used to modify the contents of the originally const buffer:
  buffer[42] = 0;
}

int main(void) {
  using Dim = alpaka::DimInt<1u>;
  using Idx = std::size_t;
  using Host = alpaka::AccCpuSerial<Dim, Idx>;

  const auto host = alpaka::getDevByIdx<Host>(0u);

  const auto buffer = alpaka::allocBuf<int, Idx>(host, alpaka::Vec<Dim, Idx>(64u));
  // buffer is const, so it's not possible to write into it:
  /*
test.cc: In function ‘int main()’:
test.cc:22:14: error: assignment of read-only location ‘((const alpaka::internal::ViewAccessOps<alpaka::BufCpu<int, std::integral_constant<long unsigned int, 1>, long unsigned int> >*)(& buffer))->alpaka::internal::ViewAccessOps<alpaka::BufCpu<int, std::integral_constant<long unsigned int, 1>, long unsigned int> >::operator[](42)’
   22 |   buffer[42] = 0;
  */
  //buffer[42] = 0;

  modify_const_buffer(buffer);

  return 0;
}
```

The issue is that the original `buffer` is `const`, but any non-`const` copy of `buffer` can be used to modify the same data.

Is this the expected behaviour ?

sbastrakov:
I believe your report concerns not const correctness, as this one actually holds for buffers and one can't modify elements of a `const` buffer. However, what alpaka does in this example is automatically casting const buffers to non-const ones e.g. here when passing to a function. Doing a similar thing implicitly (without a `const_cast`) with raw pointers would be a compiler warning/error, but not with our buffers.

I am not sure if this behavior is by design or by accident, we should look into it. I am personally inclined to not allowing it (but again there may be some intent there which i am not familiar with)

fwyzard:
[...]

PR #1746, opened 2022-06-21
Add ViewConst

bernhardmgruber:
This view wraps another view but only allows unmodifying access to the wrapped view.

Related to #1729.

### DIFF
--- a/include/alpaka/mem/view/Traits.hpp
+++ b/include/alpaka/mem/view/Traits.hpp
@@ -223,9 +223,6 @@ namespace alpaka
         static_assert(
             meta::IsIntegralSuperset<SrcIdx, ExtentIdx>::value,
             "The source view and the extent are required to have compatible index types!");
-        static_assert(
-            std::is_same_v<DstElem, std::remove_const_t<SrcElem>>,
-            "The source and the destination view are required to have the same element type!");
 
         return trait::CreateTaskMemcpy<Dim<TViewDst>, Dev<TViewDst>, Dev<TViewSrc>>::createTaskMemcpy(
             std::forward<TViewDstFwd>(viewDst),
--- a/include/alpaka/mem/view/ViewConst.hpp
+++ b/include/alpaka/mem/view/ViewConst.hpp
@@ -1,0 +1,120 @@
+/* Copyright 2022 Bernhard Manfred Gruber
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#pragma once
+
+#include <alpaka/core/Common.hpp>
+#include <alpaka/dev/Traits.hpp>
+#include <alpaka/dim/Traits.hpp>
+#include <alpaka/extent/Traits.hpp>
+#include <alpaka/idx/Traits.hpp>
+#include <alpaka/mem/view/Traits.hpp>
+#include <alpaka/mem/view/ViewAccessOps.hpp>
+#include <alpaka/offset/Traits.hpp>
+
+namespace alpaka
+{
+    //! A non-modifiable wrapper around a view. This view acts as the wrapped view, but the underlying data is only
+    //! exposed const-qualified.
+    template<typename TView>
+    struct ViewConst : internal::ViewAccessOps<ViewConst<TView>>
+    {
+        static_assert(!std::is_const_v<TView>, "ViewConst must be instantiated with a non-const type");
+        static_assert(
+            !std::is_reference_v<TView>,
+            "This is not implemented"); // It might even be dangerous for ViewConst to store a reference to the wrapped
+                                        // view, as this decouples the wrapped view's lifetime.
+
+        ALPAKA_FN_HOST_ACC ViewConst(const TView& view) : m_view(view)
+        {
+        }
+
+        ALPAKA_FN_HOST_ACC ViewConst(TView&& view) : m_view(std::move(view))
+        {
+        }
+
+        TView m_view;
+    };
+
+    template<typename TView>
+    ViewConst(TView) -> ViewConst<std::decay_t<TView>>;
+
+    namespace trait
+    {
+        template<typename TView>
+        struct DevType<ViewConst<TView>> : DevType<TView>
+        {
+        };
+
+        template<typename TView>
+        struct GetDev<ViewConst<TView>>
+        {
+            ALPAKA_FN_HOST static auto getDev(ViewConst<TView> const& view)
+            {
+                return alpaka::getDev(view.m_view);
+            }
+        };
+
+        template<typename TView>
+        struct DimType<ViewConst<TView>> : DimType<TView>
+        {
+        };
+
+        template<typename TView>
+        struct ElemType<ViewConst<TView>>
+        {
+            // const qualify the element type of the inner view
+            using type = typename ElemType<TView>::type const;
+        };
+
+        template<typename I, typename TView>
+        struct GetExtent<I, ViewConst<TView>>
+        {
+            ALPAKA_FN_HOST_ACC static auto getExtent(ViewConst<TView> const& view)
+            {
+                return alpaka::getExtent<I::value>(view.m_view);
+            }
+        };
+
+        template<typename TView>
+        struct GetPtrNative<ViewConst<TView>>
+        {
+            using TElem = typename ElemType<TView>::type;
+
+            // const qualify the element type of the inner view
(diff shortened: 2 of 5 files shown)


include/alpaka/mem/view/ViewConst.hpp
```diff
@@ -0,0 +1,97 @@
...
+        struct GetExtent<TIdxIntegralConst, ViewConst<TView>> : GetExtent<TIdxIntegralConst, TView>
+        {
+        };
+
+        template<typename TView>
+        struct GetPtrNative<ViewConst<TView>>
+        {
+            using TElem = typename ElemType<TView>::type;
+
+            ALPAKA_FN_HOST_ACC static auto getPtrNative(ViewConst<TView> const& view) -> TElem const*
```
bernhardmgruber, line 91: Here we always return a pointer with a `const` qualified pointee type.

include/alpaka/mem/view/ViewConst.hpp
```diff
@@ -0,0 +1,100 @@
...
+        struct DimType<ViewConst<TView>> : DimType<TView>
+        {
+        };
+
+        template<typename TView>
+        struct ElemType<ViewConst<TView>>
+        {
+            // const qualify the element type of the inner view
+            using type = typename ElemType<TView>::type const;
+        };
```
bernhardmgruber, line 74: Here we just `const` qualify the element type. This should make `internal::ViewAccessOps<ViewConst<TView>>` work.

bernhardmgruber:
@tonydp03 please provide feedback if `ViewConst` solves your usecase. Thank you!

tonydp03:
> @tonydp03 please provide feedback if `ViewConst` solves your usecase. Thank you!

To me it seems consistent and the interface should not be that that tricky. `auto new_buffer = alpaka::ViewConst(buffer)` creates a `const view` that does not allow to modify the original buffer (not with just an assignment, nor through copy).

My only concern is that still, on user side, if one has a `const buffer` and types `auto new_buffer = buffer`, then it can modify it. Unless we can do such that such a line would return a `ViewConst` rather than a `buffer`, I think we should document it.

bernhardmgruber:
> > @tonydp03 please provide feedback if `ViewConst` solves your usecase. Thank you!
> 
> To me it seems consistent and the interface should not be that that tricky. `auto new_buffer = alpaka::ViewConst(buffer)` creates a `const view` that does not allow to modify the original buffer (not with just an assignment, nor through copy).

So `ViewConst` solves your use case?

> My only concern is that still, on user side, if one has a `const buffer` and types `auto new_buffer = buffer`, then it can modify it. Unless we can do such that such a line would return a `ViewConst` rather than a `buffer`, I think we should document it.

This cannot be solved by `ViewConst`. We could attempt to make 
```c++
const auto buffer = ...;
auto new_buffer = buffer;
```
a compile error by deleting the copy ctor and adding a new "copy" ctor with a mutable reference argument. But that feels like a very cruel hack. Anyway, this is outside of this PR.

fwyzard:
> deleting the copy ctor and adding a new "copy" ctor with a mutable reference argument.

This sounds reasonable to me -- but it seems to prevent `std::optonal<TBuf>` from compiling :-/

tonydp03 (dismissed):

bernhardmgruber:
I extended the unit tests for immutable views now to also test copying *from* an immutable view.

bernhardmgruber:
If you use a `ViewConst` as destination for e.g. a `alpaka::memcpy`, you will rightfully get the following error:
```
/home/bgruber/dev/alpaka/include/alpaka/mem/view/Traits.hpp:210:23: error: static assertion failed: The destination view's element type must not be const!
  210 |         static_assert(!std::is_const_v<DstElem>, "The destination view's element type must not be const!");
      |                       ^~~~~~~~~~~~~~~~~~~~~~~~~
/home/bgruber/dev/alpaka/include/alpaka/mem/view/Traits.hpp:210:23: note: ‘!(bool)std::is_const_v<const float>’ evaluates to false
/home/bgruber/dev/alpaka/include/alpaka/mem/view/Traits.hpp:218:18: error: static assertion failed: The source and destination view must have the same element type!
  218 |             std::is_same_v<DstElem, std::remove_const_t<SrcElem>>,
      |             ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

tonydp03:
> If you use a `ViewConst` as destination for e.g. a `alpaka::memcpy`, you will rightfully get the following error:
> 
> ```
> /home/bgruber/dev/alpaka/include/alpaka/mem/view/Traits.hpp:210:23: error: static assertion failed: The destination view's element type must not be const!
>   210 |         static_assert(!std::is_const_v<DstElem>, "The destination view's element type must not be const!");
>       |                       ^~~~~~~~~~~~~~~~~~~~~~~~~
> /home/bgruber/dev/alpaka/include/alpaka/mem/view/Traits.hpp:210:23: note: ‘!(bool)std::is_const_v<const float>’ evaluates to false
> /home/bgruber/dev/alpaka/include/alpaka/mem/view/Traits.hpp:218:18: error: static assertion failed: The source and destination view must have the same element type!
>   218 |             std::is_same_v<DstElem, std::remove_const_t<SrcElem>>,
>       |             ~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> ```

So basically if one wants to copy a `const buffer` from a device to another, it should first use `alpaka::memcopy` with a `non-const buffer` as destination, and then create a `ViewConst` which stores a copy of it, correct?

bernhardmgruber:
> So basically if one wants to copy a const buffer from a device to another, it should first use alpaka::memcopy with a non-const buffer as destination, and then create a ViewConst which stores a copy of it, correct?

Yes.

psychocoderHPC:
Could you please ammend to your last commit and force push again? For some reason, the HZDR CI is not creating the job.

bernhardmgruber:
> Could you please ammend to your last commit and force push again? For some reason, the HZDR CI is not creating the job.

I already did that a few days ago. But let's try again.

j-stephan:
Sorry for the noise @tonydp03, I clicked the wrong button. I'm temporarily unassigning you at the request of the CI admins.

bernhardmgruber:
@j-stephan the GitLab CI has been triggered now. The culprit was indeed that I assigned someone to the PR.

tonydp03 (dismissed):

include/alpaka/mem/view/ViewConst.hpp
```diff
@@ -0,0 +1,125 @@
...
+        }
+
+        ALPAKA_FN_HOST_ACC ViewConst(TView&& view) : m_view(std::move(view))
+        {
+        }
+
+        ViewConst(const ViewConst&) = default;
+        ViewConst(ViewConst&&) noexcept = default;
+        auto operator=(const ViewConst&) -> ViewConst& = default;
+        auto operator=(ViewConst&&) noexcept -> ViewConst& = default;
```
j-stephan: Didn't we decide to remove defaulted special members?

j-stephan (changes requested):
Looks good to me, but please clarify:

include/alpaka/mem/view/Traits.hpp
```diff
@@ -223,9 +223,6 @@ namespace alpaka
         static_assert(
             meta::IsIntegralSuperset<SrcIdx, ExtentIdx>::value,
             "The source view and the extent are required to have compatible index types!");
-        static_assert(
```
psychocoderHPC, line 226: Should we not keep this assert?

psychocoderHPC (dismissed):

j-stephan:
This also needs a rebase to reflect the changes to macOS.

include/alpaka/mem/view/ViewConst.hpp
bernhardmgruber: Good catch, but for a different reason. In principle, you are right that I prefer to not have defaulted special members, if they are provided by the compiler anyway. Here, I created them originally because I only had a constructor taking a forwarding reference:
```c++
template <typename TViewFwd>
ViewConst(TViewFwd&&) { ... }
```
And that interacted with the compiler provided copy constructor unfortunately (I don't remember what it was now), so I needed to declare the copy constructors separately. And then clang-tidy complains when you don't declare the copy-assignment operators as well :)

Now however, I switched to just having two constructors with concrete types, so the problem is gone and I can remove the defaulted members.

include/alpaka/mem/view/Traits.hpp
bernhardmgruber: No because the same assert is also on line 217 :)

bernhardmgruber:
Rebased and defaulted special members removed.

j-stephan (approved):

PR #1749, opened 2022-06-22
Upgrade to Catch2 v3

### LINKED ISSUE #1745: Migrate to Catch2 v3.x
The first stable version of Catch2 v3 was released last month. We should migrate our test infrastructure (thankfully they provide a porting guide).

j-stephan:
This PR upgrades our Catch2 infrastructure to version 3. Notable changes:

* `thirdParty/catch2` now includes the full Catch2 source code so it can be built as a static library. This is the recommended approach as mentioned in the Catch2 v2 -> v3 migration guide. Users can still turn this off by using the known CMake configuration flags (nothing changed here). In this case CMake will search for a system installation of Catch2.
* Replaced all instances of `catch2/catch.hpp` with the appropriate modular headers (as required by the individual translation unit).
* Deleted the `test/catch_main` folder as it is no longer required.

Fixes #1745.


test/integ/hostOnlyAPI/src/hostOnlyAPI.cpp
```diff
@@ -58,39 +60,39 @@ TEMPLATE_LIST_TEST_CASE("hostOnlyAPI", "[hostOnlyAPI]", TestAccs)
 
     // CPU host
     auto const host = alpaka::getDevByIdx<Host>(0u);
-    INFO("using alpaka device: " << alpaka::getName(host))
+    INFO("using alpaka device: " << alpaka::getName(host));
```
j-stephan, line 63: This used to work with Catch v2; in v3 the semicolon is required.

thirdParty/CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
...
+if(BUILD_TESTING)
+    if(alpaka_USE_INTERNAL_CATCH2)
+        message(STATUS "Catch2: Using INTERNAL version 3.0.1")
+        # Force Catch2's CMake to pick up the variables we set below
+        set(CMAKE_POLICY_DEFAULT_CMP0077 NEW)
+
+        set(CATCH_CONFIG_FAST_COMPILE ON)
+        if(WIN32)
+            set(CATCH_CONFIG_WINDOWS_CRTDBG ON)
+        endif()
```
bernhardmgruber, line 20:
```suggestion
        if(WIN32)
            set(CATCH_CONFIG_WINDOWS_CRTDBG ON)
        endif()
```
Awesome!

bernhardmgruber (dismissed):

bernhardmgruber (approved):

PR #1751, opened 2022-06-24
Set policy CMP0091 to NEW

### LINKED ISSUE #1744: Switch CMP0091 to NEW
When I configure alpaka on Windows, I get this warning:
```
 CMake Deprecation Warning at CMakeLists.txt:16 (cmake_policy):
   The OLD behavior for policy CMP0091 will be removed from a future version
   of CMake.

   The cmake-policies(7) manual explains that the OLD behaviors of all
   policies are deprecated and that a policy should be set to OLD only under
   specific short-term circumstances.  Projects should be ported to the NEW
   behavior and not rely on setting a policy to OLD.
```
Because alpaka sets: `cmake_policy(SET CMP0091 OLD)`.

We should migrate to the new behavior of cmake and remove the line causing the warning.

bernhardmgruber:
This was introduced in e7fae73e7786 to fix the VS2017 build. Support for
VS2017 has been dropped in the meantime.

Fixes: #1744

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -13,7 +13,7 @@
 
 cmake_minimum_required(VERSION 3.18)
 
-cmake_policy(SET CMP0091 OLD)
+cmake_policy(SET CMP0091 NEW)
 
 include(CMakePrintHelpers)
 


CMakeLists.txt
```diff
@@ -13,8 +13,6 @@
 
 cmake_minimum_required(VERSION 3.18)
 
-cmake_policy(SET CMP0091 OLD)
-
```
j-stephan: This should be set to `NEW` instead. From the CMake documentation:

> This policy was introduced in CMake version 3.15. Use the [`cmake_policy()`](https://cmake.org/cmake/help/v3.18/command/cmake_policy.html#command:cmake_policy) command to set it to `OLD` or `NEW` explicitly. Unlike many policies, CMake version 3.18.6 does not warn when this policy is not set and simply uses `OLD` behavior.

If the `NEW` behaviour causes problems for us it would be useful to catch it now instead of being surprised later.

j-stephan (changes requested):

CMakeLists.txt
bernhardmgruber: Yeah,  that's a good idea!

j-stephan (approved):

PR #1756, opened 2022-06-29
OpenMP fixes for clang 13

### LINKED ISSUE #1755: clang-13: OpenMP support broken when using CUDA
Another bug uncovered by #1713. clang-13's default OpenMP support is broken when using CUDA:

```
/usr/lib/llvm-13/lib/clang/13.0.1/include/omp.h(487): error: function "omp_is_initial_device" has already been defined
```

This is related to clang-13 enabling OpenMP 5 by default; lowering the OpenMP version seems to silence this error. See this related issue on NVIDIA's issue tracker:

https://github.com/NVIDIA/cuda-samples/issues/102

j-stephan:
With this PR, the OpenMP version is set to 4.5 if:

* clang 13 or Xcode 13.4 are used
* CUDA is enabled (`nvcc` as device compiler)
* The OpenMP 5 back-end is NOT enabled.

If OpenMP 5 is explicitly enabled, CMake will print an informative error message instead.

I also slighty changed the way for checking for Clang / AppleClang OpenMP 5 support.

Fixes #1755.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -359,12 +359,12 @@ if(alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE OR a
         if(alpaka_ACC_ANY_BT_OMP5_ENABLED)
             if(OpenMP_CXX_VERSION VERSION_LESS 5.0)
                 message(FATAL_ERROR "alpaka_ACC_ANY_BT_OMP5_ENABLE requires compiler support for OpenMP 5.0.")
+            endif()
 
-                if((${CMAKE_CXX_COMPILER_ID} STREQUAL "AppleClang") AND (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 12.0.5))
-                    message(FATAL_ERROR "The OpenMP 5.0 back-end requires Xcode 12.5 or later")
-                elseif((${CMAKE_CXX_COMPILER_ID} STREQUAL "Clang") AND (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 11.0))
-                    message(FATAL_ERROR "The OpenMP 5.0 back-end requires clang 11.0 or later")
-                endif()
+            if((${CMAKE_CXX_COMPILER_ID} STREQUAL "AppleClang") AND (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 12.0.5))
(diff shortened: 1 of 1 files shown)


bernhardmgruber (approved):

bernhardmgruber:
This LGTM, but I would like to have a second opinion.

j-stephan:
What are you unsure about?

cmake/alpakaCommon.cmake
```diff
@@ -505,6 +518,21 @@ if(alpaka_ACC_GPU_CUDA_ENABLE)
...
                 if(NOT MSVC)
                     alpaka_set_compiler_options(DEVICE target alpaka $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-fopenmp>)
+
+                    # See https://github.com/alpaka-group/alpaka/issues/1755
+                    if((${CMAKE_CXX_COMPILER_ID} STREQUAL "AppleClang") AND
+                       (${CMAKE_CXX_COMPILER_ID} VERSION_GREATER_EQUAL 13.1.6) AND
+                       (${CMAKE_CXX_COMPILER_VERSION} VERSION_LESS 14.0.0))
+                       message(STATUS "Xcode 13.4 detected. Force-setting OpenMP to version 4.5.")
+                       alpaka_set_compiler_options(DEVICE target alpaka $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=-fopenmp-version=45>)
+                    endif()
```
bernhardmgruber, line 528: It is a bit hard to understand the interaction with your previous change. So inside this `if` clause, `alpaka_ACC_ANY_BT_OMP5_ENABLE` must be `OFF`, because that combination has been checked previously and cmake should have aborted with a fatal error. I worry that subsequent changes could forget about keeping the conditions here and further up in sync.

I kind of would like to have an `assert(alpaka_ACC_ANY_BT_OMP5_ENABLE == OFF)` here, but cmake does not have such a facility.

j-stephan: There is no interaction between the two changes. CMake parses the file line by line and immediately evaluates all function calls and variables it finds. So if we got down here it is already clear that we don't have the evil clang-13 + OMP5 + CUDA combination.

bernhardmgruber: Well, I know how cmake works :D

> So if we got down here it is already clear that we don't have the evil clang-13 + OMP5 + CUDA combination.

I find this not so clear. It is clear given the diff, because the relevant code sections are highlighted. But if we change this code in the future again, I am just afraid that we might forget something.

But anyway, I guess we can just move on, because the proposed changes are fine themselves.

PR #1767, opened 2022-07-25
HIP: use build-in `atomicAdd(double)`

psychocoderHPC:
HIP is providing for HIP 4.2+ native atomicAdd interfaces for double precision.
Use always the build-in atomicAdd so speedup atomic add.


psychocoderHPC:
Note: I am currently refactoring the atomicAdd implementations for CUDA/HIP to avoid emulating atomic's with atomicCAS if there is a build-in function available.

sbastrakov (approved):

PR #1768, opened 2022-07-27
Topic refactor cuda hip atomics

psychocoderHPC:
In https://github.com/alpaka-group/alpaka/pull/1767 I already enabled the usage of building atomicAdd for double precision.
Nevertheless, we are emulating in alpaka often atomics because it is a nightmare to maintain the list of supported atomic functions for HIP and CUDA. The list of supported types for different atomics differs between both models.

This PR introduces a "simple" way to let the compiler decide if a built-in function is available and use these. If not we will emulate the required atomics by using `atomicCas`.

- enable atomicCas for floating point data types
- emulate atomics if no builtin is available
- alpaka::AtomicCas add floating point support
- refactor atomic unit tests (remove many exceptions for CUDA/HIP)
- use builtin atomics for `hierarchy::Threads` -> mapped to `atomicXXX_block()`


include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+
+#    include <type_traits>
+
+namespace alpaka
+{
+    //! The GPU CUDA/HIP accelerator atomic ops.
+    //
+    //  Atomics can be used in the hierarchy level grids, blocks and threads.
+    //  Atomics are not guaranteed to be save between devices.
+    class AtomicUniformCudaHipBuiltIn
```
SimeonEhrig, line 27: Does it also works, if the CUDA and the HIP are enabled at the same time?

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+#    include <alpaka/core/Positioning.hpp>
+
+#    include <type_traits>
+
+namespace alpaka
+{
+    //! The GPU CUDA/HIP accelerator atomic ops.
+    //
+    //  Atomics can be used in the hierarchy level grids, blocks and threads.
+    //  Atomics are not guaranteed to be save between devices.
```
sbastrakov:
```suggestion
    //  Atomics are not guaranteed to be safe between devices.
```

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: As I know. You can not enable CUDA and HIP at the same time.

SimeonEhrig: Really, I thought Andrea implemented it? Or is it permitted by the CMake?

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+
+#        if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && !BOOST_LANG_CUDA
+#            error If ALPAKA_ACC_GPU_CUDA_ENABLED is set, the compiler has to support CUDA!
+#        endif
+
+#        if defined(ALPAKA_ACC_GPU_HIP_ENABLED) && !BOOST_LANG_HIP
+#            error If ALPAKA_ACC_GPU_HIP_ENABLED is set, the compiler has to support HIP!
+#        endif
+
+//! Inline namespace is required to check the existenz of function in global namespace via SFINAE.
```
sbastrakov:
```suggestion
//! These types must be in the global namespace for checking existence of respective functions in global namespace via SFINAE, so we use inline namespace.
```

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+inline namespace alpakaGlobal
+{
+    //! Provide an interface to builtin atomic functions.
+    //
+    // To check for the existence of builtin functions located in the global namespace :: directly.
+    // This is forbidden by C++.
```
sbastrakov: I do not understand this line. Did you mean smth like
```suggestion
    // This would not be possible without having these types in global namespace.
```
?

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+    // If the functor is inheriting from std::false_type an signature is explicitly not available. This can be used to
+    // explicitly disable builtin function in case the builtin is broken.
+    // If the functor is inheriting from std::true_type a spezialization must implement one of the following
```
sbastrakov:
```suggestion
    // If the functor is inheriting from std::true_type a specialization must implement one of the following
```

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+    // interfaces.
+    // \code{.cpp}
+    //    __device__ static T atomic( T* add, T value);
+    //    __device__ static T atomic( T* add, T compare, T value);
+    // \endcode
+    template<typename TOp, typename T, typename THierarchy, typename TSfinae = void, typename TDefer = void>
```
sbastrakov: I think doxygen comment on template parameters would be helpful here. Perhaps some parts of the comment above could be moved there.

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+    // Cas.
+    template<typename T, typename THierarchy>
+    struct AlpakaBuiltInAtomic<
+        alpaka::AtomicCas,
+        T,
+        THierarchy,
+        typename std::void_t<decltype(atomicCAS(std::declval<T*>(), std::declval<T>(), std::declval<T>()))>>
+        : std::true_type
+    {
+        __device__ static T atomic(T* add, T compare, T value)
```
sbastrakov, line 79: Why did you choose name `atomic()` and not `operator()` like math traits do? Not against that, just wondering about the reasoning

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+    {
+    };
+
+    // Cas.
+    template<typename T, typename THierarchy>
+    struct AlpakaBuiltInAtomic<
+        alpaka::AtomicCas,
+        T,
+        THierarchy,
+        typename std::void_t<decltype(atomicCAS(std::declval<T*>(), std::declval<T>(), std::declval<T>()))>>
```
sbastrakov: I believe on device side we must use `alpaka::core::declval` and not `std::declval`, please see the comment there.

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+
+// disable HIP atomicMin: see https://github.com/ROCm-Developer-Tools/hipamd/pull/40
+#        if(BOOST_LANG_HIP)
+    template<typename THierarchy>
+    struct AlpakaBuiltInAtomic<
+        alpaka::AtomicMin,
+        float,
+        THierarchy,
+        typename std::void_t<
+            decltype(atomicMin(std::declval<float*>(), std::declval<std::enable_if_t<sizeof(THierarchy), float>>()))>>
```
sbastrakov: I am not sure this part is needed. If we jsut want to hard-disable for HIP and given type, shouldn't we do it for any `TSfinae`?

Same point for specializations below.

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: As I know he is compiling a library for CUDA and a library for HIP and linking both together.
Both libs will use a different compiler.

@fwyzard Could you check if this PR is working for you. I do not exactly know your workflows.

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -0,0 +1,474 @@
...
+inline namespace alpakaGlobal
+{
+    //! Provide an interface to builtin atomic functions.
+    //
+    // To check for the existence of builtin functions located in the global namespace :: directly.
+    // This is forbidden by C++.
+    // If the functor is inheriting from std::false_type an signature is explicitly not available. This can be used to
+    // explicitly disable builtin function in case the builtin is broken.
+    // If the functor is inheriting from std::true_type a spezialization must implement one of the following
+    // interfaces.
```
sbastrakov, line 57: (Optional) we could clarify that all but CAS follow the first interface and CAS follows the second - there is no optionality as now suggested by the comment.

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: Yes! Your suggestion is more clear!

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: I expect the comment would be as useful as the comment from atomicOp. Therefore I decided to have no comment.

```
    //! Executes the given operation atomically.
    //!
    //! \tparam TOp The operation type.
    //! \tparam T The value type.
    //! \tparam TAtomic The atomic implementation type.
    //! \param addr The value to change atomically.
    //! \param value The value used in the atomic operation.
    //! \param atomic The atomic implementation.
    ALPAKA_NO_HOST_ACC_WARNING
    template<typename TOp, typename TAtomic, typename T, typename THierarchy = hierarchy::Grids>
  ALPAKA_FN_HOST_ACC auto atomicOp(
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -37,998 +30,301 @@ namespace alpaka
...
-            __device__ static auto atomicOp(
-                AtomicUniformCudaHipBuiltIn const&,
-                unsigned long int* const addr,
-                unsigned long int const& value) -> unsigned long int
-            {
-                // LLP64
-                if constexpr(std::numeric_limits<unsigned int>::max() == std::numeric_limits<unsigned long>::max())
-                    return ::atomicAdd(reinterpret_cast<unsigned int*>(addr), static_cast<unsigned int>(value));
-                else // LP64
+                //! reinterprets an adress to an 32bit value for atomicCas emulation usage
```
sbastrakov:
```suggestion
                //! reinterprets an address as a 32bit value for atomicCas emulation usage
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -37,998 +30,301 @@ namespace alpaka
...
+                template<typename TAddressType>
+                __device__ static auto reinterpretAddress(TAddressType* address)
+                    -> std::enable_if_t<sizeof(TAddressType) == 4u, unsigned int*>
```
sbastrakov, line 44: Should a fixed-width type be used instead of `unsigned int`? Then the 32-bit requirement is also more obvious. However I am not sure if it bodes well with CUDA and HIP.

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: because the method is static same as the methods in `AtomicOp`

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: Yes you are right it is not required!

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
sbastrakov: Also for later usage of the same phrase

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -37,998 +30,301 @@ namespace alpaka
...
-        template<typename THierarchy>
-        struct AtomicOp<AtomicInc, AtomicUniformCudaHipBuiltIn, int, THierarchy>
-        {
-            __device__ static auto atomicOp(AtomicUniformCudaHipBuiltIn const&, int* const addr, int const& value)
-                -> int
-            {
-                int old(*addr);
-                int assumed, eval;
-                do
+                //! reinterprets an value to be usable for the atomicCAS emulation
```
sbastrakov:
```suggestion
                //! reinterprets a value to be usable for the atomicCAS emulation
```

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
psychocoderHPC: No, we need to use a type where we know that the signature fits CUDA/HIP. If we use `std::uint32_t` it can link to `unsigned long int` or `unsigned int` depending on the platform.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -37,998 +30,301 @@ namespace alpaka
...
                     {
                         assumed = old;
-                        eval = assumed >= static_cast<unsigned long long int>(value) ? 0 : assumed + 1;
-                        old = ::atomicCAS(address_as_u, assumed, eval);
+                        T v = *(reinterpret_cast<T*>(&assumed));
+                        TOp{}(&v, value);
+                        using Cas = alpaka::trait::
+                            AtomicOp<alpaka::AtomicCas, alpaka::AtomicUniformCudaHipBuiltIn, EmulatedType, THierarchy>;
+                        old = Cas::atomicOp(ctx, addressAsIntegralType, assumed, reinterpretValue(v));
+                        // Note: uses integer comparison to avoid hang in case of NaN (since NaN != NaN)
```
sbastrakov, line 106: Do we have this case covered in tests? It should have been even before this PR, but since you are working on this part

psychocoderHPC: No, we do not cover this case.

psychocoderHPC: Our checks for atomicDec also do not test the flip over. https://github.com/alpaka-group/alpaka/issues/1769

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: I think I remember that we introduced the device API struct to avoid naming conflicts. Nevertheless, these changes are fully on the device side and should not create symbols on the host side which can conflict.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
sbastrakov: Okay, but then we should also check that `sizeof(unsigned int) == 4` ?

include/alpaka/atomic/AtomicUniformCudaHip.hpp
psychocoderHPC: I addressed it in the code snipped

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
sbastrakov: Discussed offline: maybe it's not worth as nothing would work with CUDA/HIP if `unsigned int` is not 4 bytes.

psychocoderHPC: The CUDA and HIP ecosystem require that `sizeof(int)==4`. Seeing it from the C++ standard view you are right but we do not guard it in many other places. 
We could at somewhere a global check for this condition but in reality, we will not run on systems where this is not fulfilled.

sbastrakov (dismissed):

include/alpaka/atomic/AtomicUniformCudaHip.hpp
fwyzard: Sorry, I have been doing night shifts this week, and had little time for the rest :-/

I'll try to have a look ASAP.

sbastrakov (approved):

PR #1770, opened 2022-07-28
test: extend ALPAKA_CHECK to show better error messages

psychocoderHPC:
To know where a test fails the line and file can be very useful.

before:
```
1: ALPAKA_CHECK failed because '!(equals(operandOrig, ret))'
1: ALPAKA_CHECK failed because '!(equals(operandOrig, ret))'
```

after:
```
1: ALPAKA_CHECK failed because '!(equals(operandOrig, ret))' in alpaka/test/unit/atomic/src/AtomicTest.cpp:90
1: ALPAKA_CHECK failed because '!(equals(operandOrig, ret))' in alpaka/test/unit/atomic/src/AtomicTest.cpp:96
```

### DIFF
--- a/include/alpaka/test/Check.hpp
+++ b/include/alpaka/test/Check.hpp
@@ -9,28 +9,33 @@
 
 #pragma once
 
+#include <alpaka/core/BoostPredef.hpp>
+
 #include <cstdio>
 
 // TODO: SYCL doesn't have a way to detect if we're looking at device or host code. This needs a workaround so that
 // SYCL and other back-ends are compatible.
 #ifdef ALPAKA_ACC_SYCL_ENABLED
-#    define ALPAKA_CHECK(success, expression)                                                                         \
+#    define ALPAKA_CHECK_DO(file, line, success, expression)                                                          \
         do                                                                                                            \
         {                                                                                                             \
             if(!(expression))                                                                                         \
             {                                                                                                         \
-                acc.cout << "ALPAKA_CHECK failed because '!(" << #expression << ")'\n";                               \
+                acc.cout << "ALPAKA_CHECK failed because '!(" << #expression << ")' in " << file << ":" << line       \
+                         << "\n";                                                                                     \
                 success = false;                                                                                      \
             }                                                                                                         \
         } while(0)
 #else
-#    define ALPAKA_CHECK(success, expression)                                                                         \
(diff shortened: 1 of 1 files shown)


sbastrakov (dismissed):

include/alpaka/test/Check.hpp
```diff
@@ -9,28 +9,46 @@
...
 #endif
+
+#if BOOST_LANG_HIP == BOOST_VERSION_NUMBER(4, 5, 0)
+// disable message print to avoid compiler error: 'error: stack size limit exceeded (181896) in _ZN6alpaka...'
+#    define ALPAKA_CHECK(printMsg, file, line, success, expression)                                                   \
+        do                                                                                                            \
+        {                                                                                                             \
+            if(!(expression))                                                                                         \
+            {                                                                                                         \
+                printf("ALPAKA_CHECK failed because '!(%s)' in %s:%d\n", #expression, file, line);                    \
```
bernhardmgruber, line 48: Do we need special handling for SYCL here? We don't have device side `printf` there.

j-stephan:
What is the state here?

include/alpaka/test/Check.hpp
fwyzard: We do have it now.

include/alpaka/test/Check.hpp
```diff
@@ -9,28 +9,46 @@
...
 #ifdef ALPAKA_ACC_SYCL_ENABLED
-#    define ALPAKA_CHECK(success, expression)                                                                         \
+#    define ALPAKA_CHECK_DO(printMsg, file, line, success, expression)                                                \
         do                                                                                                            \
         {                                                                                                             \
             if(!(expression))                                                                                         \
             {                                                                                                         \
-                acc.cout << "ALPAKA_CHECK failed because '!(" << #expression << ")'\n";                               \
+                acc.cout << "ALPAKA_CHECK failed because '!(" << #expression << ")' in " << file << ":" << line       \
+                         << "\n";                                                                                     \
```
bernhardmgruber, line 25:
```suggestion
                acc.cout << "ALPAKA_CHECK failed because '!(" << #expression << ")' in " << __FILE__<< ":" << __LINE__\
                         << "\n";                                                                                     \
```
You should be able to use the file and line macros directly inside the macro definition and not pass them around.

PR #1771, opened 2022-07-29
HIP: use emulated `atomicAdd(float*,float)`

### LINKED ISSUE #1657: atomicAdd workaround for ROCM
based on the request: https://github.com/alpaka-group/alpaka/issues/1656#issuecomment-1073757306

In PIConGPU we provide an atomicAdd workaround for HIP for `float` atomics which should be atomic between `blocks`. https://github.com/ComputationalRadiationPhysics/picongpu/blob/02f6d46505a55244f0ee384f53b7353126ddec31/include/pmacc/kernel/atomic.hpp#L70-L126.

I was starting to discuss the problem directly in the HIP repository but the discussion never came to a conclusion https://github.com/ROCm-Developer-Tools/HIP/pull/1816

I have not added the workaround into alpaka because it is not fully clear if the performance is only better for the access pattern used in PIConGPU where many threads within a block write randomly to a shared memory data chunk.

Should we add the workaround into alpaka? For ROCM 4.3+ we saw with PIConGPU a very bad shared memory atomic performance which is equal to the performance performing atomics into global memory.

CC-ing: @fwyzard

psychocoderHPC:
fix #1657

Use emulated float atomic for hierarchy threads. In PIConGPU builtin float atomics to shared memory are a factor 3 or more slower.

### DIFF
--- a/include/alpaka/atomic/AtomicUniformCudaHip.hpp
+++ b/include/alpaka/atomic/AtomicUniformCudaHip.hpp
@@ -141,6 +141,16 @@ inline namespace alpakaGlobal
     };
 #        endif
 
+#        if(BOOST_LANG_HIP)
+    // HIP shows bad performance with builtin atomicAdd(float*,float) for the hierarchy threads therefore we do not
+    // call the buildin method and instead use the atomicCAS emulation. For details see:
+    // https://github.com/alpaka-group/alpaka/issues/1657
+    template<>
+    struct AlpakaBuiltInAtomic<alpaka::AtomicAdd, float, alpaka::hierarchy::Threads> : std::false_type
+    {
+    };
+#        endif
+
     // Sub.
 
     template<typename T, typename THierarchy>


include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -141,6 +141,15 @@ inline namespace alpakaGlobal
     };
 #        endif
 
+#        if(BOOST_LANG_HIP)
+    // HIP shows bad performance with builtin atomicAdd(float*,float) for te hierarchy threads therefore we do not call
```
sbastrakov:
```suggestion
    // HIP shows bad performance with builtin atomicAdd(float*,float) for the hierarchy threads therefore we do not call
```

include/alpaka/atomic/AtomicUniformCudaHip.hpp
```diff
@@ -141,6 +141,15 @@ inline namespace alpakaGlobal
...
+    // the buildin method and will use the atomicCAS emulation see: https://github.com/alpaka-group/alpaka/issues/1657
```
sbastrakov:
```suggestion
    // the buildin method and instead use the atomicCAS emulation.
    // For details see: https://github.com/alpaka-group/alpaka/issues/1657
```

psychocoderHPC:
@sbastrakov ready to merge

sbastrakov (approved):

PR #1772, opened 2022-08-03
CI: refactor and extend atomic tests

### LINKED ISSUE #1769: atomic Unit tests for Dec and Inc not check edge cases
The unit test for `atomicDec` is not testing the flip around based on the values.

https://github.com/alpaka-group/alpaka/blob/0b96515b1c8a24d0c3a2904081bc301311503770/include/alpaka/atomic/Op.hpp#L131

Some for `atomicInc`

https://github.com/alpaka-group/alpaka/blob/0b96515b1c8a24d0c3a2904081bc301311503770/include/alpaka/atomic/Op.hpp#L115

We simply test if an value is incremented or decremented:

https://github.com/alpaka-group/alpaka/blob/0b96515b1c8a24d0c3a2904081bc301311503770/test/unit/atomic/src/AtomicTest.cpp#L143-L185

psychocoderHPC:
- [ ] merge after #2582 (first commit contains the fix for atomicDec)

clang CUDA is exposing `atomic*_block()` function signatures even if these can not be used by the selected architecture.
This leads to compile issues if clang is used as CUDA compiler.

Our tests cases have not checked all combinations of atomics, therefore the second commit refactors the atomic tests and fixes #1769:

- test all hierarchies which can be used for atomics
- extent test cases to test for equal, smaller, larger, and zero as left side operand
- test calling `atomic*()` and `atomicOp<*>()`

### DIFF
--- a/test/unit/atomic/src/AtomicFunctors.hpp
+++ b/test/unit/atomic/src/AtomicFunctors.hpp
@@ -1,0 +1,143 @@
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber,
+ * Antonio Di Pilato
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+//! @file This file provides functors where a atomic function and the corresponding operation availble within one
+//! object. This allwos testing of the atomic function and atomicOp<Op>() without the usage of function pointers whcih
+//! is not suppoted by all backends.
+
+#include <alpaka/atomic/Op.hpp>
+#include <alpaka/atomic/Traits.hpp>
+
+#include <utility>
+
+namespace alpaka::test::unit::atomic
+{
+    struct Add
+    {
+        using Op = alpaka::AtomicAdd;
+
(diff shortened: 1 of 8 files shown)


psychocoderHPC:
For an unknown reason, the OpenACC test shows access to invalid memory and HIP requires over 60 minutes to compile the atomic test.

test/unit/atomic/src/AtomicFunctors.hpp
```diff
@@ -0,0 +1,143 @@
...
+#include <utility>
+
+namespace alpaka::test::unit::atomic
+{
+    struct Add
+    {
+        using Op = alpaka::AtomicAdd;
+
+        template<typename... TArgs>
+        static ALPAKA_FN_ACC auto atomic(TArgs&&... args)
```
psychocoderHPC, line 27: @jkelling found out that this variadic template usage is the root of the nvhpc error (invalid memory access)

psychocoderHPC:
The fix for atomics is already merged with https://github.com/alpaka-group/alpaka/pull/1773.
Still open is to extent the test and check if we still run into the issue that compiling tests takes ages.

psychocoderHPC:
Note the reason why this PR was never merged is that old HIP compilers running into stack issues and the compile time was increasing to an unacceptable time.

psychocoderHPC:
I rebased the PR against the latest dev, including the fix in #2582

psychocoderHPC:
I refactored the test gain and hope that HIP is not running into the wall time anymore.
Each operation is now performed within its own kernel, which should simplify the work for the compiler.

PR #1777, opened 2022-08-12
Fix various problems related CUDA and memory allocations

fwyzard:
Base the version checks in `ApiCudaRt` on `CUDART_VERSION` rather than `BOOST_LANG_CUDA`:
  - `BOOST_LANG_CUDA` is defined only when compiling CUDA code, either with nvcc or with clang in CUDA mode;
  - the CUDA runtime API is available also when using a host compiler like gcc or clang in non-CUDA mode; 
  - basing the checks on `CUDART_VERSION` rather than `BOOST_LANG_CUDA` gives consistent results in both scenarios.

Add missing template parameters in a call to `allocBuf(host, extent)`.

### DIFF
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -11,8 +11,7 @@
 
 #ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
 
-#    include <alpaka/core/BoostPredef.hpp>
-
+#    include <boost/predef.h>
 #    include <cuda_runtime_api.h>
 
 namespace alpaka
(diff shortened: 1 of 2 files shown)


fwyzard:
Should I worry about the timed out tests with the HIP backend ?

sbastrakov:
@fwyzard that is an issue our CI currently has, unrelated to your PR. I restarted those jobs.

fwyzard:
@sbastrakov thanks, looks like all tests passed :+1:

sbastrakov (approved):
Looks good to me. Do you also want to have a look @j-stephan @bernhardmgruber  ?

bernhardmgruber (approved):

PR #1778, opened 2022-08-24
Make Philox random engines trivially copyable

sbastrakov:
Manually implemented copy constructor prevented it, while not implementing any non-default logic. This made it impossible to pass it to kernels by value (normally we pass those by pointer e.g. in our examples, so did not notice before).

I checked the `std::is_trivially_copyable_v` before and after the change to confirm it.

cc @jkelling who reported the issue for his branch of PIConGPU.

### DIFF
--- a/test/unit/rand/src/RandTest.cpp
+++ b/test/unit/rand/src/RandTest.cpp
@@ -1,4 +1,5 @@
-/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, René Widera, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, René Widera, Jan Stephan, Bernhard Manfred Gruber,
+ *                Sergei Bastrakov
  *
  * This file is part of alpaka.
  *
@@ -13,6 +14,8 @@
 
 #include <catch2/catch_template_test_macros.hpp>
 #include <catch2/catch_test_macros.hpp>
+
+#include <type_traits>
 
 class RandTestKernel
 {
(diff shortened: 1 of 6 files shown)


bernhardmgruber:
> I checked the std::is_trivially_copyable_v before and after the change to confirm it.

Could you please add this to the unit tests? Just a `STATIC_REQUIRE()` somewhere, thx!

sbastrakov:
@bernhardmgruber I also considered that, but I'm not sure if that should be a guarantee the class gives.

bernhardmgruber:
But then what's the point of changing the behavior to something if we don't intent to guarantee that new behavior? IIUC, @jkelling wants the type to be trivially copyable so we can pass it by value into kernels. That is a valid suggestion and you implemented that. But how can a user now rely on that behavior if we allow ourselves to break it at any moment again?

sbastrakov:
Added a comment that the type has trivially copyable guarantee and a unit test for it.

Previously PR didn't pass because of a reason I do not understand, probably same will happen now.

bernhardmgruber (dismissed):

bernhardmgruber (dismissed):

jkelling:
The CI failure with GCC is caused by a known issue. I put a workaround [here](https://github.com/sbastrakov/alpaka/pull/1).

GCC assumes types with static data members to be not mappable and makes not exception for `constexpr` ones. This is a valid interpretation of the OpenMP <= 4.5 standard. In OpenMP >= 5.0 types with any kind of static data member are mappable.

There is(or was) precedent in alpaka in working around such an issue in one place, tough I do not remember where it was. So we can either the workaroud of replacing `static constexpr` data members by functions, or remove GCC+OpenMP/OpenACC from the CI until they support OpenMP 5.

sbastrakov:
Oh I messed up the history. Will fix everything once the CI passes.

Discussed on VC: Jeffrey will make a separate PR to workaround `static` member with OpenMP 5, and then I will rebase this branch on top of that since the fix is required for the new unit test to work.

test/unit/rand/src/RandTest.cpp
```diff
@@ -95,3 +97,10 @@ TEMPLATE_LIST_TEST_CASE("defaultRandomGeneratorIsWorking", "[rand]", alpaka::tes
 
     REQUIRE(fixture(kernel));
 }
+
+TEMPLATE_LIST_TEST_CASE("defaultRandomGeneratorIsTriviallyCopyable", "[rand]", alpaka::test::TestAccs)
+{
+    using Acc = TestType;
+    using DefaultEngine = decltype(alpaka::rand::engine::createDefault(std::declval<Acc>(), 0u, 0u, 0u));
+    REQUIRE(std::is_trivially_copyable_v<DefaultEngine>);
```
jkelling: Should this not be a `static_assert`? Is it not better to fail as early as possible, i.e. at compile time?

sbastrakov: I am not sure how we approach it generally. I think runtime is better here, but would keep it uniform to how other tests handle it.

jkelling: Just noticed, that @bernhardmgruber suggested `STATIC_REQUIRE`.

bernhardmgruber:
I merged #1780, can you please rebase? Thank you!

test/unit/rand/src/RandTest.cpp
bernhardmgruber: Yes, please use `STATIC_REQUIRE` for compile time "unit tests".

sbastrakov:
Rebase done, fingers crossed for CI

sbastrakov:
In HIP the internal random engines are trivially copyable only since version 5.2.0. This is due to the way ROCRand implemented it before, it was fixed there with https://github.com/ROCmSoftwarePlatform/rocRAND/pull/252 . So for HIP < 5.2.0 and HIP accelerator alpaka's random engines also aren't trivially copyable. I will add it to the comments of the code and tests. The guards were already added and hopefully now CI passes (then i will still need to clean up a little bit in the code and history).

Thanks to @psychocoderHPC for suggestion on how to look into it.

sbastrakov:
@bernhardmgruber CI finally passed, ready for re-review.

sliwowitz (approved):

bernhardmgruber (approved):

Issue #1362, opened 2021-07-27
Memory abstraction tags


SimeonEhrig:
At the moment, I develop a prototype for a lazy evaluated linear algebra library basing on alpaka and vikunja: https://github.com/SimeonEhrig/lazyVikunjaVector

One of the design decision is, that I use mathematical objects like `vector` and `matrix` with implemented operators to do the mathematics.

```C++
Vector<int, 5> v1;
Vector<int, 5> v2;

Vector result = eval(v1 + v2);
```

The vector contains also the data. Therefore, the vector needs to know on which device it is located, e.g. `CPU 0` or `CUDA GPU 1`. This information are used for memory allocation and preventing operations between different device, e.g add a vector located on `CPU 0` with a vector located on `CUDA GPU 1` (it's just a design decision to keep the library "simple").

At the moment I use  `alpaka::AccType<TDim, TIdx>` and the device id, to decide the memory owner of the mathematical object (e.g. a vector) but @psychocoderHPC means that the memory is not bounded to the parallelization strategy. For example memory allocated with `alpaka::AccCpuSerial<Dim, std::size_t>` can be also used in a kernel, which is executed with parallelization strategy `alpaka::AccCpuOmp2Blocks<Dim, std::size_t>`, because both uses the same allocator.

My question is, there are some template tags, like `alpaka::AccCpuSerial` and `alpaka::AccCpuOmp2Blocks` for the memory, e.g. `alpaka::BufCPU` or `alpaka::BufCUDAGPU` and if yes, how I can used it for memory allocation?

This tags would be also pretty helpful for specialization of certain function, e.g.:

```C++
// generic implementation with alpaka 
template<TAcc>
Vector operator+(Vector a, Vector b){
  Vector res;
  AddKernel kernel;
  alpaka::exec<TAcc>(...);
  // ...
  return res;
}

template<>
Vector operator+<alpaka::BufCPU>(Vector a, Vector b){
   // use eigen3 for highly optimized vector addition on CPU 
}

template<>
Vector operator+<alpaka::BufCUDAGPU>(Vector a, Vector b){
   // use cuBlas for highly optimized vector addition on CUDA GPU 
}

j-stephan:
Would this tie in with our efforts on accessors? #1249

I think those have much of the functionality you ask for.

SimeonEhrig:
> Would this tie in with our efforts on accessors? #1249
> 
> I think those have much of the functionality you ask for.

[...]

PR #1781, opened 2022-08-30
add tags for each alpaka accelerator

psychocoderHPC:
We discussed today again in the alpaka meeting that it would be nice to have accelerator tags #1246 so that there is no need to use `#ifdef` to write traits for accelerators.
I had some free minutes and therefore drafted this PR.

- Provide a file `alpaka/acc/Tags.hpp` with tags for each alpaka accelerator.
- Add a concept to uniquely identifies each accelerator with an acc concept
tag.

PIConGPU example to show the possible effect of the tag.

Code snipped without tags.
```C++
template<typename T_Acc = cupla::AccThreadSeq>
struct GetDefaultStrategy
{
    using type = strategy::StridedCachedSupercells;
};

template<typename T_Acc = cupla::AccThreadSeq>
using GetDefaultStrategy_t = typename GetDefaultStrategy<T_Acc>::type;

#if(ALPAKA_ACC_GPU_CUDA_ENABLED == 1)
template<typename... T_Args>
struct GetDefaultStrategy<alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, T_Args...>>
{
    // GPU Utilization is higher compared to `StridedCachedSupercells`
    using type = strategy::CachedSupercells;
};
#endif

#if(ALPAKA_ACC_GPU_HIP_ENABLED == 1)
template<typename... T_Args>
struct GetDefaultStrategy<alpaka::AccGpuUniformCudaHipRt<alpaka::ApiHipRt, T_Args...>>
{
    // GPU Utilization is higher compared to `StridedCachedSupercells`
    using type = strategy::CachedSupercellsScaled<2>;
};
#endif
```

will be changed to without `#ifdef`

```C++
template<typename T_Acc = cupla::AccThreadSeq, typename = void>
struct GetDefaultStrategy
{
    using type = strategy::StridedCachedSupercells;
};

template<typename T_Acc = cupla::AccThreadSeq>
using GetDefaultStrategy_t = typename GetDefaultStrategy<T_Acc>::type;

template<typename T_Acc>
struct GetDefaultStrategy<T_Acc, std::enable_if_t<alpaka::concepts::ImplementsConcept<alpaka::ConceptAccCuda, T_Acc>::value>>
{
    // GPU Utilization is higher compared to `StridedCachedSupercells`
    using type = strategy::CachedSupercells;
};

template<typename T_Acc>
struct GetDefaultStrategy<T_Acc, std::enable_if_t<alpaka::concepts::ImplementsConcept<alpaka::ConceptAccHip, T_Acc>::value>>
{
    // GPU Utilization is higher compared to `StridedCachedSupercells`
    using type = strategy::CachedSupercellsScaled<2>;
};
```

CC-ing: @SimeonEhrig

### DIFF
--- a/include/alpaka/acc/AccCpuFibers.hpp
+++ b/include/alpaka/acc/AccCpuFibers.hpp
@@ -12,6 +12,7 @@
 #ifdef ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLED
 
 // Base classes.
+#    include <alpaka/acc/Tags.hpp>
 #    include <alpaka/atomic/AtomicCpu.hpp>
 #    include <alpaka/atomic/AtomicHierarchy.hpp>
 #    include <alpaka/atomic/AtomicNoOp.hpp>
@@ -78,7 +79,8 @@ namespace alpaka
         public rand::RandStdLib,
         public TimeStdLib,
         public warp::WarpSingleThread,
-        public concepts::Implements<ConceptAcc, AccCpuFibers<TDim, TIdx>>
+        public concepts::Implements<ConceptAcc, AccCpuFibers<TDim, TIdx>>,
+        public concepts::Implements<ConceptAccCpuFibers, AccCpuFibers<TDim, TIdx>>
     {
         static_assert(
             sizeof(TIdx) >= sizeof(int),
--- a/include/alpaka/acc/AccCpuOmp2Blocks.hpp
+++ b/include/alpaka/acc/AccCpuOmp2Blocks.hpp
@@ -16,6 +16,7 @@
 #    endif
 
 // Base classes.
+#    include <alpaka/acc/Tags.hpp>
 #    include <alpaka/atomic/AtomicCpu.hpp>
 #    include <alpaka/atomic/AtomicHierarchy.hpp>
 #    include <alpaka/atomic/AtomicNoOp.hpp>
@@ -79,7 +80,8 @@ namespace alpaka
         public rand::RandStdLib,
         public TimeOmp,
         public warp::WarpSingleThread,
-        public concepts::Implements<ConceptAcc, AccCpuOmp2Blocks<TDim, TIdx>>
+        public concepts::Implements<ConceptAcc, AccCpuOmp2Blocks<TDim, TIdx>>,
+        public concepts::Implements<ConceptAccCpuOmp2Blocks, AccCpuOmp2Blocks<TDim, TIdx>>
     {
         static_assert(
             sizeof(TIdx) >= sizeof(int),
--- a/include/alpaka/acc/AccCpuOmp2Threads.hpp
+++ b/include/alpaka/acc/AccCpuOmp2Threads.hpp
@@ -16,6 +16,7 @@
 #    endif
 
 // Base classes.
+#    include <alpaka/acc/Tags.hpp>
 #    include <alpaka/atomic/AtomicCpu.hpp>
 #    include <alpaka/atomic/AtomicHierarchy.hpp>
 #    include <alpaka/atomic/AtomicOmpBuiltIn.hpp>
@@ -80,7 +81,8 @@ namespace alpaka
         public rand::RandStdLib,
         public TimeOmp,
         public warp::WarpSingleThread,
-        public concepts::Implements<ConceptAcc, AccCpuOmp2Threads<TDim, TIdx>>
+        public concepts::Implements<ConceptAcc, AccCpuOmp2Threads<TDim, TIdx>>,
+        public concepts::Implements<ConceptAccCpuOmp2Threads, AccCpuOmp2Threads<TDim, TIdx>>
     {
         static_assert(
             sizeof(TIdx) >= sizeof(int),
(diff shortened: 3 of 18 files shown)


psychocoderHPC:
example how to get the real acc type out of the tag used in a trait

```C++
template<typename T_Acc>
struct GetDefaultStrategy<T_Acc, std::enable_if_t<alpaka::concepts::ImplementsConcept<alpaka::ConceptAccHip, T_Acc>::value>>
{
    // this will provide alpaka::AccGpuUniformCudaHipRt<alpaka::ApiHipRt, DimType, IdxType>
    using realAccType = typename concepts::ImplementationBase<alpaka::ConceptAccHip, T_Acc>;
};
```

fwyzard:
@psychocoderHPC does it mean that all possible accelerator tags will always be defined, even if the corresponding accelerators and backends are not available ?

bernhardmgruber:
I am not sure the accelerator tags should be provided via the concepts mechanism, since they are not concepts in the C++ sense. E.g. the CUDA and OMP2 accelerator concepts impose the same syntactic requirements. In other words, we cannot turn these into C++20 concepts at some point.

Since tags and accelerators are bijective, I also miss type functions to map between them:
```c++
using Tag = alpaka::TagFor<Acc>;
using Acc = alpaka::AccFor<Tag, Dim, Idx>;
```

Checking whether a type `Acc` is a certain accelerator, instead of:
`alpaka::concepts::ImplementsConcept<alpaka::ConceptAccCuda, Acc>`
would look like this:
`std::is_same_v<alpaka::Tag<Acc>, alpaka::ConceptAccCuda>`

Or multiple:
`mp_contains<mp_list<alpaka::ConceptOmp2, alpaka::ConceptOmp4, alpaka::Threads>, alpaka::Tag<Acc>>`

bernhardmgruber:
> @psychocoderHPC does it mean that all possible accelerator tags will always be defined, even if the corresponding accelerators and backends are not available ?

This is my understanding.

psychocoderHPC:
> @psychocoderHPC does it mean that all possible accelerator tags will always be defined, even if the corresponding accelerators and backends are not available ?

In my draft, the tags will always be available and this tag can be used to check what type a "anonymous" accelerator is.
Each tag has a 1:1 relation to an accelerator. This allows writing type traits with an available sfinae template parameter without the ugly `#ifdef` macro.

psychocoderHPC:
@bernhardmgruber 
> would look like this:
> `std::is_same_v<alpaka::Tag<Acc>, alpaka::ConceptAccCuda>`

Yes a separate trait is possible too. At the end I like to converge to the suggestion https://github.com/alpaka-group/alpaka/issues/944 that we can use the tags to write simple interfaces to allocate devices too (note: I am aware that my simple example in the issue is mixing devices and accelerators) 

```C++
constexpr size_t dim = 1;
auto numDevices = simple:acc::count(simple::acc::GpuCuda);
```

This draft is a discussion base and I am open to move into all direction to find a useful and practical solution to avoid `#ifdef`

SimeonEhrig:
~We need also a functionality to specialize functions with tags instead the acc type like in the kernel specialization example: https://github.com/alpaka-group/alpaka/blob/develop/example/kernelSpecialization/src/kernelSpecialization.cpp~

After a offline discussion, I recognized it is not useful because it only works, if the the function contains generic code and nothing acc specific, like CUDA built in functions in a kernel specialized for the CUDA acc.

SimeonEhrig:
> I am not sure the accelerator tags should be provided via the concepts mechanism, since they are not concepts in the C++ sense. E.g. the CUDA and OMP2 accelerator concepts impose the same syntactic requirements. In other words, we cannot turn these into C++20 concepts at some point.
> 
> Since tags and accelerators are bijective, I also miss type functions to map between them:
> 
> ```c++
> using Tag = alpaka::TagFor<Acc>;
> using Acc = alpaka::AccFor<Tag, Dim, Idx>;
> ```
> 
> Checking whether a type `Acc` is a certain accelerator, instead of: `alpaka::concepts::ImplementsConcept<alpaka::ConceptAccCuda, Acc>` would look like this: `std::is_same_v<alpaka::Tag<Acc>, alpaka::ConceptAccCuda>`
> 
> Or multiple: `mp_contains<mp_list<alpaka::ConceptOmp2, alpaka::ConceptOmp4, alpaka::Threads>, alpaka::Tag<Acc>>`

I already implemented the `using Tag = alpaka::TagFor<Acc>;` trait in my prototype and a constexpr function to compare an acc type with a tag type.

SimeonEhrig:
I implemented a prototype of acc tags in vinkuja: https://github.com/alpaka-group/vikunja/pull/88/commits/565823f87f7710c83786be476f73ee4275115d89

I used vikunja, because I have a use case for the acc tags and also for memory visibility tags (see this issue: https://github.com/alpaka-group/alpaka/issues/1362). I want to use both to improve the API of vikunja. Now, my plan is to integrate the prototype for acc tags (and memory visibility later) in alpaka to make it available for all alpaka users.

SimeonEhrig:
tags are implemented in PR #1804

PR #1782, opened 2022-09-01
Update the functions to allocate pinned/mapped host memory

fwyzard:
Change the interface of `allocMappedBuf()` to identify the platform with an explicit template parameter instead of deducing it from an (unused) device argument.

Add a trait for pinned/mapped memory allocation capability:
  - the constexpr template constant `hasMappedBufSupport` can be used to query if the `DevCpu` host device can allocate pinned/mapped memory accessible by the devices on the given platform;
  - the utility function `allocMappedBufIfSupported` can be used to allocate pinned/mapped memory if supported, otherwise fall back to regular memory.

Update some of the tests that allocate host buffers and copy them to an accelerator to use pinned memory if supported by the underlying platform.

Harmonise the implementation of `allocAsyncBufIfSupported` used in the integration tests, and move it into the alpaka library proper.

Update `docs/source/dev/backends.rst` to document `allocAsyncBuf()` and `allocMappedBuf()`.


fwyzard:
The change to `allocMappedBuf()` simplifies its use, waiving the requirement to create and pass around a device object that in the end was not being used.
The new interface is also clearer, since AFAIK access to the pinned host memory is guaranteed for all devices in a given platform:
  - for host CPU all memory is always accessible (obviously); 
  - for CUDA/HIP we explicitly use the `hostMallocMapped` and `hostMallocPortable` flags;
  - for SYCL/oneAPI (once implemented) `malloc_host` returns memory accessible by all devices in the same context.

A trivial detail: I've put the `TPltf` template argument first because it "sounded" right when calling the function - but if there is a convention for the order of the template arguments do let me know, it should be fine to reorder them.

For OpenMP/OpenACC, I do not know if allocating pinned/mapped host memory makes sense or not.
Any suggestions ?

fwyzard:
I plan to propose a further cleanup of the pinning functions (`pin`/`unpin`, `map`/`unmap`, `prepareForAsyncCopy`) to take a `TPltf` template argument as well.

The reason is that they are currently ambiguous when more than one GPU backend (_e.g._ CUDA and HIP) are enabled at the same time. In fact the current implementation may result in ODR-violations, which may be observed compiling two files for different backends and linking them in a single binary: a symptom of the violation is a segfault when a buffer is pinned by one backend and unpinned by the other.

Depending on the timescale, the status of the tests, and people's preferences, this may come in this PR or in a subsequent one.

fwyzard:
Let's try to re-trigger the tests...

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -124,20 +130,91 @@ namespace alpaka
...
+    template<typename TPltf, typename TElem, typename TIdx, typename TExtent>
+    ALPAKA_FN_HOST auto allocMappedBuf(DevCpu const& host, TExtent const& extent = TExtent())
     {
-        return trait::BufAllocMapped<TElem, Dim<TExtent>, TIdx, TDev>::allocMappedBuf(host, dev, extent);
+        return trait::BufAllocMapped<TPltf, TElem, Dim<TExtent>, TIdx>::allocMappedBuf(host, extent);
+    }
+
+    //! Checks if the host can allocate a pinned/mapped host memory, accessible by all devices in the given platform.
+    //!
+    //! TDev is the type of device to allocate the buffer on.
```
sbastrakov: This is probably a copy-paste leftover from the earlier checker?

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -124,20 +130,91 @@ namespace alpaka
...
+    //! \tparam TPltf The platform from which the buffer is accessible.
+    /* TODO: Remove the following pragmas once support for clang 6 is removed. They are necessary because these
+    /  clang versions incorrectly warn about a missing 'extern'. */
```
sbastrakov, line 180: Could it be a doxygen `\TODO` ?

fwyzard: No idea, I've copy-pasted it from the other places where it's already used.

include/alpaka/mem/buf/Traits.hpp
fwyzard: Ah, yes, thanks for catching it.

include/alpaka/mem/buf/Traits.hpp
fwyzard: Fixed.

fwyzard:
I've taken the opportunity to document the new memory allocation functions while fixing https://github.com/alpaka-group/alpaka/pull/1782#pullrequestreview-1093501684 .

include/alpaka/mem/buf/Traits.hpp
fwyzard: Fixed.

fwyzard:
Could somebody restart the failing test ?

sbastrakov (approved):
I like the changes!

Potentially one could consider a strategy parameter to alloc for a user to define their preference. So that the logic of `allocAsyncBufIfSupported()` is expressed by calling a general function with that parameter. However, I don't see it as a clearly better way.

sbastrakov:
@j-stephan @bernhardmgruber @psychocoderHPC do you want to review as well?

j-stephan (approved):
Thanks for the good work!

PR #1785, opened 2022-09-06
add job generator to gitlab ci

SimeonEhrig:
The new GitLab CI pipeline makes easier to cover all the possible build configurations of alpaka. The pipeline starts with a job generator, which combines the different build parameter, like CMake version, compiler (version), enabled backends and more. But the generator does not generate all possible combination, because the job number would become to big and complete run would take a long time. Instead the job generator use [all-pairs](https://en.wikipedia.org/wiki/All-pairs_testing) testing, keep a good test coverage and reduce the number of test job significantly.

The CI pipeline enables also additional features:
- add custom jobs
- distribute the jobs on stages/waves (typical 10) to enable fair use of the CI resources
- shuffle and reorder Jobs hard coded in the script
- filter and reorder CI jobs via Git commit message

# Current state

This PR contains all stages of the CI pipeline, but not every stage is implemented yet. At the moment, the CI does not generate any job via combination. Instead only custom jobs are executed.

Up coming jobs/PR's:

- create a single job and implement support for the [alpaka-groups-container](https://gitlab.hzdr.de/crp/alpaka-group-container) including the `agc-manager` tool
- add more `gcc` and `clang` jobs
- replace the `CUDA` and `HIP` custom jobs with auto generated jobs
- add more compilers, backends ...

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,67 +1,39 @@
 # stage description
-# Run Time Tests
-#   For HIP or other backends this is distributed round robin within the file.
-#   For CUDA version Major.X (where X is the latest patch of the major version) distributed round robin
-#   for each compiler within the file.
-#   There are exceptions for intermediate minor versions of CUDA
-#   stageRun0 = all CUDA version Major.0
-#   stageRun1 = all CUDA where minor != 0 and not latest minor version
-# Compile Tests
-#   stageCompileN = round robin distributed within the file
+# generate
+#  creates the actual test jobs via python script
+# run-tests
+#  execute the generated test jobs
+#  the ordering is mainly random
+#  the python generator can schedule some jobs on specific positions
 
 stages:
-  - stageCompile0
-  - stageRun0
-  - stageCompile1
-  - stageRun1
+  - generator
+  - run-test-jobs
 
 
 variables:
-  ALPAKA_GITLAB_CI_CONTAINER_VERSION: "1.4"
-  ALPAKA_CI_OS_NAME: "Linux"
-  alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE: "ON"
-  alpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE: "ON"
-  alpaka_ACC_CPU_B_SEQ_T_FIBERS_ENABLE: "OFF"
-  alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE: "OFF"
-  alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE: "ON"
-  alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE: "ON"
-  alpaka_ACC_ANY_BT_OMP5_ENABLE: "OFF"
-  alpaka_ACC_ANY_BT_OACC_ENABLE: "OFF"
-  alpaka_ACC_GPU_CUDA_ENABLE: "OFF"
-  alpaka_ACC_GPU_CUDA_ONLY_MODE: "OFF"
-  alpaka_ACC_GPU_HIP_ENABLE: "OFF"
-  alpaka_ACC_GPU_HIP_ONLY_MODE: "OFF"
-  # If ALPAKA_CI_ANALYSIS is OFF compile and execute runtime tests else compile only.
-  ALPAKA_CI_ANALYSIS: "OFF"
-  ALPAKA_CI_RUN_TESTS: "ON"
-  alpaka_CI: GITLAB
-  # needs to be enabled, that test on the GPU are executed
-  ALPAKA_FORCE_RUNTIME_TEST: "ON"
-  ALPAKA_CI_SANITIZERS: ""
-  ALPAKA_CI_INSTALL_CUDA: "OFF"
-  ALPAKA_CI_INSTALL_HIP: "OFF"
-  ALPAKA_CI_CMAKE_DIR: "$HOME/cmake"
-  BOOST_ROOT: "$HOME/boost"
-  ALPAKA_CI_BOOST_LIB_DIR: "$HOME/boost_libs"
-  ALPAKA_CI_CUDA_DIR: "$HOME/cuda"
-  ALPAKA_CI_HIP_ROOT_DIR: "$HOME/hip"
+  # container version of the generated jobs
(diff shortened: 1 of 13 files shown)


SimeonEhrig:
More information about the CI pipeline can be found in the sphinx doc: https://alpaka--1785.org.readthedocs.build/en/1785/dev/ci.html

.gitlab-ci.yml
```diff
@@ -1,67 +1,39 @@
...
-  - local: '/script/gitlabci/job_cuda11.6.yml'
-  - local: '/script/gitlabci/job_hip4.2.yml'
-  - local: '/script/gitlabci/job_hip4.3.yml'
-  - local: '/script/gitlabci/job_hip4.5.yml'
-  - local: '/script/gitlabci/job_hip5.0.yml'
-  - local: '/script/gitlabci/job_hip5.1.yml'
-  - local: '/script/gitlabci/job_hip5.2.yml'
+generate:
+  stage: generator
+  image: alpine:latest
```
bernhardmgruber, line 22: AFAIK most of the CI runs on Ubuntu. Why do we need alpine for this job? Could this run on Ubuntu as well?

docs/source/dev/ci.rst
```diff
@@ -13,9 +13,119 @@ GitLab CI
...
    :alt: Relationship between GitHub.com, GitLab.com and HZDR gitlab-ci runners
 
    Relationship between GitHub.com, GitLab.com and HZDR gitlab-ci runners
 
-To change how the tests are built and executed, modify the code in the `alpaka repository <https://github.com/alpaka-group/alpaka>`_. If the container environment with the dependencies needs to be changed, please open an issue or contribute to `alpaka-group-container repository <https://gitlab.hzdr.de/crp/alpaka-group-container>`_.
+The Container Registry
+++++++++++++++++++++++
+
+Alpaka uses containers in which as many dependencies as possible are already installed to save job execution time. The available containers can be found `here <https://gitlab.hzdr.de/crp/alpaka-group-container/container_registry>`_. Each container provides a tool called ``agc-manager`` to check if a software is installed. The documentation for ``agc-manager`` can be found `here <https://gitlab.hzdr.de/crp/alpaka-group-container/-/tree/master/tools>`_. A common way to check if a software is already installed is to use an ``if else statement``. If a software is not installed yet, you can install it every time at job runtime.
```
bernhardmgruber: Personally, I prefer to keep lines shorter also in documentation files. For example, I add a newline after each sentence.

.gitlab-ci.yml
SimeonEhrig: Because the alpine image is much smaller, than the ubuntu image. The job only generates a yaml file for the following up jobs. And the only difference in the job script compare to use an ubuntu image is, that we use `apk` instead `apt`.

bernhardmgruber: ok then!

psychocoderHPC: Is a fixed alpine version maybe better to avoid that if the image is switching to a new python version that our script stops working?

docs/source/dev/ci.rst
psychocoderHPC: Yes, this will make later diffs smaller and better readable.

psychocoderHPC:
@SimeonEhrig could you please point to the line in the diff where we now handle if a job is in a run stage (where we perform runtime tests) and where it is compiled only?

This was before the change using CPU runner only to compile only without runtime tests.

```
.base_hip_compile_only:
  extends: .basecfg_hip
  variables:
    ALPAKA_CI_RUN_TESTS: "OFF"
  tags:
    - cpuonly
    - x86_64
```

script/job_generator/generate_job_yaml.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns:
+        Dict[str, str]: Dict of {variable name : variable value}.
+    """
+    variables: Dict[str, str] = {}
+
+    return variables
+
+
+@typechecked
+def job_tags(job: Dict[str, Tuple[str, str]]) -> List[str]:
```
psychocoderHPC, line 97: Could it be that this function is currently not used? How do we like to decide in the future when the generator is creating the jobs which job must be compiled only and where we would like to execute compile only. 
Can this become a new list like the compiler list? With `perform test` and `execute only`?

.gitlab-ci.yml
SimeonEhrig: Better I fixed the Python version itself.

docs/source/dev/ci.rst
SimeonEhrig: Okay, I will change it.

script/job_generator/generate_job_yaml.py
SimeonEhrig: The function is not used at the moment, because no job is auto generated.
If the job is only compiling or also the tests executed is a point, which we need to discuss. There are two options.
1. Create a new parameter field for the job generator. It will be more or less randomly decided, if a job executes the tests or not.
2. Implement a logic in the `job_tags()` function or somewhere else, which decides which job needs to execute the tests.

SimeonEhrig:
> @SimeonEhrig could you please point to the line in the diff where we now handle if a job is in a run stage (where we perform runtime tests) and where it is compiled only?
> 
> This was before the change using CPU runner only to compile only without runtime tests.
> 
> ```
> .base_hip_compile_only:
>   extends: .basecfg_hip
>   variables:
>     ALPAKA_CI_RUN_TESTS: "OFF"
>   tags:
>     - cpuonly
>     - x86_64
> ```

Here is the global value set: https://github.com/alpaka-group/alpaka/blob/224af1e330ab9830fc5b56afe1319b83565f93da/script/job_generator/generate_job_yaml.py#L143
For specific jobs, the value is overwritten by the specific job yaml like before.

script/job_generator/generate_job_yaml.py
SimeonEhrig: In a offline discussion, we decided to implement a logic, which decides which job is compile only and which is running the tests. I added a comment in the `job_generator.py`, where the logic should be executed.

SimeonEhrig:
@psychocoderHPC I draw a flow graph for the job generator for the documentation. Can you please give feedback.

https://alpaka--1785.org.readthedocs.build/en/1785/dev/ci.html#the-job-generator

psychocoderHPC (approved):

PR #1788, opened 2022-09-14
Support std::experimental::mdspan

### LINKED ISSUE #1482: std::mdspan
~~`std::mdspan` in its [14th revision](https://wg21.link/p0009r14) passed the LEWG review last Monday, meaning its design has converged and seems to be finished for now. The paper still has to go through several votes, but I expect it to pass them, so it has a good chance of making it into C++23.~~
`std::mdspan` was voted into C++23 at the Kona meeting last November. It's coming.

I would really like to see it in alpaka as well, maybe even with some glue code, like having a `getMdspan(view)` function in addition to `getNativePointer(view)`. That would elegantly solve all the multidim buffer issues, pitches etc. `std::mdspan` supports various memory layouts, accessors (different from alpaka's current `Accessor`), slices (subspans), and works with SYCL (I asked). It is not as powerful as LLAMA, but provides a nice middle ground for many use cases.

Now I am wondering how we could include it. Kokkos has the [reference implementation](https://github.com/kokkos/mdspan), which is licensed under BSD-3. The implementation has 5KLOCs. We could include it similar to TinyMT (also BSD-3) or STLTuple (Apache). We could also have it as an optional third-party dependency and enable alpaka functionality depending on whether the user has mdspan.

bernhardmgruber:
This PR adds the reference implementation of `std::experimental::mdspan` from Kokkos to alpaka and adds a new API `alpaka::experimental::getMdSpan(TView&)`/`alpaka::experimental::getMdSpanTransposed(TView&)`, which allows to build an mdspan from any view.

Currently, there is only one memory layout supported which is hard coded inside alpaka to support arbitrary pitched allocations, which mdspan does not support with its standard layouts. ~I assume further mdspan functions sensitive to the layout (e.g. slicing) will not work.~ Submdspan works as well.

The additional dependency can be enabled via the cmake option `alpaka_USE_MDSPAN`, which is `OFF` by default. It can be set to `SYSTEM` to look for `mdspan` installed in the system, or to `FETCH` to acquire the Kokkos reference implementation via `FetchContent`.

When mdspan is enabled, it will *not* be installed with alpaka. However, when using an installed version of alpaka, you can get mdspan again via `FetchContent` or from a system installation. The behavior for getting `mdspan` is thus the same whether you use alpaka via `add_subdirectory` or from a system installation.

Fixes: #1482

Open questions:

- [x] ~Are we fine with shipping `std::experimental::mdspan` as part of alpaka?~ We will leave it out to avoid difficulties with installing it alongside alpaka.
- [x] ~If yes, how/where should we ship it? As part of the `include/alpaka` tree? It needs to be installed with alpaka.~
- [x] ~Should we offer a combined API to allocate a buffer and create the mdspan in one go? Should such a call for an ND buffer/mdspan return a ND buffer/mdspan or 1D buffer/ND mdspan. The latter is required for the standard layouts.~ Let's skip this for now.
- [x] ~When we want to have a normal mdspan with the standard layouts, we can only create it from 1D buffers/views. What should we do? Only allow custom layouts in `getMdSpan` for 1D views and have a hard coded pitched layout for ND views?~  Let's skip this for now and be happy that we can get an mdspan, even if we cannot play with layouts yet.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -909,7 +909,13 @@ elseif (alpaka_USE_MDSPAN STREQUAL "FETCH")
         GIT_REPOSITORY https://github.com/kokkos/mdspan.git
         GIT_TAG 973ef6415a6396e5f0a55cb4c99afd1d1d541681
     )
-    FetchContent_MakeAvailable(mdspan)
+    # we don't use FetchContent_MakeAvailable(mdspan) since it would also install mdspan
+    # see also: https://stackoverflow.com/questions/65527126/how-to-disable-installation-a-fetchcontent-dependency
+    FetchContent_GetProperties(mdspan)
+    if(NOT mdspan_POPULATED)
+        FetchContent_Populate(mdspan)
+        add_subdirectory(${mdspan_SOURCE_DIR} ${mdspan_BINARY_DIR} EXCLUDE_FROM_ALL)
+    endif()
     target_link_libraries(alpaka INTERFACE std::mdspan)
     target_compile_definitions(alpaka INTERFACE ALPAKA_USE_MDSPAN)
 elseif (alpaka_USE_MDSPAN STREQUAL "OFF")
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -891,3 +891,52 @@ if(alpaka_COMPILER_OPTIONS_DEVICE OR alpaka_COMPILER_OPTIONS_DEVICE)
     endif()
     message("")
 endif()
+
+#-------------------------------------------------------------------------------
+# Include mdspan
+
+set(alpaka_USE_MDSPAN "OFF" CACHE STRING "Use std::mdspan with alpaka")
+set_property(CACHE alpaka_USE_MDSPAN PROPERTY STRINGS "SYSTEM;FETCH;OFF")
+
+if (alpaka_USE_MDSPAN STREQUAL "SYSTEM")
+    find_package(mdspan REQUIRED)
+    target_link_libraries(alpaka INTERFACE std::mdspan)
+    target_compile_definitions(alpaka INTERFACE ALPAKA_USE_MDSPAN)
+elseif (alpaka_USE_MDSPAN STREQUAL "FETCH")
+    include(FetchContent)
+    FetchContent_Declare(
+        mdspan
+        GIT_REPOSITORY https://github.com/kokkos/mdspan.git
+        GIT_TAG 973ef6415a6396e5f0a55cb4c99afd1d1d541681
+    )
+    FetchContent_MakeAvailable(mdspan)
+    target_link_libraries(alpaka INTERFACE std::mdspan)
+    target_compile_definitions(alpaka INTERFACE ALPAKA_USE_MDSPAN)
+elseif (alpaka_USE_MDSPAN STREQUAL "OFF")
+else()
+    message(FATAL_ERROR "Invalid option for alpaka_USE_MDSPAN")
+endif()
+
+if (NOT alpaka_USE_MDSPAN STREQUAL "OFF")
+    if (MSVC AND (alpaka_CXX_STANDARD LESS 20))
+        message(WARNING "std::mdspan on MSVC requires C++20. Please enable C++20 via alpaka_CXX_STANDARD. Use of std::mdspan has been disabled.")
+        set(alpaka_USE_MDSPAN "OFF" CACHE STRING "Use std::mdspan with alpaka" FORCE)
+    endif ()
+
+    if (alpaka_ACC_GPU_CUDA_ENABLE AND (CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA") AND (CMAKE_CXX_COMPILER_ID STREQUAL "Clang"))
+        # this issue actually only occurs when the host compiler (not the CXX compiler) is clang, but cmake does not let us query the host compiler id
+        # see: https://gitlab.kitware.com/cmake/cmake/-/issues/20901
+        message(WARNING "std::mdspan does not work with nvcc and clang as host compiler. Use of std::mdspan has been disabled.")
+        set(alpaka_USE_MDSPAN "OFF" CACHE STRING "Use std::mdspan with alpaka" FORCE)
+    endif ()
+
+    if ((alpaka_ACC_ANY_BT_OMP5_ENABLE OR alpaka_ACC_ANY_BT_OACC_ENABLE))
+        message(WARNING "std::mdspan is not yet compatible with alpaka's OpenMP5 and OpenAcc backends. Use of std::mdspan has been disabled.")
+        set(alpaka_USE_MDSPAN "OFF" CACHE STRING "Use std::mdspan with alpaka" FORCE)
+    endif ()
+
+    if (alpaka_ACC_GPU_CUDA_ENABLE AND (NOT alpaka_CUDA_EXPT_EXTENDED_LAMBDA STREQUAL ON))
+        message(WARNING "std::mdspan requires nvcc's extended lambdas. Use of std::mdspan has been disabled.")
+        set(alpaka_USE_MDSPAN "OFF" CACHE STRING "Use std::mdspan with alpaka" FORCE)
+    endif()
+endif()
--- a/docs/source/advanced/cmake.rst
+++ b/docs/source/advanced/cmake.rst
@@ -78,6 +78,15 @@ alpaka_DEBUG_OFFLOAD_ASSUME_HOST
 
      Allow host-only contructs like assert in offload code in debug mode.
 
+alpaka_USE_MDSPAN
+  .. code-block::
+
+     Enable/Disable the use of `std::experimental::mdspan`:
+
+     "OFF" - Disable mdspan
+     "SYSTEM" - Enable mdspan and acquire it via `find_package` from your system
+     "FETCH" - Enable mdspan and download it via CMake's `FetchContent` from GitHub. The dependency will not be installed when you install alpaka.
+
 .. _cpu-serial:
 
 CPU Serial
--- a/example/bufferCopy/CMakeLists.txt
+++ b/example/bufferCopy/CMakeLists.txt
@@ -45,6 +45,11 @@ if(NOT TARGET alpaka::alpaka)
     endif()
 endif()
 
+if (alpaka_USE_MDSPAN STREQUAL "OFF")
+    message(STATUS "The bufferCopy example requires mdspan. Please set alpaka_USE_MDSPAN accordingly. Example disabled.")
+    return()
+endif ()
+
 #-------------------------------------------------------------------------------
 # Add executable.
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -50,6 +50,7 @@ concurrency:
 # alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE             : {ON, OFF}
 # alpaka_ACC_GPU_HIP_ENABLE                     : {ON, OFF}
 #   [ON] ALPAKA_CI_HIP_BRANCH                   : {rocm-4.2}
+# alpaka_USE_MDSPAN                             : {OFF, FETCH, SYSTEM}
 
 # if you add/remove any CI variable, also update docker_ci.sh
 env:
@@ -76,6 +77,7 @@ env:
   alpaka_ACC_GPU_CUDA_ONLY_MODE: OFF
   alpaka_ACC_GPU_HIP_ENABLE: OFF
   alpaka_ACC_GPU_HIP_ONLY_MODE: OFF
+  alpaka_USE_MDSPAN: OFF
   CMAKE_INSTALL_PREFIX: ${{ github.workspace }}/_install
 
 jobs:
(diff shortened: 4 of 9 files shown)


bernhardmgruber:
I also tried to prototype a single API call `allocBufAndGetMdSpan` that basically allocates a buffer and creates the mdspan in one go, returning both:

```c++
    template<
        typename TElem,
        typename TIdx,
        typename LayoutPolicy = std::experimental::layout_right,
        typename AccessorPolicy = std::experimental::default_accessor<TElem>,
        typename TDev,
        typename TExtent,
        typename... MdSpanCtorArgs>
    ALPAKA_FN_HOST auto allocBufAndMdSpan(
        TDev const& dev,
        TExtent const& extent = TExtent(),
        MdSpanCtorArgs&&... mdSpanCtorArgs)
    {
        auto buf = allocBuf(dev, Vec{getExtentProduct(extent)});
        auto extents = traits::internal::getStdExtents(extent);
        auto mdspan = std::experimental::mdspan<TElem, decltype(extents), LayoutPolicy, AccessorPolicy>{
            getPtrNative(buf),
            extents, // TODO
            std::forward<MdSpanCtorArgs>(mdSpanCtorArgs)...};
        return std::tuple{std::move(buf), std::move(mdspan)};
    }
```

However, there are far more ways to allocate buffers (and views!) in alpaka than I thought. So I don't want to duplicate all API calls. Also, the constructor overload set of mdspan is not very uniform. I can pass pointer and extent, but if the mapping is more complex, I need to pass pointer and mapping, but with the mapping instance being constructed with the extents. So I also cannot implement `allocBufAndMdSpan` generically for all layout policies.

Also notice that when you allocate an ND buffer, you will get a 1D buffer and an ND mdspan. This is needed to support mdspan's standard layouts.

fwyzard:
I'm not very familiar with `mdspan` yet, so this may not make too much sense...

From what I understood from a couple of examples, `mdspan` does not own the memory, but provides N-dimensional array access to an underlying buffer.

Then, rather than trying to get it together with the buffer, wouldn't it make more sense to have a free function that takes an existing buffer as an argument and returns an `mdspan` over it ?

bernhardmgruber:
> From what I understood from a couple of examples, `mdspan` does not own the memory, but provides N-dimensional array access to an underlying buffer.

Yes.

> Then, rather than trying to get it together with the buffer, wouldn't it make more sense to have a free function that takes an existing buffer as an argument and returns an `mdspan` over it ?

Yes, this is what `alpaka::getMdSpan(TView&)` is and I have a prototype implementation for it. The motivation for a combined API was to always create a 1D buffer, even when ND was requested. However, this may be obsolete with the recent solution discussed here: https://github.com/kokkos/mdspan/issues/117#issuecomment-1250185735. For me, having a dedicated free function to create the mdspan from an existing view is definitely preferable.

bernhardmgruber:
I had to take the non-single header version of mdspan now, because the amalgamated one straight includes `<span>`, which is C++20.

bernhardmgruber:
@SimeonEhrig since you asked about the progress here, I think it is useable enough now to be played with.

SimeonEhrig:
> @SimeonEhrig since you asked about the progress here, I think it is useable enough now to be played with.

WTF, you are like Scotty :joy:

Yesterday 
I: I like to have mdspan in the next release.
You: I'm pretty busy, maybe I have some time in December.
I: Okay, than we move it to the next release.
You: Maybe you can try it.

Today
You: Okay, I did it this night.

bernhardmgruber:
Haha, I appreciate the compliment! This PR still needs a lot of work though to push it through the CI and make it compile on all combinations. There are likely changes necessary in `mdspan` as well to make it work on MSVC. I don't have time to address all that soon, so I stick with what I said :) Still, I though yesterday that maybe I can get it working in a small test and example on at least a newer compiler with some hacks. And that's what I pushed here.

bernhardmgruber:
> If yes, how/where should we ship mdspan ? As part of the include/alpaka tree? It needs to be installed with alpaka.

Discussed during the alpaka VC. We ship mdspan as subtree placed inside the thirdparty directory. In cmake, we add a switch whether to use the internal or system mdspan (since you can install mdspan via package managers). We also need to add the appropriate install commands to cmake to include the internal mdspan.

fwyzard:
Is it possible to make `mdspan` an optional dependency, _i.e._ to make the Alpaka headers not use it if it's not available at all (neither from the local copy nor as an external package) ?

bernhardmgruber:
> Is it possible to make `mdspan` an optional dependency, _i.e._ to make the Alpaka headers not use it if it's not available at all (neither from the local copy nor as an external package) ?

Technically, it is certainly possible. Probably not even very difficult, since we just need to guard the corresponding include and trait.
But what would be the motivation for not having it?

bernhardmgruber:
So there is a bit of a problem with having mdspan reside in the thirdparty directory. I need to put the check where to look for mdspan in `alpakaCommon.cmake`. It has three options: `INTERNAL` (uses `add_subdirectory` and then `target_link_library`), `SYSTEM` (used `find_package` and then `target_link_library`) and `OFF`. However, depending on whether alpaka is installed or not, the option `INTERNAL` is not available. How should I deal with this situation? It is kind of annoying that I would need to differentiate between installed and non-installed alpaka.

psychocoderHPC:
Maybe I get your question wrong.
If alpaka is installed it should install all third-party libs too and still provide the option `INTERNAL` and `SYSTEM`. 
In case the user is installing alpaka with `SYSTEM` third-party should not be installed and using `INTERNAL` will fail.
That we need to deal with the case that that alpaka is installed without third-party and we need to handle this separate is IMO not possible to avoid because CMake allows the direct usage of codes with `add_subdirectory` and as installed liberay.

A good source for such problems is [OpenPMD-api](https://github.com/openPMD/openPMD-api), @ax3l invested already a lot of his time into handling such cases.

cmake/alpakaCommon.cmake
```diff
@@ -343,6 +343,23 @@ if(NOT alpaka_HAS_STD_MATH_CONSTANTS)
...
+
+set(alpaka_USE_MDSPAN "INTERNAL" CACHE STRING "Use std::mdspan with alpaka")
+set_property(CACHE alpaka_USE_MDSPAN PROPERTY STRINGS "SYSTEM;INTERNAL;OFF")
+
+if (alpaka_USE_MDSPAN STREQUAL "SYSTEM")
+    find_package(mdspan REQUIRED)
+    target_link_libraries(alpaka INTERFACE std::mdspan)
+    target_compile_definitions(alpaka INTERFACE ALPAKA_USE_MDSPAN)
+endif()
+if (alpaka_USE_MDSPAN STREQUAL "INTERNAL")
```
psychocoderHPC: IMO this should be an `elseif()` and the case `OFF` should be handled too.
All other cases should be watched with `else()` to help the user to find typos e.g. `cmake -Dalpaka_USE_MDSPAN=INTERNAl`

thirdParty/mdspan/examples/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
...
+
+project(example LANGUAGES CXX)
+
+if(NOT TARGET std::mdspan)
+  find_package(mdspan REQUIRED)
+endif()
+
+function(mdspan_add_example EXENAME)
+  add_executable(${EXENAME} ${EXENAME}.cpp)
+  target_link_libraries(${EXENAME} std::mdspan)
```
psychocoderHPC: `find_package` and `target_link_libraries` for `mdspan` should not be required because alapka should handle it.

psychocoderHPC (changes requested):

SimeonEhrig:
> So there is a bit of a problem with having mdspan reside in the thirdparty directory. I need to put the check where to look for mdspan in `alpakaCommon.cmake`. It has three options: `INTERNAL` (uses `add_subdirectory` and then `target_link_library`), `SYSTEM` (used `find_package` and then `target_link_library`) and `OFF`. However, depending on whether alpaka is installed or not, the option `INTERNAL` is not available. How should I deal with this situation? It is kind of annoying that I would need to differentiate between installed and non-installed alpaka.

In the worst case, you have to handle the mdspan library two times in CMake. Depending on how you include alpaka in your project, different cmake files are used. If you use the `add_subdirectory()` method, the `CMakeLists.txt` file in the root folder is used. If you install alpaka and use `find_package()` method, the `alpakaCommon.cmake` is used. There are some cmake helper functions, which "copies" some content from the `CMakeLists.txt` to the `alpakaCommon.cmake`. In this case, it looks like, that you care about some functionality only one time.

cmake/alpakaCommon.cmake
bernhardmgruber: Done.

thirdParty/mdspan/examples/CMakeLists.txt
bernhardmgruber: These are examples from the mdspan repository, not my code :)

cmake/alpakaCommon.cmake
```diff
@@ -328,6 +328,23 @@ if(NOT alpaka_HAS_STD_MATH_CONSTANTS)
...
+    find_package(mdspan REQUIRED)
+    target_link_libraries(alpaka INTERFACE std::mdspan)
+    target_compile_definitions(alpaka INTERFACE ALPAKA_USE_MDSPAN)
+elseif (alpaka_USE_MDSPAN STREQUAL "INTERNAL")
+    # this is handled in thirdParty/CMakeLists.txt
+elseif (alpaka_USE_MDSPAN STREQUAL "OFF")
+else ()
+    message(FATAL_ERROR "Invalid option for alpaka_USE_MDSPAN")
+endif()
+
```
SimeonEhrig: You should add the compile definition here, because `thirdparty/CMakeLists.txt` is not installed. So if you install alpaka and use it in another project, the compiler definition `ALPAKA_USE_MDSPAN` is not set.

```cmake
+if (NOT alpaka_USE_MDSPAN STREQUAL "OFF")
+    target_compile_definitions(alpaka INTERFACE ALPAKA_USE_MDSPAN)
+endif()
```

bernhardmgruber: But if alpaka is installed, `alpaka_USE_MDSPAN` can only be `SYSTEM;OFF`, and those cases are handled here. What problem does your suggestion solve?

bernhardmgruber:
So, it turns out, with C++20 enabled, MSVC can also compile `std::mdspan`. What should we do? Require C++20 on Windows? Or disable `std::mdspan` on Windows?

bernhardmgruber:
I started to add flags to the CI to enable/disable mdspan for certain runs, but this is touching a lot of places now. It currently seems OpenACC, OpenMP5, and MSVC with C++17 are making troubles. We could also try and detect these in cmake and disable mdspan accordingly. What is the preferred approach?

cmake/alpakaCommon.cmake
SimeonEhrig: No, you can also install mdspan separately and then set `alpaka_USE_MDSPAN=ON` that you can use the external installation.

SimeonEhrig:
> So, it turns out, with C++20 enabled, MSVC can also compile `std::mdspan`. What should we do? Require C++20 on Windows? Or disable `std::mdspan` on Windows?

Only allows to enable, if the `alpaka_CXX_STANDARD` is set to 20. Otherwise throw a warning or error.

SimeonEhrig:
> I started to add flags to the CI to enable/disable mdspan for certain runs, but this is touching a lot of places now. It currently seems OpenACC, OpenMP5, and MSVC with C++17 are making troubles. We could also try and detect these in cmake and disable mdspan accordingly. What is the preferred approach?

I think, that makes sense. Also document, what is supported and why we do not support all backends (mdspan support is a optional feature).

bernhardmgruber:
We discussed this PR in our alpaka VC today and concluded that:
* We should add all mdspan related functionality into `alpaka::experimental`
* Reexpose `std::experimental::mdspan` as `alpaka::experimental::mdspan`, so we can select between different mdspan implementations.
* Do not ship mdspan as part of alpaka, because it is too difficult to handle all the cases (alpaka usage via `add_subdirectory` and after installation). This also avoids the question whether we should install mdspan with alpaka. As replacement, we will install mdspan now via cmake's `FetchContent` when request, which works the same independently how alpaka is installed/used.

example/bufferCopy/CMakeLists.txt
```diff
@@ -45,6 +45,11 @@ if(NOT TARGET alpaka::alpaka)
     endif()
 endif()
 
+if (alpaka_USE_MDSPAN STREQUAL "OFF")
```
SimeonEhrig, line 48: I would prefer to not including the `CMakeLists.txt` of the `bufferCopy` example, if mdspan is not enabled:

`example/CMakeLists.txt`:
```cmake
if (NOT alpaka_USE_MDSPAN STREQUAL "OFF")
  add_subdirectory("bufferCopy/")
endif()
```

I think a warning should be fixable for your current build and enabling an optional feature to fix the warning sounds not like a good solution.
Also my IDE is constantly display all warnings and want's that I fix it ;-)

bernhardmgruber: If I do that, you will get a compile error then when you compile the bufferCopy example standalone, and not from the alpaka build tree. That is, you run cmake inside `example/bufferCopy`. I don't think this is better.

bernhardmgruber:
I just had a VC with @SimeonEhrig and we agreed on the following remaining tasks:

* Add the new cmake option for enabling `mdspan` to the documentation
* Do not install `mdspan` with alpaka, when enabled via `FetchContent`
* Use of `mdspan` should be disabled by default. We should enable it for at least 1 CI run.
* Convert the cmake `WARNING` to `INFO` when `mdspan` is not enabled for the buffercopy example

docs/source/advanced/cmake.rst
```diff
@@ -78,6 +78,15 @@ alpaka_DEBUG_OFFLOAD_ASSUME_HOST
...
      Allow host-only contructs like assert in offload code in debug mode.
 
+alpaka_USE_MDSPAN
+  .. code-block::
+
+     Enable/Disable the use of `std::experimental::mdspan`:
+
+     "OFF" - Disable mdspan
+     "SYSTEM" - Enable mdspan and acquire it via `find_package` from your system
+     "FETCH" - Enable mdspan and download it via CMake's `FetchContent` from GitHub
```
SimeonEhrig: Can you please note, that mdspan will not be installed

SimeonEhrig:
Do you want to keep this for some documentation purpose or this simply a forgotten artifact?

https://github.com/alpaka-group/alpaka/blob/d6a1aad8d0363e562e101f32512f44eea0fb77c3/example/bufferCopy/src/bufferCopy.cpp#L25-L31

SimeonEhrig:
I tested the PR together with vikunja. The CMake works well. Both options (`SYSTEM` and `FETCH`) works and there is nothing, which is accidentally installed.

SimeonEhrig (dismissed):

j-stephan:
Please rebase!

bernhardmgruber:
> Do you want to keep this for some documentation purpose or this simply a forgotten artifact?
> 
> https://github.com/alpaka-group/alpaka/blob/d6a1aad8d0363e562e101f32512f44eea0fb77c3/example/bufferCopy/src/bufferCopy.cpp#L25-L31

That should no longer be necessary. Removed.

docs/source/advanced/cmake.rst
bernhardmgruber: Done.

SimeonEhrig (approved):

psychocoderHPC (approved):

j-stephan (approved):

PR #1792, opened 2022-09-16
Refactor Philox to expose Counter-based RNG API

jkelling:
This PR refactors alpaka's Philox RNG implementation to expose the counter-based core RNG to the user without overhead.

## Rationale

The current alpaka RNG API is inspired by `<random>` and thereby follows the pattern of classical sequence-based RNGs. The Philox RNG is a counter-based RNG, i.e. it actually is a block cipher that encrypts a value (sequence index/counter) to produce a random number. Thereby, the counter-based RNG can generate any random number from its range for a given seed (key) arbitrarily without following any sequence.

The current API exposes a classical sequence RNG implemented based on the counter-based one by maintaining a counter for each thread while using the grid-thread-index as an offset (value in another word of the global counter) to select independent sequences. The API forces the user to keep an instance of counter state in order to use the RNG. This API limits applications in two ways:
1. There is an overhead for maintaining the counter-state in situations where the index is naturally generated from the simulation step and element each time a random number is needed.
   1. The counter is bundled with the globally-shared key. In many setups having only one global instance of the key/seed, instead of a copy per thread should be more efficient.
2. Relying on the counter will, for the same seed, produce different random numbers depending on `workDiv`, which prohibits reproducibility across platforms. A way to get reproducible runs is to derive the counter/index from the simulation step and element (e.g. time-step and updated cell or particle, this must be some parameter of the simulation continuously progressing, not based on the simulation state.)

## Solution in this PR

1. Split the Philox class hierarchy into the bare counter-based-RNG implementation and the current API and add a new user-facing generator wit default params `PhiloxStateless4x32x10Vector`.
2. Functions related to the core implementation are not using any data members of the Philox classes, hence they are made static.
3. Add a cast operator from `CudaVectorArrayWrapper` to `std::array` in order to allow for seamless transfer of keys and counter offsets between different backends.
4. Add example `counterBasedRng` which fill a multi-dimensional arrays with random numbers in a way yielding the identical results for any backend. This for illustration and constitutes a test by comparing results from one active parallel backend to some form the sequential backend.

## Review

* Please have a look at the proposed API in this PR. I am using this implementation in a current project and thus would appreciate if changes were not too breaking...
* @j-stephan @psychocoderHPC @bernhardmgruber

### DIFF
--- a/example/counterBasedRng/src/counterBasedRng.cpp
+++ b/example/counterBasedRng/src/counterBasedRng.cpp
@@ -1,0 +1,258 @@
+/* Copyright 2022 Jeffrey Kelling
+ *
+ * This file exemplifies usage of alpaka.
+ *
+ * Permission to use, copy, modify, and/or distribute this software for any
+ * purpose with or without fee is hereby granted, provided that the above
+ * copyright notice and this permission notice appear in all copies.
+ *
+ * THE SOFTWARE IS PROVIDED “AS IS” AND ISC DISCLAIMS ALL WARRANTIES WITH
+ * REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+ * MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
+ * SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+ * WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+ * ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
+ * IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+ */
+
+#include <alpaka/alpaka.hpp>
+#include <alpaka/example/ExampleDefaultAcc.hpp>
+#include <alpaka/rand/RandPhiloxStateless.hpp>
+
+#include <chrono>
+#include <iostream>
+#include <random>
+#include <typeinfo>
+
+//! A kernel that fills an array with pseudo-random numbers.
+class CounterBasedRngKernel
+{
+public:
+    template<class TAcc>
+    using Vec = alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>>;
+    template<class TAcc>
+    using Gen = typename alpaka::rand::PhiloxStateless4x32x10Vector<TAcc>;
+    template<class TAcc>
+    using Key = typename Gen<TAcc>::Key;
+    template<class TAcc>
+    using Counter = typename Gen<TAcc>::Counter;
+
+private:
+    template<unsigned int I>
+    struct ElemLoop
+    {
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename TAcc, typename TElem>
+        static ALPAKA_FN_ACC auto elemLoop(
+            TAcc const& acc,
+            alpaka::experimental::
+                BufferAccessor<TAcc, TElem, alpaka::Dim<TAcc>::value, alpaka::experimental::WriteAccess> dst,
+            Key<TAcc> const& key,
+            Vec<TAcc> const& threadElemExtent,
+            Vec<TAcc>& threadFirstElemIdx) -> void
+        {
+            auto const threadLastElemIdx = threadFirstElemIdx[I] + threadElemExtent[I];
+            auto const threadLastElemIdxClipped
+                = (dst.extents[I] > threadLastElemIdx) ? threadLastElemIdx : dst.extents[I];
+
+            constexpr auto Dim = alpaka::Dim<TAcc>::value;
+
+            const auto firstElem = threadFirstElemIdx[I];
+            if constexpr(I < Dim - 1)
+            {
+                for(; threadFirstElemIdx[I] < threadLastElemIdxClipped; ++threadFirstElemIdx[I])
+                {
+                    ElemLoop<I + 1>::elemLoop(acc, dst, key, threadElemExtent, threadFirstElemIdx);
+                }
+            }
+            else
+            {
+                Counter<TAcc> c = {0, 0, 0, 0};
+                for(unsigned int i = 0; i < Dim; ++i)
+                    c[i] = threadFirstElemIdx[i];
+
+                for(; threadFirstElemIdx[Dim - 1] < threadLastElemIdxClipped; ++threadFirstElemIdx[Dim - 1])
+                {
+                    c[Dim - 1] = threadFirstElemIdx[Dim - 1];
+                    const auto random = Gen<TAcc>::generate(c, key);
+                    // to make use of the whole random vector we would need to ensure numElement[0] % 4 == 0
+                    dst[threadFirstElemIdx] = TElem(random[0]);
+                }
+            }
+            threadFirstElemIdx[I] = firstElem;
+        }
+    };
+
+public:
+    //! The kernel entry point.
+    //!
+    //! \tparam TAcc The accelerator environment to be executed on.
+    //! \tparam TElem The matrix element type.
+    //! \param acc The accelerator to be executed on.
+    //! \param dst destimation matrix.
+    //! \param extent The matrix dimension in elements.
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TAcc, typename TElem>
+    ALPAKA_FN_ACC auto operator()(
+        TAcc const& acc,
+        alpaka::experimental::BufferAccessor<TAcc, TElem, alpaka::Dim<TAcc>::value, alpaka::experimental::WriteAccess>
+            dst,
+        Key<TAcc> const& key) const -> void
+    {
+        constexpr auto Dim = alpaka::Dim<TAcc>::value;
+        static_assert(Dim <= 4, "The CounterBasedRngKernel expects at most 4-dimensional indices!");
+
+        Vec<TAcc> const gridThreadIdx(alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc));
+        Vec<TAcc> const threadElemExtent(alpaka::getWorkDiv<alpaka::Thread, alpaka::Elems>(acc));
+        Vec<TAcc> threadFirstElemIdx(gridThreadIdx * threadElemExtent);
+
+        ElemLoop<0>::elemLoop(acc, dst, key, threadElemExtent, threadFirstElemIdx);
+    }
+};
+
+auto main() -> int
+{
+// Fallback for the CI with disabled sequential backend
+#if defined(ALPAKA_CI) && !defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+    return EXIT_SUCCESS;
+#else
+
+    // Define the index domain
+    using Dim = alpaka::DimInt<3u>;
+    using Idx = std::size_t;
+
+    // Define the accelerator
+    //
+    // It is possible to choose from a set of accelerators:
+    // - AccGpuCudaRt
+    // - AccGpuHipRt
+    // - AccCpuThreads
+    // - AccCpuFibers
+    // - AccCpuOmp2Threads
+    // - AccCpuOmp2Blocks
+    // - AccOmp5
+    // - AccCpuTbbBlocks
+    // - AccCpuSerial
+    // using Acc = alpaka::AccCpuSerial<Dim, Idx>;
+    using Acc = alpaka::ExampleDefaultAcc<Dim, Idx>;
+    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<Acc>() << std::endl;
+
+    using AccHost = alpaka::AccCpuSerial<Dim, Idx>;
+    // Get the host device for allocating memory on the host.
+    using DevHost = alpaka::DevCpu;
+
+    // Defines the synchronization behavior of a queue
+    //
+    // choose between Blocking and NonBlocking
+    using QueueProperty = alpaka::Blocking;
+    using QueueAcc = alpaka::Queue<Acc, QueueProperty>;
+    using QueueHost = alpaka::Queue<AccHost, QueueProperty>;
+
+    // Select a device
+    auto const devAcc = alpaka::getDevByIdx<Acc>(0u);
+    auto const devHost = alpaka::getDevByIdx<AccHost>(0u);
+
+    // Create a queue on the device
+    QueueAcc queueAcc(devAcc);
+    QueueHost queueHost(devHost);
+
+    // Define the work division
+    alpaka::Vec<Dim, Idx> const extent = {16, 16, 16 * 8};
+    alpaka::Vec<Dim, Idx> const elementsPerThread = {1, 1, 1};
+    alpaka::Vec<Dim, Idx> const elementsPerThreadHost = {1, 1, 8};
+
+    // Let alpaka calculate good block and grid sizes given our full problem extent
+    alpaka::WorkDivMembers<Dim, Idx> const workDivAcc(alpaka::getValidWorkDiv<Acc>(
+        devAcc,
+        extent,
+        elementsPerThread,
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted));
+    alpaka::WorkDivMembers<Dim, Idx> const workDivHost(alpaka::getValidWorkDiv<AccHost>(
+        devHost,
+        extent,
+        elementsPerThreadHost,
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted));
+
+    // Define the buffer element type
+    using Data = std::uint32_t;
+
+    // Allocate 3 host memory buffers
+    using BufHost = alpaka::Buf<DevHost, Data, Dim, Idx>;
+    BufHost bufHost(alpaka::allocBuf<Data, Idx>(devHost, extent));
+    BufHost bufHostDev(alpaka::allocBuf<Data, Idx>(devHost, extent));
+
+    // Initialize the host input vectors A and B
+    Data* const pBufHost(alpaka::getPtrNative(bufHost));
+    Data* const pBufHostDev(alpaka::getPtrNative(bufHostDev));
+
+    std::random_device rd{};
+    CounterBasedRngKernel::Key<AccHost> key = {rd(), rd()};
+
+    // Allocate buffer on the accelerator
+    using BufAcc = alpaka::Buf<Acc, Data, Dim, Idx>;
+    BufAcc bufAcc(alpaka::allocBuf<Data, Idx>(devAcc, extent));
+
+    // Create the kernel execution task.
+    auto const taskKernelAcc = alpaka::createTaskKernel<Acc>(
+        workDivAcc,
+        CounterBasedRngKernel(),
+        alpaka::experimental::writeAccess(bufAcc),
+        key);
+    auto const taskKernelHost = alpaka::createTaskKernel<AccHost>(
+        workDivHost,
+        CounterBasedRngKernel(),
+        alpaka::experimental::writeAccess(bufHost),
+        key);
+
+    // Enqueue the kernel execution task
+    alpaka::enqueue(queueHost, taskKernelHost);
+    alpaka::enqueue(queueAcc, taskKernelAcc);
+
+    // Copy the result from the device
+    alpaka::memcpy(queueAcc, bufHostDev, bufAcc);
+    const auto numElements = extent.prod();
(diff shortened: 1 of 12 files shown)


include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
```diff
@@ -154,27 +93,5 @@ namespace alpaka::rand::engine
...
-            Key key{key_in};
-            Counter counter = singleRound(counter_in, key);
-
-            // TODO: Consider unrolling the loop for performance
-            for(unsigned int n = 0; n < numRounds(); ++n)
-            {
-                key = bumpKey(key);
-                counter = singleRound(counter, key);
-            }
-            // TODO: Should the key be returned as well??? i.e. should the original key be bumped?
```
jkelling, line 175: This code moved, but this comment was also removed, so I feel the need to answer the question posed: The `bumpKey` function constitutes the key schedule of the block cipher, that is for each round a new key is derived from the actual main key. The key schedule does not carry over from block to block, otherwise we would have a stream cipher. The only use in having the round key would be adding additional rounds of encryption, but this would be done by adjusting `numRounds()`, so there is no conceivable reason to return `key`.

jkelling:
Believe @sliwowitz implemented the alpaka Philox, so maybe you should have a look.

include/alpaka/rand/Philox/PhiloxBaseCommon.hpp
sliwowitz: Good, thanks for the explanation. I went by the reference implementation, but I've seen some implementations did bump&return the key at each iteration - I thought those were wring, but wasn't sure about it so I left the comment here as a reminder I wanted to investigate more about it.

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -139,6 +139,27 @@ namespace alpaka::meta
...
+            static_assert(std::tuple_size<Other>::value == size, "Can only convert between vectors of same size.");
+            static_assert(
+                std::is_same<typename Other::value_type, value_type>::value,
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+            this->y = o[1];
+            this->z = o[2];
+            this->w = o[3];
+        }
+        operator std::array<value_type, size>() const
```
bernhardmgruber:
```suggestion
        ALPAKA_FN_HOST_ACC constexpr operator std::array<value_type, size>() const
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -139,6 +139,27 @@ namespace alpaka::meta
...
+        {
+            std::array<value_type, size> ret;
+            ret[0] = this->x;
+            ret[1] = this->y;
+            ret[2] = this->z;
+            ret[3] = this->w;
+        }
```
bernhardmgruber:
```suggestion
            ret[3] = this->w;
            return ret;
        }
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -139,6 +139,27 @@ namespace alpaka::meta
             this->z = *it++;
             this->w = *it++;
         }
+        template<class Other>
+        ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE CudaVectorArrayWrapper(const Other& o)
+        {
+            static_assert(std::tuple_size<Other>::value == size, "Can only convert between vectors of same size.");
```
bernhardmgruber:
```suggestion
            static_assert(std::tuple_size_v<Other> == size, "Can only convert between vectors of same size.");
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -139,6 +139,27 @@ namespace alpaka::meta
...
+            static_assert(
+                std::is_same<typename Other::value_type, value_type>::value,
```
bernhardmgruber:
```suggestion
                std::is_same_v<typename Other::value_type, value_type>,
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -164,6 +185,25 @@ namespace alpaka::meta
...
+            this->y = o[1];
+            this->z = o[2];
+        }
+        operator std::array<value_type, size>() const
+        {
+            std::array<value_type, size> ret;
+            ret[0] = this->x;
+            ret[1] = this->y;
+            ret[2] = this->z;
+        }
```
bernhardmgruber: Same changes as pointed out above.

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -188,6 +228,23 @@ namespace alpaka::meta
...
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+            this->y = o[1];
+        }
+        operator std::array<value_type, size>() const
+        {
+            std::array<value_type, size> ret;
+            ret[0] = this->x;
+            ret[1] = this->y;
+        }
```
bernhardmgruber: Same changes as pointed out above.

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -211,6 +268,21 @@ namespace alpaka::meta
...
+                std::is_same<typename Other::value_type, value_type>::value,
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+        }
+        operator std::array<value_type, size>() const
+        {
+            std::array<value_type, size> ret;
+            ret[0] = this->x;
+        }
+
```
bernhardmgruber: Same changes as pointed out above.

example/counterBasedRng/src/counterBasedRng.cpp
```diff
@@ -0,0 +1,258 @@
...
+                    const auto random = Gen<TAcc>::generate(c, key);
+                    // to make use of the whole random vector we would need to ensure numElement[0] % 4 == 0
+                    dst[threadFirstElemIdx] = TElem(random[0]);
+                }
+            }
+            threadFirstElemIdx[I] = firstElem;
+        }
+    };
+
+public:
```
bernhardmgruber, line 86: Is this really code a user should write? If yes, then we should try to move some parts into alpaka and provide building blocks. If no, we should try to provide an easier example.

example/counterBasedRng/src/counterBasedRng.cpp
```diff
@@ -0,0 +1,258 @@
...
+        ElemLoop<0>::elemLoop(acc, dst, key, threadElemExtent, threadFirstElemIdx);
+    }
+};
+
+auto main() -> int
+{
+// Fallback for the CI with disabled sequential backend
+#if defined(ALPAKA_CI) && !defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+    return EXIT_SUCCESS;
+#else
```
bernhardmgruber, line 118: We cannot build/run this example in the CI with other backends? why?

example/counterBasedRng/src/counterBasedRng.cpp
```diff
@@ -0,0 +1,258 @@
...
+    //     alpaka::createView(devHost, pBufHostDev, alpaka::Vec<alpaka::DimInt<1u>, Idx>(numElements)),
+    //     alpaka::createView(devAcc, alpaka::getPtrNative(bufAcc),
+    //         alpaka::Vec<alpaka::DimInt<1u>, Idx>(numElements)));
+
+    // wait in case we are using an asynchronous queue to time actual kernel runtime
+    alpaka::wait(queueHost);
+    alpaka::wait(queueAcc);
+
+    int falseResults = 0;
+    const int MAX_PRINT_FALSE_RESULTS = extent[2] * 2;
```
bernhardmgruber: Please use camel or snake case for local variables.

bernhardmgruber (changes requested):
Is the example enough for a test? I feel like we should have a unit test as well for the philox, or some of its guts.

example/counterBasedRng/src/counterBasedRng.cpp
psychocoderHPC: I think this is the problem we have with all examples. The example is using the hard-coded serial accelerator, therefore this acc is required.

example/counterBasedRng/src/counterBasedRng.cpp
```diff
@@ -0,0 +1,258 @@
...
+    alpaka::enqueue(queueHost, taskKernelHost);
+    alpaka::enqueue(queueAcc, taskKernelAcc);
+
+    // Copy the result from the device
+    alpaka::memcpy(queueAcc, bufHostDev, bufAcc);
+    const auto numElements = extent.prod();
+    // alpaka::memcpy(queueAcc,
+    //     alpaka::createView(devHost, pBufHostDev, alpaka::Vec<alpaka::DimInt<1u>, Idx>(numElements)),
+    //     alpaka::createView(devAcc, alpaka::getPtrNative(bufAcc),
+    //         alpaka::Vec<alpaka::DimInt<1u>, Idx>(numElements)));
```
bernhardmgruber: Can this be removed?

include/alpaka/rand/Philox/PhiloxStateless.hpp
```diff
@@ -0,0 +1,129 @@
...
+            return TParams::width;
+        }
+
+        static_assert(numRounds() > 0, "Number of Philox rounds must be > 0.");
+        static_assert(vectorSize() % 2 == 0, "Philox counter size must be an even number.");
+        static_assert(vectorSize() <= 16, "Philox SP network is not specified for sizes > 16.");
+        static_assert(numberWidth() % 8 == 0, "Philox number width in bits must be a multiple of 8.");
+
+        // static_assert(TWidth == 32 || TWidth == 64, "Philox implemented only for 32 and 64 bit numbers.");
+        static_assert(numberWidth() == 32, "Philox implemented only for 32 bit numbers.");
```
bernhardmgruber: Is 32bit only a new restriction? Did the previous implementation support 64bit? If no, can we delete the comment?

include/alpaka/rand/Philox/PhiloxStateless.hpp
```diff
@@ -0,0 +1,129 @@
...
+         * @param key_in initial state of the key
+         * @return result of the PRNG shuffle; has the same size as the counter
+         */
+        static ALPAKA_FN_HOST_ACC auto nRounds(Counter const& counter_in, Key const& key_in) -> Counter
+        {
+            Key key{key_in};
+            Counter counter = singleRound(counter_in, key);
+
+            // TODO: Consider unrolling the loop for performance
+            for(unsigned int n = 0; n < numRounds(); ++n)
```
bernhardmgruber:
```suggestion
            ALPAKA_UNROLL(numRounds())
            for(unsigned int n = 0; n < numRounds(); ++n)
```

include/alpaka/rand/Philox/PhiloxStatelessKeyedBase.hpp
```diff
@@ -0,0 +1,42 @@
...
+    template<typename TBackend, typename TParams>
+    struct PhiloxStatelessKeyedBase : public PhiloxStateless<TBackend, TParams>
+    {
+    public:
+        using Counter = typename PhiloxStateless<TBackend, TParams>::Counter;
+        using Key = typename PhiloxStateless<TBackend, TParams>::Key;
+
+        const Key m_key;
+
+        PhiloxStatelessKeyedBase(Key&& key) : m_key(key)
```
bernhardmgruber: Why the r-value reference? Are you missing a move?
```suggestion
        PhiloxStatelessKeyedBase(Key&& key) : m_key(std::move(key))
```

bernhardmgruber (changes requested):

example/counterBasedRng/src/counterBasedRng.cpp
jkelling: Yes and No: This is code the user should write only if they want to fill an array with random number, but this is not a common use case. Usually, the user will generate the random number and use them in the same place. Writing it into an array is only a stand-in for mapping random numbers to extrinsic properties of your problem, here this is coordinates.

I do not see part of the RNG-related logic, that could be turned into generic building blocks here.

The only that could be improved would be to have an `operator=` to replace lines 70-72.

example/counterBasedRng/src/counterBasedRng.cpp
jkelling: Indeed.

jkelling:
```suggestion
```

include/alpaka/rand/Philox/PhiloxStateless.hpp
jkelling: The comment was there in the original code, and was only moved. @sliwowitz what do you think?

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -211,6 +268,21 @@ namespace alpaka::meta
...
+            static_assert(
+                std::is_same<typename Other::value_type, value_type>::value,
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+        }
+        operator std::array<value_type, size>() const
+        {
+            std::array<value_type, size> ret;
+            ret[0] = this->x;
+        }
```
jkelling:
```suggestion
            ret[0] = this->x;
            return ret;
        }
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
jkelling:
```suggestion
            ret[1] = this->y;
            return ret;
        }
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
jkelling:
```suggestion
            ret[2] = this->z;
            return ret;
        }
```

example/counterBasedRng/src/counterBasedRng.cpp
```diff
@@ -0,0 +1,258 @@
...
+
+    if(falseResults == 0)
+    {
+        std::cout << "Execution results correct!" << std::endl;
+        return EXIT_SUCCESS;
+    }
+    else
+    {
+        std::cout << "Found " << falseResults << " false results, printed no more than " << MAX_PRINT_FALSE_RESULTS
+                  << "\n"
```
jkelling:
```suggestion
    const int maxPrintFalseResults = extent[2] * 2;

    auto aHost = alpaka::experimental::readAccess(bufHost);
    auto aAcc = alpaka::experimental::readAccess(bufHostDev);
    for(Idx z = 0; z < aHost.extents[0]; ++z)
        for(Idx y = 0; y < aHost.extents[1]; ++y)
            for(Idx x = 0; x < aHost.extents[2]; ++x)
            {
                Data const& valHost(aHost(z, y, x));
                Data const& valAcc(aAcc(z, y, x));
                if(valHost != valAcc)
                {
                    if(falseResults < maxPrintFalseResults)
                        std::cerr << "host[" << z << ", " << y << ", " << x << "] = " << valHost << " != acc[" << z
                                  << ", " << y << ", " << x << "] = " << valAcc << std::endl;
                    ++falseResults;
                }
            }

    if(falseResults == 0)
    {
        std::cout << "Execution results correct!" << std::endl;
        return EXIT_SUCCESS;
    }
    else
    {
        std::cout << "Found " << falseResults << " false results, printed no more than " << maxPrintFalseResults
                  << "\n"
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -164,6 +185,25 @@ namespace alpaka::meta
...
+        {
+            static_assert(std::tuple_size<Other>::value == size, "Can only convert between vectors of same size.");
+            static_assert(
+                std::is_same<typename Other::value_type, value_type>::value,
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+            this->y = o[1];
+            this->z = o[2];
+        }
+        operator std::array<value_type, size>() const
```
jkelling:
```suggestion
        ALPAKA_FN_HOST_ACC constexpr operator std::array<value_type, size>() const
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -188,6 +228,23 @@ namespace alpaka::meta
...
+        ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE CudaVectorArrayWrapper(const Other& o)
+        {
+            static_assert(std::tuple_size<Other>::value == size, "Can only convert between vectors of same size.");
+            static_assert(
+                std::is_same<typename Other::value_type, value_type>::value,
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+            this->y = o[1];
+        }
+        operator std::array<value_type, size>() const
```
jkelling:
```suggestion
        ALPAKA_FN_HOST_ACC constexpr operator std::array<value_type, size>() const
```

include/alpaka/meta/CudaVectorArrayWrapper.hpp
```diff
@@ -211,6 +268,21 @@ namespace alpaka::meta
...
+        template<class Other>
+        ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE CudaVectorArrayWrapper(const Other& o)
+        {
+            static_assert(std::tuple_size<Other>::value == size, "Can only convert between vectors of same size.");
+            static_assert(
+                std::is_same<typename Other::value_type, value_type>::value,
+                "Can only convert between vectors of same element type.");
+            this->x = o[0];
+        }
+        operator std::array<value_type, size>() const
```
jkelling:
```suggestion
        ALPAKA_FN_HOST_ACC constexpr operator std::array<value_type, size>() const
```

include/alpaka/rand/Philox/PhiloxStateless.hpp
```diff
@@ -0,0 +1,129 @@
...
+         * @param counter_in initial state of the counter
+         * @param key_in initial state of the key
+         * @return result of the PRNG shuffle; has the same size as the counter
+         */
+        static ALPAKA_FN_HOST_ACC auto nRounds(Counter const& counter_in, Key const& key_in) -> Counter
+        {
+            Key key{key_in};
+            Counter counter = singleRound(counter_in, key);
+
+            ALPAKA_UNROLL(numRounds())
```
jkelling, line 107:
```suggestion
            ALPAKA_UNROLL(numRounds());
```

jkelling:
> Is the example enough for a test? I feel like we should have a unit test as well for the philox, or some of its guts.

There is a test for RNGs `test/unit/rand/`, I think they use the default RNG for each backend and Philox is the default at least for OpemMP5 and OpenACC. I do not know what exactly it tests for, though. There is also `example/randomStrategies/`, which is hard-coded to use Philox.

example/counterBasedRng/src/counterBasedRng.cpp
jkelling: It is expected to be run with another backend enabled, but the serial one is to serve as the host reference, and thus must also be enabled. Technically, any CPU backend could play this role, but I do not know of a utility we have to select any CPU enabled backend (and fail if none are enabled). `alpaka::ExampleDefaultAcc<Dim, Idx>;` is used as the other, the device, backend.

bernhardmgruber: Ok, got it!

bernhardmgruber:
I will wait for @sliwowitz to review and approve now.

include/alpaka/rand/Philox/PhiloxStateless.hpp
sliwowitz: I haven't got around to implementing the 64 bit version - the relevant constants are present, but right now I'm not sure what else would have to be done to make it work. I think we can delete the comment.

jkelling:
@sliwowitz ping?

sliwowitz (dismissed):

sliwowitz:
> @sliwowitz ping?

I like it. 

BTW, tangentially related to this PR - there's been some discussion about moving the RNGs to a separate library,  we might want to revisit the idea in some upcoming meting.

bernhardmgruber (dismissed):

bernhardmgruber:
Can you please rebase your branch on the latest `develop`? Thx!

jkelling:
>  haven't got around to implementing the 64 bit version - the relevant constants are present, but right now I'm not sure what else would have to be done to make it work. I think we can delete the comment.

Done.

jkelling:
> Can you please rebase your branch on the latest `develop`? Thx!

Done. Waiting for CI.

bernhardmgruber (approved):

PR #1795, opened 2022-09-26
Make mangled CUDA kernel name as short as possible

bernhardmgruber:
This PR aims to make the mangled name of CUDA kernels launched by alpaka as short as possible and places the name of the kernel functor at the front of the template argument list, so it is more easily found in a profiler.

Example:
image

### DIFF
--- a/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
@@ -67,39 +67,38 @@
 
 namespace alpaka
 {
-    namespace uniform_cuda_hip
+    namespace detail
     {
-        namespace detail
+        //! The GPU CUDA/HIP kernel entry point.
+        // \NOTE: 'A __global__ function or function template cannot have a trailing return type.'
+        // We have put the function into a shallow namespace and gave it a short name, so the mangled name in the
+        // profiler (e.g. ncu) is as shorter as possible.
+        template<typename TKernelFnObj, typename TApi, typename TAcc, typename TDim, typename TIdx, typename... TArgs>
+        __global__ void gpuKernel(
+            Vec<TDim, TIdx> const threadElemExtent,
+            TKernelFnObj const kernelFnObj,
+            TArgs... args)
         {
-            //! The GPU CUDA/HIP kernel entry point.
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
I welcome suggestions for a better name and place instead of `alpaka::detail::kernel`, but please aim for making it short!

j-stephan:
I'm a bit afraid that this will collide with similar changes for the other back-ends. What happens when we put the OpenMP kernels into `alpaka::detail`, too?

bernhardmgruber:
> I'm a bit afraid that this will collide with similar changes for the other back-ends. What happens when we put the OpenMP kernels into `alpaka::detail`, too?

Well, you will get an error in the unit tests then. And then you will either have to choose a new name for the OpenMP entity, or rename the CUDA/HIP kernel entry point. But we can do all of that since we are in the detail namespace.

Do you have a better suggestion for a short mangled name?

j-stephan:
I do not. However, I believe my fears are unfounded. The alpaka kernel object that is passed to the actual kernel in `alpaka::detail` should have the chosen Accelerator as part of its template instantiation. So even if we pass a kernel to both OpenMP and CUDA the instantiated objects should still be different.

psychocoderHPC:
What do you think about calling the CUDA/HIP kernel `alpaka::detail::gpuKernel` to avoid naming collision and later we could name OMP kernel `alpaka::detail::omp2BlockKernel`, `alpaka::detail::omp2ThreadKernel`, `alpaka::detail::serialKernel` `alpaka::detail::omp5Kernel`, ... if needed.

fwyzard:
What about `alpaka::kernel::gpu`, `alpaka::kernel::omp2Block`, _etc._ ?

fwyzard:
Though `gpu` is maybe a bit too generic, assuming the SYCL/oneAPI kernels use a different entry point ?

bernhardmgruber:
Ok, I will rename it to `alpaka::detail::gpuKernel`.

> What about alpaka::kernel::gpu, alpaka::kernel::omp2Block, etc. ?

These names would work as well. However, I prefer using the detail namespace as it also intentionally communicates that we do not promise the name `gpuKernel` or the existence of a namespace `kernel`.

psychocoderHPC (approved):

PR #1801, opened 2022-09-28
Add a few explicit tests on pitches

bernhardmgruber:
Depends on #1610.

### DIFF
--- a/test/unit/mem/buf/src/PitchTest.cpp
+++ b/test/unit/mem/buf/src/PitchTest.cpp
@@ -1,0 +1,74 @@
+/* Copyright 2022 Bernhard Manfred Gruber
+ *
+ * This file is part of alpaka.
+ *
+ * This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ */
+
+#include <alpaka/mem/buf/Traits.hpp>
+#include <alpaka/test/acc/TestAccs.hpp>
+
+#include <boost/mp11.hpp>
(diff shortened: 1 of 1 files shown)


test/unit/mem/buf/src/PitchTest.cpp
```diff
@@ -0,0 +1,75 @@
...
+    using Idx = int;
+    using Acc = alpaka::AccGpuCudaRt<alpaka::DimInt<0>, Idx>;
+
+    const auto dev = alpaka::getDevByIdx<alpaka::Pltf<alpaka::Dev<Acc>>>(0);
+
+    checkBufferPitches(dev, alpaka::Vec<alpaka::DimInt<0>, Idx>{}, alpaka::Vec<alpaka::DimInt<0>, Idx>{});
+
+    checkBufferPitches(dev, alpaka::Vec{10}, alpaka::Vec{40});
+
+    constexpr auto pitch = 512;
```
bernhardmgruber, line 66: AFAIK, NVIDIA does not guarantee a row pitch anywhere. It is just internal magic of `cudaMallocPitch`. So this test may fail on some hardware at some point. I guess I should query the row pitch. Should we maybe add an alpaka function for that?

bernhardmgruber: Actually, we cannot really do this either, because all we could do is call `cudaMallocPitch`, since that may even use a different pitch for each call. It may not even be dependent on the arguments.

psychocoderHPC: We should rethink how to write this test. The pitch should not be set explicitly because this value is given implicitly.

bernhardmgruber:
PR #2093 will simplify how pitches work and a few similar tests. That should make this PR obsolete.

PR #1803, opened 2022-09-28
Upgrade clang/CUDA headercheck CI to clang 13 and CUDA 11.2

### LINKED ISSUE #1802: Upgrade  linux_clang-9_cuda-9.2_debug_analysis CI check
Because it emits this wrong diagnostic:
```
/home/runner/work/alpaka/alpaka/example/bufferCopy/src/bufferCopy.cpp:64:29: error: reference to __host__ function '<deduction guide for Vec><unsigned long, unsigned long, unsigned long>' in __host__ __device__ function
                            alpaka::Vec{z, y, x},
                            ^
/home/runner/work/alpaka/alpaka/include/alpaka/vec/Vec.hpp:440:5: note: '<deduction guide for Vec><unsigned long, unsigned long, unsigned long>' declared here
    Vec(TFirstIndex, TRestIndices...) -> Vec<DimInt<1 + sizeof...(TRestIndices)>, TFirstIndex>;
    ^
```
which is technically correct, since deduction guides are evaluated via a synthesized function template, but still wrong and not useful, since this function template would never be codegen-ed.

I propose to move to a newer clang and CUDA version.

bernhardmgruber:
Fixes: #1802


bernhardmgruber:
This PR requires to change the required CI checks on Github.

j-stephan:
Why did you choose CUDA 11.2?

bernhardmgruber:
> Why did you choose CUDA 11.2?

Because I chose the newest clang we support in alpaka (13) and then the newest CUDA that is supported by that clang version according to Axel's table: https://gist.github.com/ax3l/9489132.

But the choice should not matter much, since it just runs the header check.

j-stephan (approved):

PR #1804, opened 2022-09-28
Implement Acc tags

### LINKED ISSUE #1246: Introduce accelerator tags

SimeonEhrig:
Implement accelerator tags, which allows to specialize code for specific accelerators without the requirements, that the acc needs to be enabled or guarded by a preprocessor `if`.
The code inside the specialized regions needs to be compileable independent of the enabled back-end,  otherwise the prepocessor `if` is still required. For example CUDA built-in functions needs a preprocessor guard, because without CUDA compiler, the function definitions are not available.

TODOs:

- [x] test kernel specialization
- [x] make `acc_tag_match` variadic
- [x] add example


example/tagSpecialization/src/tagSpecialization.cpp
```diff
@@ -0,0 +1,136 @@
...
+std::string host_function()
+{
+    return "generic host function";
+}
+
+template<>
+std::string host_function<alpaka::AccOaccTag>()
+{
+    return "OpenAcc host function";
+}
```
bernhardmgruber: I tend to avoid full-specialization for function templates, because I have been told so by me mentor years ago. I would rather use a parameter and overloading:
```c++
template<typename Tag>
std::string host_function(Tag)
{
    return "generic host function";
}

std::string host_function(alpaka::AccOaccTag)
{
    return "OpenAcc host function";
}
```
Maybe you can add this as an additional example!

example/tagSpecialization/src/tagSpecialization.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+auto main() -> int
+{
+// Fallback for the CI with disabled sequential backend
+#if defined(ALPAKA_CI) && !defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+    return EXIT_SUCCESS;
+#else
```
bernhardmgruber, line 97: You don't need that since you don't depend on the serial backend.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,64 @@
...
+    CREATE_ACC_TAG(AccGpuHipRtTag);
+    CREATE_ACC_TAG(AccGpuSyclIntelTag);
+    CREATE_ACC_TAG(AccOaccTag);
+    CREATE_ACC_TAG(AccOmp5Tag);
+
+    template<typename TAcc>
+    struct AccToTag;
+
+    template<typename TAcc>
+    using AccToTag_t = typename AccToTag<TAcc>::type;
```
bernhardmgruber: Do we have precedence for naming an alias `_t` in alpaka?

I would rather call it `AccToTag` and rename the `struct AccToTag`, because I want to have the user facing API nicer, like:
```c++
 host_function<alpaka::AccToTag<Acc>>()
```

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    template<typename TTag, typename TDim, typename TIdx>
+    using TagToAcc_t = typename TagToAcc<TTag, TDim, TIdx>::type;
+
+    // TODO: make variadic, to support more than on tag per call
+    template<typename TAcc, typename TTag>
+    using is_acc_tag_match = std::is_same<AccToTag_t<TAcc>, TTag>;
+
+    template<typename TAcc, typename TTag>
+    inline constexpr bool is_acc_tag_match_v = alpaka::is_acc_tag_match<TAcc, TTag>::value;
```
bernhardmgruber: Variadic version:
```suggestion
    template<typename TAcc, typename... TTags>
    inline constexpr bool is_acc_tag_match_v = (alpaka::is_acc_tag_match<TAcc, TTag>::value || ...);
```
And again, I would rename it to not have the `_v` and use `CamelCase`:
```suggestion
    template<typename TAcc, typename... TTags>
    inline constexpr bool AccMatchesTags = (alpaka::is_acc_tag_match<TAcc, TTag>::value || ...);
```

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,64 @@
...
+    struct AccToTag;
+
+    template<typename TAcc>
+    using AccToTag_t = typename AccToTag<TAcc>::type;
+
+    template<typename TTag, typename TDim, typename TIdx>
+    struct TagToAcc;
+
+    template<typename TTag, typename TDim, typename TIdx>
+    using TagToAcc_t = typename TagToAcc<TTag, TDim, TIdx>::type;
```
bernhardmgruber: The user facing API should be `TagToAcc`.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,64 @@
...
+
+    // TODO: make variadic, to support more than on tag per call
+    template<typename TAcc, typename TTag>
+    using is_acc_tag_match = std::is_same<AccToTag_t<TAcc>, TTag>;
+
+    template<typename TAcc, typename TTag>
+    inline constexpr bool is_acc_tag_match_v = alpaka::is_acc_tag_match<TAcc, TTag>::value;
+
+    template<typename TAcc, typename TTag>
+    using is_acc_tag_match_t = typename is_acc_tag_match<TAcc, TTag>::type;
```
bernhardmgruber: Why do we need this one?

test/unit/acc/src/AccTagTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+    alpaka::AccOaccTag,
+    alpaka::AccGpuCudaRtTag,
+    alpaka::AccGpuHipRtTag,
+    alpaka::AccCpuSyclIntelTag,
+    alpaka::AccFpgaSyclIntelTag,
+    alpaka::AccFpgaSyclXilinxTag,
+    alpaka::AccGpuSyclIntelTag>;
+
+using AccTagTuple = std::tuple<
+    std::tuple<alpaka::test::detail::AccCpuSerialIfAvailableElseInt<Dim, Idx>, alpaka::AccCpuSerialTag>,
```
bernhardmgruber: In mp11 jargon, this is a map:
```suggestion
using AccToTagMap = std::tuple<
    std::tuple<alpaka::test::detail::AccCpuSerialIfAvailableElseInt<Dim, Idx>, alpaka::AccCpuSerialTag>,
```

test/unit/acc/src/AccTagTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+        if constexpr(std::is_same_v<TestTag, ExpectedTag>)
+        {
+            STATIC_REQUIRE((std::is_same_v<alpaka::AccToTag_t<TestAcc>, ExpectedTag>) );
+            STATIC_REQUIRE((std::is_same_v<alpaka::TagToAcc_t<TestTag, Dim, Idx>, TestAcc>) );
+            STATIC_REQUIRE((alpaka::is_acc_tag_match_v<TestAcc, TestTag>) );
+        }
+        else
+        {
+            STATIC_REQUIRE_FALSE((std::is_same_v<alpaka::AccToTag_t<TestAcc>, TestTag>) );
+            STATIC_REQUIRE_FALSE((alpaka::is_acc_tag_match_v<TestAcc, TestTag>) );
```
bernhardmgruber:
```suggestion
            STATIC_REQUIRE(!(std::is_same_v<alpaka::AccToTag_t<TestAcc>, TestTag>) );
            STATIC_REQUIRE(!(alpaka::is_acc_tag_match_v<TestAcc, TestTag>) );
```

test/unit/acc/src/AccTagTest.cpp
```diff
@@ -0,0 +1,218 @@
...
+    using ExpectedTag = std::tuple_element_t<1, TestAccTestTuple>;
+    using TestTag = std::tuple_element_t<1, TestType>;
+
+    // if the Acc is not enabled, the type is int
+    if constexpr(!std::is_same_v<TestAcc, int>)
+    {
+        std::cout << "Tested Acc: " << alpaka::getAccName<TestAcc>() << "\n"
+                  << "Expected Tag: " << ExpectedTag::get_name() << "\n"
+                  << "Tested Tag: " << TestTag::get_name() << "\n"
+                  << std::endl;
```
bernhardmgruber: Please remove unnecessary output statements. They make the CI logs longer and errors harder to find.

include/alpaka/acc/Tag.hpp
SimeonEhrig: This was my first approach but I thought `_t` feels more like C++ stl.

include/alpaka/acc/Tag.hpp
SimeonEhrig: Thanks for the variadic version. Saves me time. @psychocoderHPC means, that is maybe required to have a type and value for some traits.

include/alpaka/acc/Tag.hpp
SimeonEhrig: See comment variadic version.

test/unit/acc/src/AccTagTest.cpp
SimeonEhrig: That's a mistake. It should be `INFO()` and only print something, if the test is going wrong. But on the other site, that's not possible, because the checks are static and throws and compile time error :thinking:

include/alpaka/acc/Tag.hpp
bernhardmgruber: Wait, it should be lowercase, since it is a variable name, so `accMatchesTags`.

test/unit/acc/src/AccTagTest.cpp
SimeonEhrig: I removed it

include/alpaka/acc/Tag.hpp
SimeonEhrig: Renamed it after I found the reaction of @psychocoderHPC . 
@psychocoderHPC  You should not answer with a reaction. I didn't get a notification. It was only luck, that I saw it.

include/alpaka/acc/Tag.hpp
SimeonEhrig: new name of the Struct: `AccToTagTrait`

example/tagSpecialization/src/tagSpecialization.cpp
SimeonEhrig: I add also an example to specialize a kernel. Therefore, I think I need it.

example/tagSpecialization/src/tagSpecialization.cpp
SimeonEhrig: done

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,66 @@
...
+    CREATE_ACC_TAG(AccFpgaSyclXilinxTag);
+    CREATE_ACC_TAG(AccGenericSyclTag);
+    CREATE_ACC_TAG(AccGpuCudaRtTag);
+    CREATE_ACC_TAG(AccGpuHipRtTag);
+    CREATE_ACC_TAG(AccGpuSyclIntelTag);
+    CREATE_ACC_TAG(AccOaccTag);
+    CREATE_ACC_TAG(AccOmp5Tag);
+
+    template<typename TAcc>
+    struct AccToTagTrait;
```
psychocoderHPC: IMO all traits should be in the namespace trait and not mangled with the word trait.
Else it will not follow the structure we have already in alpaka for other rtraits.

```C++
namespace trait
{
  template<typename TAcc>
    struct AccToTag;

template<typename TTag, typename TDim, typename TIdx>
    struct TagToAcc;
}
```

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    template<typename TAcc>
+    using AccToTag = typename AccToTagTrait<TAcc>::type;
```
psychocoderHPC: This should be `AccToTagType`  and should be part of the namespace `trait` too.
We have something similar already with the trait `QueueType` in alpaka.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    template<typename TTag, typename TDim, typename TIdx>
+    struct TagToAccTrait;
+
+    template<typename TTag, typename TDim, typename TIdx>
+    using TagToAcc = typename TagToAccTrait<TTag, TDim, TIdx>::type;
```
psychocoderHPC: Same as `AccToTag` this should be `trait::TagToAccType`

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,66 @@
...
+    {                                                                                                                 \
+        static std::string get_name()                                                                                 \
+        {                                                                                                             \
+            return #tag_name;                                                                                         \
+        }                                                                                                             \
+    }
+
+namespace alpaka
+{
+    CREATE_ACC_TAG(AccCpuFibersTag);
```
psychocoderHPC: IMO this should be in the namespace `tag` then there is no need to add `Tag` to the name.
IMO namespaces are exactly for such hierarchy collections.

The usage will not be much longer `alpaka::AccCpuFibersTag` vs `alpaka::tag::AccCpuFibers`.
Or with an inline namespace `alpaka::AccCpuFibers` but this will collide with the implementation of the accelerator :-(

I am not sure if there is a need to name the tag `Acc...`, `alpaka::tag::CpuFibers` is IMO fine.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+
+/* TODO: Remove the following pragmas once support for clang 6 is removed. They are necessary because
+        these /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+    template<typename TAcc, typename... TTag>
+    inline constexpr bool accMatchesTags = (std::is_same_v<AccToTag<TAcc>, TTag> || ...);
```
psychocoderHPC: A doxygen comment would be nice.

e.g.
```
Checks if the given accelerator is equal to one of the given tags in the list.
```

There should be no need for `inline`!

psychocoderHPC: Note: In PIConGPU we write such constructs always as list operations t allow the user to create an intermediate list of e.g. tags instead of hard code these into the call.

```C++
// this list is somewhere defined in the code, not directly where the tag matcher is used
using myPrefairedAccs = std::tupel<AccCpuOmp2BlocksTag, AccCpuSerialTag>;

auto isPrefaired = accMatchesTags<CurrentAcc,myPrefairedAccs>;
// we always allow passing one element only without saying it is a list, 
// in this case, we generate a list with one element and pass it to the 
// implementation which accepts lists only
auto isHip = accMatchesTags<CurrentAcc,AccGpuHipRtTag>;
```

psychocoderHPC (changes requested):
Thanks for your PR, it is much nicer than what I did!
Some parts must be rearranged a little bit to fit into the style/concepts we already use in alpaka.
Namings should be discussed with other maintainers in this PR or latest in the next developer meeting.

include/alpaka/acc/Tag.hpp
SimeonEhrig: Makes sense. In the first implementation, it didn't made sense but now it is useful. I will change the relic ;-)

include/alpaka/acc/Tag.hpp
SimeonEhrig: Do you want to put the `AccToTagType` to the `trait` namespace and provide a flat namespace like `alpaka::NonBlocking` for `alpaka::property::NonBlocking`?

include/alpaka/acc/Tag.hpp
SimeonEhrig: I would suggest two solutions:

1. I would cover @psychocoderHPC  suggestion and provide only the tags inside the namespace `tag` without the `Acc` prefix (e.g. `alpaka::tag::CpuFibers`). A flat namespace would be not useful (e.g. `alpaka::CpuFibers`), because the name is to meaningless.
2. Prefix the names with `Tag` instead `Acc`, e.g. (`alpaka::TagCpuFibers`). We can also put the tags in the namespace `tag` for a better organization. The full API is only a little bit verbose/redundant: `alpaka::tag::TagCpuFibers`

include/alpaka/acc/Tag.hpp
SimeonEhrig: @psychocoderHPC Can you please point me to an example in PIConGPU, which we disused offline. I found an example in alpaka, but the requires also to pass the list type (e.g. `std::tuple`): https://github.com/alpaka-group/alpaka/blob/d98d7707a68e0adade3937b39c966cac9aad5e5d/include/alpaka/meta/CartesianProduct.hpp#L80-L81

test/unit/acc/src/AccTagTest.cpp
```diff
@@ -0,0 +1,269 @@
...
+    alpaka::AccCpuSyclIntelTag,
+    alpaka::AccFpgaSyclIntelTag,
+    alpaka::AccFpgaSyclXilinxTag,
+    alpaka::AccGpuSyclIntelTag>;
+
+// to many acc's triggers a compiler error in nvc++: error: excessive recursion at instantiation of class
+// looks like, that a buffer is not big enough
+#ifdef ALPAKA_ACC_ANY_BT_OACC_ENABLED
+using AccToTagMap = std::tuple<
+    std::tuple<alpaka::test::detail::AccCpuSerialIfAvailableElseInt<Dim, Idx>, alpaka::AccCpuSerialTag>,
```
j-stephan: Is there a reason why you chose `std::tuple<std::tuple<X, Y>>` over `std::tuple<std::pair<X, Y>>`?

include/alpaka/acc/Tag.hpp
j-stephan: I vote for option 2 without the extra namespace, so `alpaka::TagCpuFibers`. @bernhardmgruber's mental health is important to me and he put so much work into getting rid of additional namespaces in alpaka... but on a more serious note I think it matches our current `AccXyz` naming scheme nicely.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    template<typename TTag, typename TDim, typename TIdx>
+    struct TagToAccTrait;
+
+    template<typename TTag, typename TDim, typename TIdx>
+    using TagToAcc = typename TagToAccTrait<TTag, TDim, TIdx>::type;
+
+
+/* TODO: Remove the following pragmas once support for clang 6 is removed. They are necessary because
+        these /  clang versions incorrectly warn about a missing 'extern'. */
```
j-stephan, line 61: We should correct this comment sometime (looks like a rogue find&replace). No need to do this now, though.

include/alpaka/acc/Tag.hpp
j-stephan: > There should be no need for `inline`!

The behaviour of `inline` changed with C++17. If we leave it out the compiler will complain about `accMatchesTags` being defined multiple times.

SimeonEhrig: > Note: In PIConGPU we write such constructs always as list operations t allow the user to create an intermediate list of e.g. tags instead of hard code these into the call.
> 
> ```c++
> // this list is somewhere defined in the code, not directly where the tag matcher is used
> using myPrefairedAccs = std::tupel<AccCpuOmp2BlocksTag, AccCpuSerialTag>;
> 
> auto isPrefaired = accMatchesTags<CurrentAcc,myPrefairedAccs>;
> // we always allow passing one element only without saying it is a list, 
> // in this case, we generate a list with one element and pass it to the 
> // implementation which accepts lists only
> auto isHip = accMatchesTags<CurrentAcc,AccGpuHipRtTag>;
> ```

I like the idea, but unfortunately we have no `boost mp11` and I didn't found any implementation in alpaka, which helps me to implement the feature. I would move the feature to a new issue/PR to get this PR finish soon.

psychocoderHPC: I am fine with adding this feature to another PR and not this.

include/alpaka/acc/Tag.hpp
SimeonEhrig: What do you want to correct? It explains the clang 6 bug really well.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,69 @@
...
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+    template<typename TAcc, typename... TTag>
+    inline constexpr bool accMatchesTags = (std::is_same_v<alpaka::trait::AccToTagType<TAcc>, TTag> || ...);
+
```
psychocoderHPC, line 68: You missed to pop the diagnostics.

example/tagSpecialization/src/tagSpecialization.cpp
```diff
@@ -0,0 +1,143 @@
...
+    {
+        printf("generic kernel\n");
+        return alpaka::math::min(acc, 0, input);
+    }
+};
+
+template<typename TAcc>
+struct specialized_Kernel<
+    TAcc,
+    std::enable_if_t<alpaka::accMatchesTags<TAcc, alpaka::TagCpuOmp2Blocks, alpaka::TagCpuOmp2Blocks>>>
```
psychocoderHPC: You used twice `alpaka::TagCpuOmp2Blocks`, IMO the intention was that one of these should be `alpaka::TagCpuOmp2Threads`

include/alpaka/acc/Tag.hpp
psychocoderHPC: The todo should be doxygen syntax `/** \todo` this will help that the todo not get lost.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,69 @@
...
+        template<typename TTag, typename TDim, typename TIdx>
+        using TagToAccType = typename TagToAcc<TTag, TDim, TIdx>::type;
+    } // namespace trait
+
+
+/* TODO: Remove the following pragmas once support for clang 6 is removed. They are necessary because
+        these /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
```
psychocoderHPC, line 64: The corresponding `pop` for the stack of diagnostics is missing. All options pushed must be reverted after the end of the guarded lines.

psychocoderHPC (changes requested):

psychocoderHPC (approved):

include/alpaka/acc/Tag.hpp
psychocoderHPC: is now in the trait namespace.

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,71 @@
...
+
+
+/** \todo: Remove the following pragmas once support for clang 6 is removed. They are necessary because
+        these /  clang versions incorrectly warn about a missing 'extern'. */
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
+#endif
+    template<typename TAcc, typename... TTag>
+    inline constexpr bool accMatchesTags = (std::is_same_v<alpaka::trait::AccToTagType<TAcc>, TTag> || ...);
```
fwyzard, line 67: minor comment, even though the changes are merged already: wouldn't it be more in line with the usual c++ traits to use `acc_matches_tags` ?

include/alpaka/acc/Tag.hpp
```diff
@@ -0,0 +1,71 @@
...
+        struct AccToTag;
+
+        template<typename TAcc>
+        using AccToTagType = typename AccToTag<TAcc>::type;
+
+        template<typename TTag, typename TDim, typename TIdx>
+        struct TagToAcc;
+
+        template<typename TTag, typename TDim, typename TIdx>
+        using TagToAccType = typename TagToAcc<TTag, TDim, TIdx>::type;
```
fwyzard, line 56: other minor comment: wouldn't it be more in line with the other alpaka types (or traits) to use `alpaka::Tag<Acc>` and `alpaka::Acc<Tag, ...>` ?

include/alpaka/acc/Tag.hpp
SimeonEhrig: The alpaka name convention sad, that we use lower case camel case: https://alpaka.readthedocs.io/en/latest/dev/style.html#naming

So it's consistent with alpaka.

include/alpaka/acc/Tag.hpp
SimeonEhrig: see: #1814

Issue #1296, opened 2021-04-19
Consider adopting a metaprogramming library


bernhardmgruber:
During the accessor development #1249 I needed to implement a few meta functions on the side. Since alpaka is TMP heavy, we are going to need such metaprogramming facilities regularly so I think we should consider picking an appropriate library providing this functionality.

LLAMA uses boost::mp11 quite successfully and it provides a good feature set. Mind, that boost::mp11 is also available as a standalone library outside the usual boost distribution.

sbastrakov:
FYI we actually tried `boost::mp11` for PIConGPU 3 years ago, turned out there were issues with GPUs, and general approach not matching the needs very well back then. Doesn't mean these two factors are still relevant.

bernhardmgruber:
That's good to know! I have not encountered GPU issues yet in LLAMA, but my GPU usage is also limited.

bussmann:
Your last comment is extremely relevant. HPC is often bleeding edge and certain hardware 2-3 years behind in supporting standards, but these years make all the difference in real world applications. Thus, dependencies have to be carefully chosen. Without first hand experience this choice is difficult. One would be amazed which simple assumptions may be wrong for HPC applications.

psychocoderHPC:
> During the accessor development #1249 I needed to implement a few meta functions on the side. Since alpaka is TMP heavy, we are going to need such metaprogramming facilities regularly so I think we should consider picking an appropriate library providing this functionality.
> 
> LLAMA uses boost::mp11 quite successfully and it provides a good feature set. Mind, that boost::mp11 is also available as a standalone library outside the usual boost distribution.

Could you please link the stand-alone mp11 version? A quick search was always showing the boost version only.

bernhardmgruber:
> HPC is often bleeding edge and certain hardware 2-3 years behind in supporting standards, but these years make all the difference in real world applications.

Thankfully, mp11 is written in C++11 and thus uses a 10 year old standard. 

> Could you please link the stand-alone mp11 version? A quick search was always showing the boost version only.

https://github.com/boostorg/mp11
The README says:

> Mp11 is part of Boost, starting with release 1.66.0. It however has no Boost dependencies and can be used standalone, as a Git submodule, for instance. For CMake users, add_subdirectory is supported, as is installation and find_package(boost_mp11).

psychocoderHPC:
Thanks for the link: https://github.com/boostorg/mp11
[...]

PR #1805, opened 2022-09-30
Add `Vec::elementwise_min` and `Vec::elementwise_max` functions

fwyzard:
Return a vector where each element is set to the minimum (or maximum) of the elements in the same dimension of all the input vectors.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -454,6 +454,42 @@ namespace alpaka
     template<typename TFirstIndex, typename... TRestIndices>
     Vec(TFirstIndex&&, TRestIndices&&...) -> Vec<DimInt<1 + sizeof...(TRestIndices)>, std::decay_t<TFirstIndex>>;
 
+    //! \return The element-wise minimum of one or more vectors.
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<
+        typename TDim,
+        typename TVal,
+        typename... Vecs,
+        typename = std::enable_if_t<(std::is_same_v<Vec<TDim, TVal>, Vecs> && ...)>>
+    ALPAKA_FN_HOST_ACC constexpr auto elementwise_min(Vec<TDim, TVal> const& p, Vecs const&... qs) -> Vec<TDim, TVal>
+    {
+        Vec<TDim, TVal> r;
+        if constexpr(TDim::value > 0)
+        {
+            for(typename TDim::value_type i = 0; i < TDim::value; ++i)
+                r[i] = std::min({p[i], qs[i]...});
+        }
+        return r;
+    }
+
+    //! \return The element-wise maximum of one or more vectors.
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<
+        typename TDim,
+        typename TVal,
+        typename... Vecs,
+        typename = std::enable_if_t<(std::is_same_v<Vec<TDim, TVal>, Vecs> && ...)>>
+    ALPAKA_FN_HOST_ACC constexpr auto elementwise_max(Vec<TDim, TVal> const& p, Vecs const&... qs) -> Vec<TDim, TVal>
+    {
+        Vec<TDim, TVal> r;
+        if constexpr(TDim::value > 0)
+        {
+            for(typename TDim::value_type i = 0; i < TDim::value; ++i)
+                r[i] = std::max({p[i], qs[i]...});
+        }
+        return r;
+    }
+
     namespace trait
     {
         //! The Vec dimension get trait specialization.
--- a/test/unit/vec/src/VecTest.cpp
+++ b/test/unit/vec/src/VecTest.cpp
@@ -262,6 +262,30 @@ TEST_CASE("basicVecTraits", "[vec]")
             STATIC_REQUIRE(referenceVec == vecLessEqual);
         }
 
+        // alpaka::Vec elementwise_min
+        {
+            STATIC_REQUIRE(alpaka::elementwise_min(vec) == vec);
+
+            constexpr auto result = alpaka::elementwise_min(vec, vec3);
+            STATIC_REQUIRE(std::is_same_v<alpaka::Dim<std::decay_t<decltype(result)>>, Dim>);
+            STATIC_REQUIRE(std::is_same_v<alpaka::Idx<std::decay_t<decltype(result)>>, Idx>);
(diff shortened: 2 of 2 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -402,6 +404,34 @@ namespace alpaka
             return r;
         }
 
+        //! \return The element-wise minimum of one or more vectors.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename... Vecs, typename = std::enable_if_t<boost::mp11::mp_same<Vec, Vecs...>::value>>
```
j-stephan: We have decided against the integration of Boost.Mp11 a while back. From what I can see you need it for `mp_same`. Shouldn't it be possible to use C++17 fold expressions here?

```suggestion
        template<typename... Vecs, typename = std::enable_if_t<std::is_same_v<Vec, Vecs> && ...>>
```

CC @bernhardmgruber

bernhardmgruber: Let me rephrase: We don't have consensus yet to allow the use of Boost.Mp11. The relevant discussion is here #1296.

In this particular instance, the fold expression suggested by Jan should work equally well!

include/alpaka/vec/Vec.hpp
```diff
@@ -402,6 +404,34 @@ namespace alpaka
...
+                for(typename TDim::value_type i = 0; i < TDim::value; ++i)
+                    r[i] = std::min({p[i], qs[i]...});
+            }
+            return r;
+        }
+
+        //! \return The element-wise maximum of one or more vectors.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename... Vecs, typename = std::enable_if_t<boost::mp11::mp_same<Vec, Vecs...>::value>>
+        ALPAKA_FN_HOST_ACC friend constexpr auto elementwise_max(Vec const& p, Vecs const&... qs) -> Vec
```
bernhardmgruber: Since you seem to intend to call this functions via a qualified name `alpaka::elementwise_max` and not ADL-only: please move it out of the definition of the `Vec` class and make it an ordinary free function. This should also make your unit test compile.

bernhardmgruber (changes requested):

include/alpaka/vec/Vec.hpp
fwyzard: Ah... I read the first part of

> A name first declared in a friend declaration within a class or class template `X` becomes a member of the innermost enclosing namespace of `X`, ...

but  I missed the second part

> ... but is not visible for lookup (except argument-dependent lookup that considers `X`) unless a matching declaration at namespace scope is provided.

include/alpaka/vec/Vec.hpp
fwyzard: OK, I see.

Thanks, the fold expression does seem to work well (it only needs a couple of `(...)`).

include/alpaka/vec/Vec.hpp
bernhardmgruber: Ok, now you added declarations of the functions outside `Vec`. Why not move the entire definition outside as well? That would reduce the amount of code.

include/alpaka/vec/Vec.hpp
```diff
@@ -402,6 +402,34 @@ namespace alpaka
             return r;
         }
 
+        //! \return The element-wise minimum of one or more vectors.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename... Vecs, typename = std::enable_if_t<(std::is_same_v<Vec, Vecs> && ...)>>
+        ALPAKA_FN_HOST_ACC friend constexpr auto elementwise_min(Vec const& p, Vecs const&... qs) -> Vec
```
bernhardmgruber:
```suggestion
        template<typename... Vecs, typename = std::enable_if_t<(std::is_same_v<Vec, Vecs> && ... && true)>>
        ALPAKA_FN_HOST_ACC friend constexpr auto elementwise_min(Vec const& p, Vecs const&... qs) -> Vec
```
To support `elementwise_min` with a single argument.

test/unit/vec/src/VecTest.cpp
```diff
@@ -262,6 +262,32 @@ TEST_CASE("basicVecTraits", "[vec]")
...
+            constexpr auto vecMin = alpaka::elementwise_min(vec, vec3);
+            STATIC_REQUIRE(std::is_same_v<alpaka::Dim<std::decay_t<decltype(vecMin)>>, Dim>);
+            STATIC_REQUIRE(std::is_same_v<alpaka::Idx<std::decay_t<decltype(vecMin)>>, bool>);
+
+            constexpr alpaka::Vec<Dim, Idx> referenceVec(
+                static_cast<Idx>(0u),
+                static_cast<Idx>(8u),
+                static_cast<Idx>(3u));
+            STATIC_REQUIRE(referenceVec == vecMin);
+        }
```
bernhardmgruber: Consider adding tests for calling `elementwise_min` with a single argument.
Please also add a test calling it with more than 2 arguments. Thx!

include/alpaka/vec/Vec.hpp
fwyzard: I assumed they still need to be declared as `friend` to access the `Vec`'s members ?

bernhardmgruber: Yes they would, but IIUC you are not accessing anything private.

bernhardmgruber:
Please also rebase your PR on the latest `develop`!

include/alpaka/vec/Vec.hpp
fwyzard: Actually that's not needed:

> When a unary fold is used with a pack expansion of length zero, only the following operators are allowed:
>  1) **Logical AND (`&&`)**. The value for the empty pack is `true`
>  2) Logical OR (`||`). The value for the empty pack is `false`
>  3) The comma operator (`,`). The value for the empty pack is `void()`


But it's good to check - I'm adding a test with a single argument to the unit test.

test/unit/vec/src/VecTest.cpp
fwyzard: Done.

include/alpaka/vec/Vec.hpp
fwyzard: 👍🏻

bernhardmgruber (approved):

PR #1814, opened 2022-10-19
rename AccToTagType and TagToAccType

SimeonEhrig:
alpaka::trait::AccToTagType -> alpaka::trait::Tag
alpaka::trait::TagToAccType -> alpaka::trait::Acc

Suggestion of @fwyzard : https://github.com/alpaka-group/alpaka/pull/1804#discussion_r995653153

The PR starts as draft to discuss if we want to rename it.

There is no name collision. I looked a little bit in the code. We use the word `Acc` only in one situations, in examples: `using Acc = alpaka::AccCpuSerial<>`. The word `Tag` is not used anywhere. So I think it's pretty hard to mistake the words `Acc` and `Tag`.

### DIFF
--- a/include/alpaka/acc/Tag.hpp
+++ b/include/alpaka/acc/Tag.hpp
@@ -37,16 +37,22 @@ namespace alpaka
         template<typename TAcc>
         struct AccToTag;
 
-        template<typename TAcc>
-        using AccToTagType = typename AccToTag<TAcc>::type;
-
         template<typename TTag, typename TDim, typename TIdx>
         struct TagToAcc;
-
-        template<typename TTag, typename TDim, typename TIdx>
-        using TagToAccType = typename TagToAcc<TTag, TDim, TIdx>::type;
     } // namespace trait
 
+    /// @brief maps an acc type to a tag type
+    /// @tparam TAcc alpaka acc type
+    template<typename TAcc>
+    using AccToTag = typename trait::AccToTag<TAcc>::type;
+
+    /// @brief maps a tag type to an acc type
+    /// @tparam TTag alpaka tag type
+    /// @tparam TDim dimension of the mapped acc type
+    /// @tparam TIdx index type of the mapped acc type
+    template<typename TTag, typename TDim, typename TIdx>
+    using TagToAcc = typename trait::TagToAcc<TTag, TDim, TIdx>::type;
+
     template<typename TAcc, typename... TTag>
-    inline constexpr bool accMatchesTags = (std::is_same_v<alpaka::trait::AccToTagType<TAcc>, TTag> || ...);
+    inline constexpr bool accMatchesTags = (std::is_same_v<alpaka::AccToTag<TAcc>, TTag> || ...);
 } // namespace alpaka
--- a/test/unit/acc/src/AccTagTest.cpp
+++ b/test/unit/acc/src/AccTagTest.cpp
@@ -57,13 +57,13 @@ TEMPLATE_LIST_TEST_CASE("test all possible acc tag combinations", "[acc][tag]", 
     {
         if constexpr(std::is_same_v<TestTag, ExpectedTag>)
         {
-            STATIC_REQUIRE((std::is_same_v<alpaka::trait::AccToTagType<TestAcc>, ExpectedTag>) );
-            STATIC_REQUIRE((std::is_same_v<alpaka::trait::TagToAccType<TestTag, Dim, Idx>, TestAcc>) );
+            STATIC_REQUIRE((std::is_same_v<alpaka::AccToTag<TestAcc>, ExpectedTag>) );
+            STATIC_REQUIRE((std::is_same_v<alpaka::TagToAcc<TestTag, Dim, Idx>, TestAcc>) );
             STATIC_REQUIRE((alpaka::accMatchesTags<TestAcc, TestTag>) );
         }
         else
         {
-            STATIC_REQUIRE(!(std::is_same_v<alpaka::trait::AccToTagType<TestAcc>, TestTag>) );
+            STATIC_REQUIRE(!(std::is_same_v<alpaka::AccToTag<TestAcc>, TestTag>) );
             STATIC_REQUIRE(!(alpaka::accMatchesTags<TestAcc, TestTag>) );
         }
     }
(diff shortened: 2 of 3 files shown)


SimeonEhrig:
@bernhardmgruber @j-stephan @psychocoderHPC @sliwowitz What are you thinking?

j-stephan:
Fine for me.

bernhardmgruber (dismissed):

include/alpaka/acc/Tag.hpp
```diff
@@ -47,13 +47,13 @@ namespace alpaka
         struct AccToTag;
 
         template<typename TAcc>
-        using AccToTagType = typename AccToTag<TAcc>::type;
+        using Tag = typename AccToTag<TAcc>::type;
```
psychocoderHPC: Please add a comment to jothbtraits what they should do.

psychocoderHPC (changes requested):

include/alpaka/acc/Tag.hpp
psychocoderHPC: e.g. `maps accelerator to tag types`

psychocoderHPC: ~~As I wrote https://github.com/alpaka-group/alpaka/pull/1804#discussion_r987592586 I think the naming should end with `Type` to fit the already existing traits e.g. `QueueType`~~

include/alpaka/acc/Tag.hpp
```diff
@@ -47,13 +47,13 @@ namespace alpaka
         struct AccToTag;
```
psychocoderHPC, line 38: If we follow the naming of a similar trait `EventType` https://github.com/alpaka-group/alpaka/blob/9ec06cb61e7af09a5ac934b7c4d5a28333557e87/include/alpaka/event/Traits.hpp#L22
then `AccToTag` should be `TagType` and 

```
template<typename TAcc>
        using Tag = typename AccToTag<TAcc>::type;
```

should be

```
template<typename TAcc>
using Tag = typename trait::TagType<TAcc>::type;
```

and should be located in the alpaka namespace. @SimeonEhrig sry that I have overseen it in the original PR.

include/alpaka/acc/Tag.hpp
```diff
@@ -47,13 +47,13 @@ namespace alpaka
...
 
         template<typename TAcc>
-        using AccToTagType = typename AccToTag<TAcc>::type;
+        using Tag = typename AccToTag<TAcc>::type;
 
         template<typename TTag, typename TDim, typename TIdx>
         struct TagToAcc;
```
psychocoderHPC, line 41: should be `AccType`

include/alpaka/acc/Tag.hpp
```diff
@@ -47,13 +47,13 @@ namespace alpaka
...
 
         template<typename TTag, typename TDim, typename TIdx>
-        using TagToAccType = typename TagToAcc<TTag, TDim, TIdx>::type;
+        using Acc = typename TagToAcc<TTag, TDim, TIdx>::type;
```
psychocoderHPC: should be in alpaka namespace an

```
template<typename TTag, typename TDim, typename TIdx>
using Acc = typename trait::AccType<TTag, TDim, TIdx>::type;
```

psychocoderHPC (changes requested):
sry that I have overseen it but I found new similar traits e.g. `EventType` which should be used as an example to keep the code consistent.

SimeonEhrig:
@psychocoderHPC `alpaka::Acc` and `alpaka::trait::AccType` already exists:
https://github.com/alpaka-group/alpaka/blob/ffb8307194fbf5b7149533c1a4cfbba1577347dd/include/alpaka/acc/Traits.hpp#L58

bernhardmgruber:
> @psychocoderHPC `alpaka::Acc` and `alpaka::trait::AccType` already exists:
> 
> https://github.com/alpaka-group/alpaka/blob/ffb8307194fbf5b7149533c1a4cfbba1577347dd/include/alpaka/acc/Traits.hpp#L58

Why is this a problem? You can still add specializations that convert a tag to an acc, by specializing `alpaka::traits::AccType` and using it via `alpaka::Acc<Tag>`.

SimeonEhrig:
@bernhardmgruber The name resolution is resolved. I don't know why. I did simply rebase.

include/alpaka/acc/Tag.hpp
bernhardmgruber: I agree.

test/unit/acc/src/AccTagTest.cpp
```diff
@@ -57,13 +57,13 @@ TEMPLATE_LIST_TEST_CASE("test all possible acc tag combinations", "[acc][tag]",
     {
         if constexpr(std::is_same_v<TestTag, ExpectedTag>)
         {
-            STATIC_REQUIRE((std::is_same_v<alpaka::trait::AccToTagType<TestAcc>, ExpectedTag>) );
-            STATIC_REQUIRE((std::is_same_v<alpaka::trait::TagToAccType<TestTag, Dim, Idx>, TestAcc>) );
+            STATIC_REQUIRE((std::is_same_v<alpaka::trait::Tag<TestAcc>, ExpectedTag>) );
+            STATIC_REQUIRE((std::is_same_v<alpaka::trait::Acc<TestTag, Dim, Idx>, TestAcc>) );
```
bernhardmgruber: This is how I would like the API to look like:
```suggestion
            STATIC_REQUIRE((std::is_same_v<alpaka::Tag<TestAcc>, ExpectedTag>) );
            STATIC_REQUIRE((std::is_same_v<alpaka::Acc<TestTag, Dim, Idx>, TestAcc>) );
```

include/alpaka/acc/Tag.hpp
bernhardmgruber: I agree.

bernhardmgruber:
@SimeonEhrig please take a go at the feedback from @psychocoderHPC!

bernhardmgruber:
Internal discussion revealed, that we cannot directly use the `alpaka::Acc<T>` type function on a tag, because we would need to pass additional arguments: `alpaka::Acc<Tag, Dim, Idx>`. We could make `alpaka::Acc` variadic to accomondate for that, but maybe it's better to leave Tag <-> Acc conversions an independent feature.

SimeonEhrig:
I prefer to let it independent, because template error messages are horrible and I think it could be hard for users to understand the different usage of the 1 template parameter and 3 template parameter version of `alpaka::Acc`.

psychocoderHPC:
offline discussion with @SimeonEhrig 
Due to naming conflicts with existing traits, I suggest using the name `TagToAcc` and `AccToTag`.
IMO Using variadic templates to be able to use the naming `Acc<Tag,Dim, Idx>` and `Tag<T>` could have too many side effects.

include/alpaka/acc/Tag.hpp
SimeonEhrig: Done

SimeonEhrig:
After some discussion, we decided to do not use a variadic template. Instead I used distinguish names:

alpaka::trait::AccToTagType -> alpaka::AccToTag
alpaka::trait::TagToAccType -> alpaka::TagToAcc

bernhardmgruber (approved):

j-stephan (approved):

psychocoderHPC (approved):

PR #1816, opened 2022-10-20
Fix missing `final` keyword for accelerator inheritance

psychocoderHPC:
Most of the accelerators a marked with the keyword `final`, therefore it is not possible to inherit from these accelerators.

This PR adds the final keyword to all accelerators to avoid different behavior depending on the accelerator.

### DIFF
--- a/include/alpaka/acc/AccCpuSyclIntel.hpp
+++ b/include/alpaka/acc/AccCpuSyclIntel.hpp
@@ -36,7 +36,7 @@ namespace alpaka::experimental
     //!
     //! This accelerator allows parallel kernel execution on a oneAPI-capable Intel CPU target device.
     template<typename TDim, typename TIdx>
-    class AccCpuSyclIntel
+    class AccCpuSyclIntel final
         : public AccGenericSycl<TDim, TIdx>
         , public concepts::Implements<ConceptAcc, AccCpuSyclIntel<TDim, TIdx>>
     {
--- a/include/alpaka/acc/AccFpgaSyclIntel.hpp
+++ b/include/alpaka/acc/AccFpgaSyclIntel.hpp
@@ -35,7 +35,7 @@ namespace alpaka::experimental
     //!
     //! This accelerator allows parallel kernel execution on a oneAPI-capable Intel FPGA target device.
     template<typename TDim, typename TIdx>
-    class AccFpgaSyclIntel
+    class AccFpgaSyclIntel final
         : public AccGenericSycl<TDim, TIdx>
         , public concepts::Implements<ConceptAcc, AccFpgaSyclIntel<TDim, TIdx>>
     {
--- a/include/alpaka/acc/AccFpgaSyclXilinx.hpp
+++ b/include/alpaka/acc/AccFpgaSyclXilinx.hpp
@@ -35,7 +35,7 @@ namespace alpaka::experimental
     //!
     //! This accelerator allows parallel kernel execution on a SYCL-capable Xilinx FPGA target device.
     template<typename TDim, typename TIdx>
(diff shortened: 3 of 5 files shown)


psychocoderHPC:
In general, I do not know why we forbid that we can inherit from an accelerator. As I can see all acc's can be copied too, does anyone has an idea why it is how it is?

include/alpaka/acc/AccGenericSycl.hpp
```diff
@@ -51,7 +51,7 @@ namespace alpaka::experimental
     //!
     //! This accelerator allows parallel kernel execution on SYCL devices.
     template<typename TDim, typename TIdx>
-    class AccGenericSycl
+    class AccGenericSycl final
```
j-stephan: This must not be `final` because it is the abstract base class for the concrete SYCL implementations.

j-stephan (changes requested):

j-stephan:
My guess is that we don't want users to derive from our `Acc`s and then override parent methods. This is legal in C++ but might have bad consequences in CUDA and probably the other back-ends, too. But this is something I only can assume, maybe @BenjaminW3 still remembers.

If we cannot establish a reason for keeping them `final` the keywords should be removed IMO. Maybe there was a good reason in the past that is no longer valid.

BenjaminW3:
I also do not remember why they are final 🤷

psychocoderHPC:
The question why I am asking is that I am currently [refactoring PIConGPU/PMacc](https://github.com/ComputationalRadiationPhysics/picongpu/pull/4321) where we introduce our own context element.
Currently, I store within this PMacc context a reference to the alpaka ACC which is not nice at all.
If we would allow that I can inherit from an accelerator I could pass the PMacc context to any alpaka function and because of our alpaka concepts light, all should work nice.
So in general would like to get rid of the final keyword to extend the alpaka accelerator with user information.

PIConGPU lockstep kernel (our own programming model to handle CPU accelerators with thread restrictions for the block and GPU accelerators would than look like this.

```C++
// fixed size blocking used for the kernel (virtual threads)
// On CUDA the kernel will be processed with 128 threads per block.
// On CPU with thread restriction the kernel will be processed with 1 thread per block.
// If there will be an accelerator that can only use 3 threads per block all will work automatically
// without the need to rewrite the kernel.
auto workerCfg = lockstep::makeWorkerCfg<128>();
// start 5 blocks to process
PMACC_LOCKSTEP_KERNEL([]ALPAKA_FN_ACC(auto const& worker) -> void{
    // domain/elements size is 123
    auto forEach = pmacc::lockstep::makeForEach<123>(worker);
    forEach([&](uint32_t const idx){
      printf("Hello World element=%u worker=%u.\n", idx, worker.getWorkerIdx());
    });
},workerCfg)(5)();
```

j-stephan:
I won't oppose a PR removing `final` from our Accs. It doesn't seem that we have a good reason for keeping it around so feel free to un-finalize.

psychocoderHPC:
> I won't oppose a PR removing `final` from our Accs. It doesn't seem that we have a good reason for keeping it around so feel free to un-finalize.

Yes, I will remove the final `AccGenericSycl` as suggested because it is wrong there but all others should be there to avoid someone inheriting from the accelerator.
In a separate PR I will forbid the copy and move constructor to ensure that states e.g. for shared memory will not be invalidated.

psychocoderHPC:
ready to be merged

SimeonEhrig (approved):

bernhardmgruber (approved):

psychocoderHPC:
@j-stephan ping

j-stephan (approved):

PR #1823, opened 2022-11-03
 Create a patch if clang-format CI fails

bernhardmgruber:
I was annoyed today that I did not have clang-format X.Y installed on my notebook when preparing a PR for LLAMA. So I just made the CI not only check clang-format but also commit the changes. Then I could easily pull and squash the formatting changes into my commits.

~So, here is the same for alpaka.~

Edit: So here is something similar. The clang-format CI will create a patch file and upload it as an artifact.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -89,6 +89,14 @@ jobs:
       with:
         clangFormatVersion: 14
         exclude: './thirdParty'
+        inplace: True
+    - run: |
+        git diff --exit-code > format.patch
+    - uses: actions/upload-artifact@v3
+      if: failure()
+      with:
+        name: format.patch
+        path: format.patch
 
   ci:
     needs: clang-format
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -91,6 +91,7 @@ jobs:
         exclude: './thirdParty'
 
   ci:
+    needs: clang-format
     name: ${{ matrix.name }}
     runs-on: ${{ matrix.os }}
     env: ${{ matrix.env }}
--- a/docs/source/dev/ci.rst
+++ b/docs/source/dev/ci.rst
@@ -8,6 +8,14 @@ GitHub Actions
 
 The configuration of ``GitHub Actions`` can be found in the ``.github/workflows/`` folder. This CI uses unmodified containers from ``Docker Hub`` and sets up the environment during the test job. A caching mechanism speeds up the job times. The scripts for setting up the environment, building alpaka and running test are located in the ``script/`` folder.
 
+clang-format
+++++++++++++
+
+The first CI job run is clang-format, which will verify the formatting of your changeset.
+Only of this check passes, will the remainder of the GitHub CI continue.
+In case of a formatting failure, a patch file is attached as an artifact to the GitHub action run.
+You can apply this patch file to your changeset to fix the formatting.
+
 GitLab CI
 ---------
 


SimeonEhrig:
Just for understanding, when is the commit of the reformatted code applied? Immediately, when the job is finish or when the PR is commit. I asked, because if it is immediately done, you need care, that you local branch is different to the remote.

bernhardmgruber:
> Just for understanding, when is the commit of the reformatted code applied? Immediately, when the job is finish or when the PR is commit.

The commit is done when the CI job finishes, so before merging and potentially while other CI jobs are running. That would also restart the CI.

> I asked, because if it is immediately done, you need care, that you local branch is different to the remote.

Yes, the CI commits to your branch, so you would need to pull the changes back to your local branch. I would recommend to squash the new commits then into your own.

SimeonEhrig:
Okay, that we should do a test run, with an ill formatted commit to check, what the GitLab CI is doing.

SimeonEhrig:
Okay, that we should do a test run, with an ill formatted commit to check, what the GitLab CI is doing.

bernhardmgruber:
Currently, the CI action checks the formatting. If it is not correct, it pushes a commit to correct it. Then the CI *fails*. I first thought that was weird. But then I realized, if you want to have a clean commit history, you want the PR author to integrate the clang-format commit in their commits and repush. So the behavior is actually alright. Question: what do you think about that? Should the CI continue/pass instead when a clang-format commit is produced?

bernhardmgruber:
> Okay, that we should do a test run, with an ill formatted commit to check, what the GitLab CI is doing.

I pushed a commit with bad formatting now.

bernhardmgruber:
So apparently, the CI clang-format action pushed a commit with the right formatting. The GitHub CI did *not* continue. The GitLab CI seems to have restarted. This is a bit weird maybe. What would you expect?

SimeonEhrig:
Actual I would expect, that is doing `clang-format`, squash it to the last commit and run all tests. For me, it feels not intuitive, that the CI fails, fixes the error and is waiting for me, that I say, do it again.

I would also distinguish between 2 use case. 
1. I push one commit without clang-format. It will apply clang-format, squash it and it looks like the same, what I would do manually.
2. I expect, that clang-format is only applied on the last commit. If not, it changes my discussion point. In this scenario, I push many commits, each without clang-format. In this case, clang-format is only applied on the last commit. Therefore I have two options. Ether I keep the history it is and risk, that the last commit contains line changes, which are only exists, because I changed it in a previous commit and didn't format it or I have to rework the whole history. But therefore, I need a local clang-format*.

* There is also the workaround to push one commit after each other, that each commit is formatted by CI, but this looks like a stupid idea compare a local installation of clang-format.

bernhardmgruber:
> Actual I would expect, that is doing `clang-format`, squash it to the last commit and run all tests.

This is totally not what I would expect :D If I push 4 commits, then I would be really annoyed with clang-format would just squash the diff into the last one. I cannot reuse that commit anymore.

> For me, it feels not intuitive, that the CI fails, fixes the error and is waiting for me, that I say, do it again.

Ok. So you would expect the CI to fail, fix the format, and then fully restart?

> 1. I push one commit without clang-format. It will apply clang-format, squash it and it looks like the same, what I would do manually.

This makes sense to me.

> 2. I expect, that clang-format is only applied on the last commit. If not, it changes my discussion point. In this scenario, I push many commits, each without clang-format. In this case, clang-format is only applied on the last commit. Therefore I have two options. Ether I keep the history it is and risk, that the last commit contains line changes, which are only exists, because I changed it in a previous commit and didn't format it or I have to rework the whole history. But therefore, I need a local clang-format*.

I am not sure I understand which strategy you are arguing for. You want to have only the last commit changed?

SimeonEhrig:
> > Actual I would expect, that is doing `clang-format`, squash it to the last commit and run all tests.
> 
> This is totally not what I would expect :D If I push 4 commits, then I would be really annoyed with clang-format would just squash the diff into the last one. I cannot reuse that commit anymore.
> 
> > For me, it feels not intuitive, that the CI fails, fixes the error and is waiting for me, that I say, do it again.
> 
> Ok. So you would expect the CI to fail, fix the format, and then fully restart?
> 
> > 1. I push one commit without clang-format. It will apply clang-format, squash it and it looks like the same, what I would do manually.
> 
> This makes sense to me.
> 
> > 2. I expect, that clang-format is only applied on the last commit. If not, it changes my discussion point. In this scenario, I push many commits, each without clang-format. In this case, clang-format is only applied on the last commit. Therefore I have two options. Ether I keep the history it is and risk, that the last commit contains line changes, which are only exists, because I changed it in a previous commit and didn't format it or I have to rework the whole history. But therefore, I need a local clang-format*.
> 
> I am not sure I understand which strategy you are arguing for. You want to have only the last commit changed?

Okay, maybe I was not clear. It would be great, if each commit is separately formatted with `clang-format` and the new commit with format changes is squashed with the original commit, so that it looks like, that each commit was pushed with the correct code style and then the CI test should be executed.

But I'm not sure, if this how your job is working. I expected, that it runs clang-format on the last commit and it is fine.

bernhardmgruber:
> Okay, maybe I was not clear. It would be great, if each commit is separately formatted with clang-format and the new commit with format changes is squashed with the original commit, so that it looks like, that each commit was pushed with the correct code style and then the CI test should be executed.

I googled quickly and it seems there is no GH action available that can format each individual commit and push again. I also do not want to implement that manually.

> But I'm not sure, if this how your job is working. I expected, that it runs clang-format on the last commit and it is fine.

The current GH action runs clang-format and adds an additional commit with the formatting changes. And those should retrigger the CI (which fails for GH, but works for GL).

SimeonEhrig:
> The current GH action runs clang-format and adds an additional commit with the formatting changes. And those should retrigger the CI (which fails for GH, but works for GL).

But then, you need to manipulate the branch anyway, because we squash anything to one commit (with exceptions). So it means you have two options:
1. install clang-format on your system and push the formated commit
2. push unformated commit, pull the branch, squash and force push it again

Or am I wrong?

bernhardmgruber:
> But then, you need to manipulate the branch anyway, because we squash anything to one commit (with exceptions). So it means you have two options:
> 
> 1. install clang-format on your system and push the formated commit
> 2. push unformated commit, pull the branch, squash and force push it again
> 
> Or am I wrong?

No, you are right! And for most PRs that are just one commit, option 2 is nice because you can get the right formatting without having (the right) clang-format installed! This is the use case I am targeting.

bernhardmgruber:
Btw, it turns out that GH actions pushing commits cannot trigger new workflows when the authentication happened via the `GITHUB_TOKEN`. See: https://docs.github.com/en/actions/using-workflows/triggering-a-workflow

> When you use the repository's GITHUB_TOKEN to perform tasks, events triggered by the GITHUB_TOKEN, with the exception of workflow_dispatch and repository_dispatch, will not create a new workflow run. This prevents you from accidentally creating recursive workflow runs. For example, if a workflow run pushes code using the repository's GITHUB_TOKEN, a new workflow will not run even when the repository contains a workflow configured to run when push events occur.

In order for the CI to start running again after the commit, you would need a personal access token. I am trying this in LLAMA currently. However, I start to wonder now whether it is actually fine if the CI does not restart. Because, to maintain a good commit history, a user should either fix the formatting and force push, or pull, squash the formatting commit, and force push. Both ways, it is not fine if we merge the PR with a separate formatting commit. So it actually saves resources to not run the CI again after pushing a formatting commit.

Maybe let's discuss this in the next alpaka VC.

bernhardmgruber:
We discusses this in the alpaka VC today. We concluded that:

*  We should not format each commit individually, because that can fail with conflicts when rebasing the individual commits again. It's hard to get that right.
* Adding a new commit may be inconvenient for inexperienced git users because they need to know how to resolve that locally in git, in case they also added more changes locally.
* It is also weird that the CI would not restart after a formatting commit is created.
* It would be beneficial to have the formatting diff available as a patch file to download, if this is easily possible.
* Other CI runs should depend on the clang-format CI run and not start if the formatting failed.

bernhardmgruber:
If the CI fails now, a patch file will be uploaded as artifact:
image

bernhardmgruber:
And if the clang-format succeeds, the CI will continue as normally.

SimeonEhrig:
That workflow sounds good. Could we document it? Otherwise, it could be hard to find the patch file. Maybe adding a section here: https://alpaka.readthedocs.io/en/latest/dev/ci.html

bernhardmgruber:
> That workflow sounds good. Could we document it? Otherwise, it could be hard to find the patch file. Maybe adding a section here: https://alpaka.readthedocs.io/en/latest/dev/ci.html

I added a bit of documentation.

SimeonEhrig (approved):

PR #1826, opened 2022-11-04
fix warning calling `__host__` from `__host__ __device__` function

psychocoderHPC:
Host-only objects were annotated with the host device function specifier. Therefore warnings were shown e.g.

```
alpaka/include/alpaka/mem/view/ViewConst.hpp(93): warning #20011-D: calling a __host__ function(
"const    ::std::remove_volatile< ::alpaka::trait::ElemType<T1, void> ::type> ::type * 
alpaka::getPtrNative< ::alpaka::BufUniformCudaHipRt< ::alpaka::ApiCudaRt, float,     ::std::integral_constant<unsigned long, (unsigned long)1ul> ,
 long> > (const T1 &)") 
from a __host__ __device__ 
function("alpaka::trait::GetPtrNative< ::alpaka::ViewConst< ::alpaka::BufUniformCudaHipRt< 
::alpaka::ApiCudaRt, float,     ::std::integral_constant<unsigned long, (unsigned long)1ul> , long> > , void> ::getPtrNative") is not allowed
```

Remove the usage of `ALPAKA_NO_HOST_ACC_WARNING` this is a workaround that should only be used if there is no other way to fix these warnings.

Warnings were shown with nvcc, I do know why the CI is passing with these warnings.
```
Copyright (c) 2005-2022 NVIDIA Corporation
Built on Tue_Mar__8_18:18:20_PST_2022
Cuda compilation tools, release 11.6, V11.6.124
Build cuda_11.6.r11.6/compiler.31057947_0
```

### DIFF
--- a/include/alpaka/mem/view/Accessor.hpp
+++ b/include/alpaka/mem/view/Accessor.hpp
@@ -58,7 +58,7 @@ namespace alpaka::experimental
         struct BuildAccessor
         {
             template<typename... TAccessModes, typename TMemoryObjectForwardRef>
-            ALPAKA_FN_HOST_ACC static auto buildAccessor(TMemoryObjectForwardRef&&)
+            ALPAKA_FN_HOST static auto buildAccessor(TMemoryObjectForwardRef&&)
             {
                 static_assert(
                     meta::DependentFalseType<TMemoryObject>::value,
@@ -104,7 +104,7 @@ namespace alpaka::experimental
         typename... TAccessModes,
         typename TMemoryObject,
         typename = std::enable_if_t<!internal::IsAccessor<std::decay_t<TMemoryObject>>::value>>
-    ALPAKA_FN_HOST_ACC auto accessWith(TMemoryObject&& memoryObject)
+    ALPAKA_FN_HOST auto accessWith(TMemoryObject&& memoryObject)
     {
         return trait::BuildAccessor<std::decay_t<TMemoryObject>>::template buildAccessor<TAccessModes...>(
             memoryObject);
@@ -119,7 +119,7 @@ namespace alpaka::experimental
         typename TBufferIdx,
         std::size_t TDim,
         typename... TPrevAccessModes>
-    ALPAKA_FN_HOST_ACC auto accessWith(
+    ALPAKA_FN_HOST auto accessWith(
         Accessor<TMemoryHandle, TElem, TBufferIdx, TDim, std::tuple<TPrevAccessModes...>> const& acc)
     {
         static_assert(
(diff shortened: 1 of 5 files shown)


include/alpaka/mem/view/Accessor.hpp
```diff
@@ -131,28 +131,28 @@ namespace alpaka::experimental
     //! Constrains an existing accessor to the specified access modes.
     // constraining accessor to the same access mode again just passes through
     template<typename TNewAccessMode, typename TMemoryHandle, typename TElem, typename TBufferIdx, std::size_t TDim>
-    ALPAKA_FN_HOST_ACC auto accessWith(const Accessor<TMemoryHandle, TElem, TBufferIdx, TDim, TNewAccessMode>& acc)
+    ALPAKA_FN_HOST auto accessWith(const Accessor<TMemoryHandle, TElem, TBufferIdx, TDim, TNewAccessMode>& acc)
     {
         return acc;
     }
```
bernhardmgruber: This is a breaking change, because we could constrain an accessor in-kernel before. That is `accessWith` is allowed to be called inside a kernel.

include/alpaka/mem/view/ViewConst.hpp
```diff
@@ -31,11 +31,11 @@ namespace alpaka
             "This is not implemented"); // It might even be dangerous for ViewConst to store a reference to the wrapped
                                         // view, as this decouples the wrapped view's lifetime.
 
-        ALPAKA_FN_HOST_ACC ViewConst(const TView& view) : m_view(view)
+        ALPAKA_FN_HOST ViewConst(const TView& view) : m_view(view)
         {
         }
 
-        ALPAKA_FN_HOST_ACC ViewConst(TView&& view) : m_view(std::move(view))
+        ALPAKA_FN_HOST ViewConst(TView&& view) : m_view(std::move(view))
```
bernhardmgruber: Isn't this preventing to make a `ViewConst` inside the kernel? E.g. from a `ViewPlainPtr`.

bernhardmgruber (changes requested):
I think this PR is restricting several valid use cases, so we first need to discuss whether we disallow such use cases now.

include/alpaka/mem/view/Accessor.hpp
psychocoderHPC: I can revert it that's no problem as long as alpaka is not throwing the warning.

include/alpaka/mem/view/ViewConst.hpp
psychocoderHPC: As I know we never supported the usage of views within a kernel. All buffer, Views are host objects.

include/alpaka/mem/view/Accessor.hpp
psychocoderHPC: Mhh, `accessWith` is the reason for the warning. :-( It is not easy possible to use `ALPAKA_FN_HOST_ACC`

include/alpaka/mem/view/ViewConst.hpp
bernhardmgruber: Buffers are definitely host objects. But `std::array` as well? Or a `T*` and a `size_t`?

Anyway, I am fine with the above changes. But they add restrictions.

include/alpaka/mem/view/Accessor.hpp
bernhardmgruber: You can go ahead with your changes. I am hoping to kill those accessors at some point with `mdspan` and pointer support in the SYCL backend.

include/alpaka/mem/view/ViewConst.hpp
psychocoderHPC: Yes `std::array` is a host-only object because you can only use the `constexpr` methods because runtime methods are not annotated with `__device__`.
`T*` does not require any annotations because it is not an object with operators.

include/alpaka/mem/view/Accessor.hpp
psychocoderHPC: No problem postponing these changes a week more and discussing these.
I saw this warning and thought I need to fix it now or I will forget it until I will see the warning in a half year again.

psychocoderHPC:
For documentation:

Years ago we analyzed the behavior of `ALPAKA_NO_HOST_ACC_WARNING` because there is no documentation about this attribute available.

https://github.com/ComputationalRadiationPhysics/picongpu/pull/1707

include/alpaka/mem/view/ViewConst.hpp
bernhardmgruber: I was thinking about:
```c++
T* data = ...;
auto v = alpaka::createView(data, 10);
auto cv = alpaka::ViewConst{v};
```
inside kernel code. However, I just checked the API and we require to pass a device into `createView` and AFAIK devices are not allowed in kernel code. Therefore, you can proceed!

bernhardmgruber (dismissed):

psychocoderHPC:
@bernhardmgruber  This PR can is ready to be merged

bernhardmgruber (approved):

PR #1828, opened 2022-11-07
add math hyperbolic functions

### LINKED ISSUE #1827: No Hyperbolic Functions?
Does Alpaka have implementations of hyperbolic functions (cosh, sinh, acosh, asinh, etc.)? If not, are there any plans to support them or advice on how to get around this? I know CUDA has hyperbolic functions in their math library for example. My thought right now is to implement my own device function that calculates acosh(x) (the function I need) using the other alpaka math functions, but this probably isn't the best approach given that CUDA and others have direct implementations of these hyperbolic functions.

Thank you!

psychocoderHPC:
solve #1827

Add math functions sinh, cosh, acosh, asinh, tanh, atanh and the coresponding unit tests.

### DIFF
--- a/include/alpaka/math/MathGenericSycl.hpp
+++ b/include/alpaka/math/MathGenericSycl.hpp
@@ -32,6 +32,11 @@ namespace alpaka::experimental::math
     {
     };
 
+    //! The SYCL acosh.
+    class AcoshGenericSycl : public concepts::Implements<alpaka::math::ConceptMathAcosh, AcoshGenericSycl>
+    {
+    };
+
     //! The SYCL arg.
     class ArgGenericSycl : public concepts::Implements<alpaka::math::ConceptMathArg, ArgGenericSycl>
     {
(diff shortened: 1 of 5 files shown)


psychocoderHPC:
@jkelling Could you please check why OpenACC is not working https://github.com/alpaka-group/alpaka/actions/runs/3410869526/jobs/5674362614

I assume one of the new math functions is for an unknown reason not available in OpenACC.

bernhardmgruber:
if you spell it `Fixes: #1827` in your commit message/PR description, then Github will automatically link the corresponding issue.

bernhardmgruber (dismissed):

jkelling:
> I assume one of the new math functions is for an unknown reason not available in OpenACC.

I think the problem depends on the compiler version: The failing test uses NVHPC 21.9. There is an OpenACC test in the gitlab CI using 21.11 running on a CUDA GPU [which succeeds](https://gitlab.com/hzdr/crp/alpaka/-/jobs/3284547218). It also works for me locally with 22.1.

I suggest to bump up the nvhpc version for this check. As this PR trips up 21.9, this should be the right place to make this change. The check is defined here:
https://github.com/alpaka-group/alpaka/blob/1111fd083c3b55d23ee918c6df47041726a05013/.github/workflows/ci.yml#L260-L264
You could replace the image by `22.9-runtime-cuda11.7-ubuntu20.04`, but at the risk, of something else being broken there. I did test, 22.1, so `22.1-runtime-cuda11.5-ubuntu20.04` may be a safer bet.

psychocoderHPC:
> if you spell it `Fixes: #1827` in your commit message/PR description, then Github will automatically link the corresponding issue.

solve #XYZ should work too or was it close?

bernhardmgruber:
> solve #XYZ should work too or was it close?

Does not seem that way. GitHub did not automatically link the issue. I had to link it manually.

psychocoderHPC:
can be merged

SimeonEhrig (approved):

PR #1830, opened 2022-11-07
Refactor subDivideGridElems and test getValidWorkDiv with Idx type directly

bernhardmgruber:
Depends on:

- [x] Merge #1829 before.
- [x] Merge #1837 before.

### DIFF
--- a/include/alpaka/core/Utility.hpp
+++ b/include/alpaka/core/Utility.hpp
@@ -29,4 +29,41 @@ namespace alpaka::core
 #else
     using std::declval;
 #endif
+
+    /// Returns the ceiling of a / b, as integer.
+    template<typename Integral>
+    [[nodiscard]] ALPAKA_FN_HOST_ACC constexpr auto divCeil(Integral a, Integral b) -> Integral
+    {
+        return (a + b - 1) / b;
+    }
+
+    /// Computes the nth power of base, in integers.
+    template<typename Integral>
+    [[nodiscard]] ALPAKA_FN_HOST_ACC constexpr auto intPow(Integral base, Integral n) -> Integral
+    {
+        if(n == 0)
+            return 1;
+        auto r = base;
+        for(Integral i = 1; i < n; i++)
+            r *= base;
+        return r;
+    }
+
+    /// Computes the floor of the nth root of value, in integers.
+    template<typename Integral>
+    [[nodiscard]] ALPAKA_FN_HOST_ACC constexpr auto nthRootFloor(Integral value, Integral n) -> Integral
+    {
+        // adapted from: https://en.wikipedia.org/wiki/Integer_square_root
+        Integral L = 0;
+        Integral R = value + 1;
+        while(L != R - 1)
+        {
+            Integral const M = (L + R) / 2;
+            if(intPow(M, n) <= value)
+                L = M;
+            else
+                R = M;
+        }
+        return L;
+    }
 } // namespace alpaka::core
--- a/include/alpaka/workdiv/WorkDivHelpers.hpp
+++ b/include/alpaka/workdiv/WorkDivHelpers.hpp
@@ -12,6 +12,7 @@
 #include <alpaka/acc/Traits.hpp>
 #include <alpaka/core/Assert.hpp>
 #include <alpaka/core/Common.hpp>
+#include <alpaka/core/Utility.hpp>
 #include <alpaka/dev/Traits.hpp>
 #include <alpaka/vec/Vec.hpp>
 #include <alpaka/workdiv/WorkDivMembers.hpp>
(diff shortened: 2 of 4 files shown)


include/alpaka/core/Utility.hpp
```diff
@@ -29,4 +29,41 @@ namespace alpaka::core
...
+        while(L != R - 1)
+        {
+            Integral const M = (L + R) / 2;
+            if(intPow(M, n) <= value)
+                L = M;
+            else
+                R = M;
+        }
+        return L;
+    }
```
bernhardmgruber, line 68: In case someone wonders whether that is slower than `std::pow(static_cast<double>(value), 1.0 / static_cast<double>(n))`, it is actually 11x faster: https://quick-bench.com/q/xIDwFmTdzQVETk1jUwkpyvuoEao

bernhardmgruber, line 68: In case someone wonders whether that is slower than `std::pow(static_cast<double>(value), 1.0 / static_cast<double>(n))`, it is actually 11x faster: https://quick-bench.com/q/xIDwFmTdzQVETk1jUwkpyvuoEao

psychocoderHPC:
I will have a look into it on Monday again, this is nothing for a Friday.

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -147,177 +150,117 @@ namespace alpaka
...
-                                }
-                                else
-                                {
-                                    return a < b;
-                                }
-                            })));
-                    blockThreadExtent[minElemIdx] = blockThreadExtent[minElemIdx] / static_cast<TIdx>(2u);
+                    auto const it = std::min_element(
+                        blockThreadExtent.begin(),
+                        blockThreadExtent.end() - 1, // TODO(bgruber): why omit the last element?
```
psychocoderHPC: I do not think that this is correct, the documentation of `std::min_elemet` is saying the range is `Finds the smallest element in the range [first, last).`

psychocoderHPC (changes requested):
looks good except for the line with the TODO.

include/alpaka/workdiv/WorkDivHelpers.hpp
bernhardmgruber: I also think it's incorrect, but the previous code did:
```c++
std::min_element(
                            &blockThreadExtent[0u],
                            &blockThreadExtent[TDim::value - 1u],
                            ...
```
Which takes the address of the last element as the end of the range passed to `min_element`. And this is `.end() - 1`.

bernhardmgruber: Anyway, I added a new commit to also include the last element here. I had to update some of the reference values in the unit tests as a result. Just have a look at the last commit in isolation.

psychocoderHPC (approved):

test/unit/workDiv/src/WorkDivHelpersTest.cpp
```diff
@@ -86,7 +86,7 @@ TEMPLATE_LIST_TEST_CASE("subDivideGridElems.2D.examples", "[workDiv]", alpaka::t
                 props,
                 false,
                 alpaka::GridBlockExtentSubDivRestrictions::Unrestricted)
-            == WorkDiv{Vec{75, 5}, Vec{4, 128}, Vec{1, 1}});
+            == WorkDiv{Vec{2, 300}, Vec{256, 2}, Vec{1, 1}});
```
psychocoderHPC: For the block size, the fast-moving index X should be the larger value. The thread indices are mostly used to access the memory directly. Starting a kernel with 2 threads in x-dimension and 256 in Y-dimension will result slow performance.

test/unit/workDiv/src/WorkDivHelpersTest.cpp
```diff
@@ -111,7 +111,7 @@ TEMPLATE_LIST_TEST_CASE("subDivideGridElems.2D.examples", "[workDiv]", alpaka::t
                 props,
                 true,
                 alpaka::GridBlockExtentSubDivRestrictions::Unrestricted)
-            == WorkDiv{Vec{75, 5}, Vec{4, 120}, Vec{1, 1}});
+            == WorkDiv{Vec{2, 300}, Vec{150, 2}, Vec{1, 1}});
```
psychocoderHPC: Same here the result the block extents look wrong.

psychocoderHPC (changes requested):
I think I need a VC to understand why the results changed. The new workdiv created looks worse with in mind where this helper is used.
Typically users use this helper to fit threads onto there problem size and the problem is mapped one-to-one to memory.

test/unit/workDiv/src/WorkDivHelpersTest.cpp
bernhardmgruber: > I think I need a VC to understand why the results changed. 

The results changed because you told me to change the algorithm up here: https://github.com/alpaka-group/alpaka/pull/1830#discussion_r1033377350. If the new results look bad, then we just change the algorithm back to the old behavior. It is not my intend to change behavior with this PR.

So, revert the last commit again?

include/alpaka/workdiv/WorkDivHelpers.hpp
bernhardmgruber: I removed that commit again because it produced wrong test results. I also changed the comment into a doxygen comment by request from @psychocoderHPC.

test/unit/workDiv/src/WorkDivHelpersTest.cpp
bernhardmgruber: Last commit reverted after offline discussion.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #1843, opened 2022-11-17
Select serial accelerator for tests/examples

### LINKED ISSUE #1836: Possible to end up with no backend configured (CMake)
By running CMake twice:
```
cmake ../alpaka/ -DCMAKE_BUILD_TYPE=Release
cmake ../alpaka/ -Dalpaka_BUILD_EXAMPLES=ON -DCMAKE_BUILD_TYPE=Release
```
I managed to end up with no backend configured, and a lot of (initially) puzzling compiler errors:
```shell
In file included from alpaka/example/randomStrategies/src/randomStrategies.cpp:11:
alpaka/include/alpaka/example/ExampleDefaultAcc.hpp:60:6: warning: #warning "No supported backend selected." [-Wcpp]
   60 | #    warning "No supported backend selected."
```
```
In file included from alpaka/include/alpaka/acc/Traits.hpp:14,
                 from alpaka/include/alpaka/alpaka.hpp:35,
                 from alpaka/example/randomStrategies/src/randomStrategies.cpp:10:
alpaka/include/alpaka/core/Concepts.hpp: In instantiation of 'constexpr const auto alpaka::concepts::ImplementsConcept<alpaka::ConceptAcc, alpaka::ExampleDefaultAcc<std::integral_constant<long unsigned int, 1>, long unsigned int> >::
alpaka/include/alpaka/queue/Traits.hpp:69:11:   recursively required by substitution of 'template<class TAcc, class TProperty> struct alpaka::trait::QueueType<TAcc, TProperty, typename std::enable_if<alpaka::concepts::ImplementsConce
alpaka/include/alpaka/queue/Traits.hpp:69:11:   required by substitution of 'template<class TEnv, class TProperty> using Queue = typename alpaka::trait::QueueType<TEnv, TProperty>::type [with TEnv = alpaka::ExampleDefaultAcc<std::int
alpaka/example/randomStrategies/src/randomStrategies.cpp:42:54:   required from here
alpaka/include/alpaka/core/Concepts.hpp:31:58: error: invalid use of incomplete type 'class alpaka::ExampleDefaultAcc<std::integral_constant<long unsigned int, 1>, long unsigned int>'
   31 |         static constexpr auto value = decltype(implements(std::declval<TDerived&>()))::value;
```

I see that the serial backend is enabled if unset (for builds/tests):
```cmake
# the sequential accelerator is required for the tests and examples
if(NOT DEFINED alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE AND (alpaka_BUILD_EXAMPLES OR BUILD_TESTING))
  option(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE "enable alpaka serial accelerator" ON)
endif()
```
https://github.com/alpaka-group/alpaka/blob/develop/CMakeLists.txt#L83

so perhaps the solution is to change this to:
```cmake
# the sequential accelerator is required for the tests and examples
if ((alpaka_BUILD_EXAMPLES OR BUILD_TESTING))
  message("Sequential accelerator is required for the tests and examples")
  option(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE "enable alpaka serial accelerator" ON)
endif()
```

but perhaps there is a usecase I am missing.

Best regards,

Stewart

StewMH:
For #1836 - fail if no backend is selected.


CMakeLists.txt
```diff
@@ -81,10 +81,13 @@ set(_alpaka_INCLUDE_DIRECTORY "${_alpaka_ROOT_DIR}/include")
 set(_alpaka_SUFFIXED_INCLUDE_DIR "${_alpaka_INCLUDE_DIRECTORY}/alpaka")
 
 # the sequential accelerator is required for the tests and examples
-if(NOT DEFINED alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE AND (alpaka_BUILD_EXAMPLES OR BUILD_TESTING))
-  option(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE "enable alpaka serial accelerator" ON)
+if(alpaka_BUILD_EXAMPLES OR BUILD_TESTING)
+  if (NOT DEFINED alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE)
+    option(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE "enable alpaka serial accelerator" ON)
+  elseif(NOT alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE)
+    message(FATAL_ERROR "Serial backend must be enabled for tests and examples: delete CMakeCache.txt and rerun CMake command")
```
bernhardmgruber: AFAIK this is not necessary. I can just turn on the serial backend via `ccmake` or `cmake-gui` and reconfigure.
```suggestion
    message(FATAL_ERROR "Serial backend must be enabled for tests and examples.")
```

CMakeLists.txt
```diff
@@ -81,10 +81,13 @@ set(_alpaka_INCLUDE_DIRECTORY "${_alpaka_ROOT_DIR}/include")
...
+  endif()
```
bernhardmgruber: If the serial backend is truly necessary for tests and examples, then we should just turn it on and tell the user instead of fatally erroring. Something like (please check if it works):
```cmake
if((NOT alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE  AND (alpaka_BUILD_EXAMPLES OR BUILD_TESTING))
  option(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE "enable alpaka serial accelerator" ON)
  message(INFO "Turned on serial backend which is required to build examples or tests")
endif()
```
But I would like to have opinions from other maintainers, @psychocoderHPC or @j-stephan.

StewMH: The problem here is that once an option is set (i.e. in the first run of cmake with different options) it cannot be overridden:
```
If <variable> is already set as a normal or cache variable, then the command does nothing (see policy [CMP0077](https://cmake.org/cmake/help/latest/policy/CMP0077.html#policy:CMP0077)).
```
https://cmake.org/cmake/help/latest/command/option.html
so it doesn't complain about the line, but it doesn't change the value.

CMakeLists.txt
StewMH: Not familiar with cmake-gui, but certainly with cmake command line it will not overwrite the already set variable (see also my other comment).

psychocoderHPC: @StewMH ONe thing we should test before giving up with overwriting the variable: https://stackoverflow.com/a/39592445

```CMake
set(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE ON CACHE BOOL "enable alpaka serial accelerator" FORCE)
```

If stack overflow is correct this should work.

StewMH: This seems to work, thanks!

psychocoderHPC:
@StewMH Could you please again rebase this PR against the latest development branch? We had some trouble with the CI which is now solved.

psychocoderHPC:
Bad news: Few CI tests checking CUDA_ONLY compilation and therefore it is not allowed to enable a second backend.
IMO we need to handle `alpaka_ACC_GPU_CUDA_ONLY_MODE` and `alpaka_ACC_GPU_HIP_ONLY_MODE` and in this case, throw the warning/error we had in this PR before we explicitly activated the serial backend. :-(

```
-- Detecting CXX compile features - done
CMake Error at cmake/alpakaCommon.cmake:103 (message):
  If alpaka_ACC_GPU_CUDA_ONLY_MODE is enabled, only back-ends using CUDA can
  be enabled! This allows to mix alpaka code with native CUDA code.  However,
  this prevents any non-CUDA back-ends from being enabled.
  CMakeLists.txt:91 (include)
See also "D:/a/alpaka/alpaka/build/CMakeFiles/CMakeOutput.log".
```

StewMH:
OK, I think I handled this case.

CMakeLists.txt
```diff
@@ -81,10 +81,17 @@ set(_alpaka_INCLUDE_DIRECTORY "${_alpaka_ROOT_DIR}/include")
 set(_alpaka_SUFFIXED_INCLUDE_DIR "${_alpaka_INCLUDE_DIRECTORY}/alpaka")
 
 # the sequential accelerator is required for the tests and examples
-if(NOT DEFINED alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE AND (alpaka_BUILD_EXAMPLES OR BUILD_TESTING))
-  option(alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE "enable alpaka serial accelerator" ON)
+if(alpaka_BUILD_EXAMPLES OR BUILD_TESTING)
+  if (NOT alpaka_ACC_GPU_CUDA_ONLY_MODE OR alpaka_ACC_GPU_HIP_ONLY_MODE)
```
psychocoderHPC: I am not 100% sure but I would say this line should be

```
if (NOT (alpaka_ACC_GPU_CUDA_ONLY_MODE OR alpaka_ACC_GPU_HIP_ONLY_MODE) )
```

In your current version CMake would enable the serial accelerator if `alpaka_ACC_GPU_HIP_ONLY_MODE` is true because as I know OR has not have a higher priority than NOT.
Please correct me if I am wrong!

psychocoderHPC (dismissed):

CMakeLists.txt
StewMH: Thanks - I confirmed locally that you're correct, and either way I always prefer brackets over remembering operator precedence.

bernhardmgruber:
@StewMH, btw, if you word your PR description to contain: "Fixes: #1836", Github will automatically link the corresponding issue to this PR. I linked the issue manually now.

bernhardmgruber (dismissed):

StewMH:
Thanks @bernhardmgruber, I am not too familiar with GitHub (usually using CERN gitlab).

StewMH:
@psychocoderHPC I think I addressed your proposed change but I just wrote it myself rather than accepting your changes. I think this is why the merge is blocked. Once again I blame my inexperience with GitHub.

bernhardmgruber:
> @psychocoderHPC I think I addressed your proposed change but I just wrote it myself rather than accepting your changes. I think this is why the merge is blocked. Once again I blame my inexperience with GitHub.

The merge is blocked because @psychocoderHPC requested changes during his last review. So he needs to review again and approve or request changes again. Independently, we had an issue with the CI, which was fixed today with #1853. Since you do not have that commit yet, I would like to ask you to rebase your changes onto the current `develop` branch.

bernhardmgruber (approved):

bernhardmgruber:
@psychocoderHPC this PR is blocked because you requested changes. Have those been addressed?

j-stephan:
Hello @StewMH, thanks for the PR. From my POV it looks good. Could you please squash your commits? Afterwards I will merge.

bernhardmgruber:
> Hello @StewMH, thanks for the PR. From my POV it looks good. Could you please squash your commits? Afterwards I will merge.

@j-stephan you can also squash them!

j-stephan:
Thanks. I didn't see the option right away on the mobile app.

bernhardmgruber:
> Thanks. I didn't see the option right away on the mobile app.

You merge on the GitHub mobile app? :D

PR #1846, opened 2022-11-29
Add alpaka port of babelstream

bernhardmgruber:
Since I may need this myself for LLAMA, here is a quick port of [BabelStream ](https://github.com/UoB-HPC/BabelStream) to alpaka. This work is based on the cupla port from Jeff Young here: https://github.com/jyoung3131/BabelStream.

I took the benchmark driver (`main.cpp` and `Stream.h`) verbatim from BabelStream, which seems to be permitted by their license. This should also allow us to contribute `AlpakaStream.h/.cpp` back to BabelStream.

No performance tests done. We should at least try to see whether we can reproduce one benchmark from the paper "Evaluating GPU Programming Models for the LUMI Supercomputer". Just to make sure I did not introduce a bug while porting.

And I may need help on how to correctly attribute all authors. I added the `LICENSE` from BabelStream into the source folder and left the copyright headers untouched. I added two lines stating that Jeff wrote the cupla port and I did the alpaka port.

Contains PR #1853 to fix the CUDA CI.


bernhardmgruber:
Running a small version of this benchmark on my old notebook inside WSL:
```bash
bgruber@dixxi3:/mnt/c/dev/alpaka/build_gcc$ ./example/babelstream/babelstream -s 10240 -n 2
BabelStream
Version: 4.0
Implementation: alpaka
Running kernels 2 times
Precision: double
Array size: 0.1 MB (=0.0 GB)
Total size: 0.2 MB (=0.0 GB)
Using alpaka device Intel(R) Core(TM) i7-4710HQ CPU @ 2.50GHz
Function    MBytes/sec  Min (sec)   Max         Average
Copy        6.602       0.02482     0.02482     0.02482
Mul         6.429       0.02549     0.02549     0.02549
Add         9.545       0.02575     0.02575     0.02575
Triad       8.277       0.02969     0.02969     0.02969
Dot         0.039       4.15394     4.15394     4.15394
```

example/babelstream/src/AlpakaStream.cpp
```diff
@@ -0,0 +1,250 @@
+// Copyright (c) 2015-16 Tom Deakin, Simon McIntosh-Smith,
+// University of Bristol HPC
+//
+// For full license terms please see the LICENSE file distributed with this
+// source code
+//
+// Cupla version created by Jeff Young in 2021
+// Ported from cupla to alpaka by Bernhard Manfred Gruber in 2022
```
bernhardmgruber, line 8: So, the first part was there already. I added the last two lines.

example/babelstream/src/AlpakaStream.cpp
```diff
@@ -0,0 +1,250 @@
...
+        initA,
+        initB,
+        initC);
+    alpaka::wait(queue);
+}
+
+template<typename T>
+void AlpakaStream<T>::read_arrays(std::vector<T>& a, std::vector<T>& b, std::vector<T>& c)
+{
+    alpaka::memcpy(queue, alpaka::createView(devHost, a, static_cast<Idx>(a.size())), d_a);
```
bernhardmgruber: I was expecting `alpaka::createView(devHost, a)` to work as well, since `a` is a container, but it failed to compile. Should this work?

example/babelstream/src/AlpakaStream.cpp
```diff
@@ -0,0 +1,250 @@
...
+    alpaka::wait(queue);
+}
+
+struct DotKernel
+{
+    template<typename TAcc, typename T>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, T const* a, T const* b, T* sum, int const arraySize) const
+    {
+        // TODO - test if sharedMem bug is affecting performance here
+        auto& tb_sum = alpaka::declareSharedVar<T[TBSIZE], __COUNTER__>(acc);
```
bernhardmgruber, line 194: That comment was already there.

example/babelstream/src/main.cpp
```diff
@@ -0,0 +1,632 @@
...
+#elif defined(ACC)
+#include "ACCStream.h"
+#elif defined(SYCL)
+#include "SYCLStream.h"
+#elif defined(SYCL2020)
+#include "SYCLStream2020.h"
+#elif defined(OMP)
+#include "OMPStream.h"
+#elif defined(ALPAKA)
+#include "AlpakaStream.h"
```
bernhardmgruber: Added.

example/babelstream/src/main.cpp
```diff
@@ -0,0 +1,632 @@
...
+  stream = new SYCLStream<T>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(OMP)
+  // Use the OpenMP implementation
+  stream = new OMPStream<T>(ARRAY_SIZE, deviceIndex);
+
+#elif defined(ALPAKA)
+  // Use the alpaka implementation
+  stream = new AlpakaStream<T>(ARRAY_SIZE, deviceIndex);
+
```
bernhardmgruber: Added.

example/babelstream/src/AlpakaStream.cpp
```diff
@@ -0,0 +1,250 @@
...
+//
+// For full license terms please see the LICENSE file distributed with this
+// source code
+//
+// Cupla version created by Jeff Young in 2021
+// Ported from cupla to alpaka by Bernhard Manfred Gruber in 2022
+
+#include "AlpakaStream.h"
+
+constexpr auto TBSIZE = 1024;
```
bernhardmgruber, line 14: Block size is hardcoded here, we should make this dynamic maybe.

example/babelstream/src/AlpakaStream.cpp
psychocoderHPC: In general, it should work. I think this is a BUG!

bernhardmgruber: Hmm, I had a look and it turns out that this is not a bug. `createview(dev, cont)` requires the `cont` to specialize alpaka extents. I guess what we would need is to provide specializations of `alpaka::GetExtents` and maybe also `alpaka::Dim` for standard containers.

bernhardmgruber: Which are already implemented! :)

bernhardmgruber:
I tried to let alpaka compute the workdiv for the kernels, but that does not work for the `dot` kernel, because it is written with a specific block/SMEM size in mind. I therefore removed the benchmark from the cmake tests, so the CI does not run it. The run will fail on all backends which do not support block sizes larger than 1.

bernhardmgruber:
I checked how it is done in the Kokkos version of BabelStream. They just let Kokkos compute the workdiv. This works in their case, because Kokkos has reduction algorithms, so they can implement the `dot` kernel. We could do this with vikunja at some point or add port a small reduction implementation here ourselves. But both options are out of scope for this PR.

jyoung3131:
Very nice work, @bernhardmgruber! One of my todos has been to do a PR for the Cupla version to the "mainline" BabelStream repo. 

Would you be ok with me including your Alpaka version (with your updated `LICENSE` file) to this PR once it gets merged here?

bernhardmgruber:
> Very nice work, @bernhardmgruber!

Thank you!

> One of my todos has been to do a PR for the Cupla version to the "mainline" BabelStream repo.
> 
> Would you be ok with me including your Alpaka version (with your updated `LICENSE` file) to this PR once it gets merged here?

Sure! However, wouldn't just the alpaka version suffice? Cupla is a tool to make it easier to migrate to alpaka, but the final goal should always be to have an alpaka version. But if they accept both versions, then who am I to object? @psychocoderHPC do you have any thoughts on this?

example/babelstream/src/AlpakaStream.cpp
bernhardmgruber: With #1847, this works now.

bernhardmgruber:
I rebased on top of #1853 to check whether the CI would run through now.

example/babelstream/src/AlpakaStream.cpp
SimeonEhrig: In vikunja, the block size is also [hard coded](https://github.com/alpaka-group/vikunja/blob/d6cbf89d783cde91fe785eec4192142244c4e0e8/include/vikunja/reduce/reduce.hpp#L120) for the moment. For example [CUDA use 256](https://github.com/alpaka-group/vikunja/blob/d6cbf89d783cde91fe785eec4192142244c4e0e8/include/vikunja/workdiv/BlockBasedWorkDiv.hpp#L84). I suggest to test one run with a smaller block size. If you get a speed up, think about a possibility to set the block size via compile time parameter.

SimeonEhrig:
@bernhardmgruber Can you please add a small `README.md` to the example. The README should point to the original benchmark and give some context. Otherwise some parts of the code does not make sense as standalone alpaka example.

example/babelstream/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    if(alpaka_USE_SOURCE_TREE)
+        # Don't build the examples recursively
+        set(alpaka_BUILD_EXAMPLES OFF)
+        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../.." "${CMAKE_BINARY_DIR}/alpaka")
+    else()
+        find_package(alpaka REQUIRED)
+    endif()
+endif()
+
+alpaka_add_executable(${PROJECT_NAME} src/main.cpp src/Stream.h src/AlpakaStream.cpp src/AlpakaStream.h)
```
SimeonEhrig, line 35: You should not add header files. Actual, the compiler should find the header files automatically, because they are in the same folder like the source files.

bernhardmgruber:
> @bernhardmgruber Can you please add a small `README.md` to the example. The README should point to the original benchmark and give some context. Otherwise some parts of the code does not make sense as standalone alpaka example.

I added a small README.

bernhardmgruber:
I changed the history to first commit the cupla port with @jyoung3131 as author and then applied my changes on top. Is that fine for you?

example/babelstream/CMakeLists.txt
bernhardmgruber: I prefer it the other way around. If you don't add header files, your IDE may not find them :) And there is no harm in adding the headers.

example/babelstream/src/AlpakaStream.cpp
```diff
@@ -0,0 +1,258 @@
...
+        alpaka::getPtrNative(d_a),
+        alpaka::getPtrNative(d_b),
+        alpaka::getPtrNative(d_sum),
+        arraySize);
+    alpaka::wait(queue);
+
+    alpaka::memcpy(queue, sums, d_sum);
+    T const* sumPtr = alpaka::getPtrNative(sums);
+    // TODO(bgruber): replace by std::reduce, when gcc 9.3 is the baseline
+    return std::accumulate(sumPtr, sumPtr + DOT_NUM_BLOCKS, 0);
```
SimeonEhrig: Needs to be: `return std::accumulate(sumPtr, sumPtr + DOT_NUM_BLOCKS, static_cast<T>(0));` otherwise it cast each value to `int`.

bernhardmgruber: Thx! Fixed.

example/babelstream/src/AlpakaStream.cpp
bernhardmgruber: There is not a big difference between the block sizes:
```
BabelStream
Version: 4.0
Implementation: alpaka
Running kernels 100 times
Precision: double
Array size: 268.4 MB (=0.3 GB)
Total size: 805.3 MB (=0.8 GB)
Using alpaka device Quadro GV100

with 1024:
Function    MBytes/sec  Min (sec)   Max         Average
Copy        590977.242  0.00091     0.00096     0.00091
Mul         587894.898  0.00091     0.00094     0.00092
Add         615253.841  0.00131     0.00133     0.00131
Triad       614809.491  0.00131     0.00133     0.00132
Dot         644256.949  0.00083     0.00087     0.00084

with 512:
Function    MBytes/sec  Min (sec)   Max         Average
Copy        592219.119  0.00091     0.00091     0.00091
Mul         588827.263  0.00091     0.00092     0.00092
Add         616218.491  0.00131     0.00133     0.00131
Triad       615375.609  0.00131     0.00132     0.00131
Dot         642840.616  0.00084     0.00086     0.00084

with 256:
Function    MBytes/sec  Min (sec)   Max         Average
Copy        592287.720  0.00091     0.00091     0.00091
Mul         589497.085  0.00091     0.00092     0.00092
Add         616160.499  0.00131     0.00132     0.00131
Triad       615247.261  0.00131     0.00132     0.00131
Dot         649585.728  0.00083     0.00085     0.00083
```
But we need to revisit this anyway to support CPU devices as well. But this will be a follow-up PR.

SimeonEhrig: Okay, that's fine.

SimeonEhrig (dismissed):

SimeonEhrig:
@bernhardmgruber The GitLab CI was not triggered. Can you please push again.

bernhardmgruber:
> @bernhardmgruber The GitLab CI was not triggered. Can you please push again.

Rebased and force pushed.

SimeonEhrig (dismissed):

bernhardmgruber:
I rebased to trigger the CI again and get the latest changes from recently merged PRs.

SimeonEhrig (approved):

SimeonEhrig:
@psychocoderHPC @bernhardmgruber The default for merging is `squash and merge` but I think this is not correct for this PR, because we want to keep the author ship of the different commits. Should I run `rebase and merge`?

bernhardmgruber:
My default has always been rebase and merge, for years now. Squash makes sense for beginners that have a hard time preparing their commit history. Otherwise, please do not destroy the history.

PR #1849, opened 2022-12-01
remove boost 1.73 warning

psychocoderHPC:
alpaka requires boost 1.74.0+ therefore wrnings for the boost <1.73.0 can be removed.

### DIFF
--- a/include/alpaka/atomic/AtomicCpu.hpp
+++ b/include/alpaka/atomic/AtomicCpu.hpp
@@ -12,9 +12,6 @@
 #include <boost/version.hpp>
 
 #ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
-#    if BOOST_VERSION < 107300
-#        warning boost::atomic_ref requires Boost version 1.73 or higher. Please update your version of Boost, or disable the use of boost::atomic_ref with -DALPAKA_DISABLE_ATOMIC_ATOMICREF
-#    endif
 #    define ALPAKA_DISABLE_ATOMIC_ATOMICREF
 #endif
 


fwyzard:
for those of us not using CMake, is there a `#warning` or `#error` somewhere else, if the version of Boost is too old ?

bernhardmgruber (dismissed):

j-stephan:
I think this was introduced by CMS to have a diagnostic outside of CMake. We should keep the warning until we have better diagnostics as part of the standalone headers.

j-stephan (changes requested):
See above.

psychocoderHPC:
> I think this was introduced by CMS to have a diagnostic outside of CMake. We should keep the warning until we have better diagnostics as part of the standalone headers.

This warning has nothing todo with the minimum boost version we support. So this line should be removed.
The check is also allowing using boost 1.73.0 which is not supported by alpaka.


> for those of us not using CMake, is there a #warning or #error somewhere else, if the version of Boost is too old ?

@fwyzard I think we do not reflect the requirements as `#warning` or `#error` in the code but we should do it if someone is using standalone headers. Never the less this issue is independent of the error I remove with this PR.

psychocoderHPC:
I opened an issue that the tasks to verify the boost version in the code base is not get lost https://github.com/alpaka-group/alpaka/issues/1856

psychocoderHPC:
I rebased the PR to re-trigger the CI

fwyzard:
> Never the less this issue is independent of the error I remove with this PR.

Agreed, and thanks for opening https://github.com/alpaka-group/alpaka/issues/1856.

bernhardmgruber (dismissed):

j-stephan (dismissed):

j-stephan:
Looks like the GitLab CI was never started. Can you trigger it again?

psychocoderHPC:
I restarted the CI.

bernhardmgruber (approved):

j-stephan:
> Agreed, and thanks for opening https://github.com/alpaka-group/alpaka/issues/1856.

FYI, this is now tracked in #1636.

j-stephan (approved):

PR #1850, opened 2022-12-04
Attempts to fix queue bug

bernhardmgruber:
This PR fixes two bugs, which make TSAN stop reporting racy conditions in `queueTest`.

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -309,10 +309,10 @@ namespace alpaka::core
                     m_bDetachedFlag = true;
                 else
                 {
-                    {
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
Hmm, the CI has a lot of errors with: `Unable to return device handle for device 0. There are only 0 devices!`. That is weird.

j-stephan:
Is this ready?

bernhardmgruber:
> Is this ready?

Well, as Djikstra said, `program testing can be used very effectively to show the presence of bugs but never to show their absence.` I tried to fix the queue bug, and the CI ran through without hickup. TSAN showed a race condition before this PR and now it is gone. Whether the bug is fixed with this, we will find out :) But the state of affairs is hopefully better with this PR.

So, please go ahead and merge it!

j-stephan (approved):

PR #1851, opened 2022-12-04
Enable TSAN in CI

### LINKED ISSUE #2016: Enable TSAN in the CI
Since alpaka involves a fair amount of threading, it could help to catch bugs by enabling thread sanitizer (TSAN) in the CI. TSAN is supported by both, gcc and clang. TSAN is enabled by compiling and linking with `-fsanitize=thread`.

TSAN is also getting better/faster in each version, so it probably makes sense to always use the latest clang or gcc. For clang, PR #1851 found out that a CI run with clang-13 took 68min, and with clang-14 only 22min.

TSAN may also not work out of the box with all threading technologies. For OpenMP, it seems that we need to load additional tools into the OpenMP runtime before executing the test. For clang:
`OMP_TOOL_LIBRARIES=libarcher.so TSAN_OPTIONS='ignore_noninstrumented_modules=1' ARCHER_OPTIONS="verbose=1" <test executable>` . For the gcc OpenMP implementation, there seems to be no corresponding tool.

bernhardmgruber:
This PR enables thread sanitizer on one of the CI runs.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -184,6 +184,9 @@ jobs:
         - name: linux_clang-16_debug_ubsan
           os: ubuntu-latest
           env: {CXX: clang++, CC: clang,  ALPAKA_CI_CLANG_VER: 16,     ALPAKA_CI_STDLIB: libstdc++, CMAKE_BUILD_TYPE: Debug,   ALPAKA_BOOST_VERSION: 1.77.0, ALPAKA_CI_CMAKE_VER: 3.23.5, OMP_NUM_THREADS: 4, ALPAKA_CI_DOCKER_BASE_IMAGE_NAME: "ubuntu:22.04", CMAKE_CXX_EXTENSIONS: OFF, ALPAKA_CI_SANITIZERS: UBSan}
+        - name: linux_clang-16_debug_tsan
+          os: ubuntu-latest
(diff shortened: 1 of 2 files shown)


bernhardmgruber:
The TSan job took 68 min and generated >200MiB logs on detected race conditions. There could be some false positives, but this is definitely something that someone should look into.

bernhardmgruber:
Wow, enabling TSAN on the clang-14 instead of the clang-13 job now only took 22min.

bernhardmgruber:
So, since we merged #1850 in the meantime, the TSAN log shrunk to 150MiB. So naively, we maybe fixed one quarter of the bugs :D

bernhardmgruber:
OMG, all tests pass except the fence test :D So in light of our VC today, where we discussed memory fences, we are probably going to need a better definition/implementation of fences anyway!

j-stephan (changes requested):
Thread Sanitizer doesn't support memory fences yet: https://github.com/google/sanitizers/issues/1352

We should add an exception to the fence test case. Afterwards I think we can go ahead and merge this.

bernhardmgruber:
> Thread Sanitizer doesn't support memory fences yet: [google/sanitizers#1352](https://github.com/google/sanitizers/issues/1352)

Right. Thx for bringing this to my attention (again). I think I have even read about this at some point.

> We should add an exception to the fence test case.

I was briefly thinking about how we could maybe annotate the alpaka API with the [TSAN intrinsics](https://github.com/llvm/llvm-project/blob/ad1161f9b5ff251b80788033e4db82b5a11b187b/compiler-rt/include/sanitizer/tsan_interface.h#L24-L25) to make TSAN happy, but I don't see how to specify the address for the fence, since it could synchronize with *any* atomic somewhere else in the system. So yes, I think we should just exclude the test.

bernhardmgruber:
So, turning TSAN off for just one test turned out to be harder than I thought. I tried editing the target and directory properties to remove the TSAN flag and have the test run without TSAN, but TSAN comes in via `CMAKE_CXX_FLAGS` and I didn't dare to modify that :) So in the end, I am just removing the test entirely.

j-stephan (dismissed):

j-stephan:
The TSAN runner completes successfully. Do you want to convert this into a real PR?

bernhardmgruber:
> The TSAN runner completes successfully. Do you want to convert this into a real PR?

OMG. YESSS!! :D

j-stephan (approved):

psychocoderHPC (approved):

PR #1852, opened 2022-12-04
Refactor ConcurrentExecPool and related classes

bernhardmgruber:
This should be a pure refactoring with no functional changes. This is a preliminary refactoring to investigate the queuebug from https://github.com/alpaka-group/alpaka/pull/1850.

### DIFF
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -33,41 +33,37 @@ namespace alpaka::core
     namespace detail
     {
         template<typename T>
-        class ThreadSafeQueue : private std::queue<T>
+        struct ThreadSafeQueue
         {
-        public:
             ThreadSafeQueue() = default;
-            //! \return If the queue is empty.
+
             [[nodiscard]] auto empty() const -> bool
             {
-                return std::queue<T>::empty();
+                return m_queue.empty();
             }
-            //! Pushes the given value onto the back of the queue.
-            auto push(T&& t) -> void
-            {
-                std::lock_guard<std::mutex> lk(m_Mutex);
 
-                std::queue<T>::push(std::forward<T>(t));
+            //! Pushes the given value onto the back of the queue.
+            void push(T&& t)
+            {
+                std::lock_guard<std::mutex> lk(m_mutex);
+                m_queue.push(std::move(t));
             }
+
             //! Pops the given value from the front of the queue.
             auto pop(T& t) -> bool
             {
-                std::lock_guard<std::mutex> lk(m_Mutex);
+                std::lock_guard<std::mutex> lk(m_mutex);
 
-                if(std::queue<T>::empty())
-                {
+                if(m_queue.empty())
                     return false;
-                }
-                else
-                {
-                    t = std::move(std::queue<T>::front());
-                    std::queue<T>::pop();
-                    return true;
-                }
+                t = std::move(m_queue.front());
+                m_queue.pop();
+                return true;
             }
 
         private:
-            std::mutex m_Mutex;
+            std::queue<T> m_queue;
+            std::mutex m_mutex;
         };
 
         //! ITaskPkg.
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -73,13 +73,12 @@ namespace alpaka::core
 #    pragma clang diagnostic push
 #    pragma clang diagnostic ignored "-Wweak-vtables"
 #endif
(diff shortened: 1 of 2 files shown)


include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -33,41 +33,37 @@ namespace alpaka::core
...
             }
+
             //! Pushes the given value onto the back of the queue.
-            auto push(T&& t) -> void
+            void push(T&& t)
             {
-                std::lock_guard<std::mutex> lk(m_Mutex);
-
-                std::queue<T>::push(std::forward<T>(t));
+                std::lock_guard<std::mutex> lk(m_mutex);
```
fwyzard, line 48: Is the `<std::mutex>` part still needed with c++17 ?

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -77,118 +73,74 @@ namespace alpaka::core
...
+            void runTask() noexcept
             {
                 try
                 {
                     run();
                 }
                 catch(...)
                 {
-// Workaround: Clang can not support this when natively compiling device code. See ConcurrentExecPool.hpp.
+// Workaround: Clang can not support this when natively compiling device code.
```
fwyzard, line 89: Just wondering: is this a problem only for CUDA, or also for HIP and SYCL device compilation ?

include/alpaka/core/ConcurrentExecPool.hpp
bernhardmgruber: It shoudn't and I would prefer CTAD here, but clang gave me a warning (`-Wctad-maybe-unsupported`) so I provided the type explicitely.

include/alpaka/core/ConcurrentExecPool.hpp
bernhardmgruber: I have no idea. I could even be that the workaround is no longer necessary. It was added in c7b27ed6afed86da26146d6ec3aacc3e350a1762 in May 2016.

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
```diff
@@ -73,11 +73,8 @@ namespace alpaka
                     m_dev.registerCleanup(
                         [pool = std::weak_ptr<ThreadPool>(m_workerThread)]() noexcept
                         {
-                            auto s = pool.lock();
-                            if(s)
-                            {
-                                s = s->takeDetachHandle();
-                            }
+                            if(auto s = pool.lock())
+                                static_cast<void>(s->takeDetachHandle()); // let returned shared_ptr destroy itself
```
j-stephan, line 77:
```suggestion
                               std::ignore = s->takeDetachHandle(); // let returned shared_ptr destroy itself
```

include/alpaka/core/ConcurrentExecPool.hpp
```diff
@@ -33,41 +33,37 @@ namespace alpaka::core
...
+
             [[nodiscard]] auto empty() const -> bool
             {
-                return std::queue<T>::empty();
+                return m_queue.empty();
             }
+
             //! Pushes the given value onto the back of the queue.
-            auto push(T&& t) -> void
+            void push(T&& t)
```
j-stephan, line 46: Is this the recommended practice now?

j-stephan (changes requested):
Could you briefly describe why the refactoring is required at this point (or link to a relevant issue) in case we want to re-review this in the future?

bernhardmgruber:
> Could you briefly describe why the refactoring is required at this point (or link to a relevant issue) in case we want to re-review this in the future?

Yes, this is a preliminary refactoring to investigate the queuebug from #1850.

include/alpaka/queue/QueueGenericThreadsNonBlocking.hpp
bernhardmgruber: I know you can assign anything to `std::ignore`, but I find this even more arcane than the established cast to `void`. Would you prefer assigning to `std::ignore`? I don't care about it.

include/alpaka/core/ConcurrentExecPool.hpp
bernhardmgruber: I only use a trailing return type for non-void return types and so far nobody complained.

include/alpaka/core/ConcurrentExecPool.hpp
j-stephan: Please open an issue for the work-around. We should try to remove it.

j-stephan (approved):

PR #1861, opened 2022-12-09
Update to Catch2 v3.2.1

j-stephan:
See title.


bernhardmgruber (dismissed):

j-stephan:
There was a Catch2 update on the weekend. I updated this PR accordingly.

thirdParty/CMakeLists.txt
```diff
@@ -27,7 +27,7 @@ if(BUILD_TESTING)
         add_compile_definitions($<$<CXX_COMPILER_ID:PGI>:__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1>)
         add_subdirectory(catch2)
     else()
-        find_package(Catch2 3.1.1 CONFIG REQUIRED)
+        find_package(Catch2 3.2.1 CONFIG REQUIRED)
```
bernhardmgruber, line 30: Just because we ship Catch 3.2.1 internally, does not mean that this is the minimum version we require? Can we require a lower version number here?

j-stephan: [According to the developers](https://github.com/catchorg/Catch2/pull/2516#issuecomment-1249892636) Catch2 shouldn't be installed globally at all. I would rather remove this part entirely instead of supporting users with older Catch2 versions.

bernhardmgruber: That's a pity. I installed it globally to avoid rebuilding it, which used to happen several times a day. Anyway, I can reintall the newest one if necessary.

bernhardmgruber (approved):

bernhardmgruber (approved):

PR #1862, opened 2022-12-09
Mark CATCH cmake variables as advanced

### LINKED ISSUE #1758: [RFC] Mark CATCH cmake variables as advanced
I just noticed that we got a lot of cmake variables from Catch2 in the default view of `ccmake`:
image

Since I don't think that we should tamper with these variables, I would like to mark them as advanced in cmake, so they are hidden by default.

bernhardmgruber:
Fixes: #1758

### DIFF
--- a/thirdParty/CMakeLists.txt
+++ b/thirdParty/CMakeLists.txt
@@ -26,6 +26,14 @@ if(BUILD_TESTING)
         # still appears in NVHPC 20.7
         add_compile_definitions($<$<CXX_COMPILER_ID:PGI>:__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1>)
         add_subdirectory(catch2)
+
+        # hide Catch2 cmake variables by default in cmake gui
+        get_cmake_property(variables VARIABLES)
+        foreach (var ${variables})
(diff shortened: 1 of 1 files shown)


thirdParty/CMakeLists.txt
```diff
@@ -26,6 +26,14 @@ if(BUILD_TESTING)
...
         add_compile_definitions($<$<CXX_COMPILER_ID:PGI>:__GCC_ATOMIC_TEST_AND_SET_TRUEVAL=1>)
         add_subdirectory(catch2)
+
+    # hide Catch2 cmake variables by default in cmake gui
+    get_cmake_property(variables VARIABLES)
+    foreach (var ${variables})
+        if (var MATCHES "^CATCH_")
+            mark_as_advanced(${var})
+        endif()
+    endforeach()
```
j-stephan: The indentation is wrong here.

j-stephan (changes requested):

thirdParty/CMakeLists.txt
bernhardmgruber: Would be great to have a clang-format for cmake!

j-stephan (approved):

PR #1845, opened 2022-11-29
Rewrite the SYCL backend for the SYCL 2020 standard and USM allocations

Parsifal-2045:
Rewrite the SYCL backend to support the SYCL 2020 standard, using USM allocations instead of SYCL buffers and accessors.

Few highlights:
  - full support for the SYCL 2020 standard and the Unified Shared Memory (USM) model;
  - rewrite the N-dimensional Copy and Set memory operations to support pitched memory buffers, based on the Cpu implementation;
  - generalise the SYCL CpuSelector to non-Intel CPUs;
  - implement support for random number generators based on the Intel oneAPI libraries;
  - support compile-time sub-group sizes in SYCL kernels (see below);
  - update the documentation, tests and examples.

---

## Kernel trait for compile-time sub-group size

Introduce a new optional trait to describe at compile time the warp size that a kernel should use.  The default behaviour is to let the back-end compiler pick the preferred size.

Before launching a kernel with a compile-time sub-group size the user should query the sizes supported by the device, and choose accordingly.  If the device does not support the requested size, the SYCL runtime will throw a synchronous exception.

During just-in-time (JIT) compilation this guarantees that a kernel is compiled only for the sizes supported by the device.  During ahead-of-time (AOT) compilation this is not enough, because the device is not known at compile time.  The SYCL specification mandates that the back-end compilers should not fail if a kernel uses unsupported features, like unsupported sub-group sizes.  Unfortunately the Intel OpenCL CPU and GPU compilers currently fail with a hard error.  To work around this limitation, use the preprocessor macros defined when compiling AOT for the new SYCL targets to enable the compilation only for the sub-group sizes supported by each device.

Note: while the CPU OpenCL back-end does support a sub-group size of 64, the SYCL code currently does not.  To avoid issues with the sub-group primitives always consider the sub-group size of 64 as not supported by the device.

PR #1865, opened 2022-12-12
Make SYCL runtime objects `static`.

j-stephan:
This is a supporting effort for #1845. The alpaka SYCL platforms are now specialized using the SYCL device selector. Each specialization will contain (as `static` members) the corresponding `sycl::platform, sycl::context, std::vector<sycl::device>`.

CC @fwyzard @Parsifal-2045 @AuroraPerego

### DIFF
--- a/include/alpaka/pltf/PltfGenericSycl.hpp
+++ b/include/alpaka/pltf/PltfGenericSycl.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Jan Stephan
+/* Copyright 2023 Jan Stephan
  *
  * This file is part of Alpaka.
  *
@@ -18,7 +18,13 @@
 
 #    include <CL/sycl.hpp>
 
-#    include <iostream>
+#    include <cstddef>
+#    include <exception>
+#    if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
+#        include <iostream>
+#    endif
+#    include <mutex>
+#    include <optional>
 #    include <sstream>
 #    include <stdexcept>
 #    include <vector>
@@ -26,73 +32,143 @@
 namespace alpaka::experimental
 {
     //! The SYCL device manager.
-    class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl>
+    template<typename TSelector>
+    class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl<TSelector>>
     {
     public:
         PltfGenericSycl() = delete;
+
+        [[nodiscard]] static auto syclPlatform() -> sycl::platform&
+        {
+            auto lock = std::scoped_lock<std::mutex>{mutex};
+
+            if(!initialized)
+                do_initialization();
+
+            return *platform_opt;
+        }
+
+        [[nodiscard]] static auto syclDevices() -> std::vector<sycl::device>&
+        {
+            auto lock = std::scoped_lock<std::mutex>{mutex};
+
+            if(!initialized)
+                do_initialization();
+
+            return devices;
+        }
+
+        [[nodiscard]] static auto syclContext() -> sycl::context&
+        {
+            auto lock = std::scoped_lock<std::mutex>{mutex};
+
+            if(!initialized)
+                do_initialization();
+
+            return *context_opt;
+        }
+
+        static auto initialize() -> void
+        {
+            auto lock = std::scoped_lock<std::mutex>{mutex};
+            do_initialization();
+        }
+
+        static auto reset() -> void
+        {
+            auto lock = std::scoped_lock<std::mutex>{mutex};
+
+            if(!initialized)
+                return;
+
+            context_opt.reset();
+            devices.clear();
+            platform_opt.reset();
+            initialized = false;
+        }
+
+    private:
+        static auto do_initialization()
+        {
+            if(initialized)
+                return;
+
+            platform_opt = sycl::platform{TSelector{}};
+            devices = platform_opt->get_devices();
+            context_opt = sycl::context{
+                devices,
+                [](sycl::exception_list exceptions)
+                {
+                    auto ss_err = std::stringstream{};
+                    ss_err << "Caught asynchronous SYCL exception(s):\n";
+                    for(std::exception_ptr e : exceptions)
+                    {
+                        try
+                        {
+                            std::rethrow_exception(e);
+                        }
+                        catch(sycl::exception const& err)
+                        {
+                            ss_err << err.what() << " (" << err.code() << ")\n";
+                        }
+                    }
+                    throw std::runtime_error(ss_err.str());
+                }};
+
+            initialized = true;
+        }
+
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+        inline static std::mutex mutex;
+        inline static bool initialized{false};
+
+        inline static std::optional<sycl::platform> platform_opt{std::nullopt};
+        inline static std::vector<sycl::device> devices;
+        inline static std::optional<sycl::context> context_opt{std::nullopt};
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic pop
+#    endif
     };
 } // namespace alpaka::experimental
 
 namespace alpaka::trait
 {
     //! The SYCL platform device count get trait specialization.
-    template<typename TPltf>
-    struct GetDevCount<TPltf, std::enable_if_t<std::is_base_of_v<experimental::PltfGenericSycl, TPltf>>>
+    template<typename TSelector>
+    struct GetDevCount<alpaka::experimental::PltfGenericSycl<TSelector>>
     {
         static auto getDevCount() -> std::size_t
         {
             ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
-            static auto dev_num = []
-            {
-                auto platform = sycl::platform{typename TPltf::selector{}};
-                auto devices = platform.get_devices();
-                return devices.size();
-            }();
-
-            return dev_num;
+            return alpaka::experimental::PltfGenericSycl<TSelector>::syclDevices().size();
         }
     };
 
     //! The SYCL platform device get trait specialization.
-    template<typename TPltf>
-    struct GetDevByIdx<TPltf, std::enable_if_t<std::is_base_of_v<experimental::PltfGenericSycl, TPltf>>>
+    template<typename TSelector>
+    struct GetDevByIdx<alpaka::experimental::PltfGenericSycl<TSelector>>
     {
         static auto getDevByIdx(std::size_t const& devIdx)
         {
             ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
-            auto exception_handler = [](sycl::exception_list exceptions)
+            using SyclPltf = alpaka::experimental::PltfGenericSycl<TSelector>;
+
+            auto const dev_num = getDevCount<SyclPltf>();
+
+            if(devIdx >= dev_num)
             {
                 auto ss_err = std::stringstream{};
-                ss_err << "Caught asynchronous SYCL exception(s):\n";
-                for(std::exception_ptr e : exceptions)
-                {
-                    try
-                    {
-                        std::rethrow_exception(e);
-                    }
-                    catch(sycl::exception const& err)
-                    {
-                        ss_err << err.what() << " (" << err.code() << ")\n";
-                    }
-                }
-                throw std::runtime_error(ss_err.str());
-            };
-
-            static auto pf = sycl::platform{typename TPltf::selector{}};
-            static auto devices = pf.get_devices();
-            static auto ctx = sycl::context{devices, exception_handler};
-
-            if(devIdx >= devices.size())
-            {
-                auto ss_err = std::stringstream{};
-                ss_err << "Unable to return device handle for device " << devIdx << ". There are only "
-                       << devices.size() << " SYCL devices!";
+                ss_err << "Unable to return device handle for device " << devIdx << ". There are only " << dev_num
+                       << " SYCL devices!";
                 throw std::runtime_error(ss_err.str());
             }
 
-            auto sycl_dev = devices.at(devIdx);
+            auto sycl_dev = SyclPltf::syclDevices().at(devIdx);
 
             // Log this device.
 #    if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
@@ -100,7 +176,7 @@ namespace alpaka::trait
 #    elif ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
             std::cout << __func__ << sycl_dev.get_info<info::device::name>() << '\n';
 #    endif
-            return typename DevType<TPltf>::type{sycl_dev, ctx};
+            return typename DevType<SyclPltf>::type{sycl_dev, SyclPltf::syclContext()};
         }
 
     private:
--- a/include/alpaka/pltf/PltfCpuSyclIntel.hpp
+++ b/include/alpaka/pltf/PltfCpuSyclIntel.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Jan Stephan
+/* Copyright 2023 Jan Stephan
  *
  * This file is part of Alpaka.
  *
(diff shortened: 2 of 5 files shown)


include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,97 @@
...
+                {
+                    ss_err << err.what() << " (" << err.code() << ")\n";
+                }
+            }
+            throw std::runtime_error(ss_err.str());
+        };
+
+        inline static sycl::platform sycl_platform{TSelector {}};
+        inline static std::vector<sycl::device> sycl_devices{sycl_platform.get_devices()};
+        inline static sycl::context sycl_context{sycl_devices, async_handler};
```
fwyzard: should this part be excluded from the `#if BOOST_COMP_CLANG` blocks ?

j-stephan: No, it is actually the reason why those blocks are there. Otherwise `-Wexit-time-destructors` will be triggered.

j-stephan: IIRC the warning will always appear for non-trivial types.

fwyzard: I don't understand then: how do you build these objects at all, when `BOOST_COMP_CLANG` is not defined ?

j-stephan: Ah, now I see. That was an oversight on my part while doing the refactoring. Thanks for pointing this out.

fwyzard:
Hi @j-stephan, if I read the code correctly, these changes would build all SYCL platforms, contexts and devices at program initialisation time:
```c++
template<typename TSelector>
class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl<TSelector>> {
    ...
    inline static sycl::platform sycl_platform{TSelector{}};
    inline static std::vector<sycl::device> sycl_devices{sycl_platform.get_devices()};
    inline static sycl::context sycl_context{sycl_devices, async_handler};
};
```

We tried this approach and run into some with in the past, due to the order of initialisation and of destruction.

Now we switched to some kind of lazy initialisation, where the objects are initialised the first time they are used instead of at application initialisation time, and an explicit destruction / finalisation.

j-stephan:
> We tried this approach and run into some with in the past, due to the order of initialisation and of destruction.

Interesting. Do you remember how you approached this? According to [cppreference.com](https://en.cppreference.com/w/cpp/language/initialization#Dynamic_initialization) (see points 2 and 3) the way I implemented this here should work since the initializations happen in the right order and in the same translation unit. [See also this StackOverflow question + answer](https://stackoverflow.com/questions/56515342/static-inline-members-initialization-order) which addresses the same issue.

CC @bernhardmgruber , our local C++ lawyer.

fwyzard:
> According to [cppreference.com](https://en.cppreference.com/w/cpp/language/initialization#Dynamic_initialization) (see points 2 and 3) the way I implemented this here should work since the initializations happen in the right order and in the same translation unit.

Sorry for not being clear - I didn't mean initialisation order between the SYCL objects, I meant between them and other global objects.

For example, at some point we were having a crash because the destructor of `cout`/`cerr` was done before the destruction of the SYCL objects, while they are/were trying to print a message during the destruction.

fwyzard:
Another problem actually was with CUDA, thought it may be a general issue that affects other runtimes as well: using a `static` CUDA object would prevent the CUDA profiling tools to work.

j-stephan:
Oh, I see. Yes, that makes sense. I'll change it to the "construct on first use" idiom.

fwyzard:
More in general, we found that initialising all runtimes and objects irrespective if they are used or not may range from being sub-optimal (slow startup and extra resource usage) to outright problematic (some SYCL runtimes would crash).

bernhardmgruber:
I would also go for lazy initialization:
```c++
template<typename TSelector>
class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl<TSelector>> {
    ...
    static auto syclPlatform() {
        static sycl::platform p{TSelector{}};
        return p:
    }
...
};
```
As Andrea said, I can imagine interferrence or indeterminism if we start to construct all objects right at program startup. We may still run into issues on program termination, but this is something we need to test in practice.

fwyzard:
> As Andrea said, I can imagine interferrence or indeterminism if we start to construct all objects right at program startup. We may still run into issues on program termination, but this is something we need to test in practice.

I would suggest to add a `static void reset()` method to the SYCL platforms, so that people can call it explicitly if they run into those kind of problems. 

We would be happy to have it and call it :-)

j-stephan:
Just for clarity: What would `reset()` do? Destruct everything and then reconstruct everything? Or should it act as a shutdown mechanism?

fwyzard:
I'm just making this up as we go along... I guess it should destruct the static objects in the correct order, and leave things in a pre-initialisation state.

So if anything re-uses the objects afterwords, they would just be re-initialised from scratch.

fwyzard:
From a quick look, I'm concerned that this approach is not thread-safe: calling _e.g._ `syclDevices()` from more than one thread could create more than one collection of devices, with the first one being destroyed leading to some left-over objects.

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,135 @@
...
+        }
+
+    private:
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+        inline static std::unique_ptr<sycl::platform> platform_ptr{nullptr};
+        inline static std::unique_ptr<std::vector<sycl::device>> devices_ptr{nullptr};
+        inline static std::unique_ptr<sycl::context> context_ptr{nullptr};
```
bernhardmgruber: Why not `std::optional` instead of `std::unique_ptr`?

j-stephan: Old habits die hard...

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,150 @@
...
+    class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl<TSelector>>
     {
     public:
         PltfGenericSycl() = delete;
+
+        [[nodiscard]] static auto syclPlatform() -> sycl::platform&
+        {
+            auto lock = std::unique_lock<std::mutex>{platform_mutex};
+
+            if(!platform_opt.has_value())
```
bernhardmgruber: Optional is Boolish (i.e. contextually convertible to bool), like pointers. So it is often used like this:
```suggestion
            if(!platform_opt)
```

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,150 @@
...
+                platform_opt = std::make_optional<sycl::platform>(TSelector{});
```
bernhardmgruber: Why not this:
```suggestion
                platform_opt = sycl::platform(TSelector{});
```

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,150 @@
...
+            }
+
+            return context_opt.value();
+        }
+
+        // Use destructor attribute to make sure our static variables are destroyed in the correct order at shutdown.
+        [[gnu::destructor]] static auto reset() -> void
+        {
+            auto lock
+                = std::scoped_lock<std::mutex, std::mutex, std::mutex>{platform_mutex, devices_mutex, context_mutex};
```
bernhardmgruber: I guess you also got a clang warning here? I wonder whether we should disable that one. I would prefer:
```suggestion
            auto lock
                = std::scoped_lock{platform_mutex, devices_mutex, context_mutex};
```

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,150 @@
...
+        }
+
+        [[nodiscard]] static auto syclDevices() -> std::vector<sycl::device>&
+        {
+            auto lock = std::unique_lock<std::mutex>{devices_mutex};
+
+            if(!devices_opt.has_value())
+                devices_opt = std::make_optional<std::vector<sycl::device>>(syclPlatform().get_devices());
+
+            return devices_opt.value();
```
bernhardmgruber: This is actually a pattern and we might create a class for that at some point. I implemented something very similar a long time ago: https://github.com/DLR-SC/tigl/blob/master/src/Cache.h#L49-L103

You would wrap an optional, a mutex and a creation function together into a class and add a few functions to access the contained value and invalidate the cache. The value is recreated when accessed and the optional is empty.

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,150 @@
...
+                        }
+                        throw std::runtime_error(ss_err.str());
+                    });
+            }
+
+            return context_opt.value();
+        }
+
+        // Use destructor attribute to make sure our static variables are destroyed in the correct order at shutdown.
+        [[gnu::destructor]] static auto reset() -> void
```
bernhardmgruber: I am afraid of that attribute, because I have no experience with it. I wonder whether alpaka would also shutdown fine without it.

bernhardmgruber (dismissed):
In principle, I think this PR is fine. I have a couple of questions and suggestions though, but none of them necessarily blocks a merge.

include/alpaka/pltf/PltfGenericSycl.hpp
j-stephan: Yes, I know. I wanted to make this explicit.

include/alpaka/pltf/PltfGenericSycl.hpp
j-stephan: Yes, that CTAD warning is really annoying. One would think they would guard against the standard library.

include/alpaka/pltf/PltfGenericSycl.hpp
j-stephan: It probably would since the static vars should be destroyed in the reverse order after `main()` finishes. I can remove the attribute (which is quite useful for writing plugins).

include/alpaka/pltf/PltfGenericSycl.hpp
bernhardmgruber: That's what I thought. Fine then.

include/alpaka/pltf/PltfGenericSycl.hpp
bernhardmgruber: Does the order of destruction of the SYCL objects even matter? Does SYCL require that you hold a `sycl::platform` alive until all contextes and devices are destroyed?

I am asking because OpenCL did not require that.

fwyzard: Not that I know of.
I expect them to have internal reference counts and to guarantee the proper order of destruction - but I didn't check the implementation.
AFAIK we did not see any problems due to the order of destruction of SYCL static objects among themselves -- only due to the order of SYCL objects vs other objects.

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,150 @@
...
+
+            return platform_opt.value();
+        }
+
+        [[nodiscard]] static auto syclDevices() -> std::vector<sycl::device>&
+        {
+            auto lock = std::unique_lock<std::mutex>{devices_mutex};
+
+            if(!devices_opt.has_value())
+                devices_opt = std::make_optional<std::vector<sycl::device>>(syclPlatform().get_devices());
```
fwyzard:
```suggestion
                devices_opt = syclPlatform().get_devices();
```
should work as well ?

fwyzard:
Thanks for the changes, I _think_ this approach_should be fine.

I am vaguely concerned if we could run into a dead lock if different threads try to call different initialisation functions at the same time. Or, if one thread tries to re-initialise an object while another one is calling `reset()`.

My _naive_ approach would have bee to use a single `mutex`: then, either initialise all objects in a single method, or split the methods into a public interface (that does the locking) and a private implementation (that does the initialisation).

That said, if the current approach is deemed safe against race conditions and dead locks, I'm fine with it.

fwyzard (commented):
Thanks for the changes, I _think_ this approach should be fine.

I am vaguely concerned if we could run into a dead lock if different threads try to call different initialisation functions at the same time. Or, if one thread tries to re-initialise an object while another one is calling `reset()`.

My _naive_ approach would have bee to use a single `mutex`: then, either initialise all objects in a single method, or split the methods into a public interface (that does the locking) and a private implementation (that does the initialisation).

That said, if the current approach is deemed safe against race conditions and dead locks, I'm fine with it.

j-stephan:
Thanks for the comments. I'll address them in the afternoon, so please don't merge just yet.

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,164 @@
...
 #    include <vector>
 
 namespace alpaka::experimental
 {
     //! The SYCL device manager.
-    class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl>
+    template<typename TSelector>
+    class PltfGenericSycl : public concepts::Implements<ConceptPltf, PltfGenericSycl<TSelector>>
     {
+        using RecursiveLock = std::unique_lock<std::recursive_mutex>;
```
bernhardmgruber: Why not?
```suggestion
        using RecursiveLock = std::scoped_lock<std::recursive_mutex>;
```

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -18,89 +18,164 @@
...
+            platform_opt.reset();
+            initialized = false;
+        }
+
+    private:
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+        inline static std::recursive_mutex mutex;
```
bernhardmgruber: I don't see why a recursive mutex is needed here. Why can't this just be an ordinary `std::mutex`? Where is the reentrancy here?

include/alpaka/pltf/PltfGenericSycl.hpp
j-stephan: What would be the benefit of using `scoped_lock` in this case?

include/alpaka/pltf/PltfGenericSycl.hpp
j-stephan: I've chosen `recursive_mutex` because both `initialize()` and the entire `syclObject()` methods need to be guarded. Otherwise, T1 might call `reset()` *after* T2 is past the `initialize()` call.

What I would actually need here is a `std::recursive_shared_mutex` (for shared access in `syclObject()` and exclusive access in `initialize()`) but sadly the standard library doesn't provide this.

include/alpaka/pltf/PltfGenericSycl.hpp
bernhardmgruber: It's a less capable, simpler mechanism. It is the C++17 replacement of ` lock_guard`. `unique_lock` allows you to unlock manually as well, so has more complexity. `scoped_lock`/`lock_guard` are pure RAII solutions.

include/alpaka/pltf/PltfGenericSycl.hpp
bernhardmgruber: Ah, I see now that `initialize`  is public and also locks as well. So I see now where you would have to lock twice. Still, I would prefer if you could refactor the code to not need a `recursive_mutex`. This would reduce the mental burden. Also, `recursive_mutex` has more overhead than `mutex`, but that should not matter much here.

See also the recommendation from a stdlibc++ implementer: https://stackoverflow.com/a/14600868

bernhardmgruber (dismissed):

fwyzard (dismissed):
Thanks for the latest round of changes, everything looks good to me now.

Being paranoid about performance, you _might_ make two changes:
  - replace `std::optional<std::vector<sycl::device>>` with simply `std::vector<sycl::device>`, and leave the vector empty until it is initialised
  - replace all call top `optional.value()` with `*optional`, since they are anyway protected by `if (initialized)`

Neither of this is likely to show up in any kind of profiling, so do use the style you like better.

bernhardmgruber (approved):

PR #1868, opened 2022-12-12
fix CUDA memory allocation mapped/async

psychocoderHPC:
At least some CUDA versions return the unchanged pointer given to CUDA memory allocation if the size to allocate is zero. If the destination pointer is not initialized with `nullptr` and contains random data the random data is returned.
The created alpaka buffer is passing the pointer to the corresponding function to free the memory, in case the pointer is a random address the free is failing with an error written to stderr because alpaka guarding the CUDA call with a check which is not throwing in case an error happened.

We observed this error with PIConGPU which is using cupla to allocate memory.

### DIFF
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -274,7 +274,7 @@ namespace alpaka
 
                 auto const& dev = getDev(queue);
                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::setDevice(dev.getNativeHandle()));
-                void* memPtr;
+                void* memPtr = nullptr;
                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::mallocAsync(
                     &memPtr,
                     static_cast<std::size_t>(width) * sizeof(TElem),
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
# nvcc version
```
nvcc --version
Copyright (c) 2005-2020 NVIDIA Corporation
Built on Mon_Nov_30_19:08:53_PST_2020
Cuda compilation tools, release 11.2, V11.2.67
Build cuda_11.2.r11.2/compiler.29373293_0
```
# compile:
```
nvcc  -std=c++17 main.cu
```

# reproducer
```C++
#include <cuda_runtime_api.h>
#include <iostream>
#include <cstdint>


int main()
{

    {
        void* ptr=nullptr;
        if(cudaSuccess!=cudaMallocHost(&ptr,0))
            std::cout<<"alloc issue nullptr"<<std::endl;

        std::cout<<"ptr="<<reinterpret_cast<size_t>(ptr)<<std::endl;

        if(cudaSuccess!=cudaFreeHost(ptr))
           std::cout<<"error 0 alloc nullptr init"<<std::endl;

        cudaGetLastError();
    }
    {
        int* ptr;
        ptr = (int*)5;
        if(cudaSuccess!=cudaMallocHost((void**)&ptr,0))
            std::cout<<"alloc issue"<<std::endl;

        std::cout<<"ptr="<<reinterpret_cast<size_t>(ptr)<<std::endl;

        if(cudaSuccess!=cudaFreeHost(ptr))
           std::cout<<"error 0 alloc"<<std::endl;

        cudaGetLastError();
    }

    return 0;
}

```


# output

```
./a.out 
ptr=0
ptr=5 # should not be 5 it should be zero
error 0 alloc
```

bernhardmgruber (approved):

j-stephan (approved):

PR #1870, opened 2022-12-14
Replace ConcurrentExecPool by CallbackThread in all queues

bernhardmgruber:
This PR replaces `ConcurrentExecPool` as a background thread inside `QueueGenericThreadsNonBlockingImpl` by the much simpler `CallbackThread`. Furthermore, since it is no longer necessary, the detaching logic in `ConcurrentExecPool` is removed. The implementation for `Enqueue` for `QueueUniformCudaHipRt` is also vastly simplified.

This PR causes one change in behavior: the destructor of a non-blocking queue will now block until all tasks in the queue are completed. This is required to ensure deterministic lifetime and destruction of queues, background threads, tasks, etc. If we are fine with this new behavior, we should also cover it by a unit test.

This changeset fixes several issues reported by TSAN (see #1936), although it does not completely fix all errors. I am suspecting there could be some remaining false positives.

Depends on:

- [x] Merge #1975 first
- [x] Merge #1987 first

When this PR is merged, `ConcurrentExecPool` is only used in one place anymore, as a yielding thread pool for running CPU kernels of the Threads backend. I think we could cut down the class to it's new purpose and potentially optimize performance.

### DIFF
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -38,6 +38,11 @@ namespace alpaka::core
             }
         }
 
+        // Note: due to different std lib implementations of packaged_task, the lifetime of the passed function either
+        // ends when the packaged_task is destroyed (while the returned future is still alive) or when both are
+        // destroyed. Therefore, ensure that a submitted task does not extend the lifetime of any object that
+        // (transitively) holds the returned future. E.g. don't capture a shared_ptr to an alpaka object that stores
+        // the returned future. This is a cyclic dependency and creates a leak.
         template<typename NullaryFunction>
         auto submit(NullaryFunction&& nf) -> std::future<void>
         {
@@ -52,6 +57,7 @@ namespace alpaka::core
             auto f = task.get_future();
             {
                 std::unique_lock<std::mutex> lock{m_mutex};
+                ++m_tasksInProgress;
                 m_tasks.emplace(std::move(task));
                 if(!m_thread.joinable())
                     startWorkerThread();
@@ -60,12 +66,18 @@ namespace alpaka::core
             return f;
         }
 
+        [[nodiscard]] auto empty() const
+        {
+            return m_tasksInProgress == 0;
+        }
+
     private:
         std::thread m_thread;
         std::condition_variable m_cond;
         std::mutex m_mutex;
         bool m_stop{false};
         std::queue<Task> m_tasks;
+        std::atomic<int> m_tasksInProgress{0};
 
         auto startWorkerThread() -> void
         {
@@ -74,21 +86,24 @@ namespace alpaka::core
                 {
                     while(true)
                     {
-                        // Do not move the tasks out of the loop else the lifetime could be extended until the moment
-                        // where the callback thread is destructed.
-                        Task task;
                         {
-                            std::unique_lock<std::mutex> lock{m_mutex};
-                            m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
+                            // Do not move the tasks out of the loop else the lifetime could be extended until the
+                            // moment where the callback thread is destructed.
+                            Task task;
+                            {
+                                std::unique_lock<std::mutex> lock{m_mutex};
+                                m_cond.wait(lock, [this] { return m_stop || !m_tasks.empty(); });
 
-                            if(m_stop && m_tasks.empty())
-                                break;
+                                if(m_stop && m_tasks.empty())
+                                    break;
 
-                            task = std::move(m_tasks.front());
-                            m_tasks.pop();
+                                task = std::move(m_tasks.front());
+                                m_tasks.pop();
+                            }
+
+                            task();
                         }
-
-                        task();
+                        --m_tasksInProgress;
                     }
                 });
         }
--- a/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
+++ b/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
@@ -63,10 +63,10 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
-                // In case the device is still doing work in the queue when cuda/hip-StreamDestroy() is called, the
-                // function will return immediately and the resources associated with queue will be released
-                // automatically once the device has completed all work in queue.
-                // -> No need to synchronize here.
+                // Make sure all pending async work is finished before destroying the stream to guarantee determinism.
+                // This would not be necessary for plain CUDA/HIP operations, but we can have host functions in the
+                // stream, which reference this queue instance and its CallbackThread. Make sure they are done.
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamSynchronize(m_UniformCudaHipQueue));
                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamDestroy(m_UniformCudaHipQueue));
             }
 
@@ -185,86 +185,43 @@ namespace alpaka
         template<typename TApi, bool TBlocking, typename TTask>
         struct Enqueue<uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, TBlocking>, TTask>
         {
-            enum class CallbackState
-            {
-                enqueued,
-                notified,
-                finished,
-            };
+            using QueueImpl = uniform_cuda_hip::detail::QueueUniformCudaHipRtImpl<TApi>;
 
-            struct CallbackSynchronizationData : public std::enable_shared_from_this<CallbackSynchronizationData>
+            struct HostFuncData
             {
-                std::mutex m_mutex;
-                std::condition_variable m_event;
-                CallbackState m_state = CallbackState::enqueued;
+                // We don't need to keep the queue alive, because in it's dtor it will synchronize with the CUDA/HIP
+                // stream and wait until all host functions and the CallbackThread are done. It's actually an error to
+                // copy the queue into the host function. Destroying it here would call CUDA/HIP APIs from the host
+                // function. Passing it further to the Callback thread, would make the Callback thread hold a task
+                // containing the queue with the CallbackThread itself. Destroying the task if no other queue instance
+                // exists will make the CallbackThread join itself and crash.
+                QueueImpl& q;
+                TTask t;
             };
 
             ALPAKA_FN_HOST static void uniformCudaHipRtHostFunc(void* arg)
             {
-                // explicitly copy the shared_ptr so that this method holds the state even when the executing thread
-                // has already finished.
-                auto const spCallbackSynchronizationData
-                    = reinterpret_cast<CallbackSynchronizationData*>(arg)->shared_from_this();
-
-                // Notify the executing thread.
-                {
-                    std::unique_lock<std::mutex> lock(spCallbackSynchronizationData->m_mutex);
-                    spCallbackSynchronizationData->m_state = CallbackState::notified;
-                }
-                spCallbackSynchronizationData->m_event.notify_one();
-
-                // Wait for the executing thread to finish the task if it has not already finished.
-                std::unique_lock<std::mutex> lock(spCallbackSynchronizationData->m_mutex);
-                if(spCallbackSynchronizationData->m_state != CallbackState::finished)
-                {
-                    spCallbackSynchronizationData->m_event.wait(
-                        lock,
-                        [&spCallbackSynchronizationData]()
-                        { return spCallbackSynchronizationData->m_state == CallbackState::finished; });
-                }
-            }
+                auto data = std::unique_ptr<HostFuncData>(reinterpret_cast<HostFuncData*>(arg));
+                auto& queue = data->q;
+                auto f = queue.m_callbackThread.submit(
+                    [data = std::move(data)]() mutable
+                    {
+                        data->t();
+                        data.reset(); // destroy the task
+                    });
+                f.wait();
+            } // destroys the future `f`, destroying the packaged task and the above lambda
 
             ALPAKA_FN_HOST static auto enqueue(
                 uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, TBlocking>& queue,
                 TTask const& task) -> void
             {
-                auto spCallbackSynchronizationData = std::make_shared<CallbackSynchronizationData>();
                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::launchHostFunc(
                     queue.getNativeHandle(),
                     uniformCudaHipRtHostFunc,
-                    spCallbackSynchronizationData.get()));
-
-                // We start a new std::thread which stores the task to be executed.
-                // This circumvents the limitation that it is not possible to call CUDA/HIP methods within the CUDA/HIP
-                // callback thread. The CUDA/HIP thread signals the std::thread when it is ready to execute the task.
-                // The CUDA/HIP thread is waiting for the std::thread to signal that it is finished executing the task
-                // before it executes the next task in the queue (CUDA/HIP stream).
-                auto f = queue.getCallbackThread().submit(
-                    [spCallbackSynchronizationData, task]()
-                    {
-                        // If the callback has not yet been called, we wait for it.
-                        {
-                            std::unique_lock<std::mutex> lock(spCallbackSynchronizationData->m_mutex);
-                            if(spCallbackSynchronizationData->m_state != CallbackState::notified)
-                            {
-                                spCallbackSynchronizationData->m_event.wait(
-                                    lock,
-                                    [&spCallbackSynchronizationData]()
-                                    { return spCallbackSynchronizationData->m_state == CallbackState::notified; });
-                            }
-
-                            task();
-
-                            // Notify the waiting CUDA/HIP thread.
-                            spCallbackSynchronizationData->m_state = CallbackState::finished;
-                        }
-                        spCallbackSynchronizationData->m_event.notify_one();
-                    });
-
+                    new HostFuncData{*queue.m_spQueueImpl, task}));
                 if constexpr(TBlocking)
-                {
-                    f.wait();
-                }
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::streamSynchronize(queue.getNativeHandle()));
             }
         };
 
--- a/include/alpaka/core/ConcurrentExecPool.hpp
+++ b/include/alpaka/core/ConcurrentExecPool.hpp
@@ -282,29 +282,6 @@ namespace alpaka::core::detail
             return m_numActiveTasks == 0u;
         }
 
-        void detach(std::shared_ptr<ConcurrentExecPool>&& self)
-        {
-            m_self = std::move(self);
-            if constexpr(TisYielding)
-                m_bDetachedFlag = true;
-            else
-            {
-                std::lock_guard<TMutex> lock(this->m_mtxWakeup);
-                m_bDetachedFlag = true;
-                // we need to notify during the lock, because setting m_bDetachedFlag to true, allows another
-                // thread to delete this and thus destroy m_cvWakeup.
-                this->m_cvWakeup.notify_one();
-            }
-        }
-
-        auto takeDetachHandle() -> std::shared_ptr<ConcurrentExecPool>
-        {
-            if(m_bDetachedFlag.exchange(false))
-                return std::move(m_self);
-            else
-                return nullptr;
-        }
-
     private:
         //! The function the concurrent executors are executing.
         void concurrentExecFn()
@@ -317,11 +294,7 @@ namespace alpaka::core::detail
                     if(auto task = popTask())
                         task->runTask();
                     else
-                    {
-                        if(takeDetachHandle())
-                            return; // Pool was detached and is idle, stop and delete
                         TYield::yield();
-                    }
                 }
                 else
                 {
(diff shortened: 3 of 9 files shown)


include/alpaka/core/CallbackThread.hpp
```diff
@@ -34,46 +37,65 @@ namespace alpaka::core
...
                 if(!m_thread.joinable())
                     startWorkerThread();
             }
             m_cond.notify_one();
             return f;
         }
 
+        [[nodiscard]] auto empty() const
+        {
+            return m_tasksInProgress == 0;
```
psychocoderHPC, line 71: Is is not better to use the mutex here and check the queue if they is empty or not. Doing a second accounting even with an atomic variable will most likely introduce a moment where the queue and the counter are not synced.
The counter is only introducing another complexity.

I suggest


```C++
std::unique_lock<std::mutex> lock{m_mutex};
m_tasks.empty();
```

include/alpaka/core/CallbackThread.hpp
```diff
@@ -34,46 +37,65 @@ namespace alpaka::core
...
+        bool m_stop{false};
         std::queue<Task> m_tasks;
+        std::atomic<int> m_tasksInProgress{0};
 
         auto startWorkerThread() -> void
         {
             m_thread = std::thread(
                 [this]
                 {
-                    Task task;
```
psychocoderHPC: note: This is a bug I found yesterday. The problem is that the TAsk is always alive as long as the thread lives. This is moving the livetime of all objects stored in the task to the end of the call back thread or to the moment where the task is overwritten with a new tasks. The result of this is undefined and can create bugs.

include/alpaka/core/CallbackThread.hpp
bernhardmgruber: Yeah, that was my first attempt as well! However, consider that when the background thread dequeues the last task and starts running it, the `m_tasks` is already empty. So if you query the alpaka queues on whether they are empty inside a host func, they say true, but obviously you are still running a task :)

This case is also covered by the unit tests btw.

include/alpaka/core/CallbackThread.hpp
bernhardmgruber: The changes were merged in #1976. I rebased on the top of the changes.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -185,86 +185,43 @@ namespace alpaka
...
-                std::mutex m_mutex;
-                std::condition_variable m_event;
-                CallbackState m_state = CallbackState::enqueued;
+            using QueueImpl = uniform_cuda_hip::detail::QueueUniformCudaHipRtImpl<TApi>;
+
+            struct HostFuncData
+            {
+                QueueImpl& q; // We don't need to keep the queue alive, because in it's dtor it will synchronize with
+                              // the CUDA/HIP stream and wait until all host functions and the CallbackThread are done.
+                              // It's actually an error ot copy the queue into the host function. Destroying it here
```
fwyzard:
```suggestion
                              // It's actually an error to copy the queue into the host function. Destroying it here
```

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -185,86 +185,43 @@ namespace alpaka
...
+                              // would call CUDA/HIP APIs from the host function. Passing it further to the Callback
```
fwyzard: > Destroying it here would call CUDA/HIP APIs from the host function.

Why that would be a problem ?

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -185,86 +185,43 @@ namespace alpaka
...
-                        { return spCallbackSynchronizationData->m_state == CallbackState::finished; });
-                }
-            }
+                auto data = std::unique_ptr<HostFuncData>(reinterpret_cast<HostFuncData*>(arg));
+                auto& queue = data->q;
+                auto f = queue.m_callbackThread.submit(
+                    [data = std::move(data)]() mutable
+                    {
+                        data->t();
+                        data = nullptr; // destroy the task
```
fwyzard:
```suggestion
                        data.reset(); // destroy the task
```
looks nicer to me :-p

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
```diff
@@ -185,86 +185,43 @@ namespace alpaka
...
+                    });
+                f.wait();
```
fwyzard, line 212: Not suggesting a change, just trying to follow the logic: lines 205-212 are equivalent to
```c++
data->q.m_callbackThread.submit(...).wait();
```
?

fwyzard:
Is there some benchmark that can show the impact on the runtime performance ?

psychocoderHPC:
@bernhardmgruber based on this PR I added a few changes on top to silence all errors, I do not understand why some changes are required and would like to understand it.

https://github.com/bernhardmgruber/alpaka/pull/1

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
bernhardmgruber: Alright!

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
bernhardmgruber: Because the CUDA documentation for `cudaLaunchHostFunc` [says](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__EXECUTION.html#group__CUDART__EXECUTION_1g05841eaa5f90f27124241baafb3e856f):

> The host function must not make any CUDA API calls. Attempting to use a CUDA API may result in cudaErrorNotPermitted but this is not required.

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
bernhardmgruber: Yes. I wrote it the way I did to make the comment below be able to refer to the future `f`.

bernhardmgruber:
> @bernhardmgruber based on this PR I added a few changes on top to silence all errors, I do not understand why some changes are required and would like to understand it.
> 
> [bernhardmgruber#1](https://github.com/bernhardmgruber/alpaka/pull/1)

Thx! I will have a look when time allows!

include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp
psychocoderHPC: Could you please move the comment above the expression? This will make the lines more readable

psychocoderHPC (changes requested):

bernhardmgruber:
I found the cause of the ASAN errors. There was a cyclic reference of shared pointers in some event implementations of the form: shared_ptr -> eventImpl -> std::future -> task lambda -> captured shared_ptr. This created a leak that ASAN diagnosed.

bernhardmgruber:
> Could you please move the comment above the expression? This will make the lines more readable

done.

bernhardmgruber:
> Is there some benchmark that can show the impact on the runtime performance ?

I added one testing 1000 enqueues.

Before:
```
-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 0
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             4     2.0908 ms 
                                         5.1214 us     5.1144 us    5.13593 us 
                                        49.4037 ns    28.2211 ns    98.2177 ns 
                                                                               

-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 1
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1    78.2606 ms 
                                        961.536 us    953.695 us    966.312 us 
                                         30.501 us     21.397 us    53.1851 us 
                                                                               

-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 2
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1     1.21467 s 
                                        12.8079 ms    11.8978 ms     13.708 ms 
                                        4.61803 ms    4.24741 ms    5.02597 ms 
                                                                               

-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 3
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1    806.096 ms 
                                        8.02204 ms    7.98338 ms    8.05383 ms 
                                        178.694 us    151.852 us    202.262 us
```

After:
```
-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 0
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             4     2.1004 ms 
                                        5.05373 us    5.04876 us    5.06079 us 
                                        29.8689 ns    20.7356 ns    39.9793 ns 
                                                                               

-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 1
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1     21.604 ms 
                                        299.072 us    293.559 us    302.797 us 
                                        22.8706 us    16.7133 us    29.3113 us 
                                                                               

-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 2
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1     1.18284 s 
                                        11.5493 ms    11.5286 ms      11.57 ms 
                                        105.008 us    85.5805 us    137.062 us 
                                                                               

-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 3
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1    881.351 ms 
                                        11.0215 ms    11.0092 ms    11.0346 ms 
                                        64.5709 us    56.8905 us    77.8324 us 
```
Queue order of `alpaka::test::TestQueues` is: CPU blocking, CPU non-blocking, CUDA blocking, CUDA non-blocking.

Perf got faster for all queues except for the CUDA non-blocking.

fwyzard:
Any idea why CUDA async is somewhat slower ?
That's one of the cases we do care about :)

bernhardmgruber:
> Any idea why CUDA async is somewhat slower ?

I have no clue. `perf` shows less than 1% of the code is spent in the test executable. Most time is spent inside Catch2 and the CUDA libraries. A potential guess could be that previously, we launched the CUDA host func and enqueued to the background thread at the same time, so some kind of thread switching overhead could overlap. Now, I enqueue a host func to CUDA, which then enqueues to the callback thread. The thread switching does not overlap in this case.

Also the numbers seem to be very shaky, even though Catch2 already reports the mean of a 100 runs. Running again a couple of times, and only the async CUDA queue, now shows:

Before:
```
-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 3
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1    785.955 ms 
                                        10.7045 ms    10.6835 ms    10.7233 ms 
                                        100.935 us    76.4397 us    153.905 us 
```

After, 1st run:
```
-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 3
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1    753.501 ms 
                                        11.2029 ms    11.1918 ms    11.2191 ms 
                                        67.5108 us    49.7149 us    93.5378 us 
```

After, 2nd run:
```
-------------------------------------------------------------------------------
enqueueBenchmark - alpaka::test::TestQueues - 3
-------------------------------------------------------------------------------
/home/bgruber/dev/alpaka/test/unit/queue/src/QueueTest.cpp:201
...............................................................................

benchmark name                       samples       iterations    estimated
                                     mean          low mean      high mean
                                     std dev       low std dev   high std dev
-------------------------------------------------------------------------------
Enqueue 1000                                   100             1     524.13 ms 
                                        6.62121 ms    6.27566 ms     7.0501 ms 
                                        1.96068 ms     1.6434 ms    2.24793 ms 

```
The numbers are from my workstation. I think for a solid number one would need to run on a cluster.

> That's one of the cases we do care about :)

The entire changeset here only affects the implementation of host functions and does not affect memory copies or kernels. Are host functions really such a concern?

Also: this PR fixes several race conditions during the shutdown of alpaka and introduces consistent destruction and cleanup of resources as well as ensures all enqueued work is still finished. That is also a huge plus IMO.

jkelling:
> This PR causes one change in behavior: the destructor of a non-blocking queue will now block until all tasks in the queue are completed. This is required to ensure deterministic lifetime and destruction of queues, background threads, tasks, etc. If we are fine with this new behavior, we should also cover it by a unit test.

I do agree, that the asynchronous destruction logic of the non-blocking queue introduces a lot of complexity, but removing it does not exactly increase determinism, I think. The problem is, that the queue is a shared object and the actual destruction happens when the last reference is dropped, leaving which ever thread this happens in with the task of executing the destructor. Making the destructor block to wait for tasks to complete may cause very unpredictable blocking in multi-threaded codes., almost akin to the behavior of garbage collectors. This behavior would be quite different from that of, e.g., CUDA streams. Are we OK with this?

A way to work around this problem have a clear owner of the queue, so the destruction happens when that owner destroys it. The shared nature of the object implies a different intended use of the type, no user will not usually go for that. It could be enforced, by removing the shared aspect of the type, but this would put the burden of making sure the queue lives long enough on the user-code (.use-after-free would arise as a possible type of bug, although, the destructor could throw an exception if the queue is not empty, to catch this early.). It would also be inconsistent with all other alpaka object being shared.

bernhardmgruber:
> The problem is, that the queue is a shared object and the actual destruction happens when the last reference is dropped, leaving which ever thread this happens in with the task of executing the destructor.

This is ok, as long as that last thread is not the `CallbackThread`, which I documented.

> Making the destructor block to wait for tasks to complete may cause very unpredictable blocking in multi-threaded codes., almost akin to the behavior of garbage collectors.

Which is the behavior alpaka chose by using a `shared_ptr`. But at least the desctruction is guaranteed to take place, enqueued tasks are guaranteed to finish, and resources are not leaked.

> This behavior would be quite different from that of, e.g., CUDA streams. Are we OK with this?

I am absolutely ok with that.

> A way to work around this problem have a clear owner of the queue, so the destruction happens when that owner destroys it.

Which you can already do today by thinking more carefully about where to put your alpaka queues.

> It could be enforced, by removing the shared aspect of the type, but this would put the burden of making sure the queue lives long enough on the user-code

I think it's too late in alpaka's design to change that and make the queue e.g. a move-only type.

fwyzard:
> The entire changeset here only affects the implementation of host functions and does not affect memory copies or kernels. Are host functions really such a concern?

Yes and no :-)

For the CMS software they are a concern, because we use them to synchronise the state of the GPU work submitted by each "module" - so few hundred times per second. 

But, they are so much a concern that we are using a dedicated version instead of the stock one from Alpaka, so these changes should not impact us :-)

bernhardmgruber:
> But, they are so much a concern that we are using a dedicated version instead of the sock one from Alpaka

Wow, interesting! We would be happy if you would like to upstream them :)

> these changes should not impact us :-)

Great! I count this as a green light from your side then!

fwyzard:
Here is what we are using: https://github.com/cms-sw/cmssw/blob/master/HeterogeneousCore/AlpakaInterface/interface/HostOnlyTask.h .

The limitation is that one cannot call API functions inside the `HostOnlyTask` (because of the same limitation in CUDA).

psychocoderHPC:
> I found the cause of the ASAN errors. There was a cyclic reference of shared pointers in some event implementations of the form: shared_ptr -> eventImpl -> std::future -> task lambda -> captured shared_ptr. This created a leak that ASAN diagnosed.

Is this PR now fixing all sanitizer errors? I am wondering because the initial PR description is saying that some errors are fixed.

bernhardmgruber:
> > I found the cause of the ASAN errors. There was a cyclic reference of shared pointers in some event implementations of the form: shared_ptr -> eventImpl -> std::future -> task lambda -> captured shared_ptr. This created a leak that ASAN diagnosed.
> 
> Is this PR now fixing all sanitizer errors? I am wondering because the initial PR description is saying that some errors are fixed.

Running `ctest` with serial and CUDA backend and compiled with TSAN:
```
      Start  1: axpyTest
 1/30 Test  #1: axpyTest .........................   Passed    1.14 sec
      Start  2: hostOnlyAPITest
 2/30 Test  #2: hostOnlyAPITest ..................   Passed    1.13 sec
      Start  3: mandelbrotTest
 3/30 Test  #3: mandelbrotTest ...................   Passed    1.32 sec
      Start  4: matMulTest
 4/30 Test  #4: matMulTest .......................   Passed    1.14 sec
      Start  5: sharedMemTest
 5/30 Test  #5: sharedMemTest ....................   Passed    1.12 sec
      Start  6: zeroDimBufferTest
 6/30 Test  #6: zeroDimBufferTest ................   Passed    1.14 sec
      Start  7: accTest
 7/30 Test  #7: accTest ..........................   Passed    1.14 sec
      Start  8: atomicTest
 8/30 Test  #8: atomicTest .......................   Passed    1.14 sec
      Start  9: blockSharedTest
 9/30 Test  #9: blockSharedTest ..................   Passed    1.17 sec
      Start 10: blockSharedSharingTest
10/30 Test #10: blockSharedSharingTest ...........   Passed    1.13 sec
      Start 11: blockSyncTest
11/30 Test #11: blockSyncTest ....................   Passed    1.15 sec
      Start 12: coreTest
12/30 Test #12: coreTest .........................   Passed    0.02 sec
      Start 13: devTest
13/30 Test #13: devTest ..........................   Passed    1.12 sec
      Start 14: eventTest
14/30 Test #14: eventTest ........................***Failed    1.14 sec
      Start 15: idxTest
15/30 Test #15: idxTest ..........................   Passed    1.13 sec
      Start 16: intrinsicTest
16/30 Test #16: intrinsicTest ....................   Passed    1.16 sec
      Start 17: kernelTest
17/30 Test #17: kernelTest .......................   Passed    1.32 sec
      Start 18: mathTest
18/30 Test #18: mathTest .........................   Passed    2.00 sec
      Start 19: memBufTest
19/30 Test #19: memBufTest .......................***Failed    1.34 sec
      Start 20: bufSlicingTest
20/30 Test #20: bufSlicingTest ...................   Passed    1.16 sec
      Start 21: FenceTest
21/30 Test #21: FenceTest ........................   Passed    1.14 sec
      Start 22: memViewTest
22/30 Test #22: memViewTest ......................   Passed    1.55 sec
      Start 23: memP2PTest
23/30 Test #23: memP2PTest .......................   Passed    1.05 sec
      Start 24: metaTest
24/30 Test #24: metaTest .........................   Passed    1.14 sec
      Start 25: queueTest
25/30 Test #25: queueTest ........................***Failed   20.81 sec
      Start 26: randTest
26/30 Test #26: randTest .........................   Passed    1.13 sec
      Start 27: traitsTest
27/30 Test #27: traitsTest .......................   Passed    1.12 sec
      Start 28: vecTest
28/30 Test #28: vecTest ..........................   Passed    0.01 sec
      Start 29: warpTest
29/30 Test #29: warpTest .........................   Passed    1.28 sec
      Start 30: workDivTest
30/30 Test #30: workDivTest ......................   Passed    1.14 sec
```
The eventTest fails a Catch2 check, the memBufTest has 1 TSAN warning, the queueTest has 13 warnings. That's not completely clean, but a lot better than before.

psychocoderHPC (approved):

PR #1871, opened 2022-12-14
add support for agc-manager

SimeonEhrig:
The agc-manager is part of the alpaka-group-container and allow to search for preinstalled software.


script/before_install.sh
```diff
@@ -30,14 +31,6 @@ then
     fi
 fi
 
-#-------------------------------------------------------------------------------
-# Boost.
-echo $ALPAKA_CI_BOOST_BRANCH
```
psychocoderHPC, line 26: The way to split strings in bash without assuming always the correct number of digits is https://stackoverflow.com/a/5257398

SimeonEhrig: Do you mean, that is the sense of this chunk of code? An example for spiting strings. If yes, I can bring it back and comment it. I removed it, because it has no usage and confused me.

psychocoderHPC: I looked to the wrong side of the diff^^ NO need to reintroduce the version splitting!

script/gitlab_ci_run.sh
```diff
@@ -12,6 +12,15 @@
...
 
+# inside the agc-container, the user is root and does not require sudo
+# to compatibility to other container, fake the missing sudo command
+if ! command -v sudo &> /dev/null
+then
+    cp ${CI_PROJECT_DIR}/script/gitlabci/fake_sudo.sh /usr/bin/sudo
+    chmod +x /usr/bin/sudo
+fi
+
+# TODO(sehrig): remove me
```
j-stephan, line 23: Huh?

j-stephan (dismissed):

script/gitlab_ci_run.sh
SimeonEhrig: It is a workaround for old HIP images. When I drop support for HIP 4.x I remove this part of code. Actual, I should extend the comment but the next PR is removing the old HIP container. Therefore, I would let it as it is.

j-stephan (approved):

Issue #1753, opened 2022-06-27
clang-16: ASan fails for `memBufTest`.


j-stephan:
As part of #1713 I encountered an address sanitizer issue with clang-14. This doesn't happen with clang-13 (tested locally). Error message:

```
30: Test command: /home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest "--colour-mode" "default"
30: Test timeout computed to be: 1500
30: =================================================================
30: ==10579==ERROR: AddressSanitizer: alloc-dealloc-mismatch (operator new vs free) on 0x60700000ddf0
30:     #0 0x5570bd780102 in __interceptor_free (/home/runner/work/alpaka/alpaka/build/test/unit/mem/buf/memBufTest+0xf4102) (BuildId: dbbfef11bf08aba8be17c096a1867d9632ca5f95)
30:     #1 0x7f64a009e25d in std::invalid_argument::~invalid_argument() (/usr/lib/llvm-14/lib/libc++abi.so.1+0x2425d) (BuildId: 9e9a6e2998ed7d6781edcc2eb856fba4529c1f05)
30:     #2 0x7f64a00a05d3 in __cxa_end_catch (/usr/lib/llvm-14/lib/libc++abi.so.1+0x265d3) (BuildId: 9e9a6e2998ed7d6781edcc2eb856fba4529c1f05)
30:     #3 0x5570bd8f8895 in void testBufferAccessorAdaptor<alpaka::AccCpuSerial<std::__1::integral_constant<unsigned long, 1ul>, unsigned long> >(alpaka::Vec<alpaka::trait::DimType<alpaka::AccCpuSerial<std::__1::integral_constant<unsigned long, 1ul>, unsigned long>, void>::type, alpaka::trait::IdxType<alpaka::AccCpuSerial<std::__1::integral_constant<unsigned long, 1ul>, unsigned long>, void>::type> const&, alpaka::Vec<alpaka::trait::DimType<alpaka::AccCpuSerial<std::__1::integral_constant<unsigned long, 1ul>, unsigned long>, void>::type, alpaka::trait::IdxType<alpaka::AccCpuSerial<std::__1::integral_constant<unsigned long, 1ul>, unsigned long>, void>::type> const&) /home/runner/work/alpaka/alpaka/test/unit/mem/buf/src/BufTest.cpp:268:9
[...]

PR #1872, opened 2022-12-14
Drop legacy compilers and CUDA versions

### LINKED ISSUE #1866: Changes to GitHub Ubuntu CI runners
The GitHub Ubuntu CI runners will see some changes in the near future:

* Ubuntu 18.04 is deprecated - actions/runner-images/issues/6002
* `ubuntu-latest` will switch from Ubuntu 20.04 to Ubuntu 22.04 - actions/runner-images/issues/6399

Foreseeable changes for us when upgrading from 18.04 to 20.04:

* gcc 7, 8 will come from Ubuntu's `universe` repository. 8 appears outdated (latest version is 8.5 but `universe` only has 8.4). The [PPA](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/test) we used for installing more recent GCC versions appears out of date. We should consider switching to [another PPA](https://launchpad.net/~ubuntu-toolchain-r/+archive/ubuntu/ppa) but that one unfortunately doesn't have gcc 8 or older.
* clang 6, 7, 8 will come from Ubuntu's `universe` repository since there are no official LLVM packages for these versions. 6 and 8 seem to be up to date but 7 appears outdated (7.0.1 in `universe` vs 7.1.0 in LLVM's repository for Ubuntu 18.04).

Ubuntu 18.04 will be EOL in April 2023. My suggestion:

* Drop support for gcc 7 and 8.
* Drop support for clang 6, 7, 8.
* Drop support for CUDA 10.x since it is officially only supported on Ubuntu 18.04.

j-stephan:
This PR removes legacy compilers and CUDA versions.

* All GitHub jobs previously set to `ubuntu-latest` are now hardcoded to `ubuntu-20.04`.
* ~~Added GitLab base configurations for clang on `ìntel`, `amd` and `arm64` runners.~~ ARM will be added in a separate PR.
* GitHub job `linux_gcc-8_debug_analysis` replaced by existing GitHub job `linux_gcc-12_debug_analysis`.
* GitHub job `linux_clang-8_debug_analysis` replaced by new GitLab job `linux_clang-14_debug_analysis`.
* GitHub job `linux_gcc-7_release` replaced by existing GitHub job `linux-gcc-10_release`.
* GitHub job `linux_gcc-8_debug` replaced by existing GitHub job `linux_gcc-9_debug` which was renamed from `linux_gcc-9_debug_c++17`.
* GitHub job `linux_clang-6_release_asan_c++17` replaced by new GitLab job `linux_clang-14_release_asan`. Note that this includes a switch from `libc++` to `libstdc++`. See #1753.
* GitHub job `linux_clang-7_release_c++17` replaced by new GitLab job `linux_clang-11_release`.
* GitHub job `linux_clang-8_release` replaced by existing GitHub job `linux_clang-10_release`.
* Removed all CUDA 10.x GitLab jobs.
* Removed all occurences of the `-Wmissing-variable-declarations` suppression which was necessary for clang 6.
* Removed check for GCC <= 6 from `script/before_install.sh`.

Fixes #1866.


SimeonEhrig (dismissed):

bernhardmgruber (dismissed):

script/gitlabci/job_base.yml
```diff
@@ -145,3 +145,41 @@
...
+    alpaka_ACC_GPU_HIP_ENABLE: "OFF"
+    alpaka_ACC_GPU_HIP_ONLY_MODE: "OFF"
+  extends: .base
+  tags:
+    - cpuonly
+
+.base_clang_legacy:
+  extends: .basecfg_clang
+  tags:
+    - intel
```
SimeonEhrig: What is the intention of the flags? At the moment, you force the job to use the CPUs of the GPU runners. If you want to use a x86 CPU, you should take the tag `x86_64`. Also you should use the tag `cpuonly`. This avoids, that the limited GPU runner resources are used for a CPU only job.

j-stephan: According to [HIFIS](https://hifis.net/doc/software/gitlab/ci/#specific-runners) `x86_64` is the same as `intel`, leaving AMD CPUs out of our testing setup. Am I understanding this wrong?

Also, is there documentation for the `cpuonly` tag?

j-stephan:
Okay, this needs an architecture detection mechanism. Currently we have hardcoded `x86_64` in various places which fails our runs on HIFIS' ARM runners.

script/gitlabci/job_base.yml
SimeonEhrig: This documentation is partial updated/incomplete. Go to the runner settings* in GitLab and expand the runners section. This is our big CPU machine:

image

* you need to be logged in, that the link is working: https://gitlab.com/hzdr/crp/alpaka/-/settings/ci_cd

SimeonEhrig:
> Okay, this needs an architecture detection mechanism. Currently we have hardcoded `x86_64` in various places which fails our runs on HIFIS' ARM runners.

Ether you set an environment variable in the GitLab job yaml or you use the `CI_RUNNER_TAGS` variable.

j-stephan:
So clang 13 ASan also fails now. This warrants some investigation...

SimeonEhrig:
@j-stephan I believe you need to push again. GitHub Workflows stucks.

j-stephan:
No, the pending checks are those that are removed by this PR. @psychocoderHPC needs to disable those and then merge the PR.

script/job_generator/generate_job_yaml.py
```diff
@@ -113,7 +113,7 @@ def job_tags(job: Dict[str, Tuple[str, str]]) -> List[str]:
         and job[ALPAKA_ACC_GPU_HIP_ENABLE][VERSION] != OFF
     ):
         return ["x86_64", "rocm"]
-    return ["x86_64", "cpuonly"]
+    return ["cpuonly"]
```
SimeonEhrig: I think this is wrong. If you add only the `cpuonly` tag, the GitLab CI will decide, if a job is running on x86 or ARM. This means also, if you restart a failed job it is possible, that the job failed on x86 and is restarted on a ARM runner. At the moment, the function is not used, so it will not affect the CI. Also, the job generator needs an extension, to decide if a job is executed on a x86 CPU or a ARM CPU (it is required, that you can implement filter rules like clang x is not supporting ARM).

Because this is bigger change, I suggest to revert the change and add a todo: `# TODO(j-stephan): extend cpuonly tags for the ARM runner`

bernhardmgruber: Does it matter whether we run on x86 or ARM? We also do not differentiate between Intel and AMD CPUs, or what kind of instruction sets the CPU supports.

SimeonEhrig: No, that's not true. In practice, there is platform depending code, which is not so well tested on ARM because of the much smaller user base. 

I have some personal experience with my Raspberry PI. And it is still a problem. I had the case this weekend, that a flashing tool does not worked on the PI but on my x86 laptop. And the flashing tool should not use processor specific functions, such like vector commands.

psychocoderHPC: We should explicitly test ARM and x86. One problem is that we need different base containers, at least if we use the pre built containers @SimeonEhrig is shipping.
The pre-built containers are required to reduce the runtime of a single job.

bernhardmgruber: ok, I guess then we should pin jobs to specific architectures.

psychocoderHPC: In general, I am against supporting ARM as long as we have no access to an ARM systems to reproduce issues.

script/gitlabci/job_base.yml
```diff
@@ -145,3 +145,41 @@
...
+  tags:
+    - cpuonly
+    - x86_64 # aarch64 only supported for clang versions >= 12
+
+.base_clang: 
+  extends: .basecfg_clang
+  variables:
+    ALPAKA_CI_CMAKE_VER: 3.19.8 # Don't go below 3.19.3, there are no CMake binaries for ARM
+  tags:
+    - cpuonly
```
SimeonEhrig: It's nearly random if a job basing on this base job is executed on x86 or ARM. This means also, that the CI can run all jobs on x86.

j-stephan: Yes, randomness is what I'm going for here. We should reliably run across architectures.

script/job_generator/generate_job_yaml.py
j-stephan: We have access to an A64fx system (well, I do).

j-stephan: But we can also pin our jobs to a specific architecture if that is the broader consensus. Or just support x86 for the time being. We are somewhat democratic so this is a majority decision.

script/gitlabci/job_base.yml
SimeonEhrig: If we use the pair-wise generator, we don't need randomness anymore. The idea of the pair-wise generator is, that each pair of parameters occurs at least one time in the matrix. So if we put the processor architecture as additional parameter, the generator guaranties, that the each supported version of a compiler is tested with ARM.

SimeonEhrig (approved):

SimeonEhrig:
@j-stephan and hat a small offline discussion and decided to move the ARM support in extra PR, because the PR #1871 depends on this PR

psychocoderHPC:
The icpx compiler test shows

```
/home/runner/work/alpaka/alpaka/test/integ/axpy/src/axpy.cpp:127:46: error: implicit conversion between floating point types of different sizes [-Werror,-Wimplicit-float-size-conversion]
    std::uniform_real_distribution<Val> dist(0.0, 1.0);
```

psychocoderHPC:
> The icpx compiler test shows
> 
> ```
> /home/runner/work/alpaka/alpaka/test/integ/axpy/src/axpy.cpp:127:46: error: implicit conversion between floating point types of different sizes [-Werror,-Wimplicit-float-size-conversion]
>     std::uniform_real_distribution<Val> dist(0.0, 1.0);
> ```

should be fixed with #1884

bernhardmgruber (approved):

j-stephan:
FYI @psychocoderHPC: @SimeonEhrig @bernhardmgruber and I decided to force-merge this since the failing Intel jobs are unrelated.

PR #1878, opened 2022-12-16
Fix undefined constant with nvcc 12.0

### LINKED ISSUE #1877: Compile error with nvcc 12 and gcc 12
When I compile the unit tests with nvcc 12 and gcc 12 I get the following error:

```
/home/bgruber/dev/alpaka/include/alpaka/test/event/EventHostManualTrigger.hpp(383): error: identifier "CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS" is undefined

1 error detected in the compilation of "/home/bgruber/dev/alpaka/test/unit/event/src/EventTest.cpp".
```
Maybe this constant has been removed or moved to a different header?

bernhardmgruber:
Fixes: #1877

### DIFF
--- a/include/alpaka/test/event/EventHostManualTrigger.hpp
+++ b/include/alpaka/test/event/EventHostManualTrigger.hpp
@@ -380,7 +380,14 @@ namespace alpaka::test
             ALPAKA_FN_HOST static auto isSupported(DevCudaRt const& dev) -> bool
             {
                 int result = 0;
-                cuDeviceGetAttribute(&result, CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS, dev.getNativeHandle());
+                cuDeviceGetAttribute(
+                    &result,
+#    if CUDA_VERSION >= 12000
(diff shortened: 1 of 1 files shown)


j-stephan (commented):
Is there no replacement for the deprecated enum value? I would rather see this modernized instead of failing in the future.

bernhardmgruber:
I don't know what the replacement is and I neither have the time nor expertise to look into this. But the issue is blocking me, so here is a fix that at least allows compilation to succeed again with nvcc 12.0.

psychocoderHPC (approved):

psychocoderHPC:
I will merge this PR to avoid blocking others. It is normal that NVIDIA is removing defines. 
This is a driver API definition that is only used for testing. In general, we should maybe replace the usage in the test by using system atomics, if I remember correctly we use this flag for the queue or event test only to block a task and release it in a well-defined order.

psychocoderHPC:
I opened an issue that we should check if the usage of this driver API define can be replaced: https://github.com/alpaka-group/alpaka/issues/1881

PR #1883, opened 2022-12-19
Use `hipLaunchHostFunc` with HIP 5.4.0 and later

fwyzard:
HIP 5.2.0 [introduced](https://docs.amd.com/en-US/bundle/ROCm-Release-Notes-v5.2/page/What%E2%80%99s_New_in_This_Release.html#d678e602) `hipLaunchHostFunc` as the equivalent to `cudaLaunchHostFunc`.

However, while `hip_runtime_api.h` does define it starting from ROCm 5.2.0, `libamdhip64.so` implements it only starting from ROCm 5.4.0 .

### DIFF
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -263,8 +263,14 @@ namespace alpaka
 
         static inline Error_t launchHostFunc(Stream_t stream, HostFn_t fn, void* userData)
         {
+            // hipLaunchHostFunc is implemented only in ROCm 5.4.0 and later.
+#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(5, 4, 0)
+            // Wrap the host function using the proper calling convention.
+            return ::hipLaunchHostFunc(stream, HostFnAdaptor::hostFunction, new HostFnAdaptor{fn, userData});
+#    else
             // Emulate hipLaunchHostFunc using hipStreamAddCallback with a callback adaptor.
             return ::hipStreamAddCallback(stream, HostFnAdaptor::streamCallback, new HostFnAdaptor{fn, userData}, 0);
+#    endif
         }
 
         static inline Error_t malloc(void** devPtr, size_t size)


bernhardmgruber (dismissed):

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -263,8 +263,13 @@ namespace alpaka
 
         static inline Error_t launchHostFunc(Stream_t stream, HostFn_t fn, void* userData)
         {
+#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(5, 2, 0)
+            // Wrap the host function using the proper calling convention
+            return ::hipLaunchHostFunc(stream, HostFnAdaptor::hostFunction, new HostFnAdaptor{fn, userData});
```
psychocoderHPC, line 269: Mhh I am wondering who is freeing the user data, in this case `fn` and `userData`?
@fwyzard do you now if this is increasing the memory footprint over time of alpaka if host functions you very extensively?
This is not the scope of this PR but this question came to my mind.

psychocoderHPC (dismissed):
I have one question not related to the intention of this PR.

psychocoderHPC:
The failing icpx runs should be fixed by #1884, never the less there is a linker issue with HIP 5.2

psychocoderHPC:
I am confused the release notes 5.4.1 say 

```
New HIP API - hipLaunchHostFunc 
The following new HIP API is introduced in the ROCm v5.4.1 release.
```
https://docs.amd.com/bundle/ROCm-Release-Notes-v5.4.1/page/What%E2%80%99s_New_in_This_Release.html

but the link you provided to HIP 5.2.0 also mentioned this function.

psychocoderHPC:
I checked the source code and release 5.4.x is the first release which defines the new function: https://github.com/ROCm-Developer-Tools/hipamd/blob/rocm-5.4.x/include/hip/amd_detail/amd_hip_runtime_pt_api.h#L69

include/alpaka/core/ApiHipRt.hpp
fwyzard: > Mhh I am wondering who is freeing the user data, in this case `fn` and `userData`?

The call to `hostFunction` frees the `HostFnAdaptor` object passed to it: https://github.com/alpaka-group/alpaka/blob/39f67098802f3266fd88de58c7e1fb75da8c6a89/include/alpaka/core/ApiHipRt.hpp#L102-L107

fwyzard: `fn` is a pointer to a function, so it doesn't need to (cannot) be freed.
`userData` is a type-punned pointer to user-provided data, so it's up to the user to free it.

fwyzard: `launchHostFunc` is used in https://github.com/alpaka-group/alpaka/blob/39f67098802f3266fd88de58c7e1fb75da8c6a89/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp#L243-L246

There, `uniformCudaHipRtHostFunc` is a `static` function that notifies the worker thread to run the actual host function:
https://github.com/alpaka-group/alpaka/blob/39f67098802f3266fd88de58c7e1fb75da8c6a89/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp#L213-L225
_etc_.

The data passed to the host function is managed by the shared stated managed by the `spCallbackSynchronizationData`, which should be cleared once worker thread has run the actual host function, and the callback has complete.

fwyzard: So, it looks like all objects should be properly cleaned up.

psychocoderHPC: thanks for digging into it!

fwyzard:
Rebased to re-trigger the CIs.

fwyzard:
> I checked the source code and release 5.4.x is the first release which defines the new function: https://github.com/ROCm-Developer-Tools/hipamd/blob/rocm-5.4.x/include/hip/amd_detail/amd_hip_runtime_pt_api.h#L69

I see.

Indeed, `hip_runtime_api.h` does define it starting from ROCm 5.2.0, but `libamdhip64.so` implements it only starting from ROCm 5.4.0 .

fwyzard:
The failure seems unrelated to the changes.

j-stephan:
Yes, this is caused by a driver update in our CI. NVHPC doesn't like newer drivers unfortunately. This should be fixed by #1890 - once that is merged on Monday or so the error should go away.

fwyzard:
Rebased on top of #1890 (merged).

bernhardmgruber (dismissed):

j-stephan:
Please rebase, the CI should run again!

bernhardmgruber (approved):

j-stephan (approved):

PR #1884, opened 2022-12-20
fix icpx error implicit conversion

### LINKED ISSUE #1882: Intel oneAPI CI broken
The Intel apt repository just updated from `intel-oneapi-compiler-dpcpp-cpp-and-cpp-classic-2022.2.1` to 2023 and now the corresponding CI jobs are failing.

psychocoderHPC:
```
test/integ/axpy/src/axpy.cpp:127:46: error: implicit conversion between floating point types of different sizes [-Werror,-Wimplicit-float-size-conversion]
8627
    std::uniform_real_distribution<Val> dist(0.0, 1.0);
8628
                                        ~~~~ ^~~
```

### DIFF
--- a/test/integ/axpy/src/axpy.cpp
+++ b/test/integ/axpy/src/axpy.cpp
@@ -124,7 +124,7 @@ TEMPLATE_LIST_TEST_CASE("axpy", "[axpy]", TestAccs)
     std::random_device rd{};
     auto const seed = rd();
     std::default_random_engine eng{seed};
-    std::uniform_real_distribution<Val> dist(0.0, 1.0);
+    std::uniform_real_distribution<Val> dist(Val{0}, Val{1});
     std::cout << "using seed: " << seed << "\n";
     // Initialize the host input vectors
     for(Idx i(0); i < numElements; ++i)
--- a/test/unit/math/src/sincos.cpp
+++ b/test/unit/math/src/sincos.cpp
@@ -32,8 +32,8 @@ public:
         // (PTX kernel (float) was just empty)
         FP check_sin = alpaka::math::sin(acc, arg);
         FP check_cos = alpaka::math::cos(acc, arg);
-        FP result_sin = 0.;
-        FP result_cos = 0.;
+        auto result_sin = FP{0};
+        auto result_cos = FP{0};
         alpaka::math::sincos(acc, arg, result_sin, result_cos);
         using alpaka::test::unit::math::almost_equal;
         ALPAKA_CHECK(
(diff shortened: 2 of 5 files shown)


SimeonEhrig (dismissed):

test/integ/axpy/src/axpy.cpp
```diff
@@ -124,7 +124,7 @@ TEMPLATE_LIST_TEST_CASE("axpy", "[axpy]", TestAccs)
     std::random_device rd{};
     auto const seed = rd();
     std::default_random_engine eng{seed};
-    std::uniform_real_distribution<Val> dist(0.0, 1.0);
+    std::uniform_real_distribution<Val> dist(static_cast<Val>(0.0), static_cast<Val>(1.0));
```
bernhardmgruber: I would prefer a shorter version, but as you prefer.
```suggestion
    std::uniform_real_distribution<Val> dist(Val{0}, Val{1});
```

bernhardmgruber (dismissed):

bernhardmgruber (dismissed):

j-stephan (dismissed):

bernhardmgruber:
@psychocoderHPC I rebased your PR on develop again.

j-stephan:
This PR also needs to address the installation of TBB (or we handle this in a separate PR). Apparently we are hardcoded to 2021.4 which is no longer available.

psychocoderHPC:
> This PR also needs to address the installation of TBB (or we handle this in a separate PR). Apparently, we are hard coded to 2021.4 which is no longer available.

This is independent of the current conversion issue and should be a separate PR. I am wondering why the CI is still working if 2021.4 is not available anymore.

j-stephan:
I think most of our CI jobs might not have the TBB back-end enabled. But one of the previous runs failed because it couldn't find TBB 2021.4: https://github.com/alpaka-group/alpaka/actions/runs/3739898437/jobs/6347674062

test/unit/math/src/sincos.cpp
```diff
@@ -32,8 +32,8 @@ class SinCosTestKernel
         // (PTX kernel (float) was just empty)
         FP check_sin = alpaka::math::sin(acc, arg);
         FP check_cos = alpaka::math::cos(acc, arg);
-        FP result_sin = 0.;
-        FP result_cos = 0.;
+        FP result_sin = FP{0};
+        FP result_cos = FP{0};
```
j-stephan:
```suggestion
        auto result_sin = FP{0};
        auto result_cos = FP{0};
```

Only if the CI fails again.

bernhardmgruber (approved):

PR #1887, opened 2022-12-21
Drop MSVC 2019

### LINKED ISSUE #1879: [RFC] Reduce Windows CI
To reduce the maintenance effort and CI load, I suggest to reduce the tested combinations for MSVC. I would even go as far as to only support the latest version of Visual Studio (2022) with and without the latest version of nvcc. That is good enough for developing and testing on Windows, since I don't believe users are actually using Windows in production.

j-stephan:
This PR drops MSVC 2019 and the CUDA versions requiring MSVC 2019 from our CI. The remaining MSVC tests were renamed from `${runner}_c++17` to `${runner}`. It also fixes the entry for TBB in the supported back-end table in `README.md`. It was listed as unsupported but we are actually testing it in some CI runs.

Fixes #1879.


SimeonEhrig (dismissed):

bernhardmgruber (dismissed):

SimeonEhrig (approved):

bernhardmgruber (approved):

SimeonEhrig:
@j-stephan I think you need to hit the merge button by yourself. Because of the pending jobs (which are actual removed), the merge button is disabled for me.

PR #1889, opened 2022-12-22
add HIP to the CI generator parameter

SimeonEhrig:
The Hip jobs are the first kind of jobs, which are automatically generated by the job generator. Because we test all supported CMake and Boost versions too, the initial amount of generated jobs is much more, than the number of hand written HIP jobs, which we used before. Because of the properties of the n-wise combination generator, if we add more compilers to the generator (like nvcc, gcc, clang as C++ and CUDA compiler), each new compiler will create less additional new jobs until we arrive a stable number of maximum jobs (depends on the number of combination parameters).


script/job_generator/alpaka_globals.py
```diff
@@ -0,0 +1,17 @@
+"""This module contains constants used for the alpaka job generation.
```
j-stephan: Shouldn't we have a copyright header here?

script/job_generator/alpaka_globals.py
```diff
@@ -0,0 +1,17 @@
...
+# additional alpaka specific parameters
+BUILD_TYPE: str = "build_type"
+TEST_TYPE: str = "test_type"
+
+# possible values of BUILD_TYPE
+CMAKE_RELEASE: str = "Release"
+CMAKE_DEBUG: str = "Debug"
+BUILD_TYPES: List[str] = [CMAKE_RELEASE, CMAKE_DEBUG]
+
+# possible values of TEST_TYPE
```
j-stephan, line 24: What about analysis?

j-stephan (commented):
Nice! Just two questions:

script/job_generator/alpaka_globals.py
SimeonEhrig: I only transformed the HIP jobs from hand written to auto generated. There was no analysis job. I would add it, if we transform a job, which is doing analysis.

script/job_generator/alpaka_globals.py
SimeonEhrig: I can add it.

SimeonEhrig:
I will handle this 3 TODO's in a extra PR:

https://github.com/alpaka-group/alpaka/blob/076eb05bc1f49d06ee079e1724345bdb214d82e2/script/job_generator/generate_job_yaml.py#L171-L179

https://github.com/alpaka-group/alpaka/blob/076eb05bc1f49d06ee079e1724345bdb214d82e2/script/job_generator/generate_job_yaml.py#L245-L246

https://github.com/alpaka-group/alpaka/blob/076eb05bc1f49d06ee079e1724345bdb214d82e2/script/job_generator/generate_job_yaml.py#L247-L248

j-stephan (approved):

PR #1890, opened 2022-12-22
CUDA CI update

j-stephan:
This is a successor PR to #1713.

This PR updates the CUDA portion of our CI.

* `alpakaCommon.cmake`:
    * Update clang version requirement to clang-9. This was forgotten in #1872.
    * Updated clang-as-CUDA-compiler version requirement in to clang-14. This is the first version to officially
      support CUDA 11.x. Updated all corresponding GitLab CI jobs accordingly.
    * Added a warning for clang-as-CUDA-compiler about the usage of CUDA 11.3. Fixes #1857.
* Standalone headers:
    * Added a error message when using clang-as-CUDA-compiler together with CUDA 11.3. Fixes #1857.
* General 11.x:
  * Upgraded from gcc-7 and gcc-8 to gcc-9 (<= 11.3) and above (> 11.3).
  * Reestablished gcc-10 jobs. Fixes #1297.
  * Upgraded clang-as-host jobs from clang-8 to clang-9 (<= 11.2) and above (>= 11.6).
  * The CUDA `11.{3,4,5}` packages for Ubuntu 20.04 were never updated to not pull in gcc-9.3 and there are no
    packages for Ubuntu 22.04. The clang-as-host-jobs will be permanently disabled. Closes #1630.
* CUDA 11.1:
  * Removed clang-10-as-host job. CUDA officially only supports clang-9 in this version.
* CUDA 11.6:
  * Updated to 11.6.2.
  * Removed clang-13-as-host job. CUDA officially only supports clang-12 in this version.
* Added CUDA 11.7 and NVHPC 22.9 jobs. Fixes #1724.
* Added CUDA 11.8 and NVHPC 22.11 jobs. Fixes #1807.
* Added CUDA 12.0 jobs. Fixes #1864.
  * Upgraded Windows CUDA jobs to 12.0.
* Install gcc from `ubuntu-toolchain-r/ppa` instead of `ubuntu-toolchain-r/test`. The former features newer gcc versions.
* Removed redundant jobs (like `gcc-9_release_compile_only` vs. `gcc-9_release`).
* Updated to latest CMake point releases in all modified jobs.
* Added corresponding entries for all of the above to `README.md`.

PR #1894, opened 2023-01-07
Use `hipMallocAsync` and `hipFreeAsync` with HIP 5.2.0 and later

fwyzard:
HIP 5.2.0 [introduced](https://docs.amd.com/en-US/bundle/ROCm-Release-Notes-v5.2/page/What’s_New_in_This_Release.html#d678e211) `hipMallocAsync` and `hipFreeAsync` as the equivalent of `cudaMallocAsync` and `cudaFreeAsync`.

### DIFF
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -183,10 +183,15 @@ namespace alpaka
             return ::hipFree(devPtr);
         }
 
-        static inline Error_t freeAsync(void* /* devPtr */, Stream_t /* stream */)
+        static inline Error_t freeAsync([[maybe_unused]] void* devPtr, [[maybe_unused]] Stream_t stream)
         {
+            // hipFreeAsync is implemented only in ROCm 5.2.0 and later.
+#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(5, 2, 0)
+            return ::hipFreeAsync(devPtr, stream);
+#    else
             // Not implemented.
             return errorUnknown;
+#    endif
         }
 
         static inline Error_t funcGetAttributes(FuncAttributes_t* attr, void const* func)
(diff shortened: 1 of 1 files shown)


fwyzard:
The failure seems unrelated to the changes.

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -183,10 +183,15 @@ namespace alpaka
             return ::hipFree(devPtr);
         }
 
-        static inline Error_t freeAsync(void* /* devPtr */, Stream_t /* stream */)
+        static inline Error_t freeAsync([[maybe_unused]] void* devPtr, [[maybe_unused]] Stream_t stream)
         {
+            // hipFreeAsync is implemented only in ROCm 5.2.0 and later.
+#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(5, 2, 0)
+            return ::hipFreeAsync(devPtr, stream);
+#    else
```
j-stephan, line 191: Why not like this?

```suggestion
            // hipFreeAsync is implemented only in ROCm 5.2.0 and later.
    if constexpr(BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(5, 2, 0))
            return ::hipFreeAsync(devPtr, stream);
    else
```

fwyzard: both branches of `if constexpr` must be able to _compile_ - which is not the case on older versions of ROCm, where `hipFreeAsync` is not defined.

fwyzard: _e.g._ `test.cc`
```c++
#include <iostream>

int main(void) {

  if constexpr(false) {
    does_not_exist();
  } else {
    std::cout << "Hello world\n";
  }

  return 0;
}
```
fails to compile:
```bash
$ g++ test.cc -O2 -std=c++17 -o test
test.cc: In function ‘int main()’:
test.cc:6:5: error: ‘does_not_exist’ was not declared in this scope
    6 |     does_not_exist();
      |     ^~~~~~~~~~~~~~
```

j-stephan: Ah, right. My bad.

fwyzard:
Rebased on top of #1890 (merged).

bernhardmgruber (dismissed):

bernhardmgruber (approved):

j-stephan (approved):

psychocoderHPC (approved):

Issue #1897, opened 2023-01-09
Remove `experimental::Accessor`


bernhardmgruber:
We added `alpaka::experimental::Accessor` and some related classes to support SYCL as a backend for SYCL versions without pointers to device memory on the host (USM). Since #1482 was merged, this functionality can just as well be provided by `std::mdspan`. Furthermore, work is currently ongoing on reengineering the SYCL backend to require USM and thus also make `alpaka::experimental::Accessor` obsolete. See #1845.

Depending on whether USM will be required for the SYCL backend, we may need to add specializations for `alpaka::GetMdspan` to create `mdspan`s from SYCL buffers/accessors, or it may just work out of the box with USM pointers. In both cases, `alpaka::experimental::Accessor` is no longer required.

I would suggest we wait for the USM reengineering to finish and then remove `alpaka::experimental::Accessor`.

Depends on:

- [x] #1845

PR #1910, opened 2023-01-31
Remove experimental namespace from SYCL

Parsifal-2045:
This is a supporting PR for #1845 
It only removes all the experimental namespaces related to SYCL. Memory operations are currently not working (the bufCopy example won't even compile). That is being fixed by the main PR but seems to be related to the importing of mdspan.

### DIFF
--- a/include/alpaka/mem/fence/MemFenceGenericSycl.hpp
+++ b/include/alpaka/mem/fence/MemFenceGenericSycl.hpp
@@ -10,7 +10,7 @@
 
 #    include <CL/sycl.hpp>
 
-namespace alpaka::experimental
+namespace alpaka
 {
     namespace detail
     {
@@ -49,17 +49,17 @@ namespace alpaka::experimental
         sycl::accessor<int, 1, sycl::access_mode::read_write, sycl::target::global_buffer> m_global_dummy;
         sycl::accessor<int, 1, sycl::access_mode::read_write, sycl::target::local> m_local_dummy;
     };
-} // namespace alpaka::experimental
+} // namespace alpaka
 
 namespace alpaka::trait
 {
     template<typename TMemScope>
-    struct MemFence<experimental::MemFenceGenericSycl, TMemScope>
+    struct MemFence<MemFenceGenericSycl, TMemScope>
     {
-        static auto mem_fence(experimental::MemFenceGenericSycl const& fence, TMemScope const&)
+        static auto mem_fence(MemFenceGenericSycl const& fence, TMemScope const&)
         {
-            static constexpr auto scope = experimental::detail::SyclFenceProps<TMemScope>::scope;
-            static constexpr auto space = experimental::detail::SyclFenceProps<TMemScope>::space;
+            static constexpr auto scope = detail::SyclFenceProps<TMemScope>::scope;
+            static constexpr auto space = detail::SyclFenceProps<TMemScope>::space;
 
             // atomic_ref is already part of the SYCL spec but oneAPI has not caught up yet.
             auto dummy
--- a/include/alpaka/mem/fence/MemFenceGenericSycl.hpp
+++ b/include/alpaka/mem/fence/MemFenceGenericSycl.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Jan Stephan
+/* Copyright 2022 Jan Stephan, Luca Ferragina
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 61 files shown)


include/alpaka/mem/view/Accessor.hpp
```diff
@@ -23,7 +23,7 @@
 
 #include <tuple>
 
-namespace alpaka::experimental
```
psychocoderHPC: Please check that you move only SYCL stuff out of the experimental namespace.

include/alpaka/mem/view/ViewAccessOps.hpp
```diff
@@ -37,7 +37,7 @@ namespace alpaka::internal
     public:
         ViewAccessOps()
         {
-            static_assert(experimental::trait::internal::IsView<TView>::value);
```
psychocoderHPC: should be reverted, SYCL independent.

include/alpaka/mem/view/ViewAccessor.hpp
```diff
@@ -20,7 +20,7 @@
 
 #include <type_traits>
 
-namespace alpaka::experimental
```
psychocoderHPC: SYCL independent

psychocoderHPC (dismissed):

include/alpaka/mem/fence/MemFenceGenericSycl.hpp
```diff
@@ -54,17 +54,17 @@ namespace alpaka::experimental
...
-    struct MemFence<experimental::MemFenceGenericSycl, TMemScope>
+    struct MemFence<MemFenceGenericSycl, TMemScope>
     {
-        static auto mem_fence(experimental::MemFenceGenericSycl const& fence, TMemScope const&)
+        static auto mem_fence(MemFenceGenericSycl const& fence, TMemScope const&)
         {
-            static constexpr auto scope = experimental::detail::SyclFenceProps<TMemScope>::scope;
-            static constexpr auto space = experimental::detail::SyclFenceProps<TMemScope>::space;
+            static constexpr auto scope = detail::SyclFenceProps<TMemScope>::scope;
+            static constexpr auto space = detail::SyclFenceProps<TMemScope>::space;
```
j-stephan, line 62: Since you are touching this region anyway, could you remove the `ext::oneapi` namespace here, too?

j-stephan (changes requested):
This looks mostly good. @psychocoderHPC has already addressed most of the issues that need to be fixed. Also, please update the copyright header in all files you touched here. Additionally:

include/alpaka/mem/fence/MemFenceGenericSycl.hpp
Parsifal-2045: The last one is an easy fix. However, all the other changes are more involved as SYCL Accessor relies on the three other mentioned files. If I revert changes on those, most of the SYCL accessor will have to go back in the experimental namespace.

include/alpaka/mem/view/Accessor.hpp
Parsifal-2045: As far as I can tell, this Accessor file is included only in SYCL (buffer and accessor). I have not found a way to leave these files inside the experimental namespace without breaking SYCL

bernhardmgruber: Yes, I built the accessors a long time ago to prepare for the SYCL backend. But when we move to pointer based SYCL, they are no longer needed. Except for the use cases of @psychocoderHPC who wants to configure how memory access works, but this can also be handled by `std::mdspan`. I want to drop accessors in a separate PR, see also #1897.

bernhardmgruber:
I rebased the PR and fixed @psychocoderHPC's suggestions.

bernhardmgruber:
All examples compile with `dpcpp 2023.0.0`, the tests fail to compile with lots of errors unrelated to this PR.

bernhardmgruber:
@Parsifal-2045 do you want to make more changes to this PR or can we merge it? (The PR is in draft mode)

bernhardmgruber (dismissed):

Parsifal-2045:
> @Parsifal-2045 do you want to make more changes to this PR or can we merge it? (The PR is in draft mode)

I'll just modify the headers and remove the ext::oneapi as suggested, in the next few hours. I'll update the PR once it's ready to be merged

bernhardmgruber (approved):

j-stephan (approved):
Thanks!

j-stephan:
GitHub has some server issues at the moment. We'll hit the merge button later once they are resolved.

PR #1912, opened 2023-02-06
Disable tests if used by `add_subdirectory`

### LINKED ISSUE #1888: BUILD_TESTING enables tests when alpaka is built as an external
While starting to integrate Alpaka into [traccc](https://github.com/acts-project/traccc/) I found that the test suite was enabled when `traccc` sets BUILD_TESTING on.

As I understand it there are two options:
1)Change `BUILD_TESTING` to e.g. `alpaka_BUILD_TESTING`

2)Something like
```
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME)
    include(CTest)
endif()
```
to ensure that the tests are not built.

A workaround is also possible, as is done in traccc for Eigen:
https://github.com/acts-project/traccc/blob/main/extern/eigen3/CMakeLists.txt#L26
but this is a bit ugly.

My immediate preference is for 1) - it's just find and replace in a few files:

```
CMakeLists.txt
README_OACC.md
README_OMP5.md
README_SYCL.md
cmake/alpakaCommon.cmake
docs/source/advanced/cmake.rst
docs/source/basic/install.rst
docs/source/dev/backends.rst
docs/source/dev/test.rst
script/run_generate.sh
thirdParty/CMakeLists.txt
```
but happy to go with whatever is preferred by maintainers.

j-stephan:
Fixes #1888.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 #
-# Copyright 2022 Benjamin Worpitz, Jan Stephan, Bernhard Manfred Gruber
+# Copyright 2023 Benjamin Worpitz, Jan Stephan, Bernhard Manfred Gruber
 #
 # This file is part of alpaka.
 #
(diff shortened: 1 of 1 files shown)


bernhardmgruber:
I don't see a need for this, since it can easily be worked around by using `add_subdirectory(${alpaka_SOURCE_DIR} EXCLUDE_FROM_ALL)` instead of `add_subdirectory(${alpaka_SOURCE_DIR})`. But I also don't oppose the change. If you use alpaka via `add_subdirectory` you are likely consuming it as dependency and don't want to run the tests. So proceed as you please :)

j-stephan:
> I don't see a need for this, since it can easily be worked around

Maybe, but why force our users to work around things? And the [Modern CMake](https://cliutils.gitlab.io/modern-cmake/chapters/testing.html) recommends the change done in this PR.

bernhardmgruber:
> > I don't see a need for this, since it can easily be worked around
> 
> Maybe, but why force our users to work around things? And the [Modern CMake](https://cliutils.gitlab.io/modern-cmake/chapters/testing.html) recommends the change done in this PR.

Thx for the pointer! I have skimmed through the book a while back, but apparently missed that part.

bernhardmgruber (approved):

PR #1913, opened 2023-02-08
Raise required CMake version to 3.22

### LINKED ISSUE #1509: Raise required CMake version to 3.21
Upgrading to CMake 3.21 would offer us the following advantages:

* Support for `CUDA_SEPARABLE_COMPILATION` for clang-cuda (CMake 3.19) (important for #1467)
* Platform support for Apple Silicon (CMake 3.19)
* A more flexible `FindCUDAToolkit` module (CMake 3.19)
* Support for C++23 mode (CMake 3.20)
* `nvc++` compiler support (CMake 3.20)
* Support for the oneAPI compiler toolchain (CMake 3.20) (nice to have for #789)
* Support for Visual Studio 2022 generators (CMake 3.21)
* HIP as first-class language (CMake 3.21)
* Support for the Fujitsu compiler toolchain (CMake 3.21)

### LINKED ISSUE #1905: Clang as CUDA compiler requires CMake 3.19
CMake 3.18 has initial support for Clang as CUDA compiler but a bug causes an error, if `CMAKE_CUDA_ARCHITECTURES` is set. The bug is solved with CMake 3.19.

https://gitlab.com/hzdr/crp/alpaka/-/pipelines/749183469

I think we should handle the problem in the same way, like issue #1892

j-stephan:
Fixes #1509. Fixes #1905.


bernhardmgruber:
I find this version bump too aggressive, but I am generally in favor of using newer tools. Let's hear what others have to say.

j-stephan:
Which version would be acceptable? I just went with what Ubuntu 22.04 provides by default.

psychocoderHPC (approved):

psychocoderHPC:
I am fine with the  update even if I would like to see the dependency update more conservative but 3.22 is fixing much stuff and bringing better ROCM support, so I am for it!

PR #1914, opened 2023-02-09
Fix compilation warnings

fwyzard:
Make sure `pragma clang diagnostic` is only used when compiling with clang, and not with gcc.

With HIP version 5.2.0 or newer, include `hiprand/hiprand_kernel.h` instead of `hiprand_kernel.h`.

### DIFF
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -11,9 +11,9 @@
 
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
 
-#    include <alpaka/core/BoostPredef.hpp>
-
+#    include <boost/predef.h>
 #    include <hip/hip_runtime_api.h>
+#    include <hip/hip_version.h>
 
 namespace alpaka
 {
@@ -21,7 +21,7 @@ namespace alpaka
     {
         // Names
         static constexpr char name[] = "Hip";
-        static constexpr auto version = BOOST_LANG_HIP;
+        static constexpr auto version = BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0);
 
         // Types
         using DeviceAttr_t = ::hipDeviceAttribute_t;
@@ -83,7 +83,7 @@ namespace alpaka
         static constexpr DeviceAttr_t deviceAttributeMaxThreadsPerBlock = ::hipDeviceAttributeMaxThreadsPerBlock;
         static constexpr DeviceAttr_t deviceAttributeMultiprocessorCount = ::hipDeviceAttributeMultiprocessorCount;
 
-#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(4, 5, 0)
+#    if HIP_VERSION >= 40500000
         static constexpr Limit_t limitPrintfFifoSize = ::hipLimitPrintfFifoSize;
 #    else
         static constexpr Limit_t limitPrintfFifoSize
@@ -122,7 +122,7 @@ namespace alpaka
 
         static inline Error_t deviceGetLimit(size_t* pValue, Limit_t limit)
         {
-#    if BOOST_LANG_HIP < BOOST_VERSION_NUMBER(4, 5, 0)
+#    if HIP_VERSION < 40500000
             if(limit == limitPrintfFifoSize)
             {
                 // Implemented only in ROCm 4.5.0 and later.
@@ -186,7 +186,7 @@ namespace alpaka
         static inline Error_t freeAsync([[maybe_unused]] void* devPtr, [[maybe_unused]] Stream_t stream)
         {
             // hipFreeAsync is implemented only in ROCm 5.2.0 and later.
-#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(5, 2, 0)
+#    if HIP_VERSION >= 50200000
             return ::hipFreeAsync(devPtr, stream);
 #    else
             // Not implemented.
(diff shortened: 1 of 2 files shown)


j-stephan:
Hi @fwyzard, thanks for the submission. Please note that the AMD CI will likely fail at the moment - we had a power surge which unfortunately grilled at least one of our AMD GPUs.

fwyzard:
Ops, sorry to hear that :-(

fwyzard:
The last change is similar to what I did for CUDA some time ago, and is needed to get a consistent behaviour between `hipcc` (compiling device or host code) and `gcc` (compiling host-only code that calls the HIP API, like allocating GPU memory).

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -11,17 +11,17 @@
...
+#    include <hip/hip_version.h>
 
 namespace alpaka
 {
     struct ApiHipRt
     {
         // Names
         static constexpr char name[] = "Hip";
-        static constexpr auto version = BOOST_LANG_HIP;
+        static constexpr auto version = BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0);
```
psychocoderHPC, line 24: Why do you switch from BOOST_VERSION to the version format ROCM is shipping? Is this necessary?

I am asking because normally we use always BOOST_VERSIONS in alpaka and now we need to take care to use for HIP the HIP internal format.

fwyzard: hi @psychocoderHPC, I'm not sure I follow, are you asking why I used the BOOST_VERSION macro explicitly, or why I didn't include the third field in it ?

psychocoderHPC: I misread this line I thought you do not use `BOOST_VERSION` anymore but you use it. I was confused by the usage later, I will mark the line.

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -83,7 +83,7 @@ namespace alpaka
         static constexpr DeviceAttr_t deviceAttributeMaxThreadsPerBlock = ::hipDeviceAttributeMaxThreadsPerBlock;
         static constexpr DeviceAttr_t deviceAttributeMultiprocessorCount = ::hipDeviceAttributeMultiprocessorCount;
 
-#    if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(4, 5, 0)
+#    if HIP_VERSION >= 40500000
```
psychocoderHPC, line 86: Regarding to my last point, I am confused why you use `HIP_VERSION` and not `BOOST_LANG_VERSION` as before.
What was wrong with the original macro comparison?

fwyzard: The problem is that `BOOST_LANG_HIP` is defined only when using `hipcc` to compile host or device code, but it does not get defined when using a different compiler (_e.g._ `gcc`) to compile HIP host code.

The same can be done using the Boost version macros:
```suggestion
#    if BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0) >= BOOST_VERSION_NUMBER(4, 5, 0)
```

I can change it to this approach if it is preferred (and then I guess I should make the same change in `include/alpaka/core/ApiCudaRt.hpp` with respect to `CUDART_VERSION`).

fwyzard: Using `HIP_VERSION` is more compact :-)

psychocoderHPC: IMO we should keep `if BOOST_LANG_HIP >= BOOST_VERSION_NUMBER(4, 5, 0)` to have a single place in the code where the version number stored, in our case it is `#include <alpaka/core/BoostPredef.hpp>`

psychocoderHPC: We discussed this topic in the dev meeting and we should unify the creation of the version number in `BoostPredef.hpp` so that we do not have `BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0)` somewhere in the code expect in the predef header.

`BOOST_LANG_HIP` should be set always if `HIP_VERSION_MAJOR` is available.

fwyzard: This is the only solution that _does not work_ .

`BOOST_LANG_HIP` is defined only when compiling with `hipcc`, not when compiling with `gcc`, so it is not a suitable macro to use here.

`BOOST_LANG_HIP`'s behaviour cannot be changed, as it is used elsewhere specifically to check for a compiler with HIP support.

The viable options are
  - use `HIP_VERSION >= 40500000`
  - use `BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0) >= BOOST_VERSION_NUMBER(4, 5, 0)`
  - `#define BOOST_HIP_VERSION = BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0)` and check `BOOST_HIP_VERSION >= BOOST_VERSION_NUMBER(4, 5, 0)`

fwyzard: And, whatever the choice, it should be applied consistently here and in `include/alpaka/core/ApiCudaRt.hpp` (with respect to `BOOST_LANG_CUDA` vs `CUDART_VERSION`).

fwyzard: > `BOOST_LANG_HIP` should be set always if `HIP_VERSION_MAJOR` is available.

Maybe - but that is a _huge_ change in how Alpaka headers are written, and a departure from how Boost's own `BOOST_LANG_CUDA` behaves: `BOOST_LANG_CUDA` is defined only when compiling a `.cu` file with `nvcc` (or `clang` in cuda mode).

psychocoderHPC: OK, in that case, keep it how it is for now and we will postpone it to a future PR.

psychocoderHPC (approved):

PR #1915, opened 2023-02-10
Add ROCm 5.4 support

### LINKED ISSUE #1848: Support ROCm 5.4
https://docs.amd.com/bundle/ROCm-Release-Notes-v5.4/page/About_This_Document.html

j-stephan:
Fixes #1848. Also ensures that we use the (current) latest point releases of the AMD docker images.


script/job_generator/versions.py
```diff
@@ -32,7 +32,7 @@
         "11.8",
         "12.0",
     ],
-    HIPCC: ["5.0", "5.1", "5.2", "5.3"],
+    HIPCC: ["5.0.1", "5.1.3", "5.2.3", "5.3.3", "5.4.2"],
```
fwyzard: I think 5.4.3 was also released this week.

j-stephan: Yes, but there are no docker images (yet) :-(

https://hub.docker.com/r/rocm/dev-ubuntu-20.04/tags?page=1&name=5.4

psychocoderHPC:
@SimeonEhrig We use our [pre-compiled HIP container](https://codebase.helmholtz.cloud/crp/alpaka-group-container/container_registry) instead the AMD provided container as base correct?

SimeonEhrig:
> @SimeonEhrig We use our [pre-compiled HIP container](https://codebase.helmholtz.cloud/crp/alpaka-group-container/container_registry) instead the AMD provided container as base correct?

Yes, we are not depend on the rocm container by AMD. Our container uses a ubuntu base container and install rocm in it.

https://codebase.helmholtz.cloud/crp/alpaka-group-container/-/pipelines/191820 (enable job dependencies to see the image build order)

script/job_generator/versions.py
SimeonEhrig: HIP patch levels are not supported. I only care about the major and minor level. So, there are only docker container tags available in the shape of `alpaka-ci-ubuntu20.04-rocm5.3:<container_version>`.

If we would support different patch level for rocm, we need to store different versions of the container, like:

* alpaka-ci-ubuntu20.04-rocm5.3.0:3.0
* alpaka-ci-ubuntu20.04-rocm5.3.1:3.0
* alpaka-ci-ubuntu20.04-rocm5.3.2:3.0
* alpaka-ci-ubuntu20.04-rocm5.3.3:3.0

Or we ignore the caching mechanism and install hip in each test job like before.

I would avoid handling the patch levels until it causes problems. At the moment, it would be only additional work without benefit.

j-stephan: I don't think we should care about older patch levels. Can we create docker images that automatically pull in the latest patch release?

SimeonEhrig:
ROCm 5.4 is not available as pre build container for the moment. Therefore we have two options.

1. Add 5.4 to the alpaka-group-container repo.
2. Add an exception like for CUDA 11.7+ [1]. In this case, the job generator takes the normal ubuntu base image and install hip 5.4 in each test job.

[1] https://github.com/alpaka-group/alpaka/blob/162773e2a70d9a814d21fdc7f63bdf7a9d823f2b/script/job_generator/generate_job_yaml.py#L97-L104

script/job_generator/versions.py
SimeonEhrig: The automatically pull the latest patch level, when they are built. Means, it depends a little bit, when the last push on the master branch happened.

j-stephan:
I strongly prefer option 1.

SimeonEhrig:
ROCm 5.4 is still using Clang 15. You need to add 

```python
elif job[DEVICE_COMPILER][VERSION] == "5.4":
       variables["ALPAKA_CI_CLANG_VER"] = "15"
```

here: https://github.com/alpaka-group/alpaka/blob/759d7545770e0a4d0f3cdf6778acd90dd705a267/script/job_generator/generate_job_yaml.py#L190-L202

SimeonEhrig:
You need to wait for container version 3.1. Otherwise the rocm 5.4 job will fail, because it cannot find the required image.

j-stephan:
I see. I'll add the required changes once ROCm 5.4 is available.

SimeonEhrig (approved):

PR #1917, opened 2023-02-13
Use SPDX license identifiers

### LINKED ISSUE #1679: [RFC] Use SPDX license identifiers
All alpaka headers use a license statement at the top. This adds a lot of noise and there is a more compact way to specify the license: SPDX license identifiers. More info here: https://spdx.dev/ids/

We basically replace the existing license preamble in header files by:
`// SPDX-License-Identifier: MPL-2.0`

Are there any concerns to such a change?

j-stephan:
Fixes #1679. Replaces all header license comments with the appropriate SPDX license identifiers. In most places, this is `MPL-2.0`. In the `examples` folder this is `ISC`.

### DIFF
--- a/cmake/addExecutable.cmake
+++ b/cmake/addExecutable.cmake
@@ -1,11 +1,6 @@
 #
 # Copyright 2014-2019 Benjamin Worpitz
-#
-# This file is part of alpaka.
-#
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
 #
 
 CMAKE_MINIMUM_REQUIRED(VERSION 3.18)
--- a/example/babelstream/CMakeLists.txt
+++ b/example/babelstream/CMakeLists.txt
@@ -1,19 +1,6 @@
 #
 # Copyright 2023 Erik Zenker, Benjamin Worpitz, Jan Stephan, Bernhard Manfred Gruber
-#
-# This file exemplifies usage of alpaka.
-#
-# Permission to use, copy, modify, and/or distribute this software for any
-# purpose with or without fee is hereby granted, provided that the above
-# copyright notice and this permission notice appear in all copies.
-#
-# THE SOFTWARE IS PROVIDED “AS IS” AND ISC DISCLAIMS ALL WARRANTIES WITH
-# REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
-# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
-# SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
-# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
-# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
-# IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+# SPDX-License-Identifier: ISC
 #
 
 cmake_minimum_required(VERSION 3.22)
--- a/example/randomCells2D/src/randomCells2D.cpp
+++ b/example/randomCells2D/src/randomCells2D.cpp
@@ -1,10 +1,5 @@
 /* Copyright 2021 Jiri Vyskocil
- *
- * This file is part of alpaka.
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: ISC
  */
 
 #include <alpaka/alpaka.hpp>
--- a/example/randomStrategies/src/randomStrategies.cpp
+++ b/example/randomStrategies/src/randomStrategies.cpp
@@ -1,10 +1,5 @@
 /* Copyright 2022 Jiří Vyskočil, René Widera, Jan Stephan
- *
- * This file is part of alpaka.
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: ISC
  */
 
 #include <alpaka/alpaka.hpp>
--- a/include/alpaka/rand/TinyMT/tinymt32.h
+++ b/include/alpaka/rand/TinyMT/tinymt32.h
@@ -1,10 +1,5 @@
-/* Copyright 2020 Axel Huebl, Benjamin Worpitz, Mutsuo Saito, Bernhard Manfred Gruber
- *
- * This file is part of alpaka.
- *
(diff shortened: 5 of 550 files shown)


example/LICENSE
```diff
@@ -0,0 +1,13 @@
+Copyright 2014 - 2023 Helmholtz-Zentrum Dresden - Rossendorf e.V. (HZDR)
```
j-stephan: Since alpaka started out as a HZDR research project I believe the formal copyright holder is HZDR. Please correct me if I'm wrong @alpaka-group/alpaka-maintainers.

cmake/addExecutable.cmake
```diff
@@ -1,14 +1,9 @@
...
-# This file is part of alpaka.
-#
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+# SPDX-License-Identifier: MPL-2.0
 #
 
-CMAKE_MINIMUM_REQUIRED(VERSION 3.18)
+CMAKE_MINIMUM_REQUIRED(VERSION 3.22)
```
bernhardmgruber: I think you mixed some unrelated changes into this PR.

example/LICENSE
bernhardmgruber: I don't think this is right. This makes it look like the entire alpaka examples are copyrighted by HZDR. That is at least wrong for all LOCs I have written, which legally belong to CERN.

Why do you even need to put a LICENSE file here?

example/babelstream/CMakeLists.txt
```diff
@@ -1,19 +1,6 @@
...
-# copyright notice and this permission notice appear in all copies.
-#
-# THE SOFTWARE IS PROVIDED “AS IS” AND ISC DISCLAIMS ALL WARRANTIES WITH
-# REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
-# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL ISC BE LIABLE FOR ANY
-# SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
-# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
-# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF OR
-# IN CONNECTION WITH THE USE OR PERFORMANCE OF THIS SOFTWARE.
+# SPDX-License-Identifier: ISC
```
bernhardmgruber, line 3: Have all the examples previously already had the ISC license? That would be news to me, I thought everything was MPL-2. In order words, are you relicensing all the examples?

example/LICENSE
j-stephan: We should discuss the copyright issue in tomorrow's VC.

> Why do you even need to put a LICENSE file here?

We need to put the ISC license for the examples somewhere if we delete the header comments. Feel free to suggest a better location.

example/babelstream/CMakeLists.txt
j-stephan: The changes are shown right there ;-) As far as I can tell all examples except BabelStream have always been licensed under ISC. I didn't relicense anything (on purpose).

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -1,10 +1,5 @@
 /* Copyright 2021 Jiri Vyskocil
- *
- * This file is part of alpaka.
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: ISC
```
j-stephan, line 2: Here we have a case of accidental relicensing due to license mixing. The CMakeLists.txt is ISC-licensed but the source code is MPL. I attribute this to a copy-paste error but maybe @sliwowitz did this on purpose?

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -1,10 +1,5 @@
 /* Copyright 2022 Jiří Vyskočil, René Widera, Jan Stephan
- *
- * This file is part of alpaka.
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: ISC
```
j-stephan, line 2: And here, too.

example/LICENSE
bernhardmgruber: > We need to put the ISC license for the examples somewhere if we delete the header comments. Feel free to suggest a better location.

Do we need to do that? The ISC license doesn't state that a copy of the license must be included with the source code. We could just have the ISC SPDX license identifier in the example files and be good I think.

example/randomCells2D/src/randomCells2D.cpp
sliwowitz: Yes, it was a mistake. I did not realize we use different licenses for the examples.

sliwowitz: We can change that to ISC or whatever license we decide to use for the examples.

example/randomStrategies/src/randomStrategies.cpp
sliwowitz: Same as the the other file.

fwyzard:
Is there a good reason for the ISC license ?

According to [Wikipedia](https://en.wikipedia.org/wiki/ISC_license#Reception), ISC itself stopped using it, and moved to MPL 2.0 .

j-stephan:
This was decided upon in 2018 when alpaka was relicensed from LGPLv3 to the current MPLv2/ISC mix. You can follow the chain of reasoning [here](https://github.com/alpaka-group/alpaka/issues/637) (I only discovered this yesterday while looking into the same question). Apparently ISC is "more modern" than BSD/MIT.

fwyzard:
> Apparently ISC is "more modern" than BSD/MIT.

I don't contest that (though the differences seem minor at best).
But why not use MPL 2.0 everywhere, instead of a mix of MPL 2.0/ISC ?

Anyway, that's orthogonal to this PR... I am simply curious, and didn't know ISC was used for (some of ?) the examples.

j-stephan:
Our predecessors argued that the examples should be copy&paste-able into any code (including closed source), hence ISC for the examples. But I'm no license expert, it is entirely possible that MPL would also allow for this.

include/alpaka/rand/TinyMT/tinymt32.h
```diff
@@ -1,10 +1,5 @@
 /* Copyright 2020 Axel Huebl, Benjamin Worpitz, Mutsuo Saito, Bernhard Manfred Gruber
- *
- * This file is part of alpaka.
- *
- * This Source Code Form is subject to the terms of the Mozilla Public
- * License, v. 2.0. If a copy of the MPL was not distributed with this
- * file, You can obtain one at http://mozilla.org/MPL/2.0/.
+ * SPDX-License-Identifier: MPL-2.0
```
bernhardmgruber:
```suggestion
 * SPDX-License-Identifier: BSD-3-Clause
```
I would generally replace the author list above by the the original authors of the file's main content.

bernhardmgruber (changes requested):

include/alpaka/rand/TinyMT/tinymt32.h
j-stephan: Good catch. This was added automatically a few years back and is likely an oversight.

bernhardmgruber (approved):

PR #1924, opened 2023-02-24
Implement ShflUp, ShflDown and ShflXor 

AuroraPerego:
Following the logic of `alpaka::shfl`, I've implemented also the other three methods: `alpaka::shfl_up`, `alpaka::shfl_down`, `alpaka::shfl_xor`.
I've also added the tests for those methods.

I have two comments on that:
- is there a reason for not having templated the `alpaka::shfl` method and having defined two different methods, one for `std::int_32t` and the other one for `float`? (btw I think that also `unsigned`, `long` and `double` should be added if we don't want to template the methods) 
- SYCL doesn't support a sub group (aka warp) size different from the width in those methods. For the moment I added an assert to check that, but maybe we should think of a way to determine the `scrLane` based on the parameters given in input to `alpaka::shfl_up`, `alpaka::shfl_down`, `alpaka::shfl_xor` and then use `alpaka::shfl` when `width != sub_group_size`.

Last thing, when running the tests I noticed that, when selecting the CUDA backend, **sometimes** the kernel is not executed (I put an `assert(false)` in a kernel and all the tests passed without crashing). I don't know why, maybe that could be discussed in a separated issue.

### DIFF
--- a/include/alpaka/warp/Traits.hpp
+++ b/include/alpaka/warp/Traits.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Sergei Bastrakov, David M. Rogers, Bernhard Manfred Gruber
+/* Copyright 2022 Sergei Bastrakov, David M. Rogers, Bernhard Manfred Gruber, Aurora Perego
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -38,6 +38,18 @@ namespace alpaka::warp
         //! The shfl warp swizzling trait.
         template<typename TWarp, typename TSfinae = void>
         struct Shfl;
+
+        //! The shfl up warp swizzling trait.
+        template<typename TWarp, typename TSfinae = void>
+        struct ShflUp;
+
+        //! The shfl down warp swizzling trait.
+        template<typename TWarp, typename TSfinae = void>
+        struct ShflDown;
+
+        //! The shfl xor warp swizzling trait.
+        template<typename TWarp, typename TSfinae = void>
+        struct ShflXor;
 
         //! The active mask trait.
         template<typename TWarp, typename TSfinae = void>
@@ -162,7 +174,7 @@ namespace alpaka::warp
     //!     __shared__ int32_t values[warpsize];
     //!     values[threadIdx.x] = value;
     //!     __syncthreads();
-    //!     return values[(srcLane + width*floor(threadIdx.x/width))%width];
+    //!     return values[width*(threadIdx.x/width) + srcLane%width];
     //!
     //! However, it does not use shared memory.
     //!
@@ -182,19 +194,124 @@ namespace alpaka::warp
     //! \param  width   number of threads receiving a single value
     //! \return val from the thread index srcLane.
     ALPAKA_NO_HOST_ACC_WARNING
-    template<typename TWarp>
-    ALPAKA_FN_ACC auto shfl(TWarp const& warp, std::int32_t value, std::int32_t srcLane, std::int32_t width = 0)
+    template<typename TWarp, typename T>
+    ALPAKA_FN_ACC auto shfl(TWarp const& warp, T value, std::int32_t srcLane, std::int32_t width = 0)
     {
         using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
         return trait::Shfl<ImplementationBase>::shfl(warp, value, srcLane, width ? width : getSize(warp));
     }
 
-    //! shfl for float vals
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane with lower ID relative to caller.
+    //! The lane ID is calculated by subtracting delta from the caller’s lane ID.
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     return (threadIdx.x % width >= delta) ? values[threadIdx.x - delta] : values[threadIdx.x];
+    //!
+    //! However, it does not use shared memory.
+    //!
+    //! Notes:
+    //! * The programmer must ensure that all threads calling this
+    //!   function (and the srcLane) are executing the same line of code.
+    //!   In particular it is not portable to write if(a) {shfl} else {shfl}.
+    //!
+    //! * Commonly used with width = warpsize (the default), (returns values[threadIdx.x - delta] if threadIdx.x >=
+    //! delta)
+    //!
+    //! * Width must be a power of 2.
+    //!
+    //! \tparam TWarp   warp implementation type
+    //! \tparam T       value type
+    //! \param  warp    warp implementation
+    //! \param  value   value to broadcast
+    //! \param  offset  corresponds to the delta used to compute the lane ID
+    //! \param  width   size of the group participating in the shuffle operation
+    //! \return val from the thread index lane ID.
     ALPAKA_NO_HOST_ACC_WARNING
-    template<typename TWarp>
-    ALPAKA_FN_ACC auto shfl(TWarp const& warp, float value, std::int32_t srcLane, std::int32_t width = 0)
+    template<typename TWarp, typename T>
+    ALPAKA_FN_ACC auto shfl_up(TWarp const& warp, T value, std::uint32_t offset, std::int32_t width = 0)
     {
         using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
-        return trait::Shfl<ImplementationBase>::shfl(warp, value, srcLane, width ? width : getSize(warp));
+        return trait::ShflUp<ImplementationBase>::shfl_up(warp, value, offset, width ? width : getSize(warp));
+    }
+
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane with higher ID relative to caller.
+    //! The lane ID is calculated by adding delta to the caller’s lane ID.
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     return (threadIdx.x % width + delta < width) ? values[threadIdx.x + delta] : values[threadIdx.x];
+    //!
+    //! However, it does not use shared memory.
+    //!
+    //! Notes:
+    //! * The programmer must ensure that all threads calling this
+    //!   function (and the srcLane) are executing the same line of code.
+    //!   In particular it is not portable to write if(a) {shfl} else {shfl}.
+    //!
+    //! * Commonly used with width = warpsize (the default), (returns values[threadIdx.x+delta] if threadIdx.x+delta <
+    //! warpsize)
+    //!
+    //! * Width must be a power of 2.
+    //!
+    //! \tparam TWarp   warp implementation type
+    //! \tparam T       value type
+    //! \param  warp    warp implementation
+    //! \param  value   value to broadcast
+    //! \param  offset  corresponds to the delta used to compute the lane ID
+    //! \param  width   size of the group participating in the shuffle operation
+    //! \return val from the thread index lane ID.
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TWarp, typename T>
+    ALPAKA_FN_ACC auto shfl_down(TWarp const& warp, T value, std::uint32_t offset, std::int32_t width = 0)
+    {
+        using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
+        return trait::ShflDown<ImplementationBase>::shfl_down(warp, value, offset, width ? width : getSize(warp));
+    }
+
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane based on bitwise XOR of own lane ID.
+    //! The lane ID is calculated by performing a bitwise XOR of the caller’s lane ID with mask
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     int lane = threadIdx.x ^ mask;
+    //!     return values[lane / width > threadIdx.x / width ? threadIdx.x : lane];
+    //!
+    //! However, it does not use shared memory.
+    //!
+    //! Notes:
+    //! * The programmer must ensure that all threads calling this
+    //!   function (and the srcLane) are executing the same line of code.
+    //!   In particular it is not portable to write if(a) {shfl} else {shfl}.
+    //!
+    //! * Commonly used with width = warpsize (the default), (returns values[threadIdx.x^mask])
+    //!
+    //! * Width must be a power of 2.
+    //!
+    //! \tparam TWarp   warp implementation type
+    //! \tparam T       value type
+    //! \param  warp    warp implementation
+    //! \param  value   value to broadcast
+    //! \param  mask    corresponds to the mask used to compute the lane ID
+    //! \param  width   size of the group participating in the shuffle operation
+    //! \return val from the thread index lane ID.
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TWarp, typename T>
+    ALPAKA_FN_ACC auto shfl_xor(TWarp const& warp, T value, std::int32_t mask, std::int32_t width = 0)
+    {
+        using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
+        return trait::ShflXor<ImplementationBase>::shfl_xor(warp, value, mask, width ? width : getSize(warp));
     }
 } // namespace alpaka::warp
--- a/include/alpaka/warp/WarpGenericSycl.hpp
+++ b/include/alpaka/warp/WarpGenericSycl.hpp
@@ -1,5 +1,11 @@
 /* Copyright 2023 Jan Stephan, Luca Ferragina, Andrea Bocci, Aurora Perego
  * SPDX-License-Identifier: MPL-2.0
+ *
+ * The implementations of Shfl::shfl(), ShflUp::shfl_up(), ShflDown::shfl_down() and ShflXor::shfl_xor() are derived
+ * from Intel DPCT.
+ * Copyright (C) Intel Corporation.
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ * See https://llvm.org/LICENSE.txt for license information.
  */
 
 #pragma once
@@ -123,12 +129,9 @@ namespace alpaka::warp::trait
                The first starts at sub-group index 0 and the second at sub-group index 16. For srcLane = 4 the
                first subdivision will access the value at sub-group index 4 and the second at sub-group index 20. */
             auto const actual_group = warp.m_item_warp.get_sub_group();
-            auto const actual_item_id = static_cast<std::int32_t>(actual_group.get_local_linear_id());
-            auto const actual_group_id = actual_item_id / width;
-            auto const actual_src_id = static_cast<std::size_t>(srcLane + actual_group_id * width);
-            auto const src = sycl::id<1>{actual_src_id};
-
-            return sycl::select_from_group(actual_group, value, src);
+            std::uint32_t const w = static_cast<std::uint32_t>(width);
+            std::uint32_t const start_index = actual_group.get_local_linear_id() / w * w;
+            return sycl::select_from_group(actual_group, value, start_index + static_cast<std::uint32_t>(srcLane) % w);
         }
     };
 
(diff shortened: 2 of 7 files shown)


psychocoderHPC:
> is there a reason for not having templated the alpaka::shfl method and having defined two different methods, 

I assume older CUDA supported shuffle only for these types and the current version is supporting any type

psychocoderHPC:
> SYCL doesn't support a sub group (aka warp) size different from the width in those methods. For the moment I added an assert to check that, but maybe we should think of a way to determine the scrLane based on the parameters given in input ...

IMO As you suggested emulating the shufl_down, ... for all sub groups different than `the warp` size based in the lane is the best way.

AuroraPerego:
> I assume older CUDA supported shuffle only for these types and the current version is supporting any type

Ok, is it ok if I replace the two existing methods with the template?

AuroraPerego:
> IMO As you suggested emulating the shufl_down, ... for all sub groups different than `the warp` size based in the lane is the best way.

I'll try to implement it then

psychocoderHPC:
> > I assume older CUDA supported shuffle only for these types and the current version is supporting any type
> 
> Ok, is it ok if I replace the two existing methods with the template?

I would keep it like it is, it looks like AMD is still only shipping float and int signatures for this function.
https://github.com/ROCm-Developer-Tools/HIP/commit/04f3e3e59887a0098224513ed3011d689e1dd272
If AMD is providing a generic function with a template feel free to change it.

fwyzard:
Actually, I see the other types on a local ROCm installation:
```
$ grep -E '(\w+ +)+__shfl\>' -r /opt/rocm/include/
/opt/rocm/include/hip/amd_detail/amd_hip_cooperative_groups.h:    return __shfl(var, lane, WAVEFRONT_SIZE);
/opt/rocm/include/hip/amd_detail/amd_hip_cooperative_groups.h:    return __shfl(var, lane, WAVEFRONT_SIZE);
/opt/rocm/include/hip/amd_detail/amd_hip_cooperative_groups.h:    return __shfl(var, lane, WAVEFRONT_SIZE);
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:int __shfl(int var, int src_lane, int width = warpSize) {
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:unsigned int __shfl(unsigned int var, int src_lane, int width = warpSize) {
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:float __shfl(float var, int src_lane, int width = warpSize) {
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:double __shfl(double var, int src_lane, int width = warpSize) {
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:long __shfl(long var, int src_lane, int width = warpSize)
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:unsigned long __shfl(unsigned long var, int src_lane, int width = warpSize) {
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:long long __shfl(long long var, int src_lane, int width = warpSize)
/opt/rocm/include/hip/amd_detail/amd_warp_functions.h:unsigned long long __shfl(unsigned long long var, int src_lane, int width = warpSize) {
/opt/rocm/include/hip/nvidia_detail/nvidia_hip_runtime_api.h:#define __shfl(...)      __shfl_sync(0xffffffff, __VA_ARGS__)
/opt/rocm/include/rocprim/intrinsics/warp_shuffle.hpp:            return __shfl(v, src_lane, width);
```

See also https://github.com/ROCm-Developer-Tools/hipamd/blob/develop/include/hip/amd_detail/amd_warp_functions.h .

AuroraPerego:
I've rebased it and I think it is ready for review

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,115 @@ namespace alpaka::warp
...
+    template<typename TWarp, typename T>
+    ALPAKA_FN_ACC auto shfl(TWarp const& warp, T value, std::int32_t srcLane, std::int32_t width = 0)
     {
         using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
         return trait::Shfl<ImplementationBase>::shfl(warp, value, srcLane, width ? width : getSize(warp));
     }
 
-    //! shfl for float vals
+    //! Exchange data between threads within a warp.
+    //! It copy from a lane with lower ID relative to caller.
```
fwyzard: either
```suggestion
    //! It copies from a lane with lower ID relative to caller.
```
or
```suggestion
    //! Copy from a lane with lower ID relative to caller.
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,115 @@ namespace alpaka::warp
...
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane with lower ID relative to caller.
+    //! The lane ID is calculated by subtracting delta from the caller’s lane ID.
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     return values[(-delta + width*floor(threadIdx.x/width))%width];
```
fwyzard: This formula seems wrong:
 - `threadIdx.x/width` will be an integer between `0` and `blockDim.x / width`
 - passing it through `floor(...)` is unnecessary, and will be the same value
 - `width * floor(threadIdx.x/width)` will be a multiple of `width`
 - `(-delta + width*floor(threadIdx.x/width))` or `(width*floor(threadIdx.x/width) - delta)` will be a multiple of `width`, minus `delta`
 - then, `(...) % width` is equivalent to `(width - delta) % width`

which is the same value independently of `threadIdx.x`.

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane with lower ID relative to caller.
+    //! The lane ID is calculated by subtracting delta from the caller’s lane ID.
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     return values[width*(threadIdx.x/width) + threadIdx.x%width - delta];
```
fwyzard: `width*(threadIdx.x/width) + threadIdx.x%width - delta` is equal to `threadIdx.x - delta`, so this is simply
```suggestion
    //!     return values[threadIdx.x - delta];
```

What was it supposed to be ?

fwyzard: I think it should be something like
```c++
    //!     return (threadIdx.x % width >= delta) ? values[threadIdx.x - delta] : values[threadIdx.x];
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //! * Commonly used with width = warpsize (the default), (returns values[threadIdx.x - delta] if threadIdx.x >=
+    //! delta)
+    //!
+    //! * Width must be a power of 2.
+    //!
+    //! \tparam TWarp   warp implementation type
+    //! \tparam T       value type
+    //! \param  warp    warp implementation
+    //! \param  value   value to broadcast
+    //! \param  offset  corresponds to the delta used to compute the lane ID
```
fwyzard, line 231: Can you replace `delta` with `offset` in the comments above ?

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    ALPAKA_FN_ACC auto shfl_up(TWarp const& warp, T value, std::uint32_t offset, std::int32_t width = 0)
     {
         using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
-        return trait::Shfl<ImplementationBase>::shfl(warp, value, srcLane, width ? width : getSize(warp));
+        return trait::ShflUp<ImplementationBase>::shfl_up(warp, value, offset, width ? width : getSize(warp));
+    }
+
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane with higher ID relative to caller.
+    //! The lane ID is calculated by adding delta from the caller’s lane ID.
```
fwyzard:
```suggestion
    //! The lane ID is calculated by adding delta to the caller’s lane ID.
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     return values[width*(threadIdx.x/width) + threadIdx.x%width + delta];
```
fwyzard:
```suggestion
    //!     return (threadIdx.x % width + delta < width) ? values[threadIdx.x + delta] : values[threadIdx.x];
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    template<typename TWarp, typename T>
+    ALPAKA_FN_ACC auto shfl_down(TWarp const& warp, T value, std::uint32_t offset, std::int32_t width = 0)
+    {
+        using ImplementationBase = concepts::ImplementationBase<ConceptWarp, TWarp>;
+        return trait::ShflDown<ImplementationBase>::shfl_down(warp, value, offset, width ? width : getSize(warp));
+    }
+
+    //! Exchange data between threads within a warp.
+    //! It copies from a lane based on bitwise XOR of own lane ID.
+    //! The lane ID is calculated by performing a bitwise XOR of the caller’s lane ID with delta
```
fwyzard: `delta` should be renamed to `mask` in the comments

fwyzard: Also, according to the CUDA documentation:

> If `width` is less than `warpSize` then each group of `width` consecutive threads are able to access elements from _earlier_ groups of threads, however if they attempt to access elements from later groups of threads their own value of `var` will be returned. This mode implements a butterfly addressing pattern such as is used in tree reduction and broadcast.

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //!
+    //! Effectively executes:
+    //!
+    //!     __shared__ int32_t values[warpsize];
+    //!     values[threadIdx.x] = value;
+    //!     __syncthreads();
+    //!     return values[width*(threadIdx.x/width) + (threadIdx.x%width ^ delta)];
```
fwyzard:
```suggestion
    //!     int lane = threadIdx.x ^ mask;
    //!     return values[lane / width > threadIdx.x / width ? threadIdx.x : lane];
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //!
+    //! However, it does not use shared memory.
+    //!
+    //! Notes:
+    //! * The programmer must ensure that all threads calling this
+    //!   function (and the srcLane) are executing the same line of code.
+    //!   In particular it is not portable to write if(a) {shfl} else {shfl}.
+    //!
+    //! * Commonly used with width = warpsize (the default), (returns values[threadIdx.x^delta])
```
fwyzard:
```suggestion
    //! * Commonly used with width = warpsize (the default), (returns values[threadIdx.x^mask])
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //!
+    //! * Width must be a power of 2.
+    //!
+    //! \tparam TWarp   warp implementation type
+    //! \tparam T       value type
+    //! \param  warp    warp implementation
+    //! \param  value   value to broadcast
+    //! \param  offset  corresponds to the delta used to compute the lane ID
```
fwyzard:
```suggestion
    //! \param  mask    corresponds to the mask used to compute the lane ID
```

include/alpaka/warp/Traits.hpp
```diff
@@ -182,19 +194,123 @@ namespace alpaka::warp
...
+    //! delta)
+    //!
+    //! * Width must be a power of 2.
+    //!
+    //! \tparam TWarp   warp implementation type
+    //! \tparam T       value type
+    //! \param  warp    warp implementation
+    //! \param  value   value to broadcast
+    //! \param  offset  corresponds to the delta used to compute the lane ID
+    //! \param  width   number of threads receiving a single value
```
fwyzard:
```suggestion
    //! \param  width   size of the group participating in the shuffle operation
```
Here and in the next two functions.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -113,34 +113,75 @@ namespace alpaka::warp
...
             __device__ static auto shfl(
                 [[maybe_unused]] warp::WarpUniformCudaHipBuiltIn const& warp,
-                float val,
+                T val,
                 int srcLane,
-                std::int32_t width) -> float
+                std::int32_t width) -> T
             {
 #        if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
                 return __shfl_sync(activemask(warp), val, srcLane, width);
```
fwyzard, line 124: I know that this was already there, but shouldn't this be
```c++
return __shfl_sync(0xffffffff, val, srcLane, width);
```
instead ?

Possible preceded by a `__syncwarp()` ?

fwyzard: This comes from https://github.com/alpaka-group/alpaka/pull/1273.
@psychocoderHPC what do you think ?

fwyzard: For example, HIP does
```c++
#if CUDA_VERSION >= CUDA_9000
#define __shfl(...)      __shfl_sync(0xffffffff, __VA_ARGS__)
#define __shfl_up(...)   __shfl_up_sync(0xffffffff, __VA_ARGS__)
#define __shfl_down(...) __shfl_down_sync(0xffffffff, __VA_ARGS__)
#define __shfl_xor(...)  __shfl_xor_sync(0xffffffff, __VA_ARGS__)
#endif // CUDA_VERSION >= CUDA_9000
```

fwyzard:
@SimeonEhrig the OSX debug builds are failing with
```
Error: You are using macOS 11.
We (and Apple) do not provide support for this old version.
```

Is it expected ?

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -132,6 +131,68 @@ namespace alpaka::warp::trait
...
+        {
+            std::int32_t offset_int = static_cast<std::int32_t>(offset);
+            auto const actual_group = warp.m_item_warp.get_sub_group();
+            auto actual_item_id = static_cast<std::int32_t>(actual_group.get_local_linear_id());
+            auto const actual_group_id = actual_item_id / width;
+            auto const actual_src_id = actual_item_id - offset_int;
+            auto const src = actual_src_id >= actual_group_id * width
+                                 ? sycl::id<1>{static_cast<std::size_t>(actual_src_id)}
+                                 : sycl::id<1>{static_cast<std::size_t>(actual_item_id)};
+            return sycl::select_from_group(actual_group, value, src);
```
fwyzard: @AuroraPerego do you think it would be worth checking if `width` is the same as `get_sub_group().get_max_local_range()`, and in that case call `shift_group_right(actual_group, value, offset)` directly, instead of `select_from_group` ?

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -132,6 +131,68 @@ namespace alpaka::warp::trait
...
+        template<typename T>
+        static auto shfl_up(
+            warp::WarpGenericSycl<TDim> const& warp,
+            T value,
+            std::uint32_t offset, /* must be the same for all work-items in the group */
+            std::int32_t width)
+        {
+            std::int32_t offset_int = static_cast<std::int32_t>(offset);
+            auto const actual_group = warp.m_item_warp.get_sub_group();
+            auto actual_item_id = static_cast<std::int32_t>(actual_group.get_local_linear_id());
```
fwyzard: This one can also be `const`, since all the other ones are.

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -132,6 +131,68 @@ namespace alpaka::warp::trait
...
+
+    template<typename TDim>
+    struct ShflXor<warp::WarpGenericSycl<TDim>>
+    {
+        template<typename T>
+        static auto shfl_xor(
+            warp::WarpGenericSycl<TDim> const& warp,
+            T value,
+            std::int32_t mask,
+            std::int32_t /*width*/)
```
fwyzard: The CUDA version does make use of width:

> `__shfl_xor_sync()` calculates a source line ID by performing a bitwise XOR of the caller’s lane ID with `laneMask`: the value of `var` held by the resulting lane ID is returned. If `width` is less than `warpSize` then each group of `width` consecutive threads are able to access elements from earlier groups of threads, however if they attempt to access elements from later groups of threads their own value of `var` will be returned. This mode implements a butterfly addressing pattern such as is used in tree reduction and broadcast.

Can you implement the same behaviour ?

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -114,22 +114,81 @@ namespace alpaka::warp::trait
...
+            //     each
+            //        subdivision has the assumed index 0. The srcLane index is relative to the subdivisions.
 
-               Example: If we assume a sub-group size of 32 and a width of 16 we will receive two subdivisions:
-               The first starts at sub-group index 0 and the second at sub-group index 16. For srcLane = 4 the
-               first subdivision will access the value at sub-group index 4 and the second at sub-group index 20. */
+            //        Example: If we assume a sub-group size of 32 and a width of 16 we will receive two subdivisions:
+            //        The first starts at sub-group index 0 and the second at sub-group index 16. For srcLane = 4 the
+            //        first subdivision will access the value at sub-group index 4 and the second at sub-group
+            //        index 20. */
```
fwyzard: why the double comment ?

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -114,22 +114,81 @@ namespace alpaka::warp::trait
...
             auto const actual_group = warp.m_item_warp.get_sub_group();
-            auto const actual_item_id = static_cast<std::int32_t>(actual_group.get_local_linear_id());
-            auto const actual_group_id = actual_item_id / width;
-            auto const actual_src_id = static_cast<std::size_t>(srcLane + actual_group_id * width);
-            auto const src = sycl::id<1>{actual_src_id};
+            std::uint32_t const w = static_cast<std::uint32_t>(width);
+            unsigned int const start_index = actual_group.get_local_linear_id() / w * w;
```
fwyzard:
```suggestion
            std::uint32_t const start_index = actual_group.get_local_linear_id() / w * w;
```
for consistency

include/alpaka/warp/WarpGenericSycl.hpp
AuroraPerego: typo

include/alpaka/warp/WarpGenericSycl.hpp
AuroraPerego: I've changed also the others

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -124,12 +123,70 @@ namespace alpaka::warp::trait
...
+    {
+        template<typename T>
+        static auto shfl_up(
+            warp::WarpGenericSycl<TDim> const& warp,
+            T value,
+            std::uint32_t offset, /* must be the same for all work-items in the group */
+            std::int32_t width)
+        {
+            auto const actual_group = warp.m_item_warp.get_sub_group();
+            std::uint32_t const w = static_cast<std::uint32_t>(width);
```
fwyzard, line 149: why don't you simply use `width` directly ?

fwyzard:
I think you should update the license at the top of `include/alpaka/warp/WarpGenericSycl.hpp`:
```diff
 /* Copyright 2023 Jan Stephan, Luca Ferragina, Andrea Bocci, Aurora Perego
  * SPDX-License-Identifier: MPL-2.0
+ *
+ * The implementations of Shfl::shfl(), ShflUp::shfl_up(), ShflDown::shfl_down() and ShflXor::shfl_xor() are derived from Intel DPCT.
+ * Copyright (C) Intel Corporation
+ * SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+ * See https://llvm.org/LICENSE.txt for license information.
  */

include/alpaka/warp/WarpGenericSycl.hpp
AuroraPerego: it gives a warning when doing operations with the thread id that is unsigned:
`conversion to 'uint32_t' {aka 'unsigned int'} from 'int32_t' {aka 'int'} may change the sign of the result [-Wsign-conversion]`

fwyzard (dismissed):

fwyzard (approved):

fwyzard:
@psychocoderHPC can you merge this PR ?
I assume the Xcode failures on OSX are unrelated.

psychocoderHPC:
I removed the XCode test from the require test list and will have a short look to this PR.

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
psychocoderHPC: To be consistent with other backends may be `return __shfl_sync(0xffffffff, val, srcLane, width);` but I am not fully sure.

psychocoderHPC (approved):

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
fwyzard: I'll make a PR for this.

fwyzard: https://github.com/alpaka-group/alpaka/pull/2204

PR #1939, opened 2023-03-23
CI: fix compile issue if cpuonly job is executed on a GPU runner

psychocoderHPC:
In cases where the compile-only job is executed on a GPU runner but with different kinds of accelerators, we need to reset the variables to avoid compiling for the wrong architecture and accelerator.
Note: In general `cpuonly` jobs run on OpenStack instances but in case the OpenStack cluster is down CPU jobs will be executed on the GPU runners too.

I observed and fixed this issue in PIConGPU https://github.com/ComputationalRadiationPhysics/picongpu/pull/4517

### DIFF
--- a/script/install_hip.sh
+++ b/script/install_hip.sh
@@ -41,6 +41,13 @@ sudo update-alternatives --install /usr/bin/c++ c++ ${ROCM_PATH}/llvm/bin/clang+
 export LD_LIBRARY_PATH=${ROCM_PATH}/lib64:${ROCM_PATH}/hiprand/lib:${LD_LIBRARY_PATH}:${ROCM_PATH}/llvm/lib
 export CMAKE_PREFIX_PATH=${ROCM_PATH}:${ROCM_PATH}/hiprand:${CMAKE_PREFIX_PATH:-}
 
+if [[ "$CI_RUNNER_TAGS" =~ .*cpuonly.* ]] ; then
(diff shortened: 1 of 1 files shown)


SimeonEhrig:
@psychocoderHPC Is the default case, if `CI_GPU_ARCH` is not selected still fine or should CMake set the default?

https://github.com/alpaka-group/alpaka/blob/447781c7b89021021679e9392f6a8310102bdc6a/script/install_hip.sh#L61-L65

psychocoderHPC:
> @psychocoderHPC Is the default case, if `CI_GPU_ARCH` is not selected still fine or should CMake set the default?
> 
> https://github.com/alpaka-group/alpaka/blob/447781c7b89021021679e9392f6a8310102bdc6a/script/install_hip.sh#L61-L65

This is still fine. ROCm Cmake has a default if you do not set the architecture but in this case, we will build for several architectures which is make the build extremely slow.

SimeonEhrig:
> > @psychocoderHPC Is the default case, if `CI_GPU_ARCH` is not selected still fine or should CMake set the default?
> > https://github.com/alpaka-group/alpaka/blob/447781c7b89021021679e9392f6a8310102bdc6a/script/install_hip.sh#L61-L65
> 
> This is still fine. ROCm Cmake has a default if you do not set the architecture but in this case, we will build for several architectures which is make the build extremely slow.

That makes sense.

SimeonEhrig (approved):

PR #1941, opened 2023-04-06
[RFC] Remove OpenACC back-end

### LINKED ISSUE #1891: NVHPC 22.9 and 22.11 + OpenACC cause timeouts in test suite
NVHPC 22.9 and 22.11 + OpenACC cause reproducible timeouts in `atomicTest` and `blockSharedTest`: https://gitlab.com/hzdr/crp/alpaka/-/jobs/3521289849

Since 22.3 is working this might be a compiler regression.

j-stephan:
I motion to remove the OpenACC back-end. Reasons:

* Since the beginning of the year we are unable to test recent releases of NVHPC and the underlying bug is not looked at by anyone (see #1891).
* There are no users (that we know of) who rely on the OpenACC back-end.
* It offers no additional hardware support.

All in all, it is my opinion that the benefit of having an OpenACC back-end does not justify the maintenance burden it imposes.

A similar case can be made for the OpenMP 5 back-end, but for now it doesn't cause any problems.

Closes #1891.


bernhardmgruber:
I want to stay neutral in this decision.

fwyzard:
From the CMS point of view, we do not have any use case for OpenACC (or OpenMP 5).

bernhardmgruber (approved):

PR #1956, opened 2023-06-13
Add CI job to create amalgamted alpaka.hpp

### LINKED ISSUE #1954: autogenerate a single header file for the whole alpaka library
Hi,

while playing with Godbolt's Compiler Explorer, I realised that it would be nice to have a version of Alpaka packaged as a single header, that could be included with
```c++
#include <https://raw.githubusercontent.com/alpaka-group/alpaka/develop/include/alpaka.hpp>
```

This does not work because the system cannot resolve remote includes (the other `#include` inside `alpaka.hpp`).

Maybe the CI could be extended with a post-merge hook that
  - merges all Alpaka includes locally, keeping all the preprocessor directives intact
  - generates a single, equivalent, header file
  - commits and push it to a separate repository (or a separate branch of the main repository)

?

bernhardmgruber:
Fixes: #1954


SimeonEhrig:
I'm not sure, if you already aware about it, but there are some problems with the includes: https://godbolt.org/z/TGMn1zzj1

On the side, some includes are not resolve, because they are inside a `#if` clause. On the other side, there are also some includes are not resolved, which I cannot explain: https://github.com/alpaka-group/alpaka/blob/74461909989f0398e6583f0ccd4f58cad9c4751f/include/alpaka/alpaka.hpp#L683

Maybe because we use `#include <...>` instead `#include "..."`.

bernhardmgruber:
> I'm not sure, if you already aware about it, but there are some problems with the includes: https://godbolt.org/z/TGMn1zzj1
> 
> On the side, some includes are not resolve, because they are inside a `#if` clause. On the other side, there are also some includes are not resolved, which I cannot explain: https://github.com/alpaka-group/alpaka/blob/74461909989f0398e6583f0ccd4f58cad9c4751f/include/alpaka/alpaka.hpp#L683
> 
> Maybe because we use `#include <...>` instead `#include "..."`.

I am aware and your last sentence is what I think the problem is. I am looking whether the script can handle it or whether we need to rewrite alpaka includes to local ones. I actually prefer the latter.

SimeonEhrig:
I think, we need also to shift acc guards, so no include is inside a `#if ALPAKA_ACC_<x>_ENABLED` section.

bernhardmgruber:
> I think, we need also to shift acc guards, so no include is inside a `#if ALPAKA_ACC_<x>_ENABLED` section.

Probably related, this is a problem:
```c++
#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
#    include "alpaka/atomic/AtomicAtomicRef.hpp"
#else
#    include "alpaka/atomic/AtomicStdLibLock.hpp"
#endif // ALPAKA_DISABLE_ATOMIC_ATOMICREF
```
Because the amalgamate script will inline all headers inside the first branch and then skip them in the else branch, because they are already included :) But if `ALPAKA_DISABLE_ATOMIC_ATOMICREF` is not defined, and the else branch is taken, all headers are missing.

SimeonEhrig:
> > I think, we need also to shift acc guards, so no include is inside a `#if ALPAKA_ACC_<x>_ENABLED` section.
> 
> Probably related, this is a problem:
> 
> ```c++
> #ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
> #    include "alpaka/atomic/AtomicAtomicRef.hpp"
> #else
> #    include "alpaka/atomic/AtomicStdLibLock.hpp"
> #endif // ALPAKA_DISABLE_ATOMIC_ATOMICREF
> ```
> 
> Because the amalgamate script will inline all headers inside the first branch and then skip them in the else branch, because they are already included :) But if `ALPAKA_DISABLE_ATOMIC_ATOMICREF` is not defined, and the else branch is taken, all headers are missing.

This is, what I mean. I think, we need to change it to

```c++
#include "alpaka/atomic/AtomicAtomicRef.hpp"
#include "alpaka/atomic/AtomicStdLibLock.hpp"
```

AtomicAtomicRef.hpp
```c++
#pragma once
#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
  ...
#endif
```

AtomicStdLibLock.hpp
```c++
#pragma once
#ifdef ALPAKA_DISABLE_ATOMIC_ATOMICREF
  ...
#endif
```

bernhardmgruber:
@SimeonEhrig I did exactly that and it solved the issue. But there are a lot more. Basically every header that can disable some content including an alpaka include needs to be fixed. This is too much work for me now. If you have time, please have a go at it :)

SimeonEhrig:
> @SimeonEhrig I did exactly that and it solved the issue. But there are a lot more. Basically every header that can disable some content including an alpaka include needs to be fixed. This is too much work for me now. If you have time, please have a go at it :)

I had the fear that will happen :-( Let's see, I think it's a good task, when my brain is over ;-)

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -11,34 +11,34 @@
...
-#    include <alpaka/core/DemangleTypeNames.hpp>
-#    include <alpaka/idx/bt/IdxBtZero.hpp>
-#    include <alpaka/idx/gb/IdxGbRef.hpp>
-#    include <alpaka/intrinsic/IntrinsicCpu.hpp>
-#    include <alpaka/math/MathStdLib.hpp>
-#    include <alpaka/mem/fence/MemFenceOmp2Blocks.hpp>
-#    include <alpaka/rand/RandStdLib.hpp>
-#    include <alpaka/warp/WarpSingleThread.hpp>
-#    include <alpaka/workdiv/WorkDivMembers.hpp>
+#    include "alpaka/atomic/AtomicCpu.hpp"
```
psychocoderHPC: @bernhardmgruber Could you please push the first commit as tools author: 

```bash
GIT_AUTHOR_NAME="Tools" GIT_AUTHOR_EMAIL="alpaka@hzdr.de" git commit ...
```

psychocoderHPC (changes requested):
I love this PR :heart:

include/alpaka/acc/AccCpuOmp2Blocks.hpp
bernhardmgruber: Done

bernhardmgruber:
@SimeonEhrig I will let you take over! Just push to the PR as you please.

bernhardmgruber:
alpaka vectoradd on godbolt: https://godbolt.org/z/oG3nqEW84

psychocoderHPC:
@bernhardmgruber I assume you use currently the header `alapka.hpp` as base to create the single header. @SimeonEhrig  and I had a short discussion if it maybe makes sense to use the alpaka stand-alone header as a base and provide a hpp for each backend alpaka supports?
In this case, solving the includes and dependencies e.g. boost would be possible with e.g. a simple `g++ -E` to solve all macros.

SimeonEhrig:
> @bernhardmgruber I assume you use currently the header `alapka.hpp` as base to create the single header. @SimeonEhrig and I had a short discussion if it maybe makes sense to use the alpaka stand-alone header as a base and provide a hpp for each backend alpaka supports? In this case, solving the includes and dependencies e.g. boost would be possible with e.g. a simple `g++ -E` to solve all macros.

@psychocoderHPC Means, using some of the headers in [include/alpaka/standalone](https://github.com/alpaka-group/alpaka/tree/develop/include/alpaka/standalone) as start point and include the specefic header in godbolt to enable an acc. Should be much easier, than setting compiler arguments.

fwyzard:
Just personally, I prefer a single header that includes all the back-ends, because I often try to enable as many of them as possible :-)

But it's true that it's an unlikely use case for godbolt, since there we can use at most one back-end at a time.

bernhardmgruber:
@psychocoderHPC and @SimeonEhrig:

> @bernhardmgruber I assume you use currently the header `alapka.hpp` as base to create the single header. @SimeonEhrig and I had a short discussion if it maybe makes sense to use the alpaka stand-alone header as a base and provide a hpp for each backend alpaka supports? In this case, solving the includes and dependencies e.g. boost would be possible with e.g. a simple `g++ -E` to solve all macros.

I am strongly against making the amalgamated header behave any different than the ordinary `alpaka.hpp`. I want to take an alpaka program, paste it into godbolt, change the one include, set compiler flags, and be done.

This also gives us another way of distributing alpaka. You can just drop a single header file into your project and start using it. I know it's not "proper" dependency management. But for some quick prototyping, or a hackathon, this is actually quite useful.

Involving the compiler to generate a amalgamated header is difficult, because you don't want to expand the system includes and third-party dependencies. You want a single `alpaka.hpp` and still be able to use it with various compilers, different OSes or different Boost versions. You only want to resolve alpaka's internal includes.

@fwyzard:

> Just personally, I prefer a single header that includes all the back-ends, because I often try to enable as many of them as possible :-)
> But it's true that it's an unlikely use case for godbolt, since there we can use at most one back-end at a time.

I also prefer having all backends. Then you can just switch e.g. between serial and omp, and check whether a kernel still vectorizes the same, or whether enabling a further backend in addition causes an optimization to be missed, etc. You want the amalgamated header to behave like the ordinary alpaka library.

bernhardmgruber:
Btw, works with CUDA as well: https://godbolt.org/z/PGeoEe3Tj

bernhardmgruber:
This is ready from my perspective.

j-stephan:
Could you force-push once again? Your last commit wasn't tested on the HZDR CI, I guess it must have been swallowed by some connection error. The last pipeline was triggered a day before your most recent commit.

include/alpaka/acc/AccCpuSerial.hpp
```diff
@@ -4,39 +4,39 @@
...
+#include "alpaka/warp/WarpSingleThread.hpp"
+#include "alpaka/workdiv/WorkDivMembers.hpp"
 
 // Specialized traits.
-#    include <alpaka/acc/Traits.hpp>
-#    include <alpaka/dev/Traits.hpp>
-#    include <alpaka/idx/Traits.hpp>
-#    include <alpaka/kernel/Traits.hpp>
-#    include <alpaka/pltf/Traits.hpp>
+#include "alpaka/acc/Traits.hpp"
```
psychocoderHPC, line 25: @bernhardmgruber Do we enforce the include with `"` in our clang-format file? 
I can not see it in the diff but maybe miss something.

bernhardmgruber: AFAIK clang-format cannot enforce that. But the idea was valuable, because I also had to fix something in the `.clang-format` to recognize the quotes now used to include alpaka stuff.

psychocoderHPC:
The CI [e126342](https://github.com/alpaka-group/alpaka/pull/1956/commits/e12634281d6207e29f1dd3841d6e933f1233a20f) passed but for unknown reasons the status was not reported back to github. 
@j-stephan or me can merge it with our admin rights, no need to open a new PR to make github happy.

.github/workflows/single-header.yml
```diff
@@ -0,0 +1,34 @@
...
+      - develop
+
+jobs:
+  single-header:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - name: clone single-header
+      run: |
+        git clone -b single-header https://x-access-token:${{secrets.github_token}}@github.com/${{github.repository}}.git single-header
```
psychocoderHPC, line 17: Is the branch single-header only updated if we merge something to the develop branch or with any PR.
I am wondering because there is already a branch single header but I am not sure if @bernhardmgruber created the branch by hand or if this branch was auto-generated.

psychocoderHPC (dismissed):
One question about the time when the single-header branch update.

.github/workflows/single-header.yml
psychocoderHPC: OK my question is answered with line 8 https://github.com/alpaka-group/alpaka/pull/1956/files/e12634281d6207e29f1dd3841d6e933f1233a20f#diff-88ac08689a99a939e350a0c6fc02c3a5b93c0cf3637202361f6831ab3d0dc5aaR8

Only updated if we push to develop.

psychocoderHPC:
@bernhardmgruber IS it possible to add one line about the usage to the documentation? I am wondering if I need to set any defines to enable accelerators. e.g. if I like to compile for CUDA, serial or ROCm

.github/workflows/single-header.yml
bernhardmgruber: Yes, as you have found out yourself, the `single-header` branch is updated on each push to the `develop` branch. I have created this branch manually and put an amalgamated `alpaka.hpp` there, to test manually whether the amalgamation script works. The name of the branch from which this PR is opened is called `single_header` (underbar vs. dash) and is an unrelated name.

bernhardmgruber:
> @bernhardmgruber IS it possible to add one line about the usage to the documentation? I am wondering if I need to set any defines to enable accelerators. e.g. if I like to compile for CUDA, serial or ROCm

Done. I wanted to point to a list of all backend macros, but there exists none in the documentation. Someone should add one :)

j-stephan:
Could you squash your commits? Adding six commits for this change to the git history seems a little excessive. Maybe have one for the `Tools` user and another for all your manual changes.

bernhardmgruber:
> Could you squash your commits? Adding six commits for this change to the git history seems a little excessive. Maybe have one for the `Tools` user and another for all your manual changes.

I squashed the related changes.

include/alpaka/block/shared/dyn/BlockSharedMemDynUniformCudaHipBuiltIn.hpp
```diff
@@ -4,13 +4,13 @@
...
 
-#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+#include "alpaka/block/shared/dyn/Traits.hpp"
+#include "alpaka/core/BoostPredef.hpp"
+#include "alpaka/core/Concepts.hpp"
 
-#    include <alpaka/block/shared/dyn/Traits.hpp>
-#    include <alpaka/core/BoostPredef.hpp>
-#    include <alpaka/core/Concepts.hpp>
+#include <type_traits>
```
j-stephan, line 11: General question: Do we really want to include the standard library headers if the rest of the file is not enabled?

bernhardmgruber: Yes, this is necessary for the amalgamation tool to work, because it will remove all includes after the first one. If the first include is only conditionally included and the condition is not always true (which happens), the include never happens in the amalgamated header at all, leading to missing declarations.

psychocoderHPC (approved):

PR #1911, opened 2023-02-06
Update to Catch2 v3.3.2

j-stephan:
See title.

PR #1958, opened 2023-06-15
Disable MSVC + CUDA jobs

j-stephan:
This PR disables the MSVC + CUDA jobs while we are waiting for NVIDIA to fix `nvcc` on WIndows (see #1911 for more details). This is only intended as a temporary measure so we can finally merge #1911 which in turn is required for #1845 to continue (`icpx` doesn't like our currently used slightly outdated Catch2 version).


bernhardmgruber:
I would downgrade the nvcc version for MSVC to the last working version instead of making us completely blind on the Windows/nvcc CI. But I leave this decision up to others.

j-stephan:
I'm not certain this would fix the problem. We upgraded to CUDA 12 first, then I opened the Catch2 PR. I never tested it with older nvcc releases.

bernhardmgruber (approved):
Fine, let's merge it. But I am making you personally responsible to retry and readd the CUDA/MSVC run once there is a working combination again :)

PR #1970, opened 2023-06-21
Update SYCL CMake and remove Xilinx support

### LINKED ISSUE #1602: Add Xilinx FPGA codepaths to SYCL back-end
Right now all of the possible SYCL targets use the `parallel_for` kernel instantiation. Aside from this being currently [broken](https://github.com/triSYCL/sycl/issues/154) I believe we can achieve higher performance by using `single_task` instead and relying on Xilinx' SYCL extensions for loop unrolling, pipelining and dataflow.

### LINKED ISSUE #1935: Add SYCL compiler flags when backend is enabled
Compiling with `dpcpp` and the SYCL backend enabled, I get the following warning, turned into an error:
`icpx: error: use of 'dpcpp' is deprecated and will be removed in a future release. Use 'icpx -fsycl' [-Werror,-Wdeprecated]`

If `dpcpp` is being deprecated, we should also react in our own cmake scripts and include `-fsycl` when the SYCL backend is enabled and `icpx` is used as compiler.

j-stephan:
Since we are now requiring CMake >= 3.22 we can make use of native CMake support for `icpx` (or `IntelLLVM` in CMake lingo). This reduces the required command line to compile for Intel GPUs to this:

```bash
$ cmake -DCMAKE_CXX_COMPILER=icpx -Dalpaka_ACC_SYCL_ENABLE=ON -Dalpaka_SYCL_ONEAPI_GPU=ON -Dalpaka_SYCL_ONEAPI_GPU_DEVICES="icllp" ...
```
This PR also removes the support for Xilinx FPGAs since their USM support doesn't seem to go anywhere soon.

Closes #1602. Fixes #1935.


README_SYCL.md
```diff
@@ -40,6 +25,8 @@ The following CMake flags can be set for Intel CPUs:
 
 ### Building for Intel FPGAs
 
+Note: Intel FPGAs cannot be targeted together with other Intel hardware. This is because of different compiler trajectories as explained in the Intel oneAPI documentation.
```
bernhardmgruber, line 28: Could you add a link to the oneAPI documentation here please, thx!

bernhardmgruber (approved):
I have to trust you for the cmake changes, because I have no idea what's required there. Maybe @fwyzard wants to review this. Otherwise, we can have a second go at it when #1845 comes in.

README_SYCL.md
j-stephan: There is one at the bottom in the general notes :-)

bernhardmgruber: Fine.

fwyzard:
I do not have access to a Xilinx FPGA, so I cannot say much about this.

Assuming that eventually they implement USM support, how much work is it to add them back to Alpaka ?

j-stephan:
If they do add USM support at some point it would be a matter of adding the CMake flags, the accelerator and the `TaskKernel` I think. Most of the rest of the SYCL back-end should be the same as for oneAPI.

cmake/alpakaCommon.cmake
```diff
@@ -575,82 +561,72 @@ if(alpaka_ACC_SYCL_ENABLE)
...
-        
-        target_compile_definitions(alpaka INTERFACE "ALPAKA_SYCL_ONEAPI_GPU")
-        target_link_options(alpaka INTERFACE "SHELL:-Xsycl-target-backend=${alpaka_SYCL_ONEAPI_GPU_TARGET} \"-device ${alpaka_SYCL_ONEAPI_GPU_DEVICES}\"")
-    endif()
+        if(alpaka_SYCL_ONEAPI_GPU)
+            # Create a drop-down list (in cmake-gui) of valid Intel GPU targets. On the command line the user can specifiy
+            # additional targets, such as ranges: "Gen8-Gen12LP" or lists: "icclp;skl".
+            set(alpaka_SYCL_ONEAPI_GPU_DEVICES "bdw" CACHE STRING "Intel GPU devices / generations to compile for")
+            set_property(CACHE alpaka_SYCL_ONEAPI_GPU_DEVICES
+                        PROPERTY STRINGS "bdw;skl;kbl;cfl;bxt;glk;whl;aml;cml;icllp;lkf;ehl;tgllp;rkl;adl-s;adl-p;dg1;acm-g10;ats-m150;dg2-g10;acm-g11;ats-m75;dg2-g11;acm-g12;dg2-g12;pvc-sdv;pvc;gen11;gen12lp;gen8;gen9;xe;xe-hpc;xe-hpg")
```
SimeonEhrig, line 610: I this line something like setting the `GPU_TARGETS` in HIP?

j-stephan: Yep. As a user you provide a comma-separated list of Intel GPU targets you want to compile for.

bernhardmgruber:
I will merge this to unblock further work on the SYCL backend. Given the CI passed, I assume your changes did not break the other backends. But as I said before, I my experience on the SYCL and oneAPI compilation model is limited.

PR #1973, opened 2023-06-21
Add Xcode 14.3.1 test runner

j-stephan:


bernhardmgruber:
Given the hoops you had to jump through here to make OMP work on MacOS, I really wonder whether we should stop supporting the backend there.

j-stephan:
This is purely a CI issue. I don't know what GitHub are doing to the macOS runners, but this problem doesn't occur on normal macOS installations.

bernhardmgruber (approved):

PR #1974, opened 2023-06-21
Rework `BOOST_COMP_HIP` and `BOOST_LANG_HIP`

j-stephan:
Currently `BOOST_COMP_HIP` can only detect whether we are compiling with clang targeting HIP. This PR adds a version detection so we can have checks like these:

```c++
#if BOOST_COMP_HIP >= BOOST_VERSION_NUMBER(5, 5, 0)
```

### DIFF
--- a/include/alpaka/core/BoostPredef.hpp
+++ b/include/alpaka/core/BoostPredef.hpp
@@ -1,4 +1,5 @@
-/* Copyright 2023 Benjamin Worpitz, Matthias Werner, Jan Stephan
+/* Copyright 2023 Benjamin Worpitz, Matthias Werner, René Widera, Sergei Bastrakov, Jeffrey Kelling,
+ *                Bernhard Manfred Gruber, Jan Stephan
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -12,18 +13,16 @@
 #endif
 
 //---------------------------------------HIP-----------------------------------
-// __HIPCC__ is defined by hipcc (if either __CUDACC__ is defined)
+// __HIP__ is defined by both hip-clang and vanilla clang in HIP mode.
 // https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary
 #if !defined(BOOST_LANG_HIP)
-#    if defined(__HIPCC__) && (defined(__CUDACC__) || defined(__HIP__))
-#        include <hip/hip_runtime.h>
-// HIP defines "abort()" as "{asm("trap;");}", which breaks some kernels
-#        undef abort
+#    if defined(__HIP__)
+/* BOOST_LANG_CUDA is enabled when either __CUDACC__ (nvcc) or __CUDA__ (clang) are defined. This occurs when
+   nvcc / clang encounter a CUDA source file. Since there are no HIP source files we treat every source file
+   as HIP when we are using a HIP-capable compiler. */
+#        include <hip/hip_version.h>
+// HIP doesn't give us a patch level for the last entry, just a gitdate
 #        define BOOST_LANG_HIP BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0)
-#        if defined(BOOST_LANG_CUDA) && BOOST_LANG_CUDA
-#            undef BOOST_LANG_CUDA
-#            define BOOST_LANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
-#        endif
 #    else
 #        define BOOST_LANG_HIP BOOST_VERSION_NUMBER_NOT_AVAILABLE
 #    endif
@@ -40,10 +39,12 @@
 #    endif
 #endif
 
-// hip compiler detection
+// HIP compiler detection
 #if !defined(BOOST_COMP_HIP)
-#    if defined(__HIP__)
-#        define BOOST_COMP_HIP BOOST_VERSION_NUMBER_AVAILABLE
+#    if defined(__HIP__) // Defined by hip-clang and vanilla clang in HIP mode.
+#        include <hip/hip_version.h>
+// HIP doesn't give us a patch level for the last entry, just a gitdate
+#        define BOOST_COMP_HIP BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0)
 #    else
 #        define BOOST_COMP_HIP BOOST_VERSION_NUMBER_NOT_AVAILABLE
 #    endif
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -12,6 +12,11 @@
 #    include <intrin.h>
 #endif
 
+#if BOOST_LANG_HIP
+// HIP defines some keywords like __forceinline__ in header files.
+#    include <hip/hip_runtime.h>
+#endif
+
 //! All functions that can be used on an accelerator have to be attributed with ALPAKA_FN_ACC or ALPAKA_FN_HOST_ACC.
 //!
 //! \code{.cpp}


fwyzard:
There is already `BOOST_LANG_HIP` some 20 lines above.

j-stephan:
From what I understand that is used to detect if we are compiling a HIP file while `BOOST_COMP_HIP` is used to detect whether the compiler is HIP-capable. But I'm not fully sure why we have all the different HIP checks.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -42,8 +42,9 @@
 
 // hip compiler detection
 #if !defined(BOOST_COMP_HIP)
-#    if defined(__HIP__)
-#        define BOOST_COMP_HIP BOOST_VERSION_NUMBER_AVAILABLE
+#    if defined(__HIP__) && defined(__HIPCC__) // We are using clang to compile HIP code
+#        include <hip/hip_version.h>
+#        define BOOST_COMP_HIP BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, HIP_VERSION_PATCH)
```
fwyzard: Just a note that `HIP_VERSION_PATCH` may not be what you think it is: in HIP/ROCm 5.4.3 they use
```c++
#define HIP_VERSION_MAJOR 5
#define HIP_VERSION_MINOR 4
#define HIP_VERSION_PATCH 22804
```
so the last number is a bit complicated to use :)

j-stephan: Thanks for the warning!

fwyzard: OK, turns out that `HIP_VERSION` is defined in the same way as `BOOST_COMP_HIP` (if I counted the digits correctly).

In fact, in `include/alpaka/core/ApiHipRt.hpp` I use `HIP_VERSION` instead of relying on one of the others, because I need to check what version of HIP is available while compiling host code with a host compiler (e.g. `gcc`).

psychocoderHPC:
As @fwyzard pointed out the VERSION numbers have strange behaviour, please see an older discussion https://github.com/alpaka-group/alpaka/pull/1914#discussion_r1104141221

j-stephan:
Okay. I think a lot of (my) confusion comes from the different behaviour between CUDA and HIP:

* `BOOST_LANG_CUDA` is defined when `nvcc` or `clang` encounter a `*.cu` file.
* `BOOST_COMP_NVCC` is defined when `nvcc` is used as a compiler for either `*.cpp` or `*.cu` files.
* `BOOST_COMP_HIP` is defined when `hip-clang` is used as a compiler for a `*.cpp` (or `*.cu`?) file.
* There is no real trigger for `BOOST_LANG_HIP` since there are no `*.hip` source files. Right now this is just another (convoluted) check boiling down to the same as `BOOST_COMP_HIP`.
* What @fwyzard wanted in the February PR is an HIP API version check. This is not related to compilers / languages at all.

My suggestion: Get rid of `BOOST_LANG_HIP` and maybe introduce something like `ALPAKA_API_{CUDA; HIP}` to check for the API versions of the CUDA/HIP runtimes.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,26 +13,24 @@
...
 
 //---------------------------------------HIP-----------------------------------
-// __HIPCC__ is defined by hipcc (if either __CUDACC__ is defined)
+// __HIPCC__ is defined by hipcc when either __CUDACC__, __HIP__ or __HCC__ are defined.
 // https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary
 #if !defined(BOOST_LANG_HIP)
-#    if defined(__HIPCC__) && (defined(__CUDACC__) || defined(__HIP__))
-#        include <hip/hip_runtime.h>
-// HIP defines "abort()" as "{asm("trap;");}", which breaks some kernels
-#        undef abort
```
j-stephan, line 21: This was problematic when we targeted CUDA devices through HIP. Since we no longer do this we can remove this.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,26 +13,24 @@
...
+/* BOOST_LANG_CUDA is enabled when either __CUDACC__ (nvcc) or __CUDA__ (clang) are defined. This occurs when
+   nvcc / clang encounter a CUDA source file. Since there are no HIP source files we treat every source file
+   as HIP when we are using a HIP-capable compiler. */
+#        include <hip/hip_version.h>
+// HIP doesn't give us a patch level for the last entry, just a gitdate
 #        define BOOST_LANG_HIP BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0)
-#        if defined(BOOST_LANG_CUDA) && BOOST_LANG_CUDA
-#            undef BOOST_LANG_CUDA
-#            define BOOST_LANG_CUDA BOOST_VERSION_NUMBER_NOT_AVAILABLE
-#        endif
```
j-stephan, line 26: Same as above.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,26 +13,24 @@
...
 #    else
 #        define BOOST_LANG_HIP BOOST_VERSION_NUMBER_NOT_AVAILABLE
 #    endif
 #endif
 
 // HSA device architecture detection (HSA generated via HIP(clang))
 #if !defined(BOOST_ARCH_HSA)
-#    if defined(__HIP_DEVICE_COMPILE__) && __HIP_DEVICE_COMPILE__ == 1 && defined(__HIP__)
+#    if defined(__HIP_DEVICE_COMPILE__) && __HIP_DEVICE_COMPILE__ == 1
```
j-stephan: `__HIP_DEVICE_COMPILE__` implies `__HIP__` since we don't support other HIP compilers anymore.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -40,10 +39,12 @@
...
 
-// hip compiler detection
+// HIP compiler detection
 #if !defined(BOOST_COMP_HIP)
-#    if defined(__HIP__)
-#        define BOOST_COMP_HIP BOOST_VERSION_NUMBER_AVAILABLE
+#    if defined(__HIPCC__) // We are using a HIP-capable compiler; could be hcc, hip-clang or nvcc
+#        include <hip/hip_version.h>
+// HIP doesn't give us a patch level for the last entry, just a gitdate
+#        define BOOST_COMP_HIP BOOST_VERSION_NUMBER(HIP_VERSION_MAJOR, HIP_VERSION_MINOR, 0)
```
j-stephan, line 47: Essentially, `BOOST_COMP_HIP` and `BOOST_LANG_HIP` do the same thing since there is no special HIP language we could test for. If we use a HIP-capable compiler we assume everything it touches is a "HIP source file".

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,26 +13,24 @@
 #endif
 
 //---------------------------------------HIP-----------------------------------
-// __HIPCC__ is defined by hipcc (if either __CUDACC__ is defined)
+// __HIPCC__ is defined by hipcc when either __CUDACC__, __HIP__ or __HCC__ are defined.
 // https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary
 #if !defined(BOOST_LANG_HIP)
-#    if defined(__HIPCC__) && (defined(__CUDACC__) || defined(__HIP__))
-#        include <hip/hip_runtime.h>
```
psychocoderHPC, line 19: ~I do not think you can remove the include else `HIP_VERSION_MAJOR` a few lines later will not be defined.~~

[update] it is still in I missed it

j-stephan: The include is replaced by `<hip/hip_version.h>` a few lines below.

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,26 +13,24 @@
...
-// HIP defines "abort()" as "{asm("trap;");}", which breaks some kernels
-#        undef abort
+#    if defined(__HIPCC__)
```
psychocoderHPC: Not sure if you can remove `__HIP__` because you can compile a pure C++ file with hipcc only if you set `-x hip` it will be use th e language HIP.

IMO we should not assume only because __HIPCC__ is set what is a marker for the compiler HIP is active.
Additional if we would like to use vanilla clang `__HIPCC__` is not set. We do not test currently vanilla clang but should not remove support for it.

include/alpaka/core/BoostPredef.hpp
psychocoderHPC: As I wrote above please keep `__HIP__` to not remove vanilla clang support.

include/alpaka/core/BoostPredef.hpp
psychocoderHPC: Could be that the current CI issue is triggered by this change

```
[ 57%] Building CXX object test/unit/acc/CMakeFiles/accTest.dir/src/AccNameTest.cpp.o
cd /builds/hzdr/crp/alpaka/build/test/unit/acc && /opt/rocm/llvm/bin/clang++ -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -DALPAKA_ACC_GPU_HIP_ENABLED -DALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=47 -DALPAKA_CI -DALPAKA_DEBUG=0 -DALPAKA_OFFLOAD_MAX_BLOCK_SIZE="" -DBOOST_ALL_NO_LIB -D__HIP_PLATFORM_AMD__=1 -D__HIP_PLATFORM_HCC__=1 -I/builds/hzdr/crp/alpaka/include -isystem /builds/hzdr/crp/alpaka/thirdParty/catch2/src/catch2/.. -isystem /builds/hzdr/crp/alpaka/build/thirdParty/catch2/generated-includes -isystem /opt/boost/1.75.0/include -isystem /opt/rocm-5.0.0/hip/include -isystem /opt/rocm/include -isystem /opt/rocm-5.0.0/hip/../include -isystem /opt/rocm/llvm/lib/clang/14.0.0/include/.. -isystem /opt/rocm/rocrand/include -isystem /opt/rocm/hiprand/include -O3 -DNDEBUG -Weverything -Wno-c++98-compat -Wno-c++98-compat-pedantic -Wno-global-constructors -Wno-padded -Wno-extra-semi-stmt -Wno-unused-command-line-argument "-Wno-error=#warnings" -std=c++17 -pthread -mllvm -amdgpu-early-inline-all=true -mllvm -amdgpu-function-calls=false -x hip --offload-arch=gfx906 -std=gnu++17 -MD -MT test/unit/acc/CMakeFiles/accTest.dir/src/AccNameTest.cpp.o -MF CMakeFiles/accTest.dir/src/AccNameTest.cpp.o.d -o CMakeFiles/accTest.dir/src/AccNameTest.cpp.o -c /builds/hzdr/crp/alpaka/test/unit/acc/src/AccNameTest.cpp
/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp:25:12: error: no matching function for call to 'floatEqualExactNoWarning'
    return alpaka::math::floatEqualExactNoWarning(a, b);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/builds/hzdr/crp/alpaka/test/unit/atomic/src/AtomicTest.cpp:30:12: error: no matching function for call to 'floatEqualExactNoWarning'
    return alpaka::math::floatEqualExactNoWarning(a, b);
           ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

psychocoderHPC (changes requested):

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,18 +13,16 @@
...
 //---------------------------------------HIP-----------------------------------
-// __HIPCC__ is defined by hipcc (if either __CUDACC__ is defined)
+// __HIPCC__ is defined by hipcc when either __CUDACC__, __HIP__ or __HCC__ are defined.
 // https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary
 #if !defined(BOOST_LANG_HIP)
-#    if defined(__HIPCC__) && (defined(__CUDACC__) || defined(__HIP__))
-#        include <hip/hip_runtime.h>
-// HIP defines "abort()" as "{asm("trap;");}", which breaks some kernels
-#        undef abort
+#    if defined(__HIPCC__) && defined(__HIP__)
```
psychocoderHPC: This should be `defined(__HIPCC__) && defined(__HIP__)` but `defined(__HIP__)` is enough too.

psychocoderHPC: This is the reason why the CI is crashing.

psychocoderHPC (changes requested):

include/alpaka/core/BoostPredef.hpp
```diff
@@ -12,18 +13,16 @@
 #endif
 
 //---------------------------------------HIP-----------------------------------
-// __HIPCC__ is defined by hipcc (if either __CUDACC__ is defined)
+// __HIP__ is defined by both hip-clang and vanilla clang in HIP mode.
 // https://github.com/ROCm-Developer-Tools/HIP/blob/master/docs/markdown/hip_porting_guide.md#compiler-defines-summary
 #if !defined(BOOST_LANG_HIP)
-#    if defined(__HIPCC__) && (defined(__CUDACC__) || defined(__HIP__))
-#        include <hip/hip_runtime.h>
```
j-stephan, line 19: This was actually the reason for the CI failures. `__forceinline__` is not a HIP keyword but defined somewhere in the header files included by `hip_runtime.h`.

include/alpaka/core/Common.hpp
```diff
@@ -80,13 +80,13 @@
...
-#if BOOST_LANG_CUDA || BOOST_LANG_HIP
+#if BOOST_LANG_CUDA
 #    define ALPAKA_FN_INLINE __forceinline__
 #elif BOOST_COMP_MSVC || defined(BOOST_COMP_MSVC_EMULATED)
 // TODO: With C++20 [[msvc::forceinline]] can be used.
 #    define ALPAKA_FN_INLINE __forceinline
 #else
-// For gcc, clang, and clang-based compilers like Intel icpx
+// For gcc, clang, and clang-based compilers like Intel icpx or hip-clang
 #    define ALPAKA_FN_INLINE [[gnu::always_inline]] inline
```
j-stephan: Internally, HIP transforms `__forceinline__` to `[[gnu::always_inline]]`, so we can just use this path directly.

psychocoderHPC: I am wondering about the new error in the CI:

```
/builds/hzdr/crp/alpaka/include/alpaka/meta/CudaVectorArrayWrapper.hpp:293:28: error: an attribute list cannot appear here
        ALPAKA_FN_HOST_ACC ALPAKA_FN_INLINE constexpr value_type& operator[]([[maybe_unused]] int const k) noexcept
                           ^~~~~~~~~~~~~~~~
/builds/hzdr/crp/alpaka/include/alpaka/core/Common.hpp:90:30: note: expanded from macro 'ALPAKA_FN_INLINE'
#    define ALPAKA_FN_INLINE [[gnu::always_inline]] inline
```

The attribute is the most left before are keys e.g. static, constexpr, ... I do not understand why this error is shown.

psychocoderHPC: In llvm I can only find for CUDA the definition of `__forceinline__`

```
#define __forceinline__ __device__ __inline__ __attribute__((always_inline))
```

For HIP this is not defined in llvm.
HIPRTC contains the definition in CMake only
```
#define __forceinline__ inline __attribute__((always_inline))\n\
```

j-stephan: No matter. I restored the old behaviour which *hopefully* makes this whole mess compile.

j-stephan:
@psychocoderHPC Ping!

psychocoderHPC (approved):

PR #1977, opened 2023-06-23
Modernize CMake

### LINKED ISSUE #958: Cmake switch for -Werror
Is there a way to have a cmake switch to disable the -Werror compiler flag?

There are two reasons for this request:
1. The OpenMP4 backend triggers compiler warnings in clang about types note possibly being not mappable which might be technically correct, but  ultimately are pointless an cannot be avoided. In current clang releases, these could be turned off only with a flag that would also disable all other openmp (offload?) warnings, thus should not be done globally. Clang 11 introduces `-Wopenmp-mapping`. The problem with that hard-coded -Werror is thus, that while the code (examples/tests) can only be compiled without it, when point compiler vendors to the alpaka source, they would first get the warning elevated to an error.
The OpenMP4 backend in the develop branch also triggerst this warning my mapping `boundKernelFnObj`, which would actually not work if offloading were requested.
2. Convenience when debugging backends.

### LINKED ISSUE #1060: RelWithDebInfo & CUDA -G
The CMake option `RelWithDebInfo` is runtime performance neutral and enables all optimizations. On Linux, debug symbols are stored in a separate part of the binary and with MSVC in separate files.

Therefore, we should not add `-G` debug with `RelWithDebInfo` as this turns all device optimizations off:
https://github.com/alpaka-group/alpaka/blob/0.5.0/cmake/alpakaCommon.cmake#L514

Without checking, this is probably similar with HIP:
https://github.com/alpaka-group/alpaka/blob/0.5.0/cmake/alpakaCommon.cmake#L642

References:
- https://cmake.org/cmake/help/v3.17/variable/CMAKE_BUILD_TYPE.html
- https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html#options-for-altering-compiler-linker-behavior-generate-line-info
- details: https://reviews.llvm.org/D55056#1314996 (only with MSVC a "problem", but we inline everything we want to anyway;  https://www.brianlheim.com/2018/04/09/cmake-cheat-sheet.html#build-types)

j-stephan:
This PR is a small convenience update for our CMake build system (mostly for the tests). Notable changes:

* `rand/Traits.hpp` was missing the `<cstdint>` include which led to a build error on my system.
* `-Werror` is no longer enabled by default when `BUILD_TESTING` is set to `ON`. If desired users can set `alpaka_ENABLE_WERROR` to compile with `-Werror`; this is also done in the CI. Fixes #958.
* In `Debug` mode nvcc will now make use of the`-G` flag (device-side debug information) which CMake doesn't enable by default. Fixes #1060.
* In `Debug` mode gcc will use the `-Og` optimization level.
* In `Debug` mode clang, AppleClang and icpx will use the `-O0` optimization level.
* In `Debug` mode MSVC will use the `/Od` optimization level.
* Fixed a few minor CMake bugs and typos I found along the way.


cmake/alpakaCommon.cmake
```diff
@@ -178,6 +178,15 @@ else()
             target_link_libraries(alpaka INTERFACE ${RT_LIBRARY})
         endif()
     endif()
+
+    # Add debug optimization levels. CMake doesn't do this by default.
+    # Note that -Og is the recommended gcc optimization level for debug mode but is equivalent to -O1 for clang (and its derivates).
+    alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>>:-Og>"
+                                                          "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:Clang,AppleClang,IntelLLVM>>:-O0>"
+                                                          "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:MSVC>>:/Od>")
```
bernhardmgruber: Why can't we enable `-Og` for clang as well? IMO it should not matter which optimization it entails, as long as debug experience is not hindered.

test/common/devCompileOptions.cmake
```diff
@@ -1,187 +1,154 @@
...
+
+    # Flags added in Visual Studio 2013
+    list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CXX>:/Zc:throwingNew> $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/Zc:throwingNew>)
+    list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CXX>:/Zc:strictStrings> $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/Zc:strictStrings>)
+
+    # Flags added in Visual Studio 2015
+    list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CXX>:/permissive-> $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/permissive->)
+
+    # Flags added in Visual Studio 2017
+    list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CXX>:/Zc:twoPhase-> $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/Zc:twoPhase->)
```
bernhardmgruber: I think this one should be dropped. Why weaken standard conformance and disable two-phase lookup?

j-stephan: This was already present in the old version, you'll have to ask the original author(s) of this section.

cmake/alpakaCommon.cmake
j-stephan: See the comment above. `-Og` is equivalent to `-O1` for LLVM compilers so it might affect debug information. The GNU compiler recommends `-Og` because the enabled optimizations don't hurt the debug experience.

SimeonEhrig (dismissed):

README.md
```diff
@@ -65,20 +65,21 @@ Supported Compilers
...
+| [HIP](https://alpaka.readthedocs.io/en/latest/install/HIP.html) (clang)        | -                                               | -                                               | -                                           | :x:                    | :x:                                                        | :x:                                                   | :x:                                             | :x:                                         | :white_check_mark: (HIP 5.0 - 5.2)                | :white_check_mark: (HIP 5.3 - 5.4)        | :white_check_mark: (HIP 5.5)              | :x:                | -                                                     | -                                    |
 
 Other compilers or combinations marked with :x: in the table above may work but are not tested in CI and are therefore not explicitly supported.
 
-[^1]: Due to a [compiler bug](https://github.com/llvm/llvm-project/issues/58491) in debug mode only release builds are supported.
+[^1]: Due to a [LLVM bug](https://github.com/llvm/llvm-project/issues/58491) in debug mode only release builds are supported.
```
bernhardmgruber:
```suggestion
[^1]: Due to an [LLVM bug](https://github.com/llvm/llvm-project/issues/58491) in debug mode only release builds are supported.
```

bernhardmgruber (dismissed):

bernhardmgruber (approved):

PR #1988, opened 2023-06-26
Make alpaka platforms full objects

### LINKED ISSUE #1938: Make alpaka platform an actual object
Currently, the alpaka platform is just a type that is not instantiatable. We discussed in the alpaka VC today that it would be useful to lift this requirement and actually make it mandatory in the corresonding APIs to pass an object instead of just the type. This would allow us to store backend specific implementation that needs to exist once for the entire backend in this platform. We could get rid of some `static` variables inside alpaka this way and give the user more control when such platform global state is constructed and destroyed. It could also provide a good basis to address #1927 since we could now keep a vector of devices in the platform.

bernhardmgruber:
Fixes: #1938

For now, I have only converted the `vectorAdd` example. We should discuss how the user-facing API looks now and whether we can improve something here.

Questions:

- [x] Should alpaka have a global singleton for the host CPU platform? Answer: No, this is up to the user. Alpaka should allow for deterministic construction and destruction and therefore not use any globals or `static`s.
- [X] Should alpaka have a global singleton for the host CPU device? We had discussions in the past whether we want to have multiple host CPU devices for multi-processor systems. Answer: No, for the same reason as above.

### DIFF
--- a/example/vectorAdd/src/vectorAdd.cpp
+++ b/example/vectorAdd/src/vectorAdd.cpp
@@ -85,7 +85,8 @@ auto main() -> int
     using QueueAcc = alpaka::Queue<Acc, QueueProperty>;
 
     // Select a device
-    auto const devAcc = alpaka::getDevByIdx<Acc>(0u);
+    auto const platform = alpaka::Pltf<Acc>{};
+    auto const devAcc = alpaka::getDevByIdx(platform, 0);
 
     // Create a queue on the device
     QueueAcc queue(devAcc);
@@ -108,7 +109,8 @@ auto main() -> int
 
     // Get the host device for allocating memory on the host.
     using DevHost = alpaka::DevCpu;
-    auto const devHost = alpaka::getDevByIdx<DevHost>(0u);
+    auto const platformHost = alpaka::PltfCpu{};
+    auto const devHost = alpaka::getDevByIdx(platformHost, 0);
 
     // Allocate 3 host memory buffers
     using BufHost = alpaka::Buf<DevHost, Data, Dim, Idx>;
--- a/include/alpaka/mem/view/ViewStdArray.hpp
+++ b/include/alpaka/mem/view/ViewStdArray.hpp
@@ -29,7 +29,9 @@ namespace alpaka::trait
     {
         ALPAKA_FN_HOST static auto getDev(std::array<TElem, Tsize> const& /* view */) -> DevCpu
         {
-            return getDevByIdx<PltfCpu>(0u);
+            // Instantiating the CPU platform here is a hack we can do internally, because we know that the CPU
+            // platform does not contain any data. But it generally does not apply.
+            return getDevByIdx(PltfCpu{}, 0u);
         }
     };
 
--- a/include/alpaka/pltf/PltfGenericSycl.hpp
+++ b/include/alpaka/pltf/PltfGenericSycl.hpp
@@ -14,8 +14,6 @@
 #if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
 #    include <iostream>
 #endif
-#include <mutex>
-#include <optional>
 #include <sstream>
 #include <stdexcept>
 #include <vector>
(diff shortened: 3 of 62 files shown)


example/vectorAdd/src/vectorAdd.cpp
```diff
@@ -85,7 +85,8 @@ auto main() -> int
     using QueueAcc = alpaka::Queue<Acc, QueueProperty>;
 
     // Select a device
-    auto const devAcc = alpaka::getDevByIdx<Acc>(0u);
+    auto const platform = alpaka::Pltf<Acc>{};
+    auto const devAcc = alpaka::getDevByIdx(platform, 0);
```
bernhardmgruber, line 89: We can no longer get the nth device of an accelerator, because the creation of a platform object is now necessary.

example/vectorAdd/src/vectorAdd.cpp
```diff
@@ -108,7 +109,8 @@ auto main() -> int
 
     // Get the host device for allocating memory on the host.
     using DevHost = alpaka::DevCpu;
-    auto const devHost = alpaka::getDevByIdx<DevHost>(0u);
+    auto const platformHost = alpaka::PltfCpu{};
+    DevHost const devHost = alpaka::getDevByIdx(platformHost, 0);
```
bernhardmgruber: Same for the host platform, it has to be instantiated.

I wonder whether we should have an alias `HostPlatform` and `HostDevice` inside alpaka. Or potentially even a global object for the host platform. The latter however would not allow us again to cleanly shutdown everything object in alpaka.

include/alpaka/mem/view/ViewStdArray.hpp
```diff
@@ -29,7 +29,7 @@ namespace alpaka::trait
     {
         ALPAKA_FN_HOST static auto getDev(std::array<TElem, Tsize> const& /* view */) -> DevCpu
         {
-            return getDevByIdx<PltfCpu>(0u);
+            return getDevByIdx(PltfCpu{}, 0u);
         }
```
bernhardmgruber: This is a hack we can do internally, because we know that the CPU platform does not contain any data. But it generally does not apply.

j-stephan: Shouldn't this be a C++ comment?

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -28,150 +28,95 @@ namespace alpaka
...
-        }
-
-#    if BOOST_COMP_CLANG
-#        pragma clang diagnostic push
-#        pragma clang diagnostic ignored "-Wexit-time-destructors"
-#    endif
-        inline static std::mutex mutex;
-        inline static bool initialized{false};
-
-        inline static std::optional<sycl::platform> platform_opt{std::nullopt};
```
j-stephan, line 123: Since we're removing `std::optional` we can also remove the corresponding header, no?

include/alpaka/pltf/PltfGenericSycl.hpp
```diff
@@ -28,150 +28,95 @@ namespace alpaka
...
-                }};
-
-            initialized = true;
-        }
-
-#    if BOOST_COMP_CLANG
-#        pragma clang diagnostic push
-#        pragma clang diagnostic ignored "-Wexit-time-destructors"
-#    endif
-        inline static std::mutex mutex;
```
j-stephan, line 120: A platform object is not thread-safe with these changes. Is this intended?

j-stephan (changes requested):
Looks mostly good, two remarks:

include/alpaka/mem/view/ViewStdArray.hpp
bernhardmgruber: Possibily.

include/alpaka/pltf/PltfGenericSycl.hpp
bernhardmgruber: Right.

include/alpaka/pltf/PltfGenericSycl.hpp
bernhardmgruber: I think I provide the same guarantees as the implementation before. Where do you see a possible data race that was not possible before?

j-stephan (approved):

PR #1989, opened 2023-06-26
fix host callback unit test

psychocoderHPC:
bug introduced with #1266

Unit tests are written so that the status of a queue is well-known and should not return false negatives.
This PR reverts the changes from #1266. If we observe still issues with HIP then something is wrong in the callback implementation.

### DIFF
--- a/test/unit/queue/src/QueueTest.cpp
+++ b/test/unit/queue/src/QueueTest.cpp
@@ -25,28 +25,6 @@ using TestQueues = alpaka::meta::Concatenate<
 #endif
     >;
 
-//! Equivalent to CHECK but accounts for potential false negatives
-//!
-//! This is required when checking if an event or a queue is finished/empty.
-//! \warning This macro can be used on host only!
-//!
-//! \param count number of negative checks to be repeated
-//! \param msWait milli seconds to wait if cmd is returning false
-//! \param cmd command executed
-#define LOOPED_CHECK(count, msWait, cmd)                                                                              \
-    do                                                                                                                \
-    {                                                                                                                 \
-        bool ret = false;                                                                                             \
-        for(int i = 0; i < count; ++i)                                                                                \
-        {                                                                                                             \
-            ret = (cmd);                                                                                              \
-            if(ret)                                                                                                   \
-                break;                                                                                                \
-            std::this_thread::sleep_for(std::chrono::milliseconds(msWait));                                           \
-        }                                                                                                             \
-        CHECK(ret);                                                                                                   \
-    } while(0)
-
 TEMPLATE_LIST_TEST_CASE("queueIsInitiallyEmpty", "[queue]", TestQueues)
 {
     using DevQueue = TestType;
@@ -66,7 +44,7 @@ TEMPLATE_LIST_TEST_CASE("queueCallbackIsWorking", "[queue]", TestQueues)
 
     alpaka::enqueue(f.m_queue, [&]() { promise.set_value(true); });
 
-    LOOPED_CHECK(30, 100, promise.get_future().get());
+    CHECK(promise.get_future().get());
 }
 
 TEMPLATE_LIST_TEST_CASE("queueWaitShouldWork", "[queue]", TestQueues)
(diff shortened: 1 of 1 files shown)


test/unit/queue/src/QueueTest.cpp
```diff
@@ -65,7 +43,13 @@ TEMPLATE_LIST_TEST_CASE("queueCallbackIsWorking", "[queue]", TestQueues)
 
     alpaka::enqueue(f.m_queue, [&]() { promise.set_value(true); });
 
-    LOOPED_CHECK(30, 100, promise.get_future().get());
+    // A non-blocking queue will always stay empty because the task has been executed immediately.
```
bernhardmgruber: I disagree. A non-blocking queue will be non-empty until the background thread has finished executing the task.

test/unit/queue/src/QueueTest.cpp
```diff
@@ -65,7 +43,13 @@ TEMPLATE_LIST_TEST_CASE("queueCallbackIsWorking", "[queue]", TestQueues)
...
+    if(!alpaka::test::IsBlockingQueue<typename Fixture::Queue>::value)
+    {
+        alpaka::wait(f.m_queue);
+    }
+
+    CHECK(promise.get_future().get());
```
bernhardmgruber: No wait should be necessary, since the `std::future::get()` blocks until the promise is satisfied, and thus synchronizes with the task.

test/unit/queue/src/QueueTest.cpp
```diff
@@ -101,8 +85,11 @@ TEMPLATE_LIST_TEST_CASE(
         f.m_queue,
         [&f, &callbackFinished]() noexcept
         {
-            LOOPED_CHECK(30, 100, !alpaka::empty(f.m_queue));
+            // This callback function is in the queue therefore the queue can not be empty.
+            CHECK(!alpaka::empty(f.m_queue));
             std::this_thread::sleep_for(std::chrono::milliseconds(100u));
+            // Check again, the queue must stay not empty.
+            CHECK(!alpaka::empty(f.m_queue));
```
bernhardmgruber, line 87: So this is just a simplification of the test without a change in behavior?

test/unit/queue/src/QueueTest.cpp
```diff
@@ -113,7 +100,10 @@ TEMPLATE_LIST_TEST_CASE(
     }
 
     CHECK(callbackFinished.load() == true);
-    LOOPED_CHECK(30, 100, alpaka::empty(f.m_queue));
+
+    // The blocking queue must be empty because we wait for the non-blocking queue and the blocking is synchronized
+    // implicit.
```
bernhardmgruber:
```suggestion
    // The blocking queue must be empty because we wait for the non-blocking queue and the blocking one is synchronized
    // implicitly.
```

bernhardmgruber (commented):
I don't see what this PR fixes. I could not spot the bug that was present before.

test/unit/queue/src/QueueTest.cpp
psychocoderHPC: The background thread is an alpaka workaround which should not affect the behavior of `enqueue`. In a blocking queue, we need the guarantee that if no other thread is running the queue finished all tasks before.

test/unit/queue/src/QueueTest.cpp
psychocoderHPC: yes you are right

test/unit/queue/src/QueueTest.cpp
psychocoderHPC: No this is extending the test, we need to check that the queue stays non-empty as long as the task within the callback queue is executed.

test/unit/queue/src/QueueTest.cpp
bernhardmgruber: The same goes for CUDA. You enqueue a host func into the CUDA stream. Is the CUDA stream empty inside the host func? IMO no.

test/unit/queue/src/QueueTest.cpp
bernhardmgruber: The previous test did:
```
CHECK(!empty(queue));
wait 100ms
CHECK(!empty(queue));
wait 100ms
CHECK(!empty(queue));
wait 100ms
CHECK(!empty(queue));
wait 100ms
... // 30 times
```
The new test does only:
```
CHECK(!empty(queue));
wait 100ms
CHECK(!empty(queue));
```
IMO this is a simplification. What additional behavior is your new version covering?

psychocoderHPC: The previous implementation was only repeating the command in the case where the first check returned that the queue is empty but we are in the host function which is blocking the queue therefore it must be `!empty` else something is wrong.
The second check validates CUDA/HIP or our bookkeeping with the callback thread that if a task which is blocking the execution on the host is really blocking the queue too.

bernhardmgruber: Oh yes, you are completely right! Sorry, my mistake! So yes, your proposed change is better than what we had before!

bernhardmgruber (approved):

psychocoderHPC:
@bernhard PR is ready to be merged.

PR #1990, opened 2023-06-26
fix callback thread and unit tests

psychocoderHPC:
I did some changes on top of #1870 which workaround all thread sanitizer errors for the callback tests.
To validate these in the CI I opened this PR.

- [ ] includes #1989


psychocoderHPC:
This PR can be closed do to the refactoring of our queues there was only one left over data race which should be solved with #2011

PR #2002, opened 2023-07-11
Skip zero dimensional accelerators (continued)

fwyzard:
Follow up to #1979:
  - make `NonZeroTestDims` available along with `TestDims`
  - fix the `mapIdxPitchBytes` test


fwyzard:
The `queueTest` seems to be failing fairly consistently in all PRs.

bernhardmgruber (approved):

PR #2003, opened 2023-07-11
Update the `bufSlicingTest`

fwyzard:
General changes:
  - implement logical operations on `alpaka::Vec`;
  - update the type expected by `View::operator[]`.

Update `bufSlicingTest`:
  - update the comments and simplify the code;
  - add a second test for the memset functionality on a slice.


fwyzard:
(rebased)

fwyzard:
OK, now the only failures are with the windows cl 2022 debug builds.

Unfortunately, I'm utterly unfamiliar with Visual Studio and its error messages.
My interpretation is that now part of `test/unit/mem/copy/src/BufSlicing.cpp` are identified as "unreachable" (warning C4702) and the warning is being treated as an error (error C2220).
Is that the case ?

include/alpaka/vec/Vec.hpp
```diff
@@ -404,14 +445,52 @@ namespace alpaka
...
+            if(TDim::value > 0)
+#else
+            if constexpr(TDim::value > 0)
+#endif
+            {
+                for(typename TDim::value_type i = 0; i < TDim::value; ++i)
+                    r[i] = p[i] && q[i];
+            }
+            return r;
+        }
```
bernhardmgruber, line 475: I was thinking about two things here:

* Either this functionality is too specialized, because it only works for element types convertible to `bool` and even then it might not always make sense, like for `Vec<..., float>` etc. So it should maybe be a function outside the class.
* Or the feature is too restricted, because it assumes the result of `&&` between two element types is `bool`, and we should use something like `Vec<TDim, decltype(std::declval<TVal>() && std::declval<TVal>())>`.

I concluded however, that we can just keep it as is and refactor it when the time comes.

include/alpaka/vec/Vec.hpp
```diff
@@ -404,14 +445,52 @@ namespace alpaka
...
+        //! \return The element-wise logical or relation of two vectors.
+        ALPAKA_NO_HOST_ACC_WARNING
+        ALPAKA_FN_HOST_ACC friend constexpr auto operator||(Vec const& p, Vec const& q) -> Vec<TDim, bool>
+        {
+            Vec<TDim, bool> r;
+#if BOOST_COMP_NVCC && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 3, 0)
+            if(TDim::value > 0)
+#else
+            if constexpr(TDim::value > 0)
+#endif
```
bernhardmgruber, line 486: In my experience, nvcc 11.2 and 11.3 have several terrible bugs around `constexpr`, so maybe we should start considering dropping support.

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -46,42 +45,37 @@ struct TestContainer
...
 
-
     auto createHostBuffer(Vec extents, bool indexed) -> BufHost
     {
         BufHost bufHost(alpaka::allocBuf<TData, TIdx>(devHost, extents));
         if(indexed)
         {
-            TData* const ptr = alpaka::getPtrNative(bufHost);
-            for(TIdx i(0); i < extents.prod(); ++i)
+            TData* const ptr = bufHost.data();
```
bernhardmgruber, line 53: Why do you need to replace `getPtrNative` by `.data()`?

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -111,57 +112,113 @@ using DataTypes = std::tuple<int, float, double>;
...
+        {
+            auto mappedToND = alpaka::mapIdx<Dim::value, Dim1::value>(alpaka::Vec<Dim1, Idx>(i), extentsSubView);
+            auto addedOffset = mappedToND + offsets;
+            auto mappedTo1D = alpaka::mapIdx<Dim1::value>(
+                addedOffset,
+                extents)[0]; // take the only element in the vector
+            ptr[i] = static_cast<Data>(mappedTo1D);
+        }
+
+        // Comapre the resultBuffer with the results computed manually.
```
bernhardmgruber:
```suggestion
        // Compare the resultBuffer with the results computed manually.
```

test/unit/mem/copy/src/BufSlicing.cpp
```diff
@@ -111,57 +112,113 @@ using DataTypes = std::tuple<int, float, double>;
...
+        for(Idx i = 0; i < extents.prod(); ++i)
+        {
+            auto mappedToND = alpaka::mapIdx<Dim::value, Dim1::value>(alpaka::Vec<Dim1, Idx>(i), extents);
+            if((mappedToND >= offsets && mappedToND < offsets + extentsSubView).all())
+                ptr[i] = static_cast<Data>(0u);
+        }
 
-    // resultBuffer will be compared with the manually computed results.
-    slicingTest.compareBuffer(resultBuffer, correctResults, extentsSubView);
+        // Comapre the resultBuffer with the results computed manually.
```
bernhardmgruber:
```suggestion
        // Compare the resultBuffer with the results computed manually.
```

bernhardmgruber (changes requested):

test/unit/mem/copy/src/BufSlicing.cpp
fwyzard: I don't _need_ to, but it's shorter and it follows the approach used in STL containers.

include/alpaka/vec/Vec.hpp
fwyzard: In my local tests, 11.1 was enough to pass these tests, but I picked 11.3 for consistency with the other checks.

That said, from the CMS point of view we are using 11.5 in production, so I wouldn't mind dropping up to 11.3.

bernhardmgruber: That's good to know, we should discuss that in a future VC.

test/unit/mem/copy/src/BufSlicing.cpp
fwyzard: Fixed.

test/unit/mem/copy/src/BufSlicing.cpp
fwyzard: Fixed.

include/alpaka/vec/Vec.hpp
fwyzard: > * it only works for element types convertible to `bool` and even then it might not always make sense, like for `Vec<..., float>` etc

I think it works for types `T` for which, given `T t1, t2`, `t1 && t2` is defined and the result is convertible to `bool`.

I wouldn't mind restricting it to those types (with a `static_assert` or some SFINAE magic, since we do not have concepts in c++17) - but then we should do it for all the other operations, in case `*` does not work as expected, etc.

So... yeah, I agree with you, let's keep it like this and improve it or refactor it later, if needed.

bernhardmgruber (approved):

PR #1984, opened 2023-06-25
Add RNG support for the SYCL back-end

j-stephan:
I checked out the RNG-related changes files from #1845 and put them into an own commit.

Please merge #1982 first. Fixes #1605.

Thanks to @AuroraPerego, @Parsifal-2045 and @fwyzard for their good work.

@Parsifal-2045 Your e-mail address in the original commit(s) is not your CERN e-mail. Should it stay this way or do you want me to change it to the CERN e-mail for the RNG commit?

PR #2006, opened 2023-07-21
Silence clang-16 warnings about unsafe buffer usage.

j-stephan:
As requested by @bernhardmgruber in #1984.

### DIFF
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -244,10 +244,10 @@ namespace alpaka
                 using Allocator = AllocCpuAligned<std::integral_constant<std::size_t, alignment>>;
                 static_assert(std::is_empty_v<Allocator>, "AllocCpuAligned is expected to be stateless");
                 auto* memPtr = alpaka::malloc<TElem>(Allocator{}, static_cast<std::size_t>(getExtentProduct(extent)));
-                auto deleter = [queue = std::move(queue)](TElem* ptr) mutable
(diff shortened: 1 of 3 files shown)


include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -244,10 +244,10 @@ namespace alpaka
                 using Allocator = AllocCpuAligned<std::integral_constant<std::size_t, alignment>>;
                 static_assert(std::is_empty_v<Allocator>, "AllocCpuAligned is expected to be stateless");
                 auto* memPtr = alpaka::malloc<TElem>(Allocator{}, static_cast<std::size_t>(getExtentProduct(extent)));
-                auto deleter = [queue = std::move(queue)](TElem* ptr) mutable
+                auto deleter = [l_queue = std::move(queue)](TElem* ptr) mutable
```
j-stephan, line 247: `icpx` complains about lambda `queue` shadowing function `queue` here, hence the change.

fwyzard:
For me, this is one of the cases where the compiler is over-reaching and
warning about a construct that the programmer is using explicitly and with
a clearly intended purpose.

If this generates a warning, shouldn't all lambda captures by value
generate the same warning?
The captured value is shadowing the variable of the same name.

In fact, in this specific case I would argue that there is no shadowing:
`queue` is available _only_ through the explicit capture, so there is no
shadowing.

bernhardmgruber (approved):

bernhardmgruber:
Thx a lot!

PR #2008, opened 2023-07-22
Remove the requirement that the native handle is an `int`

fwyzard:
The SYCL backend does not use `int` as the native handle type.

### DIFF
--- a/test/unit/traits/src/NativeHandleTest.cpp
+++ b/test/unit/traits/src/NativeHandleTest.cpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Antonio Di Pilato, Jan Stephan
+/* Copyright 2023 Antonio Di Pilato, Jan Stephan, Andrea Bocci
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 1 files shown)


fwyzard:
The `queueTest` seems to be failing fairly consistently in all PRs.

test/unit/traits/src/NativeHandleTest.cpp
```diff
@@ -17,5 +17,7 @@ TEMPLATE_LIST_TEST_CASE("NativeHandle", "[handle]", alpaka::test::TestAccs)
     auto handle = alpaka::getNativeHandle(dev);
 
     STATIC_REQUIRE(std::is_same_v<alpaka::NativeHandle<Dev>, decltype(handle)>);
-    STATIC_REQUIRE(std::is_same_v<alpaka::NativeHandle<Dev>, int>); // It won't work for SYCL backend
+    /* The SYCL backend does not use an int as the native handle type
+    STATIC_REQUIRE(std::is_same_v<alpaka::NativeHandle<Dev>, int>);
+    */
```
j-stephan: This is different from the version in #1845. Is this intended?

fwyzard: I prefer this approach, because either a requirement is there or it isn't.

But I'm fine with either one (also because I expect we will review the requirement after #1988 is merged), let me know what you prefer.

j-stephan: Okay, then I'd prefer aligning with #1845. One less file to care about in that mega PR :-)

fwyzard: Done.

fwyzard: (and rebased)

bernhardmgruber (approved):

PR #2011, opened 2023-07-24
fix callback thread task lifetime

psychocoderHPC:
fix:  https://github.com/alpaka-group/alpaka/issues/2007 

Rewrite the callback thread to have a well-defined lifetime of the enqueued tasks and the moment when the future will be set after the tasks are executed.
The lifetime issue of data captured by a lambda introduced with https://github.com/alpaka-group/alpaka/pull/1870
is solved too and explicit destroying data is not required anymore.

- [x] depends on #2022

PR #2013, opened 2023-07-24
Some refactoring around QueueCpuOmp2Collective

bernhardmgruber:
Just a bunch of simplifications while studying the code.

### DIFF
--- a/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
+++ b/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
@@ -38,30 +38,35 @@ namespace alpaka
 #        pragma clang diagnostic ignored "-Wweak-vtables"
 #    endif
         //! The CPU collective device queue implementation.
-        class QueueCpuOmp2CollectiveImpl final : public cpu::ICpuQueue
+        struct QueueCpuOmp2CollectiveImpl final : cpu::ICpuQueue
 #    if BOOST_COMP_CLANG
 #        pragma clang diagnostic pop
 #    endif
         {
-        public:
-            QueueCpuOmp2CollectiveImpl(DevCpu const& dev) noexcept : m_dev(dev), m_uCurrentlyExecutingTask(0u)
+            explicit QueueCpuOmp2CollectiveImpl(DevCpu const& dev) noexcept : m_dev(dev), blockingQueue(dev)
             {
             }
-            QueueCpuOmp2CollectiveImpl(QueueCpuOmp2CollectiveImpl const&) = delete;
-            auto operator=(QueueCpuOmp2CollectiveImpl const&) -> QueueCpuOmp2CollectiveImpl& = delete;
+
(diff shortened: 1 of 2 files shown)


include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -38,30 +39,35 @@ namespace alpaka
...
             void wait(EventCpu const& ev) final
             {
                 alpaka::wait(*this, ev);
             }
 
-        public:
+            void busyWaitUntilBlockingQueueEmpty()
+            {
+                while(!empty(blockingQueue))
+                    std::this_thread::yield();
```
j-stephan: Is this safe in terms of OpenMP legalese?

bernhardmgruber: I honestly have no idea :D I know back when C++11 was new, people said to not mix `std::thread`s and OpenMP. I don't know whether that is still true. Yielding should be less probablematic I guess. I am even assuming it should not pose a problem at all, since a yield is usually just a hint to processor or OS to reschedule the thread, but otherwise has no side effect.

psychocoderHPC:
Please wait with this PR for #2011 because it is very hard if we had refactorings any time I work on fixing the queue issue.

j-stephan:
Since #2011 is merged: Do you want to rebase for another CI run or is this good to go?

include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
psychocoderHPC: Yes please do not assume that OpenMP is using standard threads. So I suggest not using `std::this_thread::yield()`

psychocoderHPC:
@bernhardmgruber This PR can be rebased #2011 is already merged.

include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
bernhardmgruber: I replaced it by a busy loop now, as in the original implementation.

j-stephan (changes requested):
The code looks good to me. But can you please squash the commits into one and give them a better description?

include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -18,6 +18,7 @@
 
 #include <atomic>
 #include <mutex>
+#include <thread>
```
j-stephan:
```suggestion
```

I don't think we still need this.

j-stephan (changes requested):

include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
bernhardmgruber: Removed.

j-stephan (approved):

PR #2014, opened 2023-07-24
Change separable compilation test from `double` to `float`

fwyzard:
Using `double` triggers an error with the SYCL backend on some targets where double precision is not supported by the hardware and is emulated in software.

The problems seems to be independent from the use of separable compilation.


fwyzard:
Using `float` sometimes gives a result that is off by 10^-6, while `epsilon` is order or 10^-7.

~Should I just bump the tolerance by a factor 10 ?~

This is fixed by forcing the use of FMA inside the loop.

bernhardmgruber (approved):

PR #2015, opened 2023-07-25
Add the fused multiply-add math functions

fwyzard:


bernhardmgruber (commented):
LGTM in principle, but could you please also call the API function from the unit tests. Then I am sure the compiler also agrees with me :) Thx!

fwyzard:
Sure :-)

Not from the separate compilation test, because making that templated on the accelerator seems quite complicated, buy I can add a new test.

Do we already have generic tests for the math functions ?

bernhardmgruber:
> Do we already have generic tests for the math functions ?

We have somewhat generic tests only for unary and binary math functions. So I guess you could add a new test file similar to what we do for `sincos`: https://github.com/alpaka-group/alpaka/blob/develop/test/unit/math/src/sincos.cpp.

fwyzard:
I've added generic tests for ternary ops, and used it for fma on float and double.
I do not think fma supports complex numbers.

psychocoderHPC (approved):

psychocoderHPC:
The failing test is the queue test which is fixed with #2011, therefore I will merge it aven CI has not passed all tests.

PR #2021, opened 2023-07-26
Fix the delagating constructor in KernelExecutionFixture

fwyzard:
Avoid using data members before they have been initialised.

### DIFF
--- a/include/alpaka/test/KernelExecutionFixture.hpp
+++ b/include/alpaka/test/KernelExecutionFixture.hpp
@@ -29,28 +29,30 @@ namespace alpaka::test
         using Acc = TAcc;
         using Dim = alpaka::Dim<Acc>;
         using Idx = alpaka::Idx<Acc>;
-        using PlatformAcc = Pltf<Acc>;
-        using DevAcc = Dev<Acc>;
-        using PltfAcc = Pltf<DevAcc>;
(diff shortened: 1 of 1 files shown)


include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -29,36 +29,38 @@ namespace alpaka::test
...
-                GridBlockExtentSubDivRestrictions::Unrestricted))
+            : m_devHost{getDevByIdx(m_platformHost, 0)}
+            , m_device{getDevByIdx(m_platform, 0)}
+            , m_queue{m_device}
+            , m_workDiv{getValidWorkDiv<Acc>(
+                  m_device,
+                  extent,
+                  Vec<Dim, Idx>::ones(),
+                  false,
+                  GridBlockExtentSubDivRestrictions::Unrestricted)}
```
bernhardmgruber, line 55: This is the bugfix. `: KernelExecutionFixture(getValidWorkDiv<Acc>(getDevByIdx(m_platformAcc, 0), ...), ...)` uses `m_platformAcc` but the constructor of the object has not run yet. And this only occured in the SYCL backend, because it is the only stateful backend. Thanks for spotting this!

bernhardmgruber (approved):

PR #2022, opened 2023-07-26
queue test: fix catch2 usages within threads

psychocoderHPC:
Catch2 is not threadsafe, therefore using the catch macros is not allowed in threads.
Use atomics to propagate checks to the main thread.


bernhardmgruber (approved):

psychocoderHPC:
The CI error in the queue test should be ignored because this should be solved with #2011

PR #2023, opened 2023-07-26
fix test CollectiveQueue

psychocoderHPC:
Better documentation of the QueueCpuOmp2Collective to reflect the special synchronization behaviours.
Fix the test CollectiveQueue, the data race was mentioned in the comments only handeled via a sleep which was shown as warning in the thread sanitizer.

### DIFF
--- a/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
+++ b/include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
@@ -73,9 +73,12 @@ namespace alpaka
     // to the user workflows.
     //
     // Within a OpenMP parallel region kernel will be performed collectively.
-    // All other operations will be performed from one thread (it is not defined which thread).
+    // All other operations will be performed from one thread (it is not defined which thread) and there will be no
+    // implicit synchronization between other operations within the parallel OpenMP parallel region. Operations
+    // executed within a OpenMP parallel region will be executed after already queued tasks before the parallel region
+    // was created.
     //
-    // Outside of a OpenMP parallel region the queue behaves like QueueCpuBlocking.
+    // Outside of an OpenMP parallel region the queue behaves like QueueCpuBlocking.
     class QueueCpuOmp2Collective final
         : public concepts::Implements<ConceptCurrentThreadWaitFor, QueueCpuOmp2Collective>
     {
--- a/test/unit/queue/src/CollectiveQueue.cpp
+++ b/test/unit/queue/src/CollectiveQueue.cpp
@@ -106,9 +106,8 @@ TEST_CASE("TestCollectiveMemcpy", "[queue]")
         auto dst = alpaka::createView(dev, results.data() + threadId, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));
         auto src = alpaka::createView(dev, &threadId, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));
 
-        // avoid that the first thread is executing the copy (can not be guaranteed)
-        size_t sleep_ms = (results.size() - static_cast<uint32_t>(threadId)) * 100u;
-        std::this_thread::sleep_for(std::chrono::milliseconds(sleep_ms));
+        // avoid that the first thread is executing the memcpy before all other threads wrote their changes
+        alpaka::wait(queue);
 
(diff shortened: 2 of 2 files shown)


include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
```diff
@@ -73,7 +73,10 @@ namespace alpaka
     // to the user workflows.
     //
     // Within a OpenMP parallel region kernel will be performed collectively.
-    // All other operations will be performed from one thread (it is not defined which thread).
+    // All other operations will be performed from one thread (it is not defined which thread) and there will be no
+    // implicit synchronization between other operations within the parallel OpenMP parallel region. Operations
+    // executed within a OpenMP parallel region will be executed after already queued tasks before the parallel region
```
bernhardmgruber, line 78:
```suggestion
    // executed within an OpenMP parallel region will be executed after already queued tasks before the parallel region
```

test/unit/queue/src/CollectiveQueue.cpp
```diff
@@ -106,9 +106,8 @@ TEST_CASE("TestCollectiveMemcpy", "[queue]")
         auto dst = alpaka::createView(dev, results.data() + threadId, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));
         auto src = alpaka::createView(dev, &threadId, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));
 
-        // avoid that the first thread is executing the copy (can not be guaranteed)
-        size_t sleep_ms = (results.size() - static_cast<uint32_t>(threadId)) * 100u;
-        std::this_thread::sleep_for(std::chrono::milliseconds(sleep_ms));
+        // avoid that the first thread is executing the memcopy before all other threads wrote there changes
```
bernhardmgruber:
```suggestion
        // avoid that the first thread is executing the memcpy before all other threads wrote their changes
```
I always make the first typo as well ;D

test/unit/queue/src/CollectiveQueue.cpp
```diff
@@ -106,9 +106,8 @@ TEST_CASE("TestCollectiveMemcpy", "[queue]")
...
+        alpaka::wait(queue);
```
bernhardmgruber: I don't understand why the wait is necessary. The sleep before is optional, only to try to make another thread as the main thread run the memcpy.

psychocoderHPC: Wait will guarantee that all threads reached the point after line 107. The old sleep was hoping that we have no data-race by waiting long enough.

test/unit/queue/src/CollectiveQueue.cpp
psychocoderHPC: fixed

include/alpaka/test/queue/QueueCpuOmp2Collective.hpp
psychocoderHPC: fixed

test/unit/queue/src/CollectiveQueue.cpp
bernhardmgruber: But this is not necessary, the following program is fine too:
```c++
#    pragma omp parallel num_threads(static_cast <int>(results.size()))
    {
        int threadId = omp_get_thread_num();
        auto dst = alpaka::createView(dev, results.data() + threadId, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));
        auto src = alpaka::createView(dev, &threadId, Vec(static_cast<Idx>(1u)), Vec(sizeof(int)));

        // no sleep and wait

        // only one thread will perform this memcpy
        alpaka::memcpy(queue, dst, src, Vec(static_cast<Idx>(1u)));
        alpaka::wait(queue);
    }
```
Where do you spot a data race that you are trying to prevent?

psychocoderHPC: You are right and it was most likely a false positive in the thread sanitizer.
I will remove the change from this PR and update the documentation of the COllectiveQueue only.

psychocoderHPC:
I opened #2025 to update the documentation to have a better branch name than `fix-...` for a documentation update.

Issue #1906, opened 2023-01-17
Clang 15 as CUDA compiler does not work (at least for the test cases)


SimeonEhrig:
If Clang 15 is set as CUDA compiler, it fails at compiling Catch2.

https://gitlab.com/hzdr/crp/alpaka/-/jobs/3613272972

bernhardmgruber:
clang-16 btw works fine as CUDA compiler. I have it in LLAMA since yesterday.

j-stephan:
Is this still an issue?

j-stephan:
It is. I just noticed that the clang-cuda-15 jobs linked here fail with the same error message as the clang-cuda-16 jobs in #1971. This is an issue in LLVM when compiling in debug mode: https://github.com/llvm/llvm-project/issues/58491

SimeonEhrig:
Clang does not provide patch releases for older major releases, than the current release. If the bug will be fixed, it will requires at least the upcoming release of  Clang 17. Because our CI filter rule does not disable the debug build for up coming major releases, it will automatically test the debug build, if we add a new major version.

j-stephan:
FYI, this has finally been fixed in LLVM: https://github.com/llvm/llvm-project/issues/58491#event-14791915581

PR #2027, opened 2023-07-27
Enable Clang-Cuda 15 for Release Builds in the CI

SimeonEhrig:
~Since we deactivate Clang-Cuda (see #1906 ), we updated Catch2. Therefore let's test it again.~

Because of a compiler bug, only builds in debug does not work. Enable Clang-Cuda 15 release builds.


j-stephan (approved):

README.md
```diff
@@ -65,19 +65,21 @@ Supported Compilers
...
+| [HIP](https://alpaka.readthedocs.io/en/latest/install/HIP.html) (clang)        | -                                           | -                                           | -                                           | :x:                    | :x:                                                    | :x:                                                     | :x:                                              | :x:                                         | :white_check_mark: (HIP 5.0 - 5.2)                | :white_check_mark: (HIP 5.3 - 5.4)     | :white_check_mark: (HIP 5.5) | :x:                | -                                                     | -                                    |
 
 Other compilers or combinations marked with :x: in the table above may work but are not tested in CI and are therefore not explicitly supported.
 
+\* due to a [compiler bug](https://github.com/llvm/llvm-project/issues/58491) in debug mode only release builds are supported
```
j-stephan, line 81: The `*` is easily overlooked. Maybe we can add `:heavy_exclamation_mark:` (renders as :heavy_exclamation_mark:) instead? Both here and in the matrix.

j-stephan (changes requested):

psychocoderHPC (approved):

README.md
j-stephan: I can also do this in the clang-16 PR so we don't have to rerun the tests.

j-stephan (approved):

README.md
SimeonEhrig: I would prefer a classic foot notes. Is `[1]` a option?

j-stephan: Yes, footnotes would also work.

bernhardmgruber (approved):

PR #2029, opened 2023-07-27
add math::log2 and math::log10

### LINKED ISSUE #2012: No Log10 Math Function?
We have a NN in our code that uses the log10 CUDA function, but it doesn't seem like Alpaka has one implemented natively. Right now I'm using the change of base formula and natural log function in Alpaka to get around this, but I was hoping that Alpaka could add native support for this math function when you get a chance. Thank you.

psychocoderHPC:
fix #2012

Implement supportfor `log2` and `log10` and the coresponding CI tests.

### DIFF
--- a/include/alpaka/math/MathGenericSycl.hpp
+++ b/include/alpaka/math/MathGenericSycl.hpp
@@ -127,6 +127,16 @@ namespace alpaka::math
     {
     };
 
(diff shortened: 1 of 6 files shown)


psychocoderHPC:
The failing CI test is the queueTest.

j-stephan (approved):

PR #2031, opened 2023-07-27
Update SYCL README

AuroraPerego:
Update the SYCL README after the rewriting of the SYCL backend.

The main updates concern:
- drop of the `sycl::stream`
- new targets passed to `-fsycl-targets`
- info on the warp size

### DIFF
--- a/README_SYCL.md
+++ b/README_SYCL.md
@@ -13,16 +13,14 @@ At the moment alpaka's SYCL back-end can only be used together with Intel oneAPI
 ### General CMake options
 
 * `alpaka_ACC_SYCL_ENABLE`: set to `ON` to enable the SYCL back-end. Requires the activation of at least one oneAPI hardware target (see below).
-* `alpaka_SYCL_ENABLE_IOSTREAM`: set to `ON` to enable device-side printing. Force-enabled if `BUILD_TESTING` is enabled.
-* `alpaka_SYCL_IOSTREAM_KIB`: Kibibytes per block reserved as output buffer for device-side printing. This cannot exceed the amount of shared memory per block. Only takes effect if `alpaka_SYCL_ENABLE_IOSTREAM` is enabled. Default: `64`.
 * `MKL_ROOT`: always required. Set to the root of your oneMKL installation. Example: `/opt/intel/oneapi/mkl/2023.1.0`.
 * `oneDPL_DIR`: always required. Set to the CMake path of your oneDPL installation. Example: `/opt/intel/oneapi/dpl/2022.1.0/lib/cmake/oneDPL`.
 
-### Building for Intel CPUs
+### Building for x86 64-bit CPUs
 
-The following CMake flags can be set for Intel CPUs:
+The following CMake flags can be set for CPUs:
 
-* `alpaka_SYCL_ONEAPI_CPU`: set to `ON` to enable compilation for Intel CPUs.
+* `alpaka_SYCL_ONEAPI_CPU`: set to `ON` to enable compilation for CPUs. Relies on the Intel OpenCL CPU runtime and ahead-of-time compiler, which support Intel and AMD CPUs.
 * `alpaka_SYCL_ONEAPI_CPU_ISA`: the Intel ISA to compile for. Look at the possible `--march` options listed in the output of `opencl-aot --help`. Default: `avx2`.
 
 ### Building for Intel FPGAs
@@ -37,7 +35,8 @@ Note: Intel FPGAs cannot be targeted together with other Intel hardware. This is
 ### Building for Intel GPUs
 
 * `alpaka_SYCL_ONEAPI_GPU`: set to `ON` to enable compilation for Intel GPUs.
-* `alpaka_SYCL_ONEAPI_GPU_DEVICES`: semicolon-separated list of one or more Intel GPUs to compile for. Check the output of `ocloc compile --help` and look at the possible values for the `-device` argument for valid values to supply here. Default: `bdw`.
+* `alpaka_SYCL_ONEAPI_GPU_DEVICES`: semicolon-separated list of one or more Intel GPUs to compile for. The possible values for the devices are listed in the [UsersManual](https://intel.github.io/llvm-docs/UsersManual.html#generic-options) under the flag `-fsycl-targets`. Default: `intel_gpu_pvc`.
+  Note: currently only one target at a time can be specified (limitation of the Intel Compiler)
 
 ## Standalone mode
 
@@ -48,10 +47,7 @@ Using the SYCL back-end always requires the following flags:
 * `-fsycl` (compiler and linker)
 * `-fsycl-standard=2020` (compiler)
 
-To enable device-side printing add the following compiler flags:
-
-* `-DALPAKA_SYCL_IOSTREAM_ENABLED`: to enable device-side printing.
-* `-DALPAKA_SYCL_IOSTREAM_KIB=<value>`: `<value>` (without the brackets) defines the kibibytes per block to be reserved for device-side printing. `<value>` cannot exceed the amount of shared memory per block.
+Device-side printing is possible with `printf`, it calls `sycl::ext::oneapi::experimental::printf` that emulates the standard one. This is an extension of the SYCL standard, still in an experimental state, therefore may not always work correctly.
 
 ### Building for x86 64-bit CPUs
 
@@ -75,8 +71,7 @@ To enable device-side printing add the following compiler flags:
 
 1. `#include <alpaka/standalone/GpuSyclIntel.hpp>` in your C++ code.
 2. Add the following flags:
-  * `-fsycl-targets=spir64_gen` (compiler and linker): to enable GPU compilation. Note: If you are using multiple SYCL hardware targets (like CPU and GPU) separate them by comma here.
-  * `-Xsycl-target-backend=spir64_gen "-device <list>"` (linker): to choose the Intel GPU(s) to compile for. Multiple devices can either be separated by comma or by supplying a range of devices. Refer to the output of `ocloc compile --help` and look for the `-device` flag for the possible values.
+  * `-fsycl-targets=intel_gpu_pvc` (compiler and linker): to enable GPU compilation. Note: If you are using multiple SYCL hardware targets (like CPU and GPU) separate them by comma here.
 
 ## Using the SYCL back-end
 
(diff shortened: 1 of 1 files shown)


README_SYCL.md
```diff
@@ -37,7 +35,8 @@ Note: Intel FPGAs cannot be targeted together with other Intel hardware. This is
 ### Building for Intel GPUs
 
 * `alpaka_SYCL_ONEAPI_GPU`: set to `ON` to enable compilation for Intel GPUs.
-* `alpaka_SYCL_ONEAPI_GPU_DEVICES`: semicolon-separated list of one or more Intel GPUs to compile for. Check the output of `ocloc compile --help` and look at the possible values for the `-device` argument for valid values to supply here. Default: `bdw`.
+* `alpaka_SYCL_ONEAPI_GPU_DEVICES`: semicolon-separated list of one or more Intel GPUs to compile for. The possible values for the devices are listed in the [UsersManual](https://intel.github.io/llvm-docs/UsersManual.html#generic-options) under the flag `-fsycl-targets`. Default: `intel_gpu_pvc`.
```
AuroraPerego, line 38: Maybe out of the scope of this PR, but it is not completely clear to me why we are asking for a semicolon-separated list of devices and then change them into comma in `cmake/alpakaCommon.cmake` instead of asking directly for a comma-separated list.

README_SYCL.md
```diff
@@ -13,14 +13,12 @@ At the moment alpaka's SYCL back-end can only be used together with Intel oneAPI
 ### General CMake options
 
 * `alpaka_ACC_SYCL_ENABLE`: set to `ON` to enable the SYCL back-end. Requires the activation of at least one oneAPI hardware target (see below).
-* `alpaka_SYCL_ENABLE_IOSTREAM`: set to `ON` to enable device-side printing. Force-enabled if `BUILD_TESTING` is enabled.
-* `alpaka_SYCL_IOSTREAM_KIB`: Kibibytes per block reserved as output buffer for device-side printing. This cannot exceed the amount of shared memory per block. Only takes effect if `alpaka_SYCL_ENABLE_IOSTREAM` is enabled. Default: `64`.
 * `MKL_ROOT`: always required. Set to the root of your oneMKL installation. Example: `/opt/intel/oneapi/mkl/2023.1.0`.
 * `oneDPL_DIR`: always required. Set to the CMake path of your oneDPL installation. Example: `/opt/intel/oneapi/dpl/2022.1.0/lib/cmake/oneDPL`.
 
-### Building for Intel CPUs
+### Building for CPUs
```
fwyzard:
```suggestion
### Building for x86 64-bit CPUs
```

README_SYCL.md
```diff
@@ -13,14 +13,12 @@ At the moment alpaka's SYCL back-end can only be used together with Intel oneAPI
...
 
-The following CMake flags can be set for Intel CPUs:
+The following CMake flags can be set for CPUs:
 
 * `alpaka_SYCL_ONEAPI_CPU`: set to `ON` to enable compilation for Intel CPUs.
```
fwyzard:
```suggestion
* `alpaka_SYCL_ONEAPI_CPU`: set to `ON` to enable compilation for CPUs. Relies on the Intel OpenCL CPU runtime and ahead-of-time compiler, which support Intel and AMD CPUs.
```

README_SYCL.md
```diff
@@ -102,3 +97,14 @@ These can be used interchangeably (some restrictions apply - see below) with the
   ```
   See [Intel's FAQ](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#feature-double-precision-emulation-fp64) for more information.
 * The FPGA back-end does not support atomics. alpaka will not check this.
+* device global variables are not supported in SYCL yet.
+
+### Choosing the sub-group size (warp size)
+
+There is a trait to specify at compile time the sub-group size to use for a kernel. The default behaviour, when no sub-group size is specified, is to let the back-end compiler pick the preferred size.
```
fwyzard:
```suggestion
Most SYCL targets support multiple sub-group sizes. There is a trait to specify at compile time the sub-group size to use for a kernel. The default behaviour, when no sub-group size is specified, is to let the back-end compiler pick the preferred size.
```

README_SYCL.md
```diff
@@ -102,3 +97,14 @@ These can be used interchangeably (some restrictions apply - see below) with the
   ```
   See [Intel's FAQ](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#feature-double-precision-emulation-fp64) for more information.
 * The FPGA back-end does not support atomics. alpaka will not check this.
+* device global variables are not supported in SYCL yet.
```
fwyzard:
```suggestion
* Device global variables (corresponding to `__device__` and `__constant__` variables in CUDA) are not supported in the SYCL back-end yet.
```

fwyzard:
Can you add a comment under the "limitations" about shared memory not working on the CPU back-end (yet) ?

README_SYCL.md
j-stephan: Convention basically. When CMake expects a list for its native facilities it is semicolon-separated. I wanted this to behave the same way.

README_SYCL.md
```diff
@@ -37,7 +35,8 @@ Note: Intel FPGAs cannot be targeted together with other Intel hardware. This is
 ### Building for Intel GPUs
 
 * `alpaka_SYCL_ONEAPI_GPU`: set to `ON` to enable compilation for Intel GPUs.
-* `alpaka_SYCL_ONEAPI_GPU_DEVICES`: semicolon-separated list of one or more Intel GPUs to compile for. Check the output of `ocloc compile --help` and look at the possible values for the `-device` argument for valid values to supply here. Default: `bdw`.
+* `alpaka_SYCL_ONEAPI_GPU_DEVICES`: semicolon-separated list of one or more Intel GPUs to compile for. The possible values for the devices are listed in the [UsersManual](https://intel.github.io/llvm-docs/UsersManual.html#generic-options) under the flag `-fsycl-targets`. Default: `intel_gpu_pvc`.
+  Note: currently only one target at a time can be specified (limitation of the Intel Compiler)
```
j-stephan, line 39: Hm, this seems like a downgrade from the previous mechanism. IIRC it was possible to specify multiple GPU targets in one pass before the change.

README_SYCL.md
```diff
@@ -102,3 +97,15 @@ These can be used interchangeably (some restrictions apply - see below) with the
   ```
   See [Intel's FAQ](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#feature-double-precision-emulation-fp64) for more information.
 * The FPGA back-end does not support atomics. alpaka will not check this.
+* Device global variables (corresponding to `__device__` and `__constant__` variables in CUDA) are not supported in the SYCL back-end yet.
+* The shared memory on CPU back-end does not work properly yet. The tests that use it (`atomicTest`, `blockSharedTest` and `blockSharedSharingTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`.
```
j-stephan:
```suggestion
* Shared memory does not yet work properly on CPU back-ends. The tests that use it (`atomicTest`, `blockSharedTest` and `blockSharedSharingTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`.
```

fwyzard (dismissed):

README_SYCL.md
```diff
@@ -102,3 +97,15 @@ These can be used interchangeably (some restrictions apply - see below) with the
...
+
+### Choosing the sub-group size (warp size)
+
+Most SYCL targets support multiple sub-group sizes. There is a trait to specify at compile time the sub-group size to use for a kernel. The default behaviour, when no sub-group size is specified, is to let the back-end compiler pick the preferred size.
```
j-stephan: Shouldn't we mention the trait here?

README_SYCL.md
```diff
@@ -102,3 +97,15 @@ These can be used interchangeably (some restrictions apply - see below) with the
...
+
+Before launching a kernel with a compile-time sub-group size the user should query the sizes supported by the device, and choose accordingly. If the device does not support the requested size, the SYCL runtime will throw a synchronous exception.
+
+During just-in-time (JIT) compilation this guarantees that a kernel is compiled only for the sizes supported by the device. During ahead-of-time (AOT) compilation this is not enough, because the device is not known at compile time. The SYCL specification mandates that the back-end compilers should not fail if a kernel uses unsupported features, like unsupported sub-group sizes. Unfortunately the Intel OpenCL CPU and GPU compilers currently fail with a hard error. To work around this limitation, use the preprocessor macros defined when compiling AOT for the new SYCL targets to enable the compilation only for the sub-group sizes supported by each device.
+
+Note: while the CPU OpenCL back-end does support a sub-group size of 64, the SYCL code currently does not. To avoid issues with the sub-group primitives always consider the sub-group size of 64 as not supported by the device.
```
j-stephan:
```suggestion
Note: while the CPU OpenCL back-end supports a sub-group size of 64, Intel's SYCL implementation currently does not. To avoid issues with the sub-group primitives always consider the sub-group size of 64 as not supported by the device.
```

j-stephan (changes requested):

README_SYCL.md
AuroraPerego: yes, but with the old targets the warp size stuff does not work because the device-specific macros are not defined. That's why we moved to the new targets.

README_SYCL.md
fwyzard:
```diff
-To avoid issues with the sub-group primitives always consider the sub-group size of 64 as not supported by the device.
+To avoid issues with the sub-group primitives, alpaka always considers the sub-group size of 64 as not supported by the device.
```

README_SYCL.md
```diff
@@ -102,3 +97,25 @@ These can be used interchangeably (some restrictions apply - see below) with the
...
+
+Most SYCL targets support multiple sub-group sizes. There is a trait to specify at compile time the sub-group size to use for a kernel that can be defined as follow:
+```cpp
+template<std::uint32_t TWarpSize, typename TAcc>
+struct alpaka::trait::WarpSize<MyKernel<TWarpSize>, TAcc>
+    : std::integral_constant<std::uint32_t, TWarpSize>
+{
+};
+```
+`MyKernel` must therefore be templated on the sub-group size.
```
fwyzard:
```suggestion
For example, if `MyKernel` requires a sub-group size of `32`, this can be declared specialising the `alpaka::trait::WarpSize`:
```cpp
struct MyKernel { ... };

template<typename TAcc>
struct alpaka::trait::WarpSize<MyKernel, TAcc>
    : std::integral_constant<std::uint32_t, 32>
{
};
```
This can be extended to kernels that support multiple sub-group sizes at compile time:
```cpp
template<std::uint32_t TWarpSize>
struct MyKernel { ... };

template<std::uint32_t TWarpSize, typename TAcc>
struct alpaka::trait::WarpSize<MyKernel<TWarpSize>, TAcc>
    : std::integral_constant<std::uint32_t, TWarpSize>
{
};
```
```

j-stephan (approved):
Thanks!

README_SYCL.md
fwyzard: From some tests it's not clear if multiple targets are supported or not... still looking into it.

fwyzard: OK, after running some more tests, I think the code _is_ compiled for multiple device of the same backend, but _only the first one_ is actually used at runtime.

PR #2033, opened 2023-07-30
Fix compilation error in allocMappedBuf for SYCL

AuroraPerego:
In the SYCL version of `allocMappedBuf` there is a call to `TPlatform::syclContext()` that has to be changed to `platform.syclContext()` now that the `Platform` is an object.
Fixes the error: `call to non-static member function without an object argument`.

`allocMappedBuf` has been modified to take also the `Platform` as an optional argument, not used for the other back-ends.

### DIFF
--- a/include/alpaka/mem/buf/BufCpu.hpp
+++ b/include/alpaka/mem/buf/BufCpu.hpp
@@ -15,6 +15,7 @@
 #include "alpaka/mem/buf/Traits.hpp"
 #include "alpaka/mem/view/ViewAccessOps.hpp"
 #include "alpaka/meta/DependentFalseType.hpp"
+#include "alpaka/platform/PlatformCpu.hpp"
 #include "alpaka/vec/Vec.hpp"
 
 #include <functional>
@@ -270,8 +271,10 @@ namespace alpaka
         struct BufAllocMapped<PlatformCpu, TElem, TDim, TIdx>
         {
             template<typename TExtent>
-            ALPAKA_FN_HOST static auto allocMappedBuf(DevCpu const& host, TExtent const& extent)
-                -> BufCpu<TElem, TDim, TIdx>
+            ALPAKA_FN_HOST static auto allocMappedBuf(
+                DevCpu const& host,
+                PlatformCpu const& /*platform*/,
+                TExtent const& extent) -> BufCpu<TElem, TDim, TIdx>
             {
                 // Allocate standard host memory.
                 return allocBuf<TElem, TIdx>(host, extent);
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -304,8 +304,10 @@ namespace alpaka
         struct BufAllocMapped<PlatformUniformCudaHipRt<TApi>, TElem, TDim, TIdx>
         {
             template<typename TExtent>
-            ALPAKA_FN_HOST static auto allocMappedBuf(DevCpu const& host, TExtent const& extent)
-                -> BufCpu<TElem, TDim, TIdx>
+            ALPAKA_FN_HOST static auto allocMappedBuf(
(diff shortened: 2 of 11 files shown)


fwyzard:
Thanks for the fix 👍🏻 
I would prefer to keep the `extent` parameter last, so the `platform` should come before it.

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -270,8 +271,10 @@ namespace alpaka
         struct BufAllocMapped<PlatformCpu, TElem, TDim, TIdx>
         {
             template<typename TExtent>
-            ALPAKA_FN_HOST static auto allocMappedBuf(DevCpu const& host, TExtent const& extent)
-                -> BufCpu<TElem, TDim, TIdx>
+            ALPAKA_FN_HOST static auto allocMappedBuf(
+                DevCpu const& host,
+                PlatformCpu /*platform*/,
```
fwyzard:
```suggestion
                PlatformCpu const& /*platform*/,
```

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -304,8 +304,10 @@ namespace alpaka
         struct BufAllocMapped<PlatformUniformCudaHipRt<TApi>, TElem, TDim, TIdx>
         {
             template<typename TExtent>
-            ALPAKA_FN_HOST static auto allocMappedBuf(DevCpu const& host, TExtent const& extent)
-                -> BufCpu<TElem, TDim, TIdx>
+            ALPAKA_FN_HOST static auto allocMappedBuf(
+                DevCpu const& host,
+                PlatformUniformCudaHipRt<TApi> /*platform*/,
```
fwyzard:
```suggestion
                PlatformUniformCudaHipRt<TApi> const& /*platform*/,
```

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -133,9 +133,12 @@ namespace alpaka
     //! \param extent The extent of the buffer.
     //! \return The newly allocated buffer.
     template<typename TPlatform, typename TElem, typename TIdx, typename TExtent>
-    ALPAKA_FN_HOST auto allocMappedBuf(DevCpu const& host, TExtent const& extent = TExtent())
+    ALPAKA_FN_HOST auto allocMappedBuf(
+        DevCpu const& host,
+        TPlatform const& platform = TPlatform(),
```
fwyzard: I do not know if allowing a default value here is a good idea or not.
In the SYCL case, what does the default constructor of the platform do?

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -167,11 +170,14 @@ namespace alpaka
     //! \param extent The extent of the buffer.
     //! \return The newly allocated buffer.
     template<typename TPlatform, typename TElem, typename TIdx, typename TExtent>
-    ALPAKA_FN_HOST auto allocMappedBufIfSupported(DevCpu const& host, TExtent const& extent = TExtent())
+    ALPAKA_FN_HOST auto allocMappedBufIfSupported(
+        DevCpu const& host,
+        TPlatform const& platform = TPlatform(),
```
fwyzard: Same here.

fwyzard (changes requested):

include/alpaka/mem/buf/Traits.hpp
fwyzard: After some more thoughts, I think it would be better to _not_ have the default:
  - the SYCL backend requires a valid platform object
  - code written to be portable should be compatible with all backends

bernhardmgruber: I agree, platforms cannot be relied upon to be default constructible.

fwyzard (approved):

j-stephan (approved):

PR #2036, opened 2023-07-31
Make the vendor random number generators optional

### LINKED ISSUE #1810: Make RNG support optional
I received a user request from @fwyzard today: Make RNG support in alpaka optional. Some back-ends introduce some really heavy (and sometimes buggy) header files so it would be nice to have CMake and preprocessor options to turn off RNG when not required.

fwyzard:
If `alpaka_DISABLE_VENDOR_RNG` is defined, alpaka will not include support for the vendor-specific random number generators (NVIDIA cuRAN, AMD hipRAND/rocRAND, Indel DPL).

Instead, all accelerators use the default configuration of the Philox RNG.

### DIFF
--- a/include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
+++ b/include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
@@ -16,6 +16,7 @@
 #include "alpaka/intrinsic/IntrinsicUniformCudaHipBuiltIn.hpp"
 #include "alpaka/math/MathUniformCudaHipBuiltIn.hpp"
 #include "alpaka/mem/fence/MemFenceUniformCudaHipBuiltIn.hpp"
+#include "alpaka/rand/RandDefault.hpp"
 #include "alpaka/rand/RandUniformCudaHipRand.hpp"
 #include "alpaka/warp/WarpUniformCudaHipBuiltIn.hpp"
 #include "alpaka/workdiv/WorkDivUniformCudaHipBuiltIn.hpp"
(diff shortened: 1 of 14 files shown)


fwyzard:
Now... how do I make it optional with a CMake flag ?

AuroraPerego:
> Now... how do I make it optional with a CMake flag ?

Something like this in `cmake/alpakaCommon.cmake` should work:
```cmake
option(alpaka_DISABLE_RANDOM "Disable the the random number generators" OFF)

# Compiler settings.
if(alpaka_DISABLE_RANDOM)
    target_compile_definitions(alpaka INTERFACE "ALPAKA_DISABLE_RANDOM")
endif()
```
and then the option `-Dalpaka_DISABLE_RANDOM` can be passed to `cmake`.

AuroraPerego:
Maybe also `randTest` should be enclosed in a `#ifndef ALPAKA_DISABLE_RANDOM` or skipped directly when the `cmake` flag is used adding to `test/unit/rand/CMakeLists.txt` something like:
```cmake
if (alpaka_DISABLE_RANDOM STREQUAL "ON")
    message(STATUS "randTest requires the random number generators libraries. Test disabled.")
    return()
endif ()
```

fwyzard:
Looks like including Philox in the CUDA and HIP accelerators leads to a recursive loop of header files inclusion... :-(

include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
```diff
@@ -64,7 +65,12 @@ namespace alpaka
         public BlockSyncUniformCudaHipBuiltIn,
         public IntrinsicUniformCudaHipBuiltIn,
         public MemFenceUniformCudaHipBuiltIn,
-        public rand::RandUniformCudaHipRand<TApi>,
+#    ifdef ALPAKA_DISABLE_VENDOR_RNG
+        public rand::RandDefault,
+#    else
+          public rand::RandUniformCudaHipRand<TApi>
+        ,
```
j-stephan: This looks odd. Did `clang-format` do this?

fwyzard: Yes... 
I changed the other files and moved the comma in front to avoid the ugly code, but I missed this one.

fwyzard: Actually, maybe I did not push that change? I'll check.

fwyzard: I missed a couple files, hopefully they look better now.

j-stephan (approved):

bernhardmgruber:
This PR may break the amalgamation, but I can fix this for you in a follow-up commit.

fwyzard:
> This PR may break the amalgamation, but I can fix this for you in a follow-up commit.

Thanks!

PR #2039, opened 2023-08-01
Remove the dependency on Intel MKL

fwyzard:


j-stephan (dismissed):
Looks good to me. Since we don't have a CI yet I assume the tests are successfully compiled on your machine?

fwyzard:
I think so... let me give it another try :-)

j-stephan (approved):

fwyzard:
Let's delay this, I got a crash while testing that _should_ be unrelated... but I'll need to check.

fwyzard:
Looks like my setup is broken, independently from this PR.
@AuroraPerego confirmed that the change works :-)

PR #2040, opened 2023-08-01
Enable release builds for gcc-9 + CUDA

j-stephan:
This PR re-enables gcc-9+CUDA release builds for compatibility with #1977. The original reason for disabling them seems to have perished in the meantime.


SimeonEhrig:
That's strange that only NVCC 11.0 is tested with GCC 9. I did a quick and found at least one valid combination for NVCC 11.0 up to 12.1 and GCC 9.

SimeonEhrig (approved):

PR #2042, opened 2023-08-01
Fix a dangling pointer in the SYCL memory buffer deleter

fwyzard:
In
```c++
template<typename TExtent>
static auto allocBuf(DevGenericSycl<TPlatform> const& dev, TExtent const& extent)
  -> BufGenericSycl<TElem, TDim, TIdx, TPlatform>
  {
    //...
    auto deleter = [&dev](TElem* ptr) { sycl::free(ptr, dev.getNativeHandle().second); };
    return BufGenericSycl<TElem, TDim, TIdx, TPlatform>(dev, memPtr, std::move(deleter), extent);
  }
```
the lambda captured `dev1 by reference.

`dev` is a `const&` to the device object used in the call to `allocBuf(dev, extent)`.

If that object is itself a temporary, the reference to `dev` captured by the lambda may be dangling by the time `deleter` is called.

This change solves the problem by capturing directly the SYCL context by value.

### DIFF
--- a/include/alpaka/mem/buf/BufGenericSycl.hpp
+++ b/include/alpaka/mem/buf/BufGenericSycl.hpp
@@ -196,9 +196,7 @@ namespace alpaka::trait
                 static_cast<std::size_t>(getExtentProduct(extent)),
                 nativeDev,
                 nativeContext);
-            // captured structured bindings are a C++20 extension
-            // auto deleter = [nativeContext](TElem* ptr) { sycl::free(ptr, nativeContext); };
-            auto deleter = [&dev](TElem* ptr) { sycl::free(ptr, dev.getNativeHandle().second); };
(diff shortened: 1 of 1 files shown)


fwyzard:
For whatever reason the `randomStrategies` example has been failing with a SEGV for me today, and this change fixes the crash.

The crash itself seems semi-random, as @AuroraPerego was not able to reproduce it running with the same tools on the same machine 🤷🏻‍♂️

bernhardmgruber (approved):

fwyzard:
(please wait before merging, I still have a crash with other changes on top of this)

bernhardmgruber:
> (please wait before merging, I still have a crash with other changes on top of this)

Sure, just mark the PR as ready for review again!

fwyzard:
False alarm, the local copy on the second machine was not up to date.

j-stephan (approved):

bernhardmgruber (approved):

PR #2041, opened 2023-08-01
GitLab CI: job generator checks if container images exists

SimeonEhrig:
- Fetch all images from the container registry when the jobs are generated.
- If an image does not exist use the best working alternative.
- The check can be disabled because it takes a lot of time and is annoying if you want to test a regex to filter jobs.

PR #2043, opened 2023-08-01
Support CUDA 12.2

### LINKED ISSUE #1995: Support CUDA 12.2
https://docs.nvidia.com/cuda/cuda-toolkit-release-notes/index.html

SimeonEhrig:
The PR will add support for CUDA 12.2. But first I want to use it to test some new job generator features.

For example automatic selection of container images, implemented in PR #2041


bernhardmgruber (approved):

j-stephan (approved):

j-stephan:
Is there a reason for marking GCC 10/11 as not supporting CUDA 12.2 in the support matrix?

SimeonEhrig:
> Is there a reason for marking GCC 10/11 as not supporting CUDA 12.2 in the support matrix?

Nope, I simply replaced CUDA 12.1 with CUDA 12.2. But I recognized, that no jobs with nvcc 12.2 and gcc 10/11 are generated. Let's have a look why.

SimeonEhrig:
> > Is there a reason for marking GCC 10/11 as not supporting CUDA 12.2 in the support matrix?
> 
> Nope, I simply replaced CUDA 12.1 with CUDA 12.2. But I recognized, that no jobs with nvcc 12.2 and gcc 10/11 are generated. Let's have a look why.

Looks like the pair-wise library does not generate the combination of nvcc 12.2 + gcc 10/11 in general. I have no idea why. Independent of the CI test, should I change the README.md? We do not disallow it, therefore it is actual supported.

j-stephan:
No, leave it as-is. The support matrix should reflect what is actually tested in the CI. We can always change it later when you have figured out why there are no tests for gcc-10/11.

SimeonEhrig:
@j-stephan ready to merge

PR #2048, opened 2023-08-02
Rewrite counterBasedRng example using mdspan

bernhardmgruber:
In order to remove the experimental accessors, we first need to rewrite the `counterBasedRng` example to use something else, e.g. `std::mdspan`.

### DIFF
--- a/example/counterBasedRng/CMakeLists.txt
+++ b/example/counterBasedRng/CMakeLists.txt
@@ -33,6 +33,11 @@ if(NOT TARGET alpaka::alpaka)
     endif()
 endif()
 
+if (alpaka_USE_MDSPAN STREQUAL "OFF")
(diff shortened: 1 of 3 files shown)


example/counterBasedRng/CMakeLists.txt
```diff
@@ -33,6 +33,11 @@ if(NOT TARGET alpaka::alpaka)
     endif()
 endif()
 
+if (alpaka_USE_MDSPAN STREQUAL "OFF")
```
j-stephan, line 36: Just a hint: You could also do this:

```suggestion
if (NOT alpaka_USE_MDSPAN)
```

j-stephan (approved):

example/counterBasedRng/CMakeLists.txt
bernhardmgruber: Are you sure? `alpaka_USE_MDSPAN` can have the values `OFF`, `SYSTEM` or `FETCH`.

PR #2056, opened 2023-08-03
Always use `std::size_t` for CUDA pitch calculations

### LINKED ISSUE #1933: BabelStream error
Running Babelstream with a workingset of 6.4 GB on a V100 with 32GB memory produces the following error:
```
bash-4.2$ make babelstream && ./example/babelstream/babelstream -s 268435456
[100%] Built target babelstream
BabelStream
Version: 4.0
Implementation: alpaka
Running kernels 100 times
Precision: double
Array size: 2147.5 MB (=2.1 GB)
Total size: 6442.5 MB (=6.4 GB)
Using alpaka device Tesla V100-SXM2-32GB
  what():  /home/bg143886/alpaka/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp(161) 'TApi::memcpyAsync( m_dstMemNative, m_srcMemNative, static_cast<std::size_t>(m_extentWidthBytes), m_uniformMemCpyKind, queue.getNativeHandle())' returned error  : 'cudaErrorInvalidValue': 'invalid argument'!
Aborted
```
This seems odd, since the allocated memory is far less then what the GPU has available.

j-stephan:
This PR hard-codes buffer and pitch sizes to `std::size_t` as there is no real reason for having other types here. It is also an attempt to fix #1933.

### DIFF
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -1,5 +1,5 @@
-/* Copyright 2022 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci, Jan Stephan,
- * Bernhard Manfred Gruber, Antonio Di Pilato
+/* Copyright 2023 Alexander Matthes, Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci, Jan Stephan,
+ *                Bernhard Manfred Gruber, Antonio Di Pilato
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -16,6 +16,7 @@
 #include "alpaka/meta/DependentFalseType.hpp"
 #include "alpaka/vec/Vec.hpp"
 
+#include <cstddef>
 #include <functional>
 #include <memory>
 #include <type_traits>
@@ -48,7 +49,7 @@ namespace alpaka
             DevUniformCudaHipRt<TApi> const& dev,
             TElem* const pMem,
             Deleter deleter,
-            TIdx const& pitchBytes,
+            std::size_t pitchBytes,
             TExtent const& extent)
             : m_dev(dev)
             , m_extentElements(getExtentVecEnd<TDim>(extent))
@@ -70,7 +71,7 @@ namespace alpaka
         DevUniformCudaHipRt<TApi> m_dev;
         Vec<TDim, TIdx> m_extentElements;
         std::shared_ptr<TElem> m_spMem;
-        TIdx m_pitchBytes;
+        std::size_t m_pitchBytes;
     };
 
     namespace trait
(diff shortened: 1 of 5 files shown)


fwyzard:
In CMSSW we use `Idx = uint32_t`, because I am fairly convinced we do not have any single object even close to 4GB.

Before making this change, I would like to have a stable code base and test if this has any impact.

j-stephan:
Okay, sure. Should this PR hurt your codebase I can also change it to accept general unsigned integer types.

bernhardmgruber:
I think using `std::size_t` in host code should generally be fine. I wonder about the pitch traits tough, since you could call them in device code and pitches (at least for the near future) should fit into a 32bit int :)

> In CMSSW we use Idx = uint32_t, because I am fairly convinced we do not have any single object even close to 4GB.

Great! I hit this limit all the time though, but I do synthetic benchmarks for my PhD, which involves arrays that should fill recent GPUs.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -41,14 +42,15 @@ namespace alpaka
             "The elem type of the buffer can not be const because the C++ Standard forbids containers of const "
             "elements!");
         static_assert(!std::is_const_v<TIdx>, "The idx type of the buffer can not be const!");
+        static_assert(std::is_unsigned_v<TIdx>, "The idx type of the buffer must be an unsigned arithmetic type!");
```
bernhardmgruber: Why is this required?

j-stephan:
This is now ready for review / checking out if it breaks any existing codebases...

bernhardmgruber:
Tried your PR on hemera, looks good:
```
bash-4.2$ make -j10 babelstream && ./example/babelstream/babelstream -s 268435456
[100%] Built target babelstream
BabelStream
Version: 4.0
Implementation: alpaka
Running kernels 100 times
Precision: double
Array size: 2147.5 MB (=2.1 GB)
Total size: 6442.5 MB (=6.4 GB)
Using alpaka device Tesla V100-SXM2-32GB
Function    MBytes/sec  Min (sec)   Max         Average     
Copy        794093.030  0.00541     0.00542     0.00541     
Mul         800666.506  0.00536     0.00537     0.00537     
Add         819746.406  0.00786     0.00800     0.00798     
Triad       819874.513  0.00786     0.00803     0.00800     
Dot         832608.915  0.00516     0.00520     0.00518   
```
So, at least #1933 should be solved with this.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -173,9 +174,20 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getPitchBytes(BufUniformCudaHipRt<TApi, TElem, TDim, TIdx> const& buf) -> TIdx
             {
-                return buf.m_pitchBytes; // TODO(bgruber): is this even correct? This reports the pitch for the TDim -
-                                         // 1 dimension, but CUDA's pitch is always the row pitch, independently of the
-                                         // buffer's dimensions.
+                constexpr auto idx = static_cast<TIdx>(DimInt<TDim::value - 1u>::value);
```
bernhardmgruber:
```suggestion
                constexpr auto idx = static_cast<TIdx>(TDim::value - 1u);
```

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -173,9 +174,20 @@ namespace alpaka
...
+                constexpr auto bufDim = Dim<BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>::value;
```
bernhardmgruber: Isn't this just:
```suggestion
                constexpr auto bufDim = TDim::value;
```

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
j-stephan: Huh, I was sure I deleted this. This shouldn't be here.

bernhardmgruber: Interesting. I made that comment a few days ago and it got flushed when I reviewed today, although the line is no longer present in the current changeset :D Well, forget about it :)

bernhardmgruber (approved):

fwyzard (commented):
can you give us a couple of days to check the impact of this PR on the pixeltrack performance, before merging it ?

j-stephan:
Sure! I'll convert it back to a draft until you are done.

bernhardmgruber:
The proposed changes fix issue #1933, great work!

fwyzard:
Sorry for the delay, I finally managed to check this, it looks good on `pixeltrack-standalone`.

fwyzard (approved):

PR #2058, opened 2023-08-03
fix events and the tests

### LINKED ISSUE #1936: `eventTest` has a race condition
The `eventTest` in alpaka has a race condition that spurriously occurs in the CI. TSAN finds it as well:
```
==================
WARNING: ThreadSanitizer: data race (pid=68572)
  Write of size 8 at 0x7b4400000490 by main thread:
    #0 operator delete(void*, unsigned long) ../../../../src/libsanitizer/tsan/tsan_new_delete.cpp:150 (libtsan.so.2+0x8cd75)
    #1 std::_Function_handler<void (), alpaka::generic::detail::QueueGenericThreadsNonBlockingImpl<alpaka::DevCpu>::~QueueGenericThreadsNonBlockingImpl()::{lambda()#1}>::_M_manager(std::_Any_data&, std::_Any_data const&, std::_Manager_operation) <null> (eventTest+0x344a5)

  Previous atomic read of size 1 at 0x7b4400000490 by thread T1:
    #0 pthread_mutex_lock ../../../../src/libsanitizer/sanitizer_common/sanitizer_common_interceptors.inc:4324 (libtsan.so.2+0x59bcf)
    #1 std::_Sp_counted_ptr_inplace<alpaka::core::detail::ConcurrentExecPool<unsigned long, std::thread, std::promise, void, std::mutex, std::condition_variable, false>, std::allocator<void>, (__gnu_cxx::_Lock_policy)2>::_M_dispose() <null> (eventTest+0x32e9d)

  Thread T1 (tid=68574, finished) created by main thread at:
    #0 pthread_create ../../../../src/libsanitizer/tsan/tsan_interceptors_posix.cpp:1001 (libtsan.so.2+0x63a69)
    #1 std::thread::_M_start_thread(std::unique_ptr<std::thread::_State, std::default_delete<std::thread::_State> >, void (*)()) <null> (libstdc++.so.6+0xdc478)
    #2 Catch::RunContext::runCurrentTest(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&, std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >&) <null> (eventTest+0x89062)
    #3 Catch::RunContext::runTest(Catch::TestCaseHandle const&) <null> (eventTest+0x8963e)
    #4 Catch::Session::runInternal() <null> (eventTest+0x4450b)
    #5 Catch::Session::run() <null> (eventTest+0x44ea6)
    #6 main <null> (eventTest+0x277b9)

SUMMARY: ThreadSanitizer: data race ../../../../src/libsanitizer/tsan/tsan_new_delete.cpp:150 in operator delete(void*, unsigned long)
==================
```
The usual suspects again, `ConcurrentExecPool` and `QueueGenericThreadsNonBlockingImpl`. May be related to the queue bug #1741.

psychocoderHPC:
- fix event tests
- add a sleep to the test EventHostManualTrigger to be able to use  `alpaka::isComplete()` without taking false negatives into account.
- fix EventGenericThreads implementation
  - The behavior for a re-enqueued event was not following the CUDA  the behavior we assume in alpaka
- extend the documenation for `alpaka::wait()` and `alpaka::enqueue()`
- add tests to check the correct behavior for `alpaka::wait(queue, event)`.

### DIFF
--- a/include/alpaka/event/EventGenericThreads.hpp
+++ b/include/alpaka/event/EventGenericThreads.hpp
@@ -297,15 +297,8 @@ namespace alpaka
                 {
                     auto oldFuture = spEventImpl->m_future;
 
-                    // Enqueue a task that waits for the given event.
-                    queueImpl.m_workerThread.submit(
-                        [spEventImpl, oldFuture]() mutable
-                        {
-                            std::unique_lock<std::mutex> lk2(spEventImpl->m_mutex);
-                            lk2.unlock();
-                            oldFuture.get();
-                            lk2.lock();
-                        });
+                    // Enqueue a task that waits for the given future of the event.
+                    queueImpl.m_workerThread.submit([oldFuture]() { oldFuture.get(); });
                 }
             }
         };
--- a/include/alpaka/event/EventGenericThreads.hpp
+++ b/include/alpaka/event/EventGenericThreads.hpp
@@ -152,7 +152,8 @@ namespace alpaka
                         // Nothing to do if it has been re-enqueued to a later position in the queue.
                         if(enqueueCount == spEventImpl->m_enqueueCount)
                         {
-                            spEventImpl->m_LastReadyEnqueueCount = spEventImpl->m_enqueueCount;
+                            spEventImpl->m_LastReadyEnqueueCount
+                                = std::max(enqueueCount, spEventImpl->m_LastReadyEnqueueCount);
                         }
                     });
             }
@@ -294,14 +295,16 @@ namespace alpaka
 
                 if(!spEventImpl->isReady())
                 {
-                    auto const enqueueCount = spEventImpl->m_enqueueCount;
+                    auto oldFuture = spEventImpl->m_future;
 
                     // Enqueue a task that waits for the given event.
                     queueImpl.m_workerThread.submit(
-                        [spEventImpl, enqueueCount]() mutable
+                        [spEventImpl, oldFuture]() mutable
                         {
                             std::unique_lock<std::mutex> lk2(spEventImpl->m_mutex);
-                            spEventImpl->wait(enqueueCount, lk2);
+                            lk2.unlock();
+                            oldFuture.get();
+                            lk2.lock();
                         });
                 }
             }
--- a/include/alpaka/queue/Traits.hpp
+++ b/include/alpaka/queue/Traits.hpp
@@ -41,6 +41,8 @@ namespace alpaka
     //!   If the event has previously been queued, then this call will overwrite any existing state of the event.
     //!   Any subsequent calls which examine the status of event will only examine the completion of this most recent
     //!   call to enqueue.
+    //!   If a queue is waiting for an event the event state of the recently captured state at the time of the API call
+    //!   to wait() will be used to release the queue.
     template<typename TQueue, typename TTask>
     ALPAKA_FN_HOST auto enqueue(TQueue& queue, TTask&& task) -> void
     {
--- a/include/alpaka/queue/Traits.hpp
+++ b/include/alpaka/queue/Traits.hpp
@@ -41,8 +41,8 @@ namespace alpaka
     //!   If the event has previously been queued, then this call will overwrite any existing state of the event.
     //!   Any subsequent calls which examine the status of event will only examine the completion of this most recent
     //!   call to enqueue.
-    //!   If a queue is waiting for an event the event state of the recently captured state at the time of the API call
-    //!   to wait() will be used to release the queue.
+    //!   If a queue is waiting for an event the latter's event state at the time of the API call to wait() will be
+    //!   used to release the queue.
     template<typename TQueue, typename TTask>
     ALPAKA_FN_HOST auto enqueue(TQueue& queue, TTask&& task) -> void
     {
--- a/include/alpaka/test/event/EventHostManualTrigger.hpp
+++ b/include/alpaka/test/event/EventHostManualTrigger.hpp
@@ -57,6 +57,8 @@ namespace alpaka::test
                     m_bIsReady = true;
                 }
                 m_conditionVariable.notify_one();
+                // Give alpaka time to update into the new state, process all events and tasks.
+                std::this_thread::sleep_for(std::chrono::milliseconds(200u));
             }
 
         public:
@@ -96,6 +98,8 @@ namespace alpaka::test
         void trigger()
         {
             m_spEventImpl->trigger();
+            // Give alpaka time to update into the new state, process all events and tasks.
+            std::this_thread::sleep_for(std::chrono::milliseconds(200u));
         }
 
     public:
@@ -282,6 +286,8 @@ namespace alpaka::test
                 // Initiate the memory set.
                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
                     cudaMemset(m_devMem, static_cast<int>(1u), static_cast<size_t>(sizeof(int32_t))));
+                // Give alpaka time to update into the new state, process all events and tasks.
+                std::this_thread::sleep_for(std::chrono::milliseconds(200u));
             }
 
         public:
@@ -315,6 +321,8 @@ namespace alpaka::test
         void trigger()
         {
             m_spEventImpl->trigger();
+            // Give alpaka time to update into the new state, process all events and tasks.
+            std::this_thread::sleep_for(std::chrono::milliseconds(200u));
         }
 
     public:
@@ -509,6 +517,8 @@ namespace alpaka::test
                 // Initiate the memory set.
                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
                     hipMemset(m_devMem, static_cast<int>(1u), static_cast<size_t>(sizeof(int32_t))));
+                // Give alpaka time to update into the new state, process all events and tasks.
+                std::this_thread::sleep_for(std::chrono::milliseconds(200u));
             }
 
         public:
@@ -542,6 +552,8 @@ namespace alpaka::test
         void trigger()
         {
             m_spEventImpl->trigger();
+            // Give alpaka time to update into the new state, process all events and tasks.
+            std::this_thread::sleep_for(std::chrono::milliseconds(200u));
         }
 
     public:
--- a/include/alpaka/wait/Traits.hpp
+++ b/include/alpaka/wait/Traits.hpp
@@ -28,8 +28,8 @@ namespace alpaka
     //! Waits the thread for the completion of the given awaited action to complete.
     //!
     //! Special Handling for events:
-    //!   If the event is re-enqueued wait will terminate when the re-enqueued event will be ready and previous enqueue
-    //!   sates of the event will be ignored.
+    //!   If the event is re-enqueued wait() will terminate when the re-enqueued event will be ready and previously
+    //!   enqueued states of the event will be ignored.
     template<typename TAwaited>
     ALPAKA_FN_HOST auto wait(TAwaited const& awaited) -> void
     {
@@ -39,9 +39,9 @@ namespace alpaka
 
     //! The waiter waits for the given awaited action to complete.
     //!
-    //! Special Handling if waiter is a queue and awaited an event:
-    //!   The waiter waits for the event state becoming ready based on the recently captured event state at the time of
-    //!   the API call even if the event is being re-enqueued later.
+    //! Special Handling if \p waiter is a queue and \p awaited an event:
+    //!   The \p waiter waits for the event state to become ready based on the recently captured event state at the
+    //!   time of the API call even if the event is being re-enqueued later.
     template<typename TWaiter, typename TAwaited>
     ALPAKA_FN_HOST auto wait(TWaiter& waiter, TAwaited const& awaited) -> void
     {
--- a/include/alpaka/wait/Traits.hpp
+++ b/include/alpaka/wait/Traits.hpp
@@ -26,6 +26,10 @@ namespace alpaka
     } // namespace trait
 
     //! Waits the thread for the completion of the given awaited action to complete.
+    //!
+    //! Special Handling for events:
+    //!   If the event is re-enqueued wait will terminate when the re-enqueued event will be ready and previous enqueue
+    //!   sates of the event will be ignored.
     template<typename TAwaited>
     ALPAKA_FN_HOST auto wait(TAwaited const& awaited) -> void
     {
@@ -34,6 +38,10 @@ namespace alpaka
     }
 
     //! The waiter waits for the given awaited action to complete.
+    //!
+    //! Special Handling if waiter is a queue and awaited an event:
+    //!   The waiter waits for the event state becoming ready based on the recently captured event state at the time of
+    //!   the API call even if the event is being re-enqueued later.
     template<typename TWaiter, typename TAwaited>
     ALPAKA_FN_HOST auto wait(TWaiter& waiter, TAwaited const& awaited) -> void
     {
--- a/test/unit/event/src/EventTest.cpp
+++ b/test/unit/event/src/EventTest.cpp
@@ -186,7 +186,7 @@ TEMPLATE_LIST_TEST_CASE("eventReEnqueueShouldBePossibleIfSomeoneWaitsFor", "[eve
             REQUIRE(alpaka::isComplete(k1));
             REQUIRE(!alpaka::isComplete(k2));
             REQUIRE(!alpaka::isComplete(e1));
-            REQUIRE(!alpaka::isComplete(e2));
+            REQUIRE(alpaka::isComplete(e2));
 
             // q1 = [e1]
             k2.trigger();
@@ -203,7 +203,6 @@ TEMPLATE_LIST_TEST_CASE("eventReEnqueueShouldBePossibleIfSomeoneWaitsFor", "[eve
         }
     }
 }
-
 
 // github issue #388
 TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[event]", TestQueues)
--- a/test/unit/event/src/EventTest.cpp
+++ b/test/unit/event/src/EventTest.cpp
@@ -150,46 +150,56 @@ TEMPLATE_LIST_TEST_CASE("eventReEnqueueShouldBePossibleIfSomeoneWaitsFor", "[eve
             alpaka::test::EventHostManualTrigger<Dev> k1(f1.m_dev);
             alpaka::test::EventHostManualTrigger<Dev> k2(f1.m_dev);
 
-            // q1 = [k1]
             alpaka::enqueue(q1, k1);
+
+            // q1 = [k1]
             REQUIRE(!alpaka::isComplete(k1));
 
-            // q1 = [k1, e1]
             alpaka::enqueue(q1, e1);
+
+            // q1 = [k1, e1]
             REQUIRE(!alpaka::isComplete(k1));
             REQUIRE(!alpaka::isComplete(e1));
 
-            // q1 = [k1, e1, k2]
             alpaka::enqueue(q1, k2);
+
+            // q1 = [k1, e1, k2]
             REQUIRE(!alpaka::isComplete(k1));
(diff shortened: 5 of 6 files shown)


test/unit/event/src/EventTest.cpp
```diff
@@ -150,46 +150,56 @@ TEMPLATE_LIST_TEST_CASE("eventReEnqueueShouldBePossibleIfSomeoneWaitsFor", "[eve
...
-            // q1 = [k2, e1]
             k1.trigger();
+
+            // q1 = [k2, e1_new]
+            // q2 = []
             REQUIRE(alpaka::isComplete(k1));
             REQUIRE(!alpaka::isComplete(k2));
             REQUIRE(!alpaka::isComplete(e1));
-            REQUIRE(!alpaka::isComplete(e2));
+            REQUIRE(alpaka::isComplete(e2));
```
psychocoderHPC, line 198: This was wrong before. It was working in CUDA because there is an event state `event not ready`. Due to the fact, that's CUDA driver was not updating the state of the event e2 fast enough the event was in the state `event not ready` what we in alpaka interpret as it is not complete (and this is correct and documented).
For testing we need relatable states and no false negatives, therefore `trigger()` is now waiting long enough to update the internal state of alpaka and CUDA objects.

psychocoderHPC: Btw the CPU backend was implementing it wrongly and was taking always the last enqueue of an event into account instead the time of the `wait(queue,event)` API call.

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+            // q3 = [->e1_new,e3]
+            REQUIRE(!alpaka::isComplete(k1_0));
+            REQUIRE(!alpaka::isComplete(k1_1));
+            REQUIRE(alpaka::isComplete(k2));
+            REQUIRE(alpaka::isComplete(k3));
+            REQUIRE(!alpaka::isComplete(e1));
+            REQUIRE(!alpaka::isComplete(e2));
+            REQUIRE(!alpaka::isComplete(e3));
+
+            // After the kernel k1_0 is released e3 is not allowed to be ready because q3 depends on the oldest e1
```
psychocoderHPC: should be `... e1_new`

include/alpaka/event/EventGenericThreads.hpp
```diff
@@ -294,14 +295,16 @@ namespace alpaka
...
                     // Enqueue a task that waits for the given event.
                     queueImpl.m_workerThread.submit(
-                        [spEventImpl, enqueueCount]() mutable
+                        [spEventImpl, oldFuture]() mutable
                         {
                             std::unique_lock<std::mutex> lk2(spEventImpl->m_mutex);
-                            spEventImpl->wait(enqueueCount, lk2);
+                            lk2.unlock();
+                            oldFuture.get();
+                            lk2.lock();
```
bernhardmgruber:
```suggestion
                            lk2.unlock();
                            oldFuture.get();
```
Why would the last lock be necessary? It is immediately unlocked again at the closing brace.

include/alpaka/queue/Traits.hpp
```diff
@@ -41,6 +41,8 @@ namespace alpaka
     //!   If the event has previously been queued, then this call will overwrite any existing state of the event.
     //!   Any subsequent calls which examine the status of event will only examine the completion of this most recent
     //!   call to enqueue.
+    //!   If a queue is waiting for an event the event state of the recently captured state at the time of the API call
+    //!   to wait() will be used to release the queue.
```
bernhardmgruber: I do not understand this statement. What is the "recently captured state"? Who captures it? The task?

include/alpaka/test/event/EventHostManualTrigger.hpp
```diff
@@ -57,6 +57,8 @@ namespace alpaka::test
                     m_bIsReady = true;
                 }
                 m_conditionVariable.notify_one();
+                // Give alpaka time to update into the new state, process all events and tasks.
+                std::this_thread::sleep_for(std::chrono::milliseconds(200u));
```
bernhardmgruber, line 61: I am very concerned if any of our multithreaded logic depends on a wait. Is there no other way to establish a synchronization?

include/alpaka/event/EventGenericThreads.hpp
psychocoderHPC: You are right it is an left over from a version I used before where I was required to change a member of the event

include/alpaka/queue/Traits.hpp
psychocoderHPC: I change the formulation. The state of the event at the API call time is used to build dependencies (wait for an event) equally to CUDA.

include/alpaka/test/event/EventHostManualTrigger.hpp
psychocoderHPC: I do not know any other way because we do not know how long CUDA or alpaka needs to run async tasks.
Calling wait or any other function is also not good because if you build your runtime engine as MPI is doing it you will always perform actions if an API call to the library is performed instead of asynchronous in the background.

IMO sleeping here is not 100% safe but OKisch. It is not affecting the core functionality of alpaka since this is a class used only for testing.

include/alpaka/queue/Traits.hpp
```diff
@@ -41,6 +41,8 @@ namespace alpaka
     //!   If the event has previously been queued, then this call will overwrite any existing state of the event.
     //!   Any subsequent calls which examine the status of event will only examine the completion of this most recent
     //!   call to enqueue.
+    //!   If a queue is waiting for an event the recently event state at the time of the API call to wait() will be
+    //!   used to release the queue.
```
j-stephan:
```suggestion
    //!   If a queue is waiting for an event the latter's state at the time of the API call to wait() will be
    //!   used to release the queue.
```

include/alpaka/test/event/EventHostManualTrigger.hpp
j-stephan: Could we wait on an atomic variable here?

include/alpaka/wait/Traits.hpp
```diff
@@ -26,6 +26,10 @@ namespace alpaka
     } // namespace trait
 
     //! Waits the thread for the completion of the given awaited action to complete.
+    //!
+    //! Special Handling for events:
+    //!   If the event is re-enqueued wait will terminate when the re-enqueued event will be ready and previous enqueue
+    //!   sates of the event will be ignored.
```
j-stephan:
```suggestion
    //!   If the event is re-enqueued wait() will terminate when the re-enqueued event will be ready and previously enqueued
    //!   states of the event will be ignored.
```

include/alpaka/wait/Traits.hpp
```diff
@@ -34,6 +38,10 @@ namespace alpaka
     }
 
     //! The waiter waits for the given awaited action to complete.
+    //!
+    //! Special Handling if waiter is a queue and awaited an event:
+    //!   The waiter waits for the event state becoming ready based on the recently captured event state at the time of
+    //!   the API call even if the event is being re-enqueued later.
```
j-stephan:
```suggestion
    //! Special Handling if \p waiter is a queue and \p awaited an event:
    //!   The \p waiter waits for the event state to become ready based on the captured event state at the time of
    //!   the API call to wait() even if the event is being re-enqueued later.
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
 
+            // q2 = [k2, ->e1, e1]
+            REQUIRE(!alpaka::isComplete(k2));
+            REQUIRE(!alpaka::isComplete(e1));
+
             // 8. kernel k2 finishes
-            // q2 = [->e1, e1]
             k2.trigger();
+            // q2 = []
+            REQUIRE(alpaka::isComplete(e1));
```
j-stephan, line 278: Shouldn't we wait for `e1` here before asking if it is ready?

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+            alpaka::enqueue(q1, k1_0);
+            alpaka::enqueue(q1, e1);
+            // q1 = [k1_0,e1]
+            alpaka::enqueue(q2, k2);
+            // q2 = [k2]
+            REQUIRE(!alpaka::isComplete(k1_0));
+
+            alpaka::wait(q2, e1);
+            alpaka::enqueue(q2, e2);
+            // q2 = [->e1,e2]
```
j-stephan:
```suggestion
            // q2 = [k2,->e1,e2]
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+            REQUIRE(alpaka::isComplete(k2));
+            REQUIRE(alpaka::isComplete(k3));
+            REQUIRE(!alpaka::isComplete(e1));
+            REQUIRE(alpaka::isComplete(e2));
+            REQUIRE(!alpaka::isComplete(e3));
+
+            k1_1.trigger();
+            // q1 = []
             // q2 = []
+            REQUIRE(alpaka::isComplete(k1_0));
```
j-stephan:
```suggestion
            // q1 = []
            // q2 = []
            // q3 = []
            REQUIRE(alpaka::isComplete(k1_0));
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
             // q2 and its current state is 'unfinished' again. q2 = [e1]
 
             // Both queues should successfully finish
             alpaka::wait(q1);
+            alpaka::wait(q2);
+        }
+        else
+        {
+            std::cerr << "Can not execute test because CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS is not supported!"
+                      << std::endl;
```
j-stephan:
```suggestion
            std::cerr << "Cannot execute test because CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS is not supported by the device!"
                      << std::endl;
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+            REQUIRE(alpaka::isComplete(k2));
+            REQUIRE(alpaka::isComplete(k3));
+            REQUIRE(alpaka::isComplete(e1));
+            REQUIRE(alpaka::isComplete(e2));
+            REQUIRE(alpaka::isComplete(e3));
+        }
+        else
+        {
+            std::cerr << "Can not execute test because CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS is not supported!"
+                      << std::endl;
```
j-stephan:
```suggestion
            std::cerr << "Cannot execute test because CU_DEVICE_ATTRIBUTE_CAN_USE_STREAM_MEM_OPS is not supported by the device!"
                      << std::endl;
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+            alpaka::test::EventHostManualTrigger<Dev> k2(f2.m_dev);
+
+            alpaka::enqueue(q1, k1_0);
+            alpaka::enqueue(q1, e1);
+            // q1 = [k1_0,e1]
+            alpaka::enqueue(q2, k2);
+            // q2 = [k2]
+            REQUIRE(!alpaka::isComplete(k1_0));
+
+            alpaka::wait(q3, e1); // let wait
```
j-stephan:
```suggestion
            alpaka::wait(q3, e1);
```

Looks like a leftover from our VC ;-)

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+            alpaka::enqueue(q3, e3);
+            // q3 = [->e1,e3]
+
+            alpaka::enqueue(q2, e1);
+            // q2 = [k2,e1(1)]
```
j-stephan:
```suggestion
            // q2 = [k2,e1_new]
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+
+            alpaka::enqueue(q2, e2);
+            // q1 = [k1_0,e1]
+            // q2 = [k2,e1_new,e3]
+            // q3 = [->e1,e3]
```
j-stephan:
```suggestion
            alpaka::enqueue(q2, e2);
            // q1 = [k1_0,e1]
            // q2 = [k2,e1_new,e2]
            // q3 = [->e1,e3]
```

test/unit/event/src/EventTest.cpp
```diff
@@ -227,44 +236,260 @@ TEMPLATE_LIST_TEST_CASE("waitForEventThatAlreadyFinishedShouldBeSkipped", "[even
...
+
+            REQUIRE(!alpaka::isComplete(k1_0));
+            REQUIRE(!alpaka::isComplete(k2));
+            REQUIRE(!alpaka::isComplete(e1));
+            REQUIRE(!alpaka::isComplete(e2));
+            REQUIRE(!alpaka::isComplete(e3));
+
+            // We release first the kernel which is blocking the most resent enqueue of event e1.
+            // Queue q3 is not allowed to be freed because these queue depends on the oldest enqueue of e1.
+            k2.trigger();
```
j-stephan:
```suggestion
            // We release first the kernel which is blocking the most recent enqueue of event e1.
            // q3 is not allowed to be freed because this queue depends on the oldest enqueue of e1.
            k2.trigger();
```

j-stephan (changes requested):
This looks mostly good to me. A few comments:

include/alpaka/event/EventGenericThreads.hpp
```diff
@@ -294,15 +295,10 @@ namespace alpaka
...
 
-                    // Enqueue a task that waits for the given event.
-                    queueImpl.m_workerThread.submit(
-                        [spEventImpl, enqueueCount]() mutable
-                        {
-                            std::unique_lock<std::mutex> lk2(spEventImpl->m_mutex);
-                            spEventImpl->wait(enqueueCount, lk2);
-                        });
+                    // Enqueue a task that waits for the given future of the event.
+                    queueImpl.m_workerThread.submit([oldFuture]() { oldFuture.get(); });
```
bernhardmgruber, line 301:
```suggestion
                    queueImpl.m_workerThread.submit([oldFuture] { oldFuture.get(); });
```
if you want

include/alpaka/queue/Traits.hpp
bernhardmgruber: I am still unsure what this means. So the queue will be destroyed as part of the event's state? But that usually results in the callback thread joining itself, which is a bug. Does `release` mean destroy here?

include/alpaka/test/event/EventHostManualTrigger.hpp
bernhardmgruber: I agree, that it is OKish because it is part of the unit tests. I assume we may see the test fail sometimes, but we can always restart the CI and cross fingers. So I think we can accept this change and fix it in the future if the CI fails to often.

include/alpaka/event/EventGenericThreads.hpp
fwyzard: keep the empty `()` please, the lambdas are messy enough with them...

include/alpaka/queue/Traits.hpp
j-stephan: "Release Queue" means "Queue no longer waits"

bernhardmgruber: Can we use "unblock" instead? "Release" carries some connotation of ownership for me :)

include/alpaka/test/event/EventHostManualTrigger.hpp
psychocoderHPC: > Could we wait on an atomic variable here?

No because we can not influence what CUDA7HIP is doing in the background and even in alpaka we can to track the state.
Please keep in mind that this class is explicitly made to emulate the asynchronous behaviors of a threaded system. We control when a kernel is ready to have a well-defined state in which events and kernels must be executed.

psychocoderHPC: > I agree, that it is OKish because it is part of the unit tests. I assume we may see the test fail sometimes, but we can always restart the CI and cross fingers. So I think we can accept this change and fix it in the future if the CI fails to often.

I do not think we will ever see it failing because of the sleep. 200ms are ages compared to what we do within alpaka and even CUDA will never be so slow that we will hit the border.

test/unit/event/src/EventTest.cpp
psychocoderHPC: `trigger()` will add `200ms` wait, after that time the state of the event should be set correctly.
If we would add a wait here and someone is refactoring something which let the test fail we will produce a never terminating (will terminate after the timeout) test. This will be extremely hard to debug compared to a clear answer from the system that the event is in the wrong state.

see also https://github.com/alpaka-group/alpaka/pull/2058#discussion_r1284372985
The test is checking that all is in an explicit state.

test/unit/event/src/EventTest.cpp
psychocoderHPC: Yes, note in PIConGPU we call it `waitOn`

include/alpaka/event/EventGenericThreads.hpp
psychocoderHPC: I will keep `()`, I personally like it also more because it cleary shows the signature of the lambda.

test/unit/event/src/EventTest.cpp
j-stephan: Okay :-)

include/alpaka/queue/Traits.hpp
psychocoderHPC: Will change it to `unblcok` because we call the queue non blocking queue.
@bernhardmgruber This is a general interface, you should at the moment you read comments here forget all your background knowledge of how something is implemented. Do not assume any callback threads or similar!

psychocoderHPC: Ohh I can not change the to `unblock` because the queue is not blocked at all. I will keep release.

psychocoderHPC:
I pushed the updates!

j-stephan (commented):
Looks good to me now. But please squash the commits, I don't think this is enough for 4 commits :-)

psychocoderHPC:
> Looks good to me now. But please squash the commits, I don't think this is enough for 4 commits :-)

I on purpose added 3 commits. 
The first commit is fixing the bug, for that the new tests are not required. 
The second commit provides new tests.
The third commit is to make also old comments in the test readable.
The last are the reviews because it is too complicated to merge the changes back into the corresponding commits.
You can auto-squash the commits by pressing merge and selecting that all should be merged! No need to waist CI time again.

j-stephan (approved):

PR #2061, opened 2023-08-04
fix accessors for the SYCL backend

AuroraPerego:
#2054 broke the SYCL backend because it's still using the `Accessor` class (in one place, we are investigating if it can be dropped completely).
This PR reintroduces few definitions that make the SYCL backend compile again.

PR #2062, opened 2023-08-04
Remove remainders of Accessor

bernhardmgruber:
- [x] Wait until #2061 is merged


include/alpaka/kernel/TaskKernelGenericSycl.hpp
```diff
@@ -111,16 +110,6 @@ namespace alpaka::detail
...
 
-    template<typename TElem, typename TIdx, std::size_t TDim, typename TAccessModes>
-    inline auto require(
-        sycl::handler& cgh,
-        experimental::Accessor<SyclAccessor<TElem, DimInt<TDim>::value, TAccessModes>, TElem, TIdx, TDim, TAccessModes>
-            acc,
-        special)
-    {
-        cgh.require(acc.m_accessor);
-    }
```
bernhardmgruber: @j-stephan I don't quite understand what the replacement is here after the USM merge. Do we still need to require things anymore? And if yes, where do they come from?

include/alpaka/kernel/TaskKernelGenericSycl.hpp
```diff
@@ -111,15 +110,6 @@ namespace alpaka::detail
...
 
-    template<typename TElem, typename TIdx, std::size_t TDim, typename TAccessModes>
-    inline auto require(
-        sycl::handler& cgh,
-        Accessor<SyclAccessor<TElem, DimInt<TDim>::value, TAccessModes>, TElem, TIdx, TDim, TAccessModes> acc,
-        special)
-    {
-        cgh.require(acc.m_accessor);
-    }
-
```
j-stephan, line 122: You can also remove the other `require` overloads I think. We needed them for SYCL accessors to global memory but we don't have those anymore.

AuroraPerego:
Since this is affecting the SYCL back-end that is not checked in the CI yet, can you compile some tests for SYCL to test the changes?
Right now it's not compiling, there is an `#include "alpaka/mem/buf/sycl/Accessor.hpp"` in `include/alpaka/mem/buf/BufGenericSycl.hpp` that must be removed as well.

include/alpaka/kernel/TaskKernelGenericSycl.hpp
AuroraPerego: yes, and therefore also lines 154-155

AuroraPerego:
Since you are touching that file, also the `struct kernel` struct can be removed [here](https://github.com/alpaka-group/alpaka/pull/2062/files#diff-a73804fe988e37b9a98027f85583a10625529d1c6353e8ca7c0b3180462699e7L101-L104).
In short, everything in the `namespace alpaka::detail` is no longer used

include/alpaka/kernel/TaskKernelGenericSycl.hpp
bernhardmgruber: thx, done!

bernhardmgruber:
@AuroraPerego I applied your changes.

j-stephan:
I added a missing `<cstdint>`. After adding this the tests compile successfully on my laptop (except for two which lead to a segmentation fault in the GPU compiler - I believe this is unrelated to the changes here).

j-stephan (approved):

PR #2063, opened 2023-08-04
minor change: change const variables to constexpr

chaever:
Certain variables are declared as `const` when they could be `constexpr`.

### DIFF
--- a/include/alpaka/platform/PlatformUniformCudaHipRt.hpp
+++ b/include/alpaka/platform/PlatformUniformCudaHipRt.hpp
@@ -134,8 +134,8 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
-                constexpr std::size_t KiB{1024};
-                constexpr std::size_t MiB = kiB * kiB;
+                constexpr auto KiB = std::size_t{1024};
+                constexpr auto MiB = KiB * KiB;
                 std::cout << "name: " << devProp.name << std::endl;
                 std::cout << "totalGlobalMem: " << devProp.totalGlobalMem / MiB << " MiB" << std::endl;
                 std::cout << "sharedMemPerBlock: " << devProp.sharedMemPerBlock / KiB << " KiB" << std::endl;
--- a/include/alpaka/platform/PlatformUniformCudaHipRt.hpp
+++ b/include/alpaka/platform/PlatformUniformCudaHipRt.hpp
@@ -1,4 +1,5 @@
-/* Copyright 2022 Benjamin Worpitz, René Widera, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di Pilato, Christian Kaever
+/* Copyright 2022 Benjamin Worpitz, René Widera, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di Pilato,
+ *                Christian Kaever
  * SPDX-License-Identifier: MPL-2.0
  */
 
--- a/include/alpaka/platform/PlatformUniformCudaHipRt.hpp
+++ b/include/alpaka/platform/PlatformUniformCudaHipRt.hpp
@@ -134,8 +134,8 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
-                std::size_t const kiB(1024);
-                std::size_t const miB(kiB * kiB);
+                std::size_t constexpr kiB(1024);
+                std::size_t constexpr miB(kiB * kiB);
                 std::cout << "name: " << devProp.name << std::endl;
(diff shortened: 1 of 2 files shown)


include/alpaka/rand/TinyMT/tinymt32.h
```diff
@@ -352,9 +352,9 @@ void tinymt32_init(tinymt32_t * random, uint32_t seed) {
  */
 void tinymt32_init_by_array(tinymt32_t * random, uint32_t init_key[],
                             int key_length) {
-    const unsigned int lag = 1;
-    const unsigned int mid = 1;
-    const unsigned int size = 4;
+    constexpr unsigned int lag = 1;
+    constexpr unsigned int mid = 1;
+    constexpr unsigned int size = 4;
```
j-stephan: This is third-party code; I'd like to avoid to change it.

include/alpaka/platform/PlatformUniformCudaHipRt.hpp
```diff
@@ -134,8 +134,8 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
-                std::size_t const kiB(1024);
-                std::size_t const miB(kiB * kiB);
+                std::size_t constexpr kiB(1024);
+                std::size_t constexpr miB(kiB * kiB);
```
j-stephan: Since you're touching these lines anyway:

```suggestion
                constexpr auto KiB = std::size_t{1024};
                constexpr auto MiB = KiB * KiB;
```

Also the renaming from `{k,m}iB` to `{K,M}iB` needs to be applied where necessary.

j-stephan (changes requested):
Hi @chaever, thanks for the PR. General comments:

* Please edit the header comments of the files you touched and add your name to the copyright list.
* If you are paid for working on/with alpaka please make sure that the e-mail address in the commit is your professional e-mail address. I haven't checked this - if this is already the case you don't need to change anything here :-)

Two comments about the code:

include/alpaka/rand/TinyMT/tinymt32.h
bernhardmgruber: I agree. Is there a particular reason you need to have these files changes? If static analysis flags those, we may be able to disable it for this file.

chaever: I had originally found the other code line first, and then looked for other similar changes via `grep -r`.
The change is now revoked and should be back to the original.

include/alpaka/platform/PlatformUniformCudaHipRt.hpp
chaever: This should be included now.

include/alpaka/platform/PlatformUniformCudaHipRt.hpp
```diff
@@ -134,11 +134,11 @@ namespace alpaka
             {
                 ALPAKA_DEBUG_FULL_LOG_SCOPE;
 
-                std::size_t constexpr kiB(1024);
-                std::size_t constexpr miB(kiB * kiB);
+                constexpr std::size_t KiB{1024};
```
j-stephan: Is there a reason for preferring `KiB{1024}` over `KiB = std::size_t{1024}`?

chaever: I did that out of force of habit, sorry!
Both variables should be `auto` now.

j-stephan (dismissed):

include/alpaka/platform/PlatformUniformCudaHipRt.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2022 Benjamin Worpitz, René Widera, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di Pilato
+/* Copyright 2022 Benjamin Worpitz, René Widera, Andrea Bocci, Bernhard Manfred Gruber, Antonio Di Pilato, Christian Kaever
  * SPDX-License-Identifier: MPL-2.0
  */
```
j-stephan: Your name needs to be in a new line and aligned to `Benjamin Worpitz` above. Otherwise clang-format will complain since you are exceeding the character limit.

chaever: Thanks for the tip!   clang-fomat should no longer complain, when running
`find -iname '*.cpp' -o -iname '*.hpp' | xargs clang-format-14 -i` (taken from [here](https://github.com/alpaka-group/alpaka/blob/develop/CONTRIBUTING.md)) clang-format changes only files that are in `thirdParty/catch2`.

j-stephan (approved):

PR #2067, opened 2023-08-07
Ensure CallbackThread/ThreadPool propagate exceptions

### LINKED ISSUE #1994: Validity check of workDiv for TaskKernelCpuSerial
The CPU serial task checks the validity of the work div only at the start of kernel execution, throwing an exception if the check fails:
https://github.com/alpaka-group/alpaka/blob/375f4f0e161839caa47efa3342f815be1181c5aa/include/alpaka/kernel/TaskKernelCpuSerial.hpp#L82
In conjunction with the non-blocking CPU queue, such an exception is stored away in a TaskPkg, with no apparent way for the calling code/user to check for it:
https://github.com/alpaka-group/alpaka/blob/375f4f0e161839caa47efa3342f815be1181c5aa/include/alpaka/core/ConcurrentExecPool.hpp#L75-L82
This leads to kernels with invalid workdiv to *silently* no be executed, presenting a major stumbling block for new users.
[`TaskKernelCpuOmp2Blocks`](https://github.com/alpaka-group/alpaka/blob/375f4f0e161839caa47efa3342f815be1181c5aa/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp#L827-L830) implements the same behavior.

For [`TaskKernelGpuUniformCudaHipRt`](https://github.com/alpaka-group/alpaka/blob/375f4f0e161839caa47efa3342f815be1181c5aa/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp#L212) the queue's Enqueue trait is specialized to perform the check at enqueue-time.

bernhardmgruber:
As a side effect, this replaces the `std::packaged_task` inside `CallbackThread` by a custom virtualization mechanism, so exceptions are propagated.

Fixes: #1994

### DIFF
--- a/include/alpaka/core/CallbackThread.hpp
+++ b/include/alpaka/core/CallbackThread.hpp
@@ -4,6 +4,9 @@
 
 #pragma once
 
+#include "alpaka/core/BoostPredef.hpp"
+
+#include <cassert>
 #include <condition_variable>
 #include <functional>
 #include <future>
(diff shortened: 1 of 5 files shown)


j-stephan:
Note that the GitLab pipeline is still running (https://gitlab.com/hzdr/crp/alpaka/-/pipelines/958241668). GitHub is showing an older pipeline state here.

include/alpaka/core/CallbackThread.hpp
```diff
@@ -16,10 +19,40 @@ namespace alpaka::core
 {
     class CallbackThread
     {
-        // std::packaged_task is used because std::function requires that the wrapped callable is copyable.
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wweak-vtables"
+#endif
+        // A custom class is used because std::function<F> requires F to be copyable, and std::packed_task provides a
+        // std::future which will keep the task alive and we can not control the moment the future is set.
```
j-stephan:
```suggestion
        // A custom class is used because std::function<F> requires F to be copyable, and std::packaged_task provides a
        // std::future which will keep the task alive and we cannot control the moment the future is set.
```

j-stephan (changes requested):

j-stephan (approved):

PR #2071, opened 2023-08-09
remove dummy atomics from MemFence

### LINKED ISSUE #2069: memFence and atomics
Today during the meeting we have discussed about the way fences work and how they are implemented in alpaka.
[According to the c++ standard](https://en.cppreference.com/w/cpp/atomic/atomic_thread_fence), they should be used together with atomics to be meaningful and the SYCL standard behaves in the same way. For this reason both in the serial and in the SYCL back-end the fences have been surrounded with atomics.
However, if a user needs a fence in his code this means that he is already using atomics, therefore adding another one in alpaka should not be necessary.
Considering this, the proposal is to [drop the atomics](https://github.com/AuroraPerego/alpaka/commit/5b70b353de2ef5f2c9d6915313a59f13a970fbb8) in the fences and to document properly their usage in the documentation.

FYI @fwyzard

AuroraPerego:
Addresses #2069

### DIFF
--- a/include/alpaka/mem/fence/MemFenceCpu.hpp
+++ b/include/alpaka/mem/fence/MemFenceCpu.hpp
@@ -54,13 +54,7 @@ namespace alpaka
                  *   a == 10 && b == 20
                  */
 
-                static auto dummy = std::atomic<int>{42};
-
-                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
-                 * these dummy operations to ensure this.*/
-                auto x = dummy.load(std::memory_order_relaxed);
                 std::atomic_thread_fence(std::memory_order_acq_rel);
-                dummy.store(x, std::memory_order_relaxed);
             }
         };
     } // namespace trait
--- a/include/alpaka/mem/fence/MemFenceCpuSerial.hpp
+++ b/include/alpaka/mem/fence/MemFenceCpuSerial.hpp
@@ -42,14 +42,7 @@ namespace alpaka
             static auto mem_fence(MemFenceCpuSerial const&, TMemScope const&)
             {
                 /* Enable device fences because we may want to synchronize with other (serial) kernels. */
-
-                static std::atomic<int> dummy{42};
-
-                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
-                 * these dummy operations to ensure this.*/
-                auto x = dummy.load(std::memory_order_relaxed);
                 std::atomic_thread_fence(std::memory_order_acq_rel);
-                dummy.store(x, std::memory_order_relaxed);
             }
         };
     } // namespace trait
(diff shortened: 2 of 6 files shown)


include/alpaka/mem/fence/MemFenceCpu.hpp
```diff
@@ -54,13 +54,10 @@ namespace alpaka
...
                  */
 
-                static auto dummy = std::atomic<int>{42};
-
-                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
-                 * these dummy operations to ensure this.*/
-                auto x = dummy.load(std::memory_order_relaxed);
+                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them.
+                 * When using the fence, ensure that this condition is met in the code.
+                 */
```
bernhardmgruber: Please also drop this comment, since it's no longer accurate. Fences still need atomics, but the user must now provide them. You may add a documentation comment somewhere else, but not her :)
```suggestion
```

include/alpaka/mem/fence/MemFenceCpuSerial.hpp
```diff
@@ -41,15 +41,11 @@ namespace alpaka
...
-
-                static std::atomic<int> dummy{42};
-
-                /* ISO C++ fences are only clearly defined if there are atomic operations surrounding them. So we use
-                 * these dummy operations to ensure this.*/
-                auto x = dummy.load(std::memory_order_relaxed);
+                /* Enable device fences because we may want to synchronize with other (serial) kernels.
+-                * ISO C++ fences are only clearly defined if there are atomic operations surrounding them.
+                 * When using the fence, ensure that this condition is met in the code.
+                 */
```
bernhardmgruber:
```suggestion
                // Enable device fences because we may want to synchronize with other (serial) kernels.
```

include/alpaka/mem/fence/MemFenceGenericSycl.hpp
```diff
@@ -63,17 +49,14 @@ namespace alpaka::trait
     template<typename TMemScope>
     struct MemFence<MemFenceGenericSycl, TMemScope>
     {
-        static auto mem_fence(MemFenceGenericSycl const& fence, TMemScope const&)
+        static auto mem_fence(MemFenceGenericSycl const&, TMemScope const&)
         {
+            /* Fences implementation in SYCL follow the ISO C++ fences logic, so they are only clearly defined if
+             * there are atomic operations surrounding them. When using the fence, ensure that this condition is met
+             * in the code.
+             */
```
bernhardmgruber: Maybe just:
```suggestion
            // Fences implementation in SYCL follow the ISO C++ fences logic
```
Or remove the comment entirely.

bernhardmgruber (commented):
Looks mostly good. I would remove the comments on fences requiring atomics somewhere else. This should go to the documentation of the API, maybe here: https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/mem/fence/Traits.hpp#L51

bernhardmgruber (approved):

j-stephan (approved):

j-stephan:
There are outages in the GitLab CI right now. I'll restart the currently failing jobs once our IT department fixed the issue(s).

PR #1467, opened 2021-11-15
Add `alpaka_RELOCATABLE_DEVICE_CODE` option

j-stephan:
Fixes #1292. Needs to be backported to 0.8 after merging.

PR #2072, opened 2023-08-09
Simplify `alpaka_add_{executable,library}`

j-stephan:
This PR refactors and modernizes the `alpaka_add_{executable,library}` CMake macros. A lot of special handling is no longer necessary since CMake introduced native support for CUDA.

It also serves as a prerequisite for #1467.


bernhardmgruber (approved):
Great work!

Do we have a test for an alpaka library by any chance? If not, we should open an issue.

j-stephan:
> Do we have a test for an alpaka library by any chance? If not, we should open an issue.

We don't. We are also not using `alpaka_add_library` anywhere in this repository. Maybe PIConGPU or cupla make use of it?

PR #2076, opened 2023-08-10
Reduce console output

bernhardmgruber:
A few tests and examples are a bit noisy when inspecting the CI logs, so here is a small reduction of their output.

### DIFF
--- a/example/helloWorld/src/helloWorld.cpp
+++ b/example/helloWorld/src/helloWorld.cpp
@@ -142,8 +142,8 @@ auto main() -> int
     // Thus, a thread can process data element size wise with its
     // vector processing unit.
     using Vec = alpaka::Vec<Dim, Idx>;
-    Vec const elementsPerThread(Vec::all(static_cast<Idx>(1)));
-    Vec const threadsPerGrid(Vec::all(static_cast<Idx>(8)));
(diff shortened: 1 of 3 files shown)


example/helloWorld/src/helloWorld.cpp
```diff
@@ -142,8 +142,8 @@ auto main() -> int
     // Thus, a thread can process data element size wise with its
     // vector processing unit.
     using Vec = alpaka::Vec<Dim, Idx>;
-    Vec const elementsPerThread(Vec::all(static_cast<Idx>(1)));
-    Vec const threadsPerGrid(Vec::all(static_cast<Idx>(8)));
+    auto const elementsPerThread = Vec::all(static_cast<Idx>(1));
+    auto const threadsPerGrid = Vec{4, 2, 4};
```
j-stephan, line 146: Why the change from `(8,8,8)` to `(4,2,4)` here?

bernhardmgruber: To reduce the number of `printf` spamming the console from 512 to 32. I picked 32 to fill a warp, but it should actually not matter and could be even less.

j-stephan (approved):

Issue #2077, opened 2023-08-10
Update clang-format


bernhardmgruber:
Alpaka currently uses clang-format-14. clang-format-17 is about to be released shortly, so I think we should upgrade to version 15, or even 16.

j-stephan:
This will require voting on the new rules introduced in clang-format-16 and -17. I'll look into it.

j-stephan:
I have created a PR for all changes up to clang-format-16. See [here](https://github.com/ComputationalRadiationPhysics/contributing/pull/66).

bernhardmgruber:
So it turns out that due to the bad formatting introduces in PR #2080, nvcc (or the host compiler?) yield a LOT of warnings and the compilation errors are not visible anymore in the CI, because the log exceeds a limit:) See e.g.: https://gitlab.com/hzdr/crp/alpaka/-/jobs/4929025306.

We discussed in the VC yesterday that we may want to delay the clang-format upgrade until the last moment before the next release. But given the problem I see now in the CI, it may be worthwhile to upgrade sooner.

j-stephan:
No, those have been present before. They come from us adding `-pedantic` in our test warnings which clashes with nvcc's internal source reformatting before the CUDA compilation pass.

bernhardmgruber:
> No, those have been present before. They come from us adding `-pedantic` in our test warnings which clashes with nvcc's internal source reformatting before the CUDA compilation pass.

Oh really? I have only observed this now after #2080 was merged. In that case though, can we suppress that warning? As I said, I can no longer read the error messages, because the log explodes :)

PR #2080, opened 2023-08-16
Simplify extent and offset APIs

### LINKED ISSUE #2079: Squash extent/offset/pitch traits
Alpaka's extent traits take the requested dimension as template argument:
```c++
auto e = getExtent<1>(buffer);
```
This creates unnecessary friction as dealing with template paramters is less convenient in user code than ordinary function parameters:
```c++
auto e = getExtent(1, buffer);
```
Arguably, the latter should be similarly efficient than the first version, since the compiler still sees the constant and `getExtent` can be fully inlined.

Furthermore, since extents are not huge objects, just returning all extents as an array could just squash a whole set of APIs into one:
```c++
auto e = getExtents(buffer)[1];
auto es = getExtents(); // all extents
auto [depth, height, width] = getExtents(); // how beautiful!
```
The latter function already exists as `getExtentsVec`, but is not a trait. It just builds upon `getExtent<I>(...)`.

I propose to replace all extent functions by a single, customizable (trait) `getExtents(thing) -> Vec/Array`.

The same logic can be applied to offset and pitch traits as well.

bernhardmgruber:
* Add getExtents and GetExtents returning extents as alpaka::Vec
* Deprecate getExtent, GetExtent and getExtentVec
* Add getOffsets and GetOffsets returning offsets as alpaka::Vec
* Deprecate getOffset, GetOffset and getOffsetVec

Addresses part of #2079. This is a breaking change.


bernhardmgruber:
Yeah, we are going to need that clang-format (#2077) upgrade for this PR.

include/alpaka/mem/view/ViewPlainPtr.hpp
```diff
@@ -251,13 +245,13 @@ namespace alpaka
...
-        struct GetOffset<TIdxIntegralConst, ViewPlainPtr<TDev, TElem, TDim, TIdx>>
+        template<typename TDev, typename TElem, typename TDim, typename TIdx>
+        struct GetOffsets<ViewPlainPtr<TDev, TElem, TDim, TIdx>>
         {
             ALPAKA_FN_HOST
-            static auto getOffset(ViewPlainPtr<TDev, TElem, TDim, TIdx> const&) -> TIdx
+            auto operator()(ViewPlainPtr<TDev, TElem, TDim, TIdx> const&) const -> Vec<TDim, TIdx>
             {
-                return 0u;
+                return {};
```
psychocoderHPC: IMO should be `return {0}` to not rely on any default initialization. Same as in [ViewStdArray.hpp](https://github.com/alpaka-group/alpaka/pull/2080/files#diff-7a7202f988bb3302cc16ddcbead90f348b1107b34b83968c3749f0134158a806)

psychocoderHPC (changes requested):

include/alpaka/mem/view/ViewPlainPtr.hpp
bernhardmgruber: But wouldn't this be a compile error for a zero-dimensional view? And `return {0}` would suggest that the returned vector is 1-dimensional. Even though `{0}` would also initialize any other excess dimensions with `0`. My version covers all dimension.

bernhardmgruber: I replaced it by `Vec<TDim, TIdx>::zeros()`.

bernhardmgruber:
> Yeah, we are going to need that clang-format (#2077) upgrade for this PR.

We discussed this in the VC today and @psychocoderHPC proposed, with no objection, to merge the clang-format upgrade later.

include/alpaka/mem/buf/BufCpu.hpp
```diff
@@ -288,12 +285,12 @@ namespace alpaka
...
-        template<typename TIdxIntegralConst, typename TElem, typename TDim, typename TIdx>
-        struct GetOffset<TIdxIntegralConst, BufCpu<TElem, TDim, TIdx>>
+        template<typename TElem, typename TDim, typename TIdx>
+        struct GetOffsets<BufCpu<TElem, TDim, TIdx>>
         {
-            ALPAKA_FN_HOST static auto getOffset(BufCpu<TElem, TDim, TIdx> const&) -> TIdx
+            ALPAKA_FN_HOST auto operator()(BufCpu<TElem, TDim, TIdx> const&) const -> Vec<TDim, TIdx>
             {
-                return 0u;
+                return {};
```
psychocoderHPC: Sry missed htis in the first review:  Should be `Vec<TDim, TIdx>::zeros()`

include/alpaka/mem/buf/BufGenericSycl.hpp
```diff
@@ -208,12 +206,12 @@ namespace alpaka::trait
...
-    template<typename TIdxIntegralConst, typename TElem, typename TDim, typename TIdx, typename TPlatform>
-    struct GetOffset<TIdxIntegralConst, BufGenericSycl<TElem, TDim, TIdx, TPlatform>>
+    template<typename TElem, typename TDim, typename TIdx, typename TPlatform>
+    struct GetOffsets<BufGenericSycl<TElem, TDim, TIdx, TPlatform>>
     {
-        static auto getOffset(BufGenericSycl<TElem, TDim, TIdx, TPlatform> const&) -> TIdx
+        auto operator()(BufGenericSycl<TElem, TDim, TIdx, TPlatform> const&) const -> Vec<TDim, TIdx>
         {
-            return 0u;
+            return {};
```
psychocoderHPC: Should be `Vec<TDim, TIdx>::zeros()`

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -331,12 +333,13 @@ namespace alpaka
...
-        struct GetOffset<TIdxIntegralConst, BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>
+        template<typename TApi, typename TElem, typename TDim, typename TIdx>
+        struct GetOffsets<BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>
         {
-            ALPAKA_FN_HOST static auto getOffset(BufUniformCudaHipRt<TApi, TElem, TDim, TIdx> const&) -> TIdx
+            ALPAKA_FN_HOST auto operator()(BufUniformCudaHipRt<TApi, TElem, TDim, TIdx> const&) const
+                -> Vec<TDim, TIdx>
             {
-                return 0u;
+                return {};
```
psychocoderHPC: Should be `Vec<TDim, TIdx>::zeros()`

psychocoderHPC (changes requested):
Sry missed some places in the first review which should be adjusted.

psychocoderHPC (approved):

include/alpaka/extent/Traits.hpp
```diff
@@ -22,121 +23,183 @@ namespace alpaka
...
-        };
-
-        //! The extent set trait.
-        template<typename TIdxIntegralConst, typename TExtent, typename TExtentVal, typename TSfinae = void>
-        struct SetExtent;
+        struct [[deprecated("Specialize GetExtents instead")]] GetExtent{
+            ALPAKA_NO_HOST_ACC_WARNING ALPAKA_FN_HOST_ACC static auto getExtent(TExtent const&)
+                ->Idx<TExtent>{return static_cast<Idx<TExtent>>(1);
     } // namespace trait
+}; // namespace alpaka
```
psychocoderHPC, line 30: This comment and the comment above is wrong!

Please remove the comments.

include/alpaka/offset/Traits.hpp
```diff
@@ -19,100 +20,168 @@ namespace alpaka
...
-        };
-
-        //! The x offset set trait.
-        template<typename TIdx, typename TOffsets, typename TOffset, typename TSfinae = void>
-        struct SetOffset;
+        struct [[deprecated("Specialize GetOffsets instead")]] GetOffset{
+            ALPAKA_NO_HOST_ACC_WARNING ALPAKA_FN_HOST_ACC static auto getOffset(TOffsets const&)
+                ->Idx<TOffsets>{return static_cast<Idx<TOffsets>>(0);
     } // namespace trait
+}; // namespace alpaka
```
psychocoderHPC, line 27: This comment and the comment above is wrong, please remove the comment.

psychocoderHPC (dismissed):

include/alpaka/extent/Traits.hpp
```diff
@@ -22,121 +23,183 @@ namespace alpaka
...
-            ALPAKA_FN_HOST_ACC static auto getExtent(TExtent const&) -> Idx<TExtent>
-            {
-                return static_cast<Idx<TExtent>>(1);
-            }
-        };
-
-        //! The extent set trait.
-        template<typename TIdxIntegralConst, typename TExtent, typename TExtentVal, typename TSfinae = void>
-        struct SetExtent;
+        struct [[deprecated("Specialize GetExtents instead")]] GetExtent{
```
j-stephan, line 26: Oh dear, I see what you mean...

include/alpaka/extent/Traits.hpp
j-stephan: Maybe squashing this segment to `namespace alpaka::trait` improves things a bit...

include/alpaka/extent/Traits.hpp
```diff
@@ -22,121 +23,183 @@ namespace alpaka
...
+//! \return The extent in the given dimension.
+ALPAKA_NO_HOST_ACC_WARNING
+template<std::size_t Tidx, typename TExtent>
+[[deprecated("use getExtents(extent)[Tidx] instead")]] ALPAKA_FN_HOST_ACC auto getExtent(
+    TExtent const& extent = TExtent()) -> Idx<TExtent>
+{
+#if BOOST_COMP_CLANG || BOOST_COMP_GNUC
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored "-Wdeprecated-declarations"
+#endif
```
j-stephan, line 50: What is this? We are using `[[deprecated]]` but are turning off the related warning?

include/alpaka/extent/Traits.hpp
```diff
@@ -22,121 +23,183 @@ namespace alpaka
...
-        };
-        //! The unsigned integral width set trait specialization.
-        template<typename TExtent, typename TExtentVal>
-        struct SetExtent<DimInt<0u>, TExtent, TExtentVal, std::enable_if_t<std::is_integral_v<TExtent>>>
+            return Vec{i};
+        }
+    };
+    //! The unsigned integral width set trait specialization.
+    template<typename TExtent, typename TExtentVal>
+    struct SetExtent<DimInt<0u>, TExtent, TExtentVal, std::enable_if_t<std::is_integral_v<TExtent>>> // WTF?
```
j-stephan: Why the `WTF` here?

include/alpaka/offset/Traits.hpp
j-stephan: There is also another namespace squasing opportunity here :-)

j-stephan (changes requested):

include/alpaka/extent/Traits.hpp
bernhardmgruber: Exactly.

include/alpaka/extent/Traits.hpp
bernhardmgruber: I have put a deprecation annotation on the function and its trait. Because function uses the trait, the warning needs to be disabled here. Otherwise, the intent of the warning is to tell users, that they should switch to the new API.

Alternatively, we can delete the old functions entirely.

include/alpaka/extent/Traits.hpp
bernhardmgruber: Because I question this specialization in its entirety. Good you saw that, let's have a discussion.

What is the extent of an `int`? Is it `0`? Is it `1`? What if the `int` holds the value `42`, is `42` the extent of the `int` then?

bernhardmgruber:
@j-stephan:

> Maybe squashing this segment to namespace alpaka::trait improves things a bit...

> There is also another namespace squasing opportunity here :-)

No there isn't. clang-format puts the comments there wrongly. Those braces do not close a namespace, but a struct. This is fixed in newer versions of clang-format.

include/alpaka/offset/Traits.hpp
bernhardmgruber: @psychocoderHPC, clang-format adds those comments. If I remove them, the formatting fails. This is solved in newer versions of clang-format.

include/alpaka/extent/Traits.hpp
fwyzard: > What is the extent of an `int`? Is it `0`?

yes

bernhardmgruber: > > What is the extent of an `int`? Is it `0`?
> 
> yes

Well, too bad! alpaka defines it (since at least 2015) to be the int's value :)

include/alpaka/extent/Traits.hpp
bernhardmgruber: @j-stephan I don't know what you mean with your thumbs up. Keep deprecated functions or remove them?

j-stephan: Oh, sorry. This was just in reference to your explanation. Keep the deprecated version for now, we don't want to immediately break things on the user side.

include/alpaka/extent/Traits.hpp
j-stephan: Mh, I don't see a compelling reason to keep this specialization. Do things break when removing it?

bernhardmgruber: I had a quick look if we can take the specializations for integrals away, but they are way too deeply used inside alpaka. workdivs, buffer allocation, etc. All that uses the extents API to work with integer sizes etc.

j-stephan: Okay, that is what I was afraid of - some arcane usage. So we are also relying on this functionality of returning the integer value?

bernhardmgruber: I removed the `WTF` comment, and will leave extents for integers as they are.

bernhardmgruber: @j-stephan totally! `alpaka::allocBuf(.., 42)` uses that to turn `42` into a `Vec` internally. You could refactor this but you would have to touch a few places.

j-stephan (approved):

PR #2081, opened 2023-08-17
Disable recursive macro expansion warning for icpx

bernhardmgruber:
This silences warnings with icpx inside the `ALPAKA_CHECK` and `printf` macros.


j-stephan (approved):
Thanks!

PR #2087, opened 2023-08-22
Remove setExtent and setOffset APIs

bernhardmgruber:
These APIs are not used anywhere within our repository and are also only specialized for CUDA any SYCL built-in types, as well as `Vec`.

If any user uses those APIs, it's a breaking change for them.

Relates to #2079.

### DIFF
--- a/include/alpaka/core/CudaHipCommon.hpp
+++ b/include/alpaka/core/CudaHipCommon.hpp
@@ -137,62 +137,6 @@ namespace alpaka
             }
         };
 
-        //! The CUDA/HIP vectors extent set trait specialization.
-        template<typename TExtent, typename TExtentVal>
-        struct SetExtent<
-            DimInt<Dim<TExtent>::value - 1u>,
-            TExtent,
-            TExtentVal,
-            std::enable_if_t<alpaka::detail::isCudaHipBuiltInType<TExtent> && (Dim<TExtent>::value >= 1)>>
-        {
-            ALPAKA_NO_HOST_ACC_WARNING
-            ALPAKA_FN_HOST_ACC static auto setExtent(TExtent const& extent, TExtentVal const& extentVal) -> void
-            {
-                extent.x = extentVal;
-            }
-        };
-        //! The CUDA/HIP vectors extent set trait specialization.
-        template<typename TExtent, typename TExtentVal>
-        struct SetExtent<
-            DimInt<Dim<TExtent>::value - 2u>,
-            TExtent,
-            TExtentVal,
-            std::enable_if_t<alpaka::detail::isCudaHipBuiltInType<TExtent> && (Dim<TExtent>::value >= 2)>>
-        {
-            ALPAKA_NO_HOST_ACC_WARNING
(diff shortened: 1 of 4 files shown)


j-stephan:
> If any user uses those APIs, it's a breaking change for them.

I think we should discuss this first in the next alpaka meeting before introducing this change.

bernhardmgruber:
> > If any user uses those APIs, it's a breaking change for them.
> 
> I think we should discuss this first in the next alpaka meeting before introducing this change.

Alright. What question do you hope to get answered?

j-stephan:
Whether to remove it entirely or just deprecate the API.

bernhardmgruber:
> Whether to remove it entirely or just deprecate the API.

Yeah, that's a good question. Let's hear what everybody thinks!

psychocoderHPC:
We can not simply remove `SetExtent` or `GetExtent`. These traits are allow using non alpaka types e.g. a CUDA `uin3`, ...

example can be found in cupla:

https://github.com/alpaka-group/cupla/blob/7dc3d027b6554fc8336971182604d77c73c45816/include/cupla/datatypes/uint.hpp#L90-L130

I am fine if the interface is changing as long as we are able to support non-alpaka types for offsets and extents.

psychocoderHPC:
It could be that we do not use `SetOffset` in alpaka, in that case removing it should also not affect dependent projects.

bernhardmgruber:
@psychocoderHPC could you please point me to a place in cuple where you actually use `alpaka::setOffset` or `alpaka::setExtent`?

psychocoderHPC:
> @psychocoderHPC could you please point me to a place in cuple where you actually use `alpaka::setOffset` or `alpaka::setExtent`?

We do not use it but specialize it that our own type can be passed to alpaka to define the workdiff extents, ...
but from a short grep it looks like alpaka is not using `setExtent` anymore.
So there is no problem removing it.
It looks like the bigger problem would be if we can not define `GetOffset` anymore because there we need to permutate the access for alpaka because alpaka `Vec` order is (z,y,x) and cupla/cuda uses `x,y,z`. The trait was originally made to support this permutation.
I lost the overview about the refactorings in alpaka so I need to test next week if we broke it.

https://github.com/alpaka-group/cupla/blob/7dc3d027b6554fc8336971182604d77c73c45816/include/cupla/datatypes/uint.hpp#L90-L130

bernhardmgruber:
> > @psychocoderHPC could you please point me to a place in cuple where you actually use `alpaka::setOffset` or `alpaka::setExtent`?
> 
> We do not use it but specialize it that our own type can be passed to alpaka to define the workdiff extents, ...
> but from a short grep it looks like alpaka is not using `setExtent` anymore. So there is no problem removing it.

Ok great, so let's remove it, or at least mark it as deprecated!

> It looks like the bigger problem would be if we can not define `GetOffset` anymore ...

I am not talking about removing `getOffset`. This is needed for all the views. This PR only suggests to remove `setOffset`.

j-stephan (approved):

PR #2090, opened 2023-08-23
Unify ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK macros

bernhardmgruber:
This PR unifies the definition of `ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK**` macros across MSVC and other compilers.

It also fixes an unrelated warning by clang.

### DIFF
--- a/include/alpaka/core/UniformCudaHip.hpp
+++ b/include/alpaka/core/UniformCudaHip.hpp
@@ -9,7 +9,7 @@
 #include "alpaka/core/Cuda.hpp"
 #include "alpaka/core/Hip.hpp"
 
-#include <array>
+#include <initializer_list>
 #include <stdexcept>
 #include <string>
 #include <tuple>
@@ -47,23 +47,19 @@ namespace alpaka::uniform_cuda_hip::detail
     }
 
     //! CUDA/HIP runtime API error checking with log and exception, ignoring specific error values
-    // NOTE: All ignored errors have to be convertible to TApi::Error_t.
-    template<typename TApi, bool TThrow, typename... TErrors>
+    template<typename TApi, bool TThrow>
     ALPAKA_FN_HOST inline void rtCheckIgnore(
         typename TApi::Error_t const& error,
         char const* cmd,
         char const* file,
         int const& line,
-        TErrors&&... ignoredErrorCodes) noexcept(!TThrow)
+        std::initializer_list<typename TApi::Error_t> ignoredErrorCodes) noexcept(!TThrow)
     {
         if(error != TApi::success)
         {
-            std::array<typename TApi::Error_t, sizeof...(ignoredErrorCodes)> const aIgnoredErrorCodes{
-                {ignoredErrorCodes...}};
-
             // If the error code is not one of the ignored ones.
-            if(std::find(std::cbegin(aIgnoredErrorCodes), std::cend(aIgnoredErrorCodes), error)
-               == std::cend(aIgnoredErrorCodes))
+            if(std::find(std::cbegin(ignoredErrorCodes), std::cend(ignoredErrorCodes), error)
+               == std::cend(ignoredErrorCodes))
             {
                 rtCheck<TApi, TThrow>(error, ("'" + std::string(cmd) + "' returned error ").c_str(), file, line);
             }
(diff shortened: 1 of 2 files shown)


psychocoderHPC (dismissed):

j-stephan:
We disabled the MSVC+CUDA checks a while back. Did you verify that this still compiles for this combination?

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -85,89 +81,53 @@ namespace alpaka::uniform_cuda_hip::detail
...
+                __FILE__,                                                                                             \
+                __LINE__,                                                                                             \
+                {__VA_ARGS__});                                                                                       \
+        } while(0)
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic pop
 #    endif
 
 //! CUDA/HIP runtime error checking with log and exception.
 #    define ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(cmd) ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE(cmd)
```
psychocoderHPC: handle clang diagnostics should not be required if we change this to 

```
#define ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(cmd) ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE(cmd,)
# define ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(cmd) 
 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE_NOEXCEPT(cmd, )
```

If the user is calling `ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE()` or `ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_IGNORE_NOEXCEPT` without any ignore list the warning will be shown again but this is nothing alpaka needs to handle.
If the user do not like to ignore errors the macro without `*_IGNORE_*` should be used.

psychocoderHPC: I am fine with the current PR but would like to point out that we most likely do not need to handle compiler diagnostics here.

bernhardmgruber:
> We disabled the MSVC+CUDA checks a while back. Did you verify that this still compiles for this combination?

Absolutely not :D I work mostly on my Ubuntu workstation these days, so I rarely test with MSVC.

Also: I really think Google got it right with their ["Beyoncé Rule"](https://abseil.io/resources/swe-book/html/ch11.html#:~:text=We%20have%20a%20name%20for,changes%20across%20the%20entire%20codebase.), saying: “If you liked it, then you shoulda put a test on it.”.
So, given there is no test, I assume you didn't like MSVC+CUDA support :)

j-stephan:
We disabled the tests because of a nvcc bug in combination with Catch2.

include/alpaka/core/UniformCudaHip.hpp
bernhardmgruber: I just retained the warning suppression since it was there before. But your suggestion with `...IGNORE(cmd,)` is gold! That got rid of the warning entirely!

bernhardmgruber:
> We disabled the tests because of a nvcc bug in combination with Catch2.

Right and I remember that we played some time to get a working version of MSVC+nvcc+Catch2.

However, I think my point still stands. If we don't have test coverage for a specific combination and we continue to accept changes, like we did for the last months, then any PR could have broken that combination. My PR is not special in that regard.

j-stephan:
> My PR is not special in that regard.

It is (IMO) because you are removing special MSVC+CUDA handling. That makes it much more likely to break the combination than a random code change.

j-stephan (changes requested):
Please add `/Zc:preprocessor` to the general MSVC build flags:

```cmake
# Make MSVC's preprocessor more conforming to other compilers' preprocessors
alpaka_set_compiler_options(HOST_DEVICE target alpaka $<$AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CXX>>:/Zc:preprocessor>
                                                      $<$AND:$<CXX_COMPILER_ID:MSVC>,$<COMPILE_LANGUAGE:CUDA>>:-Xcompiler=/Zc:preprocessor>)
```

See the [description](https://learn.microsoft.com/en-us/cpp/build/reference/zc-preprocessor?view=msvc-170). At least on godbolt this makes `##__VA_ARGS__` work the same way on all compilers so we wouldn't need the special handling for MSVC.

Edit: Somewhere around https://github.com/alpaka-group/alpaka/blob/873b1685ee2f6c58db4a48e5af61a79ce5377f3a/cmake/alpakaCommon.cmake#L196.

include/alpaka/mem/buf/uniformCudaHip/Set.hpp
```diff
@@ -94,8 +94,6 @@ namespace alpaka
             template<typename TQueue>
             auto enqueue(TQueue& queue) const -> void
             {
-                using Idx = Idx<TExtent>;
-
```
j-stephan, line 98: Is this a related change?

bernhardmgruber: No, as documented in the PR description and the separate commit.

bernhardmgruber:
> Please add `/Zc:preprocessor` to the general MSVC build flags:
>  ...

This is not needed for this PR because I refactored the code to not need the comma eating behavior of `##__VA_ARGS__`. Check the changeset, there is just `__VA_ARGS__` nows.

But feel free to propose a new PR which adds `/Zc:preprocessor`! But then, aren't we maybe enforcing too much for alpaka users? This basically forces a non-default, different (but conforming) preprocessor implementation on all MSVC users.

j-stephan:
Ah, I missed that. In that case let's hope this doesn't break MSVC + CUDA.

j-stephan (approved):

PR #2091, opened 2023-08-24
extend inline macro documentation

psychocoderHPC:
The inline macro should be left of any keyword because depending of the alpaka backend std attribute specifiers will be used.

### DIFF
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -14,10 +14,13 @@
 
 //! All functions that can be used on an accelerator have to be attributed with ALPAKA_FN_ACC or ALPAKA_FN_HOST_ACC.
 //!
+//! \code{.cpp}
 //! Usage:
 //! ALPAKA_FN_ACC
 //! auto add(std::int32_t a, std::int32_t b)
 //! -> std::int32_t;
+//! \endcode
+//! @{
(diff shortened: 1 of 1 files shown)


include/alpaka/core/Common.hpp
```diff
@@ -14,10 +14,13 @@
 
 //! All functions that can be used on an accelerator have to be attributed with ALPAKA_FN_ACC or ALPAKA_FN_HOST_ACC.
 //!
+//! @code{.cpp}
```
psychocoderHPC: Sry for sneaking in this change, I first commented by accident on this part and realized that only ALPAKAFN_INLINE is affected but would like to keep this change due to the correct usage of doxygen.

include/alpaka/core/Common.hpp
```diff
@@ -74,6 +78,8 @@
 #endif
 
 //! Macro defining the inline function attribute.
+//!
+//! Macro should stay left hand side of keyword e.g. 'static', 'constexpr', 'explicit' or the return type.
```
j-stephan:
```suggestion
//! The macro should stay on the left hand side of keywords, e.g. 'static', 'constexpr', 'explicit' or the return type.
```

j-stephan (changes requested):

bernhardmgruber (approved):

j-stephan (approved):

PR #2092, opened 2023-08-25
Simplify pitch API

### LINKED ISSUE #2079: Squash extent/offset/pitch traits

bernhardmgruber:
* Add getPitchesInBytes and GetPitchesInBytes returning extents as alpaka::Vec.
* Deprecate getPitchBytes, GetPitchBytes and getPitchBytesVec.
* Includes 2 unrelated refactorings and an improvement to the reporting if the bytes of a view were correctly set.

Fixes: #2079

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -215,10 +215,10 @@ auto main() -> int
     // padding between rows/planes of multidimensional memory allocations.
     // Therefore the pitch (distance between consecutive rows/planes) may be
     // greater than the space required for the data.
-    Idx const deviceBuffer1Pitch(alpaka::getPitchBytes<2u>(deviceBuffer1) / sizeof(Data));
-    Idx const deviceBuffer2Pitch(alpaka::getPitchBytes<2u>(deviceBuffer2) / sizeof(Data));
-    Idx const hostBuffer1Pitch(alpaka::getPitchBytes<2u>(hostBuffer) / sizeof(Data));
-    Idx const hostViewPlainPtrPitch(alpaka::getPitchBytes<2u>(hostViewPlainPtr) / sizeof(Data));
+    Idx const deviceBuffer1Pitch(alpaka::getPitchesInBytes(deviceBuffer1)[2] / sizeof(Data));
+    Idx const deviceBuffer2Pitch(alpaka::getPitchesInBytes(deviceBuffer2)[2] / sizeof(Data));
+    Idx const hostBuffer1Pitch(alpaka::getPitchesInBytes(hostBuffer)[2] / sizeof(Data));
+    Idx const hostViewPlainPtrPitch(alpaka::getPitchesInBytes(hostViewPlainPtr)[2] / sizeof(Data));
 
     // Test device Buffer
     //
--- a/example/randomCells2D/src/randomCells2D.cpp
+++ b/example/randomCells2D/src/randomCells2D.cpp
@@ -201,16 +201,16 @@ auto main() -> int
     RandomEngineVector<Acc>* const ptrBufAccRandV{alpaka::getPtrNative(bufAccRandV)};
 
     InitRandomKernel initRandomKernel;
-    auto pitchBufAccRandS = alpaka::getPitchBytes<1u>(bufAccRandS);
+    auto pitchBufAccRandS = alpaka::getPitchesInBytes(bufAccRandS)[1];
     alpaka::exec<Acc>(queue, workdiv, initRandomKernel, extent, ptrBufAccRandS, pitchBufAccRandS);
     alpaka::wait(queue);
 
-    auto pitchBufAccRandV = alpaka::getPitchBytes<1u>(bufAccRandV);
+    auto pitchBufAccRandV = alpaka::getPitchesInBytes(bufAccRandV)[1];
     alpaka::exec<Acc>(queue, workdiv, initRandomKernel, extent, ptrBufAccRandV, pitchBufAccRandV);
     alpaka::wait(queue);
 
-    auto pitchHostS = alpaka::getPitchBytes<1u>(bufHostS);
-    auto pitchHostV = alpaka::getPitchBytes<1u>(bufHostV);
+    auto pitchHostS = alpaka::getPitchesInBytes(bufHostS)[1];
+    auto pitchHostV = alpaka::getPitchesInBytes(bufHostV)[1];
 
     for(Idx y = 0; y < numY; ++y)
     {
@@ -221,7 +221,7 @@ auto main() -> int
         }
     }
 
-    auto pitchBufAccS = alpaka::getPitchBytes<1u>(bufAccS);
+    auto pitchBufAccS = alpaka::getPitchesInBytes(bufAccS)[1];
     alpaka::memcpy(queue, bufAccS, bufHostS);
     RunTimestepKernelSingle runTimestepKernelSingle;
     alpaka::exec<Acc>(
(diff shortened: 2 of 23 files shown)


include/alpaka/mem/view/Traits.hpp
```diff
@@ -138,9 +180,24 @@ namespace alpaka
...
+#    pragma GCC diagnostic pop
+#endif
+    }
+
+    //! \return The pitches in bytes. This is the distance in bytes between two consecutive elements in the given
+    //! dimension.
+    template<typename TView>
+    ALPAKA_FN_HOST auto getPitchesInBytes(TView const& view) -> Vec<Dim<TView>, Idx<TView>>
+    {
+        return trait::GetPitchesInBytes<TView>{}(view);
```
bernhardmgruber, line 200: Let the bikeshedding begin. I am open to name suggestions for this function.

j-stephan: `getByteDistance`?

bernhardmgruber: I would prefer if it had the word `pitch` or `stride` in it. `getByteDistance` sounds too generic to me.

j-stephan: Why not `getBytePitches` or just `getPitches`? Is it important that we clarify in the function name that a pitch contains some number of bytes?

bernhardmgruber: Right, this is similar to what @psychocoderHPC suggested. In principle, `getPitches` is OK.

I got quite confused some time ago, because it was not clear to me that the pitch was expressed in numbers of bytes and not in number elements, `getExtents` for example returnes the number of elements in each dimension. So to make the distinction clear, I preferred `getPitchBytes` in the past.

By now, I have a better understanding of what a pitch is, and in a CUDA, alpaka, picongpu, or image context, it has a clear meaning to me. Still, there is some overlap with the word stride and `std::mdspan` expresses the stride in number of elements. So, defensively, I opted for `getPitchesInBytes`, to make it very clear what is intended. The name is a bit long but clearer than `getBytePitches`, since we are not returning the pitches of a byte, but of a view in bytes :)

I am still ok with naming it `getPitches` if everybody wants so, since that is more consistent with `getExtents` and `getOffsets`. I would then make sure the documentation indicates clearly that the values are in bytes.

psychocoderHPC: I am ok with `getPitches` and `getPitchesInBytes`.

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -215,10 +215,10 @@ auto main() -> int
     // padding between rows/planes of multidimensional memory allocations.
     // Therefore the pitch (distance between consecutive rows/planes) may be
     // greater than the space required for the data.
-    Idx const deviceBuffer1Pitch(alpaka::getPitchBytes<2u>(deviceBuffer1) / sizeof(Data));
-    Idx const deviceBuffer2Pitch(alpaka::getPitchBytes<2u>(deviceBuffer2) / sizeof(Data));
-    Idx const hostBuffer1Pitch(alpaka::getPitchBytes<2u>(hostBuffer) / sizeof(Data));
-    Idx const hostViewPlainPtrPitch(alpaka::getPitchBytes<2u>(hostViewPlainPtr) / sizeof(Data));
+    Idx const deviceBuffer1Pitch(alpaka::getPitchesInBytes(deviceBuffer1)[2] / sizeof(Data));
```
psychocoderHPC, line 218: `deviceBuffer1` is a 3-dimensional buffer and this line calculates the number of elements per row assuming that pitch is a multiple of the element size.
IMO this should be `Idx const deviceBuffer1Pitch(alpaka::getPitchesInBytes(deviceBuffer1)[1] / sizeof(Data));`
because the bytes to jump within the y-dimension is required to calculate number of elements per row.
The value of `getPitchBytes(buffer)[2]` for a 3-dimensional buffer is the size of an element aka how many bytes do I need to add to go to the next column.

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -201,16 +201,16 @@ auto main() -> int
     RandomEngineVector<Acc>* const ptrBufAccRandV{alpaka::getPtrNative(bufAccRandV)};
 
     InitRandomKernel initRandomKernel;
-    auto pitchBufAccRandS = alpaka::getPitchBytes<1u>(bufAccRandS);
+    auto pitchBufAccRandS = alpaka::getPitchesInBytes(bufAccRandS)[1];
```
psychocoderHPC, line 204: `bufAcccRandS` is a 2-diensional buffer therefore this should be `auto pitchBufAccRandS = alpaka::getPitchesInBytes(bufAccRandS)[0];`

psychocoderHPC (changes requested):
I annotated only two lines but from our developer discussion last week I thought the pitch will be defined as the number of bytes to jump within the dimension, so
3-dim: `(y-x-plane-bytes, x-line-bytes, element-bytes)` if so it looks like the changes are wrong, if my definition is wrong than I missed the new definition of `getPitchInBytes()`

psychocoderHPC:
> I annotated only two lines but from our developer discussion last week I thought the pitch will be defined as the number of bytes to jump within the dimension, so 3-dim: `(y-x-plane-bytes, x-line-bytes, element-bytes)` if so it looks like the changes are wrong, if my definition is wrong than I missed the new definition of `getPitchInBytes()`

Ignore this, I have not seen that there is a second PR https://github.com/alpaka-group/alpaka/pull/2093 that is changing the behavior.

include/alpaka/mem/view/Traits.hpp
bernhardmgruber: Any other opinions? I have @j-stephan for `getPitches`, me for `getPitchesInBytes` and @psychocoderHPC for both. I need a tiebreaker.

fwyzard: I like that `getPitchesInBytes` is more explicit and that `getPitches` is shorter...

If "pitch" is universally measured in bytes, I would favour `getPitches`.

The [wikipedia page](https://en.wikipedia.org/wiki/Stride_of_an_array) on array strides mixes up stride _vs_ pitch and elements _vs_ bytes, so it's not very helpful.

According to ChatGPT 3.5
> In a 2D array, the term "pitch" typically refers to the number of elements or bytes between the start of one row and the start of the next row in memory. It's often used in the context of optimizing memory access patterns, especially in languages like C or C++.

Again not helpful.

Given the overall confusion, I would go with `getPitchesInBytes`.

bernhardmgruber: Thx a lot for this analysis :D I would like to stick with `getPitchesInBytes` then.

bernhardmgruber:
The TSAN failure in the `coreTest` seems unreleated, but I don't have time to look into it the next days.

psychocoderHPC (approved):

PR #2093, opened 2023-08-25
Redefine pitches

### LINKED ISSUE #2083: Rework pitches
Pitches in alpaka are, IMO, suprising. Given the extents `{42, 10, 2}`, a 3D buffer of `float` (4 bytes) on the CPU has the following pitches: `{4, 3360, 80, 8}`. Yes, the pitch vector has 4 dimensions and this is actively used in the implementation of `ViewSubView`. 

The documentation of `getPitchBytes` says:

> The pitch in bytes. This is the distance in bytes between two consecutive elements in the given dimension.

Which is clearly wrong. Then, the implementation of `GetPitchBytes` says:

> This is the distance in bytes in the linear memory between two consecutive elements in the next higher dimension (TIdx-1).

Which is actually true (except for the 1 + highest dimension), but contradicts the documentation of `getPitchBytes`.

I strongly think the situation should be fixed. In my understanding, for the extent  `{42, 10, 2}`, the pitch vector should be `{80, 8, 4}`, and thus correspond to the documentation of `getPitchBytes`. However, such a change would be a silent breaking, since user code would still compile but have different behavior. But then, the function is already wrong wrt. to it's documentation, so such a change can also be seen as a fix. Alternativelty, `getPitchBytes` could be deprecated and replaced by a new function, like `getStrideBytes`, which has the correct behavior. Also, we could replace everything by a function `getPitchesInBytes` (see also #2049) that returns the proposed vector.

bernhardmgruber:
This PR shifts the values returned from getPitchesInBytes to be consistent with std::mdspan (except in bytes).

Example: the pitch vector for the extent {42, 10, 2} changes:

Before: {4, 3360, 80, 8}
After: {80, 8, 4}

The new meaning is that the pitch value is the number of bytes to jump from one element to the next in the given dimension.

Fixes: #2083

~~Waiting on #2092 to be merged before rebase.~~

This is a **silent breaking change**. However, to a function that was only introduced a few days ago with #2092.

### DIFF
--- a/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
+++ b/include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
@@ -182,16 +182,18 @@ namespace alpaka
         struct GetPitchesInBytes<BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>
         {
             ALPAKA_FN_HOST auto operator()(BufUniformCudaHipRt<TApi, TElem, TDim, TIdx> const& buf) const
+                -> Vec<TDim, TIdx>
             {
                 Vec<TDim, TIdx> v{};
                 if constexpr(TDim::value > 0)
                 {
+                    v.back() = sizeof(TElem);
                     if constexpr(TDim::value > 1)
-                        v.back() = buf.m_rowPitchInBytes;
-                    else
-                        v.back() = buf.m_extentElements.back() * sizeof(TElem);
-                    for(int i = static_cast<int>(TDim::value) - 2; i >= 0; i--)
-                        v[i] = buf.m_extentElements[i] * v[i + 1];
+                    {
+                        v[TDim::value - 2] = static_cast<TIdx>(buf.m_rowPitchInBytes);
+                        for(TIdx i = TDim::value - 2; i > 0; i--)
+                            v[i - 1] = buf.m_extentElements[i] * v[i];
+                    }
                 }
                 return v;
             }
--- a/include/alpaka/mem/buf/cpu/Set.hpp
+++ b/include/alpaka/mem/buf/cpu/Set.hpp
@@ -23,6 +23,8 @@ namespace alpaka
         template<typename TDim, typename TView, typename TExtent>
         struct TaskSetCpuBase
         {
+            static_assert(TDim::value > 0);
+
             using ExtentSize = Idx<TExtent>;
             using DstSize = Idx<TView>;
             using Elem = alpaka::Elem<TView>;
@@ -31,15 +33,16 @@ namespace alpaka
             TaskSetCpuBase(TViewFwd&& view, std::uint8_t const& byte, TExtent const& extent)
                 : m_byte(byte)
                 , m_extent(getExtents(extent))
-                , m_extentWidthBytes(m_extent[TDim::value - 1u] * static_cast<ExtentSize>(sizeof(Elem)))
+                , m_extentWidthBytes(m_extent.back() * static_cast<ExtentSize>(sizeof(Elem)))
 #if(!defined(NDEBUG)) || (ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL)
                 , m_dstExtent(getExtents(view))
 #endif
                 , m_dstPitchBytes(getPitchesInBytes(view))
                 , m_dstMemNative(reinterpret_cast<std::uint8_t*>(getPtrNative(view)))
             {
-                ALPAKA_ASSERT((castVec<DstSize>(m_extent) <= m_dstExtent).foldrAll(std::logical_or<bool>()));
-                ALPAKA_ASSERT(m_extentWidthBytes <= m_dstPitchBytes[TDim::value - 1u]);
+                ALPAKA_ASSERT((castVec<DstSize>(m_extent) <= m_dstExtent).all());
+                if constexpr(TDim::value > 1)
+                    ALPAKA_ASSERT(m_extentWidthBytes <= m_dstPitchBytes[TDim::value - 2]);
             }
 
 #if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
@@ -80,9 +83,8 @@ namespace alpaka
 #endif
                 // [z, y, x] -> [z, y] because all elements with the innermost x dimension are handled within one
                 // iteration.
-                Vec<DimMin1, ExtentSize> const extentWithoutInnermost(subVecBegin<DimMin1>(this->m_extent));
-                // [z, y, x] -> [y, x] because the z pitch (the full idx of the buffer) is not required.
-                Vec<DimMin1, DstSize> const dstPitchBytesWithoutOutmost(subVecEnd<DimMin1>(this->m_dstPitchBytes));
+                Vec<DimMin1, ExtentSize> const extentWithoutInnermost = subVecBegin<DimMin1>(this->m_extent);
+                Vec<DimMin1, DstSize> const dstPitchBytesWithoutOutmost = subVecBegin<DimMin1>(this->m_dstPitchBytes);
 
                 if(static_cast<std::size_t>(this->m_extent.prod()) != 0u)
                 {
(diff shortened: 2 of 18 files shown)


include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -32,29 +32,44 @@ namespace alpaka
...
+        {
+            std::size_t m_rowPitchInBytes;
+        };
+    } // namespace detail
+
     //! The CUDA/HIP memory buffer.
     template<typename TApi, typename TElem, typename TDim, typename TIdx>
-    class BufUniformCudaHipRt : public internal::ViewAccessOps<BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>
+    struct BufUniformCudaHipRt
+        : detail::PitchHolder<TDim>
```
fwyzard: why a new base class instead of a data member ?

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -32,29 +32,44 @@ namespace alpaka
...
         ALPAKA_FN_HOST BufUniformCudaHipRt(
             DevUniformCudaHipRt<TApi> const& dev,
             TElem* const pMem,
             Deleter deleter,
-            std::size_t pitchBytes,
-            TExtent const& extent)
-            : m_dev(dev)
-            , m_extentElements(getExtentVecEnd<TDim>(extent))
+            TExtent const& extent,
+            std::size_t pitchBytes)
```
fwyzard: why the swap in the order of the arguments ?

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -165,26 +178,23 @@ namespace alpaka
             }
         };
 
-        //! The BufUniformCudaHipRt pitch get trait specialization.
         template<typename TApi, typename TElem, typename TDim, typename TIdx>
-        struct GetPitchBytes<DimInt<TDim::value - 1u>, BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>
+        struct GetPitchesInBytes<BufUniformCudaHipRt<TApi, TElem, TDim, TIdx>>
         {
-            ALPAKA_FN_HOST static auto getPitchBytes(BufUniformCudaHipRt<TApi, TElem, TDim, TIdx> const& buf) -> TIdx
+            ALPAKA_FN_HOST auto operator()(BufUniformCudaHipRt<TApi, TElem, TDim, TIdx> const& buf) const
```
fwyzard, line 184: can you add back the return type ?

include/alpaka/mem/buf/cpu/Set.hpp
```diff
@@ -35,11 +35,11 @@ namespace alpaka
 #if(!defined(NDEBUG)) || (ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL)
                 , m_dstExtent(getExtents(view))
 #endif
-                , m_dstPitchBytes(getPitchBytesVec(view))
+                , m_dstPitchBytes(getPitchesInBytes(view))
                 , m_dstMemNative(reinterpret_cast<std::uint8_t*>(getPtrNative(view)))
             {
                 ALPAKA_ASSERT((castVec<DstSize>(m_extent) <= m_dstExtent).foldrAll(std::logical_or<bool>()));
-                ALPAKA_ASSERT(m_extentWidthBytes <= m_dstPitchBytes[TDim::value - 1u]);
+                //                ALPAKA_ASSERT(m_extentWidthBytes <= m_dstPitchBytes[TDim::value - 1u]);
```
fwyzard: ?

include/alpaka/mem/view/ViewSubView.hpp
```diff
@@ -127,9 +129,32 @@ namespace alpaka
         }
 
     public:
+        ALPAKA_FN_HOST auto computeNativePtr()
+        {
+#if BOOST_COMP_GNUC
+#    pragma GCC diagnostic push
+#    pragma GCC diagnostic ignored                                                                                    \
+        "-Wcast-align" // "cast from 'std::uint8_t*' to 'TElem*' increases required alignment of target type"
```
fwyzard:
```suggestion
            // "cast from 'std::uint8_t*' to 'TElem*' increases required alignment of target type"
#    pragma GCC diagnostic ignored "-Wcast-align" 
```
?

fwyzard (commented):
why the change from
```c++
struct Action {
  static auto action(...) { ... }
};
```
to
```c++
struct Action {
  auto operator()(...) const { ... }
};
```
?

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
bernhardmgruber: Because the data member is not needed for 0-dim and 1-dim buffers.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
bernhardmgruber: Because I initially thought about having a ctor overload without the `pitch`, but then I decided against it. I can order them back.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
bernhardmgruber: sure!

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
fwyzard: ah, I see

include/alpaka/mem/view/ViewSubView.hpp
bernhardmgruber: thx!

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
bernhardmgruber: I restored the original order.

include/alpaka/mem/buf/cpu/Set.hpp
bernhardmgruber: I readded the check under a guard.

bernhardmgruber:
> why the change from
> 
> ```c++
> struct Action {
>   static auto action(...) { ... }
> };
> ```
> 
> to
> 
> ```c++
> struct Action {
>   auto operator()(...) const { ... }
> };
> ```
> 
> ?

I tried hard to find the corresponding issue, but probably it was just a long discussion on a PR somewhere. I argued for using `operator()` consistently to avoid repreating the name of the functionality in the customization point struct name (trait) and the member function. I think we concluded to move to `operator()` in the future. All math customizations are already using this scheme btw. But the current situation in alpaka is inconsistent.

I don't care too much what we do. `operator()` is what I would prefer, but I won't lose any sleep over something else. What do other people think?

fwyzard:
> I don't care too much what we do. `operator()` is what I would prefer, but I won't lose any sleep over something else. What do other people think?

If the reason is to avoid duplicating the name you could use something like `static auto apply()` for each class.

But I don't really mind, I was only curious about the change.

include/alpaka/mem/buf/BufUniformCudaHipRt.hpp
```diff
@@ -64,8 +64,8 @@ namespace alpaka
             DevUniformCudaHipRt<TApi> const& dev,
             TElem* const pMem,
             Deleter deleter,
-            TExtent const& extent,
-            std::size_t pitchBytes)
+            std::size_t pitchBytes,
+            TExtent const& extent)
```
bernhardmgruber: This should be swapped back.

bernhardmgruber:
I really appreciate how a MD-subscript operation into a buffer is now the dot product between index and pitch vector <3. This is also consistent with some MD array talk I have seen.

test/integ/matMul/src/matMul.cpp
```diff
@@ -240,16 +240,24 @@ TEMPLATE_LIST_TEST_CASE("matMul", "[matMul]", TestAccs)
               << alpaka::test::integ::measureRunTimeMs([&] { alpaka::memcpy(queueAcc, bufCAcc, bufCHost); }) << " ms"
               << std::endl;
 
-    alpaka::memcpy(queueAcc, bufCAcc, bufCHost);
+    alpaka::memcpy(queueAcc, bufCAcc, bufCHost); // FIXME(bgruber): this is necessary a second time?
```
psychocoderHPC: What do you mean by this comment?

psychocoderHPC (dismissed):
only one question but nothing against the PR at all.

test/integ/matMul/src/matMul.cpp
bernhardmgruber: This `memcpy` is also run 3 lines above, so it looks like the second call here is not necessary.

bernhardmgruber: I removed the line.

bernhardmgruber:
Hmm. Something got stuck with the readthedocs build. The error saids: "This build was terminated due to inactivity. If you continue to encounter this error, file a support request with and reference this build id (21768482)."

psychocoderHPC (approved):

PR #2095, opened 2023-08-25
Add a more comprehensible test for ViewSubView

bernhardmgruber:
This PR adds a test for `ViewSubView` that I used to understand its behavior. I still think it makes a nice unit test as well, so here it is.

### DIFF
--- a/test/unit/mem/view/src/ViewSubViewTest.cpp
+++ b/test/unit/mem/view/src/ViewSubViewTest.cpp
@@ -3,6 +3,7 @@
  */
 
 #include <alpaka/core/BoostPredef.hpp>
+#include <alpaka/mem/view/Traits.hpp>
 #include <alpaka/mem/view/ViewSubView.hpp>
 #include <alpaka/test/Extent.hpp>
 #include <alpaka/test/acc/TestAccs.hpp>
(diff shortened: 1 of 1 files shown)


test/unit/mem/view/src/ViewSubViewTest.cpp
```diff
@@ -166,3 +167,69 @@ TEMPLATE_LIST_TEST_CASE("viewSubViewOffsetConstTest", "[memView]", alpaka::test:
...
+    using Acc = TestType;
+    using Dev = alpaka::Dev<Acc>;
+    using Dim = alpaka::DimInt<2>;
+    using Idx = alpaka::Idx<Acc>;
+    using Vec = alpaka::Vec<Dim, Idx>;
+
+    auto const platform = alpaka::Platform<Acc>{};
+    auto const dev = alpaka::getDevByIdx(platform, 0);
+
+    auto const extent = Vec{4, 4};
```
fwyzard: just a suggestion, could you make this non-square, to better show the rows and columns ?

bernhardmgruber: Ver good idea, done!

bernhardmgruber:
Please ignore the failing TSAN test, this is fixed in #2103.

j-stephan (approved):

PR #2096, opened 2023-08-25
Silence warnings

j-stephan:
This PR silences two rather annoying outputs:

1. When using `icpx` it will complain about every single Catch2 file with the following "remark" (which apparently is something different from a warning):

```
icpx: remark: Note that use of '-g' without any optimization-level option will turn off most compiler optimizations similar to use of '-O0'; use '-Rno-debug-disables-optimization' to disable this remark [-Rdebug-disables-optimization]
```

2. When using `gcc` as a CUDA host compiler our usage of `-pedantic` will flood the logs with the following message:

```
/builds/hzdr/crp/alpaka/include/alpaka/core/ClipCast.hpp:23:3: warning: style of line directive is a GCC extension
   23 |         auto constexpr min = static_cast<V>(std::numeric_limits<alpaka::meta::HigherMin<T, V>>::min());
      |   ^ 
```
These messages seem to occur for every single line and therefore exceed the maximum log size limit, making logs rather useless when an error occurs later on.


psychocoderHPC (approved):

fwyzard:
Thanks for fixing the GCC one, I've just run into it in a different project as well.

About the icpx one, why is Catch2 built without optimisations?

j-stephan:
About `icpx`: This only occurs while compiling in debug mode. CMake adds `-g` to the command line, but not `-Og` (gcc) or `-O0` (clang). In this case, both compilers assume `-O0` (which is fine). Intel apparently decided to ~~warn~~ remark about this behaviour.

In alpaka we manually add `-Og` / `-O0` when compiling in debug mode but this doesn't propagate to Catch2 (or any third-party code).

PR #2098, opened 2023-08-28
use alpaka.png instead alpaka.pdf as logo in the sphix doc latex build

SimeonEhrig:
The readthedocs builder fails at building a pdf version of the documentation, if more than one pdf file is located in the build folder. Before this PR, it copies the `alpaka.pdf` (logo) to the build folder and generates the `alpaka-doc.pdf` (complete documentation).


bernhardmgruber (approved):
I find it weird that we cannot have vector graphics in PDF format when generating the documentation via latex. However, if this PR fixes building the documentation, we should go for it. Please create an issue though, that the latex build should use the pdf image so someone can have a look at it in the future.

SimeonEhrig:
Actual, there is already PR, which should allow to use multiple pdf's in the build folder: https://github.com/readthedocs/readthedocs.org/pull/10438

SimeonEhrig:
> I find it weird that we cannot have vector graphics in PDF format when generating the documentation via latex. However, if this PR fixes building the documentation, we should go for it. Please create an issue though, that the latex build should use the pdf image so someone can have a look at it in the future.

Done in #2100 If this PR solves the build problem, we can also enable the pdf build for PRs. So we will see immediately, if the pdf build fails.

bernhardmgruber:
> If this PR solves the build problem, we can also enable the pdf build for PRs. So we will see immediately, if the pdf build fails.

Please go ahead with this! We should know when the documentation can no longer be built!

SimeonEhrig:
This PR solved the build failure on the development branch. I also enabled pdf and epub builds for PR previews. At least I think I did it. The descriptions is a little bit small.

Issue #639, opened 2018-09-12
Why clang(CUDA) + OpenMP is not supported?


tdd11235813:
what was the reason for disabling CUDA(clang) + OpenMP? Does not clang support this or was this alpaka related?
( [corresponding PR](https://github.com/ComputationalRadiationPhysics/alpaka/pull/571) )

BenjaminW3:
It was already disabled before this PR. I do not remember all the details but as far as I remember, it is simply not supported by clang to do both at once.

tdd11235813:
it seems, clang disables OpenMP, when clang(cuda) is involved. So no `_OPENMP` and pragma omp's are defined. When CUDA is disabled, then it works again. Tested with clang5+cuda8 and clang6+cuda9.0.
Opened an issue over [there](https://github.com/clang-ykt/clang/issues/28).

ax3l:
From a quick test, `nvcc -std=c++11 -Xcompiler=-fopenmp` (9.2) works and `clang -x cuda -std=c++11 -fopenmp` as well... (trunk)

tdd11235813:
I will test with the new versions next week, when our Intel/Nvidia developer system is up.
(On the cluster modules are not built properly or missing)

tdd11235813:
Using clang7 and CUDA10.0 I got a known issue:
```
clang-7: error: cannot find libdevice for sm_30. Provide path to different CUDA installation via --cuda-path, or pass -nocudalib to build without linking with libdevice.
```
libdevice is where clang [expects it](https://bugs.llvm.org/show_bug.cgi?id=26966#c6).
```
$ ls /usr/local/cuda/nvvm/libdevice
libdevice.10.bc
```
However, as you can see, there is only that 1.0 thing. As of CUDA 9 it no more ships different libdevices. See also [this forum thread on NVidia](https://devtalk.nvidia.com/default/topic/1025985/cuda-setup-and-installation/fedora-25-cuda-9-missing-libdevice-files/).
The "one-size-fits-all" "solution" is recognized [by clang](https://github.com/llvm-mirror/clang/blob/master/lib/Driver/ToolChains/Cuda.cpp#L162-L172), but maybe clang still does not like CUDA 10.0.

[...]

PR #2099, opened 2023-08-28
[Refactor] CI job generator: define in the version.py which backend combinations should be tested

SimeonEhrig:
Before, the backends was enabled depending on the host and device compiler.
For each host device compiler combination only one backend combination was possible.
Now, the backend combinations are defined in the versions.py and each combination can be combined with a host device compiler combination, if no filter rule forbid it.

Is preparation for enabling GCC and Clang CPU tests in the GitLab CI.
~Should solve also issue #639~ -> no, see [here](https://github.com/alpaka-group/alpaka/issues/639#issuecomment-1695652937)

### DIFF
--- a/script/job_generator/generate_job_yaml.py
+++ b/script/job_generator/generate_job_yaml.py
@@ -219,6 +219,23 @@ def job_image(
 
 
 @typechecked
+def append_backend_variables(
+    variables: Dict[str, str], job: Dict[str, Tuple[str, str]]
+):
+    """Searches for enabled back-ends in the job parameters and appends the back-end
+    variable to variables to enable it in the CI job.
+
+    Args:
+        variables (Dict[str, str]): variables of the CI job
+        job (Dict[str, Tuple[str, str]]): job with parameters from the pair-wise
+            generator.
+    """
+    for backend in BACKENDS_LIST:
+        if backend in job:
+            variables[backend] = "ON"
+
+
+@typechecked
 def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
     """Add variables to the job depending of the job dict.
 
@@ -270,14 +287,12 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
     variables["ALPAKA_CI_CMAKE_VER"] = job[CMAKE][VERSION]
     variables["ALPAKA_BOOST_VERSION"] = job[BOOST][VERSION]
 
-    # TODO(SimeonEhrig) at the moment, not all backends are available in
-    # versions.sw_versions[BACKENDS], therefore we have to set each backend explicit
-    # Later, we can iterate over versions.sw_versions[BACKENDS] and check, if a Backend needs to be
-    # enabled
+    # all back-ends are disabled by default
+    # back-ends are conditionally enabled depending on the job parameters
     variables[ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE] = "OFF"
     variables[ALPAKA_ACC_CPU_B_SEQ_T_FIBERS_ENABLE] = "OFF"
     variables[ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE] = "OFF"
-    variables[ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE] = "ON"
+    variables[ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE] = "OFF"
     variables[ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE] = "OFF"
     variables[ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE] = "OFF"
     variables[ALPAKA_ACC_GPU_CUDA_ENABLE] = "OFF"
@@ -290,8 +305,9 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
     else:
         variables["ALPAKA_TEST_MDSPAN"] = "OFF"
 
+    append_backend_variables(variables, job)
+
     if job[DEVICE_COMPILER][NAME] == HIPCC:
-        variables[ALPAKA_ACC_GPU_HIP_ENABLE] = "ON"
         variables["CC"] = "clang"
         variables["CXX"] = "clang++"
         variables["GPU_TARGETS"] = "${CI_GPU_ARCH}"
@@ -322,7 +338,6 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
         ALPAKA_ACC_GPU_CUDA_ENABLE in job
         and job[ALPAKA_ACC_GPU_CUDA_ENABLE][VERSION] != OFF_VER
     ):
-        variables[ALPAKA_ACC_GPU_CUDA_ENABLE] = "ON"
         variables["ALPAKA_CI_STDLIB"] = "libstdc++"
         variables["CMAKE_CUDA_ARCHITECTURES"] = job[SM_LEVEL][VERSION]
         variables["ALPAKA_CI_CUDA_VERSION"] = job[ALPAKA_ACC_GPU_CUDA_ENABLE][VERSION]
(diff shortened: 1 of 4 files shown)


SimeonEhrig:
The failed TSAN job is independent of this PR, see #2101.

SimeonEhrig:
There is no rule, which requires an enabled OpenMP2 backend for `nvcc` as device compiler. Therefore both backend combinations are tested.

https://github.com/alpaka-group/alpaka/blob/9a30ae807d761bc0bf278735714a5ac10a04ae93/script/job_generator/versions.py#L37-L53

@psychocoderHPC Any opinion about the behavior?

psychocoderHPC:
> There is no rule, which requires an enabled OpenMP2 backend for `nvcc` as device compiler. Therefore both backend combinations are tested.
> 
> https://github.com/alpaka-group/alpaka/blob/9a30ae807d761bc0bf278735714a5ac10a04ae93/script/job_generator/versions.py#L37-L53
> 
> @psychocoderHPC Any opinion about the behavior?

@SimeonEhrig looks good for me.

script/job_generator/versions.py
```diff
@@ -30,11 +29,40 @@
...
+            ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE,
+            ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE,
+            ALPAKA_ACC_GPU_CUDA_ENABLE,
+        ],
+        # clang-cuda
+        # OpenMP is not supported for clang as cuda compiler
+        # https://github.com/alpaka-group/alpaka/issues/639
+        # therefore a extra combination is required
+        [
+            ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE,
```
psychocoderHPC, line 50: How can I see that this is for clang-cuda except by the comment? So I am wondering how the generator know that for clang-cuda this configuration should be used and not the first entry in `BACKENDS`?

psychocoderHPC (dismissed):

script/job_generator/versions.py
SimeonEhrig: You cannot see it directly in code, because it is decided by filter rules. So `clang-cuda` requires `ALPAKA_ACC_GPU_CUDA_ENABLE` and does not allow the `OMP2` backends:

https://github.com/alpaka-group/alpaka/blob/9a30ae807d761bc0bf278735714a5ac10a04ae93/script/job_generator/alpaka_filter.py#L48-L56

You can check if a compiler backend passes the filter with `ajc-validate` tool (except the alpaka filter, I need to implement support for this :thinking: ).

script/job_generator/generate_job_yaml.py
```diff
@@ -218,6 +218,23 @@ def verify_image(
     return "".join(verified_container_url)
 
 
+@typechecked
+def append_backend_variables(
+    variables: Dict[str, str], job: Dict[str, Tuple[str, str]]
+):
+    """Is searching for enabled back-ends in the job and appends the back-end
+    variable to variables to enable it in the CI job.
```
j-stephan:
```suggestion
    """Searches for enabled back-ends in the job parameters and appends the back-end
    variable to variables to enable it in the CI job.
```

Does the Python documentation have something similar to Doxygen's `\p` to mark function parameters? I needed to read it a few times before I realized that `variables` was referring to the parameter.

script/job_generator/generate_job_yaml.py
```diff
@@ -270,14 +287,12 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
     variables["ALPAKA_CI_CMAKE_VER"] = job[CMAKE][VERSION]
     variables["ALPAKA_BOOST_VERSION"] = job[BOOST][VERSION]
 
-    # TODO(SimeonEhrig) at the moment, not all backends are available in
-    # versions.sw_versions[BACKENDS], therefore we have to set each backend explicit
-    # Later, we can iterate over versions.sw_versions[BACKENDS] and check, if a Backend needs to be
-    # enabled
+    # by default, all backends are disabled
+    # backends will be enabled depending the job parameters
```
j-stephan:
```suggestion
    # all back-ends are disabled by default
    # back-ends are conditionally enabled depending on the job parameters
```

script/job_generator/generate_job_yaml.py
```diff
@@ -290,8 +305,10 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
     else:
         variables["ALPAKA_TEST_MDSPAN"] = "OFF"
 
+    # enabled backend depending on the job parameters
```
j-stephan: Do we need this comment here again?

script/job_generator/versions.py
```diff
@@ -30,11 +29,40 @@
     ],
     HIPCC: ["5.0", "5.1", "5.2", "5.3", "5.4", "5.5"],
     ICPX: ["2023.1.0", "2023.2.0"],
+    # Defines all backends, which will be enabled.
+    # Special cases for ALPAKA_ACC_GPU_CUDA_ENABLE and ALPAKA_ACC_GPU_HIP_ENABLE.
+    # The will be combine with nvcc and hip versions.
```
j-stephan:
```suggestion
    # Contains all enabled back-ends.
    # There are special cases for ALPAKA_ACC_GPU_CUDA_ENABLE and ALPAKA_ACC_GPU_HIP_ENABLE
    # which have to be combined with nvcc and hipcc versions.
```

script/job_generator/versions.py
```diff
@@ -85,24 +113,38 @@ def get_compiler_versions(clang_cuda: bool = True) -> List[Tuple[str, str]]:
 
 @typechecked
 def get_backend_matrix() -> List[List[Tuple[str, str]]]:
-    """Generate backend list, where only backend is active on the same time.
+    """Generates a backend list, with different combinations of enable backends.
```
j-stephan:
```suggestion
    """Generates a back-end list which contains different combinations of enabled back-ends.
```

script/job_generator/versions.py
```diff
@@ -85,24 +113,38 @@ def get_compiler_versions(clang_cuda: bool = True) -> List[Tuple[str, str]]:
...
-    if NVCC in sw_versions:
-        for cuda_version in sw_versions[NVCC]:
-            combination_matrix.append([(ALPAKA_ACC_GPU_CUDA_ENABLE, cuda_version)])
-
-    if ICPX in sw_versions:
-        combination_matrix.append([(ALPAKA_ACC_SYCL_ENABLE, ON_VER)])
+    for backend_list in sw_versions[BACKENDS]:
+        combination: List[Tuple[str, str]] = []
+        # all backends except ALPAKA_ACC_GPU_HIP_ENABLE and
+        # ALPAKA_ACC_GPU_CUDA_ENABLE can be only enabled
```
j-stephan:
```suggestion
        # all back-ends except ALPAKA_ACC_GPU_HIP_ENABLE and
        # ALPAKA_ACC_GPU_CUDA_ENABLE can be enabled without having a version number attached to them
```

script/job_generator/versions.py
```diff
@@ -85,24 +113,38 @@ def get_compiler_versions(clang_cuda: bool = True) -> List[Tuple[str, str]]:
...
+        for backend_name in backend_list:
+            if backend_name not in [
+                ALPAKA_ACC_GPU_HIP_ENABLE,
+                ALPAKA_ACC_GPU_CUDA_ENABLE,
+            ]:
+                combination.append((backend_name, ON_VER))
+
+        # add a backend entry for each CUDA SDK version
```
j-stephan:
```suggestion
        # add a back-end entry for each CUDA SDK version
```

script/job_generator/versions.py
```diff
@@ -85,24 +113,38 @@ def get_compiler_versions(clang_cuda: bool = True) -> List[Tuple[str, str]]:
...
+        if ALPAKA_ACC_GPU_CUDA_ENABLE in backend_list:
+            for cuda_version in sw_versions[NVCC]:
+                combination_copy = combination[:]
+                combination_copy.append((ALPAKA_ACC_GPU_CUDA_ENABLE, cuda_version))
+                combination_matrix.append(combination_copy)
+        # add a backend entry for each HIP SDK version
```
j-stephan:
```suggestion
        # add a back-end entry for each HIP SDK version
```

j-stephan (changes requested):
Just a few suggestions:

script/job_generator/generate_job_yaml.py
SimeonEhrig: No, I believe it was a left over, before I moved the code to an extra function.

j-stephan (dismissed):

SimeonEhrig:
I stopped the merge process, because I found a bug in #2102 which already exist in this PR. The bug is simply not triggered, because we don't use a specific combination of enabled backends.

SimeonEhrig:
> I stopped the merge process, because I found a bug in #2102 which already exist in this PR. The bug is simply not triggered, because we don't use a specific combination of enabled backends.

@j-stephan The bug is fixed in the `alpaka-job-coverage library 1.5.3.` and tests passed: https://github.com/alpaka-group/alpaka/pull/2102

Therefore the PR is ready to merge. Since your last review I only increased `alpaka-job-coverage library` version.

j-stephan (approved):

PR #2103, opened 2023-08-29
Disable part of the threadpool test for TSAN

### LINKED ISSUE #2101: coreTest fails with TSAN
The clang-16 TSAN CI job failed a few times at the coreTest. See for example [here](https://github.com/alpaka-group/alpaka/actions/runs/5999816286/job/16270849037?pr=2092#step:5:4976).

PR #2103 disabled the offending test. We should revisit this in the future and make the test work.

bernhardmgruber:
Fixes: #2101.

### DIFF
--- a/test/unit/core/src/ThreadPool.cpp
+++ b/test/unit/core/src/ThreadPool.cpp
@@ -6,6 +6,12 @@
 
 #include <catch2/catch_test_macros.hpp>
 
+#if defined(__has_feature)
+#    if __has_feature(thread_sanitizer)
+#        define TSAN
+#    endif
+#endif
+
 TEST_CASE("threadpool", "[core]")
 {
     alpaka::core::detail::ThreadPool tp{2};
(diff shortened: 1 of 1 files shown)


test/unit/core/src/ThreadPool.cpp
```diff
@@ -16,6 +22,10 @@ TEST_CASE("threadpool", "[core]")
 
     CHECK_THROWS_AS(f1.get(), std::runtime_error);
 
+#ifdef TSAN
+#    warning "Part of the threadpool test fails the TSAN CI and is therefore disabled."
+    // TODO(bgruber): Revisit this in the future on a new CI image. This problem does not happen locally.
```
j-stephan, line 27: Interesting. So probably some Ubuntu 20.04 issue then?

j-stephan (approved):

test/unit/core/src/ThreadPool.cpp
bernhardmgruber: I have no idea. It does not happen on my Ubuntu 23.04. Also interestingly, the CI reports a higher `BOOST_LIB_STD_GNU` than my local run :D In any case, all I can do is disable the test and add this note. Let's have a look again on newer Ubuntu images.

PR #2106, opened 2023-08-30
Remove STLTuple

j-stephan:
The SYCL back-end and the late OpenMP 5 / OpenACC back-ends made use of Codeplay's tuple implementation to pass parameter packs into kernels. STLTuple is no longer maintained, the repository has been archived. Tests on my local machine have shown that it is now possible to use `std::tuple` instead. This PR replaces all usages of STLTuple with `std::tuple`.


bernhardmgruber (approved):

bernhardmgruber:
Those are my favorite PRs!

include/alpaka/kernel/TaskKernelGenericSycl.hpp
```diff
@@ -44,7 +42,7 @@
                 sycl::nd_item<TDim::value> work_item) [[intel::reqd_sub_group_size(sub_group_size)]]                  \
             {                                                                                                         \
                 auto acc = TAcc{item_elements, work_item, dyn_shared_accessor, st_shared_accessor};                   \
-                core::apply(                                                                                          \
+                std::apply(                                                                                           \
```
j-stephan, line 45: Side note: These white spaces behind the opening `(` are added by clang-format. Removing them makes the CI job complain.

bernhardmgruber: Well, they are necessary to align the `\` at the end of the line.

psychocoderHPC:
Ohh we are currently building code based on alpaka's Tuple implementation because we would like to pass classes with tuple inside onto the GPU what is not possible with `std::tupel` because it is not trivially copyable.

j-stephan:
I see. I'll convert this to a draft so we can discuss on how to proceed.

psychocoderHPC:
We can always pull the tuple implementation into PIConGPU but avoid this because alpaka is shipping it anyway.

bernhardmgruber:
I would argue that this dependence could move downstream into picongpu, so removing it from alpaka is fair.

j-stephan:
I agree with Bernhard here. STLTuple wasn't intended to be exposed as a public interface - user code depending on it should pull it in on its own.

j-stephan:
@psychocoderHPC Would it be okay for you to pull in STLTuple directly into the affected code bases?

psychocoderHPC (approved):

psychocoderHPC:
offline discussed: we decided in the developer meeting to remove the code because we do not provide CI tests.
We discussed introducing an own implementation in the future but this requires that we test the implementation.

PR #2107, opened 2023-08-30
Pass `alpaka_ENABLE_WERROR` from environment to CMake

j-stephan:
In #1977 I forgot to ensure that `alpaka_ENABLE_WERROR` is also passed to CMake in the generation step. This PR fixes this oversight.


SimeonEhrig (dismissed):

SimeonEhrig:
The change looks good. But no test job passed. I believe there is a lot to do :grimacing:

test/common/devCompileOptions.cmake
```diff
@@ -10,7 +10,6 @@ if(alpaka_ACC_GPU_CUDA_ENABLE AND (CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA"))
     if(alpaka_ENABLE_WERROR)
         list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wreorder>")
         list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wdefault-stream-launch>")
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wmissing-launch-bounds>")
```
j-stephan, line 13: I removed this warning since it warns about an optional CUDA feature (enforcing a compile-time grid / block size) we do not make use of.

bernhardmgruber:
Since so many warnings occur in the code generated by nvcc, I wonder why we need to enable `-Werror` on the host compiler as well. Shouldn't it be enough to just enable `--Werror` for nvcc for CUDA, and `-Werror` for CXX? And we just let nvcc generate whatever it wants. I think we just have to fix too much otherwise.

bernhardmgruber:
The changeset of this PR is also included in #2163. If that PR gets merged, this PR becomes obsolete.

j-stephan:
Superseded by #2163.

PR #2108, opened 2023-08-30
fix CUDA/HIP device wait for event

### LINKED ISSUE #2060: CUDA/HIP device wait for an event is wrong working
alpaka is providing the functionality that all streams of a device can be dependent on an event. This is nothing CUDA is providing via an API call.
alpaka is implementing it by using the stream `zero`

https://github.com/alpaka-group/alpaka/blob/ac7b41daf2be8d34cb2212535b866b3e41e16ce7/include/alpaka/event/EventUniformCudaHipRt.hpp#L218-L232

alpaka is creating the streams so that they do not wait for the legacy blocking stream zero.

https://github.com/alpaka-group/alpaka/blob/ac7b41daf2be8d34cb2212535b866b3e41e16ce7/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp#L58

Therefore our implementation is wrong because there will be no synchronization.

A solution is to implement it the same way we do it for other backends. Da device must track a list with all stream/queues and in the implementation call `alpaka::wait(queue, event)` for all queues of the device.

psychocoderHPC:
fix #2060

The implementationto let a device wait for an event was wrongly implemented.

**new implementation**

Collect all queues of an device within the device itself, equal to the CPU device implementation, and let each stream wait for the event.

### DIFF
--- a/include/alpaka/dev/DevUniformCudaHipRt.hpp
+++ b/include/alpaka/dev/DevUniformCudaHipRt.hpp
@@ -8,10 +8,12 @@
 #include "alpaka/core/Cuda.hpp"
 #include "alpaka/core/Hip.hpp"
 #include "alpaka/dev/Traits.hpp"
+#include "alpaka/dev/common/QueueRegistry.hpp"
 #include "alpaka/mem/buf/Traits.hpp"
 #include "alpaka/platform/Traits.hpp"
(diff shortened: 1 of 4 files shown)


j-stephan (approved):

j-stephan:
`sudo merge` because the Read the Docs build failed.

PR #2113, opened 2023-08-31
Fix integer conversions

j-stephan:
This PR fixes the following warning unconvered by #2107:

```
/home/runner/work/alpaka/alpaka/include/alpaka/test/mem/view/ViewTest.hpp:136:29: error: useless cast to type 'long unsigned int' [-Werror=useless-cast]
  136 |                             static_cast<long unsigned>(i),
      |                             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~            
```

In addition, the affected `printf` command now uses the correct conversion specifiers for the types passend into it.

### DIFF
--- a/include/alpaka/test/mem/view/ViewTest.hpp
+++ b/include/alpaka/test/mem/view/ViewTest.hpp
@@ -1,5 +1,4 @@
-/* Copyright 2022 Axel Huebl, Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber, Antonio Di
- * Pilato
+/* Copyright 2023 Benjamin Worpitz, Sergei Bastrakov, René Widera, Bernhard Manfred Gruber, Jan Stephan
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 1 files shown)


include/alpaka/test/mem/view/ViewTest.hpp
```diff
@@ -131,11 +130,7 @@ namespace alpaka::test
                 {
                     if(pBytes[i] != byte)
                     {
-                        printf(
-                            "Byte at offset %lu is different: %u != %u\n",
-                            static_cast<long unsigned>(i),
-                            static_cast<unsigned>(pBytes[i]),
-                            static_cast<unsigned>(byte));
+                        printf("Byte at offset %zu is different: %hhu != %hhu\n", i, pBytes[i], byte);
```
bernhardmgruber, line 141: `zu` is unfortunately not supported by nvcc :/

j-stephan: How is it compiling and running then? Does `nvcc` silently replace this with something else?

bernhardmgruber: It didn't the other day with my nvcc 12.x at work. It said that the format specifier `z` is not supported. So I changed it to the `long unsigned` here.

Mind that if the unit tests run successfully, this code is never printed. So successful completion of the test suite does not tell you whether this change is correct.

bernhardmgruber: I quickly tried to add a test case for that but it's not so easy to capture the printf output. Also some refactoring of the test suite is necessary. I postpone that effort.

bernhardmgruber (approved):

PR #2128, opened 2023-09-05
Turn off fast math for Catch2 build

j-stephan:
`icpx` defaults to `-fp-model=fast` which generates the following warning while building Catch2:

```
/home/stepha27/workspace/caravan/alpaka/thirdParty/catch2/src/catch2/catch_approx.cpp:55:82: warning: explicit comparison with infinity in fast floating point mode [-Wtautological-constant-compare]
            || marginComparison(m_value, other, m_epsilon * (m_scale + std::fabs(std::isinf(m_value)? 0 : m_value)));
                                                                                 ^~~~~~~~~~~~~~~~~~~
1 warning generated.
```

For alpaka we manually set this to `-fp-model=precise`. This PR does the same for the Catch2 build.

### DIFF
--- a/test/common/devCompileOptions.cmake
+++ b/test/common/devCompileOptions.cmake
@@ -146,7 +146,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "Inte
 
     if(${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
         # fast math is turned on by default with ICPX, which breaks our unit tests
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-ffp-model=precise")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-fp-model=precise")
 
         if (alpaka_ACC_SYCL_ENABLE)
             # avoid: warning: disabled expansion of recursive macro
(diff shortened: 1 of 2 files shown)


test/common/devCompileOptions.cmake
```diff
@@ -146,7 +146,7 @@ if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "Inte
 
     if(${CMAKE_CXX_COMPILER_ID} STREQUAL "IntelLLVM")
         # fast math is turned on by default with ICPX, which breaks our unit tests
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-ffp-model=precise")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-fp-model=precise")
```
j-stephan, line 149: The option is called `-fp-model`, not `-ffp-model`. I guess the latter is an alias for compatibility.

bernhardmgruber: `-ffp-model` is the flag in clang, see: https://clang.llvm.org/docs/UsersManual.html#cmdoption-ffp-model
Your change seems right according to Intel's documentation: https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/fp-model-fp.html

bernhardmgruber (approved):

PR #2131, opened 2023-09-05
GitLab CI: enable runtime tests for gcc and clang cpu jobs

SimeonEhrig:


j-stephan (approved):
LGTM. Question: Would it be possible to request ARM runners here, too?

SimeonEhrig:
> LGTM. Question: Would it be possible to request ARM runners here, too?

Short answer no. Long answer see: #2134

PR #2133, opened 2023-09-05
GitLab CI: set `alpaka_DEBUG`

### LINKED ISSUE #2130: Add CI jobs with `alpaka_DEBUG=2`
Currently, all (but one analysis) Debug CI jobs run with `alpaka_DEBUG=0`. This means, that extra debugging code is never tested by the CI. We should add at least a few CI runs testing different debug levels.

SimeonEhrig:
fixes: #2130 

- [x] requires #2131 to be merged before

### DIFF
--- a/script/job_generator/generate_job_yaml.py
+++ b/script/job_generator/generate_job_yaml.py
@@ -3,7 +3,7 @@ SPDX-License-Identifier: MPL-2.0
 
 Create GitLab-CI job description written in yaml from the job matrix."""
 
-from typing import List, Dict, Tuple
+from typing import List, Dict, Tuple, Union
 from typeguard import typechecked
 import os, yaml
 import gitlab
@@ -221,7 +221,7 @@ def job_image(
 
 @typechecked
 def append_backend_variables(
-    variables: Dict[str, str], job: Dict[str, Tuple[str, str]]
+    variables: Dict[str, Union[str, int]], job: Dict[str, Tuple[str, Union[str, int]]]
 ):
     """Searches for enabled back-ends in the job parameters and appends the back-end
     variable to variables to enable it in the CI job.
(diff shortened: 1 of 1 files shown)


script/job_generator/generate_job_yaml.py
```diff
@@ -287,6 +288,11 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
     variables["ALPAKA_CI_CMAKE_VER"] = job[CMAKE][VERSION]
     variables["ALPAKA_BOOST_VERSION"] = job[BOOST][VERSION]
 
+    if job[BUILD_TYPE][VERSION] == CMAKE_DEBUG:
+        variables["alpaka_DEBUG"] = 2
```
j-stephan, line 292: There is also `alpaka_DEBUG = 1`. Or is this enabled somewhere else?

SimeonEhrig: If we want to test both, I need to add the alpaka debug variable to the parameter set of the job generator. Or extend the `BUILD_TYPE` type to `CMAKE_DEBUG1` and `CMAKE_DEBUG2`.

j-stephan: I think the parameter set would be the best fit here.

bernhardmgruber:
Please rebase, I think you are missing my PR that fixed one of the bugs: #2127

SimeonEhrig:
@bernhardmgruber I rebased. But there is still the problem with the heavy job output.

bernhardmgruber:
@SimeonEhrig thx! I don't have a quick solution ready and unfortunately no time the next days. Let's discuss this in the next alpaka VC.

bernhardmgruber:
@SimeonEhrig let's think about this after the release!

psychocoderHPC:
@SimeonEhrig Is this PR outdated or should it be finished?

SimeonEhrig:
close because `alpaka_Debug` is set in PR #2315. A better selection mechanism for the debug level will be implemented later.

PR #2135, opened 2023-09-05
Improve nvcc's handling of `-Werror`

j-stephan:
Another round of fixes for #2107. `nvcc` treats `-Werror` differently than the other compilers; this PR fixes the behaviour. Example error:

```
nvcc -Werror -Wdouble-promotion [and so on]
nvcc fatal   : Value '-Wdouble-promotion' is not defined for option 'Werror'
```

### DIFF
--- a/test/common/devCompileOptions.cmake
+++ b/test/common/devCompileOptions.cmake
@@ -7,7 +7,16 @@
 # Compiler settings.
 #-------------------------------------------------------------------------------
 if(alpaka_ACC_GPU_CUDA_ENABLE AND (CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA"))
-    list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:-Wdefault-stream-launch -Werror=default-stream-launch>)
+    if(alpaka_ENABLE_WERROR)
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:--Wreorder>)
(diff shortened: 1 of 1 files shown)


test/common/devCompileOptions.cmake
```diff
@@ -113,7 +123,9 @@ endif()
 # Clang, AppleClang, ICPX
 if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" OR CMAKE_CXX_COMPILER_ID STREQUAL "IntelLLVM")
     if(alpaka_ENABLE_WERROR)
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-Werror")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CXX>:-Werror>
+                                               $<$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>:-Xcompiler -Werror>
+                                               $<$<COMPILE_LANG_AND_ID:CUDA,AppleClang,Clang,IntelLLVM>:-Werror>)
```
bernhardmgruber: Does only nvcc not support `-Werror` directly? Or are you omitting `-Werror` on purpose for CUDA compilers other than NVIDIA, AppleClang, Clang and IntelLLVM? I don't understand why the last clause is necessary.

Why would this not work?
```suggestion
        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<IF:$<COMPILE_LANG_AND_ID:CUDA,NVIDIA>,-Xcompiler -Werror,-Werror>)
```

j-stephan: It was a long day...

bernhardmgruber (approved):

PR #2136, opened 2023-09-05
Refactor `mapIdx` and `mapIdxPitchBytes`

bernhardmgruber:
I had a quick look at the failing warpTest and noticed that the implementations of `mapIdx` and `mapIdxPitchBytes` are way too complex. So here is a refactoring.

### DIFF
--- a/include/alpaka/idx/MapIdx.hpp
+++ b/include/alpaka/idx/MapIdx.hpp
@@ -12,237 +12,82 @@
 
 namespace alpaka
 {
-    namespace detail
-    {
-        //! Maps a linear index to a N dimensional index.
-        template<std::size_t TidxDimOut, std::size_t TidxDimIn, typename TSfinae = void>
-        struct MapIdx;
-        //! Maps a N dimensional index to the same N dimensional index.
-        template<std::size_t TidxDim>
-        struct MapIdx<TidxDim, TidxDim>
-        {
-            // \tparam TElem Type of the index values.
-            // \param idx Idx to be mapped.
-            // \param extent Spatial size to map the index to.
-            // \return A N dimensional vector.
-            ALPAKA_NO_HOST_ACC_WARNING
-            template<typename TElem>
-            ALPAKA_FN_HOST_ACC static auto mapIdx(
-                Vec<DimInt<TidxDim>, TElem> const& idx,
-                [[maybe_unused]] Vec<DimInt<TidxDim>, TElem> const& extent) -> Vec<DimInt<TidxDim>, TElem>
-            {
-                return idx;
-            }
-        };
-        //! Maps a 1 dimensional index to a N dimensional index.
-        template<std::size_t TidxDimOut>
-        struct MapIdx<TidxDimOut, 1u, std::enable_if_t<(TidxDimOut > 1u)>>
-        {
-            // \tparam TElem Type of the index values.
-            // \param idx Idx to be mapped.
-            // \param extent Spatial size to map the index to
-            // \return A N dimensional vector.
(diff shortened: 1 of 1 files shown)


include/alpaka/idx/MapIdx.hpp
```diff
@@ -12,237 +12,82 @@
...
-                Vec<DimInt<TidxDimIn>, TElem> const&,
-                Vec<DimInt<TidxDimExtents>, TElem> const&) -> Vec<DimInt<0u>, TElem>
-            {
-                return {};
-            }
-        };
-    } // namespace detail
-
-    //! Maps a N dimensional index to a N dimensional position.
+    //! Maps a N dimensional index to a N dimensional position. At least one dimension must always be 1 or zero.
```
j-stephan:
```suggestion
    //! Maps an N-dimensional index to an N-dimensional position. At least one dimension must always be 1 or zero.
```

include/alpaka/idx/MapIdx.hpp
```diff
@@ -12,237 +12,82 @@
...
     //!
-    //! \tparam TidxDimOut Dimension of the index vector to map to.
-    //! \tparam TidxDimIn Dimension of the index vector to map from.
-    //! \tparam TElem Type of the elements of the index vector to map from.
+    //! \tparam TDimOut Dimension of the index vector to map to.
+    //! \param in The index vector to map from.
+    //! \param extent The extents of the space with dimension higher than 1.
```
j-stephan: This sentence sounds a bit weird.

include/alpaka/idx/MapIdx.hpp
```diff
@@ -12,237 +12,82 @@
...
-                return {idx.back() + (subVecBegin<DimMinusOne>(pitch) * subVecBegin<DimMinusOne>(idx)).sum()};
+                out[d] = static_cast<TElem>(flat % extent[d]);
+                flat /= extent[d];
             }
-        };
+            out.front() = static_cast<TElem>(flat);
+            return out;
+        }
+        else
+            static_assert(sizeof(TElem) == 0, "Not implemented");
```
j-stephan: Why not

```suggestion
            static_assert(!sizeof(TElem), "Not implemented");
```

like we do elsewhere?

include/alpaka/idx/MapIdx.hpp
```diff
@@ -12,237 +12,82 @@
...
-    template<std::size_t TidxDimOut, std::size_t TidxDimIn, std::size_t TidxDimPitch, typename TElem>
-    ALPAKA_FN_HOST_ACC auto mapIdxPitchBytes(
-        Vec<DimInt<TidxDimIn>, TElem> const& idx,
-        Vec<DimInt<TidxDimPitch>, TElem> const& pitch) -> Vec<DimInt<TidxDimOut>, TElem>
-    {
-        return detail::MapIdxPitchBytes<TidxDimOut, TidxDimIn>::mapIdxPitchBytes(idx, pitch);
+            return result;
+        }
+        else
+            static_assert(sizeof(TElem) == 0, "Not implemented");
```
j-stephan: Also here.

j-stephan (changes requested):

include/alpaka/idx/MapIdx.hpp
bernhardmgruber: Because the first version requires less cycles to parse in my brain. But I don't mind changing it.

include/alpaka/idx/MapIdx.hpp
bernhardmgruber: Redrafted.

j-stephan (approved):

PR #2138, opened 2023-09-05
Check for C++20 features

j-stephan:
In some places we include C++20 headers if available. On MSVC these headers are available even for C++ versions before C++20 but they will issue the following warning:

```
C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Tools\MSVC\14.36.32532\include\bit(12): warning STL4038: The contents of <bit> are available only with C++20 or later. [D:\a\alpaka\alpaka\build\test\unit\workDiv\workDivTest.vcxproj]
```

This PR ensures that the headers are definitely only included when the C++20 standard is active.

Uncovered by #2107.

### DIFF
--- a/include/alpaka/intrinsic/IntrinsicCpu.hpp
+++ b/include/alpaka/intrinsic/IntrinsicCpu.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Sergei Bastrakov, Bernhard Manfred Gruber
+/* Copyright 2023 Sergei Bastrakov, Bernhard Manfred Gruber, Jan Stephan
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -10,7 +10,10 @@
 
 #include <bitset>
 #include <climits>
-#if __has_include(<bit>)
+#if __has_include(<version>) // Not part of the C++17 standard but all major standard libraries include this
+#    include <version>
+#endif
+#ifdef __cpp_lib_bitops
 #    include <bit>
 #endif
 
--- a/include/alpaka/math/Traits.hpp
+++ b/include/alpaka/math/Traits.hpp
@@ -1,5 +1,5 @@
 /* Copyright 2023 Benjamin Worpitz, Matthias Werner, Jan Stephan, Bernhard Manfred Gruber, Sergei Bastrakov,
- * Andrea Bocci
+ *                Andrea Bocci, René Widera
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -10,7 +10,10 @@
 
 #include <cmath>
 #include <complex>
-#if __has_include(<numbers>)
+#if __has_include(<version>) // Not part of the C++17 standard but all major standard libraries include this
+#    include <version>
+#endif
+#ifdef __cpp_lib_math_constants
 #    include <numbers>
 #endif
 


include/alpaka/intrinsic/IntrinsicCpu.hpp
```diff
@@ -10,7 +10,7 @@
 
 #include <bitset>
 #include <climits>
-#if __has_include(<bit>)
+#if(__cplusplus >= 202002L) && __has_include(<bit>)
 #    include <bit>
 #endif
```
bernhardmgruber: This is a pessimization in general, because compilers can provide the `<bit>` header in c++2a mode already, and not bump `__cplusplus`.

This is also dangerous on MSVC, because `__cplusplus` in C++20 mode is still `199711L`, unless the `/Zc:__cplusplus` flag is specified. So any user not defining this will also not get the `<bit>` header.

The right thing to do here IMO is to silence the warning.

j-stephan: > This is also dangerous on MSVC, because `__cplusplus` in C++20 mode is still `199711L`

Wow, that is amazingly stupid. Alright, I'll just silence the warning then.

j-stephan:
Okay, this is a fun one. `STL4038` is implemented in terms of `#pragma message` ([Documentation](https://learn.microsoft.com/en-us/cpp/preprocessor/message?view=msvc-170)) and AFAIK there is no way to suppress this. I opened an issue here: https://github.com/microsoft/STL/issues/4010

Until there is a nice solution I see the following possibilities:

* Keep the current PR changes and add `/Zc:__cplusplus` to alpaka's `INTERFACE` flags. I don't think we care about `c++2a`anyway, only "complete" standards.
* Turn off the warning by `#define`ing some Microsoft STL internal macros. This is dangerous.
* Turn off the warning by `#define`ing `message(...)` to nothing. This is probably dangerous.

bernhardmgruber:
> I don't think we care about c++2a anyway, only "complete" standards.

I care about those a lot, since compiler vendors implement various features at their own pace, even across standards. But I know this is not the mental model of many users :)

Also, a compiler accepting e.g. `-std=c++20` and providing the value `202002` for the macro `__cplusplus`, does not mean, that all C++20 features are available. Clang provides these values, but still lacks some stuff around modules or CTAD for aliases, see: https://clang.llvm.org/cxx_status.html. This is all a best effort, there is no "complete" standard (implementation). Because of this, feature test macros were introduced, so we must check those in addition to anything else we do.

Maybe we need to go with the version I have commented on your issue at the MS STL:
```c++
#if __has_include(<version>)
#    include <version>
#endif
// later:
#ifdef __cpp_lib_bitops
#    include <bit>
#endif
// later:
#ifdef __cpp_lib_bitops
    std::popcount(value);
#else
    // fallback implementation
#endif
```
But I want to see what STL replies.

bernhardmgruber:
Alright, his reply did not help me resolve this case. So I kind of want to recommend my solution above that pulls in the `<version>` header when available, etc. But it's also really ugly. So I guess I will just leave it to you how you want to solve this. And then we kick that out when we move to C++20.

bernhardmgruber (approved):

Issue #2068, opened 2023-08-08
static shared memory in SYCL


AuroraPerego:
The shared memory in SYCL currently is allocated with `sycl::local_accessor` that must be instantiated before the kernel and then their pointer is passed as an argument to the kernel. The logic that shared memory allocations in SYCL follow is the same as for the serial back-end. Some memory (47kb) is preallocated for the static shared memory and this causes two problems: 
- three failures in the tests for the SYCL CPU back-end with a `PI_ERROR_OUT_OF_RESOURCES` error (`atomicTest`, `blockSharedTest` and `blockSharedSharingTest`)
- a huge slow-down in the kernel launch (spotted when profiling [pixeltrack-standalone](https://github.com/cms-patatrack/pixeltrack-standalone))

An alternative to the accessors is the allocation of shared memory directly in the kernel with `sycl::ext::oneapi::group_local_memory_for_overwrite`, a possible implementation is [here](https://github.com/AuroraPerego/alpaka/tree/shared_memory).
This solves both the problems (all the tests pass and the application is faster) but now the `blockSharedTest` fails.
The failure is related to the fact that when allocating some memory with the same type and size it is expected to return always the same pointer. This is done automatically in CUDA, but cannot be done easily in SYCL since static variables cannot be used inside kernels.
This raised also some concerns on the fact that the way this is currently implemented with the `__COUNTER__` may not work properly when dealing with different translation units, e.g. when the user has an header file with the shared variable that is included in multiple files. We have not come up yet with a different approach to this problem

FYI @fwyzard

psychocoderHPC:
note: the behaviour of `sycl::ext::oneapi::group_local_memory_for_overwrite` has changed with onapi 2025.2.
Allocation of the same data type is providing aliased memory equal to the cuda behaviour.

I found it in alpaka3 while updating to oneapi 2025.2
https://github.com/alpaka-group/alpaka3/pull/195

fwyzard:
It seems that 2025.2 also introduces a new [`work_group_static` extension](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_work_group_static.asciidoc), which may be better for implementing "shared" memory.

fwyzard:
I think the change is a consequence of https://github.com/intel/llvm/pull/16356 .

fwyzard:
[...]

PR #2140, opened 2023-09-05
update SYCL README

AuroraPerego:
I updated the README because the reason of `PI_ERROR_OUT_OF_RESOURCES` on the CPU is the OpenCL runtime and not the shared memory.

### DIFF
--- a/README_SYCL.md
+++ b/README_SYCL.md
@@ -97,7 +97,8 @@ These can be used interchangeably (some restrictions apply - see below) with the
   See [Intel's FAQ](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#feature-double-precision-emulation-fp64) for more information.
 * The FPGA back-end does not support atomics. alpaka will not check this.
 * Device global variables (corresponding to `__device__` and `__constant__` variables in CUDA) are not supported in the SYCL back-end yet.
-* Shared memory does not yet work properly on CPU back-ends. The tests that use it (`atomicTest`, `blockSharedTest` and `blockSharedSharingTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`.
+* Shared memory works but on the GPU it is very slow.
+* The latest Intel OpenCL CPU runtime does not work properly. Some tests (`atomicTest`, `blockSharedTest`, `blockSharedSharingTest` and `warpTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`. The only runtime version that seems to work is 2022.14.8.0.04 (can be downloaded [here](https://github.com/intel/llvm/releases/download/2022-WW33/oclcpuexp-2022.14.8.0.04_rel.tar.gz) apart from a bug with `all_of_group` / `any_of_group` that requires the warp size being equal to the block size as a workaround.
 
 ### Choosing the sub-group size (warp size)
 


README_SYCL.md
```diff
@@ -97,7 +97,8 @@ These can be used interchangeably (some restrictions apply - see below) with the
   See [Intel's FAQ](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#feature-double-precision-emulation-fp64) for more information.
 * The FPGA back-end does not support atomics. alpaka will not check this.
 * Device global variables (corresponding to `__device__` and `__constant__` variables in CUDA) are not supported in the SYCL back-end yet.
-* Shared memory does not yet work properly on CPU back-ends. The tests that use it (`atomicTest`, `blockSharedTest` and `blockSharedSharingTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`.
+* Shared memory works but on the GPU it's very slow.
```
j-stephan:
```suggestion
* Shared memory works but on the GPU it is very slow.
```

Is the reason for this alpaka's implementation or is this a general Intel thing?

README_SYCL.md
```diff
@@ -97,7 +97,8 @@ These can be used interchangeably (some restrictions apply - see below) with the
...
+* The latest Intel OpenCL CPU runtime does not work properly. Some tests (`atomicTest`, `blockSharedTest`, `blockSharedSharingTest` and `warpTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`. The only runtime that seems to work is the 2022.14.8.0.04 (can be downloaded [here](https://github.com/intel/llvm/releases/download/2022-WW33/oclcpuexp-2022.14.8.0.04_rel.tar.gz) apart from a bug with `all_of_group` / `any_of_group` that requires the warp size being equal to the block size as a workaround.
```
j-stephan: Do Intel know about this?

j-stephan (changes requested):
Thank you! Just one thing:

README_SYCL.md
AuroraPerego: the alpaka implementation of the shared memory is slower than the native one

README_SYCL.md
AuroraPerego: yes..

README_SYCL.md
j-stephan: I see. Have you been able to identify a cause yet?

AuroraPerego: yes, we are allocating a lot of shared memory for each kernel even if we do not use it and this causes a slow down, see #2068

README_SYCL.md
```diff
@@ -97,7 +97,8 @@ These can be used interchangeably (some restrictions apply - see below) with the
   See [Intel's FAQ](https://github.com/intel/compute-runtime/blob/master/opencl/doc/FAQ.md#feature-double-precision-emulation-fp64) for more information.
 * The FPGA back-end does not support atomics. alpaka will not check this.
 * Device global variables (corresponding to `__device__` and `__constant__` variables in CUDA) are not supported in the SYCL back-end yet.
-* Shared memory does not yet work properly on CPU back-ends. The tests that use it (`atomicTest`, `blockSharedTest` and `blockSharedSharingTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`.
+* Shared memory works but on the GPU it is very slow.
+* The latest Intel OpenCL CPU runtime does not work properly. Some tests (`atomicTest`, `blockSharedTest`, `blockSharedSharingTest` and `warpTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`. The only runtime that seems to work is the 2022.14.8.0.04 (can be downloaded [here](https://github.com/intel/llvm/releases/download/2022-WW33/oclcpuexp-2022.14.8.0.04_rel.tar.gz) apart from a bug with `all_of_group` / `any_of_group` that requires the warp size being equal to the block size as a workaround.
```
j-stephan:
```suggestion
* The latest Intel OpenCL CPU runtime does not work properly. Some tests (`atomicTest`, `blockSharedTest`, `blockSharedSharingTest` and `warpTest`) fail with a `PI_ERROR_OUT_OF_RESOURCES`. The only runtime version that seems to work is 2022.14.8.0.04 (can be downloaded [here](https://github.com/intel/llvm/releases/download/2022-WW33/oclcpuexp-2022.14.8.0.04_rel.tar.gz) apart from a bug with `all_of_group` / `any_of_group` that requires the warp size being equal to the block size as a workaround.
```

j-stephan (changes requested):
Almost there, just one last thing:

j-stephan (approved):

PR #2142, opened 2023-09-06
Gitlab CI: enable test for the sycl CPU backend

SimeonEhrig:
I didn't expect, that it will work now, but it is interesting.

### DIFF
--- a/script/job_generator/generate_job_yaml.py
+++ b/script/job_generator/generate_job_yaml.py
@@ -15,7 +15,7 @@ from alpaka_globals import *  # pylint: disable=wildcard-import,unused-wildcard-
 from util import print_warn, exit_error
 
 JOB_COMPILE_ONLY = "compile_only_job"
-JOB_RUNTIME = "runtime_job_gpu"
+JOB_GPU_RUNTIME = "runtime_job_gpu"
 JOB_CPU_RUNTIME = "runtime_job_cpu"
 JOB_ROCM_RUNTIME = "rocm_runtime_job"
 JOB_NVCC_GCC_RUNTIME = "nvcc_gcc_runtime_job"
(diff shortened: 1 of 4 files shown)


SimeonEhrig:
@fwyzard @AuroraPerego FYI

AuroraPerego:
Good! The tests that fail are due to the OpenCL CPU runtime :(

SimeonEhrig:
> Good! The tests that fail are due to the OpenCL CPU runtime :(

Do you know the problem? At the moment, I'm not sure if this a bug in alpaka or a problem of the driver of the CI runner. I would need to reproduce it locally to find it out.

AuroraPerego:
Something is broken in the runtime. I do not think is a bug in alpaka because all the tests pass with an older runtime (see #2140).

SimeonEhrig:
> Something is broken in the runtime. I do not think is a bug in alpaka because all the tests pass with an older runtime (see #2140).

I see. So ether we wait until Intel fixed the problem or we add an older additional OneAPI runtime version for runtime tests. Or is there any feature missing, which disallow this?

AuroraPerego:
As I wrote in the README, the only problem is that there is a bug in `all_of_group`/`any_of_group` methods (that does not affect the results of the tests). For that we have a workaround, for the `PI_ERROR_OUT_OF_RESOURCES` no and wait for Intel to fix it means potentially wait a lot.. I would go with adding an older runtime version for the tests

fwyzard:
(rebased to re-trigger the tests)

fwyzard:
I consistently see a failure from babelstream on SYCL CPU:
```
The following tests FAILED:
	 18 - babelstream (SEGFAULT)
```

SimeonEhrig:
> I consistently see a failure from babelstream on SYCL CPU:
> 
> ```
> The following tests FAILED:
> 	 18 - babelstream (SEGFAULT)
> ```

I will start implementing SYCL support in alpaka 2 the next days. During the work, I can have a look on this problem. I need to setup a SYCL CPU and GPU environment. Hopefully I can reproduce the problem locally.

fwyzard:
OK, with enough fixes it works :-)

Let's merge the other PRs first.

SimeonEhrig:
@fwyzard You need to approve because I opened the PR.

Merge after PR #2468

fwyzard (approved):

PR #2146, opened 2023-09-07
Improve handling of CMake generator expressions

j-stephan:
Fixes a bug introduced by #2135 (which makes it related to #2107) and also follows the recommended way of writing generator expressions.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -165,7 +165,7 @@ endif()
 if(MSVC)
     # warning C4505: '__cudaUnregisterBinaryUtil': unreferenced local function has been removed
     if(alpaka_ACC_GPU_CUDA_ONLY_MODE)
-        target_compile_options(alpaka INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/wd4505>)
+        target_compile_options(alpaka INTERFACE "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler /wd4505>")
     endif()
 else()
     # For std::future we need to pass the correct pthread flag for the compiler and the linker:
@@ -189,13 +189,13 @@ else()
 
     # Add debug optimization levels. CMake doesn't do this by default.
     # Note that -Og is the recommended gcc optimization level for debug mode but is equivalent to -O1 for clang (and its derivates).
-    alpaka_set_compiler_options(HOST_DEVICE target alpaka $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:-Og>
-                                                          $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CUDA>>:-Xcompiler -Og>
-                                                          $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:Clang,AppleClang,IntelLLVM>>:-O0>
-                                                          $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:MSVC>>:/Od>)
+    alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:SHELL:-Og>"
+                                                          "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CUDA>>:SHELL:-Xcompiler -Og>"
+                                                          "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:Clang,AppleClang,IntelLLVM>>:SHELL:-O0>"
+                                                          "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:MSVC>>:SHELL:/Od>")
     
-    target_link_options(alpaka INTERFACE $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>>:-Og>
-                                         $<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:Clang,AppleClang,IntelLLVM>>:-O0>) 
+    target_link_options(alpaka INTERFACE "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>>:SHELL:-Og>"
+                                         "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:Clang,AppleClang,IntelLLVM>>:SHELL:-O0>")
 endif()
 
 #-------------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


test/common/CMakeLists.txt
```diff
@@ -15,20 +15,18 @@ message(STATUS "alpaka_DEV_COMPILE_OPTIONS: ${alpaka_DEV_COMPILE_OPTIONS}")
...
 if(alpaka_ACC_GPU_CUDA_ENABLE)
     # CUDA driver API is used by EventHostManualTrigger
     target_link_libraries(${_COMMON_TARGET_NAME} INTERFACE CUDA::cuda_driver)
     target_compile_definitions(${_COMMON_TARGET_NAME} INTERFACE "CUDA_API_PER_THREAD_DEFAULT_STREAM")
 endif()
-if(alpaka_ACC_GPU_CUDA_ENABLE AND CMAKE_CUDA_COMPILER_ID STREQUAL "NVIDIA")
-    if(alpaka_ENABLE_WERROR)
-        target_compile_options(${_COMMON_TARGET_NAME} INTERFACE $<$<COMPILE_LANGUAGE:CUDA>:-Werror=all-warnings>)
-    endif()
-endif()
```
j-stephan, line 31: This section is handled by `devCompileOptions.cmake`.

cmake/alpakaCommon.cmake
```diff
@@ -423,67 +423,64 @@ if(alpaka_ACC_GPU_CUDA_ENABLE)
...
-                    alpaka_set_compiler_options(DEVICE target alpaka $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/openmp>)
+                    alpaka_set_compiler_options(DEVICE target alpaka "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler /openmp>")
                 endif()
             endif()
 
             # Always add warning/error numbers which can be used for suppressions
-            alpaka_set_compiler_options(DEVICE target alpaka $<$<COMPILE_LANGUAGE:CUDA>:-Xcudafe=--display_error_number>)
-
-            # avoids warnings on host-device signatured, default constructors/destructors
-            alpaka_set_compiler_options(DEVICE target alpaka $<$<COMPILE_LANGUAGE:CUDA>:-Xcudafe=--diag_suppress=esa_on_defaulted_function_ignored>)
```
j-stephan, line 473: We don't have any `default`ed methods which are annotated with `ALPAKA_FN{_HOST}_ACC` so I don't see a reason to keep this warning suppressed.

test/common/CMakeLists.txt
```diff
@@ -15,20 +15,18 @@ message(STATUS "alpaka_DEV_COMPILE_OPTIONS: ${alpaka_DEV_COMPILE_OPTIONS}")
 target_compile_options(${_COMMON_TARGET_NAME} INTERFACE ${alpaka_DEV_COMPILE_OPTIONS})
 
 if(MSVC)
-    target_compile_options(${_COMMON_TARGET_NAME} INTERFACE $<$<COMPILE_LANGUAGE:CXX>:/wd4996> $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/wd4996>) # This function or variable may be unsafe. Consider using <safe_version> instead.
-    target_compile_options(${_COMMON_TARGET_NAME} INTERFACE $<$<COMPILE_LANGUAGE:CXX>:/bigobj> $<$<COMPILE_LANGUAGE:CUDA>:-Xcompiler=/bigobj>)
+    # This function or variable may be unsafe. Consider using <safe_version> instead.
+    target_compile_options(${_COMMON_TARGET_NAME} INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:SHELL:/wd4996>"
+                                                            "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler /wd4996>")
+    target_compile_options(${_COMMON_TARGET_NAME} INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:SHELL:/bigobj>"
+                                                            "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler /bigobj>")
```
bernhardmgruber, line 22:
```suggestion
    target_compile_options(${_COMMON_TARGET_NAME} INTERFACE "$<$<COMPILE_LANGUAGE:CXX>:SHELL:/wd4996 /bigobj>"
                                                            "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:-Xcompiler /wd4996 -Xcompiler /bigobj>")
```
Please avoid duplicated logic.

test/common/devCompileOptions.cmake
```diff
@@ -8,48 +8,54 @@
...
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:--Wreorder>)
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:--Wdefault-stream-launch>)
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:--Wmissing-launch-bounds>)
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:--Wext-lambda-captures-this>)
-        list(APPEND alpaka_DEV_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CUDA>:--Werror all-warnings>)
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wreorder>")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wdefault-stream-launch>")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wmissing-launch-bounds>")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Wext-lambda-captures-this>")
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "$<$<COMPILE_LANGUAGE:CUDA>:SHELL:--Werror all-warnings>")
```
bernhardmgruber, line 15: I think we can also combine all of those into one generator expression. Applies to lots of other places as well.

test/unit/math/CMakeLists.txt
```diff
@@ -23,11 +23,11 @@ target_include_directories(
...
-set_target_properties(
+target_compile_options(
     ${_TARGET_NAME}
-    PROPERTIES
-    COMPILE_OPTIONS
-    $<$<CXX_COMPILER_ID:clang>:"-ffp-contract=off"> # ffp-contract: https://llvm.org/docs/CompileCudaWithLLVM.html#id5
+    PRIVATE
+    # ffp-contract: https://llvm.org/docs/CompileCudaWithLLVM.html#flags-that-control-numerical-code
+    "$<$<COMPILE_LANG_AND_ID:CUDA,AppleClang,Clang,IntelLLVM>:SHELL:-ffp-contract=off>"
     )
```
bernhardmgruber, line 31: So this is not a problem with icpx compiling ordinary C++?

j-stephan: I don't know. The previous version only seemed to care about CUDA so I did as well.

test/common/CMakeLists.txt
j-stephan: I'm just following the original layout here: One warning per line. I guess this was intended for better maintenance of this file.

test/common/devCompileOptions.cmake
j-stephan: Again, one warning per line seems to be the original intention of the file.

test/common/CMakeLists.txt
bernhardmgruber: That's what I thought. I had just hoped you could also push a little cleanup as well. It is harder to read for me with the redundancy, because I actually need to check whether all lines do really start the same ..

test/unit/math/CMakeLists.txt
bernhardmgruber: Where do you see that the previous version only cared about CUDA? It just checks for whether the compiler id is clang. That comment next to it could have just been outdated and added when we did not support icpx yet.

But the even bigger question: is that flag even necessary, since the CI seems to pass for icpx. We should revisit this.

bernhardmgruber (approved):

PR #2150, opened 2023-09-08
Use CMake's native HIP support

### LINKED ISSUE #1462: Use CMake's native HIP support
CMake 3.21 [introduced native support for HIP](https://cmake.org/cmake/help/v3.21/release/3.21.html#id9). We should check this out as soon as we can migrate to a CMake version >= 3.21.

j-stephan:
This PR refactors the HIP portion of our CMake scripts to utilize CMake's native HIP support.

Fixes #1462.


bernhardmgruber (commented):
In principle, LGTM. But I want a second opinion.

j-stephan:
I'm shifting this to the `alpaka-next` project. Because of a bug the native CMake support for ROCm 5.0 is broken and I don't want to remove hardware support so close to the release.

psychocoderHPC:
@j-stephan Could you please rebase this PR that it can run the CI again, since we have a release branch there should be no reason anymore to not merge this PR if it passes the CI.

cmake/addLibrary.cmake
```diff
@@ -36,4 +36,23 @@ macro(alpaka_add_library libraryName)
             set_property(TARGET ${In_Name} PROPERTY CUDA_SEPARABLE_COMPILATION OFF)
         endif()
     endif()
+
+    if(alpaka_ACC_GPU_HIP_ENABLE)
+        enable_language(HIP)
+        foreach(_file ${ARGN})
+            if((${_file} MATCHES "\\.cpp$") OR
+               (${_file} MATCHES "\\.cxx$") OR
+               (${_file} MATCHES "\\.cu$") OR
```
StewMH: Unless I miss something you probably don't want .cu to be marked as HIP?

cmake/addExecutable.cmake
```diff
@@ -35,4 +35,23 @@ macro(alpaka_add_executable In_Name)
             set_property(TARGET ${In_Name} PROPERTY CUDA_SEPARABLE_COMPILATION OFF)
         endif()
     endif()
+
+    if(alpaka_ACC_GPU_HIP_ENABLE)
+        enable_language(HIP)
+        foreach(_file ${ARGN})
+            if((${_file} MATCHES "\\.cpp$") OR
+               (${_file} MATCHES "\\.cxx$") OR
+               (${_file} MATCHES "\\.cu$") OR
```
StewMH: same here, probably not cu

SimeonEhrig: Thanks for the hint. Looks like a copy paste error. I will overtake the PR from @j-stephan  since he is not available for the next time. Let's see, if I find some time this week.

SimeonEhrig:
In a offline discussion with @psychocoderHPC , we decide:
* HIP 5.0 support should be removed, if nobody is using it
* I try to fix the bug for HIP 5.1 . If it is to complicated, maybe we also remove the support

SimeonEhrig:
@fwyzard Can we drop the support for HIP 5.0? If I remember correctly, LUMI is still using HIP 5.2 therefore I'm not sure, if you need HIP 5.0.

fwyzard:
Yes, we can drop 5.0 (and 5.1, if we have it).

SimeonEhrig:
> Yes, we can drop 5.0 (and 5.1, if we have it).

Great. We don't need to drop 5.1. There was only a small bug, which I fixed. HIP 5.0 triggers a bug in CMake, which is at least hard to fix.

psychocoderHPC:
This PR is finished #2215 and uses native CMake HIP support.

PR #2151, opened 2023-09-08
Remove unnecessary member initialization calls

bernhardmgruber:
This should be a non-functional change.

### DIFF
--- a/include/alpaka/acc/AccGenericSycl.hpp
+++ b/include/alpaka/acc/AccGenericSycl.hpp
@@ -81,16 +81,10 @@ namespace alpaka
             : WorkDivGenericSycl<TDim, TIdx>{threadElemExtent, work_item}
             , gb::IdxGbGenericSycl<TDim, TIdx>{work_item}
             , bt::IdxBtGenericSycl<TDim, TIdx>{work_item}
-            , AtomicHierarchy<AtomicGenericSycl, AtomicGenericSycl, AtomicGenericSycl>{}
-            , math::MathGenericSycl{}
             , BlockSharedMemDynGenericSycl{dyn_shared_acc}
             , BlockSharedMemStGenericSycl{st_shared_acc}
             , BlockSyncGenericSycl<TDim>{work_item}
-            , IntrinsicGenericSycl{}
-            , MemFenceGenericSycl{}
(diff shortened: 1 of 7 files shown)


include/alpaka/acc/AccGenericSycl.hpp
```diff
@@ -81,16 +81,10 @@ namespace alpaka
...
             , BlockSharedMemStGenericSycl{st_shared_acc}
             , BlockSyncGenericSycl<TDim>{work_item}
-            , IntrinsicGenericSycl{}
-            , MemFenceGenericSycl{}
-#    ifdef ALPAKA_DISABLE_VENDOR_RNG
-            , rand::RandDefault{}
-#    else
+#    ifndef ALPAKA_DISABLE_VENDOR_RNG
             , rand::RandGenericSycl<TDim>{work_item}
 #    endif
```
fwyzard, line 89: I understand you wanted to keep only the strictly necessary initialisations.
Just for this case where one of the two `#ifdef` branches requires initialisation, would it make sense to keep also the other ?

bernhardmgruber: It does not make sense for me. I would only keep the code that really runs something.

fwyzard: So... I should go ahead and remove all the comments and all the code that is there only to silence warnings ?

fwyzard: Anyway, whatever.

bernhardmgruber: > So... I should go ahead and remove all the code that is there only to silence warnings ?

I also dislike that sort of code. I am saddened that we have to add it to workaround compiler warnings.

But I am always happy if we can agree on dropping older compilers to remove some of that baggage!

j-stephan: IMO this is a question of style / preference. Let us discuss this in tomorrow's meeting.

SimeonEhrig (approved):

include/alpaka/acc/AccGenericSycl.hpp
bernhardmgruber: Discussed during today's alpaka meeting: no objections.

PR #2153, opened 2023-09-08
Enable OpenMP back-ends for clang-cuda

### LINKED ISSUE #639: Why clang(CUDA) + OpenMP is not supported?

j-stephan:
Fixes #639.


bernhardmgruber (dismissed):

j-stephan:
I'm putting this into the `alpaka-next` release. Every fix I make just produces two more issues and warnings. I'm honestly not sure if we should even attempt to fix this since the list of work-arounds just gets longer and longer...

bernhardmgruber:
With a quick look at the diff, yeah, let's maybe push this off into the future.

(I do see the combination of CUDA+OpenMP with one compiler as useful though, so let's not drop the idea)

PR #847, opened 2019-09-24
NEW Added tests for existing math operators

jkrude:
Rework of #828
Tests were run on a cuda-device and std-backend.
See also https://github.com/ComputationalRadiationPhysics/alpaka/pull/828 for the previous version.
Implemented functors and abstraction, so that one code is used for all operators.

PR #2155, opened 2023-09-11
Remove FP contraction for math test

bernhardmgruber:
This PR removes the flag `-ffp-contract=off` from the math tests for certain compilers. It's not needed anymore (although I made no attempt to find out why).


j-stephan (approved):
Interesting, it looks like we never wanted this in the first place: https://github.com/alpaka-group/alpaka/pull/847#discussion_r339738972

Somehow it was still merged in.

PR #2157, opened 2023-09-12
Address `nvcc` + `clang++` issues

j-stephan:
This PR fixes another set of warnings / errors uncovered by #2107:

* Silences warnings caused by nvcc-generated code which `clang++ -Weverything` doesn't like.
* Adds `ALPAKA_UNREACHABLE` to `extents/Traits.hpp`.

### DIFF
--- a/include/alpaka/extent/Traits.hpp
+++ b/include/alpaka/extent/Traits.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "alpaka/core/Common.hpp"
+#include "alpaka/core/Unreachable.hpp"
 #include "alpaka/dim/DimIntegralConst.hpp"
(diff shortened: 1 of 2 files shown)


include/alpaka/extent/Traits.hpp
```diff
@@ -97,6 +98,8 @@ namespace alpaka
             return getExtents(extent)[Dim<TExtent>::value - 1u];
         else
             return 1;
+
+        ALPAKA_UNREACHABLE({});
```
bernhardmgruber, line 102: Man, I despise those. I think this is only a problem in older nvcc versions. I hope we can drop those soon!

bernhardmgruber (approved):

PR #2162, opened 2023-09-14
refactor template order `allocMappedBufIfSupported`

psychocoderHPC:
Move template `TPlatform` as the last template. There is no need to provide the platform template signature if we pass the platform as an instance.

@j-stephan This PR should be cherry-picked into the current open-release branch!

### DIFF
--- a/include/alpaka/mem/buf/Traits.hpp
+++ b/include/alpaka/mem/buf/Traits.hpp
@@ -164,14 +164,14 @@ namespace alpaka
     //! this function is provided for convenience in the cases where the difference is not relevant,
     //! and the pinned/mapped memory is only used as a performance optimisation.
     //!
-    //! \tparam TPlatform The platform from which the buffer is accessible.
     //! \tparam TElem The element type of the returned buffer.
     //! \tparam TIdx The linear index type of the buffer.
     //! \tparam TExtent The extent type of the buffer.
+    //! \tparam TPlatform The platform from which the buffer is accessible.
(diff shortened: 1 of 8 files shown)


include/alpaka/mem/buf/Traits.hpp
```diff
@@ -164,15 +164,15 @@ namespace alpaka
...
     //! \tparam TIdx The linear index type of the buffer.
     //! \tparam TExtent The extent type of the buffer.
+    //! \tparam TPlatform The platform from which the buffer is accessible.
     //! \param host The host device to allocate the buffer on.
     //! \param extent The extent of the buffer.
     //! \return The newly allocated buffer.
-    template<typename TPlatform, typename TElem, typename TIdx, typename TExtent>
-    ALPAKA_FN_HOST auto allocMappedBufIfSupported(
+    template<typename TElem, typename TIdx, typename TExtent, typename TPlatform>
+    ALPAKA_FN_HOST static auto allocMappedBufIfSupported(
```
bernhardmgruber: Why do you need the `static` here? It should not be needed.
```suggestion
    template<typename TElem, typename TIdx, typename TExtent, typename TPlatform>
    ALPAKA_FN_HOST auto allocMappedBufIfSupported(
```

psychocoderHPC: Misunderstanding from my side. I will remove it.

bernhardmgruber (approved):

PR #2164, opened 2023-09-15
Fix a few warnings

bernhardmgruber:
* Removed unused aliases triggering warnings
* Only add clang warning flags when the compiler version supports them, avoiding warnings
* Workaround nvcc warnings on missing return statements when using `if constexpr`.


psychocoderHPC (approved):

psychocoderHPC:
Could you please add some description what kind of warning will be fixed and for which compiler.

bernhardmgruber:
> Could you please add some description what kind of warning will be fixed and for which compiler.

Done. Thx!

PR #2165, opened 2023-09-15
Workaround gcc warning on uninitialized PlatformCpu

bernhardmgruber:
This is a workaround for g++-11 bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96295
g++-11 complains in *all* places where a PlatformCpu is used, that it "may be used uninitialized"

### DIFF
--- a/include/alpaka/platform/PlatformCpu.hpp
+++ b/include/alpaka/platform/PlatformCpu.hpp
@@ -16,6 +16,12 @@ namespace alpaka
     //! The CPU device platform.
     struct PlatformCpu : concepts::Implements<ConceptPlatform, PlatformCpu>
     {
+#if defined(BOOST_COMP_GNUC) && BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(11, 0, 0)                                     \
+    && BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(12, 0, 0)
+        // This is a workaround for g++-11 bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96295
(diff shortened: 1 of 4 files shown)


include/alpaka/platform/PlatformCpu.hpp
```diff
@@ -16,6 +16,12 @@ namespace alpaka
     //! The CPU device platform.
     struct PlatformCpu : concepts::Implements<ConceptPlatform, PlatformCpu>
     {
+#if defined(BOOST_COMP_GNUC) && BOOST_COMP_GNUC >= BOOST_VERSION_NUMBER(11, 0, 0)                                     \
+    && BOOST_COMP_GNUC < BOOST_VERSION_NUMBER(12, 0, 0)
+        // This is a workaround for g++-11 bug: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=96295
+        // g++-11 complains in *all* places where a PlatformCpu is used, that it "may be used uninitialized"
+        char c = {};
```
fwyzard, line 23: according to the bug report, it should be enough to add
```c++
PlatformCpu() {}
```
instead of adding a data member.

bernhardmgruber: That's what the bug report says, but unfortunately it did not work in our case and the warning persisted.

psychocoderHPC (approved):

fwyzard:
I would have preferred a workaround that does not involve adding spurious data members.

bernhardmgruber:
> I would have preferred a workaround that does not involve adding spurious data members.

I understand that my proposed workaround feels unorthodox. I will gladly review a PR with a better solution!

PR #2168, opened 2023-09-18
Strip namespaces from mathTest

bernhardmgruber:
Just ran into the deeply nested namespaces inside the `mathTest` and decided to remove them to save indentation.

### DIFF
--- a/test/unit/math/src/Buffer.hpp
+++ b/test/unit/math/src/Buffer.hpp
@@ -10,101 +10,91 @@
 
 #include <ostream>
 
-namespace alpaka
+namespace mathtest
 {
-    namespace test
+    //! Provides alpaka-style buffer with arguments' data.
+    //! TData can be a plain value or a complex data-structure.
+    //! The operator() is overloaded and returns the value from the correct Buffer,
+    //! either from the host (index) or device buffer (index, acc).
+    //! Index out of range errors are not checked.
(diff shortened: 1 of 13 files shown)


bernhardmgruber:
Make sure you hide whitespace changes when reviewing!

fwyzard:
What about the other `alpaka::test` namespace usage ?

bernhardmgruber:
> What about the other `alpaka::test` namespace usage ?

My PR just concerns the `mathTest` internals. I think `alpaka::test` is used more widespread. Especially also in public headers of the alpaka library itself.

psychocoderHPC:
Is it really a good idea to move all code in the hpp-files into the global namespace? There is a small possibility of naming conflicts.
I tend to have my own code always encapsulated within a namespace.

bernhardmgruber:
> Is it really a good idea to move all code in the hpp-files into the global namespace? There is a small possibility of naming conflicts. I tend to have my own code always encapsulated within a namespace.

My thought process was to reduce all the nesting to a nested namespace specifier. Then I figured it is still too long when prefixing entities from the nested namespace. I looked into other unit tests where I saw no namespaces being used. Then I decided to remove them.

I honestly don't care whether we have no namespace or a small one for local header files in unit tests. From the best practices, maybe having a namespace is better than declaring in the global namespace. I can add one then.

psychocoderHPC (approved):

Issue #2074, opened 2023-08-10
Add a test using `alpaka_add_library`


bernhardmgruber:
> Do we have a test for an alpaka library by any chance? If not, we should open an issue.

We don't. We are also not using `alpaka_add_library` anywhere in this repository. Maybe PIConGPU or cupla make use of it?

_Originally posted by @j-stephan in https://github.com/alpaka-group/alpaka/issues/2072#issuecomment-1672825182_

bernhardmgruber:
Thinking further about this, can we actually enable/disable accelerators for different libraries (but maybe sharing the same implementation)? Like:

```
enable CUDA
alpaka_add_library(cuda lib.cpp)
disable CUDA
enable OMP2
alpaka_add_library(omp lib.cpp) // same source
disable OMP2
add_executable(program main.cpp)
target_link_libraries(program cuda omp)
```

j-stephan:
I don't think so (but maybe I'm wrong). But I had an idea for an alternative approach based on source file properties which might allow us to get rid of these special functions and just use the native `add_{library, executable}` directly.

bernhardmgruber:
In any case, we should have a small example involving multiple libraries that show how a larger application with alpaka could be struktured. I only thought, since CMS also compiles a library with alpaka multiple times with different backends to dynamically load the right one from a driver executable, that this may be an interesting small test.

SimeonEhrig:
Vikunja is testing `alpaka_add_library` in a special integration test: https://github.com/alpaka-group/vikunja/blob/9613b24ef5f9e0d66ef6d33e1781700c38d6f56c/ci/gitlab_scripts/integration_test/addsubdirectory.cmake#L47

In general, I implemented to different integration tests, to test the vikunja cmake with an external sample project. Maybe we should overtake the approach for alpaka.

PR #2175, opened 2023-09-20
Add alpaka-ls

bernhardmgruber:
I needed to print the devices available for an accelerator, so I added a ~~test~~ small tool called `alpaka-ls` as example. ~~But maybe this should just be an example. What do you think?~~


bernhardmgruber:
Thinking about this again, SYCL has a utility called `sycl-ls`, which for example prints this on my system:
```
bgruber@graviola:~$ sycl-ls
[opencl:acc:0] Intel(R) FPGA Emulation Platform for OpenCL(TM), Intel(R) FPGA Emulation Device 1.2 [2023.16.7.0.21_160000]
[opencl:cpu:1] Intel(R) OpenCL, AMD Ryzen 9 5950X 16-Core Processor             3.0 [2023.16.7.0.21_160000]
```
I think it could be useful to have an `alpaka-ls`, that lists similar information.

fwyzard:
That's a good idea, and it could also be a good project for a new student.

bernhardmgruber:
So, will this PR be merged or replaced by an example?

SimeonEhrig:
Actual I would need the `alpaka-ls` tool for this issue: https://github.com/alpaka-group/alpaka/issues/2074

It was thinking about to implement CMake integration tests to test if our CMake is correctly working. Therefore I would implement a test, which builds `alpaka-ls` with several enabled backends. Then a bash script would execute it and check if the output matches with the enabled backends.

The background is, that I want to overwork the CMake totally, to support different correctly CMake targets and allows to enable all backends in parallel and compile a translation unit with a specific backend.

bernhardmgruber:
Alright, here is the tool. Prints e.g. on my machine:
```
AccCpuSerial<1,int>
	AMD Ryzen 9 5950X 16-Core Processor
AccCpuOmp2Blocks<1,int>
	AMD Ryzen 9 5950X 16-Core Processor
AccGpuCudaRt<1,int>
	NVIDIA GeForce RTX 2060
```

SimeonEhrig:
On HAL

```
AccCpuSerial<1,int>
	AMD EPYC 7452 32-Core Processor
AccCpuThreads<1,int>
	AMD EPYC 7452 32-Core Processor
AccGpuHipRt<1,int>
	AMD Radeon VII
	AMD Radeon VII
```

and

```
AccCpuSerial<1,int>
	AMD EPYC 7452 32-Core Processor
AccCpuThreads<1,int>
	AMD EPYC 7452 32-Core Processor
AccCpuOmp2Blocks<1,int>
	AMD EPYC 7452 32-Core Processor
AccCpuOmp2Threads<1,int>
	AMD EPYC 7452 32-Core Processor
AccGpuCudaRt<1,int>
	NVIDIA A30
```

Next week, also the Intel GPU is coming. So my goal after the renovation of the CMake is to display all 3 different GPUs with `alpaka-ls` :-)

example/ls/CMakeLists.txt
```diff
@@ -0,0 +1,21 @@
...
+        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../.." "${CMAKE_BINARY_DIR}/alpaka")
+    else()
+        find_package(alpaka REQUIRED)
+    endif()
+endif()
+
+alpaka_add_executable(${PROJECT_NAME} src/ls.cpp)
+target_link_libraries(${PROJECT_NAME} PUBLIC alpaka::alpaka)
+set_target_properties(${PROJECT_NAME} PROPERTIES FOLDER example)
+add_test(NAME ${PROJECT_NAME} COMMAND ${CMAKE_PROJECT_NAME})
```
SimeonEhrig:
```suggestion
add_test(NAME ${PROJECT_NAME} COMMAND ${PROJECT_NAME})
```

Fixes the broken test.

bernhardmgruber:
@SimeonEhrig thx a lot for testing!

example/ls/CMakeLists.txt
bernhardmgruber: Applied

psychocoderHPC (approved):

PR #2179, opened 2023-09-27
Remove unnecessary `-fintelfpga` flag

j-stephan:
I tested the `1.0.0-rc1` branch on Intel's DevCloud and noticed that `-fintelfpga` now conflicts with the `-fsycl-targets=spir64_fpga` flag. According to the [documentation](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-2/fintelfpga.html) those two are equivalent so I removed `-fintelfpga`. I couldn't do further tests because Intel's DevCloud doesn't provide an up-to-date installation of oneAPI (most recent version is 2023.0 - alpaka requires at least 2023.1).

This will also be cherry-picked into the `1.0.0-rc1` branch.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -614,7 +614,6 @@ if(alpaka_ACC_SYCL_ENABLE)
 
         if(alpaka_SYCL_ONEAPI_FPGA)
             target_compile_definitions(alpaka INTERFACE "ALPAKA_SYCL_ONEAPI_FPGA")
-            alpaka_set_compiler_options(DEVICE target alpaka "-fintelfpga")
 
             if(alpaka_SYCL_ONEAPI_FPGA_MODE STREQUAL "emulation")
                 target_compile_definitions(alpaka INTERFACE "ALPAKA_FPGA_EMULATION")


bernhardmgruber:
> Intel's DevCloud doesn't provide an up-to-date installation of oneAPI

:D

bernhardmgruber (approved):

fwyzard:
> I couldn't do further tests because Intel's DevCloud doesn't provide an up-to-date installation of oneAPI (most recent version is 2023.0 - alpaka requires at least 2023.1).

Last I checked they had 2023.2.1, which was the latest version.

Just now I tried to sign in and they asked me to pay...

j-stephan:
Interesting. This afternoon `module avail` only listed some 2022.x versions and 2023.0. I wonder whether they changed their modules or if the available modules are maybe different on FPGA nodes.

j-stephan:
`module avail` on FPGA nodes:

```
------------------------------------------------------------------------------------- /glob/module-files/intel-oneapi --------------------------------------------------------------------------------------
2022.1.2  2022.3.1  2023.0
```

`module avail` on GPU nodes:

```
------------------------------------------------------------------------------------- /glob/module-files/intel-oneapi --------------------------------------------------------------------------------------
2022.1.2  2022.3.1  2023.0
```

The individual modules like `compiler` also stop at 2023.0. So it seems they rolled back the available versions for some reason.

PR #2196, opened 2023-11-27
Check for HIP versions in min/max range

StewMH:
With the current logic, cmake checks first for HIP 5.5, then for 5.0:
```
CMake Warning at cmake/alpakaCommon.cmake:506 (find_package):
  By not providing "Findhip.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "hip", but
  CMake did not find one.

  Could not find a package configuration file provided by "hip" (requested
  version 5.5) with any of the following names:

    hipConfig.cmake
    hip-config.cmake

  Add the installation prefix of "hip" to CMAKE_PREFIX_PATH or set "hip_DIR"
  to a directory containing one of the above files.  If "hip" provides a
  separate development package or SDK, be sure it has been installed.
  CMakeLists.txt:94 (include)


-- Could not find HIP v5.5. Now searching for HIP v5.0
CMake Warning at cmake/alpakaCommon.cmake:509 (find_package):
  By not providing "Findhip.cmake" in CMAKE_MODULE_PATH this project has
  asked CMake to find a package configuration file provided by "hip", but
  CMake did not find one.

  Could not find a package configuration file provided by "hip" (requested
  version 5.0) with any of the following names:

    hipConfig.cmake
    hip-config.cmake

  Add the installation prefix of "hip" to CMAKE_PREFIX_PATH or set "hip_DIR"
  to a directory containing one of the above files.  If "hip" provides a
  separate development package or SDK, be sure it has been installed.
  CMakeLists.txt:94 (include)


CMake Error at cmake/alpakaCommon.cmake:514 (message):
  Optional alpaka dependency HIP could not be found!
  CMakeLists.txt:94 (include)
```

I think the behaviour should give an inclusive range but I don't immediately have a way to test this (only having a somewhat broken HIP 5.1 installation).

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -503,11 +503,7 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
     # supported HIP version range
     set(_alpaka_HIP_MIN_VER 5.0)
     set(_alpaka_HIP_MAX_VER 5.5)
-    find_package(hip "${_alpaka_HIP_MAX_VER}")
-    if(NOT TARGET hip)
-        message(STATUS "Could not find HIP v${_alpaka_HIP_MAX_VER}. Now searching for HIP v${_alpaka_HIP_MIN_VER}")
-        find_package(hip "${_alpaka_HIP_MIN_VER}")
-    endif()
+    find_package(hip "${_alpaka_HIP_MIN_VER}...<${_alpaka_HIP_MAX_VER}")
 
     if(NOT TARGET hip)
         message(FATAL_ERROR "Optional alpaka dependency HIP could not be found!")
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -506,7 +506,7 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
     find_package(hip "${_alpaka_HIP_MIN_VER}...${_alpaka_HIP_MAX_VER}")
 
     if(NOT TARGET hip)
-        message(STATUS "Could not find HIP <=v${_alpaka_HIP_MAX_VER}. Now searching for unsupported HIP >v${_alpaka_HIP_MAX_VER}")
+        message(WARNING "Could not find HIP <=v${_alpaka_HIP_MAX_VER}. Now searching for unsupported HIP >v${_alpaka_HIP_MAX_VER}")
         find_package(hip "${_alpaka_HIP_MAX_VER}")
     endif()
 
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -503,7 +503,12 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
     # supported HIP version range
     set(_alpaka_HIP_MIN_VER 5.0)
     set(_alpaka_HIP_MAX_VER 5.5)
-    find_package(hip "${_alpaka_HIP_MIN_VER}...<${_alpaka_HIP_MAX_VER}")
+    find_package(hip "${_alpaka_HIP_MIN_VER}...${_alpaka_HIP_MAX_VER}")
+
+    if(NOT TARGET hip)
+        message(STATUS "Could not find HIP <=v${_alpaka_HIP_MAX_VER}. Now searching for unsupported HIP >v${_alpaka_HIP_MAX_VER}")
+        find_package(hip "${_alpaka_HIP_MAX_VER}")
+    endif()
 
     if(NOT TARGET hip)
         message(FATAL_ERROR "Optional alpaka dependency HIP could not be found!")


cmake/alpakaCommon.cmake
```diff
@@ -503,11 +503,7 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
     # supported HIP version range
     set(_alpaka_HIP_MIN_VER 5.0)
     set(_alpaka_HIP_MAX_VER 5.5)
-    find_package(hip "${_alpaka_HIP_MAX_VER}")
-    if(NOT TARGET hip)
-        message(STATUS "Could not find HIP v${_alpaka_HIP_MAX_VER}. Now searching for HIP v${_alpaka_HIP_MIN_VER}")
-        find_package(hip "${_alpaka_HIP_MIN_VER}")
-    endif()
+    find_package(hip "${_alpaka_HIP_MIN_VER}...<${_alpaka_HIP_MAX_VER}")
```
psychocoderHPC: Nice, this feature has been available since CMake 3.19 and we forgot about it after we required a newer version of CMake.

psychocoderHPC: `<` should be removed else the max version will be excluded. see https://cmake.org/cmake/help/latest/command/find_package.htm

cmake/alpakaCommon.cmake
```diff
@@ -503,11 +503,7 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
     # supported HIP version range
     set(_alpaka_HIP_MIN_VER 5.0)
     set(_alpaka_HIP_MAX_VER 5.5)
-    find_package(hip "${_alpaka_HIP_MAX_VER}")
-    if(NOT TARGET hip)
```
psychocoderHPC, line 507: Please keep this additional search. In case there will be a newer version of HIP available and we have not tested this version the user should still be able to use it even if we have not tested the version. 

Could you please rephrase the message to something e.g.

`message(STATUS "Could not find HIP <=v${_alpaka_HIP_MAX_VER}. Now searching for unsopported HIP >v${_alpaka_HIP_MIN_VER}")`

psychocoderHPC (changes requested):

cmake/alpakaCommon.cmake
StewMH: Thanks, I hadn't quite followed the < documentation.

cmake/alpakaCommon.cmake
```diff
@@ -503,10 +503,11 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
     # supported HIP version range
     set(_alpaka_HIP_MIN_VER 5.0)
     set(_alpaka_HIP_MAX_VER 5.5)
-    find_package(hip "${_alpaka_HIP_MAX_VER}")
+    find_package(hip "${_alpaka_HIP_MIN_VER}...${_alpaka_HIP_MAX_VER}")
```
SimeonEhrig, line 506: The upper bound is excluded. Therefore `_alpaka_HIP_MAX_VER` needs to be `5.6`. Maybe we should also change the name of the variable.

https://cmake.org/cmake/help/latest/command/find_package.html#id7

I tested it with ROCm 5.4.6, 5.5.0, 5.5.3, 5.6 and 5.7. It is working in general. Only the warn message for ROCm 5.5.x is wrong.

cmake/alpakaCommon.cmake
```diff
@@ -503,10 +503,11 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
...
+
     if(NOT TARGET hip)
-        message(STATUS "Could not find HIP v${_alpaka_HIP_MAX_VER}. Now searching for HIP v${_alpaka_HIP_MIN_VER}")
-        find_package(hip "${_alpaka_HIP_MIN_VER}")
+        message(STATUS "Could not find HIP <=v${_alpaka_HIP_MAX_VER}. Now searching for unsupported HIP >v${_alpaka_HIP_MAX_VER}")
```
SimeonEhrig:
```suggestion
        message(WARNING "Could not find HIP <=v${_alpaka_HIP_MAX_VER}. Now searching for unsupported HIP >v${_alpaka_HIP_MAX_VER}")
```

SimeonEhrig (commented):
I could verify it with ROCm 5.5 and ROCm 5.7 on our dev system.

cmake/alpakaCommon.cmake
StewMH: Interesting, the find_package documentation suggests we should be getting an inclusive range.
```
A version range with the format versionMin...[<]versionMax where versionMin and versionMax have the same format and constraints on components being integers as the single version. By default, both end points are included. By specifying <, the upper end point will be excluded. 
```
Which CMake do you have?

SimeonEhrig: Your are right. I missed the part that the upper bound is only excluded if the `<` is set. I used CMake 3.27.4

SimeonEhrig (approved):
I checked it again locally. For unknown reason the warning with HIP 5.5 and different CMake versions appears only on my dev system. The CI is fine and does not trow a warning. Also the documentation says, that the upper version should be included. Therefor I will merge it :-)

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #2198, opened 2023-12-01
add trait `IsKernelArgumentTriviallyCopyable`

### LINKED ISSUE #2195: std::is_trivially_copyable too strong as a constraint
The line

https://github.com/alpaka-group/alpaka/blob/4c981f0a1aee4cd59edb8ebc70399249409bf0cd/include/alpaka/kernel/Traits.hpp#L227

constrains types to copyable ones by just using `std::is_trivially_copyable`. Our users sometimes have types, which are trivially copyable, but don't fulfill the requirements of `std::is_trivially_copyable`. It is not allowed to specialize `std::is_trivially_copyable` itself.
Is it possible to get a trait for this property, so that a user is able to specify trivial copyability for particular types?

psychocoderHPC:
fix #2195

Provide an alpaka specific trait to validate kernel argument conditions which can be specialized by the user for their own types and at their own risk.

### DIFF
--- a/include/alpaka/kernel/Traits.hpp
+++ b/include/alpaka/kernel/Traits.hpp
@@ -202,6 +202,29 @@ namespace alpaka
         "-Wdocumentation" // clang does not support the syntax for variadic template arguments "args,..."
 #endif
 
+
+    //! Check if a type used as kernel argument is trivially copyable
+    //!
+    //! \attention In case this trait is specialized for a user type the user should be sure that the result of calling
+    //! the copy constructor is equal to use memcpy to duplicate the object. An existing destructor should be free
+    //! of side effects.
+    //!
+    //! It's implementation defined whether the closure type of a lambda is trivially copyable.
+    //! Therefor the default implementation is true for trivially copyable or empty (stateless) types.
+    //!
+    //! @tparam T type to check
+    //! @{
+    template<typename T, typename = void>
+    struct IsKernelArgumentTriviallyCopyable
+        : std::bool_constant<std::is_empty_v<T> || std::is_trivially_copyable_v<T>>
+    {
+    };
+
+    template<typename T>
+    inline constexpr bool isKernelArgumentTriviallyCopyable = IsKernelArgumentTriviallyCopyable<T>::value;
+
+    //! @}
+
     namespace detail
     {
         //! Check that the return of TKernelFnObj is void
@@ -221,13 +244,8 @@ namespace alpaka
         template<typename T>
         inline void assertKernelArgIsTriviallyCopyable()
         {
-            static_assert(
-                // it's implementation defined whether the closure type of a lambda is trivially copyable. But they
-                // should be at least empty then (stateless).
-                std::is_empty_v<T> || std::is_trivially_copyable_v<T>,
-                "The kernel argument T must be trivially copyable!");
+            static_assert(isKernelArgumentTriviallyCopyable<T>, "The kernel argument T must be trivially copyable!");
         }
-
     } // namespace detail
 
 //! Creates a kernel execution task.


include/alpaka/kernel/Traits.hpp
```diff
@@ -202,6 +202,29 @@ namespace alpaka
...
+    /// \attention In case this trait is specialized for a user type the user should be sure that the result of calling
+    /// the copy constructor is equal to use memcopy to duplicate the object. An existing destructor should be side
+    /// effect free.
+    ///
+    /// It's implementation defined whether the closure type of a lambda is trivially copyable.
+    /// Therefor the default implementation is true for trivially copyable or empty (stateless) types.
+    ///
+    /// @tparam T type to check
+    /// @{
+    template<typename T>
```
MaxSagebaum: Adding an additional template parameter `typename = void` makes specialization usually simpler.

```suggestion
    template<typename T, typename = void>
```

Instead of creating a specialization for every type, it can be generalized like:
```
template<typename Real>
struct IsKernelArgumentTriviallyCopyable<Real, typename std::enable_if<codi::isForwardType<Real>::value>::type
```

fwyzard: @MaxSagebaum thanks for the suggestion, I think it makes sense

include/alpaka/kernel/Traits.hpp
```diff
@@ -202,6 +202,29 @@ namespace alpaka
...
+    struct IsKernelArgumentTriviallyCopyable
+        : std::bool_constant<std::is_empty_v<T> || std::is_trivially_copyable_v<T>>
+    {
+    };
+
+    template<typename T>
+    constexpr bool IsKernelArgumentTriviallyCopyable_v = IsKernelArgumentTriviallyCopyable<T>::value;
```
fwyzard: Following the styles of the other alpaka traits (like `isQueue<T>`, _etc_.) this should probably be
```suggestion
    inline constexpr bool isKernelArgumentTriviallyCopyable = IsKernelArgumentTriviallyCopyable<T>::value;
```

fwyzard (changes requested):
@psychocoderHPC I would follow the suggestion from @MaxSagebaum , and also change the variable template name to `isKernelArgumentTriviallyCopyable<T>`.

include/alpaka/kernel/Traits.hpp
```diff
@@ -202,6 +202,29 @@ namespace alpaka
         "-Wdocumentation" // clang does not support the syntax for variadic template arguments "args,..."
 #endif
 
+
+    /// Check if a type used as kernel argument is trivially copyable
+    ///
+    /// \attention In case this trait is specialized for a user type the user should be sure that the result of calling
+    /// the copy constructor is equal to use memcopy to duplicate the object. An existing destructor should be side
+    /// effect free.
```
bernhardmgruber:
```suggestion
    /// \attention In case this trait is specialized for a user type the user should be sure that the result of calling
    /// the copy constructor is equal to use memcpy to duplicate the object. An existing destructor should be free
    /// of side effects.
```

include/alpaka/kernel/Traits.hpp
bernhardmgruber: You guys are going to love `requires` clauses in C++20.

include/alpaka/kernel/Traits.hpp
bernhardmgruber: Agreed.

include/alpaka/kernel/Traits.hpp
psychocoderHPC: good point I changed the template signature

include/alpaka/kernel/Traits.hpp
psychocoderHPC: done

include/alpaka/kernel/Traits.hpp
```diff
@@ -221,13 +244,8 @@ namespace alpaka
         template<typename T>
         inline void assertKernelArgIsTriviallyCopyable()
         {
-            static_assert(
-                // it's implementation defined whether the closure type of a lambda is trivially copyable. But they
-                // should be at least empty then (stateless).
-                std::is_empty_v<T> || std::is_trivially_copyable_v<T>,
-                "The kernel argument T must be trivially copyable!");
+            static_assert(IsKernelArgumentTriviallyCopyable_v<T>, "The kernel argument T must be trivially copyable!");
```
bernhardmgruber:
```suggestion
            static_assert(isKernelArgumentTriviallyCopyable<T>, "The kernel argument T must be trivially copyable!");
```

bernhardmgruber (dismissed):

include/alpaka/kernel/Traits.hpp
bernhardmgruber: Just one fix to get it compiling.

psychocoderHPC: I changed the usage to `isKernelArgumentTriviallyCopyable`, alpaka uses different naming compared to `std`.
e.g. `isQueue`,`isPlatform`, ...

fwyzard (approved):

bernhardmgruber (approved):

PR #2199, opened 2023-12-01
remove `ALPAKA_ASSERT_OFFLOAD`, introduce `ALPAKA_ASSERT_ACC`

### LINKED ISSUE #2001: improve `ALPAKA_ASSERT_OFFLOAD`
It's not clear to me what the meaning of `ALPAKA_ASSERT_OFFLOAD` is :-)

The current definition is
```c++
#define ALPAKA_ASSERT(...) assert(__VA_ARGS__)

#if defined(ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) || defined(SYCL_EXT_ONEAPI_ASSERT)
#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) ALPAKA_ASSERT(EXPRESSION)
#elif defined __AMDGCN__ && (!defined NDEBUG)
#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
        do                                                                                                            \
        {                                                                                                             \
            if(!(EXPRESSION))                                                                                         \
                __builtin_trap();                                                                                     \
        } while(false)
#else
#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
        do                                                                                                            \
        {                                                                                                             \
        } while(false)
#endif
```

So, if I interpret it correctly:
  - if `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST` is defined, `ALPAKA_ASSERT_OFFLOAD` expands to `ALPAKA_ASSERT`, which expands to `assert`, which expands to a runtime check only if `NDEBUG` is not defined;

  - if `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST` is not defined:
      - for the CPU backends, `ALPAKA_ASSERT_OFFLOAD` always expands to a no-op;
      - for the CUDA backend, `ALPAKA_ASSERT_OFFLOAD` always expands to a no-op;
      - for the ROCm backend, `ALPAKA_ASSERT_OFFLOAD` expands to a runtime check if `NDEBUG` is not defined;
      - for the SYCL backend, `ALPAKA_ASSERT_OFFLOAD` always expands to `assert`, so it expands to a runtime check if `NDEBUG` is not defined.

It looks like the behaviour is different for the different backends, which seems counter intuitive.

So... what is the expected behaviour ?
  1. always compile to a runtime check
  2. compile to a runtime check if `NDEBUG` is not defined
  3. compile to a runtime check if some other macro is not defined
  4. always compile to a runtime check on the CPU, and compile to a runtime check on the other devices if `NDEBUG` is not defined
  5. always compile to a runtime check on the CPU, and compile to a runtime check on the other devices if some other macro is not defined
  6. something else
  7. always compile to a no-op

### LINKED ISSUE #2186: Discuss `ALPAKA_ASSERT_OFFLOAD`
I find the current implementation of `ALPAKA_ASSERT_OFFLOAD` very confusing.

It's currently defined as
```c++
#if defined(ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) || defined(SYCL_EXT_ONEAPI_ASSERT)
#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) ALPAKA_ASSERT(EXPRESSION)
#elif defined __AMDGCN__ && (!defined NDEBUG)
#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                         \
        do                                                                            \
        {                                                                             \
            if(!(EXPRESSION))                                                         \
                __builtin_trap();                                                     \
        } while(false)
#else
#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                         \
        do                                                                            \
        {                                                                             \
        } while(false)
#endif
```

My understanding is that
  - while building CPU code - including code for CPU-based accelerators - `ALPAKA_ASSERT_OFFLOAD()` becomes `assert()`, that is controlled by `NDEBUG`;
  - while building CUDA code, `ALPAKA_ASSERT_OFFLOAD()` is always disabled;
  - while building HIP/ROCm code, `ALPAKA_ASSERT_OFFLOAD()` is controlled by `NDEBUG`;
  - while building SYCL/oneAPI code, `ALPAKA_ASSERT_OFFLOAD()` becomes `assert()`, that is controlled by `NDEBUG`.

However, if `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST` is defined, `ALPAKA_ASSERT_OFFLOAD()` always becomes `assert()`, and is controlled by `NDEBUG`.

I find two aspects problematic:
  - `ALPAKA_ASSERT_OFFLOAD()` is disabled by default for CUDA targets, and it can only be enabled by setting `ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST`;
  - `ALPAKA_ASSERT_OFFLOAD()` cannot be disabled for GPU targets without setting `NDEBUG`, which disabled asserts also for the standard CPU compilation.

In the CMS software we are using a somewhat different approach:
  - `assert()` are always enabled for CPU code (because we never define `NDEBUG`);
  - `assert()` are disabled by default in CUDA code, and can be enabled by setting the `GPU_DEBUG` macro.

The rationale is that in host code `assert()` are cheap enough and the benefit from the error checking outweighs the minor performance hit we may have, while in CUDA code the impact can be significant[^1] so we enable `assert()` only while debugging.

[^1]: this is due in part due to the extra checks and `trap()` functionality, in part due to the impact on register usage caused by potentially printing messages.

As we are porting our code from CUDA to Alpaka, we are looking to implement a similar functionality:
  - `assert()` are always enabled for CPU code (because we never define `NDEBUG`);
  - `assert()` are disabled by default in CUDA, HIP/ROCm and SYCL/oneAPI code, and can be enabled setting an appropriate preprocessor macro.

Is this an approach that would be useful in general for Alpaka ?

If so, we can make a PR to rework `ALPAKA_ASSERT_OFFLOAD()`; otherwise we can implement it in CMSSW.

psychocoderHPC:
fix #2186
fix #2001

Provide a device side assert `ALPAKA_ASSERT_ACC` that can be disabled by defining `ALPAKA_DISABLE_ASSERT_ACC` in the C++ code or by the CMake option `alpaka_ASSERT_ACC_ENABLE`

In #2186 the footnote says that assert on the device can have a huge impact on the performance, therefore I added the possibility to force disable the device side asserts.
The naming for the CMake option follows the name schema we use for other options in CMake.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -139,6 +139,12 @@ if(alpaka_DISABLE_VENDOR_RNG)
     target_compile_definitions(alpaka INTERFACE "ALPAKA_DISABLE_VENDOR_RNG")
 endif()
 
+# Device side assert
+option(alpaka_ASSERT_ACC_ENABLE "Enable device side asserts. In case  value is OFF device side asserts will be disabled even if NDEBUG is not defined." ON)
+if(!alpaka_ASSERT_ACC_ENABLE)
+    target_compile_definitions(alpaka INTERFACE "ALPAKA_DISABLE_ASSERT_ACC")
+endif()
+
 #-------------------------------------------------------------------------------
 # Debug output of common variables.
 if(${alpaka_DEBUG} GREATER 1)
@@ -731,9 +737,6 @@ if(alpaka_ACC_SYCL_ENABLE)
 endif()
 
 target_compile_definitions(alpaka INTERFACE "ALPAKA_DEBUG=${alpaka_DEBUG}")
-if(alpaka_DEBUG_OFFLOAD_ASSUME_HOST)
-   target_compile_definitions(alpaka INTERFACE "ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST")
-endif()
 
 target_compile_definitions(alpaka INTERFACE "ALPAKA_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=${alpaka_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB}")
 
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -39,7 +39,7 @@ struct TestBufferKernel
         for(size_t z = idx[0]; z < data.extent(0); z += gridSize[0])
             for(size_t y = idx[1]; y < data.extent(1); y += gridSize[1])
                 for(size_t x = idx[2]; x < data.extent(2); x += gridSize[2])
-                    ALPAKA_ASSERT_OFFLOAD(
+                    ALPAKA_ASSERT_ACC(
                         data(z, y, x)
                         == alpaka::mapIdx<1u>(Vec{z, y, x}, Vec{data.extent(0), data.extent(1), data.extent(2)})[0]);
     }
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynMember.hpp
@@ -42,7 +42,7 @@ namespace alpaka
     public:
         BlockSharedMemDynMember(std::size_t sizeBytes) : m_dynPitch(getPitch(sizeBytes))
         {
-            ALPAKA_ASSERT_OFFLOAD(static_cast<std::uint32_t>(sizeBytes) <= staticAllocBytes());
+            ALPAKA_ASSERT_ACC(static_cast<std::uint32_t>(sizeBytes) <= staticAllocBytes());
         }
 
         auto dynMemBegin() const -> uint8_t*
--- a/include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
+++ b/include/alpaka/block/shared/st/detail/BlockSharedMemStMemberImpl.hpp
@@ -39,7 +39,7 @@ namespace alpaka::detail
             : m_mem(mem)
             , m_capacity(static_cast<std::uint32_t>(capacity))
         {
-            ALPAKA_ASSERT_OFFLOAD((m_mem == nullptr) == (m_capacity == 0u));
+            ALPAKA_ASSERT_ACC((m_mem == nullptr) == (m_capacity == 0u));
         }
 #else
         BlockSharedMemStMemberImpl(std::uint8_t* mem, std::size_t) : m_mem(mem)
@@ -52,12 +52,12 @@ namespace alpaka::detail
         {
             // Add meta data chunk in front of the user data
             m_allocdBytes = varChunkEnd<MetaData>(m_allocdBytes);
-            ALPAKA_ASSERT_OFFLOAD(m_allocdBytes <= m_capacity);
+            ALPAKA_ASSERT_ACC(m_allocdBytes <= m_capacity);
             auto* meta = getLatestVarPtr<MetaData>();
 
             // Allocate variable
             m_allocdBytes = varChunkEnd<T>(m_allocdBytes);
-            ALPAKA_ASSERT_OFFLOAD(m_allocdBytes <= m_capacity);
+            ALPAKA_ASSERT_ACC(m_allocdBytes <= m_capacity);
 
             // Update meta data with id and offset for the allocated variable.
             meta->id = id;
@@ -87,7 +87,7 @@ namespace alpaka::detail
                 // Adjust offset to be aligned
                 std::uint32_t const alignedMetaDataOffset
                     = varChunkEnd<MetaData>(off) - static_cast<std::uint32_t>(sizeof(MetaData));
-                ALPAKA_ASSERT_OFFLOAD(
+                ALPAKA_ASSERT_ACC(
                     (alignedMetaDataOffset + static_cast<std::uint32_t>(sizeof(MetaData))) <= m_allocdBytes);
                 auto* metaDataPtr = reinterpret_cast<MetaData*>(m_mem + alignedMetaDataOffset);
                 off = metaDataPtr->offset;
--- a/include/alpaka/core/Assert.hpp
+++ b/include/alpaka/core/Assert.hpp
@@ -9,22 +9,47 @@
 #include <cassert>
 #include <type_traits>
 
+//! The assert can be explicit disabled by defining NDEBUG
 #define ALPAKA_ASSERT(...) assert(__VA_ARGS__)
 
-#if defined(ALPAKA_DEBUG_OFFLOAD_ASSUME_HOST) || defined(SYCL_EXT_ONEAPI_ASSERT)
-#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION) ALPAKA_ASSERT(EXPRESSION)
-#elif defined __AMDGCN__ && (!defined NDEBUG)
-#    define ALPAKA_ASSERT_OFFLOAD(EXPRESSION)                                                                         \
-        do                                                                                                            \
-        {                                                                                                             \
-            if(!(EXPRESSION))                                                                                         \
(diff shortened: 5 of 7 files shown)


fwyzard:
Hi @psychocoderHPC, thanks for the PR.

Is it the intent that people should always use `ALPAKA_ASSERT_ACC` in device code ?

I ask because I'm wondering if the check on `SYCL_EXT_ONEAPI_ASSERT` should be moved into `ALPAKA_ASSERT` instead ?

Something like
```c++
/// The assert can be explicitly disabled by defining NDEBUG
#if defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
// SYCL device compilation
#    if defined(SYCL_EXT_ONEAPI_ASSERT)
#        define ALPAKA_ASSERT(...) assert(__VA_ARGS__)
#    else
#        define ALPAKA_ASSERT(...) ALPAKA_NOOP(__VA_ARGS__)
#    endif
#else
// non-SYCL device compilation
#    define ALPAKA_ASSERT(...) assert(__VA_ARGS__)
#endif

/// Assert for usage within device code which can be disabled independently from NDEBUG.
#if defined(ALPAKA_DISABLE_ASSERT_ACC)
#    define ALPAKA_ASSERT_ACC(...) ALPAKA_NOOP(__VA_ARGS__)
#else
#    define ALPAKA_ASSERT_ACC(...) ALPAKA_ASSERT(__VA_ARGS__)
#endif
```

fwyzard:
Another thing: when the "device" is actually the cpu (_e.g._ device code compiled for `AccCpuSerial`) I would prefer to keep the `assert` enabled.

So, if none of `__CUDA_ARCH__`, `__HIP_DEVICE_COMPILE__` or `__SYCL_DEVICE_ONLY__` are defined, `ALPAKA_ASSERT_ACC` should be equivalent to plain `assert`.

psychocoderHPC:
> Hi @psychocoderHPC, thanks for the PR.
> 
> Is it the intent that people should always use `ALPAKA_ASSERT_ACC` in device code ?
> 
> I ask because I'm wondering if the check on `SYCL_EXT_ONEAPI_ASSERT` should be moved into `ALPAKA_ASSERT` instead ?
> 
> Something like
> 
> ```c++
> /// The assert can be explicitly disabled by defining NDEBUG
> #if defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
> // SYCL device compilation
> #    if defined(SYCL_EXT_ONEAPI_ASSERT)
> #        define ALPAKA_ASSERT(...) assert(__VA_ARGS__)
> #    else
> #        define ALPAKA_ASSERT(...) ALPAKA_NOOP(__VA_ARGS__)
> #    endif
> #else
> // non-SYCL device compilation
> #    define ALPAKA_ASSERT(...) assert(__VA_ARGS__)
> #endif
> 
> /// Assert for usage within device code which can be disabled independently from NDEBUG.
> #if defined(ALPAKA_DISABLE_ASSERT_ACC)
> #    define ALPAKA_ASSERT_ACC(...) ALPAKA_NOOP(__VA_ARGS__)
> #else
> #    define ALPAKA_ASSERT_ACC(...) ALPAKA_ASSERT(__VA_ARGS__)
> #endif
> ```

Yes from my side it was intended that we have an assert for the host site and the usage within the device code.
To control the assert behaviour typically `NDEBUG` is used. I do not want to change these default behavior for host code.
My intention was that we only to a non-standard way for device-side asserts.

Never the less I can change it if needed.

psychocoderHPC:
> Another thing: when the "device" is actually the cpu (_e.g._ device code compiled for `AccCpuSerial`) I would prefer to keep the `assert` enabled.
> 
> So, if none of `__CUDA_ARCH__`, `__HIP_DEVICE_COMPILE__` or `__SYCL_DEVICE_ONLY__` are defined, `ALPAKA_ASSERT_ACC` should be equivalent to plain `assert`.

We can handle asserts differently depending on which accelerator is enabled. The drawback would be that in this case, a kernel with asserts executed with the CUDA backend will behave slightly differently compared to the kernel that is running with a CPU accelerator. 
If you use SYCL and compile for CPU than it is hard to know for the user if asserts are enabled or disabled.

fwyzard:
> If you use SYCL and compile for CPU than it is hard to know for the user if asserts are enabled or disabled.

Indeed, by "CPU" I really meant "compiled by the standard compiler, like `gcc`".

fwyzard:
> Never the less I can change it if needed.

I would say that if we document something like
> use
>   - `ALPAKA_ASSERT` for a check that can can be used anywhere, and is controlled by `NDEBUG` 
>   - `ALPAKA_ASSERT_ACC` for a check that can can be used anywhere, but is disabled by default on backends where it may have a high cost

then it should be changed.

If instead we document
> use
>   - `ALPAKA_ASSERT` for a check that can can be used on the host, and is controlled by `NDEBUG` 
>   - `ALPAKA_ASSERT_ACC` for a check that can can be used on the device, and is disabled by default on backends where it may have a high cost

there is no need to change it.

fwyzard:
With some delay, here is what I have in mind:
```c++
// ALPAKA_ASSERT_ACC_IMPL is an assert-like macro.
// It can be disabled setting the ALPAKA_DISABLE_ASSERT_ACC preprocessor symbol or the NDEBUG preprocessor symbol.
#if !defined(ALPAKA_DISABLE_ASSERT_ACC)
#    define ALPAKA_ASSERT_ACC_IMPL(...) ALPAKA_ASSERT(__VA_ARGS__)
#else
#    define ALPAKA_ASSERT_ACC_IMPL(...) ALPAKA_NOOP(__VA_ARGS__)
#endif

// ALPAKA_ASSERT_ACC is an assert-like macro.
// In device code for a GPU or SYCL backend it can be disabled setting the ALPAKA_DISABLE_ASSERT_ACC preprocessor
// symbol or the NDEBUG preprocessor symbol. In device code for a native C++ CPU backend and in host code, it is
// equivalent to ALPAKA_ASSERT, and can be disabled setting the NDEBUG preprocessor symbol.
#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && defined(__CUDA_ARCH__)
// CUDA device code
#    define ALPAKA_ASSERT_ACC(...) ALPAKA_ASSERT_ACC_IMPL(__VA_ARGS__)
#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED) && defined(__HIP_DEVICE_COMPILE__)
// HIP/ROCm device code
#    define ALPAKA_ASSERT_ACC(...) ALPAKA_ASSERT_ACC_IMPL(__VA_ARGS__)
#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
// SYCL/oneAPI device code
#    if defined(SYCL_EXT_ONEAPI_ASSERT)
#        define ALPAKA_ASSERT_ACC(...) ALPAKA_ASSERT_ACC_IMPL(__VA_ARGS__)
#    else
#        define ALPAKA_ASSERT_ACC(...) ALPAKA_NOOP(__VA_ARGS__)
#    endif
// add here any other #elif conditions for non-CPU backends
// ...
#else
// CPU backend, or host code
#    define ALPAKA_ASSERT_ACC(...) ALPAKA_ASSERT(__VA_ARGS__)
#endif
```

I've split off the check on `ALPAKA_DISABLE_ASSERT_ACC` into a separate macro `ALPAKA_ASSERT_ACC_IMPL` to make the code more readable - but that's supposed to be just an implementation detail, not something the user code would call.

psychocoderHPC:
I updated the PR with the suggestion from @fwyzard and added him as a co-author.

fwyzard (approved):

psychocoderHPC:
PR can be merged, the failing CI test is not required and is already fixed with https://github.com/alpaka-group/alpaka/pull/2208

PR #2201, opened 2023-12-01
named access to `Vec`

### LINKED ISSUE #2185: Proposal: add `.x()`, `.y()`, ... accessors to `alpaka::Vec`
As we gain more experience rewriting CUDA code and porting it to Alpaka, one of the aspects that people are struggling a lot with is the use of OpenCL-like indices for accessing `alpaka::Vec` variables, when they are used for the work division, execution index, _etc._

For 1 dimensional kernel things are easy
```c++
dim3 id = ...;
f(id.x);
```
maps to
```c++
Vec<DimInt<1u>, float> id = ...'
f(id[0]);
```

For a 2 dimensional kernel things are already complicated:
```c++
dim3 id = ...;
f(id.x, id.y);
```
maps to
```c++
Vec<DimInt<2u>, float> id = ...'
f(id[1], id[0]);
```

The problem is not so much that `.x`, `.y` map to `[1]`, `[0]`...

The biggest problem is that `.x` maps to `[N-1]`, `.y` maps to `[N-2]`, _etc_.
This means that `.x` maps to `[0]` in a 1-dimensional kernel, but `.x` maps to `[1]` in a 2-Dimensional kernel, _etc._ - which is very confusing for users, when they mix 1D and 2D kernels.

Even though we briefly entertained the possibility, it's clear that we cannot change the ordering of the indices used in Alpaka, as it would break all existing code.

But we _could_ _add_ a more familiar syntax to `Vec` variables: `.x()`, `.y()`, `.z()` accessors.

```c++
Vec<DimInt<1u>, float> id = ...'
f(id.x());
```

```c++
Vec<DimInt<2u>, float> id = ...'
f(id.x(), id.y());
```

We should be able to use SFINAE or partial specialisation to make `.x()` available only when N > 0, `.y()` available only when N > 1, _etc_.

What do you think ?

psychocoderHPC:
fix #2185

Provide access to the vector components via named methods. `x()`, `y()`, `z()` and `w()`.
The available names depend on the dimension of the vector.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -170,6 +170,60 @@ namespace alpaka
             return m_data[Dim::value - 1];
         }
 
+        //! access elements by name
+        //!
+        //! names: x,y,z,w
+        //! @{
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 1, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) x() const
+        {
+            return m_data[Dim::value - 1];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 1, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) x()
+        {
+            return m_data[Dim::value - 1];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 2, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) y() const
+        {
+            return m_data[Dim::value - 2];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 2, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) y()
+        {
+            return m_data[Dim::value - 2];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 3, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) z() const
+        {
+            return m_data[Dim::value - 3];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 3, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) z()
+        {
+            return m_data[Dim::value - 3];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 4, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) w() const
+        {
+            return m_data[Dim::value - 4];
+        }
+
+        template<typename TDefer = Dim, std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 4, int> = 0>
+        ALPAKA_FN_HOST_ACC constexpr decltype(auto) w()
+        {
+            return m_data[Dim::value - 4];
+        }
+
+        //! @}
+
         //! Value reference accessor at the given non-unsigned integer index.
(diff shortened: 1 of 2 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -170,6 +170,68 @@ namespace alpaka
             return m_data[Dim::value - 1];
         }
 
+        //! access elements by name
+        //!
+        //! names: x,y,z,w
+        //! @{
+        template<typename TDefer = Dim>
+        ALPAKA_FN_HOST_ACC constexpr auto x() const
+            -> std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 1, TVal const&>
```
bernhardmgruber: I am personally a fan of putting `enable_if` into the template parameter list for better readability. This may also remove the need to delay instantiation as you did with `TDefer`. I have not tested it though.
```suggestion
        template<std::enable_if_t<std::is_same_v<TDefer, Dim> && Dim::value >= 1, int> = 0>
        ALPAKA_FN_HOST_ACC constexpr auto x() const -> TVal const&
```

test/unit/vec/src/VecTest.cpp
```diff
@@ -388,3 +388,67 @@ TEST_CASE("deductionGuide", "[vec]")
...
     STATIC_REQUIRE(std::is_same_v<decltype(v3l), alpaka::Vec<alpaka::DimInt<3>, long>>);
 }
+
+TEST_CASE("accessByName", "[vec]")
+{
+    // dim == 1
+    auto v1 = alpaka::Vec{1};
+    auto const v1const = v1;
+    CHECK(v1.x() == 1); // non-const overload
+    CHECK(v1const.x() == 1); // const overload
```
bernhardmgruber: Only if you want:
```suggestion
    auto v1 = alpaka::Vec{1};
    CHECK(v1.x() == 1); // non-const overload
    CHECK(std::as_const(v1).x() == 1); // const overload
```

bernhardmgruber (dismissed):
LGTM, here are a few suggestions if you like:

include/alpaka/vec/Vec.hpp
psychocoderHPC: I took your suggestion because this allow the usage of `decltype(auto)`

test/unit/vec/src/VecTest.cpp
psychocoderHPC: I applied your changes and extended the test slidly to check if we can correctly manipulate the values.

bernhardmgruber (approved):

include/alpaka/vec/Vec.hpp
bernhardmgruber: Why would you want to use `decltype(auto)` over just `TVal&` and `TVal const&`?

psychocoderHPC: Avoids copy past errors^^. IMO defining `RefType` and `ConstRefType` or similar would be better than writing everywhere `TVal&`.

bernhardmgruber: > Avoids copy past errors^^. IMO defining `RefType` and `ConstRefType` or similar would be better than writing everywhere `TVal&`.

That's even worse in my opinion. I value readability and `auto x() -> TVal&` is more readable than `decltype(auto) x()` or `auto x() -> RefType`. The former requires me to look at the function body, the latter requires me to lookup what `RefType` is.

But in any case, I was just curious. I leave this up to you!

fwyzard: `static_assert()` works without the need for `TDefer`, and lets us provide a more readable error message:
```c++
        ALPAKA_FN_HOST_ACC constexpr decltype(auto) x() const
        {
            static_assert(Dim::value >= 1, ".x() is defined only for Vec types with a dimension of 1 or more.");
            return m_data[Dim::value - 1];
        }
```

psychocoderHPC: @fwyzard Your suggestion will prevent you can check from the user code side if `Vec` is allowing the call to the method `z()`. I know that the defer in the template signature is ugly but the function is cleanly disabled if not supported.

fwyzard: What do you mean, to check from user code ?

bernhardmgruber: The version from @psychocoderHPC is SFINAE-friendly, whereas a `static_assert` is not:
```c++
if constexpr (requires { v.y() }) ...
```
Works with the version here for a 1D vector `v`, but fails with a hard error using a `static_assert`.

Issue #1019, opened 2020-05-29
CMake target for each accelerator


psychocoderHPC:
Based on [this](https://github.com/alpaka-group/alpaka/pull/942#issuecomment-593970673) I suggest that we add CMake target for each accelerator backend.
Currently alpaka CMake behaves to optimal fits the CI workflow and is activating all available backends. 
Backends should be select able during the install of alpaka and store the dependencies including there locations. After alpaka is installed e.g. for CUDA than there is no requirement to search for a CUDA library again because dependencies must be checked during the install or when `add_subdirectory` is called.

j-stephan:
> After alpaka is installed e.g. for CUDA than there is no requirement to search for a CUDA library again because dependencies must be checked during the install or when add_subdirectory is called.

What if I don't have CUDA installed at first but install it after alpaka? Do we want to force a reinstall?

psychocoderHPC:
Than you can not use CUDA. If you install MPI before you install CUDA you can not have a CUDA-aware MPI.
If you activate CUDA after you installed alpaka you can not guarantee that you alpaka works fine with CUDA.
The general workflow to install software is:

```
- configure
- make
- make test
- make install
``` 

To switch the CUDA version after installing alpaka sound good but will avoid that you are on a reproducible environment.

psychocoderHPC:
You sill have the possibility to use alpaka via `add_subdirectory` to be able to switch accelerator backends.

j-stephan:
But CUDA is a build dependency for CUDA-aware MPI, right? This isn't true for alpaka since we are header-only.

I'm also not convinced that there is a problem with our current approach. Right now we check for dependencies at build / installation / `add_subdirectory` as well as `find_package` (i.e. after installing alpaka). The latter comes with virtually no overhead. Why should we remove it?

psychocoderHPC:
With the current approche the admin who provides alpaka as module has hard times to guarantee that alpaka works because what 'make test' checks is not what the iser later uses as environment.
We should discuss it with more alpaka devs in a face to face meeting. 
It is hard for me to explain here all because I need maybe intruduce how HPC system handle there software and how they IMO should do it without moving software/environment tests to there users.

psychocoderHPC:
A possible way would be to keep the live search in cmake after the installation but only allow backends those was activated and therefore could be tested during the install and disable all other.
This would still allow you to switch e.g. between different cuda versions if cuda was enabled during the install.

ax3l:
[...]

Issue #919, opened 2020-01-27
CMake Modernization


j-stephan:
I'd like to improve the current CMake infrastructure and adjust it to modern CMake idioms (as seen [here](https://www.youtube.com/watch?v=y7ndUhdQuU8) and [here](https://www.youtube.com/watch?v=y9kSr5enrSk)).

Goals to achieve:

* make Alpaka installable to the file system (`/usr/local` comes to mind)
* reduce current complexity
* ease the integration of new backends (SYCL)

My current idea is to make everything (Alpaka headers, test cases, examples, third-party dependencies) a target, add some high-level targets (like a common target for all unit tests, another for all integration tests, and so on) and let CMake handle the build and usage requirements. Example: CUDA would be a third-party target. The user would link his project to `alpaka::cuda` which would then pull in a CUDA requirement. If he prefers OpenMP4, he'd link to `alpaka::omp4` without the need for CUDA.

There are some questions remaining that I'd like to discuss with the Alpaka developers:

* Which CMake version should we aim for? I originally aimed for CMake 3.15 (which I had to compile myself for Ubuntu 19.10), but at least one proposal (#858) would require CMake 3.16.
* Which Alpaka headers are we considering public (should be used by the user) and which private (for internal use only)?
* `FindCUDA.cmake` has been deprecated for more than two years now. Unfortunately, there is still no native support for clang-cuda. Should we still use the deprecated script or instead ditch clang-cuda for the time being? Native support has been worked on for [3 years](https://gitlab.kitware.com/cmake/cmake/issues/16586) but there is still no solution in sight.

j-stephan:
@ax3l Are you still working on #478?

BenjaminW3:
> Which Alpaka headers are we considering public (should be used by the user) and which private (for internal use only)?

nearly everything is public

> Should we still use the deprecated script or instead ditch clang-cuda for the time being?

I really like clang-cuda and would not want to drop support for it.

> Which CMake version should we aim for?

I have no problem requiring CMake 3.15. I am not sure if we should already require 3.16. If it provides a huge value then it should be possible to even require 3.16.

BenjaminW3:
> My current idea is to make everything (Alpaka headers, test cases, examples, third-party dependencies) a target

This should already be the case.

> add some high-level targets (like a common target for all unit tests, another for all integration tests, and so on)

[...]

PR #2203, opened 2023-12-06
add CMake presets

SimeonEhrig:
add CMake presets to make it easier to compile alpaka for different backends

# TODO

- [x] add missing backends
- [x] write documentation

# Usage

```bash
cd <alpaka_root>
cmake --preset gpu-cuda-nvcc-gcc
cmake --build --preset gpu-cuda-nvcc-gcc
ctest --preset gpu-cuda-nvcc-gcc 
```

# Testing

Because of more urgent other tasks I will add tests later during working on PR #919 and #1019. If the CMake presets get broken, nothing else will be affected.

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,198 @@
+{
+    "version": 3,
+    "cmakeMinimumRequired": {
+        "major": 3,
+        "minor": 22,
+        "patch": 0
+    },
+    "configurePresets": [
+        {
+            "name": "develop",
+            "description": "enable tests and examples",
+            "hidden": true,
+            "binaryDir": "${sourceDir}/build/${presetName}",
+            "cacheVariables": {
+                "alpaka_BUILD_EXAMPLES": "ON",
+                "BUILD_TESTING": "ON"
+            }
+        },
+        {
+            "name": "cpu-serial",
+            "description": "enable serial back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "cpu-omp2b",
+            "description": "enable OpenMP 2 Grid Block back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "cpu-omp2t",
+            "description": "enable OpenMP 2 Block Thread back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "cpu-stdthreads",
+            "description": "enable std::thread back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "cpu-tbb",
+            "description": "enable Intel TBB back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "cpu-all",
+            "description": "enable all cpu back-ends (serial, OpenMP 2 Grid Block and Block Thread, std::thread and Intel TBB back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
+                "alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE": "ON",
+                "alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE": "ON",
+                "alpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE": "ON",
+                "alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "gpu-hip",
+            "description": "use HIP back-end on AMD GPUs",
+            "inherits": "develop",
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER": "hipcc",
+                "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
+                "alpaka_ACC_GPU_HIP_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "gpu-sycl-intel",
+            "description": "use Sycl back-end on Intel GPUs",
+            "inherits": "develop",
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER": "icpx",
+                "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
+                "alpaka_ACC_SYCL_ENABLE": "ON",
+                "alpaka_SYCL_ONEAPI_GPU": "ON",
+                "alpaka_SYCL_ONEAPI_GPU_DEVICES": "spir64"
+            }
+        },
+        {
+            "name": "gpu-cuda-nvcc",
+            "description": "use cuda back-end on Nvidia GPUs with nvcc device compiler",
+            "inherits": "develop",
+            "cacheVariables": {
+                "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
+                "alpaka_ACC_GPU_CUDA_ENABLE": "ON"
+            }
+        },
+        {
+            "name": "gpu-cuda-clang",
+            "description": "use cuda back-end on Nvidia GPUs with clang++ device compiler",
+            "inherits": "develop",
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER": "clang++",
+                "CMAKE_CUDA_COMPILER": "clang++",
+                "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
+                "alpaka_ACC_GPU_CUDA_ENABLE": "ON"
+            }
+        }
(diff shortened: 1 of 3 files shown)


CMakePresets.json
```diff
@@ -0,0 +1,102 @@
...
+            "cacheVariables": {
+                "alpaka_BUILD_EXAMPLES": {
+                    "type": "BOOL",
+                    "value": "ON"
+                },
+                "BUILD_TESTING": {
+                    "type": "BOOL",
+                    "value": "ON"
+                }
+            }
```
bernhardmgruber: Does it need to be this verbose? In LLAMA, I am fine with something like:
```suggestion
            "cacheVariables": {
                "alpaka_BUILD_EXAMPLES": "ON",
                "BUILD_TESTING": "ON"
            }
```

SimeonEhrig: By default `BUILD_TESTING` is off. I thought it make sense to enable tests and examples if you want to build alpaka.

SimeonEhrig:
@bernhardmgruber @psychocoderHPC Can you please have a look on the name scheme of the profiles and also the enabled arguments.

CMakePresets.json
bernhardmgruber: Maybe you misunderstood me. I meant why do you need to specify the `"type"` in an extra nested JSON object. Every cache variable could just be one line, which is far easier to read IMO.

CMakePresets.json
```diff
@@ -0,0 +1,407 @@
...
+        },
+        {
+            "name": "cpu-serial-gcc",
+            "description": "use g++ compiler with serial back-end",
+            "inherits": "develop",
+            "cacheVariables": {
+                "CMAKE_CXX_COMPILER": {
+                    "type": "STRING",
+                    "value": "g++"
+                },
```
bernhardmgruber: I assume this would break my workflow. I configure with:
`CXX=g++-13 cmake ..`. For clang, I do not even have a `clang++` installed, only `clang++-version`.

bernhardmgruber (commented):
I think that except for the `gpu-sycl-intel` preset, where the cmake options are harder to figure out, this CMake presets file would not help me much :S

In LLAMA, I use this for example:
```
    {
      "name": "default",
      "toolchainFile": "~/dev/vcpkg/scripts/buildsystems/vcpkg.cmake",
      "cacheVariables": {
        "BUILD_TESTING": "ON",
        "LLAMA_BUILD_EXAMPLES": "ON",
        "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
        "Boost_NO_WARN_NEW_VERSIONS": "ON"
      }
    }
```
Because specifying the toolchain file annoying, I don't have tests and examples enabled by default, but this is what I do locally almost 100%. I suppress Boost warnings in CMake and I want to usually have the alpaka serial accelerator, so all examples compile.

For alpaka, I would rather use a similar config, e.g.:
```
    {
      "name": "minimal",
      "cacheVariables": {
        "BUILD_TESTING": "ON",
        "alpaka_BUILD_EXAMPLES": "ON",
        "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
        "Boost_NO_WARN_NEW_VERSIONS": "ON"
      }
    }
```
Or one to have all CPU backends:
```
    {
      "name": "all_cpu",
      "cacheVariables": {
        "BUILD_TESTING": "ON",
        "alpaka_BUILD_EXAMPLES": "ON",
        TODO: enable all CPU backends
        "alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE": "ON",
        "Boost_NO_WARN_NEW_VERSIONS": "ON"
      }
    }
```

I think the CMake presets should not provide yet another configuration mechanism to enable/disable backends, `ccmake` works good for this, but rather support common work flows.

CMakePresets.json
SimeonEhrig: You version would hard work together with IDEs such like Visual Studio Code. In special for vsc on remote systems, it is pretty hard to configure environment variables.

SimeonEhrig:
The basic idea of the presets are the following:

- use it directly for the alpaka project -> for own projects using alpaka, you should write your own presets.
- show how to enable the different backends. In special the sycl backend is pretty hard to enable

I would also not enable all CPU backends at the same time. This requires that you install OpenMP and Intel TBB on your system on the same time.

And I avoid to use `ccmake` where I can because it introduce a new source of errors. It causes undocumented changes in the cmake configuration. Better would be to create a new user specific preset. With this preset you know what you did every time. Also you can simply delete the build and configure it again.

CMakePresets.json
bernhardmgruber: For CLion, this was easy to configure in the Toolchain menu. But anyhow, I just made a suggestion. I don't need the preset file here, so if you want to hardcode `g++` than I don't mind.

SimeonEhrig: The basic idea was, that you can easily overwrite the parameter with `cmake --preset cmake --preset cpu-serial-gcc -DCMAKE_CXX_COMPILER=g++-13`. But we can also remove the gcc and clang version and provides only one configuration per cpu back-end.

bernhardmgruber:
> The basic idea of the presets are the following:
> 
> * use it directly for the alpaka project -> for own projects using alpaka, you should write your own presets.

Right, so it's intended for developers and not downstream users.

> * show how to enable the different backends. In special the sycl backend is pretty hard to enable

We don't need that, because this is what documentation is for.

> I would also not enable all CPU backends at the same time. This requires that you install OpenMP and Intel TBB on your system on the same time.

Which I have and whenever I want to test all backends, I would want to have a preset for this :)

> And I avoid to use `ccmake` where I can because it introduce a new source of errors. It causes undocumented changes in the cmake configuration. Better would be to create a new user specific preset. 

I completely disagree, it's the bread and butter of configurating. It's very approachable and nicely shows what kind of options a project has.

> With this preset you know what you did every time. Also you can simply delete the build and configure it again.

That's a fair point and that's why I configure with a fixed command line when doing reproducible benchmarks. But for normal coding, `ccmake` and change whatever I want.

bernhardmgruber:
Anyhow, I let @psychocoderHPC take over this review.

SimeonEhrig:
> We don't need that, because this is what documentation is for.

I prefer an interactive documentation compare to text. It's true, I can compile all information from the README. But why I should do it, if the preset is already doing it for me?

> Which I have and whenever I want to test all backends, I would want to have a preset for this :)

Fine, I will add one. I will not hurt.

> I completely disagree, it's the bread and butter of configurating. It's very approachable and nicely shows what kind of options a project has.

> That's a fair point and that's why I configure with a fixed command line when doing reproducible benchmarks. But for normal coding, ccmake and change whatever I want.

We work completely different. I only use directly `cmake` if I want to build a project. If I want to develop a project, vsc is doing all the stuff. I don't want to think about cmake configuration. The computer should do this. I would get mad if I need to care about. I use a least 4 different systems and switches quickly ;-)

SimeonEhrig:
@bernhardmgruber After our discussion, I would do the following things:

- remove the types to make the json much easier
- unify each cpu backend version of gcc and clang to a single cpu backend version without explicit CXX compiler
- add a cpu backend configuration which enables: serial, both OpenMP 2, threads? [1], Intel TBB

[1] I'm not sure if we should enabled `std::threads`, because the execution time of the examples and tests are horrible

I would keep setting the CXX compiler for Intel Sycl and AMD HIP because this is a big source of errors to forget it.

My idea of the presets is, that you have a preset for each backend that you have an easy start point to build the examples and tests for the backend. And you can use the presets to inherent from it with your `CMakeUserPresets.json` or configure an build and modify it afterwards with `ccmake`.

Personally, I will not use the `CMakePresets.json` directly. I want to inherent from it via `CMakeUserPresets.json`, set `ninja` as build generator and create a debug and release version of each backend, which I can switch easily in vsc.

bernhardmgruber:
@SimeonEhrig do what works best for you! And I like your suggestions. I also did not know about `CMakeUserPresets.json`.

CMakePresets.json
SimeonEhrig: Now, I only set the generic device compiler (`clang++`, `hipcc`, `ipcx`) where it is required.

docs/source/advanced/cmake.rst
```diff
@@ -1,12 +1,99 @@
...
+                    "value": "RELEASE"
+                }
+            }
+        }
+      ]
+   }
+
+.. hint::
+
+   Many IDEs like `Visual Studio Code <https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-presets.md>`_ and `CLion <https://www.jetbrains.com/help/clion/cmake-presets.html>`_ supports CMake presets.
```
bernhardmgruber:
```suggestion
Many IDEs like `Visual Studio Code <https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-presets.md>`_ and `CLion <https://www.jetbrains.com/help/clion/cmake-presets.html>`_ supports CMake presets.
```
Whitespace on purpose?

Also:
```suggestion
   Many IDEs like `Visual Studio Code <https://github.com/microsoft/vscode-cmake-tools/blob/main/docs/cmake-presets.md>`_ and `CLion <https://www.jetbrains.com/help/clion/cmake-presets.html>`_ support CMake presets.
```

bernhardmgruber (dismissed):

bernhardmgruber:
@psychocoderHPC please also share your opinion on how you would want to use presets, or not :)

@SimeonEhrig: btw, great work on the documentation!

docs/source/advanced/cmake.rst
SimeonEhrig: The whitespaces are required by reStructuredText that it correctly renders the hint box.

SimeonEhrig:
@bernhardmgruber René wrote in the internal chat, that he cannot give feedback and is fine with your review.

bernhardmgruber:
> @bernhardmgruber René wrote in the internal chat, that he cannot give feedback and is fine with your review.

Great! However, I do not have the permissions to merge a PR with failing tests or with @psychocoderHPC's review pending. You have to ask @psychocoderHPC to merge it.

psychocoderHPC (approved):

PR #2215, opened 2024-01-03
Use CMake's native HIP support Nr. 2

SimeonEhrig:
Open PR #2150 again, because I want to finish the PR and I cannot squash commits, if I'm not the owner of the branch.


cmake/alpakaCommon.cmake
```diff
@@ -506,31 +507,38 @@ endif()
...
 
-    if(NOT TARGET hip)
-        message(FATAL_ERROR "Optional alpaka dependency HIP could not be found!")
-    else()
-        target_link_libraries(alpaka INTERFACE hip::host hip::device)
+
+        target_link_libraries(alpaka INTERFACE "$<$<LINK_LANGUAGE:CXX>:hip::host>")
+        alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<COMPILE_LANGUAGE:CXX>:-D__HIP_PLATFORM_AMD__>")
+        # only required for HIP 5.1 and maybe older
+        alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<COMPILE_LANGUAGE:CXX>:-D__HIP_PLATFORM_HCC__>")
```
psychocoderHPC: If this is only required for HIP 5.1 we should apply the fix only for this version.

```
if(${_hip_MAJOR_MINOR_VERSION} VERSION_EQUAL "5.1")
        alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<COMPILE_LANGUAGE:CXX>:-D__HIP_PLATFORM_HCC__>")
endif()
```

psychocoderHPC (changes requested):

psychocoderHPC (approved):

PR #2216, opened 2024-01-09
Add `alpaka::getPreferredWarpSize(dev)`

fwyzard:
`alpaka::getPreferredWarpSize(dev)` returns one of the possible warp sizes supported by the device.

On devices that support a single work size (cpu, CUDA gpu, ROCm gpu), `getPreferredWarpSize(dev)` avoids the overhead of wrapping that value in an `std::vector`.

On devices that support multiple warp sizes, the value returned by `getPreferredWarpSize(dev)` is unspecified. Currently it returns the largest supported value -- but this could change in a future version of alpaka.

Add a test for `alpaka::getPreferredWarpSize(dev)`.

### DIFF
--- a/include/alpaka/dev/DevGenericSycl.hpp
+++ b/include/alpaka/dev/DevGenericSycl.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2023 Jan Stephan, Antonio Di Pilato, Luca Ferragina, Aurora Perego
+/* Copyright 2024 Jan Stephan, Antonio Di Pilato, Luca Ferragina, Aurora Perego, Andrea Bocci
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -18,6 +18,7 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <functional>
 #include <memory>
 #include <mutex>
 #include <shared_mutex>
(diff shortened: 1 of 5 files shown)


bernhardmgruber (approved):

bernhardmgruber:
I am just curious on the purpose of this API. Is the main goal to avoid the heap allocation of `auto getWarpSizes(TDev const& dev) -> std::vector<std::size_t>`?

Because we could just change the API to either return e.g. a `boost::small_vector` or just cache the warp sizes in the device and return a `std::span`. The latter assumes that a device does not change its warp size during program execution.

include/alpaka/dev/DevGenericSycl.hpp
```diff
@@ -181,10 +182,22 @@ namespace alpaka::trait
             auto find64 = std::find(warp_sizes.begin(), warp_sizes.end(), 64);
             if(find64 != warp_sizes.end())
                 warp_sizes.erase(find64);
+            // Sort the warp sizes in decreasing order
+            std::sort(warp_sizes.begin(), warp_sizes.end(), std::greater<>{});
```
mehmetyusufoglu, line 186: [Nit] If the vector size can be large somehow,  "initially sorting then finding the element in vector" (logarithmic time) is faster compared to "finding deleting then sorting" (linear time). After sorting, using std::lower_bound function finds in log(n) time.

fwyzard: The largest vector I encountered had 5 elements: `{ 4, 8, 16, 32, 64 }`.

include/alpaka/dev/DevGenericSycl.hpp
```diff
@@ -181,10 +182,22 @@ namespace alpaka::trait
             auto find64 = std::find(warp_sizes.begin(), warp_sizes.end(), 64);
             if(find64 != warp_sizes.end())
                 warp_sizes.erase(find64);
```
bernhardmgruber, line 184: Btw, in C++20, this should be just `std::erase(warp_sizes, 64)`. Looking forward to the upgrade :)

bernhardmgruber:
@fwyzard if you want the PR merged, please mark the PR as Ready for review, thx!

fwyzard:
Thanks for the review.

I've marked it as a draft because I want to figure out first how it interacts with caching the device information.

include/alpaka/dev/DevGenericSycl.hpp
bernhardmgruber: @mehmetyusufoglu your analysis is correct, but I am with @fwyzard: the supported warp sizes are probably a small set here :) Also, binary search is slower than linear for small sizes due to data dependent access pattern. So @fwyzard's version is probably faster for our use case :D

PR #2219, opened 2024-01-15
[WIP] Use  executionTask+enqueue instead of exec in examples

mehmetyusufoglu:
There are 2 ways of running a kernel. Some of the examples use `alpaka::exec<Acc>(queue, workdiv...) + wait(queue)` some others use `createTaskKernel<Acc> and enqueue` functions. 

All of the exec and wait function call pairs converted to `createTaskKernel + enqueue` calls.

**Tests**
All example results are the same as before these changes.

**Clarification Needed:** 

-  If there is a function call between `exec` call and `wait` call, how the change above could be done?  See the change in **MonteCarlo** example.

- **Note from @psychocoderHPC :**
_I checked the code. And enqueue requires that we bind together already the kernel with the work division. 
Where exec allows to pass the work division and kernel separate more similar to how CUDA works.
I like the exec way more because very often the kernel will be started with different work division size. IN that case the user must always call_  

```
alpaka::createTaskKernel<Acc>(
        workDiv2,
        kernel2, ...
)
``` 
 _for any new block and grid configuration. So the user code becomes longer.
IMO we should provide additionally an enqueue which has the same signature as exec has and provide two interfaces of enqueue. Let us have a short discussion in the next developer meeting._

### DIFF
--- a/example/babelstream/src/AlpakaStream.cpp
+++ b/example/babelstream/src/AlpakaStream.cpp
@@ -51,8 +51,7 @@ void AlpakaStream<T>::init_arrays(T initA, T initB, T initC)
 {
     auto const workdiv = WorkDiv{arraySize / blockSize, blockSize, 1};
     // auto const workdiv = alpaka::getValidWorkDiv(devAcc, arraySize);
-    alpaka::exec<Acc>(
-        queue,
+    auto const taskKernel = alpaka::createTaskKernel<Acc>(
         workdiv,
         InitKernel{},
         alpaka::getPtrNative(d_a),
@@ -61,7 +60,7 @@ void AlpakaStream<T>::init_arrays(T initA, T initB, T initC)
         initA,
         initB,
         initC);
-    alpaka::wait(queue);
+    alpaka::enqueue(queue, taskKernel);
 }
 
 template<typename T>
@@ -87,8 +86,10 @@ void AlpakaStream<T>::copy()
 {
     auto const workdiv = WorkDiv{arraySize / blockSize, blockSize, 1};
     // auto const workdiv = alpaka::getValidWorkDiv(devAcc, arraySize);
-    alpaka::exec<Acc>(queue, workdiv, CopyKernel{}, alpaka::getPtrNative(d_a), alpaka::getPtrNative(d_c));
-    alpaka::wait(queue);
+    auto const taskKernel
+        = alpaka::createTaskKernel<Acc>(workdiv, CopyKernel{}, alpaka::getPtrNative(d_a), alpaka::getPtrNative(d_c));
+    // Enqueue the kernel
+    alpaka::enqueue(queue, taskKernel);
 }
 
 struct MulKernel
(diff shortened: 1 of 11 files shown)


fwyzard:
@mehmetyusufoglu these changes are not equivalent:
  - `alpaka::exec<Acc>(queue, ...);` is equivalent to `task = alpaka::createTaskKernel<Acc>(...);  alpaka::enqueue(queue, task);`
  - the call to `alpaka::wait(queue);` is needed in either case, if one wants to wait for the kernel execution to complete.

That said, why do you want to migrate back from `exec` to `createTaskKernel + enqueue` ?
I'd rather always use `exec` and avoid the explicit creation of the temporary task.

fwyzard (changes requested):
Please do not replace `exec` with `createTaskKernel + enqueue`.

mehmetyusufoglu:
> @mehmetyusufoglu these changes are not equivalent:
> 
>     * `alpaka::exec<Acc>(queue, ...);` is equivalent to `task = alpaka::createTaskKernel<Acc>(...);  alpaka::enqueue(queue, task);`
> 
>     * the call to `alpaka::wait(queue);` is needed in either case, if one wants to wait for the kernel execution to complete.
> 
> 
> That said, why do you want to migrate back from `exec` to `createTaskKernel + enqueue` ? I'd rather always use `exec` and avoid the explicit creation of the temporary task.

. 

> @mehmetyusufoglu these changes are not equivalent:
> 
>     * `alpaka::exec<Acc>(queue, ...);` is equivalent to `task = alpaka::createTaskKernel<Acc>(...);  alpaka::enqueue(queue, task);`
> 
>     * the call to `alpaka::wait(queue);` is needed in either case, if one wants to wait for the kernel execution to complete.
> 
> 
> That said, why do you want to migrate back from `exec` to `createTaskKernel + enqueue` ? I'd rather always use `exec` and avoid the explicit creation of the temporary task.

These changes are requested by Rene while we were talking about "why there are 2 different ways of calling kernels in examples". This is WIP, to be discussed. I can close the pull request if it is not needed. 

About `wait()`, yes, I was doubted about it certainly. Thanks.

mehmetyusufoglu:
Closed, not needed currently.

psychocoderHPC:
> @mehmetyusufoglu these changes are not equivalent:
> 
> * `alpaka::exec<Acc>(queue, ...);` is equivalent to `task = alpaka::createTaskKernel<Acc>(...);  alpaka::enqueue(queue, task);`
> * the call to `alpaka::wait(queue);` is needed in either case, if one wants to wait for the kernel execution to complete.
> 
> That said, why do you want to migrate back from `exec` to `createTaskKernel + enqueue` ? I'd rather always use `exec` and avoid the explicit creation of the temporary task.

@fwyzard Independent of the wrong removed waits. I am wondering if it makes more sense to remove `alpaka::exec<Acc>()` and provide `alpaka::enqueue<Acc>(queue, workdiv, kernel)` or `alpaka::enqueue(alpaka::TagGpuCudaRt, queue, workdiv, kernel)`?
I do not see a real reason why we have two different naming to enqueue an kernel. That we have different function signatures makes IMO sense.

PR #2220, opened 2024-01-15
1D Convolution example

mehmetyusufoglu:
This pull request consisting 1D convolution implementation is an exercise for a 2D or 3D convolution example.

### DIFF
--- a/example/convolution1D/src/convolution1D.cpp
+++ b/example/convolution1D/src/convolution1D.cpp
@@ -1,0 +1,178 @@
+/* Copyright 2023  Bernhard Manfred Gruber, Simeon Ehrig, Rene Widera, Mehmet Yusufoglu.
+ * SPDX-License-Identifier: ISC
+ */
+
+#include <alpaka/alpaka.hpp>
+#include <alpaka/example/ExampleDefaultAcc.hpp>
+
+#include <cmath>
+#include <iomanip>
+#include <iostream>
+#include <limits>
+#include <type_traits>
+
+//! Convolution Example
+//!
+//! 1D convolution example: Creates two 1D arrays, applies convolution filter.
+//! Array sizes are hardcoded.
+//!
+
+/**
+ * @brief The ConvolutionKernel function-object
+ * Calculates 1D convolution using input and filter arrays.
+ */
+struct ConvolutionKernel
+{
+    /** @brief Main convolution code
+     *  @param Accelerator
+     *  @param Input array, first input of convolution integral
+     *  @param Filter array, second input of convolution integral
+     *  @param Empty output array to be filled
+     *  @param Input array size
+     *  @param Filter size
+     */
+    template<typename TAcc, typename TElem>
+    ALPAKA_FN_ACC auto operator()(
+        TAcc const& acc,
+        TElem const* const input,
+        TElem const* const filter,
+        TElem* const output,
+        const std::size_t inputSize,
+        const std::size_t filterSize) const -> void
+    {
+        auto const globalThreadIdxX = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+
+        // Since the kernel is launched 1-D calculating linearizedGlobalThreadIdx line is unnecessary.
+        // globalThreadIdx[0] can be used to map all the threads.
+        if(globalThreadIdxX < inputSize)
+        {
+            int32_t const halfFilterSize = filterSize / 2;
+            TElem result = 0.0f;
+            // Calculate sum of multiplications of corresponding elements
+            auto const start
+                = static_cast<int32_t>(std::max(static_cast<int32_t>(globalThreadIdxX) - halfFilterSize, 0));
+            auto const stop = std::min(globalThreadIdxX + halfFilterSize, inputSize - 1);
+            for(int32_t i = start; i <= stop; ++i)
+                result += input[i] * filter[i + halfFilterSize - static_cast<int32_t>(globalThreadIdxX)];
+            output[globalThreadIdxX] = result;
+        }
+    }
+};
+
+auto FuzzyEqual(float a, float b) -> bool
+{
+    return std::fabs(a - b) < std::numeric_limits<float>::epsilon() * 10.0f;
+}
+
+auto main() -> int
+{
+    // Size of 1D arrays to be used in convolution integral
+    // Here instead of "convolution kernel" the term "filter" is used because kernel has a different meaning in GPU
+    // programming. Secondly filter array is not reversed. Implemented like a convolutional layer in CNN.
+    constexpr size_t filterSize = 3;
+    using DataType = float;
+    constexpr size_t inputSize = 8;
+    constexpr std::array<DataType, inputSize> expectedOutput = {0.8f, 1.4f, 2.0f, 2.6f, 3.2f, 3.8f, 4.4f, 2.3f};
+
+    // Define the index domain
+    using Dim = alpaka::DimInt<1u>;
+    // Index type
+    using Idx = std::size_t;
+
+    // Define the accelerator
+    using DevAcc = alpaka::ExampleDefaultAcc<Dim, Idx>;
+    using QueueProperty = alpaka::Blocking;
+    using QueueAcc = alpaka::Queue<DevAcc, QueueProperty>;
+    using BufAcc = alpaka::Buf<DevAcc, DataType, Dim, Idx>;
+
+    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<DevAcc>() << '\n';
+
+    auto const platformHost = alpaka::PlatformCpu{};
+    auto const devHost = alpaka::getDevByIdx(platformHost, 0);
+
+    // Select a device
+    auto const platformAcc = alpaka::Platform<DevAcc>{};
+    auto const devAcc = alpaka::getDevByIdx(platformAcc, 0);
+
+    // Create a queue on the device
+    QueueAcc queue(devAcc);
+
+    // Allocate memory host input
+    auto hostInputMemory = alpaka::allocBuf<DataType, Idx>(devHost, inputSize);
+
+    // Fill array with data
+    for(size_t i = 0; i < inputSize; i++)
+        hostInputMemory[i] = static_cast<DataType>(i + 1);
+
+    // Allocate memory host filter
+    auto hostFilterMemory = alpaka::allocBuf<DataType, Idx>(devHost, filterSize);
+
+    // Fill array with any data
+    for(size_t i = 0; i < filterSize; i++)
+        hostFilterMemory[i] = static_cast<DataType>(i + 1) / 10.0f;
+
+    // Allocate memory in device
+    BufAcc inputDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, inputSize);
+    BufAcc filterDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, filterSize);
+    BufAcc outputDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, static_cast<Idx>(inputSize));
+
+    // Copy input and filter (convolution kernel array) from host to device
+    alpaka::memcpy(queue, inputDeviceMemory, hostInputMemory, inputSize);
+    alpaka::memcpy(queue, filterDeviceMemory, hostFilterMemory, filterSize);
+    // Make sure memcpy finished.
+    alpaka::wait(queue);
+    using Vec = alpaka::Vec<Dim, Idx>;
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+
+    auto const elementsPerThread = Vec::all(static_cast<Idx>(1));
+    // Grid size
+    auto const threadsPerGrid = inputSize;
+    WorkDiv const workDiv = alpaka::getValidWorkDiv<DevAcc>(
+        devAcc,
+        threadsPerGrid,
+        elementsPerThread,
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
+
+    // Instantiate the kernel (gpu code) function-object
+    ConvolutionKernel convolutionKernel;
+
+    // Native pointers needed for the kernel execution function
+    DataType* nativeFilterDeviceMemory = alpaka::getPtrNative(filterDeviceMemory);
+    DataType* nativeInputDeviceMemory = alpaka::getPtrNative(inputDeviceMemory);
+    DataType* nativeOutputDeviceMemory = alpaka::getPtrNative(outputDeviceMemory);
+
+    // Run the kernel
+    alpaka::exec<DevAcc>(
+        queue,
+        workDiv,
+        convolutionKernel,
+        nativeInputDeviceMemory,
+        nativeFilterDeviceMemory,
+        nativeOutputDeviceMemory,
+        inputSize,
+        filterSize);
+
+    // Allocate memory on host
+    auto resultGpuHost = alpaka::allocBuf<DataType, Idx>(devHost, inputSize);
+    // Copy from device memory to host
+    alpaka::memcpy(queue, resultGpuHost, outputDeviceMemory, inputSize);
+    alpaka::wait(queue);
+
+    bool allEqual{true};
+    // Print result array at the host
+    for(size_t i{0}; i < inputSize; i++)
+    {
+        std::cout << "output[" << i << "]:" << std::setprecision(3) << resultGpuHost[i] << "\n";
+        // Compare with the reference output
+        bool fuzzyEqual = FuzzyEqual(resultGpuHost[i], expectedOutput[i]);
+        allEqual = allEqual && fuzzyEqual;
+    }
+    if(!allEqual)
+    {
+        std::cout << "Error: Some convolution results doesn't match!\n";
+        return EXIT_FAILURE;
+    }
+    std::cout << "All results are correct!\n";
+    return EXIT_SUCCESS;
+}
--- a/example/CMakeLists.txt
+++ b/example/CMakeLists.txt
@@ -17,6 +17,7 @@ project("alpakaExamples" LANGUAGES CXX)
 
 add_subdirectory("bufferCopy/")
 add_subdirectory("complex/")
+add_subdirectory("convolution1D/")
 add_subdirectory("counterBasedRng/")
 add_subdirectory("heatEquation/")
 add_subdirectory("helloWorld/")
--- a/example/convolution1D/CMakeLists.txt
+++ b/example/convolution1D/CMakeLists.txt
@@ -1,0 +1,47 @@
+#
+# Copyright 2023 Erik Zenker, Benjamin Worpitz, Jan Stephan
+# SPDX-License-Identifier: ISC
+#
+
+################################################################################
+# Required CMake version.
+
+cmake_minimum_required(VERSION 3.22)
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+################################################################################
+# Project.
+
+set(_TARGET_NAME convolution1D)
+
+project(${_TARGET_NAME} LANGUAGES CXX)
+
+#-------------------------------------------------------------------------------
+# Find alpaka.
+
+if(NOT TARGET alpaka::alpaka)
+    option(alpaka_USE_SOURCE_TREE "Use alpaka's source tree instead of an alpaka installation" OFF)
+
+    if(alpaka_USE_SOURCE_TREE)
+        # Don't build the examples recursively
+        set(alpaka_BUILD_EXAMPLES OFF)
+        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../.." "${CMAKE_BINARY_DIR}/alpaka")
+    else()
+        find_package(alpaka REQUIRED)
+    endif()
+endif()
+
+#-------------------------------------------------------------------------------
+# Add executable.
+
+alpaka_add_executable(
+    ${_TARGET_NAME}
+    src/convolution1D.cpp)
+target_link_libraries(
+    ${_TARGET_NAME}
+    PUBLIC alpaka::alpaka)
+
+set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER example)
+
+add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME})


SimeonEhrig:
@mehmetyusufoglu A hint, you can use a draft instead a pull request, if you nobody should merge to early. There is option on the right corner to covert the PR to a draft.

image

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+    template<typename TAcc, typename TElem>
+    ALPAKA_FN_ACC auto operator()(
+        TAcc const& acc,
+        TElem const* const input,
+        TElem const* const filter,
+        TElem* output,
+        const std::size_t inputSize,
+        const std::size_t filterSize) const -> void
+    {
+        using Dim = alpaka::Dim<TAcc>;
```
bernhardmgruber:
```suggestion
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+        using Vec = alpaka::Vec<Dim, Idx>;
+        using VecInt = alpaka::Vec<alpaka::DimInt<1u>, Idx>;
+
+
+        Vec const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+        Vec const globalThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+
+        // Map the three dimensional thread index into a
+        // one dimensional thread index space
+        auto const linearizedGlobalThreadIdx = alpaka::mapIdx<1u>(globalThreadIdx, globalThreadExtent)[0];
```
bernhardmgruber: There should not be a need to map indices, because you already launch the kernel as 1D.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+        if(linearizedGlobalThreadIdx < static_cast<unsigned>(inputSize))
+        {
+            int halfFilterSize = filterSize / 2;
```
bernhardmgruber:
```suggestion
            int const halfFilterSize = filterSize / 2;
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // Define the index domain
+    using Dim = alpaka::DimInt<1u>;
+    // Index type
+    using Idx = std::size_t;
+
+    // Define the accelerator
+    using DevAcc = alpaka::ExampleDefaultAcc<Dim, Idx>;
+    using QueueProperty = alpaka::Blocking;
+    using QueueAcc = alpaka::Queue<DevAcc, QueueProperty>;
+    using DataType = float;
```
bernhardmgruber: Since you hardcoded `float` and used FP constents, e.g. `1.4f`, in several other places as well, you can just use `float` everywhere.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    // Fill array with any data
+    for(size_t i = 0; i < FilterSize; i++)
+        nativeHostFilterMemory[i] = static_cast<DataType>(i + 1) / 10.0f;
+
+    // Allocate memory in device
+    alpaka::Buf<DevAcc, DataType, Dim, Idx> inputDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, InputSize);
+    alpaka::Buf<DevAcc, DataType, Dim, Idx> filterDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, FilterSize);
+    alpaka::Buf<DevAcc, DataType, Dim, Idx> outputDeviceMemory
+        = alpaka::allocBuf<DataType, Idx>(devAcc, static_cast<Idx>(InputSize));
```
bernhardmgruber: I don't think the buffer type helps here. I would have the compiler deduce it:
```suggestion
    auto inputDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, InputSize);
    auto filterDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, FilterSize);
    auto outputDeviceMemory = alpaka::allocBuf<DataType, Idx>(devAcc, static_cast<Idx>(InputSize));
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    // Copy input and filter (convolution kernel array) from host to device
+    alpaka::memcpy(queue, inputDeviceMemory, hostInputMemory, InputSize);
+    alpaka::memcpy(queue, filterDeviceMemory, hostFilterMemory, FilterSize);
+
+    DataType* nativeFilterDeviceMemory = alpaka::getPtrNative(filterDeviceMemory);
+    DataType* nativeInputDeviceMemory = alpaka::getPtrNative(inputDeviceMemory);
+    DataType* nativeOutputDeviceMemory = alpaka::getPtrNative(outputDeviceMemory);
```
bernhardmgruber: Move those variables closer to before they are used.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+
+    auto const elementsPerThread = Vec::all(static_cast<Idx>(1));
+    auto const threadsPerGrid = gridSize;
+    WorkDiv const workDiv = alpaka::getValidWorkDiv<DevAcc>(
+        devAcc,
+        threadsPerGrid,
+        elementsPerThread,
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
```
bernhardmgruber: I would avoid some of these copies to make the code denser. 
```suggestion
    using Vec = alpaka::Vec<Dim, Idx>;
    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;

    auto const elementsPerThread = Vec::all(static_cast<Idx>(1));
    WorkDiv const workDiv = alpaka::getValidWorkDiv<DevAcc>(
        devAcc,
        InputSize,
        elementsPerThread,
        false,
        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+        nativeInputDeviceMemory,
+        nativeFilterDeviceMemory,
+        nativeOutputDeviceMemory,
+        InputSize,
+        FilterSize);
+    alpaka::wait(queue);
+
+    // Allocate memory on host
+    alpaka::Buf<alpaka::DevCpu, DataType, Dim, Idx> resultGpuHost
+        = alpaka::allocBuf<DataType, Idx>(devHost, InputSize);
```
bernhardmgruber:
```suggestion
    auto resultGpuHost = alpaka::allocBuf<DataType, Idx>(devHost, InputSize);
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // Copy from device memory to host
+    alpaka::memcpy(queue, resultGpuHost, outputDeviceMemory, InputSize);
+
+    bool allEqual{true};
+    // Print result array at the host
+    for(size_t i{0}; i < InputSize; i++)
+    {
+        std::cout << "output[" << i << "]:" << std::setprecision(3) << resultGpuHost[i] << std::endl;
```
bernhardmgruber: No need for a stream flush in every iteration.
```suggestion
        std::cout << "output[" << i << "]:" << std::setprecision(3) << resultGpuHost[i] << "\n";
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+ * @brief Function template to compare floating point values
+ *
+ */
+template<typename FloatingType>
+typename std::enable_if<std::is_floating_point<FloatingType>::value, bool>::type FuzzyEqual(
+    FloatingType a,
+    FloatingType b)
+{
+    return std::fabs(a - b) < std::numeric_limits<FloatingType>::epsilon() * 10;
+}
```
bernhardmgruber: Just inline the code at the callsite to keep it simpler.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,204 @@
...
+    // Copy from device memory to host
+    alpaka::memcpy(queue, resultGpuHost, outputDeviceMemory, InputSize);
+
+    bool allEqual{true};
+    // Print result array at the host
+    for(size_t i{0}; i < InputSize; i++)
+    {
+        std::cout << "output[" << i << "]:" << std::setprecision(3) << resultGpuHost[i] << std::endl;
+        // Compare with the reference output
+        bool fuzzyEqual = FuzzyEqual<DataType>(resultGpuHost[i], ExpectedOutput[i]);
```
bernhardmgruber:
```suggestion
        bool fuzzyEqual = std::abs(resultGpuHost[i] - ExpectedOutput[i]) < std::numeric_limits<FloatingType>::epsilon() * 10;
```

bernhardmgruber (commented):
Here is some feedback for a start.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: There are usually aliases used in the examples, like BufAcc below. Instead of auto or an alias, I prefered explicit use since this is an example, being explicit is better?
using BufAcc = alpaka::Buf<DevAcc, DataType, Dim, Idx>

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: Ok, for preventing flushing.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: Removed one variable (`gridSize`), kept the other one ( `threadsPerGrid` ) since this is an example, seems more informative.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: I changed to an alias now, shorter, like in almost all examples. I remember Rene was talking about failures of some of the compilers.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: I wanted the user use another function/operator when the data type is not floating point.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: I had removed that line from the kernel. Only constexpr result array is float, which is used for comparison of result.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+//!
+//! 1D convolution example: Creates two 1D arrays, calculates the convolution integral using those arrays.
+//! Array sizes are hardcoded.
+//!
+
+// Size of 1D arrays to be used in convolution integral
+// In signal processing domain, the term "kernel" is used for the matrix or array used in convolution integral.
+// Here instead of "kernel" the term "filter" is used because kernel has a different meaning in GPU programming.
+constexpr size_t FilterSize = 3;
+constexpr size_t InputSize = 8;
```
SimeonEhrig: I this example, it will not cause any problems, but I would avoid to use globals where it is possible. Most of the people simply copies the code and don't think about bad practice.

I suggest to move it in the begin of the `main()`.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+constexpr float ExpectedOutput[InputSize] = {0.8f, 1.4f, 2.0f, 2.6f, 3.2f, 3.8f, 4.4f, 2.3f};
```
SimeonEhrig: I suggest to use `std::array`.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+     *  @param Empty output array to be filled
+     *  @param Input array size
+     *  @param Filter size
+     */
+    template<typename TAcc, typename TElem>
+    ALPAKA_FN_ACC auto operator()(
+        TAcc const& acc,
+        TElem const* const input,
+        TElem const* const filter,
+        TElem* output,
```
SimeonEhrig:
```suggestion
        TElem* const output,
```

mehmetyusufoglu: Ok the pointer is const.

SimeonEhrig: Yes, this is a confusing part of C :grimacing:

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: done, thanks.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: done thanks.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+            // Calculate sum of multiplications of corresponding elements
+            for(size_t i = 0; i < filterSize; ++i)
+            {
+                uint32_t inputIndex = globalThreadIdx[0] - halfFilterSize + i;
+                if(inputIndex >= 0 && inputIndex < inputSize)
+                {
+                    result += input[inputIndex] * filter[i];
+                }
+            }
+            output[globalThreadIdx[0]] = result;
```
bernhardmgruber: I would avoid rechecking the condition inside the loop and adjust the loop range. Also, your code contains an underflow (`globalThreadIdx[0] - halfFilterSize` may be below 0, which would then be a huge number, but the `+ i` would then overflow again), which works by accident because `uint32_t` has defined wrap-around semantics, and `inputSize` is not close to `UINT_MAX`.
```suggestion
            uint32_t const halfFilterSize = filterSize / 2;
            TElem result = 0.0f;
            // Calculate sum of multiplications of corresponding elements
            auto const start = static_cast<uint32_t>(std::max(static_cast<int>(globalThreadIdx[0]) - halfFilterSize, 0));
            auto const stop = std::min(globalThreadIdx[0] + halfFilterSize, inputSize);
            for(size_t i = start; i <= stop; ++i)
                    result += input[i] * filter[i-globalThreadIdx[0]];
```
This may not be the nicest version and you could consider still letting the loop run from 0 to `filterSize` and offset the index into `input` instead.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+ * @brief Function template to compare floating point values
+ *
+ */
+template<typename FloatingType>
+typename std::enable_if<std::is_floating_point<FloatingType>::value, bool>::type FuzzyEqual(
+    FloatingType a,
+    FloatingType b)
+{
+    return std::fabs(a - b) < std::numeric_limits<FloatingType>::epsilon() * 10;
+}
```
bernhardmgruber:
```suggestion
void FuzzyEqual(float a, float b)
{
    return std::fabs(a - b) < std::numeric_limits<float>::epsilon() * 10;
}
```
I would really simplify this to make the example less complex.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+auto main() -> int
+{
+// Fallback for the CI with disabled sequential backend
+#if defined(ALPAKA_CI) && !defined(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED)
+    return EXIT_SUCCESS;
+#else
```
bernhardmgruber: Is this actually still needed? I think you cannot even compile the unit tests without the sequential backend, so when could this ever trigger in the CI?

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // Create a queue on the device
+    QueueAcc queue(devAcc);
+
+    // Allocate memory host input
+    auto hostInputMemory = alpaka::allocBuf<DataType, Idx>(devHost, inputSize);
+    DataType* nativeHostInputMemory = alpaka::getPtrNative(hostInputMemory);
+
+    // Fill array with data
+    for(size_t i = 0; i < inputSize; i++)
+        nativeHostInputMemory[i] = static_cast<DataType>(i + 1);
```
bernhardmgruber: I think we can directly access host buffers here:
```suggestion
    auto hostInputMemory = alpaka::allocBuf<DataType, Idx>(devHost, inputSize);

    // Fill array with data
    for(size_t i = 0; i < inputSize; i++)
        hostInputMemory[i] = static_cast<DataType>(i + 1);
```
Applies to other places as well.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: yes, i had that in mind first. Then in the last commit today, I converted int to uint32_t by mistake after seeing "use types with explicit sizes" in the documentation. I will simply use int32_t.  Thanks.

mehmetyusufoglu: I think it is ok now. Thanks good point.

mehmetyusufoglu: About checking condition inside the loop: is it because this is kernel code ( warp coalescence etc) or just for optimization in general?

example/convolution1D/src/convolution1D.cpp
psychocoderHPC: I agree simpler should be better in the example.


```C++
auto FuzzyEqual(float a, float b) ...
```

psychocoderHPC: `10` should be a `10f`

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+        nativeOutputDeviceMemory,
+        inputSize,
+        filterSize);
+    alpaka::wait(queue);
+
+    // Allocate memory on host
+    auto resultGpuHost = alpaka::allocBuf<DataType, Idx>(devHost, inputSize);
+    // Copy from device memory to host
+    alpaka::memcpy(queue, resultGpuHost, outputDeviceMemory, inputSize);
+
```
psychocoderHPC, line 161: `alpaka::wait(queue);` is missing to be sure that the copied values are on the host.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,200 @@
...
+    alpaka::exec<DevAcc>(
+        queue,
+        workDiv,
+        convolutionKernel,
+        nativeInputDeviceMemory,
+        nativeFilterDeviceMemory,
+        nativeOutputDeviceMemory,
+        inputSize,
+        filterSize);
+    alpaka::wait(queue);
```
psychocoderHPC: no `wait` is required because the next operation is in the same queue

psychocoderHPC (changes requested):

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: ok I added wait after the second operation. Thanks a lot.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: I checked now, this exists in around 10 examples.  I will check CI procedure. Thanks.

example/convolution1D/src/convolution1D.cpp
bernhardmgruber: > I will simply use int32_t.

You can also just use `int`.

bernhardmgruber: > About checking condition inside the loop: is it because this is kernel code ( warp coalescence etc) or just for optimization in general?

General optimization.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: Great, thanks.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: int is not allowed in [coding guidelines ](https://alpaka.readthedocs.io/en/latest/dev/style.html)of alpaka.

bernhardmgruber: Hmm, I think the coding guidelines need a rewrite then. I think there is nothing wrong with using an `int` as a loop counter or distance measure.

bernhardmgruber: #2231

mehmetyusufoglu: Inside the new loop you suggested, (i - globalThreadIndx(0) ) becomes  -1.  I needed another check in the loop. Another solution is padding, I did it in the 2d example therefore this can be left like this.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: done, thanks.

example/convolution1D/src/convolution1D.cpp
bernhardmgruber: > Inside the new loop you suggested, (i - globalThreadIndx(0) ) becomes -1.

Right, because I forgot a `+halfFilterSize` and a `-1` :D Consider the code above (and below) pseudocode. Try:
```c++
            auto const start = static_cast<uint32_t>(std::max(static_cast<int>(globalThreadIdx[0]) - halfFilterSize, 0));
            auto const stop = std::min(globalThreadIdx[0] + halfFilterSize, inputSize-1);
            for(size_t i = start; i <= stop; ++i)
                    result += input[i] * filter[i+halfFilterSize-globalThreadIdx[0]];
```
> I needed another check in the loop.

It should not be necessary if the loop bounds are correct.

> Another solution is padding, I did it in the 2d example therefore this can be left like this.

Right, which may be a better solution, since it makes the kernel simpler.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,188 @@
...
+        const std::size_t inputSize,
+        const std::size_t filterSize) const -> void
+    {
+        using Idx = alpaka::Idx<TAcc>;
+        using Vec = alpaka::Vec<alpaka::Dim<TAcc>, Idx>;
+        using VecInt = alpaka::Vec<alpaka::DimInt<1u>, Idx>;
+
+
+        auto const globalThreadIdxX = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        Vec const globalThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
```
bernhardmgruber:
```suggestion
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,188 @@
...
+        TAcc const& acc,
+        TElem const* const input,
+        TElem const* const filter,
+        TElem* const output,
+        const std::size_t inputSize,
+        const std::size_t filterSize) const -> void
+    {
+        using Idx = alpaka::Idx<TAcc>;
+        using Vec = alpaka::Vec<alpaka::Dim<TAcc>, Idx>;
+        using VecInt = alpaka::Vec<alpaka::DimInt<1u>, Idx>;
```
bernhardmgruber:
```suggestion
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,188 @@
...
+            output[globalThreadIdxX] = result;
+        }
+    }
+};
+
+/**
+ * @brief Function template to compare floating point values
+ *
+ */
+auto FuzzyEqual(float a, float b) -> bool
```
bernhardmgruber:
```suggestion
auto FuzzyEqual(float a, float b) -> bool
```
It's no longer a `template` and I would also not need the comment, since the function name suggests what the function is about.

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+auto main() -> int
+{
+    // Size of 1D arrays to be used in convolution integral
+    // Here instead of "convolution kernel" the term "filter" is used because kernel has a different meaning in GPU
+    // programming. Secondly filter array is not reversed. Implemented like a convolutional layer in CNN.
+    constexpr size_t filterSize = 3;
+    constexpr size_t inputSize = 8;
+    using DataType = float;
+    constexpr std::array<DataType, inputSize> expectedOutput = {0.8f, 1.4f, 2.0f, 2.6f, 3.2f, 3.8f, 4.4f, 2.3f};
```
bernhardmgruber:
```suggestion
    using DataType = float;
    constexpr std::array<DataType, inputSize> expectedOutput = {0.8f, 1.4f, 2.0f, 2.6f, 3.2f, 3.8f, 4.4f, 2.3f};
    constexpr size_t inputSize = expectedOutput.size();
```
Don't repeat yourself :)

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,188 @@
...
+    // Index type
+    using Idx = std::size_t;
+
+    // Define the accelerator
+    using DevAcc = alpaka::ExampleDefaultAcc<Dim, Idx>;
+    using QueueProperty = alpaka::Blocking;
+    using QueueAcc = alpaka::Queue<DevAcc, QueueProperty>;
+    using BufAcc = alpaka::Buf<DevAcc, DataType, Dim, Idx>;
+
+    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<DevAcc>() << std::endl;
```
bernhardmgruber:
```suggestion
    std::cout << "Using alpaka accelerator: " << alpaka::getAccName<DevAcc>() << '\n';
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+    auto const elementsPerThread = Vec::all(static_cast<Idx>(1));
+    // Grid size
+    auto const threadsPerGrid = inputSize;
+    WorkDiv const workDiv = alpaka::getValidWorkDiv<DevAcc>(
+        devAcc,
+        threadsPerGrid,
+        elementsPerThread,
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
```
bernhardmgruber, line 135:
```suggestion
    WorkDiv const workDiv = alpaka::getValidWorkDiv<DevAcc>(
        devAcc,
        threadsPerGrid,
        elementsPerThread,
        false);
```
If you want, since the last argument is the default.

bernhardmgruber (commented):
Last piece of feedback. The rest looks good.

Please squash all commits as well!

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: done, thanks.

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: explicit is better i think

example/convolution1D/src/convolution1D.cpp
mehmetyusufoglu: done thans.

bernhardmgruber (dismissed):

example/convolution1D/src/convolution1D.cpp
```diff
@@ -0,0 +1,178 @@
+/* Copyright 2023 Benjamin Worpitz, Erik Zenker, Bernhard Manfred Gruber, Jan Stephan
```
bernhardmgruber: You may want to add yourself here. Also, I am not sure if everybody on this list contributed to your example :)

mehmetyusufoglu: What is the policy about this? So I am adding only current team members contributed like 
_"Copyright 2023  Bernhard Manfred Gruber, Simeon Ehrig, Rene Widera, Mehmet Yusufoglu."_

bernhardmgruber: I cannot give you legal advice. In principle, if you create something original, then you hold copyright for this creation. I assume you created at least a part of the 1D Convolution here, so you hold copyright to this "invention". If you copied some pieces of code from somewhere else, other people may hold copyright too, which you have to include. If you just included small changes from elsewhere, like a few lines of code that cannot be called an "original work", you do not need to name the author. Here are some [guidelines from the GPU project](https://www.gnu.org/prep/maintain/html_node/Legally-Significant.html).

bernhardmgruber: However, you may find some people on the team, that are pedantic about adding anybody to the copyright header who flipped a single character in the source file :) It's ultimately also some amount of personal judgement.

But for this case here: if you didn't fully copy your example from the web, I am sure it contains original work from you!

bernhardmgruber (approved):

psychocoderHPC (approved):

PR #2226, opened 2024-01-16
Include global memory in AccDevProps

### LINKED ISSUE #2194: `AccDevProps` should include global memory available
I was looking for the available global memory of a device and was disappointed that `AccDevProps` does not offer a corresponding data member. The shared memory size is available. Please consider including an additional member for global memory.

bernhardmgruber:
Fixes: #2194

### DIFF
--- a/include/alpaka/acc/AccDevProps.hpp
+++ b/include/alpaka/acc/AccDevProps.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2020 Benjamin Worpitz, Bernhard Manfred Gruber
+/* Copyright 2024 Benjamin Worpitz, Bernhard Manfred Gruber
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -6,9 +6,6 @@
 
 #include "alpaka/core/Common.hpp"
 #include "alpaka/vec/Vec.hpp"
-
-#include <string>
-#include <vector>
 
 namespace alpaka
 {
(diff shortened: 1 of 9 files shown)


include/alpaka/acc/AccDevProps.hpp
```diff
@@ -23,37 +20,14 @@ namespace alpaka
...
-            : m_gridBlockExtentMax(gridBlockExtentMax)
-            , m_blockThreadExtentMax(blockThreadExtentMax)
-            , m_threadElemExtentMax(threadElemExtentMax)
-            , m_gridBlockCountMax(gridBlockCountMax)
-            , m_blockThreadCountMax(blockThreadCountMax)
-            , m_threadElemCountMax(threadElemCountMax)
-            , m_multiProcessorCount(multiProcessorCount)
-            , m_sharedMemSizeBytes(sharedMemSizeBytes)
-        {
-        }
```
fwyzard, line 44: Is this removed in favour of aggregate initialisation ?

I understand that it simplifies the class, but it also makes it easier to break the code if somebody reorders the data members.

bernhardmgruber: Right, but I can also break the class by reordering constructor arguments. And the second benefit of `AccDevProps` being an aggregate is that we can soon use designated initializers:
```c++
return {
    .m_multiProcessorCount = ...,
    .m_gridBlockExtentMax = ...,
    ... };
```
which is superior to the comments we have now:
```c++
            return {// m_multiProcessorCount
                    alpaka::core::clipCast<TIdx>(device.template get_info<sycl::info::device::max_compute_units>()),
                    // m_gridBlockExtentMax
                    getExtentVecEnd<TDim>(Vec<DimInt<3u>, TIdx>(
                        // WARNING: There is no SYCL way to determine these values
                        std::numeric_limits<TIdx>::max(),
                        std::numeric_limits<TIdx>::max(),
                        std::numeric_limits<TIdx>::max())),
                    // m_gridBlockCountMax
                    std::numeric_limits<TIdx>::max(),
                    // m_blockThreadExtentMax
                    max_threads_dim_vec,
                    // m_blockThreadCountMax
                    alpaka::core::clipCast<TIdx>(device.template get_info<sycl::info::device::max_work_group_size>()),
                    // m_threadElemExtentMax
                    Vec<TDim, TIdx>::all(std::numeric_limits<TIdx>::max()),
                    // m_threadElemCountMax
                    std::numeric_limits<TIdx>::max(),
                    // m_sharedMemSizeBytes
                    device.template get_info<sycl::info::device::local_mem_size>(),
                    // m_globalMemSizeBytes
                    getMemBytes(dev)};
```

fwyzard: > Right, but I can also break the class by reordering constructor arguments.

Well, yes, and that's true for any function declaration.

I would expect that somebody changing a function signature would be wary of checking and updating the call sites; somebody reordering data members may not realise the impact it will have.

bernhardmgruber: > I would expect that somebody changing a function signature would be wary of checking and updating the call sites;

Fully agree.

> somebody reordering data members may not realise the impact it will have.

Well, if the type is an aggregate you should also check all uses, but I know this is harder, since how would you grep for a `return { x, z, y };` or a `v.push_back({x, z, y});` when changing a 3 component struct named `Vec3`? 

If you feel safer, I can add a unit test for the order of members.

fwyzard: Maybe it would be enough to add a comment reminding the future developers that this class uses aggregate initialisation, so care should be taken if reordering the data members.

bernhardmgruber: Added comment and test

psychocoderHPC (approved):

PR #2228, opened 2024-01-22
2D Convolution example using global and shared memory

mehmetyusufoglu:
An example: A 2D Convolutional filter applied to a matrix. The values of filter-matrix were initially kept in constant memory at the first commit. But due to Gitlab pipeline error "_The SYCL backend does not support global device constants_"; in the second commit, constant memory usage has been removed.

 - Kernel1: Global memory is used, without tiling.
 - Kernel2: Uses tiling. Block size is assumed to be equal to tile size. First, the tile is copied to shared memory, since an element in the tile would be accessed many times.  Each block works on the domain of one tile. But at the border of the tile, some external matrix values are needed ( at the border with another tile) then those matrix values are taken from the global memory.


example/convolution2D/src/convolution2D.cpp
```diff
@@ -0,0 +1,374 @@
...
+        auto const blockThreadIdx = alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc);
+        auto const blockThreadIdx1D = alpaka::mapIdx<1u>(blockThreadIdx, blockThreadExtent)[0u];
+        // Get elements from 2-element arrays
+        auto const [blockThreadExtentY, blockThreadExtentX] = blockThreadExtent;
+        auto const [blockThreadY, blockThreadX] = blockThreadIdx;
+        auto const [gridBlockExtentY, gridBlockExtentX] = gridBlockExtent;
+        // Allocate shared memory
+        auto* const sharedN = alpaka::getDynSharedMem<TElem>(acc);
+        // Fill shared memory of device so that tile items are accessed from shared memory
+        if(row < matrixHeight && col < matrixWidth && blockThreadIdx1D < blockThreadExtent.prod())
```
psychocoderHPC, line 126: On the host side you use `getValidWorkDiv` This means you will have one thread for some alpaka accelerators.
I know you wrote that the block size must be equal to the tile size but you do not enforce  it e.g. with an `ALPAKA_VERIFY`
If you have only one thread in the block you can simply iterate over the shared memory to fill it.

example/convolution2D/src/convolution2D.cpp
```diff
@@ -0,0 +1,374 @@
...
+            sharedN[blockThreadIdx1D] = 0.0f;
+        }
+
+        // Wait for the block fills the shared memory with the tile of the main matrix
+        alpaka::syncBlockThreads(acc);
+
+        if(col < matrixWidth && row < matrixHeight)
+        {
+            TElem pValue{0.0f};
+            for(int32_t fRow = 0; fRow < filterWidth; fRow++)
```
psychocoderHPC, line 141: Not sure if all of this will work with a single thread per block. Have you checked the code with the `serial` backend?

example/convolution2D/src/convolution2D.cpp
```diff
@@ -0,0 +1,374 @@
...
+    // Index type
+    using Idx = std::uint32_t;
+    using Vec = alpaka::Vec<Dim, Idx>;
+    // Define the accelerator
+    using DevAcc = alpaka::ExampleDefaultAcc<Dim, Idx>;
+    using QueueAcc = alpaka::Queue<DevAcc, alpaka::NonBlocking>;
+
+    using DataType = float;
+    static constexpr Idx filterWidth = 5;
+    static constexpr Idx matrixWidth = 10;
```
psychocoderHPC: I am not sure if this very small size is good for the example. In case  there is an error in the implementation if you use more than one block this will not show up because `10x10` will be computed mostly with a single block without block paralism.

psychocoderHPC: Can we at least use `128x128`?

psychocoderHPC (changes requested):

example/convolution2D/src/convolution2D.cpp
mehmetyusufoglu: Yes, It was tested. I tested again. Works.

example/convolution2D/src/convolution2D.cpp
mehmetyusufoglu: Yes but if there is one thread per block it means it is not a GPU (or it is not good to use GPU); so we dont know which level of memory is used ?

example/convolution2D/src/convolution2D.cpp
mehmetyusufoglu: Done, thanks.

example/convolution2D/src/convolution2D.cpp
mehmetyusufoglu: If there is one block, whole block is loaded into the shared memory in the code.

example/convolution2D/src/convolution2D.cpp
```diff
@@ -0,0 +1,381 @@
...
+
+auto main() -> int
+{
+    // Define the index domain
+    using Dim = alpaka::DimInt<2>;
+    // Index type
+    using Idx = std::uint32_t;
+    using Vec = alpaka::Vec<Dim, Idx>;
+    // Define the accelerator
+    using DevAcc = alpaka::AccCpuSerial<Dim, Idx>;
```
psychocoderHPC: Let us have an offline chat, I am wondering why you use the serial accelerator here. We should not write examples only for CPU or only for GPU. 
An example should run on all platforms.

psychocoderHPC (changes requested):

example/convolution2D/src/convolution2D.cpp
```diff
@@ -0,0 +1,381 @@
...
+//! Results can be tested by comparing with the results of the Matlab call: Y =
+//! filter2(FilterMatrix,InputMatrix,'same');
+
+/**
+ * @brief 2D Convolutional Filter using only global memory for the input-matrix and the filter-matrix
+ */
+struct ConvolutionKernel2DGlobalMemory
+{
+    /**
+    @tparam TAcc Accelerator type
```
psychocoderHPC: please use all doxygen command starting with `\` in alpaak.

example/convolution2D/src/convolution2D.cpp
```diff
@@ -0,0 +1,381 @@
...
+            TElem*, // output array
+            int32_t const, // matrixWidth
+            int32_t const, // matrixHeight
+            TElem const* const, // filter
+            int32_t const, // filter width
+            int32_t const, // allocated input width
+            int32_t const) // allocated filter width
+        {
+            // Reserve the buffer, buffers size is the number of elements in a block (tile)
+            return static_cast<std::size_t>(1u * blockThreadExtent.prod() * threadElemExtent.prod()) * sizeof(TElem);
```
psychocoderHPC: `1u` can be removed

psychocoderHPC (changes requested):

example/convolution2D/src/convolution2D.cpp
mehmetyusufoglu: done, thanks.

example/convolution2D/src/convolution2D.cpp
mehmetyusufoglu: This was for a test run. I changed to _ExampleDefaultAcc_. Thanks.

psychocoderHPC (approved):

PR #2232, opened 2024-01-23
Revise coding guidelines

### LINKED ISSUE #2231: Update coding guidelines to match reality
The current [alpaka coding guidelines](https://alpaka.readthedocs.io/en/latest/dev/style.html) may be out of date. For example, during a review @mehmetyusufoglu rightfully pointed out, that the use of non-sized built-in types like `int` or `unsigned` is **forbidden** in alpaka.

I think this is nonsensical.

I propose the following changes:
* Remove all rules that are handled by clang-format.
* Do not recommend variable initialization using paranthesis: `int i(0);` should not be the default and even discouraged IMO. I think we also regularly break the rule against `int i = 0;`. I write lots code like that, and people seemed fine with it? Encouraging parenthesis, `int i{0};`, is fine.
* Drop suggesting comment blocks before type or function definitions. It only adds noise.
* Drop the recommendation to put comments next to includes telling which feature is included from that header. This is unmaintainable, does not help IMO, and will go away in C++23 anyway.
* In the future, we should also remove all rules that clang-tidy is taking care of, e.g. identifier namings.

bernhardmgruber:
This PR tries to get the coding guidelines up to the status-quo, but does not propose any rule changes. I reworded some sentences while reading.

Fixes: #2231

### DIFF
--- a/docs/source/dev/style.rst
+++ b/docs/source/dev/style.rst
@@ -3,148 +3,83 @@
 Coding Guidelines
 ==================
 
-.. attention::
-   The Coding Guidelines are currently revised
+Formatting
+----------
 
-General
--------
-
-* Use the ``.clang-format`` file supplied in alpaka's top-level directory to format your code. This will handle indentation,
-whitespace and braces automatically. Usage:
-
-.. code-block:: bash
-
-  clang-format-16 -i <sourcefile>
-
-* If you want to format the entire code base execute the following command from alpaka's top-level directory:
-
-.. code-block:: bash
-
-  find example include test -name '*.hpp' -o -name '*.cpp' | xargs clang-format-16 -i
-
-Windows users should use `Visual Studio's native clang-format integration
-<https://devblogs.microsoft.com/cppblog/clangformat-support-in-visual-studio-2017-15-7-preview-1/>`.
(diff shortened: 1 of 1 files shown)


docs/source/dev/style.rst
```diff
@@ -3,148 +3,83 @@
...
+Use the ``.clang-format`` file supplied in alpaka's top-level directory to format your code.
+This will handle indentation, whitespace and braces automatically.
+Checkout ``CONTRIBUTING.md`` on how to run it.
 
 Naming
 ------
 
 * Types are always in PascalCase (KernelExecCuda, BufT, ...) and singular.
-* Variables are always in camelCase (memBufHost, ...) and plural for collections and singular else.
+* Variables are always in camelCase (memBufHost, ...) and singlar by default. Use plural for collections.
```
psychocoderHPC: Mehmet mentioned that there is a typo `singlar` should be `singular`

psychocoderHPC (changes requested):

docs/source/dev/style.rst
bernhardmgruber: Fixed.

psychocoderHPC (approved):

SimeonEhrig (approved):

PR #2236, opened 2024-01-30
Allow non-integral types in Vec generator constructor

sliwowitz:
The `explicit Vec(F&& generator, std::integer_sequence<TVal, Is...>)` accepted a parameter pack of `TVal` which is defined to be the inner vector value and can possible be non-integral, which makes it invalid in pre-C++20 codebases. We've restricted the parameter to an integral type, thus disallowing this construction method for non-integral inner types.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -87,18 +87,22 @@ namespace alpaka
         ALPAKA_FN_HOST_ACC constexpr explicit Vec(
             F&& generator,
             std::void_t<decltype(generator(std::integral_constant<std::size_t, 0>{}))>* ignore = nullptr)
+            : Vec(std::forward<F>(generator), std::make_index_sequence<TDim::value>{})
+        {
+            static_cast<void>(ignore);
+        }
 #else
         template<typename F, std::enable_if_t<std::is_invocable_v<F, std::integral_constant<std::size_t, 0>>, int> = 0>
         ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator)
-#endif
-            : Vec(std::forward<F>(generator), std::make_integer_sequence<TVal, TDim::value>{})
+            : Vec(std::forward<F>(generator), std::make_index_sequence<TDim::value>{})
         {
         }
+#endif
 
     private:
-        template<typename F, TVal... Is>
-        ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator, std::integer_sequence<TVal, Is...>)
-            : m_data{generator(std::integral_constant<TVal, Is>{})...}
+        template<typename F, std::size_t... Is>
+        ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator, std::index_sequence<Is...>)
+            : m_data{generator(std::integral_constant<std::size_t, Is>{})...}
         {
         }
 
--- a/test/unit/vec/src/VecTest.cpp
+++ b/test/unit/vec/src/VecTest.cpp
@@ -468,3 +468,27 @@ TEST_CASE("accessByNameConstexpr", "[vec]")
     STATIC_REQUIRE(v4.z() == 3);
     STATIC_REQUIRE(v4.w() == 4);
 }
+
+TEMPLATE_TEST_CASE("Vec generator constructor", "[vec]", std::size_t, int, unsigned, float, double)
+{
+    // Define a generator function
+    auto generator = [](auto index) { return static_cast<TestType>(index.value + 1); };
+
+    // Create a Vec object using the generator function
+    alpaka::Vec<alpaka::DimInt<5>, TestType> vec(generator);
+
+    // Check that the values in the Vec object are as expected
+    for(std::size_t i = 0; i < 5; ++i)
+    {
+        // Floating point types require a precision check instead of an exact == match
+        if constexpr(std::is_floating_point<TestType>::value)
+        {
+            auto const precision = std::numeric_limits<TestType>::epsilon() * 5; // Arbitrary precision requirement
+            CHECK(std::abs(vec[i] - static_cast<TestType>(i + 1)) < precision);
+        }
+        else
+        {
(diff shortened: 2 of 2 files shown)


mehmetyusufoglu:
Is it possible to add a test for it?

sliwowitz:
I suppose so. I've looked into the current `Vec` test, and it seems we're only testing integral types. Was this class ever supposed to handle floating point values?

include/alpaka/vec/Vec.hpp
```diff
@@ -96,9 +96,9 @@ namespace alpaka
         }
 
     private:
-        template<typename F, TVal... Is>
-        ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator, std::integer_sequence<TVal, Is...>)
-            : m_data{generator(std::integral_constant<TVal, Is>{})...}
+        template<typename F, typename TIdx, TIdx... Is, typename = std::enable_if_t<std::is_integral_v<TIdx>>>
+        ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator, std::integer_sequence<TIdx, Is...>)
+            : m_data{generator(std::integral_constant<TIdx, Is>{})...}
```
bernhardmgruber: It seems very odd to me to constrain a private API, since we basically control all callsites. This looks rather like a design error. Fortunately, I wrote that code so I remember what I tried to do :)

This constructor supports the one above it and should mimic [`std::simd`'s generator constructor](https://en.cppreference.com/w/cpp/experimental/simd/simd/simd), which uses `std::size_t` for the index sequence. For some reason I used `TVal` here. Probably because my brain was thinking about `Vec`'s of integers only.

The correct fix IMO would be to just use `std::index_sequence` and `std::size_t`:
```c++
#else
        template<typename F, std::enable_if_t<std::is_invocable_v<F, std::integral_constant<std::size_t, 0>>, int> = 0>
        ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator)
#endif
            : Vec(std::forward<F>(generator), std::make_index_sequence<TDim::value>{})
        {
        }

    private:
        template<typename F, std::size_t... Is>
        ALPAKA_FN_HOST_ACC constexpr explicit Vec(F&& generator, std::index_sequence<Is...>)
            : m_data{generator(std::integral_constant<std::size_t, Is>{})...}
        {
        }
```

psychocoderHPC: @sliwowitz COuld you have again a look into what @bernhardmgruber wrote?

sliwowitz:
Seems @bernhardmgruber's idea works fine too. Does anyone have tips on what kind of additional tests to write for this? Maybe extend the [current tests](https://github.com/alpaka-group/alpaka/blob/develop/test/unit/vec/src/VecTest.cpp) so that they'd consturct `Vec`s of additional types beside `size_t`, and add a test for the generator constructor (I can't see one in there)

psychocoderHPC:
> Seems @bernhardmgruber's idea works fine too. Does anyone have tips on what kind of additional tests to write for this? Maybe extend the [current tests](https://github.com/alpaka-group/alpaka/blob/develop/test/unit/vec/src/VecTest.cpp) so that they'd consturct `Vec`s of additional types beside `size_t`, and add a test for the generator constructor (I can't see one in there)

Yes sounds good. We should have a test where a `Vec<float,...>` is created and for generators.

bernhardmgruber:
Sure, please add tests for the generator constructor!

bernhardmgruber (dismissed):

bernhardmgruber:
@sliwowitz Could you please add a test so I can proceed merging this PR? Thanks!

psychocoderHPC (dismissed):

psychocoderHPC:
@sliwowitz  ping

test/unit/vec/src/VecTest.cpp
```diff
@@ -468,3 +468,27 @@ TEST_CASE("accessByNameConstexpr", "[vec]")
...
+        // Floating point types require a precision check instead of an exact == match
+        if constexpr(std::is_floating_point<TestType>::value)
+        {
+            auto const precision = std::numeric_limits<TestType>::epsilon() * 5; // Arbitrary precision requirement
+            CHECK(std::abs(vec[i] - static_cast<TestType>(i + 1)) < precision);
+        }
+        else
+        {
+            CHECK(vec[i] == static_cast<TestType>(i + 1));
+        }
```
bernhardmgruber, line 492: Why do you need a comparison with an epsilon? All values fit within the representation of a `float`. Furthermore, if there were any rounding errors, they would be deterministic.

This should be enough:
```suggestion
            CHECK(vec[i] == static_cast<TestType>(i + 1));
```

sliwowitz: This will trigger a warning for unsafe == float comparison, and the test will fail because of that. An alternative would be to put a pragma disabling the warning, or to use `floatEqualExactNoWarning` which basically does exactly that

sliwowitz:
Just a side note, Bernhard's solution actually allows non-integral types in the Vec generator constructor, so I renamed the PR to reflect that..

test/unit/vec/src/VecTest.cpp
bernhardmgruber: Sigh ... alright then.

bernhardmgruber (approved):

test/unit/vec/src/VecTest.cpp
sliwowitz: Yeah... Here it's imho better option than disabling the pragmas that disable the warning - I'd prefer not to use them unless necessary, and it might be argued, that strict equality makes sense in test where the fp number is not being modified (allocations, initializations) whereas here you could imagine some future architecture, that would make the assignment i+1 imprecise, or someone reusing the test for bigger numbers. Not really a solid argument though, just trying to rationalize one of the options (eps comparison vs. pragma disabling the warning) since I don't like either.

psychocoderHPC (approved):

Issue #1264, opened 2021-03-01
Performance test suite


bernhardmgruber:
Alpaka's "aim is to provide performance portability across accelerators through the abstraction (not hiding!) of the underlying levels of parallelism." We should therefore also pay attention to possible performance impacts of changes applied to this library. In order to better assess such performance impacts I want to propose the creation of a performance test suite.

Performance tests are similar to the existing unit tests with some differences:
* We do not need assertions to verify the correctness of the code. That is the job of unit tests.
* Perform enough work so we can rule out noise. But keep them short enough so that the CI does not take ages. A good test length is probably around a few seconds.

When we have a few of these tests, we should setup a CI pipeline which can visualize how runtime performance changes with commits and PRs.

psychocoderHPC:
IMO the performance test should always check for correct results. 
Such tests should be executable without a CI too and in that case, it could be that someone is testing different code versions without running all unit tests.
Additionally, if you write an e.g. `forEach` as your performance test and one implementation has a logical programming mistake which results in a compiler optimization where the `hot` code is optimized out you would not realize the mistake if the performance difference is small.

bernhardmgruber:
> Such tests should be executable without a CI too and in that case, it could be that someone is testing different code versions without running all unit tests.

Absolutely!

> Additionally, if you write an e.g. forEach as your performance test and one implementation has a logical programming mistake which results in a compiler optimization where the hot code is optimized out you would not realize the mistake if the performance difference is small.

You would not notice it when running the performance test suite, yes. But you would notice it when running the unit tests. And I think this is fine, because you also run your unit tests either before committing or at least as part of the CI build later before merging. Adding additional code to verify the results just makes the performance tests more complicated.

Example: filling a few GB buffers with random numbers. Would you want to run the number generator again on the host to verify that all random numbers are generated correctly? Or that the random number distribution is fine? That could easily double or triple the test runtime.

Furthermore, if the performance test consists of largely hot code that you want to benchmark, you can also easily perf it without modifying the code (i.e. disabling the computation of reference results and comparisons).

psychocoderHPC:
> Example: filling a few GB buffers with random numbers. Would you want to run the number generator again on the host to verify that all random numbers are generated correctly? Or that the random number distribution is fine? That could easily double or triple the test runtime.

A result verification is not required to be a point-wise comparison of values. It depends on the test what is correct or wrong. For random numbers, you would most likely check the standard derivation, ...

[...]

PR #2237, opened 2024-01-31
Create benchmarks directory and move babelstream into it

mehmetyusufoglu:
A simple PR. A directory called "_benchmarks_" is created and _babelstream_ example is copied into it. There is a new cmake flag _alpaka_BUILD_BENCHMARKS_.  If this flag is ON then _alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE_ is turned ON (Like _alpaka_BUILD_EXAMPLES_ flag)

The codes under benchmark directory is compiled. But Babelstream example is not run at the CI, as it was in the Examples directory before.


SimeonEhrig:
Maybe we should add a cmake target `all_benchmarks` where we can register and execute all benchmarks.

The benchmarks need to be build in the CI. I see no reason, why we should not enable the benchmarks for all builds. Therefore we can add the CMake argument here:
https://github.com/alpaka-group/alpaka/blob/dcc87b3edc8493796a748e5e551c319c875774c5/script/run_generate.sh#L80-L94

benchmarks/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+#
+# Copyright 2023 Benjamin Worpitz, Jan Stephan
```
SimeonEhrig: Please set your name.

bernhardmgruber:
I generally like the idea of separating benchmarks and examples, but could you please elaborate a bit on your motivation for doing this? Specifically, are you going to add more benchmarks? Are you planning to build the benchmarks differently than examples? Thx!

mehmetyusufoglu:
In my opinion, examples could be designed for any reason, pedagogical or showing implementation of a new feature etc. Benchmarks will mainly focus on performance and visualising it's change through time with CI will show general performance effects of each PR merged.

bernhardmgruber:
> Benchmarks will mainly focus on performance and visualising it's change through time with CI will show general performance effects of each PR merged.

Alright, so you are preparing for some kind of performance CI? Here is a ticket for that: #1264

SimeonEhrig:
> > Benchmarks will mainly focus on performance and visualising it's change through time with CI will show general performance effects of each PR merged.
> 
> Alright, so you are preparing for some kind of performance CI? Here is a ticket for that: #1264

We discussed it last week. At the moment, a CI is not possible because of lacking resources. But we want to have benchmarks to run regression benchmarks locally on laptops, workstation or server.

For example, we thought about to use `mdspan` for tensors in kernels. This makes the usage easier instead using raw pointers. But maybe the performance overhead is to high, which means we need also to implement an interface with raw pointers.

sliwowitz:
There's also https://github.com/alpaka-group/alpaka/pull/1723 which I've just rebased on top of actual develop. It uses Catch2 for benchmarking infrastructure (thus integrated with e.g. ctest). I tried to implement a generaic fixture for benchmarking kernels that would allow us to write simple benchmarks for basic features, but I didn't implement any other use case than the random generator, so I didn't know what would be some actual sensible requirements for such a fixture.

Using Catch2 to handle the benchmarks is IMHO still a good idea since we're already using it to handle tests.

psychocoderHPC (approved):

PR #2238, opened 2024-02-05
Remove CUDA vector array wrapper

sliwowitz:
The `CudaVectorArrayWrapper` class used to provide `operator[]` and some other STL-like features to small CUDA vectors (like `uint4`, `float2`, etc.) It was used in GPU Philox implementations as storage for the counter and key. The CPU implementation used a `std::array` which could not be used in device code until very recent CUDA releases which won't be our lowest supported versions for a long time to come. This posed a portability issue where we couldn't guarantee the same initialization properties in certain edge cases.

The counter and key members were replaced with our `alpaka::Vec` in Philox, making the `CudaVectorArrayWrapper`, which hasn't been used outside Philox, obsolete. Therefore it can be removed, and the Philox implementation simplified since it doesn't need to select the vector data backend on its own, as this responsibility lies with the `alpaka::Vec` class now.

- [x] depends on https://github.com/alpaka-group/alpaka/pull/2236


include/alpaka/math/FloatEqualExact.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2021 Jiri Vyskocil
+/* Cop.hppyright 2021 Jiri Vyskocil
```
SimeonEhrig: typo

SimeonEhrig (dismissed):

bernhardmgruber (dismissed):

sliwowitz:
The removal of `CudaVectorArrayWrapper` also allowed me to remove a couple layers that were doing the data structure backend selection since now it's unified. I also added some unit tests for basic Philox functionality.

SimeonEhrig:
Please rebase. The tsan job is fixed in #2250

psychocoderHPC (approved):

PR #2240, opened 2024-02-19
simplification of workdiv creation

mehmetyusufoglu:
Workdiv constructor needs 2 class template parameters and 3 _alpaka::Vec_ arguments. 
`auto const workdiv = alpaka::WorkDivMembers<Dim, Idx>{arraySize / blockSize, (int)blockSize, 1u};`

The aim is being able to create workdiv, without template parameters, for example 
    `auto const workdiv = createWorkDiv(arraySize / blockSize, blockSize, 1);`

**Update**: Factory method is not used, a new constructor is introduced, no need for explicit type params in the new constructor call. All 3 input types must be alpaka::Vec with the same TDim type.  `auto const workDiv2D = alpaka::WorkDivMembers(blocksPerGrid2D, threadsPerBlock2D, elementsPerThread2D);`

Another PR [2243] is provided onto this one, for using std::array instead of alpaka vectors as arguments. Class template parameters are not needed.

### DIFF
--- a/include/alpaka/workdiv/WorkDivMembers.hpp
+++ b/include/alpaka/workdiv/WorkDivMembers.hpp
@@ -21,6 +21,7 @@ namespace alpaka
     public:
         ALPAKA_FN_HOST_ACC WorkDivMembers() = delete;
 
+        //! Accepts different alpaka vector types and takes the last TDim number of items.
         ALPAKA_NO_HOST_ACC_WARNING
         template<typename TGridBlockExtent, typename TBlockThreadExtent, typename TThreadElemExtent>
         ALPAKA_FN_HOST_ACC explicit WorkDivMembers(
@@ -30,6 +31,18 @@ namespace alpaka
             : m_gridBlockExtent(getExtentVecEnd<TDim>(gridBlockExtent))
             , m_blockThreadExtent(getExtentVecEnd<TDim>(blockThreadExtent))
             , m_threadElemExtent(getExtentVecEnd<TDim>(threadElemExtent))
+        {
+        }
+
+        //! \brief Accepts single specific type and is called without explicit template parameters.
+        ALPAKA_NO_HOST_ACC_WARNING
+        ALPAKA_FN_HOST_ACC WorkDivMembers(
+            alpaka::Vec<TDim, TIdx> const& gridBlockExtent,
+            alpaka::Vec<TDim, TIdx> const& blockThreadExtent,
+            alpaka::Vec<TDim, TIdx> const& elemExtent)
+            : m_gridBlockExtent(gridBlockExtent)
+            , m_blockThreadExtent(blockThreadExtent)
+            , m_threadElemExtent(elemExtent)
         {
         }
 
@@ -82,6 +95,14 @@ namespace alpaka
         Vec<TDim, TIdx> m_blockThreadExtent;
         Vec<TDim, TIdx> m_threadElemExtent;
     };
+
+    //! Deduction guide for the constructor which can be called without explicit template type parameters
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TDim, typename TIdx>
+    ALPAKA_FN_HOST_ACC WorkDivMembers(
+        alpaka::Vec<TDim, TIdx> const& gridBlockExtent,
+        alpaka::Vec<TDim, TIdx> const& blockThreadExtent,
+        alpaka::Vec<TDim, TIdx> const& elemExtent) -> WorkDivMembers<TDim, TIdx>;
 
     namespace trait
     {
--- a/test/unit/workDiv/src/WorkDivHelpersTest.cpp
+++ b/test/unit/workDiv/src/WorkDivHelpersTest.cpp
@@ -139,3 +139,46 @@ TEMPLATE_LIST_TEST_CASE("isValidWorkDiv", "[workDiv]", alpaka::test::TestAccs)
     REQUIRE(alpaka::isValidWorkDiv(alpaka::getAccDevProps<Acc>(dev), workDiv));
     REQUIRE(alpaka::isValidWorkDiv<Acc>(dev, workDiv));
 }
+
+//! Test the constructors of WorkDivMembers using 3D extent, 3D extent with zero elements and 2D extents
+TEST_CASE("WorkDivMembers", "[workDiv]")
+{
+    using Idx = std::size_t;
+    using Dim3D = alpaka::DimInt<3>;
+    using Vec3D = alpaka::Vec<Dim3D, Idx>;
+
+    auto const elementsPerThread3D = Vec3D::all(static_cast<Idx>(1u));
+    auto const threadsPerBlock3D = Vec3D{2u, 2u, 2u};
+    auto blocksPerGrid3D = Vec3D{1u, 1u, 1u};
+
+    auto ref3D = alpaka::WorkDivMembers<Dim3D, Idx>{blocksPerGrid3D, threadsPerBlock3D, elementsPerThread3D};
+    // call WorkDivMembers without explicit class template types
+    auto workDiv3D = alpaka::WorkDivMembers(blocksPerGrid3D, threadsPerBlock3D, elementsPerThread3D);
+    CHECK(workDiv3D == ref3D);
+
+    // change blocks per grid, assign zero to an element
+    blocksPerGrid3D = Vec3D{3u, 3u, 0u};
+    ref3D = alpaka::WorkDivMembers<Dim3D, Idx>{blocksPerGrid3D, threadsPerBlock3D, elementsPerThread3D};
+    // call without explicit template parameter types
+    workDiv3D = alpaka::WorkDivMembers(blocksPerGrid3D, threadsPerBlock3D, elementsPerThread3D);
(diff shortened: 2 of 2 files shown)


SimeonEhrig (dismissed):

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -381,4 +381,17 @@ namespace alpaka
...
+    //! \return Returns the work division; namely the WorkDivMembers instance.
+    template<typename TIdx, typename TDim>
+    auto createWorkDiv(
+        alpaka::Vec<TDim, TIdx> const& gridBlockExtent,
+        alpaka::Vec<TDim, TIdx> const& blockThreadExtent,
+        alpaka::Vec<TDim, TIdx> const& elemExtent) -> alpaka::WorkDivMembers<TDim, TIdx>
+    {
+        return alpaka::WorkDivMembers<TDim, TIdx>(gridBlockExtent, blockThreadExtent, elemExtent);
+    }
+
```
bernhardmgruber: Why can't we just add such a constructor to `WorkDivMembers` and a deduction guide if necessary?

bernhardmgruber:
Just tried it, if you add the constructor:
```c++
        ALPAKA_FN_HOST_ACC WorkDivMembers(
            alpaka::Vec<TDim, TIdx> const& gridBlockExtent,
            alpaka::Vec<TDim, TIdx> const& blockThreadExtent,
            alpaka::Vec<TDim, TIdx> const& elemExtent)
            : m_gridBlockExtent(gridBlockExtent)
            , m_blockThreadExtent(blockThreadExtent)
            , m_threadElemExtent(elemExtent)
        {
        }
```
your unit test can be written using:
```c++
    auto ref = alpaka::WorkDivMembers(blocksPerGrid3D, threadsPerBlock3D, elementsPerThread3D);
```
and compiles and passes the tests.

include/alpaka/workdiv/WorkDivHelpers.hpp
mehmetyusufoglu: Yes, but I wanted to use "create..." since I will add new similar functions. Using create functions seemed cleaner for conversions from stl container types and initializer lists. 

By the way; I think you mean _deduction guides_ for constructors as the one below, since they don't have a return type?:

```
template<class A, class B>
struct Agg
{
    A a;
    B b;
};
// implicitly-generated guides are formed from default, copy, and move constructors
 
template<class A, class B>
Agg(A a, B b) -> Agg<A, B>;
```

mehmetyusufoglu: Currently we cannot use stl containers, argument types are constrained to alpaka vector and buffer.

bernhardmgruber: > Yes, but I wanted to use "create..." since I will add new similar functions. Using create functions seemed cleaner for conversions from stl container types and initializer lists.

I beg to disagree. Constructors are the C++ way to construct an object. There are limited use cases for factory functions, but if a constructor will do fine, use one. You can overload them if you need to convert from various sources.

> By the way; I think you mean _deduction guides_ for constructors as the one below, since they don't have a return type?:
> 
> ```
> template<class A, class B>
> struct Agg
> {
>     A a;
>     B b;
> };
> // implicitly-generated guides are formed from default, copy, and move constructors
>  
> template<class A, class B>
> Agg(A a, B b) -> Agg<A, B>;
> ```

Yes, I mean this kind of deduction guide. But it turns out, that one is not necessary here.

> Currently we cannot use stl containers, argument types are constrained to alpaka vector and buffer.

Alright, you don't need one here :)

mehmetyusufoglu: Ok, I moved the _createWorkDiv_ function as a new constructor. This PR could be merged, then I will prepare a second PR for the _array_ or _initilizer list_ type of inputs without explicit type parameters.

mehmetyusufoglu: It needed a deduction guide (hence a seperate constructor declaration with deduction guide) for some compilers.

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -33,6 +35,18 @@ namespace alpaka
         {
         }
 
+        //! Accepts single specific type and is called without explicit template parameters.
+        ALPAKA_NO_HOST_ACC_WARNING
+        ALPAKA_FN_HOST_ACC explicit WorkDivMembers(
```
bernhardmgruber:
```suggestion
        ALPAKA_FN_HOST_ACC WorkDivMembers(
```

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -21,6 +21,8 @@ namespace alpaka
     public:
         ALPAKA_FN_HOST_ACC WorkDivMembers() = delete;
 
+        //! Accepts different alpaka vector types and takes the last TDim number of items. TDim is explicitly specified
+        //! at the constructor call.
```
bernhardmgruber: You don't specify TDim as part of the constructor call, but as part of the type. I think this suffices:
```suggestion
        //! Accepts different alpaka vector types and takes the last TDim number of items.
```

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -83,6 +97,14 @@ namespace alpaka
...
     };
 
+    //! Deduction guide for the constructor which can be called without explicit template type parameters
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TDim, typename TIdx>
+    ALPAKA_FN_HOST_ACC explicit WorkDivMembers(
+        alpaka::Vec<TDim, TIdx> const& gridBlockExtent,
+        alpaka::Vec<TDim, TIdx> const& blockThreadExtent,
+        alpaka::Vec<TDim, TIdx> const& elemExtent) -> WorkDivMembers<TDim, TIdx>;
+
```
bernhardmgruber:
```suggestion
    template<typename TDim, typename TIdx>
    ALPAKA_FN_HOST_ACC WorkDivMembers(
        alpaka::Vec<TDim, TIdx> const& gridBlockExtent,
        alpaka::Vec<TDim, TIdx> const& blockThreadExtent,
        alpaka::Vec<TDim, TIdx> const& elemExtent) -> WorkDivMembers<TDim, TIdx>;
```
I am not even sure if doxygen supports documenting deduction guides. I would just leave it here.

Different question: I did not need a deduction guide for g++-13. Is the deduction guide provided as a workaround for some compilers? If yes, could you maybe document them here, so we can remove the deduction guide when we phase out those compilers?

mehmetyusufoglu: There were many compilers, yes I can document. Thanks. ( I might return to create... functions idea because now I am trying to call WorkDivMembers constructor without constraining the arguments to alpaka vector.  )

mehmetyusufoglu: The comment is important for the developers

psychocoderHPC (dismissed):

psychocoderHPC:
we need to fix the OSX CI before we can merge this.

test/unit/workDiv/src/WorkDivHelpersTest.cpp
```diff
@@ -139,3 +139,46 @@ TEMPLATE_LIST_TEST_CASE("isValidWorkDiv", "[workDiv]", alpaka::test::TestAccs)
     REQUIRE(alpaka::isValidWorkDiv(alpaka::getAccDevProps<Acc>(dev), workDiv));
     REQUIRE(alpaka::isValidWorkDiv<Acc>(dev, workDiv));
 }
+
+//! Test the constructors of WorkDivMembers using 3D extent, 3D extent with zero elements and 2D extents
+TEMPLATE_LIST_TEST_CASE("WorkDivMembers", "[workDiv]", alpaka::test::TestAccs)
```
bernhardmgruber:
```suggestion
TEST_CASE("WorkDivMembers", "[workDiv]")
```
This test does not depend on any accelerator, so no need to run it multiple times.

bernhardmgruber (changes requested):

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -83,6 +96,14 @@ namespace alpaka
         Vec<TDim, TIdx> m_threadElemExtent;
     };
 
+    //! Deduction guide for the constructor which can be called without explicit template type parameters
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TDim, typename TIdx>
+    ALPAKA_FN_HOST_ACC explicit WorkDivMembers(
```
bernhardmgruber: Why is the deduction guide explicit? I would make it implicit, the same way as the constructor.
```suggestion
    ALPAKA_FN_HOST_ACC WorkDivMembers(
```

psychocoderHPC:
@mehmetyusufoglu could you please rebase this against the develop branch? I merged the OSX CI fix

bernhardmgruber (dismissed):
One more small nit:

bernhardmgruber (approved):

PR #2241, opened 2024-02-20
Change cheatsheet to clarify some alpaka structures

mehmetyusufoglu:
Some simple clarifications for the novice alpaka user are added to the cheatsheet and library file. Alpaka vector is a static array, Buffer is a dynamic array. Pitch is the allocated memory with padding (till the alignment point) and extent is actually dimensions.

### DIFF
--- a/docs/source/basic/cheatsheet.rst
+++ b/docs/source/basic/cheatsheet.rst
@@ -115,9 +115,10 @@ Create a CPU device for memory allocation on the host side
 
 Allocate a buffer in host memory
   .. code-block:: c++
-
-     Vec<Dim, Idx> extent = value;
-     using BufHost = Buf<DevHost, DataType, Dim, Idx>;
+     // Use alpaka vector as a static array for the extents
+     alpaka::Vec<Dim, Idx> extent = value;
+     // Allocate memory for the alpaka buffer, which is a dynamic array
+     using BufHost = alpaka::Buf<DevHost, DataType, Dim, Idx>;
      BufHost bufHost = allocBuf<DataType, Idx>(devHost, extent);
 
 (Optional, affects CPU – GPU memory copies) Prepare it for asynchronous memory copies
@@ -129,7 +130,8 @@ Create a view to host memory represented by a pointer
   .. code-block:: c++
 
      using Dim = alpaka::DimInt<1u>;
-     Vec<Dim, Idx> extent = size;
+     // Create an alpaka vector which is a static array
+     alpaka::Vec<Dim, Idx> extent = size;
      DataType* ptr = ...;
      auto hostView = createView(devHost, ptr, extent);
 
@@ -231,7 +233,7 @@ Access multi-dimensional indices and extents of blocks, threads, and elements
      // Origin: Grid, Block, Thread
      // Unit: Blocks, Threads, Elems
 
-Access components of and destructuremulti-dimensional indices and extents
+Access components of and destructure multi-dimensional indices and extents
   .. code-block:: c++
 
      auto idxX = idx[0];
--- a/docs/source/basic/library.rst
+++ b/docs/source/basic/library.rst
@@ -211,7 +211,8 @@ Memory Management
 
 The memory allocation function of the *alpaka* library (``alpaka::allocBuf<TElem>(device, extents)``) is uniform for all devices, even for the host device.
 It does not return raw pointers but reference counted memory buffer objects that remove the necessity for manual freeing and the possibility of memory leaks.
-Additionally the memory buffer objects know their extents, their pitches as well as the device they reside on.
+Additionally, the memory buffer objects know their extents, their pitches as well as the device they reside on.
+Due to padding, the allocated number of bytes may be more than the required storage; the pitch value gives the correct stride for each dimension for row-major access.
 This allows buffers that possibly reside on different devices with different pitches to be copied only by providing the buffer objects as well as the extents of the region to copy (``alpaka::memcpy(bufDevA, bufDevB, copyExtents``).
 
 Kernel Execution


docs/source/basic/cheatsheet.rst
```diff
@@ -231,7 +233,7 @@ Access multi-dimensional indices and extents of blocks, threads, and elements
      // Origin: Grid, Block, Thread
      // Unit: Blocks, Threads, Elems
 
-Access components of and destructuremulti-dimensional indices and extents
+Access components of and destructuremulti-dimensional indices and extents (dimensions)
```
bernhardmgruber:
```suggestion
Access components of and destructure multi-dimensional indices and extents (dimensions)
```

docs/source/basic/library.rst
```diff
@@ -211,7 +211,7 @@ Memory Management
 
 The memory allocation function of the *alpaka* library (``alpaka::allocBuf<TElem>(device, extents)``) is uniform for all devices, even for the host device.
 It does not return raw pointers but reference counted memory buffer objects that remove the necessity for manual freeing and the possibility of memory leaks.
-Additionally the memory buffer objects know their extents, their pitches as well as the device they reside on.
+Additionally the memory buffer objects know their extents, their pitches (space allocated in including the padding) as well as the device they reside on.
```
bernhardmgruber: This is not correct. The pitches are strides. They are the number of elements to skip to reach the next element/row/slice, etc.
The mental model for accessing a multidimensional buffer/array is: `*(ptr + dot_product(index, pitches))`.

docs/source/basic/cheatsheet.rst
```diff
@@ -115,9 +115,10 @@ Create a CPU device for memory allocation on the host side
 
 Allocate a buffer in host memory
   .. code-block:: c++
-
-     Vec<Dim, Idx> extent = value;
-     using BufHost = Buf<DevHost, DataType, Dim, Idx>;
+     // Use alpaka vector as a static array for the extent(dimensions)
+     alpaka::Vec<Dim, Idx> extent = value;
+     // Allocate memory for the alpaka buffer which is a dynamic array
+     using BufHost = alpaka::Buf<DevHost, DataType, Dim, Idx>;
```
bernhardmgruber: Be careful to keep the cheatsheet short! It is rather not an introduction, but really a cheatsheet for the API. At least that's my understanding.

docs/source/basic/cheatsheet.rst
```diff
@@ -115,9 +115,10 @@ Create a CPU device for memory allocation on the host side
 
 Allocate a buffer in host memory
   .. code-block:: c++
-
-     Vec<Dim, Idx> extent = value;
-     using BufHost = Buf<DevHost, DataType, Dim, Idx>;
+     // Use alpaka vector as a static array for the extent(dimensions)
```
bernhardmgruber:
```suggestion
     // Use alpaka vector as a static array for the extent(dimensions)
```
I prefer the word extents over dimensions. In my mental model, if you have a 2D array of size 10x12 elements, it has 2 dimensions and extents 10 and 12. So dimensions for me, are the number of extents.

docs/source/basic/library.rst
mehmetyusufoglu: Yes, with data+padding it makes a stride. I should have written explicitly.

docs/source/basic/cheatsheet.rst
mehmetyusufoglu: No. When we say "dimensions of an A4 paper" or "dimensions of a box" we don't mean "2 dimensional" or "3 dimentional" we just mean height,width and depth values.

fwyzard: Well, an A4 paper has 2 dimensions, and those dimensions are 29.7 cm and 21 cm.

To be pedantic, one could say that an A4 paper has a dimensionality of 2, and its size along those dimensions are 29.7 cm and 21 cm.

In any case, I don't know if there is an official nomenclature in C/C++.
The STL has traits for them, and uses [rank](https://en.cppreference.com/w/cpp/types/rank) for the number of dimensions, and [extent](https://en.cppreference.com/w/cpp/types/extent) the size along each dimension.

docs/source/basic/library.rst
mehmetyusufoglu: "due to padding, allocated number of bytes may be more than the data, the pitch value gives the correct stride for row-major access"

docs/source/basic/cheatsheet.rst
mehmetyusufoglu: _Extent_ and _pitch_ are Cuda concepts mainly, C++ developers know "dimensions", "padding", "alignment".  Therefore I have just wanted to add a few words to clarify.

docs/source/basic/cheatsheet.rst
mehmetyusufoglu: Sure, these are really important explanations. I am not planning to add more.

docs/source/basic/cheatsheet.rst
mehmetyusufoglu: Ok I will do "array of dimensions" :)

bernhardmgruber: In addition to what @fwyzard said, there is also [`std::extents`](https://en.cppreference.com/w/cpp/container/mdspan/extents).

fwyzard: since C++23 ... I still live in C++17 :-)

mehmetyusufoglu:
1. I think you mix up "number of dimensions" and the "dimensions".

2. I haven't replaced or proposed to replace the term extent with the
dimensions, just tried to make the document more understandable for the
people who are new to alpaka.

20 Şub 2024 Sal 18:07 tarihinde Andrea Bocci ***@***.***>
şunu yazdı:

docs/source/basic/cheatsheet.rst
mehmetyusufoglu: 1. I think you mix up "number of dimensions" and the "dimensions". 

2. I haven't replaced or proposed replacing the term extent with the
dimensions, just tried to make the document more understandable for the
people who are new to alpaka.

bernhardmgruber: While this mostly counts for me as well, as library designers or architects, we need to stay current with the technical evolution and design accordingly. I also want my libraries to fit the future ;)

mehmetyusufoglu: Alpaka uses of Dim and TDim, which are misleading I think. For example Dim can be NDim or NumDim ( e.g. matlab or python function names to get number of dimensions)

docs/source/basic/cheatsheet.rst
```diff
@@ -129,7 +130,8 @@ Create a view to host memory represented by a pointer
   .. code-block:: c++
 
      using Dim = alpaka::DimInt<1u>;
-     Vec<Dim, Idx> extent = size;
+     // create an alpaka vector which is a static array
```
bernhardmgruber: Please use a consistent case when starting a comment
```suggestion
     // Create an alpaka vector which is a static array
```

docs/source/basic/cheatsheet.rst
```diff
@@ -115,9 +115,10 @@ Create a CPU device for memory allocation on the host side
 
 Allocate a buffer in host memory
   .. code-block:: c++
-
-     Vec<Dim, Idx> extent = value;
-     using BufHost = Buf<DevHost, DataType, Dim, Idx>;
+     // Use alpaka vector as a static array for the extent(the array of dimensions)
```
bernhardmgruber:
```suggestion
     // Use alpaka vector as a static array for the extents (the array of dimensions)
```

docs/source/basic/cheatsheet.rst
```diff
@@ -115,9 +115,10 @@ Create a CPU device for memory allocation on the host side
...
+     alpaka::Vec<Dim, Idx> extent = value;
+     // Allocate memory for the alpaka buffer which is a dynamic array
```
bernhardmgruber:
```suggestion
     // Allocate memory for the alpaka buffer, which is a dynamic array
```

docs/source/basic/library.rst
```diff
@@ -211,7 +211,7 @@ Memory Management
 
 The memory allocation function of the *alpaka* library (``alpaka::allocBuf<TElem>(device, extents)``) is uniform for all devices, even for the host device.
 It does not return raw pointers but reference counted memory buffer objects that remove the necessity for manual freeing and the possibility of memory leaks.
-Additionally the memory buffer objects know their extents, their pitches as well as the device they reside on.
+Additionally the memory buffer objects know their extents, their pitches (due to padding, allocated number of bytes may be more than the data; the pitch value gives the correct stride for row-major access) as well as the device they reside on.
```
bernhardmgruber: I would reword into two sentences:
```suggestion
Additionally, the memory buffer objects know their extents, their pitches as well as the device they reside on.
Due to padding, the allocated number of bytes may be more than the required storage; the pitch value gives the correct stride for each dimension for row-major access.
```

bernhardmgruber (commented):
Small improvements, but generally LGTM

docs/source/basic/cheatsheet.rst
mehmetyusufoglu: Lets agree on this. Number of Dimensions = Number of Extents. Dimensions could be misleading as you said. I accept. Why, because it could mean a set of dimension names like {XDimension,YDimension,ZDimension} or it could mean as in daily usage {width,depth, height}.  I remove it.

fwyzard: > Number of Dimensions = Number of Extents.

Agreed.
It is also called "rank" (in C++) or "dimensionality".

bernhardmgruber (approved):

PR #2242, opened 2024-02-22
Add support for device and constant global variables in the SYCL backend

### LINKED ISSUE #2070: device global variables in SYCL
That is a tough topic.. first a quick look into the usage of device global variables in SYCL (since [oneAPI 2023.2](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_device_global.asciidoc)):
```cpp
// declaration of the variable
sycl::ext::oneapi::experimental::device_global<int> myGlobVar;

int main(){
  // ...
  // there are dedicated memset and memcpy to/from global variables
  Q.memcpy(myGlobVar, &hostVar).wait();

  Q.submit([&](sycl::handler &cgh) {
     cgh.parallel_for(sycl::nd_range<1>( 1, 1), [=](sycl::nd_item<1> item) {
       // myGlobVar is of type device_global, it can be used directly:
       int b = myGlobVar + item.get_global_linear_id();
       // or with .get() to obtain the actual type (myGlobVar.get() is of type int)
       printf("value of myGlobVar: %d\n", myGlobVar.get());
    });
  });
```
None of that is easy to integrate into alpaka and in any case will break the current implementation.

To declare the variable a possibility is to change the current macro to take the type as arguments, like this:
```cpp
#define DEVICE_GLOBAL(T) sycl::ext::oneapi::experimental::device_global<T> 
```
but then it would need an explicit `.get()` inside the kernels.

The following example already takes care of the `.get()`, but it does not work, because trying to use `V` inside a kernel would result in a compile time error:
 ```cpp
#ifdef __SYCL_DEVICE_ONLY__
  #define DEVICE_GLOBAL(T,V)                                            \  
    sycl::ext::oneapi::experimental::device_global<T> V##_proxy;        \
    T& V = V##_proxy.get();
#endif
```

For `memcpy` / `memset`, dedicated methods should be added, with the additional issue that the current implementation checks that the types of dst and src are the same and this is not true when device global variables specific methods are used.
This is one of the implementations of the `memcpy` in SYCL:
```cpp
// memcpy to device_global
template <typename T, typename PropertyListT>
event memcpy(device_global<T, PropertyListT>& dest,
   const void *src, size_t numBytes = sizeof(T), size_t offset = 0);
```

To conclude, it's not clear yet what is the best way to do the `.get()` inside kernels.

Any suggestion on how to implement all of this is appreciated :)

Note: in alpaka there are also constant device global variables (`__constant__` in CUDA). In SYCL the constant memory has been deprecated, but this can be achieved by creating a `const T` device global:
```cpp
sycl::ext::oneapi::experimental::device_global<const int> myGlobVar;
```

FYI @fwyzard

AuroraPerego:
This tries to fix #2070.
This implementation changes the current API.
The declaration of device memory is done with the macro `ALPAKA_STATIC_ACC_MEM_GLOBAL(type, name)` that wraps the CUDA/HIP/serial variables in a struct `alpaka::DevGlobal` and declares a `sycl::ext::oneapi::experimental::device_global<type>` for SYCL. The `inline` attribute is used to ensure that only one instance of that variable exists across different translation units.
`name` is used for the `memcpy`, while `name.get()` must be used in the kernel to align with the behavior of the SYCL backend.
The `memcpy` has been specialized for the device global variables.

The test with the SYCL backend failed with the original `KernelExecutionFixture` because it creates a new `queue` instead of using the one used for the `memcpy`. I added a constructor that takes in input also the `queue`.

Another issue with the test is that being compiled with the flags to enable the SYCL backend, the macro for the device global variable expands to the SYCL one (`sycl::ext::oneapi::experimental::device_global<type>`) and therefore it fails when running on the `AccCpuSerial`. I have disabled this accelerator just for this test.
Note that this happens also with the CUDA backend (with the macro expanded to `__device__ alpaka::DevGlobal<type> name`), but for some strange reason it works on the serial backend.

Thanks to @fwyzard for the help :)


include/alpaka/core/Common.hpp
```diff
@@ -98,12 +98,11 @@
...
-//!   ALPAKA_STATIC_ACC_MEM_GLOBAL int i;
+//!   ALPAKA_STATIC_ACC_MEM_GLOBAL(int, variable);
 //!
-//! Those variables behave like ordinary variables when used in file-scope.
-//! They have external linkage (are accessible from other compilation units).
-//! If you want to access it from a different compilation unit, you have to declare it as extern:
-//!   extern ALPAKA_STATIC_ACC_MEM_GLOBAL int i;
+//! Those variables behave like ordinary variables when used in file-scope,
+//! but inside kernels the get() method must be used to access the variable.
+//! They are declared inline to be accessible from other compilation units.
```
fwyzard:
```suggestion
//! They are declared inline to resolve to a single instance across multiple
//! translation units.
```

include/alpaka/core/Common.hpp
```diff
@@ -112,36 +111,71 @@
...
 //! \attention It is not allowed to initialize the variable together with the declaration.
-//!            To initialize the variable alpaka::createStaticDevMemView and alpaka::memcpy must be used.
+//!            To initialize the variable alpaka::memcpy must be used.
 //! \code{.cpp}
-//! ALPAKA_STATIC_ACC_MEM_GLOBAL int foo;
+//! ALPAKA_STATIC_ACC_MEM_GLOBAL(int, foo);
+//!
+//! struct DeviceMemoryKernel
+//! {
+//!    ALPAKA_NO_HOST_ACC_WARNING
```
fwyzard, line 121: why `ALPAKA_NO_HOST_ACC_WARNING` ?

include/alpaka/core/Common.hpp
```diff
@@ -112,36 +111,71 @@
...
-//!   ALPAKA_STATIC_ACC_MEM_CONSTANT int i;
+//!   ALPAKA_STATIC_ACC_MEM_CONSTANT(int, variable);
 //!
-//! Those variables behave like ordinary variables when used in file-scope.
-//! They have external linkage (are accessible from other compilation units).
-//! If you want to access it from a different compilation unit, you have to declare it as extern:
-//!   extern ALPAKA_STATIC_ACC_MEM_CONSTANT int i;
+//! Those variables behave like ordinary variables when used in file-scope,
+//! but inside kernels the get() method must be used to access the variable.
+//! They are declared inline to be accessible from other compilation units.
```
fwyzard:
```suggestion
//! They are declared inline to resolve to a single instance across multiple
//! translation units.
```

include/alpaka/core/Common.hpp
```diff
@@ -150,25 +184,40 @@
...
 //! \attention It is not allowed to initialize the variable together with the declaration.
-//!            To initialize the variable alpaka::createStaticDevMemView and alpaka::memcpy must be used.
+//!            To initialize the variable alpaka::memcpy must be used.
 //! \code{.cpp}
-//! ALPAKA_STATIC_ACC_MEM_CONSTANT int foo;
+//! ALPAKA_STATIC_ACC_MEM_CONSTANT(int, foo);
+//!
+//! struct DeviceMemoryKernel
+//! {
+//!    ALPAKA_NO_HOST_ACC_WARNING
```
fwyzard, line 170: why `ALPAKA_NO_HOST_ACC_WARNING` ?

include/alpaka/core/Common.hpp
```diff
@@ -150,25 +184,40 @@
...
-//!     alpaka::memcpy(queue, viewGlobalMemUninitialized, bufHost, extent);
+//!     alpaka::memcpy(queue, foo, bufHost, extent);
 //! }
 //! \endcode
 #if((BOOST_LANG_CUDA && BOOST_COMP_CLANG_CUDA) || (BOOST_LANG_CUDA && BOOST_COMP_NVCC && BOOST_ARCH_PTX)              \
     || BOOST_LANG_HIP)
-#    define ALPAKA_STATIC_ACC_MEM_CONSTANT __constant__
+#    define ALPAKA_STATIC_ACC_MEM_CONSTANT(type, name)                                                                \
+        template<typename TAcc>                                                                                       \
+        inline __constant__ alpaka::DevGlobal<type> name
```
fwyzard: would
```suggestion
        inline __constant__ alpaka::DevGlobal<const type> name
```
work ?

(also for the CPU case)

include/alpaka/mem/global/DeviceGlobalCpu.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <type_traits>
+
+// memcpy specialization for device global variables
+namespace alpaka
+{
+    template<typename TViewSrc, typename TViewDstFwd, typename TQueue>
+    ALPAKA_FN_HOST auto memcpy(TQueue& queue, alpaka::DevGlobal<TViewDstFwd>& viewDst, TViewSrc const& viewSrc) -> void
+    {
+        typedef std::remove_all_extents_t<TViewDstFwd> T;
```
fwyzard: I would prefer to use
```suggestion
        using T = std::remove_all_extents_t<TViewDstFwd>;
```
everywhere, instead of `typedef`.

And actually `Type` would make more sense than `T`, which is normally used for template arguments.

include/alpaka/mem/global/DeviceGlobalCpu.hpp
```diff
@@ -0,0 +1,68 @@
...
+        auto extent = getExtents(viewSrc);
+        auto view = alpaka::ViewPlainPtr<DevCpu, T, alpaka::Dim<decltype(extent)>, alpaka::Idx<decltype(extent)>>(
+            reinterpret_cast<T*>(&viewDst),
```
fwyzard: IIUC, this relies on the fact that `DevGlobal<Type>` and `DevGlobal<Type>::value` are at the same address in memory.
If that is the case, please add a comment to explain it.

include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2023 Benjamin Worpitz, Andrea Bocci, Bernhard Manfred Gruber, Jan Stephan
+/* Copyright 2023 Benjamin Worpitz, Andrea Bocci, Bernhard Manfred Gruber, Jan Stephan, Aurora Perego
```
fwyzard:
```suggestion
/* Copyright 2024 Benjamin Worpitz, Andrea Bocci, Bernhard Manfred Gruber, Jan Stephan, Aurora Perego
```

include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -34,18 +34,35 @@ namespace alpaka::test
...
+            , m_workDiv{getValidWorkDiv<Acc>(
+                  m_device,
+                  extent,
+                  Vec<Dim, Idx>::ones(),
+                  false,
+                  GridBlockExtentSubDivRestrictions::Unrestricted)}
+        {
+        }
+
+        KernelExecutionFixture(Queue& queue, WorkDiv workDiv) : m_queue{queue}, m_workDiv{std::move(workDiv)}
```
fwyzard:
```suggestion
        KernelExecutionFixture(Queue queue, WorkDiv workDiv) : m_queue{std::move(queue)}, m_workDiv{std::move(workDiv)}
```
to support the case where the `queue` argument is a temporary and cannot bind to a non-const reference.

include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -34,18 +34,35 @@ namespace alpaka::test
...
+        {
+        }
+
+        template<typename TExtent>
+        KernelExecutionFixture(Queue& queue, TExtent const& extent)
+            : m_queue{queue}
```
fwyzard:
```suggestion
        KernelExecutionFixture(Queue queue, TExtent const& extent)
            : m_queue{std::move(queue)}
```
Idem.

test/unit/mem/view/src/ViewStaticAccMem.cpp
```diff
@@ -1,5 +1,5 @@
-/* Copyright 2023 Axel Huebl, Benjamin Worpitz, Matthias Werner, Bernhard Manfred Gruber, Jan Stephan, Andrea Bocci
- * SPDX-License-Identifier: MPL-2.0
+/* Copyright 2023 Axel Huebl, Benjamin Worpitz, Matthias Werner, Bernhard Manfred Gruber, Jan Stephan, Andrea Bocci,
```
fwyzard:
```suggestion
/* Copyright 2024 Axel Huebl, Benjamin Worpitz, Matthias Werner, Bernhard Manfred Gruber, Jan Stephan, Andrea Bocci,
```

include/alpaka/core/Common.hpp
```diff
@@ -1,4 +1,4 @@
-/* Copyright 2023 Axel Hübl, Benjamin Worpitz, Matthias Werner, Jan Stephan, René Widera, Andrea Bocci
+/* Copyright 2023 Axel Hübl, Benjamin Worpitz, Matthias Werner, Jan Stephan, René Widera, Andrea Bocci, Aurora Perego
```
fwyzard:
```suggestion
/* Copyright 2024 Axel Hübl, Benjamin Worpitz, Matthias Werner, Jan Stephan, René Widera, Andrea Bocci, Aurora Perego
```

fwyzard (commented):
Many thanks for this work !

As it introduces a breaking changes to the API, I would ask for enough time to consider all impacts before merging the changes.

include/alpaka/core/Common.hpp
AuroraPerego: copied from the test (https://github.com/alpaka-group/alpaka/blob/develop/test/unit/mem/view/src/ViewStaticAccMem.cpp#L31)

include/alpaka/core/Common.hpp
AuroraPerego: same as above

include/alpaka/core/Common.hpp
AuroraPerego: From a quick try, I'd say no because `T value` inside the `struct DevGlobal` is mapped to `const type value` that then needs to be initialized when the variable is declared

include/alpaka/mem/global/DeviceGlobalCpu.hpp
AuroraPerego: (If I did it correctly) there is the overload of `operator&` that returns the address of `value`

include/alpaka/core/Common.hpp
fwyzard: Maybe you could store `std::remove_const_t<T>` as a data member (so `memcpy` can write to it), but return a `T&` or `T*` (including the `const`) in `T& get()` and `T* operator&()` ?

include/alpaka/mem/global/DeviceGlobalCpu.hpp
fwyzard: Mhm, ok. Why is the `reinterpret_cast<T*>` needed, then ?

include/alpaka/mem/view/Traits.hpp
```diff
@@ -1,4 +1,5 @@
-/* Copyright 2023 Axel Hübl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2023 Axel Hübl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan, Bernhard Manfred Gruber,
+ * Aurora Perego
```
fwyzard:
```suggestion
/* Copyright 2024 Axel Hübl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan, Bernhard Manfred Gruber,
 *                Aurora Perego
```

include/alpaka/core/Common.hpp
AuroraPerego: do you mean
```cpp
    template<typename T>
    struct DevGlobal
    {
        using Type = std::remove_const_t<T>;
        Type value; // backend specific value
    }
```
?
The `memcpy` sees `T`, not `Type`, and fails :(

fwyzard: Some options:
  - add a `private` non-`const` `get()` method, and make the function that calls `memcpy` a `friend` of `DevGlobal` so it can use it
  - use `const_cast<std::remove_const_t<T>*>()` in the call to `memcpy` (this should be safe because the underlying `value` is not `const`)
  - use two different objects, one with a `const` `get()` on the device and one with a non-`const` `get()` on the host (this is just a random idea, not sure it's actually feasible)

include/alpaka/mem/global/DeviceGlobalCpu.hpp
AuroraPerego: because `&viewDst` is a pointer to `TViewDstFwd` that can have an extent, while `T` is without the extent.
In the test for example `TViewDstFwd=unsigned int[3][2]`, `&viewDst` has type `unsigned int (*)[3][2]`, while `T=unsigned int`

include/alpaka/core/Common.hpp
AuroraPerego: I went for the second option

include/alpaka/mem/global/DeviceGlobalGenericSycl.hpp
```diff
@@ -0,0 +1,97 @@
...
+#include "alpaka/mem/global/Traits.hpp"
+#include "alpaka/queue/sycl/QueueGenericSyclBase.hpp"
+
+#ifdef ALPAKA_ACC_SYCL_ENABLED
+
+#    include <sycl/sycl.hpp>
+
+namespace alpaka
+{
+    using sycl::ext::oneapi::experimental::device_global;
```
fwyzard: This makes `device_global` visible to all in the `alpaka` namespace, which may not be a good idea.

Could you move it at least inside the `detail` namespace (and use it as `detail::device_global` below) ?

Or, just use it fully expanded everywhere.

include/alpaka/mem/global/DeviceGlobalUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,120 @@
...
+        struct DevGlobalTrait<TagGpuCudaRt, T>
+        {
+            // CUDA implementation
+            using Type = detail::DevGlobalImplGeneric<TagGpuCudaRt, T>;
+        };
+
+        template<typename T>
+        struct DevGlobalTrait<TagGpuHipRt, T>
+        {
+            // CUDA implementation
```
fwyzard:
```suggestion
            // HIP/ROCm implementation
```

include/alpaka/mem/global/DeviceGlobalCpu.hpp
```diff
@@ -0,0 +1,109 @@
...
+        {
+            using Type = detail::DevGlobalImplGeneric<TagCpuThreads, T>;
+        };
+    } // namespace detail
+
+    template<typename TAcc, typename TViewSrc, typename TViewDstFwd, typename TQueue>
+    ALPAKA_FN_HOST auto memcpy(
+        TQueue& queue,
+        alpaka::detail::DevGlobalImplGeneric<TAcc, TViewDstFwd>& viewDst,
+        TViewSrc const& viewSrc) -> void
```
fwyzard: This approach seems too generic, as in principle it can match also for a CUDA `DevGlobalImplGeneric`.

Also, `TAcc` should be `TTag`.

include/alpaka/mem/global/DeviceGlobalUniformCudaHipBuiltIn.hpp
```diff
@@ -0,0 +1,120 @@
...
+            using Type = detail::DevGlobalImplGeneric<TagGpuHipRt, T>;
+        };
+    } // namespace detail
+
+    // from device to host
+    template<typename TAcc, typename TApi, bool TBlocking, typename TViewDst, typename TViewSrc>
+    ALPAKA_FN_HOST auto memcpy(
+        uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, TBlocking>& queue,
+        TViewDst& viewDst,
+        alpaka::detail::DevGlobalImplGeneric<TAcc, TViewSrc>& viewSrc)
```
fwyzard: `TAcc` should be `TTag`, it should somehoe match the `TApi` (we don't want to use CUDA on a HIP/ROCm global variable).

include/alpaka/mem/global/DeviceGlobalCpu.hpp
fwyzard: Can you add a constrain that `TAcc` (actually `TTag`) is one of `TagCpuOmp2Blocks`, `TagCpuOmp2Threads`, `TagCpuSerial`, `TagCpuTbbBlocks`, `TagCpuThreads` ?

There may be a smarter way, but at least this should avoid the wrong matches.

fwyzard: By the way, IIUC `TViewDstFwd` here is not really a `View`, but the underlying type of the global variable.
If that's the case, could you rename it to `TType` ?

include/alpaka/mem/global/DeviceGlobalGenericSycl.hpp
AuroraPerego: I've used it fully expanded

include/alpaka/mem/global/DeviceGlobalCpu.hpp
AuroraPerego: > Can you add a constrain that `TAcc` (actually `TTag`) is one of `TagCpuOmp2Blocks`, `TagCpuOmp2Threads`, `TagCpuSerial`, `TagCpuTbbBlocks`, `TagCpuThreads` ?

ok, I'll try (and I'll probably also have to prevent the test from running on `AccCpuSerial` when it is compiled with the flags to enable CUDA/HIP/SYCL)

> By the way, IIUC `TViewDstFwd` here is not really a `View`, but the underlying type of the global variable. If that's the case, could you rename it to `TType` ?

You are right, I'll change it

include/alpaka/mem/global/DeviceGlobalUniformCudaHipBuiltIn.hpp
AuroraPerego: I don't know if there is a map between `TTag` and `TApi`, but I'll look into it

AuroraPerego:
Device global variables in oneAPI versions `2023.*` require explicitly the device image scope in the constructor, while from `2024.0` this is not needed anymore. For this reason, the required compiler version has been changed to `2024.0`.

Now the only failures in the tests are some warnings (that become errors when compiling the tests) in the SYCL `headers` / `footers`.

fwyzard:
> Now the only failures in the tests are some warnings (that become errors when compiling the tests) in the SYCL `headers` / `footers`.

@psychocoderHPC can we remove `-Wreserved-identifier -Wold-style-cast` from the flags passed to `icpx` ?
The offending code is autogenerated, and at least for `-Wreserved-identifier` is included before any user or alpaka code, so we cannot disable the warnings with a `#pragma` in the code.

psychocoderHPC:
> > Now the only failures in the tests are some warnings (that become errors when compiling the tests) in the SYCL `headers` / `footers`.
> 
> @psychocoderHPC can we remove `-Wreserved-identifier -Wold-style-cast` from the flags passed to `icpx` ? The offending code is autogenerated, and at least for `-Wreserved-identifier` is included before any user or alpaka code, so we cannot disable the warnings with a `#pragma` in the code.

offline discussed: yes we can remove the options if nessesary

test/unit/mem/view/src/ViewStaticAccMem.cpp
```diff
@@ -16,92 +16,119 @@ using Elem = std::uint32_t;
...
-
-    WARN("The SYCL backend does not support global device variables.");
-
-#endif // !defined(ALPAKA_ACC_SYCL_ENABLED)
+        std::vector<Elem> data2(6, 0u);
+        auto bufHost2 = alpaka::createView(devHost, data2.data(), extent);
+        alpaka::memcpy(
+            queueAcc,
+            bufHost2,
+            g_globalMemory2DUninitialized<typename alpaka::trait::AccToTag<Acc>::type>,
```
SimeonEhrig: I would prefer passing the `Acc` type directly instead using the type trait `alpaka::trait::AccToTag`. I think the common case is, that we have the `Acc` type available and not the tag type, therefore we would have a lot of copy/past code.

Depending on if it makes sense to provide the possibility to use tags in the user API, you can ether move `typename alpaka::trait::AccToTag<Acc>::type` to the implementation (no support for tags in the user API) or do some kind of overload and check if the given type is a alpaka accelerator type or tag.

AuroraPerego: I've moved it to the implementation
```cpp
    using DevGlobal = typename detail::DevGlobalTrait<typename alpaka::trait::AccToTag<TAcc>::type, T>::Type;
```

SimeonEhrig (approved):

psychocoderHPC (approved):

psychocoderHPC:
@alpaka-group/alpaka-maintainers IMO we can merge this PR. Are there any voices against merging it?

fwyzard:
OK for me.

Any further developments can happen in follow up PRs.

BenjaminW3 (approved):

test/unit/mem/view/src/ViewStaticAccMem.cpp
```diff
@@ -16,92 +16,101 @@ using Elem = std::uint32_t;
 using Dim = alpaka::DimInt<2u>;
 using Idx = std::uint32_t;
 
-#if !defined(ALPAKA_ACC_SYCL_ENABLED)
+ALPAKA_STATIC_ACC_MEM_GLOBAL alpaka::DevGlobal<TAcc, Elem[3][2]> g_globalMemory2DUninitialized;
```
psychocoderHPC, line 19: @AuroraPerego We need to review this change again because of the bug shown https://github.com/alpaka-group/alpaka/issues/2259

In the original code we had constant global memory only and in the new code we have global memory and constant memory.
With your changes is there a difference between `ALPAKA_STATIC_ACC_MEM_GLOBAL` and `ALPAKA_STATIC_ACC_MEM_CONSTANT`?

psychocoderHPC: IMO in this section we should only use `ALPAKA_STATIC_ACC_MEM_CONSTANT` and later we have a test for `ALPAKA_STATIC_ACC_MEM_GLOBAL`

AuroraPerego: > In the original code we had constant global memory only and in the new code we have global memory and constant memory.

They were different also in the original code, with `ALPAKA_STATIC_ACC_MEM_GLOBAL` expanding to `__device__` and `ALPAKA_STATIC_ACC_MEM_CONSTANT` expanding to `__constant__`.

> With your changes is there a difference between ALPAKA_STATIC_ACC_MEM_GLOBAL and ALPAKA_STATIC_ACC_MEM_CONSTANT?

Yes, they are using the `__device__` and `__constant__` attributes respectively (plus the `inline` attribute)

> IMO in this section we should only use `ALPAKA_STATIC_ACC_MEM_CONSTANT` and later we have a test for `ALPAKA_STATIC_ACC_MEM_GLOBAL`

ok, but why? we were testing both here also before this PR

psychocoderHPC: > ok, but why? we were testing both here also before this PR

This is a good question.

What if we define `ALPAKA_STATIC_ACC_MEM_GLOBAL` static instead of `inline` this should create a single instance too.

AuroraPerego: it compiles and also solves the issue in #2259, I can make a PR with the change if you think it's correct (I don't know enough about `static`, `inline` and `extern` to be sure of that)

fwyzard: In general, what macros should be used to support all use cases ?

### `file1.cu`
```c++
static __device__ int i = 0;
```

and

### `file2.h`
```c++
extern __device__ int i;
```

### `file2.cu`
```c++
#include "file2.h"

__device__ int i = 0;
```

?

AuroraPerego: Thinking about it, maybe `static` is not what we want. The ideal would be to make the `extern` keyword work also with the new implementation to allow the `file2.h`/`file2.cu` case, which IIUC is not possible with `static`.

PR #2243, opened 2024-02-26
 workDivMembers construction with a user defined type

mehmetyusufoglu:
`WorkDivMembers` class constructor could accept any vector type if the necessary traits and functions related to that type are provided.

This PR uses a user-defined vector, related traits and specializations to test the  of `WorkDivMembers` constructors. 

- Tests are added to test the usage of the user-defined vectors in `WorkDivMembers` construction.
-  Already existing `WorkDivMembers` construction tests are improved and refactored. 
-  Initializer list test is added in construction of `WorkDivMembers`.  `auto const workDiv2DUsingInitList = alpaka::WorkDivMembers<Dim2D, Idx>({6, 9}, {4, 6}, {1, 1});` initializes 3 alpaka vectors as arguments.

### DIFF
--- a/test/unit/workDiv/CMakeLists.txt
+++ b/test/unit/workDiv/CMakeLists.txt
@@ -10,6 +10,11 @@ append_recursive_files_add_to_src_group("src/" "src/" "cpp" _FILES_SOURCE)
 alpaka_add_executable(
     ${_TARGET_NAME}
     ${_FILES_SOURCE})
+
+target_include_directories(
+ ${_TARGET_NAME}
+ PRIVATE "src")
+
 target_link_libraries(
     ${_TARGET_NAME}
     PRIVATE common)
--- a/test/unit/workDiv/src/FooVec.hpp
+++ b/test/unit/workDiv/src/FooVec.hpp
@@ -1,0 +1,112 @@
+/* Copyright 2022 Sergei Bastrakov, Jan Stephan, Mehmet Yusufoglu
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include <alpaka/dim/Traits.hpp>
+#include <alpaka/elem/Traits.hpp>
+#include <alpaka/extent/Traits.hpp>
+#include <alpaka/vec/Vec.hpp>
+
+#include <array>
+#include <iostream>
+
+//! User defined vector for testing the usability of any vector type.
+//!
+//! \tparam TVal The data type.
+//! \tparam N Vector size as a non-type parameter.
+template<typename TVal, std::size_t N>
+class FooVec
+{
+public:
+    static_assert(N <= 3, "Size must be 3 or smaller");
+    std::array<TVal, N> arr;
+
+    // Default Constructor
+    FooVec()
+    {
+        arr.fill(TVal());
+    }
+
+    // Constructor with initializer list
+    FooVec(std::initializer_list<TVal> initList)
+    {
+        if(initList.size() <= N)
+        {
+            std::copy(initList.begin(), initList.end(), arr.begin());
+        }
+        else
+        {
+            throw std::out_of_range("Initializer list size exceeds array size");
+        }
+    }
+
+    // Example member function to print the contents of the array
+    void printArray() const
+    {
+        for(auto const& element : arr)
+        {
+            std::cout << element << ' ';
+        }
+        std::cout << std::endl;
+    }
+};
+
+namespace alpaka::trait
+{
+
+    //! The DimType specialization for the user defined vector
+    //! \tparam TVal The data type.
+    //! \tparam N Vector size as a non-type parameter.
+    template<typename TVal, size_t N>
+    struct DimType<FooVec<TVal, N>>
+    {
+        using type = alpaka::DimInt<N>;
+    };
+
+    //! The ElemType specialization for the user defined vector
+    //! \tparam TVal The data type.
+    //! \tparam N Vector size as a non-type parameter.
+    template<typename TVal, size_t N>
+    struct ElemType<FooVec<TVal, N>>
+    {
+        using type = TVal;
+    };
+
+    //! The IdxType specialization for the user defined vecto
+    //! \tparam TVal The data type.
+    //! \tparam N Vector size as a non-type parameter.
+    template<typename TVal, size_t N>
+    struct IdxType<FooVec<TVal, N>>
+    {
+        using type = std::size_t;
+    };
+
+    //! Specialization for the user defined vector type FooVec. This specialization makes the vector usable in
+    //! WorkDivMembers construction. Since alpaka vectors use z-y-x order, FooVec is reversed.
+    //! \tparam TVal The element type of the vector type
+    //! \tparam N The size of the vector type
+    template<typename TVal, size_t N>
+    struct GetExtents<FooVec<TVal, N>>
+    {
+        ALPAKA_NO_HOST_ACC_WARNING
+        ALPAKA_FN_HOST_ACC constexpr auto operator()(FooVec<TVal, N> const& extent) const
+            -> alpaka::Vec<DimInt<N>, TVal>
+        {
+            alpaka::Vec<DimInt<N>, TVal> v{};
+#if BOOST_COMP_NVCC && BOOST_COMP_NVCC < BOOST_VERSION_NUMBER(11, 3, 0)
+            if(DimInt<N>::value > 0)
+#else
+            if constexpr(DimInt<N>::value > 0)
+#endif
+            {
+                // Reverse the vector since the dimensions ordered as z-y-x in alpaka
+                for(unsigned i = 0; i < DimInt<N>::value; i++)
+                    v[i] = extent.arr[DimInt<N>::value - i - 1];
+            }
+
+            return v;
+        }
+    };
+} // namespace alpaka::trait
--- a/test/unit/workDiv/src/WorkDivHelpersTest.cpp
+++ b/test/unit/workDiv/src/WorkDivHelpersTest.cpp
@@ -1,6 +1,8 @@
-/* Copyright 2022 Sergei Bastrakov, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2022 Sergei Bastrakov, Jan Stephan, Bernhard Manfred Gruber, Mehmet Yusufoglu
  * SPDX-License-Identifier: MPL-2.0
  */
+
+#include "FooVec.hpp"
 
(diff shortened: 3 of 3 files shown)


fwyzard:
Rather than `std::array`s (or in addition to them, if there is any concrete use case), could you add support for `[std::initializer_list](https://en.cppreference.com/w/cpp/utility/initializer_list)` arguments ?

That would support `alpaka::WorkDivMembers{{1u, 1u}, {2u, 2u}, {1u, 1u}};`.

include/alpaka/vec/Vec.hpp
```diff
@@ -656,6 +656,25 @@ namespace alpaka
         return r;
     }
 
+    //! \brief Converts std::array to alpaka Vec
+    //! \tparam TVal is the element type of the array
+    //! \tparam N is the size of the array
+    template<typename TVal, size_t N>
+    ALPAKA_FN_HOST_ACC constexpr auto arrayToVec(std::array<TVal, N> const& a) -> Vec<DimInt<N>, TVal>
```
psychocoderHPC: There is no need for an additional method that can only handle std array.
We have the traits `GetExtents` to provide the translation that alpaka can work with alpaka foreign types.

psychocoderHPC (changes requested):

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -33,6 +35,33 @@ namespace alpaka
...
+            , m_threadElemExtent(elemExtent)
+        {
+        }
+
+        //! \brief Accepts single specific type and is called without explicit template parameters.
+        //! \tparam T is the element type of the array
+        //! \tparam N is the size of the array
+        template<class T, size_t N>
+        ALPAKA_FN_HOST_ACC explicit WorkDivMembers(
+            std::array<T, N> const gridBlockExtent,
```
psychocoderHPC: The goal should not be to add a constructor which understand `std::array`. We like to stay interpotable with any alpaka foreign type which represents extents. `std::array` is one example it could also be `std::vector` or users type like in PIConGPU `pmacc::math::Vector<>`. 
To reach this goal we have the traits these give the alpaka user the possibility to be specialize for there own types.

This allows alpaka to be very generic and therefore we do not need to write constructors for any typical used type. As sson as we have C++20 we could write everything on a much nicer why but until than we need to life with the trais.

bernhardmgruber:
> That would support `alpaka::WorkDivMembers{{1u, 1u}, {2u, 2u}, {1u, 1u}};`.

This may already work with #2240, if the right constructor is selected. I think `alpaka::Vec` can be created from an initializer list.

bernhardmgruber:
> > That would support `alpaka::WorkDivMembers{{1u, 1u}, {2u, 2u}, {1u, 1u}};`.
> 
> This may already work with #2240, if the right constructor is selected. I think `alpaka::Vec` can be created from an initializer list.

It turns out it doesn't work, and I am not exactly sure why. The compiler fails to deduce the template parameters of the deduction guide. It has something to do with `{1u, 1u}` being an initializer list, because `alpaka::WorkDivMembers{alpaka::Vec{1u, 1u}, {2u, 2u}, {1u, 1u}};` works successfully.

mehmetyusufoglu:
Yes, i also tried. Did not work.

26 Şub 2024 Pzt 21:23 tarihinde Bernhard Manfred Gruber <
***@***.***> şunu yazdı:

include/alpaka/vec/Vec.hpp
mehmetyusufoglu: Ok I used getExtents. There is already one implemented for 1D array arguments in "ViewPlanePtr" therefore I had to check the dimension. I will refine the solution. Thanks a lot.

include/alpaka/workdiv/WorkDivMembers.hpp
```diff
@@ -33,6 +33,32 @@ namespace alpaka
...
 
+        //! Accepts different alpaka vector types and takes the last TDim number of items. TDim is explicitly specified
+        //! at the constructor call.
+        ALPAKA_NO_HOST_ACC_WARNING
+        template<typename TGridBlockExtent>
+        ALPAKA_FN_HOST_ACC explicit WorkDivMembers(
+            TGridBlockExtent const& gridBlockExtent = TGridBlockExtent(),
+            TGridBlockExtent const& blockThreadExtent = TGridBlockExtent(),
+            TGridBlockExtent const& threadElemExtent = TGridBlockExtent())
+            : m_gridBlockExtent(getExtents<TGridBlockExtent>(gridBlockExtent))
```
psychocoderHPC: this should be `: m_gridBlockExtent(getExtents(gridBlockExtent))`

mehmetyusufoglu: This part is left to another PR. For now `getExtents` is not called directly at the constructor.

include/alpaka/extent/Traits.hpp
```diff
@@ -55,10 +55,10 @@ namespace alpaka
 
     //! \return The extents of the given object.
     ALPAKA_NO_HOST_ACC_WARNING
-    template<typename T>
-    ALPAKA_FN_HOST_ACC auto getExtents(T const& object) -> Vec<Dim<T>, Idx<T>>
+    template<typename TObject>
```
psychocoderHPC: changes from this file should not be part of this PR

test/unit/workDiv/CMakeLists.txt
```diff
@@ -6,10 +6,13 @@
 set(_TARGET_NAME "workDivTest")
 
 append_recursive_files_add_to_src_group("src/" "src/" "cpp" _FILES_SOURCE)
+append_recursive_files_add_to_src_group("src/" "src/" "hpp" _FILES_HEADER)
```
psychocoderHPC: This should not be required because the file is directly included the cpp.

test/unit/workDiv/CMakeLists.txt
```diff
@@ -6,10 +6,13 @@
...
 
 alpaka_add_executable(
     ${_TARGET_NAME}
-    ${_FILES_SOURCE})
+    ${_FILES_SOURCE}
+    ${_FILES_HEADER})
+
 target_link_libraries(
```
psychocoderHPC, line 18: Additionally 

```
target_include_directories(${_TARGET_NAME} PRIVATE "src")
```

is required to provide the target with the include directory where the hpps are located.

test/unit/workDiv/src/WorkDivHelpersTest.cpp
```diff
@@ -34,6 +34,71 @@ namespace
     }
 } // namespace
 
+namespace alpaka::trait
```
psychocoderHPC: could you move this to the FooVec, please? No need to have this code here in the test file since it corresponds to the FooVec class.

psychocoderHPC (changes requested):

test/unit/workDiv/CMakeLists.txt
mehmetyusufoglu: Yes, sure. Actually,  I noticed a different usage in   https://github.com/alpaka-group/alpaka/blob/develop/test/unit/math/CMakeLists.txt  and used the same approach.

mehmetyusufoglu:
> Rather than `std::array`s (or in addition to them, if there is any concrete use case), could you add support for `[std::initializer_list](https://en.cppreference.com/w/cpp/utility/initializer_list)` arguments ?
> 
> That would support `alpaka::WorkDivMembers{{1u, 1u}, {2u, 2u}, {1u, 1u}};`.

Currently the following are possible due to a merged PR:  `alpaka::WorkDivMembers<Dim2D,Idx>{{1u, 1u}, {2u, 2u}, {1u, 1u}};` .  OR  `alpaka::WorkDivMembers{Vec{1u, 1u}, Vec{2u, 2u}, Vec{1u, 1u}};`  Vec is alpaka::vec with specific Dim and Idx. 

I will consider your suggestion together with C++20, seems more reasonable I think.

test/unit/workDiv/src/FooVec.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    template<typename T>
+    struct has_member_arr<T, std::void_t<decltype(&T::arr)>> : std::true_type
+    {
+    };
+
+    //! The DimType specialization for the user defined vector which has a member called "arr"
+    //! \tparam T The vector type
+    template<typename T>
+    struct DimType<T, std::enable_if_t<has_member_arr<T>::value>>
```
bernhardmgruber:
```suggestion
    //! The DimType specialization for the user defined vector which has a member called "arr"
    //! \tparam T The vector type
    template<typename T>
    struct DimType<FooVec>
```
Why don't you just specialize for `FooVec` directly? I don't see a need for SFINAE here.

Applies to other traits as well.

test/unit/workDiv/src/FooVec.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+    // Move Assignment Operator
+    FooVec& operator=(FooVec&& other) noexcept
+    {
+        if(this != &other)
+        {
+            arr = std::move(other.arr);
+        }
+        return *this;
+    }
```
bernhardmgruber:
```suggestion
```
I don't see a reason for implementing special member functions.

bernhardmgruber (commented):
Here are a few comments:

test/unit/workDiv/src/FooVec.hpp
mehmetyusufoglu: FooVec is a template, not a type. But i will try to formulate your approach.

test/unit/workDiv/src/FooVec.hpp
mehmetyusufoglu: Ok removed. Might be misleading. Thanks.

test/unit/workDiv/src/FooVec.hpp
bernhardmgruber: Sorry, my bad. Should be:
```c++
template<typename TVec, std::size_t Size>
struct DimType<FooVec<TVec, Size>>
```

mehmetyusufoglu:
Yes, i've already did like this actually. Testing,.

Bernhard Manfred Gruber ***@***.***>, 8 Mar 2024 Cum, 13:37
tarihinde şunu yazdı:

test/unit/workDiv/src/FooVec.hpp
mehmetyusufoglu: done thanks.

psychocoderHPC (approved):

Issue #2192, opened 2023-11-21
`alpaka::getWarpSizes` incurs a noticeable overhead


fwyzard:
While porting the CMS pixel reconstruction from native CUDA to Alpaka, it was noticed that the use of the `alpaka::getWarpSizes(device)` function incurs a noticeable overhead.

See https://github.com/cms-sw/cmssw/pull/43064#issuecomment-1817590926 for the discussion.

A possible workaround is to cache the warp size in our code, instead of querying it for every event.

However, it would seem natural to cache this information within the Alpaka device objects, instead of querying the underlying back-end each time.

fwyzard:
I think that caching the warp sizes inside the device object would require
  - either filling it at construction time
  - or using a mutex to avoid setting the cache concurrently

psychocoderHPC:
IMO caching makes sense, we should store the value during the device creation then there will be no need for a mutex.

bernhardmgruber:
Is there a CUDA device with a warpSize not 32? I am almost in favor of hardcoding it ... Otherwise, we could just collect and cache the entire device properties (i.e. `cudaDeviceProp`), so we can also serve other values faster.

fwyzard:
Not that I know of.

But HIP devices can have a warp size of 32 or 64, depending on the GPU model and potentially on the environment settings.

psychocoderHPC:
Partly solved by #2246. Never the less we should cache all over runtime constant device properties within the device, than there is no need to query the API multiple times.

sbaldu:
I think this was also addressed by PR #2492

PR #2246, opened 2024-03-06
Fix slow getWarpSize problem

mehmetyusufoglu:
It was noticed that the use of the `alpaka::getWarpSizes(device)` function incurs a noticeable overhead. (The Issue #2192 )

**Solution:** 

- Cuda: The fastest option is returning 32, it is not changed. But specialisation is taken to the DevCudaRt.hpp file, which is much cleaner than using nested preprocessor statements. 

- Hip:  Rather than getting the whole device properties and selecting the attribute from that struct; getting directly the attribute seems a faster solution. `hipDeviceGetAttribute(&size, hipDeviceAttributeWarpSize, device)` (credits to @fwyzard )

**Notes:**
deviceAttributeWarpSize is added to ApiCudaRt.hpp but is not used.

### DIFF
--- a/include/alpaka/dev/DevUniformCudaHipRt.hpp
+++ b/include/alpaka/dev/DevUniformCudaHipRt.hpp
@@ -175,10 +175,11 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getPreferredWarpSize(DevUniformCudaHipRt<TApi> const& dev) -> std::size_t
             {
-                typename TApi::DeviceProp_t devProp;
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::getDeviceProperties(&devProp, dev.getNativeHandle()));
+                int warpSize = 0;
 
-                return static_cast<std::size_t>(devProp.warpSize);
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
+                    TApi::deviceGetAttribute(&warpSize, TApi::deviceAttributeWarpSize, dev.getNativeHandle()));
+                return static_cast<std::size_t>(warpSize);
             }
         };
 
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -75,6 +75,7 @@ namespace alpaka
         static constexpr DeviceAttr_t deviceAttributeMaxSharedMemoryPerBlock = ::cudaDevAttrMaxSharedMemoryPerBlock;
         static constexpr DeviceAttr_t deviceAttributeMaxThreadsPerBlock = ::cudaDevAttrMaxThreadsPerBlock;
         static constexpr DeviceAttr_t deviceAttributeMultiprocessorCount = ::cudaDevAttrMultiProcessorCount;
+        static constexpr DeviceAttr_t deviceAttributeWarpSize = ::cudaDevAttrWarpSize;
 
         static constexpr Limit_t limitPrintfFifoSize = ::cudaLimitPrintfFifoSize;
         static constexpr Limit_t limitMallocHeapSize = ::cudaLimitMallocHeapSize;
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -78,6 +78,7 @@ namespace alpaka
             = ::hipDeviceAttributeMaxSharedMemoryPerBlock;
         static constexpr DeviceAttr_t deviceAttributeMaxThreadsPerBlock = ::hipDeviceAttributeMaxThreadsPerBlock;
         static constexpr DeviceAttr_t deviceAttributeMultiprocessorCount = ::hipDeviceAttributeMultiprocessorCount;
+        static constexpr DeviceAttr_t deviceAttributeWarpSize = ::hipDeviceAttributeWarpSize;
 
 #    if HIP_VERSION >= 40'500'000
         static constexpr Limit_t limitPrintfFifoSize = ::hipLimitPrintfFifoSize;


fwyzard:
Hi @mehmetyusufoglu,
thanks for looking into this.

For the CUDA backend, I think you can leave the warp size hard coded to 32, since all CUDA GPUs have that warp size.

For the HIP backend, you can use [`hipDeviceGetAttribute(&size, hipDeviceAttributeWarpSize, device)`](https://docs.amd.com/projects/HIP/en/latest/doxygen/html/group___device.html#ga7080a145a4239a7276e0dc22062026c1).

mehmetyusufoglu:
> Hi @mehmetyusufoglu, thanks for looking into this.
> 
> For the CUDA backend, I think you can leave the warp size hard coded to 32, since all CUDA GPUs have that warp size.
> 
> For the HIP backend, you can use [`hipDeviceGetAttribute(&size, hipDeviceAttributeWarpSize, device)`](https://docs.amd.com/projects/HIP/en/latest/doxygen/html/group___device.html#ga7080a145a4239a7276e0dc22062026c1).

Thank you. Sure, I am fixing.

mehmetyusufoglu:
I could not understand the reason of the compile error which only appears at clang14_debug_analysis compiler.

Screenshot from 2024-03-06 14-33-09

mehmetyusufoglu:
This compilation error happens only at clang-14_cuda-11.2_debug_analysis compilation setup. I could not get the reason. 

Screenshot from 2024-03-06 14-33-09

AuroraPerego:
> This compilation error happens only at clang-14_cuda-11.2_debug_analysis compilation setup. I could not get the reason.

When you are compiling for CUDA the other backends are not enabled, therefore backend-specific objects (e.g. `ApiHipRt`) are not defined (the `Tag`s are the only exception).
That's why the specialization for CUDA was wrapped in an `#ifdef ALPAKA_ACC_GPU_CUDA_ENABLED`. 
The correct way of writing `getPreferredWarpSize()` should be something like:
```cpp
ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::deviceGetAttribute(
    &warpSize,
    TApi::deviceAttributeWarpSize,
    dev.getNativeHandle()));
```
with the generic `TApi` instead of `if`/`else`. `TApi` will then be specialized to either `ApiCudaRt` or `ApiHipRt` depending on the target you are compiling for.

However, as Andrea suggested, you should keep the specialization for CUDA with the 32 hard-coded, since it's the fastest way of obtaining the warp size.

mehmetyusufoglu:
Thanks, i am fixing.

6 Mart 2024 Çarşamba tarihinde Aurora Perego ***@***.***>
yazdı:

include/alpaka/dev/DevHipRt.hpp
```diff
@@ -13,6 +13,23 @@ namespace alpaka
...
+
+    namespace trait
+    {
+        template<typename TApi>
+        struct GetPreferredWarpSize<DevUniformCudaHipRt<TApi>, std::enable_if_t<std::is_same_v<ApiHipRt, TApi>>>
+        {
+            ALPAKA_FN_HOST static auto getPreferredWarpSize(DevUniformCudaHipRt<TApi> const& dev) -> std::size_t
+            {
+                // initial value is set to 32
+                int warpSize{32};
```
fwyzard:
```suggestion
                int warpSize = 32;
```

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -171,30 +172,15 @@ namespace alpaka
...
-        struct GetPreferredWarpSize<DevUniformCudaHipRt<ApiCudaRt>>
-        {
-            ALPAKA_FN_HOST static constexpr auto getPreferredWarpSize(DevUniformCudaHipRt<ApiCudaRt> const& /* dev */)
-                -> std::size_t
-            {
-                // All CUDA GPUs to date have a warp size of 32 threads.
-                return 32u;
+                // Same value for all cuda architectures
+                constexpr size_t warpSize{32};
+                return warpSize;
```
fwyzard: please revert this to
```suggestion
                // All CUDA GPUs to date have a warp size of 32 threads.
                return 32u;
```

fwyzard:
Why do you use a different approach for the CUDA and HIP cases ?

Either define them both in `include/alpaka/dev/DevUniformCudaHipRt.hpp`, or define each of them in `include/alpaka/dev/DevCudaRt.hpp` and `include/alpaka/dev/DevHipRt.hpp`.

fwyzard:
For example, the minimal changes would have been to change just
```c++
                typename TApi::DeviceProp_t devProp;
                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::getDeviceProperties(&devProp, dev.getNativeHandle()));
```
to
```c++
                int warpSize = 0;
                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
                    TApi::deviceGetAttribute(&warpSize, TApi::deviceAttributeWarpSize, dev.getNativeHandle()));
                return static_cast<std::size_t>(warpSize);
```

mehmetyusufoglu:
> ```c++
> TApi::deviceGetAttribute
> ```

Thanks. Let me finish, I don't want to return 32 from the deviceGetAttribute of Cuda, i will specialize  GetPreferredWarpSize only. As you said in 2 different files.

include/alpaka/dev/DevUniformCudaHipRt.hpp
psychocoderHPC: @mehmetyusufoglu Could you please address both suggestions from @fwyzard

mehmetyusufoglu:
> For example, the minimal changes would have been to change just
> 
> ```c++
>                 typename TApi::DeviceProp_t devProp;
>                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::getDeviceProperties(&devProp, dev.getNativeHandle()));
> ```
> 
> to
> 
> ```c++
>                 int warpSize = 0;
>                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
>                     TApi::deviceGetAttribute(&warpSize, TApi::deviceAttributeWarpSize, dev.getNativeHandle()));
>                 return static_cast<std::size_t>(warpSize);
> ```

@fwyzard  Do you think changing deviceGetAttibute is better

mehmetyusufoglu:
> ```c++
> int warpSize = 0;
>                 ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
>                     TApi::deviceGetAttribute(&warpSize, TApi::deviceAttributeWarpSize, dev.getNativeHandle()));
>                 return static_cast<std::size_t>(warpSize);
> ```

@fwyzard  What is your opinion, in my opinion, changing `deviceGetAttribute` function of `ApiCudaTr`  if the argument (the attribute) is `deviceAttributeWarpSize`, is not a clean solution. In the current solution I proposed; there is no if-else or nested #ifs anymore. Your suggestion needs less changes for sure.

fwyzard:
> @fwyzard Do you think changing deviceGetAttibute is better

no, `TApi::deviceGetAttibute` is a wrapper around the underlying API function, and it should keep the same behaviour and interface.

However, we don't _need_ to call `TApi::deviceGetAttibute` for the CUDA architecture, because we know that the warp size is always 32.

include/alpaka/dev/DevHipRt.hpp
```diff
@@ -13,6 +14,23 @@ namespace alpaka
...
+
+    namespace trait
+    {
+        //! The HIP RT preferred device warp size get trait specialization.
+        template<>
+        struct GetPreferredWarpSize<DevUniformCudaHipRt<ApiHipRt>>
+        {
+            ALPAKA_FN_HOST static auto getPreferredWarpSize(DevUniformCudaHipRt<ApiHipRt> const& dev) -> std::size_t
+            {
+                int warpSize{0};
```
fwyzard: please use
```suggestion
                int warpSize = 0;
```

include/alpaka/dev/DevHipRt.hpp
```diff
@@ -13,6 +14,23 @@ namespace alpaka
...
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_TAPI(
+                    ApiHipRt,
+                    ApiHipRt::deviceGetAttribute(&warpSize, ApiHipRt::deviceAttributeWarpSize, dev.getNativeHandle()));
```
fwyzard: instead of adding `ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_TAPI` you could do
```suggestion
                using TApi = ApiHipRt;
                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
                    ApiHipRt::deviceGetAttribute(&warpSize, ApiHipRt::deviceAttributeWarpSize, dev.getNativeHandle()));
```

include/alpaka/dev/DevCudaRt.hpp
```diff
@@ -13,6 +13,22 @@ namespace alpaka
...
+    namespace trait
+    {
+        //! The CUDA RT preferred device warp size get trait specialization.
+        template<>
+        struct GetPreferredWarpSize<DevUniformCudaHipRt<ApiCudaRt>>
+        {
+            ALPAKA_FN_HOST static auto getPreferredWarpSize(DevUniformCudaHipRt<ApiCudaRt> const& /* dev */)
+                -> std::size_t
+            {
+                // Same value for all cuda architectures
```
fwyzard: I think the old comment was more descriptive:
```suggestion
                // All CUDA GPUs to date have a warp size of 32 threads.
```

fwyzard:
There is one potential problem with this approach: a file that includes `alpaka/dev/DevUniformCudaHipRt.hpp` but does not include explicitly `alpaka/dev/DevHipRt.hpp` or `alpaka/dev/DevCudaRt.hpp` will miss the `getPreferredWarpSize()` specialisations.

Basically this change makes `DevUniformCudaHipRt.hpp` an internal implementation detail, that other code shuld not use directly.

@psychocoderHPC do you think this is a problem ?

include/alpaka/dev/DevHipRt.hpp
mehmetyusufoglu: Ok, thanks.

mehmetyusufoglu:
> There is one potential problem with this approach: a file that includes `alpaka/dev/DevUniformCudaHipRt.hpp` but does not include explicitly `alpaka/dev/DevHipRt.hpp` or `alpaka/dev/DevCudaRt.hpp` will miss the `getPreferredWarpSize()` specialisations.
> 
> Basically this change makes `DevUniformCudaHipRt.hpp` an internal implementation detail, that other code shuld not use directly.
> 
> @psychocoderHPC do you think this is a problem ?

Ok, I selected the shorter solution suggested by Andrea, @fwyzard . Specialisations in 2 different files is cleaner but needed many changes.

fwyzard (approved):

psychocoderHPC (approved):

PR #2250, opened 2024-03-19
Fix tsan error: change memory randomization

mehmetyusufoglu:
Decrease mmap randomization to fix the tsan error `FATAL: ThreadSanitizer: unexpected memory mapping`

### DIFF
--- a/script/ci.sh
+++ b/script/ci.sh
@@ -9,6 +9,11 @@ source ./script/set.sh
 
 ./script/print_env.sh
 source ./script/before_install.sh
+if [ -n "$GITHUB_ACTIONS" ] && [ "$ALPAKA_CI_OS_NAME" = "Linux" ]; then
+  # Workaround for the error: ThreadSanitizer: unexpected memory mapping
+  # change the configuration of the address space layout randomization
+  sudo sysctl vm.mmap_rnd_bits=28
+fi
 
 if [ "$ALPAKA_CI_OS_NAME" = "Linux" ]
 then


psychocoderHPC:
This looks like the similar issue: https://github.com/google/sanitizers/issues/1716

psychocoderHPC:
If I see it correctly the bug should be solved in llvm 17+ see https://github.com/llvm/llvm-project/commit/fb77ca05ffb4f8e666878f2f6718a9fb4d686839

SimeonEhrig:
> If I see it correctly the bug should be solved in llvm 17+ see [llvm/llvm-project@fb77ca0](https://github.com/llvm/llvm-project/commit/fb77ca05ffb4f8e666878f2f6718a9fb4d686839)

No, it does not solve the problem. I already tested it: https://github.com/alpaka-group/alpaka/pull/2249/commits/812035ea3733a8854501deec9454b7dda3652ffe

I think, this fixes another bug.

mehmetyusufoglu:
this is a workaround for linux OS. Tsan does not fail anymore.
`sudo sysctl vm.mmap_rnd_bits=28`

I will put a condition to set this only for tsan or only for linux.

script/set.sh
```diff
@@ -11,4 +11,6 @@
 # -u: treat unset variables as an error
 # -v: Print shell input lines as they are read
 # -x: Print command traces before executing command
+sudo sysctl vm.mmap_rnd_bits=28
```
SimeonEhrig: You was faster. I would also suggest it :-) It works.

Please guard it with:

```bash
if [ -n "$GITHUB_ACTIONS" ] && [ "$ALPAKA_CI_OS_NAME" = "Linux" ]; then
    sudo sysctl vm.mmap_rnd_bits=28
fi
```

An move it to `ci.sh`. `set.sh` is sourced many times and it's meaning is really executing the `set` command for the bash.

SimeonEhrig (changes requested):

psychocoderHPC (approved):

psychocoderHPC:
@mehmetyusufoglu please update the description of this PR

mehmetyusufoglu:
> @mehmetyusufoglu please update the description of this PR

Done, thanks.

SimeonEhrig (approved):

Issue #2222, opened 2024-01-15
`getValidWorkDiv()` and `isValidWorkDiv()` is broken


psychocoderHPC:
Both functions using the device properties (`TApi::getDeviceProperties()`) to check if a work division is correct for the device.
In the case of CUDA and ROCm this is not valid because `TApi::funcGetAttributes(&funcAttrs, kernelName)` must be used.
If a kernel is using a high amount of registers the number of threads per block can be less than the maximal number of threads a device can handle.
The debug message shown for any kernel execution using `TApi::funcGetAttributes(&funcAttrs, kernelName)` and show therefore the correct restrictions.

To fix this bug it is required to change the interface of `getValidWorkDiv()` because the kernel is required for the query.

Two of my students found the issue during porting LULESH to alpaka:

# Kernel 1 can use 1024 threads:

![Screenshot from 2024-01-12 20-33-29](https://github.com/alpaka-group/alpaka/assets/4037125/1238540d-5ffa-4a22-923f-0fa5a0b9e423)

# Kernel 2 used to many registers and can therefor run only with 512 threads

![Screenshot from 2024-01-12 20-45-19](https://github.com/alpaka-group/alpaka/assets/4037125/9a512fde-54f7-4d04-b98f-aba1e554d7e8)

psychocoderHPC:
suggestion for the interface change:

```C++
    template<
        typename TAcc,
        typename TDev,
        typename TGridElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>,
        typename TThreadElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>>
    ALPAKA_FN_HOST auto getValidWorkDiv(
        [[maybe_unused]] TDev const& dev,
        [[maybe_unused]] TGridElemExtent const& gridElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>::ones(),
        [[maybe_unused]] TThreadElemExtent const& threadElemExtents = Vec<Dim<TAcc>, Idx<TAcc>>::ones(),
        [[maybe_unused]] bool blockThreadMustDivideGridThreadExtent = true,
        [[maybe_unused]] GridBlockExtentSubDivRestrictions gridBlockExtentSubDivRestrictions
        = GridBlockExtentSubDivRestrictions::Unrestricted)
        -> WorkDivMembers<Dim<TGridElemExtent>, Idx<TGridElemExtent>>
```

to

```C++
    template<
        typename TAcc,
[...]

PR #2251, opened 2024-03-21
Fix getValidWorkDiv for CUDA and ROCm [I] #2222

mehmetyusufoglu:
`getValidWorkDiv`  has a bug. It only considers device _hard_ properties ( `TApi::getDeviceProperties()`); does not consider the kernel function. Actually kernel function properties can limit number of threads per block. 

In the case of CUDA and ROCm, the function `getValidWorkDiv` should  use  `TApi::funcGetAttributes(&funcAttrs, kernelName)`  together with  `TApi::getDeviceProperties()`.

 I have added 3 functions.  `getFunctionAttributes`,`getValidWorkDivKernel` and `isValidWorkDivKernel`.   One function intrinsic to Alpaka changed: `subDivideGridElems`. Now has an additional argument; "threads per block", if the value is zero the function uses device hard properties as before.

**Next PR:** The 2 functions `getValidWorkDivKernel` and `isValidWorkDivKernel` will be used in Examples and Tests instead of getValidWorkDiv and isValidWorkDiv at the next PR. The names can be determined.

Credits to @psychocoderHPC  for the PR.
**Tests:** The tests for 1D and 2D workdivs were added.

### DIFF
--- a/include/alpaka/kernel/KernelBundle.hpp
+++ b/include/alpaka/kernel/KernelBundle.hpp
@@ -1,0 +1,65 @@
+/* Copyright 2022 Benjamin Worpitz, Bert Wesarg, René Widera, Sergei Bastrakov, Bernhard Manfred Gruber, Mehmet
+ * Yusufoglu SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include <alpaka/core/RemoveRestrict.hpp>
+
+#include <tuple>
+#include <type_traits>
+
+namespace alpaka
+{
+    //! \brief The class used to bind kernel function object and arguments together. Once an instance of this class is
+    //! created, arguments are not needed to be separately given to functions who need kernel function and arguments.
+    //! \tparam TKernelFn The kernel function object type.
+    //! \tparam TArgs Kernel function object invocation argument types as a parameter pack.
+    template<typename TKernelFn, typename... TArgs>
+    class KernelBundle
+    {
+    public:
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wdocumentation" // clang does not support the syntax for variadic template
+                                                       // arguments "args,...". Ignore the error.
+#endif
+        //! \param kernelFn The kernel function-object
+        //! \param args,... The kernel invocation arguments.
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
+        KernelBundle(TKernelFn const& kernelFn, TArgs&&... args)
+            : m_kernelFn(kernelFn)
+            , m_args(std::forward<TArgs>(args)...)
+        {
+        }
+
+        //! The function object type
+        using KernelFn = TKernelFn;
+        //! Tuple type to encapsulate kernel function argument types and argument values
+        using ArgTuple = std::tuple<remove_restrict_t<std::decay_t<TArgs>>...>;
+
+        KernelFn m_kernelFn;
+        ArgTuple m_args;
+    };
+
+    //! \brief User defined deduction guide with trailing return type. For CTAD during the construction.
+    //! \tparam TKernelFn The kernel function object type.
+    //! \tparam TArgs Kernel function object argument types as a parameter pack.
+    //! \param kernelFn The kernel object
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic push
+#    pragma clang diagnostic ignored "-Wdocumentation" // clang does not support the syntax for variadic template
+                                                       // arguments "args,...". Ignore the error.
+#endif
+    //! \param args,... The kernel invocation arguments.
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
+    //! \return Kernel function bundle. An instance of KernelBundle which consists the kernel function object and its
+    //! arguments.
+    template<typename TKernelFn, typename... TArgs>
+    KernelBundle(TKernelFn const& kernelFn, TArgs&&... args) -> KernelBundle<TKernelFn, TArgs...>;
+
+} // namespace alpaka
--- a/include/alpaka/kernel/KernelFunctionAttributes.hpp
+++ b/include/alpaka/kernel/KernelFunctionAttributes.hpp
@@ -1,0 +1,25 @@
+/* Copyright 2022 René Widera, Mehmet Yusufoglu
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include <cstddef>
+
+namespace alpaka
+{
+    //! Kernel function attributes struct. Attributes are filled by calling the API of the accelerator using the kernel
+    //! function as an argument. In case of a CPU backend, maxThreadsPerBlock is set to 1 and other values remain zero
+    //! since there are no correponding API functions to get the values.
+    struct KernelFunctionAttributes
+    {
+        std::size_t constSizeBytes{0};
+        std::size_t localSizeBytes{0};
+        std::size_t sharedSizeBytes{0};
+        int maxDynamicSharedSizeBytes{0};
+        int numRegs{0};
+        // This field is ptx or isa version if the backend is GPU
+        int asmVersion{0};
+        int maxThreadsPerBlock{0};
+    };
+} // namespace alpaka
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -17,6 +17,8 @@
 #include "alpaka/core/OmpSchedule.hpp"
 #include "alpaka/dev/DevCpu.hpp"
 #include "alpaka/idx/MapIdx.hpp"
+#include "alpaka/kernel/KernelBundle.hpp"
+#include "alpaka/kernel/KernelFunctionAttributes.hpp"
 #include "alpaka/kernel/Traits.hpp"
 #include "alpaka/workdiv/WorkDivMembers.hpp"
 
@@ -946,6 +948,30 @@ namespace alpaka
         {
             using type = TIdx;
         };
+
+        //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDim The dimensionality of the accelerator device properties.
+        //! \tparam TIdx The idx type of the accelerator device properties.
+        //! \tparam TKernelFn Kernel function object type.
+        //! \tparam TArgs Kernel function object argument types as a parameter pack.
+        template<typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
+        struct FunctionAttributes<AccCpuOmp2Blocks<TDim, TIdx>, KernelBundle<TKernelFn, TArgs...>>
+        {
+            //! \param kernelBundle Kernel bundeled with it's arguments. The function attributes of this kernel will be
+            //! determined. Max threads per block is one of the attributes.
+            //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
+            //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+                [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
+                -> alpaka::KernelFunctionAttributes
+            {
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+                kernelFunctionAttributes.maxThreadsPerBlock = 1u;
+                return kernelFunctionAttributes;
+            }
+        };
+
     } // namespace trait
 } // namespace alpaka
 
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -975,6 +975,8 @@ namespace alpaka
                 // properties function.
                 auto const& props = alpaka::getAccDevProps<AccCpuOmp2Blocks<TDim, TIdx>>(dev);
                 kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(props.m_blockThreadCountMax);
+                kernelFunctionAttributes.maxDynamicSharedSizeBytes
+                    = static_cast<int>(alpaka::BlockSharedDynMemberAllocKiB * 1024);
                 return kernelFunctionAttributes;
             }
         };
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -20,6 +20,7 @@
 #include "alpaka/kernel/KernelBundle.hpp"
 #include "alpaka/kernel/KernelFunctionAttributes.hpp"
 #include "alpaka/kernel/Traits.hpp"
+#include "alpaka/platform/PlatformCpu.hpp"
 #include "alpaka/workdiv/WorkDivMembers.hpp"
 
 #include <functional>
@@ -961,13 +962,17 @@ namespace alpaka
             //! determined. Max threads per block is one of the attributes.
             //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
             //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
-            ALPAKA_NO_HOST_ACC_WARNING
-            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+            ALPAKA_FN_HOST static auto getFunctionAttributes(
                 [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
                 -> alpaka::KernelFunctionAttributes
             {
                 alpaka::KernelFunctionAttributes kernelFunctionAttributes;
-                kernelFunctionAttributes.maxThreadsPerBlock = 1u;
+                using Acc = AccCpuOmp2Blocks<TDim, TIdx>;
+                auto const platformAcc = alpaka::Platform<Acc>{};
+                auto const dev = alpaka::getDevByIdx(platformAcc, 0);
+                // set function properties to device properties
+                auto const& props = alpaka::getAccDevProps<Acc>(dev);
+                kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(props.m_blockThreadCountMax);
                 return kernelFunctionAttributes;
             }
         };
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
@@ -951,27 +951,29 @@ namespace alpaka
         };
 
         //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDev The device type.
         //! \tparam TDim The dimensionality of the accelerator device properties.
         //! \tparam TIdx The idx type of the accelerator device properties.
         //! \tparam TKernelFn Kernel function object type.
         //! \tparam TArgs Kernel function object argument types as a parameter pack.
-        template<typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
-        struct FunctionAttributes<AccCpuOmp2Blocks<TDim, TIdx>, KernelBundle<TKernelFn, TArgs...>>
+        template<typename TDev, typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
+        struct FunctionAttributes<AccCpuOmp2Blocks<TDim, TIdx>, TDev, KernelBundle<TKernelFn, TArgs...>>
         {
+            //! \param dev The device instance
             //! \param kernelBundle Kernel bundeled with it's arguments. The function attributes of this kernel will be
             //! determined. Max threads per block is one of the attributes.
             //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
             //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
             ALPAKA_FN_HOST static auto getFunctionAttributes(
+                TDev const& dev,
                 [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
                 -> alpaka::KernelFunctionAttributes
             {
                 alpaka::KernelFunctionAttributes kernelFunctionAttributes;
-                using Acc = AccCpuOmp2Blocks<TDim, TIdx>;
-                auto const platformAcc = alpaka::Platform<Acc>{};
-                auto const dev = alpaka::getDevByIdx(platformAcc, 0);
-                // set function properties to device properties
-                auto const& props = alpaka::getAccDevProps<Acc>(dev);
+
+                // set function properties for maxThreadsPerBlock to device properties, since API doesn't have function
+                // properties function.
+                auto const& props = alpaka::getAccDevProps<AccCpuOmp2Blocks<TDim, TIdx>>(dev);
                 kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(props.m_blockThreadCountMax);
                 return kernelFunctionAttributes;
             }
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
@@ -15,6 +15,8 @@
 #include "alpaka/acc/AccCpuOmp2Threads.hpp"
 #include "alpaka/core/Decay.hpp"
 #include "alpaka/dev/DevCpu.hpp"
+#include "alpaka/kernel/KernelBundle.hpp"
+#include "alpaka/kernel/KernelFunctionAttributes.hpp"
 #include "alpaka/kernel/Traits.hpp"
 #include "alpaka/meta/NdLoop.hpp"
 #include "alpaka/workdiv/WorkDivMembers.hpp"
@@ -192,6 +194,30 @@ namespace alpaka
         {
             using type = TIdx;
         };
+
+        //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDim The dimensionality of the accelerator device properties.
+        //! \tparam TIdx The idx type of the accelerator device properties.
+        //! \tparam TKernelFn Kernel function object type.
+        //! \tparam TArgs Kernel function object argument types as a parameter pack.
+        template<typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
+        struct FunctionAttributes<AccCpuOmp2Threads<TDim, TIdx>, KernelBundle<TKernelFn, TArgs...>>
+        {
+            //! \param kernelBundle Kernel bundeled with it's arguments. The function attributes of this kernel will be
+            //! determined. Max threads per block is one of the attributes.
+            //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
+            //! fields. For CPU, the field of max threads allowed by the given kernel function for the block is 1.
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+                [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
+                -> alpaka::KernelFunctionAttributes
+            {
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+                kernelFunctionAttributes.maxThreadsPerBlock = 1u;
+                return kernelFunctionAttributes;
+            }
+        };
+
     } // namespace trait
 } // namespace alpaka
 
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
@@ -221,6 +221,8 @@ namespace alpaka
                 // properties function.
                 auto const& props = alpaka::getAccDevProps<AccCpuOmp2Threads<TDim, TIdx>>(dev);
                 kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(props.m_blockThreadCountMax);
+                kernelFunctionAttributes.maxDynamicSharedSizeBytes
+                    = static_cast<int>(alpaka::BlockSharedDynMemberAllocKiB * 1024);
                 return kernelFunctionAttributes;
             }
         };
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
@@ -19,6 +19,7 @@
 #include "alpaka/kernel/KernelFunctionAttributes.hpp"
 #include "alpaka/kernel/Traits.hpp"
 #include "alpaka/meta/NdLoop.hpp"
+#include "alpaka/platform/PlatformCpu.hpp"
 #include "alpaka/workdiv/WorkDivMembers.hpp"
 
 #include <functional>
@@ -208,12 +209,17 @@ namespace alpaka
             //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
             //! fields. For CPU, the field of max threads allowed by the given kernel function for the block is 1.
             ALPAKA_NO_HOST_ACC_WARNING
-            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+            ALPAKA_FN_HOST static auto getFunctionAttributes(
                 [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
                 -> alpaka::KernelFunctionAttributes
             {
                 alpaka::KernelFunctionAttributes kernelFunctionAttributes;
-                kernelFunctionAttributes.maxThreadsPerBlock = 1u;
+                using Acc = AccCpuOmp2Threads<TDim, TIdx>;
+                auto const platformAcc = alpaka::Platform<Acc>{};
+                auto const dev = alpaka::getDevByIdx(platformAcc, 0);
+                // set function properties to device properties
+                auto const& props = alpaka::getAccDevProps<Acc>(dev);
+                kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(props.m_blockThreadCountMax);
                 return kernelFunctionAttributes;
             }
         };
--- a/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
@@ -197,28 +197,29 @@ namespace alpaka
         };
 
         //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDev The device type.
         //! \tparam TDim The dimensionality of the accelerator device properties.
         //! \tparam TIdx The idx type of the accelerator device properties.
         //! \tparam TKernelFn Kernel function object type.
         //! \tparam TArgs Kernel function object argument types as a parameter pack.
-        template<typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
-        struct FunctionAttributes<AccCpuOmp2Threads<TDim, TIdx>, KernelBundle<TKernelFn, TArgs...>>
+        template<typename TDev, typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
+        struct FunctionAttributes<AccCpuOmp2Threads<TDim, TIdx>, TDev, KernelBundle<TKernelFn, TArgs...>>
         {
+            //! \param dev The device instance
             //! \param kernelBundle Kernel bundeled with it's arguments. The function attributes of this kernel will be
             //! determined. Max threads per block is one of the attributes.
             //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
-            //! fields. For CPU, the field of max threads allowed by the given kernel function for the block is 1.
-            ALPAKA_NO_HOST_ACC_WARNING
+            //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
             ALPAKA_FN_HOST static auto getFunctionAttributes(
+                TDev const& dev,
                 [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
                 -> alpaka::KernelFunctionAttributes
             {
                 alpaka::KernelFunctionAttributes kernelFunctionAttributes;
-                using Acc = AccCpuOmp2Threads<TDim, TIdx>;
-                auto const platformAcc = alpaka::Platform<Acc>{};
-                auto const dev = alpaka::getDevByIdx(platformAcc, 0);
-                // set function properties to device properties
-                auto const& props = alpaka::getAccDevProps<Acc>(dev);
+
+                // set function properties for maxThreadsPerBlock to device properties, since API doesn't have function
+                // properties function.
+                auto const& props = alpaka::getAccDevProps<AccCpuOmp2Threads<TDim, TIdx>>(dev);
                 kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(props.m_blockThreadCountMax);
                 return kernelFunctionAttributes;
             }
--- a/include/alpaka/kernel/TaskKernelCpuThreads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuThreads.hpp
@@ -17,6 +17,8 @@
 #include "alpaka/core/Decay.hpp"
 #include "alpaka/core/ThreadPool.hpp"
 #include "alpaka/dev/DevCpu.hpp"
+#include "alpaka/kernel/KernelBundle.hpp"
+#include "alpaka/kernel/KernelFunctionAttributes.hpp"
 #include "alpaka/kernel/Traits.hpp"
 #include "alpaka/meta/NdLoop.hpp"
 #include "alpaka/workdiv/WorkDivMembers.hpp"
@@ -200,6 +202,30 @@ namespace alpaka
         {
             using type = TIdx;
         };
+
+        //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDim The dimensionality of the accelerator device properties.
+        //! \tparam TIdx The idx type of the accelerator device properties.
+        //! \tparam TKernelFn Kernel function object type.
+        //! \tparam TArgs Kernel function object argument types as a parameter pack.
+        template<typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
+        struct FunctionAttributes<AccCpuThreads<TDim, TIdx>, KernelBundle<TKernelFn, TArgs...>>
+        {
+            //! \param kernelBundle Kernel bundeled with it's arguments. The function attributes of this kernel will be
+            //! determined. Max threads per block is one of the attributes.
+            //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
+            //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+                [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
+                -> alpaka::KernelFunctionAttributes
+            {
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+                kernelFunctionAttributes.maxThreadsPerBlock = 1u;
+                return kernelFunctionAttributes;
+            }
+        };
+
     } // namespace trait
 } // namespace alpaka
 
--- a/include/alpaka/kernel/TaskKernelCpuThreads.hpp
+++ b/include/alpaka/kernel/TaskKernelCpuThreads.hpp
@@ -205,28 +205,29 @@ namespace alpaka
         };
 
         //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDev The device type.
         //! \tparam TDim The dimensionality of the accelerator device properties.
         //! \tparam TIdx The idx type of the accelerator device properties.
         //! \tparam TKernelFn Kernel function object type.
         //! \tparam TArgs Kernel function object argument types as a parameter pack.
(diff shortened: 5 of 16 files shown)


psychocoderHPC:
offline discussed in the developer meeting:
https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaFuncAttributes.html#structcudaFuncAttributes provides more properties than the maximum number of threads per block.
We should at least design the API so that we can get more than one property. 

`maxThreadsPerBlock` and `maxDynamicSharedSizeBytes` are what all our backends can provide. I am not fully sure if for CPU backends `maxDynamicSharedSizeBytes` must be reduced by the compile time shared memory, if so it will be hard to provide the correct amount of dynamic shared memory.

include/alpaka/kernel/Traits.hpp
```diff
@@ -69,6 +77,42 @@ namespace alpaka
...
+            template<typename TDim, typename TIdx>
+            ALPAKA_FN_HOST_ACC static auto getMaxThreadsInBlockForKernel(
+                [[maybe_unused]] TKernelFnObj const& kernelFnObj,
+                [[maybe_unused]] TArgs const&... args) -> std::size_t
+            {
+                // auto kernelName = alpaka::detail::
+                //     gpuKernel<TKernelFnObj, TApi, TAcc, TDim, TIdx, remove_restrict_t<std::decay_t<TArgs>>...>;
+                // // Log the function attributes.
+                printf("gWorkDivForKernel test default. Return zero\n");
+#if BOOST_ARCH_PTX && (BOOST_ARCH_PTX < BOOST_VERSION_NUMBER(2, 0, 0))
```
psychocoderHPC: This check is not required because CUDA is not supporting sm_20 anymore.

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -10,19 +10,57 @@
...
 #include <functional>
 #include <set>
+#include <tuple>
 #include <type_traits>
 
 //! The alpaka library.
 namespace alpaka
 {
+    template<typename TAcc, typename TKernelFn, typename... TArgs>
+    class KernelBundle
```
psychocoderHPC: why do you moved this class from the file `KernelBundle` into `WorkDivHelper`?

psychocoderHPC (changes requested):

include/alpaka/workdiv/WorkDivHelpers.hpp
mehmetyusufoglu: this is for a temporary test

mehmetyusufoglu:
> offline discussed in the developer meeting: https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaFuncAttributes.html#structcudaFuncAttributes provides more properties than the maximum number of threads per block. We should at least design the API so that we can get more than one property.
> 
> `maxThreadsPerBlock` and `maxDynamicSharedSizeBytes` are what all our backends can provide. I am not fully sure if for CPU backends `maxDynamicSharedSizeBytes` must be reduced by the compile time shared memory, if so it will be hard to provide the correct amount of dynamic shared memory.

I

mehmetyusufoglu:
> offline discussed in the developer meeting: https://docs.nvidia.com/cuda/cuda-runtime-api/structcudaFuncAttributes.html#structcudaFuncAttributes provides more properties than the maximum number of threads per block. We should at least design the API so that we can get more than one property.
> 
> `maxThreadsPerBlock` and `maxDynamicSharedSizeBytes` are what all our backends can provide. I am not fully sure if for CPU backends `maxDynamicSharedSizeBytes` must be reduced by the compile time shared memory, if so it will be hard to provide the correct amount of dynamic shared memory.

Done. A structure is returned for GPU and CPUs. For CPUs only maxthreadsperblock field is used and it is 1, for GPU backends all the values accessible by the Cuda and Hip APIs are set and usable.

include/alpaka/kernel/TaskKernelCpuOmp2Blocks.hpp
```diff
@@ -945,7 +947,31 @@ namespace alpaka
         struct IdxType<TaskKernelCpuOmp2Blocks<TDim, TIdx, TKernelFnObj, TArgs...>>
         {
             using type = TIdx;
+            1
```
mehmetyusufoglu:
```suggestion
```

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    auto const& bundeledKernel = alpaka::makeKernelBundle<Acc>(kernel, kernelArg);
+
+    // Test getValidWorkDivForKernel for 1024 threads per grid
+    auto const workDiv = alpaka::getValidWorkDivForKernel(dev, bundeledKernel, Vec{1024}, Vec{1});
+    // Test isValidWorkDivKernel function
+    auto const isWorkDivValidForKernel = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, workDiv);
+    CHECK(isWorkDivValidForKernel == true);
+
+// If Acc is Cuda or Hip
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
```
psychocoderHPC: This is not checking if the Acc is CUDA or HIP.  I do not understand if you like to active the code if CUDA is enabled in CMake or if you like to go into this code only if the ACC is CUDA

psychocoderHPC: IMO the macro guards are not required at all.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,182 @@
...
+    auto const& bundeledKernel = alpaka::makeKernelBundle<Acc>(kernel, kernelArg);
+
+    // Test getValidWorkDivForKernel function for 1024x2 threads per grid.
+    auto const workDiv = alpaka::getValidWorkDivForKernel(dev, bundeledKernel, Vec{1024, 2}, Vec{1, 1});
+    // Test isValidWorkDivKernel function
+    auto const isWorkDivValidForKernel = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, workDiv);
+    CHECK(isWorkDivValidForKernel == true);
+
+// If Acc is Cuda or Hip
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
```
psychocoderHPC: same here I think the macro guards are not required

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,182 @@
...
+        // Expected valid workdiv for this kernel.
+        CHECK(workDiv == WorkDiv{Vec{4, 1}, Vec{256, 2}, Vec{1, 1}});
+    }
+    // For ApiHipRt
+    else if constexpr(std::is_same_v<
+                          alpaka::AccGpuUniformCudaHipRt<alpaka::ApiHipRt, alpaka::Dim<Acc>, alpaka::Idx<Acc>>,
+                          Acc>)
+    {
+        // Expected valid workdiv with this kernel
+        CHECK(workDiv == WorkDiv{Vec{2, 1}, Vec{512, 2}, Vec{1, 1}});
```
psychocoderHPC: IMO we should check that the blocksize is not 1024 because we do not know what the compiler is producing for different CUDA architectures.

psychocoderHPC (changes requested):

include/alpaka/kernel/KernelBundle.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include <type_traits>
+
+namespace alpaka
+{
+    //! \brief The class used to bind kernel function object and arguments together. Once an instance of this class is
+    //! created, arguments are not needed to be separately given to functions who need kernel function and arguments.
+    //! \tparam TAcc The accelerator type.
+    //! \tparam TKernelFn The kernel function object type.
+    //! \tparam TArgs Kernel function object invocation argument types as a parameter pack.
+    template<typename TAcc, typename TKernelFn, typename... TArgs>
```
psychocoderHPC: I know I wrote this object but after looking again on the code I think we should remove Acc from this object.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: When I remove precompile directives and only use `if constexpr` I get "type undefined" error... Ok I will try to solve by traits someow merging `is_same` and "if type defined" checks.

```
if constexpr(std::is_same_v<
                     alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, alpaka::Dim<Acc>, alpaka::Idx<Acc>>,
                     Acc>)
```

psychocoderHPC: use tags here instead of `alpaka::AccGpuUniformCudaHipRt`

psychocoderHPC:
I check our current code. All implementations of `TaskKernel*` using a tuple to store the arguments internally therefore using the "kernelBoundle" object which is unifying this approche will reduce the possibility of implementation issues and should not be negative for the performance compared to the current state of the code.

include/alpaka/kernel/KernelBundle.hpp
```diff
@@ -0,0 +1,65 @@
...
+    //! \tparam TArgs Kernel function object argument types as a parameter pack.
+    //! \param kernelFn The kernel object
+    //! \param args,... The kernel invocation arguments.
+    //! \return Kernel function bundle. An instance of KernelBundle which consists the kernel function object and its
+    //! arguments.
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
+    template<typename TKernelFn, typename... TArgs>
+    inline auto makeKernelBundle(TKernelFn const& kernelFn, TArgs&&... args) -> KernelBundle<TKernelFn, TArgs...>
```
psychocoderHPC: not required anymore you can write

```C++
KernelBundle foo = {[](){},1,3.0};
```

psychocoderHPC (changes requested):

include/alpaka/kernel/KernelFunctionAttributes.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include <cstddef>
+
+namespace alpaka
+{
+    //! Kernel function attributes struct. Attributes are filled by calling the API of the accelerator using the kernel
```
psychocoderHPC, line 11: please add a comment with the behaviour if an entry is not filled by the underlaying api e.g. on CPU

include/alpaka/kernel/KernelBundle.hpp
mehmetyusufoglu: removed, thanks. But i had to add a deduction guide instead.

include/alpaka/kernel/KernelBundle.hpp
```diff
@@ -0,0 +1,65 @@
...
+                                                       // arguments "args,...". Ignore the error.
+#endif
+    //! \param args,... The kernel invocation arguments.
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
+    //! \return Kernel function bundle. An instance of KernelBundle which consists the kernel function object and its
+    //! arguments.
+    template<typename TKernelFn, typename... TArgs>
+    KernelBundle(TKernelFn const& kernelFn, TArgs&&... args) -> KernelBundle<TKernelFn, TArgs...>;
```
psychocoderHPC, line 63: Is this deduction guide required? IMO there should be no need, if there is a requirement maybe add the compiler into the description why we need it. My tests on godbolt compiler explorer worked without it.

include/alpaka/kernel/KernelFunctionAttributes.hpp
psychocoderHPC: please add a description to which values like localSizeBytes will be set if those values are not available for a backend/accelerator.

include/alpaka/kernel/KernelFunctionAttributes.hpp
```diff
@@ -0,0 +1,24 @@
...
+    //! function as an argument. In case of a CPU backend, maxThreadsPerBlock is set to 1 since there are no
+    //! correponding API functions to get the values.
+    struct KernelFunctionAttributes
+    {
+        std::size_t constSizeBytes{0};
+        std::size_t localSizeBytes{0};
+        std::size_t sharedSizeBytes{0};
+        int maxDynamicSharedSizeBytes{0};
+        int numRegs{0};
+        int ptxVersion{0};
```
psychocoderHPC: maybe call it asm version and add a description that this can be ISA or PTX version depending on the GPU

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +392,81 @@ namespace alpaka
...
+                    TKernelFn,
+                    TApi,
+                    AccGpuUniformCudaHipRt<TApi, TDim, TIdx>,
+                    TDim,
+                    TIdx,
+                    remove_restrict_t<std::decay_t<TArgs>>...>;
+
+
+#        if BOOST_ARCH_PTX && (BOOST_ARCH_PTX < BOOST_VERSION_NUMBER(2, 0, 0))
+#            error "Device capability >= 2.0 is required!"
```
psychocoderHPC: This can be removed all current supported CUDA compiler compile for `sm_5X` or higher all other are not supported.

psychocoderHPC (changes requested):
Please update the cheat sheet too! https://alpaka.readthedocs.io/en/latest/basic/cheatsheet.html#kernel-execution

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,204 @@
...
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+    auto const platform = alpaka::Platform<Acc>{};
+    auto const dev = alpaka::getDevByIdx(platform, 0);
+
+    TestKernelWithManyRegisters kernel;
+    size_t val(200ul);
+    auto const& bundeledKernel = alpaka::KernelBundle(kernel, val);
+
+    // Test getValidWorkDivForKernel function for 1024x2 threads per grid.
+    auto const workDiv = alpaka::getValidWorkDivForKernel<Acc>(dev, bundeledKernel, Vec{1024, 2}, Vec{1, 1});
```
psychocoderHPC: please change all block, grid and thread sizes to `{Y,X}` in this case `{2,1024}. The indexing in alpaka is **NOT** `X,Y`

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,204 @@
...
+        auto const threadsPerBlock = workDiv.m_blockThreadExtent[0];
+        // Get hard limits
+        auto const kernelFunctionAttributes = alpaka::getFunctionAttributes<Acc>(bundeledKernel);
+        auto const threadsPerBlockLimit = kernelFunctionAttributes.maxThreadsPerBlock;
+        // Compare the workdiv value calculated using the kernel function to the device max limits for threads per
+        // block. The calculated value using kernel function is certainly less than or equal to the device max limits.
+        // For a common Nvidia GPU, expected values for threadsPerBlock and threadsPerBlockLimit are 512 and 1024.
+        // Depending on the GPU or the compiler the test below might fail because threadsPerBlock can be equal to
+        // threadsPerBlockLimit, which is the max device limit.
+        CHECK(threadsPerBlock < static_cast<Idx>(threadsPerBlockLimit));
```
psychocoderHPC, line 111: I am wondering how this can become true? The function `getValidWorkDivForKernel` in line 90 to get the `workDiv` is using internally `kernelFunctionAttributes()` and in line 100 use call `kernelFunctionAttributes()` directly and compare `threadsPerBlock` with `threadsPerBlock`

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,204 @@
...
+    using Vec = alpaka::Vec<Dim, Idx>;
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+    auto const platform = alpaka::Platform<Acc>{};
+    auto const dev = alpaka::getDevByIdx(platform, 0);
+
+    TestKernelWithManyRegisters kernel;
+    auto const& bundeledKernel = alpaka::KernelBundle(kernel, 200ul);
+
+    // Test getValidWorkDivForKernel for 1024 threads per grid
+    auto const workDiv = alpaka::getValidWorkDivForKernel<Acc>(dev, bundeledKernel, Vec{1024}, Vec{1});
```
psychocoderHPC: IMO you need to query the maximum number of threads the divice can handle instead of hard code `1024`. `1024` is only a limit for our GPU in the CI if I remember correctly A100+ support 2048 threads.

psychocoderHPC (changes requested):

include/alpaka/kernel/KernelBundle.hpp
mehmetyusufoglu: Yes, needed: 

- GodBolt:  https://godbolt.org/z/5hEqPP8v8
- Local Compile: `Error: cannot deduce class template arguments
      auto const& bundeledKernel = alpaka::KernelBundle(kernel, val);`

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: This was chosen as an easy to understand example, any number that cannot be fit by local kernel in a block would be ok.  Anyway, as you suggested I will use max device limits.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: Yes, I have just noticed, instead of device properties function properties are used by mistake.

include/alpaka/kernel/KernelFunctionAttributes.hpp
mehmetyusufoglu: For CPU non of them except `maxThreadsPerBlock` is used for now. All values except `maxThreadsPerBlock` are zero for CPU.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: Thanks. I use `props.m_blockThreadCountMax` now.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: Done.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: removed since architecture dependent, thanks.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,224 @@
...
+    }
+    else if constexpr(alpaka::accMatchesTags<Acc, alpaka::TagGpuHipRt>)
+    {
+        // Get calculated threads per block from the workDiv found by examining the kernel function
+        auto const threadsPerBlock = workDiv.m_blockThreadExtent.prod();
+        // Get hard limits
+        auto const threadsPerBlockLimit = props.m_blockThreadCountMax;
+        // Depending on the GPU type or the compiler this test might fail because threadsPerBlock can be less than
+        // threadsPerBlockLimit, which is the max device limit.
+        CHECK(threadsPerBlock == static_cast<Idx>(threadsPerBlockLimit));
```
psychocoderHPC: CI is saying this is failing, IMO you should do the same you did for CUDA. So you can most likely merge the CUDA and HIP if branch.

```
: /builds/hzdr/crp/alpaka/test/unit/workDiv/src/WorkDivForKernelTest.cpp:198: FAILED:
48:   CHECK( threadsPerBlock == static_cast<Idx>(threadsPerBlockLimit) )
48: with expansion:
48:   736 (0x2e0) == 1024 (0x400)
```

psychocoderHPC (changes requested):

include/alpaka/kernel/KernelBundle.hpp
psychocoderHPC: I will keep this point open that we see this if we revisit this PR.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+                          alpaka::TagCpuTbbBlocks>)
+    {
+        // CPU must have only 1 thread per block. In other words, number of blocks is equal to number of threads.
+        CHECK(workDiv == WorkDiv{Vec{threadsPerGridTestValue}, Vec{1}, Vec{1}});
+        // Test a new 1D workdiv. Threads per block can not be larger than 1 for CPU. Hence 2 is not valid.
+        auto const& workDiv1DUsingInitList = WorkDiv{Vec{threadsPerGridTestValue / 2}, Vec{2}, Vec{1}};
+        auto const isWorkDivValidForCpu
+            = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, workDiv1DUsingInitList);
+        CHECK(isWorkDivValidForCpu == false);
+    }
```
psychocoderHPC, line 151: please add a else branch with an exception, to see if we miss an accelerator.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+        auto const threadsPerBlockLimit = props.m_blockThreadCountMax;
+        // Depending on the GPU type or the compiler the test below might fail because threadsPerBlock can be equal to
+        // threadsPerBlockLimit, which is the max device limit.
+        CHECK(threadsPerBlock == static_cast<Idx>(threadsPerBlockLimit));
+    }
+    else if constexpr(alpaka::accMatchesTags<
+                          Acc,
+                          alpaka::TagCpuSerial,
+                          alpaka::TagCpuThreads,
+                          alpaka::TagCpuOmp2Threads,
```
psychocoderHPC, line 117: this should be `alpaka::TagCpuOmp2Blocks`

`alpaka::TagCpuOmp2Threads,` max threads depend on runtime. If the user is setting the environment variable `OMP_NUM_THREADS=1` than one thread is supported.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+        auto const validWorkDiv = WorkDiv{Vec{8, threadsPerGridTestValue / 8}, Vec{1, threadsPerBlock}, Vec{1, 1}};
+        isWorkDivValidForHip = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, validWorkDiv);
+        CHECK(isWorkDivValidForHip == true);
+    }
+    else if constexpr(alpaka::accMatchesTags<
+                          Acc,
+                          alpaka::TagCpuSerial,
+                          alpaka::TagCpuThreads,
+                          alpaka::TagCpuOmp2Threads,
```
psychocoderHPC: must be `alpaka::TagCpuOmp2Blocks`

include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
```diff
@@ -192,6 +194,30 @@ namespace alpaka
...
+            //! determined. Max threads per block is one of the attributes.
+            //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
+            //! fields. For CPU, the field of max threads allowed by the given kernel function for the block is 1.
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+                [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
+                -> alpaka::KernelFunctionAttributes
+            {
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+                kernelFunctionAttributes.maxThreadsPerBlock = 1u;
```
psychocoderHPC: BUG: This is wrong: The number of threads can be `>=1` depending of how the user is setting the environment variable `OMP_NUM_THREADS`
You must query the device properties to get the correct number.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+                          alpaka::TagCpuTbbBlocks>)
+    {
+        // CPU must have only 1 thread per block. In other words, number of blocks is equal to number of threads.
+        CHECK(workDiv == WorkDiv{Vec{8, threadsPerGridTestValue / 8}, Vec{1, 1}, Vec{1, 1}});
+        // Test a new 2D workdiv. Threads per block can not be larger than 1 for CPU. Hence 2x1 threads is not valid.
+        auto const& invalidWorkDiv2D = WorkDiv{Vec{1, 2048}, Vec{1, 2}, Vec{1, 1}};
+        auto const isWorkDivValidForCpu = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, invalidWorkDiv2D);
+        CHECK(isWorkDivValidForCpu == false);
```
psychocoderHPC, line 249: You missed to check if the work division is valid.

psychocoderHPC (changes requested):

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: the `workDiv` variable is already checked for all acc types before the if-else branches.

include/alpaka/kernel/TaskKernelCpuOmp2Threads.hpp
mehmetyusufoglu: I queried device properties for all CPU accs.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: Ok, done. I used 4 types:  ```
alpaka::TagCpuSerial,
                          alpaka::TagCpuOmp2Blocks,
                          alpaka::TagCpuTbbBlocks,
                          alpaka::TagCpuSycl
```

File: test/unit/workDiv/src/WorkDivForKernelTest.cpp (hunk 17 above)
Comment (mehmetyusufoglu) [reply test/unit/workDiv/src/WorkDivForKernelTest.cpp:151]: added, thanks.

File: test/unit/workDiv/src/WorkDivForKernelTest.cpp [hunk 22]
```diff
@@ -0,0 +1,257 @@
...
+        double var0 = 1.0;
+        double var1 = 2.0;
+        double var2 = 3.0;
+
+        // Define many variables and use some calculations in order to prevent compiler optimization and make the
+        // kernel use many registers (around 80 on sm_52). Using many registers per SM decreases the max number of
+        // threads per block while this kernel is being run.
+
+        // TODO: Use function templates to parametrize and shorten the code!
+        double var3 = var2 + fmod(var2, 5);
```
psychocoderHPC: I missed it before, please use `alpaka::math::fmod(acc,...)`

psychocoderHPC (changes requested):

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,257 @@
...
+                          Acc,
+                          alpaka::TagCpuSerial,
+                          alpaka::TagCpuOmp2Blocks,
+                          alpaka::TagCpuTbbBlocks,
+                          alpaka::TagCpuSycl>)
+    {
+        // CPU must have only 1 thread per block. In other words, number of blocks is equal to number of threads.
+        CHECK(workDiv == WorkDiv{Vec{8, threadsPerGridTestValue / 8}, Vec{1, 1}, Vec{1, 1}});
+        // Test a new 2D workdiv. Threads per block can not be larger than 1 for CPU. Hence 2x1 threads is not valid.
+        auto const& invalidWorkDiv2D = WorkDiv{Vec{1, 2048}, Vec{1, 2}, Vec{1, 1}};
```
psychocoderHPC: please do not use `auto const& ` here

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,257 @@
...
+    using Vec = alpaka::Vec<Dim, Idx>;
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+    auto const platform = alpaka::Platform<Acc>{};
+    auto const dev = alpaka::getDevByIdx(platform, 0);
+
+    TestKernelWithManyRegisters kernel;
+    auto const& bundeledKernel = alpaka::KernelBundle(kernel, 200ul);
+
+    // Get hard limits for test
+    auto const& props = alpaka::getAccDevProps<Acc>(dev);
```
psychocoderHPC: I would skip the reference, I know the life time should be ok but IMO it is a common source of mistakes.

`auto const& props` -> `auto & props` 

Some for the line above.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,257 @@
...
+    auto const platform = alpaka::Platform<Acc>{};
+    auto const dev = alpaka::getDevByIdx(platform, 0);
+
+    TestKernelWithManyRegisters kernel;
+    // A random value
+    size_t val(200ul);
+    auto const& bundeledKernel = alpaka::KernelBundle(kernel, val);
+
+    // Get hard limits for test
+    auto const& props = alpaka::getAccDevProps<Acc>(dev);
```
psychocoderHPC: some here.

include/alpaka/kernel/KernelBundle.hpp
```diff
@@ -0,0 +1,65 @@
...
+        {
+        }
+
+        //! The function object type
+        using KernelFn = TKernelFn;
+        //! Tuple type to encapsulate kernel function argument types and argument values
+        using ArgTuple = std::tuple<remove_restrict_t<std::decay_t<TArgs>>...>;
+
+        KernelFn m_kernelFn;
+        ArgTuple m_args;
```
fwyzard, line 44: Shouldn't data members be `private` ?

include/alpaka/kernel/TaskKernelCpuThreads.hpp
```diff
@@ -200,6 +203,35 @@ namespace alpaka
...
+            //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
+            //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+                [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle)
+                -> alpaka::KernelFunctionAttributes
+            {
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+                using Acc = AccCpuThreads<TDim, TIdx>;
+                auto const platformAcc = alpaka::Platform<Acc>{};
```
fwyzard:
```suggestion
                auto const platformAcc = alpaka::PlatformCpu{};
```
We know the type, we don't need to extract it from the accelerator.

include/alpaka/kernel/TaskKernelGenericSycl.hpp
```diff
@@ -276,6 +278,34 @@ namespace alpaka::trait
...
+        //! determined. Max threads per block is one of the attributes.
+        //! \return KernelFunctionAttributes instance. The default version always returns an instance with zero
+        //! fields. For CPU, the field of max threads allowed by kernel function for the block is 1.
+        ALPAKA_NO_HOST_ACC_WARNING
+        ALPAKA_FN_HOST_ACC static auto getFunctionAttributes(
+            [[maybe_unused]] KernelBundle<TKernelFn, TArgs...> const& kernelBundle) -> alpaka::KernelFunctionAttributes
+        {
+            alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+            using Acc = AccGenericSycl<TDim, TIdx>;
+            auto const platformAcc = alpaka::Platform<Acc>{};
```
fwyzard: IMHO this is wrong: the SYCL platform objects have a state, with well-defined device and context objects associated to them. We should not create a new platform with the associated objects every time.

include/alpaka/kernel/TaskKernelGenericSycl.hpp
```diff
@@ -276,6 +278,34 @@ namespace alpaka::trait
...
+            auto const dev = alpaka::getDevByIdx(platformAcc, 0);
```
fwyzard: This is wrong: the different SYCL devices may have different properties, and so return different properties for the kernel.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +388,76 @@ namespace alpaka
...
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+
+                kernelFunctionAttributes.constSizeBytes = funcAttrs.constSizeBytes;
+                kernelFunctionAttributes.localSizeBytes = funcAttrs.localSizeBytes;
+                kernelFunctionAttributes.sharedSizeBytes = funcAttrs.sharedSizeBytes;
+                kernelFunctionAttributes.maxDynamicSharedSizeBytes = funcAttrs.maxDynamicSharedSizeBytes;
+
+                kernelFunctionAttributes.numRegs = funcAttrs.numRegs;
+                kernelFunctionAttributes.asmVersion = funcAttrs.ptxVersion;
+                kernelFunctionAttributes.maxThreadsPerBlock = static_cast<int>(funcAttrs.maxThreadsPerBlock);
```
fwyzard, line 438: These two parameters (`asmVersion` and `maxThreadsPerBlock`) may depend on the current CUDA device.
It would be better to select a device explicitly before calling this function.

include/alpaka/kernel/KernelFunctionAttributes.hpp
fwyzard: For the CPU backends (serial, threads, TBB, OMP2), `maxDynamicSharedSizeBytes` should be set to the value of `alpaka::BlockSharedDynMemberAllocKiB * 1024`, as defined in `alpaka/block/shared/dyn/BlockSharedDynMemberAllocKiB.hpp`.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +388,76 @@ namespace alpaka
...
+#            pragma GCC diagnostic ignored "-Wconditionally-supported"
+#        endif
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
+                    TApi::funcGetAttributes(&funcAttrs, reinterpret_cast<void const*>(kernelName)));
+#        if BOOST_COMP_GNUC
+#            pragma GCC diagnostic pop
+#        endif
+
+                alpaka::KernelFunctionAttributes kernelFunctionAttributes;
+
```
fwyzard:
```suggestion
```

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +388,76 @@ namespace alpaka
...
+                kernelFunctionAttributes.constSizeBytes = funcAttrs.constSizeBytes;
+                kernelFunctionAttributes.localSizeBytes = funcAttrs.localSizeBytes;
+                kernelFunctionAttributes.sharedSizeBytes = funcAttrs.sharedSizeBytes;
+                kernelFunctionAttributes.maxDynamicSharedSizeBytes = funcAttrs.maxDynamicSharedSizeBytes;
+
```
fwyzard:
```suggestion
```

include/alpaka/kernel/Traits.hpp
```diff
@@ -69,6 +70,25 @@ namespace alpaka
...
+        {
+            //! \param kernelBundle The kernel object instance, which includes the kernel function object and it's
+            //! invocation arguments.
+            //! \return KernelFunctionAttributes data structure instance. The default version always returns the
+            //! instance with fields which are set to zero.
+            ALPAKA_NO_HOST_ACC_WARNING
+            ALPAKA_FN_HOST_ACC static auto getFunctionAttributes([[maybe_unused]] TKernelBundle const& kernelBundle)
+                -> alpaka::KernelFunctionAttributes
+            {
+                return alpaka::KernelFunctionAttributes();
```
fwyzard: Maybe this should `throw std::invalid_argument{...}` with a message about not being implemented for the given back-end, instead of returning a default-constructed object.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
fwyzard: Shouldn't this be `<=` instead of `<` ?
Running with the same number of threads as `threadsPerBlockLimit` should be OK.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,253 @@
...
+        // for different GPUs. CHECK(workDiv == WorkDiv{Vec{8, 729444}, Vec{1, 736}, Vec{1, 1}});
+
+        // Get calculated threads per block from the workDiv that was found by examining kernel function
+        auto const threadsPerBlock = workDiv.m_blockThreadExtent.prod();
+        // Get hard limits
+        auto const threadsPerBlockLimit = props.m_blockThreadCountMax;
+
+        // Depending on the GPU type or the compiler the test below might fail because threadsPerBlock can be equal to
+        // threadsPerBlockLimit, which is the max device limit.
+        CHECK(threadsPerBlock < static_cast<Idx>(threadsPerBlockLimit));
```
fwyzard, line 195:
```suggestion
        CHECK(threadsPerBlock <= static_cast<Idx>(threadsPerBlockLimit));
```

Using the maximum value should be fine.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,253 @@
...
+        auto isWorkDivValidForCuda = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, invalidWorkDiv);
+        CHECK(isWorkDivValidForCuda == false);
+
+        auto const validWorkDiv = WorkDiv{Vec{8, threadsPerGridTestValue / 8}, Vec{1, threadsPerBlock}, Vec{1, 1}};
+        isWorkDivValidForCuda = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, validWorkDiv);
+        CHECK(isWorkDivValidForCuda == true);
+    }
+    else if constexpr(alpaka::accMatchesTags<
+                          Acc,
+                          alpaka::TagGpuHipRt,
```
fwyzard, line 209: Why the HIP case is different from the CUDA case ?

fwyzard:
IMHO the current approach is not correct: the values returned by `getFunctionAttributes`, and so in turn the values returned by `getValidWorkDivForKernel`, may depend on the current device, not just on the kernel.

`getValidWorkDivForKernel` and `getFunctionAttributes` should take an extra argument that is the Device to be queried.

include/alpaka/kernel/TaskKernelGenericSycl.hpp
mehmetyusufoglu: Ok you mean there could certainly be more devices than 1. I see.

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -315,6 +357,79 @@ namespace alpaka
...
+        typename TThreadElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>>
+    ALPAKA_FN_HOST auto getValidWorkDivForKernel(
+        [[maybe_unused]] TDev const& dev,
+        TKernelBundle const& kernelBundle,
+        [[maybe_unused]] TGridElemExtent const& gridElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>::ones(),
+        [[maybe_unused]] TThreadElemExtent const& threadElemExtents = Vec<Dim<TAcc>, Idx<TAcc>>::ones(),
+        [[maybe_unused]] bool blockThreadMustDivideGridThreadExtent = true,
+        [[maybe_unused]] GridBlockExtentSubDivRestrictions gridBlockExtentSubDivRestrictions
+        = GridBlockExtentSubDivRestrictions::Unrestricted)
+        -> WorkDivMembers<Dim<TGridElemExtent>, Idx<TGridElemExtent>>
```
psychocoderHPC: maybe it is better to use 

```C++
template<
        typename TAcc,
        typename TDev,
        typename TKernelBundle,
        typename TGridElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>,
        typename TThreadElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>>
    ALPAKA_FN_HOST auto getValidWorkDivForKernel(
        [[maybe_unused]] TDev const& dev,
        TKernelBundle const& kernelBundle,
        [[maybe_unused]] TGridElemExtent const& gridElemExtent = Vec<Dim<TAcc>, Idx<TAcc>>::ones(),
        [[maybe_unused]] TThreadElemExtent const& threadElemExtents = Vec<Dim<TAcc>, Idx<TAcc>>::ones(),
        [[maybe_unused]] bool blockThreadMustDivideGridThreadExtent = true,
        [[maybe_unused]] GridBlockExtentSubDivRestrictions gridBlockExtentSubDivRestrictions
        = GridBlockExtentSubDivRestrictions::Unrestricted)
        -> std::endable_if_t<Dim<TGridElemExtent>::value == Dim<Acc>::value && Dim<TThreadElemExtent>::value == Dim<Acc>::value && std::is_same_v<Idx<TGridElemExtent>, Idx<Acc>> && std::is_same_v<Idx<TThreadElemExtent>, Idx<Acc>>,WorkDivMembers<Dim<TGridElemExtent>, Idx<TGridElemExtent>>>
```

instead of having aserts within the function.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
mehmetyusufoglu: These are function attributes not device properties; I think we need to stick to cuda's choice.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: I wanted to show that function attributes for this specific kernel function is less than the device properties. This is risky in a test but this is the idea of fixing the problem.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: BEcause Hip block size experimentally not less than the max.

include/alpaka/kernel/KernelBundle.hpp
mehmetyusufoglu: I wanted to give user some flexibility (since bundle creates some "states" which is not good for dynamic scenarios; ) May be some variables might need to be updated by the user. But can make private currently will not change anything.

include/alpaka/kernel/KernelFunctionAttributes.hpp
mehmetyusufoglu: Done, thanks, good info.

include/alpaka/kernel/Traits.hpp
mehmetyusufoglu: done, thanks a lot.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: done thanks.

psychocoderHPC:
please rebase against dev to get the fix from #2294 in and remove draft from the PR if it is ready

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -315,12 +357,86 @@ namespace alpaka
...
+            std::is_same_v<Idx<TThreadElemExtent>, Idx<Acc>>,
+            "The idx type of Acc and the idx type of TThreadElemExtent have to be identical!");
+
+        // Get max number of threads per block depending on the kernel function attributes.
+        // For GPU backend; number of registers used by the kernel, local and shared memory usage of the kernel
+        // determines the max number of threads per block. This number could be equal or less than the max number of
+        // threads per block defined by device properties.
+        auto const kernelFunctionAttributes = getFunctionAttributes<Acc>(dev, kernelBundle);
+        auto const threadsPerBlock = kernelFunctionAttributes.maxThreadsPerBlock;
+        // printf("%lu", std::get<0>(kernelBundle.m_args));
```
psychocoderHPC: please remove

mehmetyusufoglu:
> please rebase against dev to get the fix from #2294 in and remove draft from the PR if it is ready

Done, thanks.

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -0,0 +1,261 @@
...
+
+    // Test isValidWorkDivKernel function
+    auto const isValid = alpaka::isValidWorkDivKernel<Acc>(dev, bundeledKernel, workDiv);
+    CHECK(isValid == true);
+
+    if constexpr(alpaka::accMatchesTags<Acc, alpaka::TagGpuCudaRt>)
+    {
+        // Expected valid workdiv for this kernel. These values might change depending on the GPU type and compiler
+        // therefore commented out. Number of registers used by kernel might limit threads per block value differently
+        // for different GPUs. CHECK(workDiv == WorkDiv{Vec{8, 729444}, Vec{1, 736}, Vec{1, 1}});
```
psychocoderHPC: this comment looks outdated with this hard coded values

psychocoderHPC (changes requested):
two tiny things, than it is IMO mergable

test/unit/workDiv/src/WorkDivForKernelTest.cpp
mehmetyusufoglu: Not outdated, but i changed the comment since could be confusing and values would change for different GPU versions.

psychocoderHPC (approved):

PR #2253, opened 2024-03-28
Set up `pre-commit`

chillenzer:
We've had good experience with [pre-commit](https://pre-commit.com) in PIConGPU, so I was asked to add it here, too. In very short, it's a tool that makes git run linters, formatters and other checks before various events (commits and pushes in our case).

The given configuration is mostly taken over from PIConGPU and was adjusted in cooperation with @mehmetyusufoglu. It was described in https://github.com/ComputationalRadiationPhysics/picongpu/pull/4792, https://github.com/ComputationalRadiationPhysics/picongpu/pull/4817, and https://github.com/ComputationalRadiationPhysics/picongpu/pull/4850. For more details, you can also look [here](https://pre-commit.com/hooks.html) and following the links in the `.pre-commit-config.yaml`. I have no strong opinions about the details of the configuration, so any suggestions for changes, additions or deletion are welcome and can probably easily be incorporated.

The PR consists of three types of changes:
* Adding the config file `.pre-commit-config.yaml` (that's all the setup needed).
* Adding documentation about this in CONTRIBUTING.md and the docs.
* Actually running the hooks on all files and fixing the failures.

You might want to review these changes separately.

As a next step, we should also run this in CI which can easily be implement via [pre-commit.ci](https://pre-commit.ci) and adds a few nice features like automatic updating of the hooks and automatic commits with fixes.


psychocoderHPC (approved):

bernhardmgruber (approved):

bernhardmgruber:
This is awesome! I didn't know there were hooks for clang-format, which is great!

There may be an issue though: If you don't run the exactly same version of clang-format locally than the CI, the locally enforced hook will be rejected by the CI, in which case you may have to skip pre-commit with `git commit --no-verify`. I am currious how often we will encounter issues in this regard. But this is no reason to block any of the work here!

In the past: For mixed clang-format versions, if the CI fails the clang-format check, you get a diff that you can apply locally. I needed this a few times in the past, since my local clang-format is usually the latest.

chillenzer:
Thanks for having a look! The above mentioned problem of version mismatch should resolve if we move to [pre-commit.ci](https://pre-commit.ci) for such kinds of checks such that the `.pre-commit-config.yaml` configured version becomes the ground truth. This is (up to technical details) what we did in PIConGPU: https://github.com/ComputationalRadiationPhysics/picongpu/pull/4817

This is next on my list.

fwyzard:
> The above mentioned problem of version mismatch should resolve if we move to [pre-commit.ci](https://pre-commit.ci) for such kinds of checks such that the `.pre-commit-config.yaml` configured version becomes the ground truth.

How does this solve the issue of a developer having a different version of `clang-format` on his or her own system ?
Or having the correct version, but with a different name ?

For example, I have
```
fwyzard@fools:~$ clang-format --version
Ubuntu clang-format version 19.0.0 (++20240407042218+bc8726b16bac-1~exp1~20240407042328.1604)
```
but I also have
```
fwyzard@fools:~$ clang-format-1<TAB><TAB>
clang-format-12  clang-format-14  clang-format-15  clang-format-16  clang-format-17  clang-format-18  clang-format-19  
```

chillenzer:
Well, `pre-commit` installs and uses its own version of `clang-format`. That is true on your local system as well as in CI and these two versions are hence consistent.

So, the worst case scenario you can get is that your IDE is set up to auto-format with a different version such that your commits might fail on first attempt and you have to `git add` the changes made by `pre-commit`. That is admittedly an annoying scenario (I've had that in other contexts) but as soon as your change enter `git` world they are definitely formatted with the correct version. And the setup of a local IDE would probably be considered out of scope for Alpaka.

PS: Theoretically, another scenario would be that you manually run another version frequently but in that case just use `pre-commit run [...]` to run the version installed by `pre-commit`.

fwyzard:
> Well, `pre-commit` installs and uses its own version of `clang-format`. That is true on your local system as well as in CI and these two versions are hence consistent.

I see.

Sorry, no, not on my system.

PR #2263, opened 2024-05-03
Implement accelerator traits for single vs multi-threads per block

fwyzard:
`alpaka::isSingleThreadAccelerator<T>` evaluates to `true` if the given type is an accelerator that supports only one thread per block, and to to `false` if it supports multiple threads per block.

`alpaka::isMultiThreadAccelerator<T>` evaluates to `true` if the given type is an accelerator that supports multiple threads per block, and to to `false` if it supports only one thread per block.

If the given type is not an accelerator, the constants are not defined.

Implement a unit test for `alpaka::isSingleThreadAccelerator<T>` and `alpaka::isMultiThreadAccelerator<T>`.

### DIFF
--- a/include/alpaka/acc/AccCpuOmp2Blocks.hpp
+++ b/include/alpaka/acc/AccCpuOmp2Blocks.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Axel Huebl, Benjamin Worpitz, René Widera, Jan Stephan, Bernhard Manfred Gruber
+/* Copyright 2024 Axel Huebl, Benjamin Worpitz, René Widera, Jan Stephan, Bernhard Manfred Gruber, Andrea Bocci
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -114,6 +114,18 @@ namespace alpaka
         struct AccType<AccCpuOmp2Blocks<TDim, TIdx>>
         {
             using type = AccCpuOmp2Blocks<TDim, TIdx>;
+        };
+
+        //! The CPU OpenMP 2.0 block single thread accelerator type trait specialization.
+        template<typename TDim, typename TIdx>
+        struct IsSingleThreadAcc<AccCpuOmp2Blocks<TDim, TIdx>> : std::true_type
+        {
+        };
+
+        //! The CPU OpenMP 2.0 block multi thread accelerator type trait specialization.
+        template<typename TDim, typename TIdx>
+        struct IsMultiThreadAcc<AccCpuOmp2Blocks<TDim, TIdx>> : std::false_type
+        {
         };
 
         //! The CPU OpenMP 2.0 block accelerator device properties get trait specialization.
--- a/include/alpaka/acc/Traits.hpp
+++ b/include/alpaka/acc/Traits.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2022 Benjamin Worpitz, Bernhard Manfred Gruber
+/* Copyright 2024 Benjamin Worpitz, Bernhard Manfred Gruber, Andrea Bocci
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -36,6 +36,26 @@ namespace alpaka
         template<typename T, typename TSfinae = void>
         struct AccType;
 
+        //! The single thread accelerator trait.
+        //!
+        //! If TAcc is an accelerator that supports only a single thread per block, inherit from std::true_type.
+        //! If TAcc is not an accelerator, or an accelerator that supports multiple threads per block, inherit from
+        //! std::false_type.
+        template<typename TAcc, typename TSfinae = void>
+        struct IsSingleThreadAcc : std::false_type
+        {
+        };
+
+        //! The multi thread accelerator trait.
+        //!
+        //! If TAcc is an accelerator that supports multiple threads per block, inherit from std::true_type.
+        //! If TAcc is not an accelerator, or an accelerator that supports only a single thread per block, inherit from
+        //! std::false_type.
+        template<typename TAcc, typename TSfinae = void>
+        struct IsMultiThreadAcc : std::false_type
+        {
+        };
+
         //! The device properties get trait.
         template<typename TAcc, typename TSfinae = void>
         struct GetAccDevProps;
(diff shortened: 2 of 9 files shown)


fwyzard:
@psychocoderHPC I use these traits inside the grid and block strided loops, to avoid the thread-level loop if the accelerator supports only one loop.

Comments, and suggestions for better names are welcome :-)

fwyzard:
The failing test has a few instances of
```
22: D:\a\alpaka\alpaka\test\unit\acc\src\AccTraitTest.cpp(29): FAILED:
22:   REQUIRE( devProps.m_blockThreadCountMax > 1 )
22: with expansion:
22:   1 > 1
```

Is there a way to see what back-end is giving problems with the check ?

fwyzard:
Temporarily adding log messages to see what accelerators are failing.

fwyzard:
`AccCpuOmp2Threads` reports a single thread per block on the Windows tests.

include/alpaka/acc/Traits.hpp
```diff
@@ -36,6 +36,22 @@ namespace alpaka
         template<typename T, typename TSfinae = void>
         struct AccType;
 
+        //! The single thread accelerator trait.
+        //!
+        //! If TAcc is a accelerator that supports only a single thread per block, inherit from std::true_type.
```
mehmetyusufoglu:
```suggestion
        //! If TAcc is an accelerator that supports only a single thread per block, inherit from std::true_type.
```

mehmetyusufoglu:
> @psychocoderHPC I use these traits inside the grid and block strided loops, to avoid the thread-level loop if the accelerator supports only one loop.
> 
> Comments, and suggestions for better names are welcome :-)



> `alpaka::isSingleThreadAccelerator<T>` evaluates to `true` if the given type is an accelerator that supports only one thread per block, and to to `false` if it supports multiple threads per block.
> 
> `alpaka::isMultiThreadAccelerator<T>` evaluates to `true` if the given type is an accelerator that supports multiple threads per block, and to to `false` if it supports only one thread per block.
> 
> If the given type is not an accelerator, the constants are not defined.
> 
> Implement a unit test for `alpaka::isSingleThreadAccelerator<T>` and `alpaka::isMultiThreadAccelerator<T>`.

Only one of the 2 traits above is enough? One is the complement of the other?

fwyzard:
> Only one of the 2 traits above is enough? One is the complement of the other?

For an Accelerator, yes.

For an arbitrary type, no: a type can also not be an accelerator at all.

include/alpaka/acc/Traits.hpp
fwyzard: Thanks.

psychocoderHPC:
> > Only one of the 2 traits above is enough? One is the complement of the other?
> 
> For an Accelerator, yes.
> 
> For an arbitrary type, no: a type can also not be an accelerator at all.

Could you please give an example where both traits are required? The trait documentation is saying that the traits are only defined for accelerators. I try to find an example where one is not the complement of the other.

psychocoderHPC:
> `AccCpuOmp2Threads` reports a single thread per block on the Windows tests.

https://github.com/alpaka-group/alpaka/blob/1b8146ef300259c1206a3a56341be3e64d96515d/include/alpaka/acc/AccCpuOmp2Threads.hpp#L131-L133 if the environment variable `OMP_NUM_THREDAS` is set to `1` what is sometimes the default for an OS this backend is limiting the number of threads per block.

This is interesting and something I had not in mind.

fwyzard:
> > > Only one of the 2 traits above is enough? One is the complement of the other?
> > 
> > 
> > For an Accelerator, yes.
> > For an arbitrary type, no: a type can also not be an accelerator at all.
> 
> Could you please give an example where both traits are required? The trait documentation is saying that the traits are only defined for accelerators. I try to find an example where one is not the complement of the other.

You are right - after trying to use them in an example, I realise that they should be `false` for non-accelerator types, rather than undefined.

psychocoderHPC:
> You are right - after trying to use them in an example, I realise that they should be `false` for non-accelerator types, rather than undefined.

I expect that you update this PR and change the default to false instead of undefined. The question in the toom is still if both traits are useful or if a single trait is enough.

fwyzard:
> I expect that you update this PR and change the default to false instead of undefined. 

Indeed, done.

> The question in the room is still if both traits are useful or if a single trait is enough.

Now they are.

For example, if only `isSingleThreadAccelerator` is defined, and one wants to check for the other case, one needs to use:
```c++
if constexpr (alpaka::isAccelerator<TAcc> and not alpaka::isSingleThreadAccelerator<TAcc>) {
  ...
}
```

If both are defined, one can directly use
```c++
if constexpr (alpaka::isMultiThreadAccelerator<TAcc>) {
  ...
}
```

That said, either one can be derived from the other plus existing traits, so if there is a preference to have only one, I can remove `isMultiThreadAccelerator`.

fwyzard:
Could you let me know
  - if the current approach is OK;
  - if I should remove `isMultiThreadAccelerator` or keep it;
  - if I should remove the specialisations on `IsSingleThreadAccelerator` and `IsMultiThreadAccelerator` that inherit from `false_type` and let them use the general implementation ?
  
In any case I will squash the 3rd commit into the 1st.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -116,6 +116,18 @@ namespace alpaka
             using type = AccCpuOmp2Blocks<TDim, TIdx>;
         };
 
+        //! The CPU OpenMP 2.0 block single thread accelerator type trait specialization.
+        template<typename TDim, typename TIdx>
+        struct IsSingleThreadAccelerator<AccCpuOmp2Blocks<TDim, TIdx>> : std::true_type
```
psychocoderHPC: Could you please rename the traits to `isMultiThreadAcc` to follow the typical naming of alpaka.

psychocoderHPC (changes requested):
Discussed in the dev meeting. We are fine with both traits but I like to change `Accelerator` to `Acc`

include/alpaka/acc/AccCpuOmp2Blocks.hpp
fwyzard: Will do.

fwyzard: Done

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -116,6 +116,18 @@ namespace alpaka
...
 
+        //! The CPU OpenMP 2.0 block single thread accelerator type trait specialization.
+        template<typename TDim, typename TIdx>
+        struct IsSingleThreadAcc<AccCpuOmp2Blocks<TDim, TIdx>> : std::true_type
+        {
+        };
+
+        //! The CPU OpenMP 2.0 block multi thread accelerator type trait specialization.
+        template<typename TDim, typename TIdx>
+        struct IsMultiThreadAcc<AccCpuOmp2Blocks<TDim, TIdx>> : std::false_type
```
psychocoderHPC, line 127: in principel we do not need to define this here because the default is false.

psychocoderHPC (approved):

SimeonEhrig (approved):

include/alpaka/acc/AccCpuOmp2Blocks.hpp
psychocoderHPC: offline discussed: we decided to keep these specializations to be explicit.

PR #2269, opened 2024-05-13
implement alpaka::meta::isList, alpaka::meta::ToList and alpaka::meta::toTuple

SimeonEhrig:
- `alpaka::meta::isTuple`: checks if a given type is a `std::tuple` or not
- `alpaka::meta::toTuple`: pack a arbitrary number of types in a `std::tuple`. If the given type is a `std::tuple` the created type is the same.

- [x] check if it working with PR #2180 , because the function is used there

### DIFF
--- a/include/alpaka/meta/TypeListOps.hpp
+++ b/include/alpaka/meta/TypeListOps.hpp
@@ -4,6 +4,7 @@
 
 #pragma once
 
+#include <tuple>
 #include <type_traits>
 
 namespace alpaka::meta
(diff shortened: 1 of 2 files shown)


psychocoderHPC:
@SimeonEhrig please update the description of this PR, `isTuple`,... is not valid anymore.

include/alpaka/meta/TypeListOps.hpp
```diff
@@ -35,4 +36,60 @@ namespace alpaka::meta
...
+        struct ToListImpl<TListType, TList, std::enable_if_t<alpaka::meta::isList<TList>>>
+        {
+            using type = TList;
+        };
+    } // namespace detail
+
+    //! \brief Takes an arbitrary number of types (T) and creates a type list of type TListType with the types (T). If
+    //! T is a single template parameter and it satisfies alpaka::meta::isList, the type of the structure is T (no type
+    //! change).
+    //! \tparam TListType type of the created list
```
psychocoderHPC, line 73: maybe add `e.g. std::tuple` as an example

include/alpaka/meta/TypeListOps.hpp
```diff
@@ -35,4 +36,60 @@ namespace alpaka::meta
...
+    template<template<typename...> class TListType, typename T, typename... Ts>
+    struct ToList<TListType, T, Ts...>
+    {
+        using type = TListType<T, Ts...>;
+    };
+
+    //! \brief If T is a single argument and a type list (fullfil alpaka::meta::isList), the return type is T.
+    //! Otherwise, std::tuple is returned with T types as template parameters.
+    template<typename... T>
+    using toTuple = typename ToList<std::tuple, T...>::type;
```
psychocoderHPC: Must be `ToTuple` because it creates the type only not the instance.

psychocoderHPC (changes requested):
ONly one name change is required and maybe a small comment extension.

psychocoderHPC (approved):

PR #2270, opened 2024-05-16
refactor template order `allocMappedBuf`

SimeonEhrig:
Move template TPlatform as the last template. There is no need to provide the platform template signature if we pass the platform as an instance.

follow up of #2162

### DIFF
--- a/docs/source/dev/backends.rst
+++ b/docs/source/dev/backends.rst
@@ -195,7 +195,7 @@ Depending on the cmake argument ``ALPAKA_ACC_GPU_CUDA_ONLY_MODE`` the function a
     +----------+-------------------------------------+
     | CUDA     | alpaka                              |
     +==========+=====================================+
-    | ``dim3`` | ``alpaka::Vec< TDim, TVal >``  |
+    | ``dim3`` | ``alpaka::Vec< TDim, TVal >``       |
     +----------+-------------------------------------+
 
 
@@ -353,7 +353,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaFreeAsync              | n/a (automatic memory management with reference counted memory handles)                    |
     +----------------------------+--------------------------------------------------------------------------------------------+
-    | cudaFreeHost               | n/a (automatic memory management with reference counted memory handles)                                                                                       |
+    | cudaFreeHost               | n/a (automatic memory management with reference counted memory handles)                    |
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaFreeMipmappedArray     | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
@@ -363,7 +363,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaGetSymbolSize          | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
-    | cudaHostAlloc              | alpaka::allocMappedBuf<TPlatform, TElement>(host, extents) 1D, 2D, 3D supported!           |
+    | cudaHostAlloc              | alpaka::allocMappedBuf<TElement, TIdx>(host, platform, extents) 1D, 2D, 3D supported! [1]  |
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaHostGetDevicePointer   | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
@@ -383,7 +383,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaMallocAsync            | alpaka::allocAsyncBuf<TElement>(queue, extents1D)                                          |
     +----------------------------+--------------------------------------------------------------------------------------------+
-    | cudaMallocHost             | alpaka::allocMappedBuf<TPlatform, TElement>(host, extents) 1D, 2D, 3D supported!           |
+    | cudaMallocHost             | alpaka::allocMappedBuf<TElement, TIdx>(host, platform, extents) 1D, 2D, 3D supported! [1]  |
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaMallocManaged          | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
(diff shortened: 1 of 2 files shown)


psychocoderHPC (dismissed):

psychocoderHPC:
please update https://github.com/alpaka-group/alpaka/blob/8dbe7c42c647e4b67f22c10a39c36b160f654ecf/docs/source/dev/backends.rst?plain=1#L386 too.
It was also forgotten during the change from platform type to platform instance.

Maybe we should use here `allocMappedBufIfSupported()`

psychocoderHPC:
please update https://github.com/alpaka-group/alpaka/blob/8dbe7c42c647e4b67f22c10a39c36b160f654ecf/docs/source/dev/backends.rst?plain=1#L366 too


```c++
alpaka::allocMappedBuf<TElement>(host, platform, extents) 1D, 2D, 3D supported!  
```

SimeonEhrig:
> please update
> 
> https://github.com/alpaka-group/alpaka/blob/8dbe7c42c647e4b67f22c10a39c36b160f654ecf/docs/source/dev/backends.rst?plain=1#L366
> too
> 
> ```c++
> alpaka::allocMappedBuf<TElement>(host, platform, extents) 1D, 2D, 3D supported!  
> ```

I `alpaka::allocMappedBuf<TElement, TIdx>(host, platform, extents) 1D, 2D, 3D supported!` should be correct.

docs/source/dev/backends.rst
```diff
@@ -363,7 +363,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaGetSymbolSize          | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
-    | cudaHostAlloc              | alpaka::allocMappedBuf<TPlatform, TElement>(host, extents) 1D, 2D, 3D supported!           |
+    | cudaHostAlloc              | alpaka::allocMappedBuf<TElement, TIdx>(host, extents) 1D, 2D, 3D supported!                |
```
psychocoderHPC: Sry my comment was wrong this should be `allocMappedBufIfSupported`. There is no strong need that this memory is mapped.
`allocMappedBuf` is not available for any accelerator or device.

docs/source/dev/backends.rst
```diff
@@ -363,7 +363,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaGetSymbolSize          | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
-    | cudaHostAlloc              | alpaka::allocMappedBuf<TPlatform, TElement>(host, extents) 1D, 2D, 3D supported!           |
+    | cudaHostAlloc              | alpaka::allocMappedBufIfSupported<TElement, TIdx>(host, extents) 1D, 2D, 3D supported!     |
```
fwyzard: This too should be just
```suggestion
    | cudaHostAlloc              | alpaka::allocMappedBuf<TElement, TIdx>(host, extents) 1D, 2D, 3D supported!                |
```

psychocoderHPC: I do not think that `allocMappedBuf` is correct here. `allocMappedBuf` can fail if mapped memory is not supported for the current device, so the code will not run with all devices. `allocMappedBufIfSupported` will be safe for all kinds of devices.

The CUDA API function `cudaHostAlloc` is pinning memory but does not guarantee that the memory is visible on the GPU device, so the best fitting alpaka equivalent is `allocMappedBufIfSupported`

fwyzard: > The CUDA API function `cudaHostAlloc` is pinning memory but does not guarantee that the memory is visible on the GPU device, so the best fitting alpaka equivalent is `allocMappedBufIfSupported`

It does guarantee that. From the [CUDA Runtime API](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gb65da58f444e7230d3322b6126bb4902):

> `__host__ cudaError_t cudaHostAlloc ( void** pHost, size_t size, unsigned int  flags )`
>
> #### Description
>
> Allocates size bytes of host memory that is page-locked and accessible to the device.

fwyzard: In any case, [`cudaMallocHost()`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__HIGHLEVEL.html#group__CUDART__HIGHLEVEL_1gd5c991beb38e2b8419f50285707ae87e) and [`cudaHostAlloc()`](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1gb65da58f444e7230d3322b6126bb4902) are identical, so they should map to the same functionality in alpaka.

psychocoderHPC: Ohh I was not aware that `cudaMallocHost` is the C++-API equivalent of `cudaHostAlloc`. I used always `cudaHostAlloc` with flags to create mapped memory.

In this case, IMO both should be equal. The problem I had is that for both CUDA methods, you can configure the behavior via a flag where we have only named functions. Maybe we should add  `allocMappedBufIfSupported`, ` allocMappedBuf`, and `allocMappedBuf` into the documentation and the user must decide what fitts best for their problem because the method used for the replacement depends on the flags used in the native CUDA function call.

SimeonEhrig: I changed it back to `alpaka::allocMappedBuf<TPlatform, TElement>(host, extents)` and wrote a for foot note for `alpaka::allocMappedBufIfSupported<TPlatform, TElement>(host, extents)`

docs/source/dev/backends.rst
```diff
@@ -363,7 +363,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +----------------------------+--------------------------------------------------------------------------------------------+
     | cudaGetSymbolSize          | --                                                                                         |
     +----------------------------+--------------------------------------------------------------------------------------------+
-    | cudaHostAlloc              | alpaka::allocMappedBuf<TPlatform, TElement>(host, extents) 1D, 2D, 3D supported!           |
+    | cudaHostAlloc              | alpaka::allocMappedBuf<TElement, TIdx>(host, extents) 1D, 2D, 3D supported! [1]            |
```
fwyzard:
```suggestion
    | cudaHostAlloc              | alpaka::allocMappedBuf<TElement, TIdx>(host, platform, extents) 1D, 2D, 3D supported! [1]  |
```

fwyzard: also below

fwyzard (approved):

psychocoderHPC (approved):

PR #2271, opened 2024-05-17
set and read the alpaka buffer directly in examples

mehmetyusufoglu:
- Alpaka buffer items can be set and read directly, Casting the buffer to a pointer is not needed. Some examples were misleading for a novice user, I will use examples in a demo so I wanted to fix.

- All `alpaka::getPtrNative` calls replaced with `std::data` in all examples and all tests under `test/integ`, at a unit test and benchmarks


example/monteCarloIntegration/src/monteCarloIntegration.cpp
```diff
@@ -105,12 +105,11 @@ auto main() -> int
 
     // Setup buffer.
     BufHost bufHost{alpaka::allocBuf<uint32_t, Idx>(devHost, extent)};
-    uint32_t* const ptrBufHost{alpaka::getPtrNative(bufHost)};
     BufAcc bufAcc{alpaka::allocBuf<uint32_t, Idx>(devAcc, extent)};
     uint32_t* const ptrBufAcc{alpaka::getPtrNative(bufAcc)};
```
fwyzard: if you like it better, you could also use
```suggestion
    uint32_t* const ptrBufAcc = std::data(bufAcc);
```
or
```suggestion
    uint32_t* const ptrBufAcc = bufAcc.data();
```

psychocoderHPC: I like this, it looks more like STL and explicitly says that we like to use the buffer as contiguous array.

mehmetyusufoglu: Thanks, great. I replaced all `alpaka::getPtrNative `calls with` std::data `in all examples and tests under `test/integ`. ( in 2 different commits. )

benchmarks/babelstream/src/AlpakaStream.cpp
```diff
@@ -51,16 +51,8 @@ void AlpakaStream<T>::init_arrays(T initA, T initB, T initC)
...
-        workdiv,
-        InitKernel{},
-        alpaka::getPtrNative(d_a),
-        alpaka::getPtrNative(d_b),
-        alpaka::getPtrNative(d_c),
-        initA,
-        initB,
-        initC);
+    alpaka::exec<
+        Acc>(queue, workdiv, InitKernel{}, std::data(d_a), std::data(d_b), std::data(d_c), initA, initB, initC);
```
mehmetyusufoglu, line 55: clang-format changes the call format since line is shorter I think

SimeonEhrig (dismissed):

SimeonEhrig (approved):

PR #2273, opened 2024-05-21
fix alpaka_add_library relocatable device code

ichinii:
alpaka_add_library did not enable relocatable device code, because the wrong identifier for the library name was used.

I changed the identifier to be consistent with contents of addExecutable.cmake


SimeonEhrig (approved):

SimeonEhrig:
@ichinii Thanks for fixing the bug. I found out, that you CI does not cover relocatable device code for `alpaka_add_library`.

PR #2275, opened 2024-05-21
Document different alpaka terms and it relationships

SimeonEhrig:
During my work on PR #2180 I had some trouble to add the memory visibility on the correct concepts. Therefore I had a offline discussion with @psychocoderHPC and started to write down some terms, their meanings and their relationships.

It is pretty hard to document this stuff, therefore I think my current documentation is not complete and misses precisely documentation. Please comment it.


docs/source/basic/terms.rst
```diff
@@ -0,0 +1,57 @@
...
+- Depending on the platform, it also contains a runtime context.
+- A ``platform`` has a least one device but it can also has many device.
+- Each ``platform`` can be used with N ``accelerator``. ``platforms`` and ``accelerator`` cannot freely combined. An ``accelerator`` supports only a specific ``platform``.
+ 
+Device
+------
+
+- A ``device`` represent a compute unit, such as a CPU or a GPU.
+- Each ``device`` is bounded to a specific ``platform``.
+- Each ``device`` can have N ``queues``.
```
fwyzard: In general, I would write "any number" instead of "N":
```suggestion
- Each ``device`` can have any number of ``queues``.
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,57 @@
...
+
+Accelerator
+-----------
+
+- A ``accelerator`` is a index mapping function. It distributes the index space to the chunks. The ``accelerator`` maps a continues index space to a blocked index domain decomposition.
```
fwyzard: It is much more than that.
For example, the OpenMP and TBB accelerators implement different kind of parallelism for the Cpu device.

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,57 @@
...
+- It is not allowed to create an instance of an ``accelerator``.
+- A ``accelerator`` is bounded to a specific ``platform``.
+
+Queue
+-----
+
+- Stores operations which should be executed on a ``device``.
+- Operations can be ``TaskKernels``, ``Events``, ``Sets`` and ``Copies``.
```
fwyzard: Also host tasks (that are not kernels).

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,57 @@
...
+Accelerator
+-----------
+
+- A ``accelerator`` is a index mapping function. It distributes the index space to the chunks. The ``accelerator`` maps a continues index space to a blocked index domain decomposition.
+- It is not allowed to create an instance of an ``accelerator``.
+- A ``accelerator`` is bounded to a specific ``platform``.
+
+Queue
+-----
+
```
fwyzard, line 40: I would add that a queue can be either "synchronous" (blocking) or "asynchronous" (non-blocking), and that this may determine how the work is submitted to the device, and if a synchronisation is performed after each operation.

docs/source/basic/terms.rst
fwyzard: Memory allocations can also be submitted to a queue.

docs/source/basic/terms.rst
SimeonEhrig: Yes, I know. Accelerator is pretty hard to explain. Actual I only wrote down, what @psychocoderHPC told me, because my personal description is to inprecisely.

docs/source/basic/terms.rst
SimeonEhrig: Is host task also a trait/concept in alpaka? I never heard about this.

fwyzard: Mhm, I'm not sure about a concept.

I think you can submit anything that can be called with no arguments:
```c++
#include <chrono>
#include <iostream>
using namespace std::chrono_literals;

#include <alpaka/alpaka.hpp>

int main(void) {

  using Platform = alpaka::PlatformCudaRt;
#else
  using Platform = alpaka::PlatformCpu;
#endif  // ALPAKA_ACC_GPU_CUDA_ENABLED
  Platform platform;

  // use the first available device
  using Device = alpaka::Dev<Platform>;
  Device device = alpaka::getDevByIdx(platform, 0u);

  // create an asynchronous queue
  using Queue = alpaka::Queue<Device, alpaka::NonBlocking>;
  Queue queue{device};

  std::cout << "before submitting\n";

  // submit a host task to an asynchronous queue
  alpaka::enqueue(queue, [](){
    // asynchronous host task executed from a device queue
    std::cout << "async hello world from " << alpaka::core::demangled<Device> << '\n';
  });

  // sleep for 1s
  std::this_thread::sleep_for(1000ms);

  std::cout << "after submitting, before waiting\n";

  // wait for the asynchronous task to complete
  alpaka::wait(queue);

  std::cout << "after waiting\n";

  return 0;
}
```

Compiled with
```bash
$ g++ -std=c++17 -O2 -DALPAKA_ACC_GPU_CUDA_ENABLED -DALPAKA_HOST_ONLY -I/home/fwyzard/src/alpaka-group/alpaka/include -I/usr/local/cuda/include test.cc -L/usr/local/cuda/lib64 -lcudart -lcuda -lrt -o test
```
should print
```
$ ./test 
before submitting
async hello world from alpaka::DevUniformCudaHipRt<alpaka::ApiCudaRt>
after submitting, before waiting
after waiting
```

Compiled with
```bash
$ g++ -std=c++17 -O2 -DALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED -I/home/fwyzard/src/alpaka-group/alpaka/include test.cc -L/usr/local/cuda/lib64 -lrt -o test
```
should print
```
$ ./test 
before submitting
async hello world from alpaka::DevCpu
after submitting, before waiting
after waiting
```

SimeonEhrig: I was not aware, that is possible. I use a task kernel and a cpu device for it. But I found the feature in the queue tests.

I will add it. Looks like my documentation makes sense ;-)

fwyzard: > I use a task kernel and a cpu device for it.

That also works. 

The point of using a host task with a device queue is to run the host code only after the device operations are complete.

A similar functionality in CUDA is provided by `cudaLaunchHostFunc`, but the alpaka version does not have the limitation that it cannot call other device API functions.

docs/source/basic/terms.rst
SimeonEhrig: I would discuss the documentation of accelerator in the next alpaka VC. For me, this is a complicated subject.

mehmetyusufoglu:
The attached image is great, don't we need to show a relation from accelerator to device since in the code we use accelerator type not the platform get the device?

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,65 @@
...
+- A ``device`` represent a compute unit, such as a CPU or a GPU.
+- Each ``device`` is bounded to a specific ``platform``.
+- Each ``device`` can have any number of ``queues``.
+
+Accelerator
+-----------
+
+- A ``accelerator`` is a index mapping function. It distributes the index space to the chunks. The ``accelerator`` maps a continues index space to a blocked index domain decomposition.
+- It is not allowed to create an instance of an ``accelerator``.
+- A ``accelerator`` is bounded to a specific ``platform``.
```
mehmetyusufoglu, line 28: This could be important. "Accelerator makes the kernel device-agnostic". Accelerator maps device side functions and provides access to shared memory ?

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,65 @@
...
+- Depending on the platform, it also contains a runtime context.
+- A ``platform`` has a least one device but it can also has many device.
+- Each ``platform`` can be used with any number of ``accelerator``. ``platforms`` and ``accelerator`` cannot freely combined. An ``accelerator`` supports only a specific ``platform``.
+ 
+Device
+------
+
+- A ``device`` represent a compute unit, such as a CPU or a GPU.
+- Each ``device`` is bounded to a specific ``platform``.
+- Each ``device`` can have any number of ``queues``.
```
mehmetyusufoglu: Device can be attached to data (in an alpaka buf or view) which makes easy memcpy calls.

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,65 @@
...
+- It is not allowed to create an instance of an ``accelerator``.
+- A ``accelerator`` is bounded to a specific ``platform``.
+
+Queue
+-----
+
+- Stores operations which should be executed on a ``device``.
+- Operations can be ``TaskKernels``, ``HostTasks``, ``Events``, ``Sets`` and ``Copies``.
+- Each ``queue`` is bounded to a specific ``device``.
+- A Queue can be ``Blocking`` (host thread is waiting for finishing the API call) or ``NonBlocking`` (host thread continues after calling the API independent if the call finished or not).
```
mehmetyusufoglu: [nit] Blocking queues only block the caller thread of exec or enqueue, does not block other queues

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,65 @@
...
+- All operations in a queue will be executed sequentiell.
```
mehmetyusufoglu:
```suggestion
- All operations in a queue will be executed sequentially.
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,65 @@
...
+-----
+
+- A ``event`` is a marker in the ``queue``.
+- ``events`` can be used to describe dependencies between different ``queues``.
+- A ``event`` allows to wait until a specific time point.
+
+Set
+---
+
+- A ``Set`` set byte wise a memory to a specific value.
```
mehmetyusufoglu:
```suggestion
- A ``Set`` sets a memory to a specific value byte-wise.
```

fwyzard:
> in the code we use accelerator type not the platform get the device?

This is actually the wrong way of using accelerators, platforms and devices.

docs/source/basic/terms.rst
SimeonEhrig: Okay, I see why the functionality makes sense.

docs/source/basic/terms.rst
SimeonEhrig: There is also a relation between accelerator and TaskKernel, which I didn't add it. And I'm not sure if I should add it in this diagram or in second. Same for your suggestion.

SimeonEhrig:
> > in the code we use accelerator type not the platform get the device?
> 
> This is actually the wrong way of using accelerators, platforms and devices.

This actual on of the motivations of this documentation. Our examples are written in an accelerator centered way, but this actual not correct. And the idea that the accelerator is the central element causes some problems, if you want to implement new features.

fwyzard:
> Our examples are written in an accelerator centered way, but this actual not correct.

It does make sense to write the unit tests like that: we want to test the accelerator, so we need to use the corresponding device.

But in general, it makes more sense for an application to start from the platform, enumerate the available devices, pick one, create a queue, and choose the accelerator that best matches that device and the problem at hand.

SimeonEhrig:
@alpaka-group/alpaka-maintainers I'm not able to finish the PR because I do not really understand the alpaka concepts. Somebody can overtake the PR. If the PR is not finished until next release it will be closed without merge.

psychocoderHPC:
@alpaka-group/alpaka-maintainers Since @SimeonEhrig asked for help I changed some points which IMO are important. 
My goal is not to have the perfect description because what @SimeonEhrig did is already a huge improvement compared to the current state.

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+Work division
+-------------
+
+- Describes the domain decomposition of a contiguous N-dimensional index domain in ``blocks``, ``threads`` within a ``block``, and ``elements`` per ``thread``.
+- A ``work division`` has limitations depending on the ``kernel`` function and ``accelerator``.
+
+Accelerator
+-----------
+
+- Describes "how" a kernel work division is mapped to device threads.
```
mehmetyusufoglu: We give work division separately to the kernel, not use the workdiv from the acc?

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+------
+
+- A ``device`` represent a compute unit, such as a CPU or a GPU.
+- Each ``device`` is bounded to a specific ``platform``.
+- Each ``device`` can be used by many specific ``accelerators``.
+
+Work division
+-------------
+
+- Describes the domain decomposition of a contiguous N-dimensional index domain in ``blocks``, ``threads`` within a ``block``, and ``elements`` per ``thread``.
```
SimeonEhrig:
```suggestion
- Describes the domain decomposition of a contiguous N-dimensional index domain in ``blocks``, ``threads`` and ``elements``. A ``block`` contains one ore more ``threads`` and a ``thread`` one or more ``elements``.
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+- A ``work division`` has limitations depending on the ``kernel`` function and ``accelerator``.
+
+Accelerator
+-----------
+
+- Describes "how" a kernel work division is mapped to device threads.
+  - N-dimensional work divisions (1D, 2D, 3D) are supported.
+  - Holds implementations of shared memory, atomic operations, math operations etc.
+- ``Accelerators`` are instantiated only when a kernel is executed, and can only be accessed in device code.
+  - Each device function can (should) be templated on the accelerator type, and take an accelerator as its first argument.
```
SimeonEhrig: Looks like the indentation is not correct. It does not render as sub items.

docs/source/basic/terms.rst
psychocoderHPC: The user is giving the work division to the kernel exec call but at the end this is only the interface. I do not like describe the interface in words, my intention is to describe what different concepts do.
In this case the accelerator performs the mapping from indices to physical threads.

docs/source/basic/terms.rst
psychocoderHPC: fixed

docs/source/basic/terms.rst
psychocoderHPC: fixed

psychocoderHPC:
@alpaka-group/alpaka-maintainers ping

SimeonEhrig:
@psychocoderHPC I can only approve it only via text because I'm the author of the PR.

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+------
+
+* A ``device`` represent a compute unit, such as a CPU or a GPU.
+* Each ``device`` is bounded to a specific ``platform``.
+* Each ``device`` can be used by many specific ``accelerators``.
+
+Work division
+-------------
+
+* Describes the domain decomposition of a contiguous N-dimensional index domain in ``blocks``, ``threads`` and ``elements``. A ``block`` contains one or more ``threads`` and a ``thread`` process one or more ``elements``.
```
fwyzard, line 23:
```suggestion
* Describes the domain decomposition of a contiguous N-dimensional index domain in ``blocks``, ``threads`` and ``elements``. A ``block`` contains one or more ``threads`` and each ``thread`` processes one or more ``elements``.
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+    * The accelerator object can be used to extract the ``work division`` and indices of the current block and thread.
+    * The accelerator type can be used to implement per-accelerator behaviours.
+* An ``accelerator`` is bounded to a specific ``platform``.
+
+Queue
+-----
+
+* Stores tasks which should be executed on a ``device``.
+* Operations can be ``TaskKernels``, ``HostTasks``, ``Events``, ``Sets`` and ``Copies``.
+* A Queue can be ``Blocking`` (host thread is waiting for finishing the API call) or ``NonBlocking`` (host thread continues after calling the API independent if the call finished or not).
```
fwyzard, line 43:
```suggestion
* A Queue can be ``Blocking`` (the host thread waits for each API call or task to finish) or ``NonBlocking`` (the host thread continues after making an API call or submitting a task without waiting for it to finish, or even start).
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+* All operations in a queue will be executed sequential in FIFO order.
```
fwyzard, line 44:
```suggestion
* All operations in a queue will be executed sequentially in FIFO order.
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+* Operations in different queues can run in parallel even on blocking queues.
```
fwyzard, line 45: With blocking queues, this requires using different host tasks, right ?

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+* ``wait()`` can be executed for queues to block the caller host thread until all previous enqueued work is finished.
+* Each ``queue`` is bounded to a specific ``device``.
```
fwyzard, line 47:
```suggestion
* Each ``queue`` is bound to a specific ``device``.
```

docs/source/basic/terms.rst
```diff
@@ -0,0 +1,70 @@
...
+Task
+----
+
+* A ``TaskKernel`` contains the algorithm which should be executed on a ``device``.
+* A ``HostTask`` is a functor without ``acc`` argument, which can be enqueued and is always executed on the host device. 
+
+Event
+-----
+
+* A ``event`` is a marker in a ``queue``.
```
fwyzard, line 58:
```suggestion
* An ``event`` is a marker in a ``queue``.
```

fwyzard:
Should it be `Mem Set` and `Mem Copy` instead of just `Set` and `Copy` ?

fwyzard:
@SimeonEhrig can you remove `[RFC] ` from the title of the PR ?

fwyzard:
Ah, no matter, I didn't know I could do it myself.

SimeonEhrig:
@psychocoderHPC Do you want to finish this PR or do we simply close it? I already mentioned it, if we don't merge it until release, it will be deleted.

SimeonEhrig:
Closed because alpaka 1.2.0 is released.

PR #2280, opened 2024-05-28
run examples with all enabled accelerators

psychocoderHPC:
Remove the usage of `ExampleDefaultAcc` in all examples.
 Examples will now be executed for all enabled accelerators. 
Fix that `ExampleDefaultAcc` was used in the test `MapIdxPitchBytes`.

note: The stream benchmark is still using `ExampleDefaultAcc`, this will be addressed in a separate PR.

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -4,7 +4,7 @@
  */
 
 #include <alpaka/alpaka.hpp>
-#include <alpaka/example/ExampleDefaultAcc.hpp>
+#include <alpaka/example/ExecuteForEachAccTag.hpp>
 
 #include <cstdint>
 #include <iostream>
@@ -64,24 +64,19 @@ struct FillBufferKernel
     }
 };
 
-auto main() -> int
+// In standard projects, you typically do not execute the code with any available accelerator.
+// Instead, a single accelerator is selected once from the active accelerators and the kernels are executed with the
+// selected accelerator only. If you use the example as the starting point for your project, you can rename the
+// example() function to main() and move the accelerator tag to the function body.
+template<typename TAccTag>
+auto example(TAccTag const&) -> int
 {
     // Define the index domain
     using Dim = alpaka::DimInt<3u>;
     using Idx = std::size_t;
 
     // Define the device accelerator
-    //
-    // It is possible to choose from a set of accelerators:
-    // - AccGpuCudaRt
-    // - AccGpuHipRt
-    // - AccCpuThreads
-    // - AccCpuOmp2Threads
-    // - AccCpuOmp2Blocks
-    // - AccCpuTbbBlocks
-    // - AccCpuSerial
-    // using Acc = alpaka::AccCpuSerial<Dim, Idx>;
-    using Acc = alpaka::ExampleDefaultAcc<Dim, Idx>;
+    using Acc = alpaka::TagToAcc<TAccTag, Dim, Idx>;
     std::cout << "Using alpaka accelerator: " << alpaka::getAccName<Acc>() << std::endl;
     // Defines the synchronization behavior of a queue
     //
@@ -90,12 +85,6 @@ auto main() -> int
     using DevQueue = alpaka::Queue<Acc, AccQueueProperty>;
 
     // Define the device accelerator
-    //
-    // It is possible to choose from a set of accelerators:
-    // - AccCpuThreads
-    // - AccCpuOmp2Threads
-    // - AccCpuOmp2Blocks
-    // - AccCpuSerial
     using Host = alpaka::AccCpuSerial<Dim, Idx>;
     // Defines the synchronization behavior of a queue
     //
@@ -256,4 +245,20 @@ auto main() -> int
     std::cout << std::endl;
 
     return EXIT_SUCCESS;
+}
+
+auto main() -> int
+{
+    // Execute the example once for each enabled accelerator.
+    // If you would like to execute it for a single accelerator only you can use the following code.
+    //  \code{.cpp}
+    //  auto tag = TagCpuSerial;
+    //  return example(tag);
+    //  \endcode
+    //
+    // valid tags:
+    //   TagCpuSerial, TagGpuHipRt, TagGpuCudaRt, TagCpuOmp2Blocks, TagCpuTbbBlocks,
+    //   TagCpuOmp2Threads, TagCpuSycl, TagCpuTbbBlocks, TagCpuThreads,
+    //   TagFpgaSyclIntel, TagGenericSycl, TagGpuSyclIntel
+    return alpaka::executeForEachAccTag([=](auto const& tag) { return example(tag); });
 }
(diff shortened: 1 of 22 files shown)


example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -257,3 +257,9 @@ auto main() -> int
 
     return EXIT_SUCCESS;
 }
+
+auto main() -> int
+{
+    // execute the example once for each enabled accelerator
+    return std::apply([](auto const&... tags) { return (example(tags) || ...); }, alpaka::EnabledAccTags{});
```
mehmetyusufoglu: Users will be challenged and might think that this is the way alpaka programs are written. It is an extra layer of difficulty for a novice user. ( I thought a solution by using `std::variant `and `std::visit `but seems not simpler than this solution. )

mehmetyusufoglu: Can't we solve this at CI level?

mehmetyusufoglu: Call `ctest -v`  many times in run_test.sh with different arguments? ( Add arguments to example's main function (i.e. argv) in cpp and use  CMAKE_CTEST_ARGUMENTS somehow? )

psychocoderHPC: Adding arguments will increase the complexity much more. The problem is we write code which can be executed on many devices so it is not like other projects where a simple main is enough.
This PR solves an issue that we are not able to execute our examples on all devices.
The now-shown CI issues show how easy it is to run the example with all accelerators.

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -64,7 +63,8 @@ struct FillBufferKernel
     }
 };
 
-auto main() -> int
+template<typename TAccTag>
```
SimeonEhrig, line 71: Maybe you should a small comment.

```c++
// In standard projects, not every kernel is executed with all active accelerators.
// Instead, a single accelerator is selected once from the active accelerators and the kernels are executed on the selected accelerator.
// If you use the example as the starting point for your project, you can rename the example() function to main() and move the accelerator type to the function body.
```

example/bufferCopy/src/bufferCopy.cpp
fwyzard: > Users will be challenged and might think that this is the way alpaka programs are written. It is an extra layer of difficulty for a novice user.

I fully agree.

If you want to write tests like this, OK... I personally strongly dislike this, but I won't object to obscure unreadable code in the tests, as long as the compiler and CI are happy.

But this should definitely not be how we write an example.

fwyzard:
@psychocoderHPC, please do not make the _examples_ even harder to understand :-(

psychocoderHPC:
> @psychocoderHPC, please do not make the _examples_ even harder to understand :-(

But I like to solve the issue we have that only a magic accelerator is used and as you can see the examples was not able to run with other accelerators because we never tested it.

fwyzard:
> But I like to solve the issue we have that only a magic accelerator is used and as you can see the examples was not able to run with other accelerators because we never tested it.

I don't disagree.

Do you think we could move the logic to build and test different accelerators to the build system ?

psychocoderHPC:
> Do you think we could move the logic to build and test different accelerators to the build system ?

I thought about it to but it look to me very complicated and will add a lot of code.

I decided to restructure my changes and add a helper to hide the folding expression.

psychocoderHPC:
I updated all examples and hopefully simplified the main by removing the folding expression.

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -257,3 +246,19 @@ auto main() -> int
...
+    //  \code{.cpp}
+    //  auto tag = TagCpuSerial;
+    //  return example(tag);
+    //  \endcode
+    //
+    // valid tags:
+    //   TagCpuSerial, TagGpuHipRt, TagGpuCudaRt, TagCpuOmp2Blocks, TagCpuTbbBlocks,
+    //   TagCpuOmp2Threads, TagCpuSycl, TagCpuTbbBlocks, TagCpuThreads,
+    //   TagFpgaSyclIntel, TagGenericSycl, TagGpuSyclIntel
+    return alpaka::executeForEachAccTag([=](auto const& tag) { return example(tag); });
```
SimeonEhrig, line 263: To honest, I'm not sure if we should use tags directly. Now it looks like, that tags are required to develop an alpaka application.

Maybe we should use the tags only internal in the `alpaka::executeForEachAccTag` function, which should become `alpaka::executeForEachAcc`.

psychocoderHPC: I do not see a way to implement it as easy as it is now. The reason is that we can not create an instance of Acc, so we can not pass it to the execution lambda.
I also prefer that we advertise tags more because of the reason that you are not allowed to instantiate tags on the host side. This avoids using accelerators as an argument in a lambda.
Additionally accelerators have the problem that the definition are not available if the backend is not activated in CMake.

psychocoderHPC: There is another advantage of using tags. Tags only say to which accelerator you would like to map your problem but is not say anything about the dimensionality of the index domain. This provides the freedom to switch dimensionality for each kernel depending on the needs.

SimeonEhrig:
```suggestion
    return alpaka::executeForEachAccTag([=](auto const& tag) { return example<alpaka::TagToAcc<decltype(TagToAcc)>>(); });
```

```c++
template<typename TAcc>
auto example() -> int {
// ...
using Acc = TAcc;
// ..
}
```

psychocoderHPC: IMO

```C++
return example(tag);
```
vs.
```C++
return example<alpaka::TagToAcc<decltype(TagToAcc)>>();
```

The problem is any way that the dimension and index type for the accelerator is required to create an accelerator.
So it will be 

```C++
return example<alpaka::TagToAcc<decltype(TagToAcc), Dim, Idx>();
```

Where Dim and Idx must moved away from the problem into the main function.

SimeonEhrig (approved):

psychocoderHPC:
@alpaka-group/alpaka-maintainers please merge

include/alpaka/example/ExecuteForEachAccTag.hpp
```diff
@@ -0,0 +1,27 @@
...
+    //
+    // @param callable callable which can be invoked with an accelerator tag
+    // @return disjunction of all invocation results
+    //
+    template<typename TCallable>
+    inline auto executeForEachAccTag(TCallable&& callable)
+    {
+        // Execute the callable once for each enabled accelerator.
+        // Pass the tag as first argument to the callable.
+        return std::apply([=](auto const&... tags) { return (callable(tags) || ...); }, alpaka::EnabledAccTags{});
```
fwyzard, line 25: @psychocoderHPC sorry, I hadn't thought about this before: does the `||` operator obey the usual short-circuit rules here ?

That is, as soon as one of the `callable` returns `true`, will the rest be skipped ?

psychocoderHPC: Yes short-circuit should be active but keep in mind that return zero is the case where we like to execute all. In case non-zero is returned what is an error we will stop the execution.

fwyzard: OK, so we don't want to continue running with the other tags if one fail ?

psychocoderHPC: Yes exact, I think we should not waiste compute time if we failed. 
I failure can also bring the device into a state that all followup work fails, in that case the last error will only be an artefact of the failing first run but the user will first look to the last failure.

PR #2283, opened 2024-05-31
Throw error from device to host with a msg  #2258

### LINKED ISSUE #2258: portable device-side `abort()` or `throw` ?
In a few places in the CMS code we would like to signal an error from the device side to the host code.

Something similar to a device-side `assert(false)`, but with a more meaningful error message and always available (`ALPAKA_ASSERT_ACC` may not be implemented on some back-ends).

For example, in one case we have something like
```c++
#if defined(__CUDACC__) && defined(__CUDA_ARCH__)
#define CMS_DEVICE_THROW(MSG)           \
  {                                     \
    printf("%s\n", (MSG));              \
    __trap();                           \
  }
#elif defined(__HIPCC__) && defined(__HIP_DEVICE_COMPILE__)
#define CMS_DEVICE_THROW(MSG)           \
  {                                     \
    printf("%s\n", (MSG));              \
    abort();                            \
  }
#else
#define CMS_DEVICE_THROW(MSG)           \
  {                                     \
    throw std::runtime_error(MSG);      \
  }
#endif
```

Does Alpaka already provide something like this ?

If not, would it be interesting to have it in Alpaka ?

mehmetyusufoglu:
Macros to make signalling an error from the device side to the host code.

In case of Cuda, Hip, Sycl a user defined message is added to the abort. For other backends `std::runtime_error` exception with the user message is thrown.

 Testing:
- Tests could be done for Accs: CpuThreads and CpuBlocks by catching the `runtime_error` exceptions thrown during exec. 

- Aborts can not be catched  from Cuda, Hip, Sycl as we call exec.  (Only tested by running a temporary fail test at CI)
But for Cuda; `__trap` triggers `runtime_error` can be catched during the wait(queue).

- For Hip, abort() is used. Which triggers  __builtin_trap()

- For sycl assert(false) is used. std::abort generated a runtime error at our local development system no compile error. https://github.com/alpaka-group/alpaka/pull/2283#issuecomment-2169733670

- I turned out to be OpenMP specification mandates std::runtime errors should be handled by the same thread otherwise it is converted to abort. I checked with a signal handler and `SIGABRT` is fired.  (Therefore openmp cases could not be tested other than a one time fail test at CI.)

**Issue**

fix #2258


mehmetyusufoglu:
Acc should be input of macros...

mehmetyusufoglu:
There is an nvcc compile error since `__trap` function is called in the KERNEL

`calling a __device__ function("__trap") from a __host__ __device__ function("operator()") is not allowed`

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,21 @@
+/* Copyright 2022 Andrea Bocci, Mehmet Yusufoglu */
+#pragma once
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
```
psychocoderHPC: The documentation string is missing.
There should be also a note that you can not catch the error like you can do it with exceptions.

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,34 @@
+/* Copyright 2022 Andrea Bocci, Mehmet Yusufoglu */
```
psychocoderHPC: This should be, see other files

```
/* Copyright 2022 Andrea Bocci, Mehmet Yusufoglu
 */
```

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,34 @@
...
+#pragma once
```
psychocoderHPC, line 5: please add an empty line above and after the pragma

psychocoderHPC (changes requested):

test/unit/kernel/src/KernelThrow.cpp
```diff
@@ -0,0 +1,85 @@
...
+        using Vec = alpaka::Vec<Dim, Idx>;
+        Vec const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+        Vec const globalThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+
+        // Map the three dimensional thread index into a one dimensional thread index space.
+        auto const linearizedGlobalThreadIdx = alpaka::mapIdx<1u>(globalThreadIdx, globalThreadExtent);
+
+        if(static_cast<unsigned>(linearizedGlobalThreadIdx[0u]) == 0u)
+        {
+            printf("linearizedGlobalThreadIdx: %lu\n", linearizedGlobalThreadIdx[0]);
```
psychocoderHPC: Windows CI fails here: 

```
 D:\a\alpaka\alpaka\test\unit\kernel\src\KernelThrow.cpp(29,20): error C2220: the following warning is treated as an error [D:\a\alpaka\alpaka\build\test\unit\kernel\kernelTest.vcxproj]
Z D:\a\alpaka\alpaka\test\unit\kernel\src\KernelThrow.cpp(29,20): 
warning C4477: 'printf' : format string '%lu'
 requires an argument of type 'unsigned long', but variadic argument 1 has type 'TVal' [D:\a\alpaka\alpaka\build\test\unit\kernel\kernelTest.vcxproj]
```

mehmetyusufoglu: solved by changing  integer literal suffix in  `printf` to `u` (from lu)  and Acc index type to `uint32_t` from `size_t` .

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: added, thanks.

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,38 @@
...
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf("%s [ALPAKA_DEVICE_THROW: Calling abort(). Hip backend is enabled.]\n", (MSG));                    \
+            abort();                                                                                                  \
+        }
+#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf("%s [ALPAKA_DEVICE_THROW: Calling std::abort(). Sycl backend is enabled.]\n", (MSG));              \
+            std::abort();                                                                                             \
```
fwyzard: Are you sure we can call `std::abort()` from inside a SYCL kernel running _e.g._ on a GPU ?

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,38 @@
+/* Copyright 2022  Andrea Bocci, Mehmet Yusufoglu
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+//! ALPAKA_DEVICE_THROW either aborts(terminating the program and creating a core dump) or throws std::runtime_error
+//! depending on the Acc. It is impossible to "catch" aborts; but the std::runtime_error exception can be catched in
```
fwyzard: While we cannot catch the `abort()` or `__trap()` on the device, those should result in raising an asynchronous error in the GPU runtime; we should be able to detect that and throw and exception, that _can_ be caught.

Is this model not working as expected ?

test/unit/kernel/src/KernelThrow.cpp
```diff
@@ -0,0 +1,85 @@
...
+        using Vec = alpaka::Vec<Dim, Idx>;
+        Vec const globalThreadIdx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
+        Vec const globalThreadExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Threads>(acc);
+
+        // Map the three dimensional thread index into a one dimensional thread index space.
+        auto const linearizedGlobalThreadIdx = alpaka::mapIdx<1u>(globalThreadIdx, globalThreadExtent);
+
+        if(static_cast<unsigned>(linearizedGlobalThreadIdx[0u]) == 0u)
+        {
+            printf("linearizedGlobalThreadIdx: %u\n", linearizedGlobalThreadIdx[0]);
```
fwyzard: This could be simplified to
```suggestion
        if(globalThreadIdx == Vec::zeros())
        {
```

test/unit/kernel/src/KernelThrow.cpp
```diff
@@ -0,0 +1,85 @@
...
+        }
+        alpaka::syncBlockThreads(acc);
+    }
+};
+
+template<typename T, typename Acc>
+void checkThrow()
+{
+    if constexpr(alpaka::accMatchesTags<Acc, T>)
+    {
```
fwyzard: Why the extra complexity of using a `Tag` ?

Why not simply
```suggestion
template<typename Acc>
void checkThrow()
{
```
?

test/unit/kernel/src/KernelThrow.cpp
```diff
@@ -0,0 +1,85 @@
...
+        using Idx = alpaka::Idx<Acc>;
+        using Dim = alpaka::Dim<Acc>;
+        using Vec = alpaka::Vec<Dim, Idx>;
+        using QueueProperty = alpaka::Blocking;
+        using Queue = alpaka::Queue<Acc, QueueProperty>;
```
fwyzard:
```suggestion
        using Queue = alpaka::Queue<Acc, alpaka::Blocking>;
```

test/unit/kernel/src/KernelThrow.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+        Queue queue(devAcc);
+        auto const workDiv = alpaka::WorkDivMembers<Dim, Idx>{Vec{8}, Vec{1}, Vec{1}};
+
+        KernelWithThrow kernel;
+        auto runner = [&]() { alpaka::exec<Acc>(queue, workDiv, kernel); };
+        try
+        {
+            runner();
+        }
```
fwyzard: why not more simply
```suggestion
        try
        {
            alpaka::exec<Acc>(queue, workDiv, KernelWithThrow{});
        }
```
?

test/unit/kernel/src/KernelThrow.cpp
```diff
@@ -0,0 +1,85 @@
...
+    // Test runtime-error exceptions. Aborts can not be catched.
+    checkThrow<alpaka::TagCpuThreads, Acc>();
+    checkThrow<alpaka::TagCpuSerial, Acc>();
+
+    // The OpenMP specification mandates that exceptions thrown by some thread must be handled by the same thread and
+    // within the same parallel region. Therefore std::runtime_error aborts the program.
+
+    // checkThrow<alpaka::TagCpuOmp2Blocks, Acc>();
+    // checkThrow<alpaka::TagCpuOmp2Threads, Acc>();
+}
```
fwyzard: Aren't there any tests for the GPU runtimes ?

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: I tried to get errors using cudaDeviceSynchronize  and cudaGetLastError but I can not get the error type.  But I have not tried async handlers. Because they will abort the test ?

For example; I used a signal handler for OpemMp to check whether OpenMP converts runtime_error to abort. Yes signal handler catched the SIGABRT signal. But when  I add that handler, in the test the test aborts eventually.

fwyzard: I don't have any experience with OpenMP, so I cannot comment in that regard.
However, using signal handlers seems way outside the scope of this feature and of this test.

test/unit/kernel/src/KernelThrow.cpp
mehmetyusufoglu: Because `checkThrow` is called seperately for each enabled Acc.   it is called in a block by a `TestAccs` parameter type (tuple of Accs) which includes all enabled types.
```

using TestAccs = alpaka::test::EnabledAccs<alpaka::DimInt<1u>, std::uint32_t>;

TEMPLATE_LIST_TEST_CASE("ThrowForCpuThreadAndSerial", "[kernel]", TestAccs)
{
    using Acc = TestType;
    // Test runtime-error exceptions. Aborts can not be catched.
    checkThrow<alpaka::TagCpuThreads, Acc>();
```

include/alpaka/core/RuntimeMacros.hpp
fwyzard: For the CUDA case, here is an example that illustrates the behaviour I have in mind:
```c++
#include <iostream>
#include <string>

#include <alpaka/alpaka.hpp>

using Dim = alpaka::DimInt<1u>;
using Idx = uint32_t;

struct KernelWithThrow {
  template<typename TAcc>
  ALPAKA_FN_ACC void operator()(TAcc const& acc) const {
    auto idx = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
    if (idx == Vec::zeros()) {
#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && defined(__CUDA_ARCH__)
      __trap();
#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED) && defined(__HIP_DEVICE_COMPILE__)
      abort();
#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
      abort();
#else
      throw std::runtime_error();
#endif
    }
  }
};


int main(void) {
  using Platform = alpaka::PlatformCudaRt;
  using Device = alpaka::DevCudaRt;
  using Queue = alpaka::Queue<Device, alpaka::Blocking>;
  using Tag = alpaka::TagGpuCudaRt;

  Platform platform{};
  Device device = alpaka::getDevByIdx(platform, 0);
  Queue queue{device};

  auto const workDiv = alpaka::WorkDivMembers<Dim, Idx>{Vec{8}, Vec{1}, Vec{1}};

  try {
    alpaka::exec<alpaka::TagToAcc<Tag, Dim, Idx>>(queue, workDiv, KernelWithThrow{});
  } catch(std::runtime_error& e) {
    std::cerr << "Caught an error in alpaka::exec:\n" << e.what() << '\n';
  }

  // note that this kind of error is "sticky", and will not be reset by a call to cudaGetLastError()

  try {
    alpaka::wait(queue);
  } catch(std::runtime_error& e) {
    std::cerr << "Caught an error in alpaka::wait:\n" << e.what() << '\n';
  }

  try {
    alpaka::wait(queue);
  } catch(std::runtime_error& e) {
    std::cerr << "Caught an error in alpaka::wait:\n" << e.what() << '\n';
  }

  return 0;
}
```

Compiled with
```bash
/usr/local/cuda/bin/nvcc -std=c++17 -O3 -g -I/home/fwyzard/src/alpaka-group/alpaka/include --expt-relaxed-constexpr -arch sm_86 -DALPAKA_ACC_GPU_CUDA_ENABLED test.cu -o test
```

Results in
```bash
./test 
Caught an error in alpaka::wait:
/home/fwyzard/src/alpaka-group/alpaka/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp(175) 'TApi::streamSynchronize(queue.getNativeHandle())' A previous API call (not this one) set the error  : 'cudaErrorLaunchFailure': 'unspecified launch failure'!
Caught an error in alpaka::wait:
/home/fwyzard/src/alpaka-group/alpaka/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp(175) 'TApi::streamSynchronize(queue.getNativeHandle())' returned error  : 'cudaErrorLaunchFailure': 'unspecified launch failure'!
/home/fwyzard/src/alpaka-group/alpaka/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp(74) 'TApi::streamSynchronize(m_UniformCudaHipQueue)' returned error  : 'cudaErrorLaunchFailure': 'unspecified launch failure'!
/home/fwyzard/src/alpaka-group/alpaka/include/alpaka/queue/cuda_hip/QueueUniformCudaHipRt.hpp(75) 'TApi::streamDestroy(m_UniformCudaHipQueue)' returned error  : 'cudaErrorLaunchFailure': 'unspecified launch failure'!
```

fwyzard: I'm not sure why we don't get the error directly in `alpaka::exec` even with a blocking queue - but we do get the error in `alpaka::wait`.

mehmetyusufoglu: So code did not abort in this case?

test/unit/kernel/src/KernelThrow.cpp
mehmetyusufoglu: No because I can catch the "cudaErrorLaunchFailure" string in the exception correctly. But the test eventually makes the other tests fail in the same executable. I will try with putting to a separate exe.

fwyzard: Yes, because `__trap()` causes an error that will not be reset.

Maybe one could try destroying the device and re-initialising it, but I'm not sure it's worth it.

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: We only support one-api and it is tested on CI and HAL. .  I wanted to do assert(false) but assert turned out to be calling std::abort from kernel in sycl ([url](https://github.com/intel/llvm/pull/3461/files/641d07178f240f1b3f4d63e9f1836095f274eeae#diff-27b10bd82fa1537002ee64fb2aa6b289ac9aa6527cc17d51b5ceb850c384d202))

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: ok I have added wait for Cuda and added a comment to the test. Thanks.

test/unit/kernel/src/KernelThrow.cpp
mehmetyusufoglu: Yes I don't think it is worth for now. I have moved the test to a standalone file to be able to test cuda backend as well. Thanks.

fwyzard:
I tried calling `abort()` from a SYCL kernel running on an Intel GPU, with oneAPI 2024.1, but it doesn't seem to work:
```c++
#include <cstdlib>
#include <exception>
#include <iostream>

#include <sycl.hpp>

// Exception handler for asynchronous SYCL exceptions
void exception_handler(sycl::exception_list exceptions) {
    for (const auto& e : exceptions) {
        try {
            std::rethrow_exception(e);
        } catch (const sycl::exception& e) {
            std::cout << "Caught asynchronous SYCL exception: " << e.what() << std::endl;
        }
    }
}

int main() {
    try {
        // Create a SYCL queue with an exception handler
        sycl::queue queue(exception_handler);

        // Submit a command group to the queue
        queue.submit([&](sycl::handler& cgh) {
            // Define the kernel
            cgh.single_task([=]() {
                sycl::ext::oneapi::experimental::printf("%d\n", 1);
                std::abort();
                sycl::ext::oneapi::experimental::printf("%d\n", 2);
            });
        });

        // Wait for the queue to finish executing
        queue.wait_and_throw();
    } catch (sycl::exception const& e) {
        std::cout << "Caught synchronous SYCL exception: " << e.what() << std::endl;
    } catch (std::exception const& e) {
        std::cout << "Cuaght std exception:" << e.what() << '\n';
    } catch (...) {
        std::cout << "Caught unexpected exception\n";
    }

    return 0;
}
```

fails at runtime (when the JIT compiler is called) with
```
Caught synchronous SYCL exception: The program was built for 1 devices
Build program log for 'Intel(R) Data Center GPU Flex 170':
Module <0x3610e90>:  Unresolved Symbol <abort>
Module <0x3610e90>:  Unresolved Symbol <abort> -999 (Unknown PI error)
```

I suspect that `abort()` (or `std::abort()`) works on the OpenCL CPU backend because it finds the system-level function, but is not a portable feature of SYCL/oneAPI.

mehmetyusufoglu:
@fwyzard  Thanks. I have started a CI run for `TagGpuSyclIntel` case. Then I will start for `TagCpuSycl`. I mean the accelerators.  We will see. 

I Tried on HAL computer now. There are no compile errors. But I got the same result with you at run-time.

```
 spack load cuda@12.2
 spack load intel-oneapi-compilers@2023.1.0
 spack load intel-oneapi-tbb@2021.10.0
 module load rocm-5.4.6
 sycl-ls
[opencl:acc:0] Intel(R) FPGA Emulation Platform for OpenCL(TM), Intel(R) FPGA Emulation Device 1.2 [2023.15.3.0.20_160000]
[opencl:cpu:1] Intel(R) OpenCL, AMD EPYC 7452 32-Core Processor                 3.0 [2023.15.3.0.20_160000]
[opencl:gpu:2] Intel(R) OpenCL Graphics, Intel(R) Arc(TM) A770 Graphics 3.0 [23.52.28202.52]
[ext_oneapi_level_zero:gpu:0] Intel(R) Level-Zero, Intel(R) Arc(TM) A770 Graphics 1.3 [1.3.28202]
[ext_oneapi_cuda:gpu:0] NVIDIA CUDA BACKEND, NVIDIA A30 0.0 [CUDA 12.4]
[ext_oneapi_hip:gpu:0] AMD HIP BACKEND, AMD Radeon RX 7900 XTX 0.0 [HIP 50422.80]

COMPILE
 icpx -fsycl test_abort.cpp
RUN
yusufo81@hal8999:~$ ./a.out
Caught synchronous SYCL exception: The program was built for 1 devices
Build program log for 'Intel(R) Arc(TM) A770 Graphics':
Module <0x39adf80>:  Unresolved Symbol <abort>
Module <0x39adf80>:  Unresolved Symbol <abort> -11 (PI_ERROR_BUILD_PROGRAM_FAILURE)
```

fwyzard:
I've tried looking into the specification of SYCL and the implementation of oneAPI, and I suspect the only portable solution may be `assert(false)`.

Hover
  - on my laptop integrated GPU that actually does not do anything :-(
  - on a Flax 170 datacenter GPU it causes the program to abort in a way that cannot be caught.

I'll bring it up with Intel at the next occasion - but for the time being, that seems better than nothing.

mehmetyusufoglu:
> I've tried looking into the specification of SYCL and the implementation of oneAPI, and I suspect the only portable solution may be `assert(false)`.
> 
> Hover
> 
>     * on my laptop integrated GPU that actually does not do anything :-(
> 
>     * on a Flax 170 datacenter GPU it causes the program to abort in a way that cannot be caught.
> 
> 
> I'll bring it up with Intel at the next occasion - but for the time being, that seems better than nothing.

Ok I have added `assert(false)` , thanks a lot. I can not catch the exception at HAL computer so i did not test it in the test code.

```
1
2
AssertHandler::printMessage
test_abort.cpp:27: auto main()::(anonymous class)::operator()(sycl::handler &)::(anonymous class)::operator()() const: global id: [0,0,0], local id: [0,0,0] Assertion `false` failed.
```

fwyzard:
> can not catch the exception at HAL computer so i did not test it in the test code.

Unrelated to the test, but what is the HAL computer ?

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,39 @@
...
+//! ALPAKA_DEVICE_THROW either aborts(terminating the program and creating a core dump) or throws std::runtime_error
+//! depending on the Acc. The std::runtime_error exception can be catched in the catch block.
+//!
+//! CUDA __trap function triggers std::runtime_error but can be catched during wait not exec.
+//!
+//! The OpenMP specification mandates that exceptions thrown by some thread must be handled by the same thread.
+//! Therefore std::runtime_error thrown by ALPAKA_DEVICE_THROW aborts the the program for OpenMP backends. If needed
+//! the SIGABRT signal can be catched by signal handler.
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && defined(__CUDA_ARCH__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
```
fwyzard: Does this need to be a macro ?
Can we use a device function instead ?

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,39 @@
...
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED) && defined(__HIP_DEVICE_COMPILE__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf("%s [ALPAKA_DEVICE_THROW: Calling abort(). Hip backend is enabled.]\n", (MSG));                    \
+            abort();                                                                                                  \
+        }
+#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf("%s [ALPAKA_DEVICE_THROW: Calling std::abort(). Sycl backend is enabled.]\n", (MSG));              \
```
fwyzard: This doesn't call any more `std::abort`

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,39 @@
...
+//!
+//! CUDA __trap function triggers std::runtime_error but can be catched during wait not exec.
+//!
+//! The OpenMP specification mandates that exceptions thrown by some thread must be handled by the same thread.
+//! Therefore std::runtime_error thrown by ALPAKA_DEVICE_THROW aborts the the program for OpenMP backends. If needed
+//! the SIGABRT signal can be catched by signal handler.
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && defined(__CUDA_ARCH__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf("%s [ALPAKA_DEVICE_THROW: Calling __trap(). Cuda backend is enabled.]\n", (MSG));                  \
```
fwyzard: I think a message like
```
%s [ALPAKA_DEVICE_THROW: Calling __trap(). Cuda backend is enabled.]\n
```
is fine for debugging the implementation - but once this goes in production can we print just the user-supplied message ?

Or, if we do want to add more information, something like
```
alpaka encountered a user-defined error condition while running on the ___ back-end:\n
%s
```

mehmetyusufoglu: ok, changed. Thanks.

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: Will not make much difference since only 2 lines are inlined. On the other hand Isn't it better to call the aborts or exceptions directly ?

mehmetyusufoglu:
> > can not catch the exception at HAL computer so i did not test it in the test code.
> 
> Unrelated to the test, but what is the HAL computer ?

A HPC system at HZDR :)  Has different kind of gpus and easily configurable.

include/alpaka/core/RuntimeMacros.hpp
fwyzard: @psychocoderHPC what do you prefer here, functions or macros ?

psychocoderHPC: Functions are always prefered above macros. We should only use macros if we need `__LINE__`, ...

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,45 @@
...
+//! ALPAKA_DEVICE_THROW either aborts(terminating the program and creating a core dump) or throws std::runtime_error
+//! depending on the Acc. The std::runtime_error exception can be catched in the catch block.
+//!
+//! CUDA __trap function triggers std::runtime_error but can be catched during wait not exec.
+//!
+//! The OpenMP specification mandates that exceptions thrown by some thread must be handled by the same thread.
+//! Therefore std::runtime_error thrown by ALPAKA_DEVICE_THROW aborts the the program for OpenMP backends. If needed
+//! the SIGABRT signal can be catched by signal handler.
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) && defined(__CUDA_ARCH__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
```
psychocoderHPC: I was on the way to say `change it to a C++ function and than this should be deviceThrow() within the alpaka namespace.`
**BUT** than it should be a trait which gets the acc as argument with the coresponding class equal to the trait name which is than specialized for each acc. Device functions always take the acc as first argument.

Never the less I would handle it equal to `ALPAKA_ASSERT_ACC` and would add the line and error message to the printf.
To follow the naming schema this macro should be named `ALPAKA_THROW_ACC`

psychocoderHPC (changes requested):

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,45 @@
...
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the Cuda back-end: \n %s ",       \
```
fwyzard:
```suggestion
                "alpaka encountered a user-defined error condition while running on the CUDA back-end:\n%s",       \
```
(just being pedantic)

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,45 @@
...
+                (MSG));                                                                                               \
+            __trap();                                                                                                 \
+        }
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED) && defined(__HIP_DEVICE_COMPILE__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the HIP back-end: \n %s ",        \
```
fwyzard:
```suggestion
                "alpaka encountered a user-defined error condition while running on the HIP back-end:\n%s",        \
```
_etc_

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,45 @@
...
+                (MSG));                                                                                               \
+            abort();                                                                                                  \
+        }
+#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the Sycl back-end: \n %s ",       \
+                (MSG));                                                                                               \
```
fwyzard: I don't think `printf` works like this in SYCL code running on Intel GPUs.

AuroraPerego: you need to include `alpaka/core/Sycl.hpp` for the `printf` to work in SYCL

mehmetyusufoglu:
Thanks, i  removed that line.

18 Haz 2024 Sal 03:07 tarihinde Aurora Perego ***@***.***>
şunu yazdı:

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: Thanks, I removed the line.

AuroraPerego: I think you can keep this line and add the include, it's useful to have a message that explains the reason for the throw.

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,42 @@
...
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the HIP back-end: \n %s ",        \
+                (MSG));                                                                                               \
+            abort();                                                                                                  \
+        }
+#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
```
fwyzard, line 40: can you add back the `printf` here (with the proper includes) ?

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: Hence; according to you it should be a macro called ALPAKA_THROW_ACC without an Acc argument?

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: Ok, done. Thanks.

test/unit/runtime/src/KernelThrow.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+using TestAccs = alpaka::test::EnabledAccs<alpaka::DimInt<1u>, std::uint32_t>;
+
+TEMPLATE_LIST_TEST_CASE("ThrowForCpuThreadAndSerial", "[runtime]", TestAccs)
+{
+    using Acc = TestType;
+    // Test runtime-error exceptions. Aborts can not be catched.
+    checkThrow<alpaka::TagCpuThreads, Acc>("Exception thrown by the kernel");
+    checkThrow<alpaka::TagCpuSerial, Acc>("Exception thrown by the kernel");
+    checkThrow<alpaka::TagGpuCudaRt, Acc>("cudaErrorLaunchFailure");
```
fwyzard, line 83: can you add the tests for the other backends ?

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,48 @@
...
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the HIP back-end:\n%s",           \
+                (MSG));                                                                                               \
+            abort();                                                                                                  \
+        }
+#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the Sycl back-end:\n%s",          \
```
fwyzard:
```suggestion
                "alpaka encountered a user-defined error condition while running on the SYCL back-end:\n%s",          \
```

test/unit/runtime/src/KernelThrow.cpp
mehmetyusufoglu: @fwyzard  can you test **hip/rocm** abort() if not difficult for you, or if you don't mind. I tested but can not trigger abort with abort() function. But i tested for only a specific setup. ( On the other hand, for `OpenMp` I explained the issue at PR explanation, the `sycl` assert(false) can not be catched. Only `Hip` can be added but I assumed it is not catchable somehow, may be due to Rene's explanations. )

mehmetyusufoglu: `__assert_fail` did not work for hip/rocm.

psychocoderHPC:
@mehmetyusufoglu please rebase against develop branch to fix the CI issues

test/unit/runtime/src/KernelThrow.cpp
fwyzard: > @fwyzard can you test **hip/rocm** abort() if not difficult for you, or if you don't mind. I tested but can not trigger abort with abort() function.

I've tried using `abort()` inside a ROCm kernel.

The kernel did stop at the point of the call to `abort()`, but it did not print any messages, and it did not cause any errors on the host :-/

fwyzard: `__assert_fail()` is identical to `abort()`

mehmetyusufoglu: Thanks a lot, good info.  I tried  `asm volatile("s_trap 0x1;"); return; ` but that did not stop the kernel. There is no `raise` function to `raise(SIGABRT).` Null pointer access to create _segmentation fault_ stops the kernel, then host program stalls, but does not exit.

include/alpaka/core/RuntimeMacros.hpp
psychocoderHPC: yes this will follow our macro `ALPAKA_ASSERT_ACC` and avoid introducing a new trait. IMO this function/macro will be used in seldem cases therefore I would like to avoid increasing the code base because of this tiny throw.

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,48 @@
...
+                (MSG));                                                                                               \
+            __trap();                                                                                                 \
+        }
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED) && defined(__HIP_DEVICE_COMPILE__)
+#    define ALPAKA_DEVICE_THROW(MSG)                                                                                  \
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the HIP back-end:\n%s",           \
+                (MSG));                                                                                               \
+            abort();                                                                                                  \
```
psychocoderHPC, line 36: @mehmetyusufoglu reported offline that aborting is not working. We use `__builtin_trap();` what is working fine.

test/unit/runtime/src/KernelThrow.cpp
mehmetyusufoglu: ` __builtin_trap()` is recommended by  @psychocoderHPC  . I will try.

AuroraPerego: Andrea already tried and said it behaved like `abort()`. Looking at the implementation they are the same:
```cpp
// abort
extern "C" __device__ inline __attribute__((weak))
void abort() {
  __builtin_trap();
}
```

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: `__builtin_trap();`  halts the main thread (host) but does not exit from the program.  

The only working command seems to be `assert(false)`. I used assert(false) because the code below aborted with a core dump.  Thrown exception not be catched as an` std::excception` or `hipError`.



```C++
#include <hip/hip_runtime.h>
#include < iostream >  
#include <signal.h>

// Kernel definition
__global__ void myKernel(int *data, int threshold) {
  int idx = blockIdx.x * blockDim.x + threadIdx.x;

  // Trigger a segmentation fault by accessing invalid memory
  // asm volatile("s_trap 0x1;");
  //            __builtin_trap();a
  if (idx == 0) {
    printf("Aborting kernel at index %d\n", idx);

    assert(false);
  }
  // Kernel operation
  data[idx] *= 2;
}

int main() {
  const int dataSize = 10;
  int h_data[dataSize] = {1, 2, 3, 4, 5, 6, 7, 8, 9, 10};
  int threshold = 7;

  int *d_data;
  hipMalloc(&d_data, dataSize * sizeof(int));
  hipMemcpy(d_data, h_data, dataSize * sizeof(int), hipMemcpyHostToDevice);

  dim3 blockDim(5);
  dim3 gridDim((dataSize + blockDim.x - 1) / blockDim.x);
  try {
    // Launch the kernel
    hipLaunchKernelGGL(myKernel, gridDim, blockDim, 0, 0, d_data, threshold);
    // hipDeviceSynchronize();
    // Wait for GPU to finish before accessing on host
  } catch (...) {
    printf("catched from try");
  }
  hipError_t err = hipDeviceSynchronize();
  //  Check for errors
  if (err != hipSuccess) {
    std::cerr << "Kernel execution aborted or encountered an error: "
              << hipGetErrorString(err) << std::endl;
    std::exit(EXIT_FAILURE);
  }
  printf("copy data back to host:\n");

  // Copy data back to host
  //  hipMemcpy(h_data, d_data, dataSize * sizeof(int), hipMemcpyDeviceToHost);

  // Print the results
  //    for (int i = 0; i < dataSize; i++) {
  //      std::cout << h_data[i] << " ";
  // }
  std::cout << std::endl;

  // Cleanup
  hipFree(d_data);

  return 0;
}
// compiled by command: hipcc kernel_abort.cpp -o kernel_abort
// rocm version: rocm-5.7.2
```

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: Ok, then now we have a macro which is taking a string, throwing or aborting depending on backend.

mehmetyusufoglu: macro name changed to ALPAKA_THROW_ACC. Thanks.

test/unit/runtime/src/KernelThrow.cpp
mehmetyusufoglu: Yes, I used assert(false) eventually.

fwyzard (dismissed):

SimeonEhrig (dismissed):

psychocoderHPC (approved):

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,48 @@
...
+        }
+#elif defined(ALPAKA_ACC_SYCL_ENABLED) && defined(__SYCL_DEVICE_ONLY__)
+#    define ALPAKA_THROW_ACC(MSG)                                                                                     \
+        {                                                                                                             \
+            printf(                                                                                                   \
+                "alpaka encountered a user-defined error condition while running on the SYCL back-end:\n%s",          \
+                (MSG));                                                                                               \
+            assert(false);                                                                                            \
```
psychocoderHPC: Note: assert is a noop incase `-DNDEBUG` is set.

psychocoderHPC: We should use `std::abort` even if it is not ending the application similar to HIP.

psychocoderHPC (changes requested):

include/alpaka/core/RuntimeMacros.hpp
```diff
@@ -0,0 +1,48 @@
...
+// Implementation details
+#include "alpaka/core/Sycl.hpp"
+
+//! ALPAKA_THROW_ACC either aborts(terminating the program and creating a core dump) or throws std::runtime_error
+//! depending on the Acc. The std::runtime_error exception can be catched in the catch block.
+//!
+//! CUDA __trap function triggers std::runtime_error but can be catched during wait not exec.
+//!
+//! The OpenMP specification mandates that exceptions thrown by some thread must be handled by the same thread.
+//! Therefore std::runtime_error thrown by ALPAKA_THROW_ACC aborts the the program for OpenMP backends. If needed
```
psychocoderHPC, line 20: please add HIP and SYCL behaviour too

psychocoderHPC (changes requested):

include/alpaka/core/RuntimeMacros.hpp
AuroraPerego: `std::abort()` cannot be used inside a SYCL kernel. It works on the CPU, but on the GPU it gives:
```shell
Caught synchronous SYCL exception: The program was built for 1 devices
Build program log for 'Intel(R) Data Center GPU Flex 170':
Module <0x3610e90>:  Unresolved Symbol <abort>
Module <0x3610e90>:  Unresolved Symbol <abort> -999 (Unknown PI error)
```
According to Intel developers `assert(false)` is the suggested approach (even though they know the behaviour is different from `__trap` and they may implement something more similar to it in the future).

mehmetyusufoglu: My tests previously done had generated the same runtime error:

https://github.com/alpaka-group/alpaka/pull/2283#issuecomment-2169733670

include/alpaka/core/RuntimeMacros.hpp
fwyzard: (wrong box)

include/alpaka/core/RuntimeMacros.hpp
fwyzard: this is a oneAPI implementation details, but you could try calling 
```c++
#ifdef __GLIBC__
extern "C" {
extern __DPCPP_SYCL_EXTERNAL void __assert_fail(const char *expr,
                                                const char *file,
                                                unsigned int line,
                                                const char *func);
}
#elif defined(_WIN32)
extern "C" {
// TODO: documented C runtime library APIs must be recognized as
//       builtins by FE. This includes _dpcomp, _dsign, _dtest,
//       _fdpcomp, _fdsign, _fdtest, _hypotf, _wassert.
//       APIs used by STL, such as _Cosh, are undocumented, even though
//       they are open-sourced. Recognizing them as builtins is not
//       straightforward currently.
extern __DPCPP_SYCL_EXTERNAL void _wassert(const wchar_t *wexpr,
                                           const wchar_t *wfile, unsigned line);
}
#endif
```

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: Added. For sycl it turned out to be my tests and @AuroraPerego and @fwyzard's tests both generated a runtime error saying "...SYCL exception: Unresolved Symbol <abort> " etc. Therefore assert(false) is used and related comments are added.

include/alpaka/core/RuntimeMacros.hpp
mehmetyusufoglu: assert(false) works as I tested before:
```
spack load intel-oneapi-compilers@2023.1.0
$ icpx -fsycl test_abort.cpp
$ ./a.out 
AssertHandler::printMessage
test_abort.cpp:27: auto main()::(anonymous class)::operator()(sycl::handler &)::(anonymous class)::operator()() const: global id: [0,0,0], local id: [0,0,0] Assertion `false` failed.
Aborted (core dumped)
```

fwyzard: It probably does not work if the user defines `NDEBUG`.

Can you try compiling with `-DNDEBUG` ?

mehmetyusufoglu: yes, when compiled with -DNDEBUG `assert(false)` does not abort`  icpx -fsycl test_abort.cpp -DNDEBUG`. Hence assert(false) can not be used I will use `abort() `because at least there is an exception although it is not abort.

```
COMPILE code with std::Abort
 icpx -fsycl test_abort.cpp
RUN
yusufo81@hal8999:~$ ./a.out
Caught synchronous SYCL exception: The program was built for 1 devices
Build program log for 'Intel(R) Arc(TM) A770 Graphics':
Module <0x39adf80>:  Unresolved Symbol <abort>
Module <0x39adf80>:  Unresolved Symbol <abort> -11 (PI_ERROR_BUILD_PROGRAM_FAILURE)
```

psychocoderHPC (approved):

PR #2286, opened 2024-06-04
Align shared memory using alignas(16)

fwyzard:
No changes expected.

### DIFF
--- a/include/alpaka/block/shared/dyn/BlockSharedMemDynUniformCudaHipBuiltIn.hpp
+++ b/include/alpaka/block/shared/dyn/BlockSharedMemDynUniformCudaHipBuiltIn.hpp
@@ -8,6 +8,7 @@
 #include "alpaka/core/BoostPredef.hpp"
 #include "alpaka/core/Concepts.hpp"
 
+#include <cstddef>
 #include <type_traits>
 
 #if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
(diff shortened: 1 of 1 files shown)


include/alpaka/block/shared/dyn/BlockSharedMemDynUniformCudaHipBuiltIn.hpp
```diff
@@ -38,10 +38,12 @@ namespace alpaka
...
-                // we have to use the widest possible type to have all types aligned correctly.
-                // See: http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#shared
-                // http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#vector-types
-                extern __shared__ float4 shMem[];
+                // we use the widest possible alignment supported by CUDA types to have
+                // all types aligned correctly.
+                // See:
+                //   - http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#shared
+                //   - http://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#vector-types
+                extern __shared__ std::byte shMem[] alignas(16);
```
mehmetyusufoglu: Hardcoded because there is no  `max_align_t` type in Cuda?

fwyzard: That's a good idea, but it doesn't seem to work :-/

The regular CUDA headers do not define `max_align_t`.

The NVIDIA C++ STD-like headers in `/usr/local/cuda/targets/x86_64-linux/include/cuda/std/detail/libcxx/include/cstddef` has
```c++
#if defined(__CLANG_MAX_ALIGN_T_DEFINED) || defined(_GCC_MAX_ALIGN_T) || \
    defined(__DEFINED_max_align_t) || defined(__NetBSD__)
// Re-use the compiler's <stddef.h> max_align_t where possible.
using ::max_align_t;
#else
typedef long double max_align_t;
#endif
```

So it uses the host definition of `max_align_t` is available, and falls back to `long double` otherwise.

So we could use the host version of `max_align_t`, but at least on my machine that is `32`, which is more than is needed for the CUDA types, and I'd rather not _increase_ the shared memory alignment :-/

fwyzard: Actually I made a mistake: `sizeof(max_align_t)` is 32, but `alignof(max_align_t)` is 16, so we can use that.

fwyzard: @mehmetyusufoglu thank you for the suggestion.

psychocoderHPC (approved):

Issue #2287, opened 2024-06-04
Why is `AtomicAtomicRef` always disabled?


chillenzer:
Hi,
quick question: What's the reason for [this line](https://github.com/alpaka-group/alpaka/blame/c2a37cc8c9dd82b45a203318bf6000febc460f45/include/alpaka/atomic/AtomicCpu.hpp#L10)? Explicitly:
```C++
#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
#    define ALPAKA_DISABLE_ATOMIC_ATOMICREF
#endif
```
This ALWAYS AND UNCONDITIONALLY disables the `AtomicAtomicRef` path.

fwyzard:
Ops...

It used to be wrong in a different way:
```c++
#ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
#    if BOOST_VERSION < 107300
#        warning boost::atomic_ref requires Boost version 1.73 or higher. Please update your version of Boost, or disable the use of boost::atomic_ref with -DALPAKA_DISABLE_ATOMIC_ATOMICREF
#    endif
#    define ALPAKA_DISABLE_ATOMIC_ATOMICREF
#endif
```

Then, when we removed the Boost version check, we left the wrong definition.

fwyzard:
https://github.com/alpaka-group/alpaka/pull/2288 should fix it.

psychocoderHPC:
fixed with #2288 and #2294

PR #2288, opened 2024-06-04
Re-enable AtomicAtomicRef path

fwyzard:


fwyzard:
fixes #2287

psychocoderHPC (dismissed):

fwyzard:
OK, for me this "error"
> ```
> /home/runner/work/alpaka/alpaka/include/alpaka/atomic/AtomicAtomicRef.hpp:222:36: error: comparing floating point with == or != is unsafe [-Werror,-Wfloat-equal]
> ```
is going too far.

What is the supposedly "safe" way to compare floats in a CAS loop ?
Check if they haven't changed too much ?

mehmetyusufoglu:
I might not have understood the question correctly, but what about this: 

```
 float value = atomicFloat.load();
    while (fuzzyEqual(value, oldValue)) { // some standard fuzzy equal
        if (atomicFloat.compare_exchange_weak(value, newValue)) {
            return true;
        }
    }
```

fwyzard:
The comparison in a CAS loop _relies_ by construction on bitwise identical values, not on "the values are close enough".

IMHO this should not be a warning, and definitely never be an error.

mehmetyusufoglu:
@SimeonEhrig  For gcc and clang,  it is easy to turn float comparison error OFF (by  adding -Wno-float-equal) but what can be done for Windows?

fwyzard:
Windows does not seem to complain about this:
image

psychocoderHPC:
We should silent this warning for the coresponding compilers via 

```
#    if BOOST_COMP_CLANG
#        pragma clang diagnostic push
#        pragma clang diagnostic ignored "-Wexit-time-destructors"
#    endif
...
#    if BOOST_COMP_CLANG
#        pragma clang diagnostic pop
#    endif
```

This code is perfromance critical and it is common in scientific software to compare floating point values with `==`.

psychocoderHPC:
We can also silent this warning here

https://github.com/alpaka-group/alpaka/blob/develop/test/common/devCompileOptions.cmake

psychocoderHPC:
note: unsafe comparison of floating point values is used for OpenMP atomics t00

https://github.com/alpaka-group/alpaka/blob/5f3b91c07ebe0e0eca4f16c65a06d30ca4fd74b0/include/alpaka/atomic/AtomicOmpBuiltIn.hpp#L270-L273

chillenzer:
I would also go for the narrow version. Alternatively, one could make this more explicit by defining a function `bitwiseIdentical<T>` such that the reader of the code knows that this is not an unsafe floating point comparison of some numerical values but a bitwise "nothing has changed with this piece of memory". That function template would, of course, specialise to the same code but one could then use it everywhere and turn off that warning in a single place only.

fwyzard:
> Alternatively, one could make this more explicit by defining a function `bitwiseIdentical<T>` such that the reader of the code knows that this is not an unsafe floating point comparison of some numerical values but a bitwise "nothing has changed with this piece of memory". That function template would, of course, specialise to the same code but one could then use it everywhere and turn off that warning in a single place only.

No.

C++ already provides this feature, it's called `operator==`.

fwyzard:
Unless there are concerns about NaN ?

psychocoderHPC:
> Unless there are concerns about NaN ?

IMO that's not an issue.

fwyzard:
The two failures are a bit puzzling.

~~On Windows, the cl-2022 debug build works, while the cl-2022 release build compiles and fails with a segmentation fault at runtime. To make it more puzzling, it looks like many (all ?) workflows fail, even those that do not use any atomic operation.~~

It looks like the Windows cl-2022 release build has problems since a while, unrelated to this PR.

On Linux, so far only the NVCC 11.2 / clang 9 build fails (in release mode) - while the NVCC 11.8 / clang 9 build works (in debug mode).

@psychocoderHPC would it be OK to disable the AtomicAtomicRef path on ~~MSVC 2022 and~~ clang 9 ?
@SimeonEhrig do you think we could add a more recent windows compiler to the tests ?

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -219,7 +219,23 @@ namespace alpaka
                 T result;
                 do
                 {
+#    if BOOST_COMP_GNUC
+#        pragma GCC diagnostic push
```
mehmetyusufoglu, line 223: `GCC` is understood by _clang_ compiler hence this would work (similar to usage at /alpaka/offset/Traits.hpp:42):

```
#if BOOST_COMP_CLANG || BOOST_COMP_GNUC
#    pragma GCC diagnostic push
#    pragma GCC diagnostic ignored "-Wfloat_equal"
#endif
         result = ((old == compare) ? value : old);
#if BOOST_COMP_CLANG || BOOST_COMP_GNUC
#endif
```

fwyzard: I see, thanks for the suggestion !

SimeonEhrig:
> @SimeonEhrig do you think we could add a more recent windows compiler to the tests ?

Yes, we can upgrade the MS compiler. Because of lacking Windows developers we decided that is fine to only support the latest version. Unfortunately Bernhard left us, which was the last person with Windows experience. Therefore nobody keep the Windows CI up to date.

fwyzard:
I'm confused, the nvcc 11/clang 10 test used to work, and now it failed... I'll try rerunning it, just in case:

image

fwyzard:
OK, let's try disabling `boost::atomic_ref` also for clang 10...

psychocoderHPC:
I remove the CI job ` windows_cl-2022_release ` from the required test list.

fwyzard:
> I remove the CI job `windows_cl-2022_release` from the required test list.

Thank you @psychocoderHPC.

That's now the only failing test - could you go ahead and merge the PR ?

psychocoderHPC (approved):

PR #2293, opened 2024-06-16
An example using MdSpan to pass 2D data

mehmetyusufoglu:
A new example is prepared by converting Convolution2D example. conv2DWithMdspan example. 

Uses mdspan data structure for input matrix, input filter matrix and output matrix. Hence pitch values are not needed to be used. Cleaner than the other version.

- [x] should take all accs into account equal to https://github.com/alpaka-group/alpaka/pull/2280


example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
```diff
@@ -0,0 +1,221 @@
...
+        }
+    }
+};
+
+auto FuzzyEqual(float a, float b) -> bool
+{
+    return std::fabs(a - b) < std::numeric_limits<float>::epsilon() * 1000.0f;
+}
+
+auto main() -> int
```
psychocoderHPC: This example should test all accelerators equal to https://github.com/alpaka-group/alpaka/pull/2280

We need to rebase and update this PR before it can be merged, the goal is to remove the ExampleDefaultAcc asap.

psychocoderHPC (changes requested):

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
mehmetyusufoglu: Yes, doing now. ... (I saw, the related PR merged.)

mehmetyusufoglu: Ok done. Tested locally, good work thanks.

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
```diff
@@ -0,0 +1,243 @@
...
+                {
+                    // Position of input matrix element to be multiplied with the corresponding element at filter
+                    auto const exactRow = static_cast<int32_t>(row) - filterWidth / 2 + fRow;
+                    auto const exactCol = static_cast<int32_t>(col) - filterWidth / 2 + fCol;
+                    if(exactRow >= 0 && exactRow < matrixHeight && exactCol >= 0 && exactCol < matrixWidth)
+                    {
+                        pValue += filter(fRow, fCol) * input(exactRow, exactCol);
+                    }
+                }
+                output(row, col) = pValue;
```
fwyzard: shouldn't this be outside the loop on `fRow` ?

mehmetyusufoglu: Yes, fixed.  (Good catch. It was not changing the result therefore can not be detected.)

fwyzard:
Looks good to me, though I've never used `mdspan` yet, so I may be missing something.

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
```diff
@@ -0,0 +1,243 @@
...
+    //! \param filter Filter-matrix
+
+    template<typename TAcc, typename MdSpan>
+    ALPAKA_FN_ACC auto operator()(TAcc const& acc, MdSpan const input, MdSpan output, MdSpan const filter) const
+        -> void
+    {
+        auto matrixWidth = input.extent(0);
+        auto matrixHeight = input.extent(1);
+        // Filter matrix is square
+        auto filterWidth = filter.extent(0);
```
fwyzard:
```suggestion
        auto filterWidth = static_cast<int32_t>(filter.extent(0));
```
and remove the `static_cast` below ?

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+        // Get thread index, the center of filter-matrix is positioned to the item on this index.
+        auto const [row, col] = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
```
fwyzard: Is there a way to force these to be `int32_t` and remove the `static_cast` below ?

SimeonEhrig:
> Looks good to me, though I've never used `mdspan` yet, so I may be missing something.

The usage of `mdspan` looks fine for me. At least, this is how I use it.

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
```diff
@@ -0,0 +1,243 @@
...
+    using QueueAcc = alpaka::Queue<DevAcc, alpaka::NonBlocking>;
+
+    using DataType = float;
+    static constexpr Idx filterWidth = 5;
+    static constexpr Idx matrixWidth = 128;
+    static constexpr Idx matrixHeight = 128;
+
+    static_assert(
+        alpaka::Dim<DevAcc>::value == 2u,
+        "The accelerator used for the Alpaka Kernel has to be 2 dimensional!");
```
fwyzard: This seems redundant here, given that the `DevAcc` has a dimension of 2 by construction.

fwyzard: Though it could make sense to have this check inside the kernel.

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
```diff
@@ -0,0 +1,243 @@
...
+        alpaka::experimental::getMdSpan(bufFilterAcc));
+
+    // Allocate memory on host to receive the resulting matrix as an array
+    auto resultGpuHost = alpaka::allocBuf<DataType, Idx>(devHost, extent);
+    // Copy result from device memory to host
+    alpaka::memcpy(queueAcc, resultGpuHost, outputDeviceMemory, extent);
+    alpaka::wait(queueAcc);
+
+    //  Print results
+    std::string const kernelType{"ConvolutionKernelMdspan2D"};
```
SimeonEhrig: Why is the kernel name an extra variable?

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
mehmetyusufoglu: For greater than zero check i had to convert to int32, i can change the loop borders but this is cleaner I think.

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
mehmetyusufoglu: done, thanks.

example/conv2DWithMdspan/src/conv2DWithMdspan.cpp
mehmetyusufoglu: ok, directly assigned to int32_t and removed all static_casts lets see if there would be a compiler warning or not.

SimeonEhrig (approved):

psychocoderHPC (approved):

PR #2294, opened 2024-06-17
fix clang9/nvcc11.2 boost bug

psychocoderHPC:
This PR fix current CI issues with clang + nvcc and disable workarounds for C++20 and newer.

```
/boost/boost/atomic/detail/bitwise_cast.hpp(113): error: type name is not allowed
```

- disable atomic_ref for clang<12&nvcc<11.6

The second commit fixes Windows compile issues those popped up today, maybe due to another MSVC update in the github workflow windows containers.

Issues introduced with #2288 where we enabled atomic ref after we introduced a bug that it was always disabled.

PR #2295, opened 2024-06-17
Fix memcpy warning

chillenzer:
Fixes a compiler warning about `std::string` that apparently happens when a `std::string` is added to a normal C-string. See https://gcc.gnu.org/bugzilla/show_bug.cgi?id=105651. It was also found in https://github.com/cms-sw/cmssw/issues/44096.

Sorry, there are two tiny unrelated formatting changes because pre-commit didn't allow me to push otherwise. 

- [x] rebase against #2294 after V is merged to solve CI issues

### DIFF
--- a/include/alpaka/core/UniformCudaHip.hpp
+++ b/include/alpaka/core/UniformCudaHip.hpp
@@ -61,7 +61,11 @@ namespace alpaka::uniform_cuda_hip::detail
             if(std::find(std::cbegin(ignoredErrorCodes), std::cend(ignoredErrorCodes), error)
                == std::cend(ignoredErrorCodes))
             {
-                rtCheck<TApi, TThrow>(error, ("'" + std::string(cmd) + "' returned error ").c_str(), file, line);
+                rtCheck<TApi, TThrow>(
+                    error,
+                    (std::string("'") + std::string(cmd) + "' returned error ").c_str(),
+                    file,
+                    line);
             }
             else
             {
--- a/include/alpaka/core/UniformCudaHip.hpp
+++ b/include/alpaka/core/UniformCudaHip.hpp
@@ -61,11 +61,8 @@ namespace alpaka::uniform_cuda_hip::detail
             if(std::find(std::cbegin(ignoredErrorCodes), std::cend(ignoredErrorCodes), error)
                == std::cend(ignoredErrorCodes))
             {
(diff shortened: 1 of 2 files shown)


psychocoderHPC (dismissed):

psychocoderHPC:
@chillenzer and I observed the same warning in a prototype for the upcoming mallocMC change. With this fix the warning get resolved.

include/alpaka/core/UniformCudaHip.hpp
```diff
@@ -61,7 +61,11 @@ namespace alpaka::uniform_cuda_hip::detail
             if(std::find(std::cbegin(ignoredErrorCodes), std::cend(ignoredErrorCodes), error)
                == std::cend(ignoredErrorCodes))
             {
-                rtCheck<TApi, TThrow>(error, ("'" + std::string(cmd) + "' returned error ").c_str(), file, line);
+                rtCheck<TApi, TThrow>(
+                    error,
+                    (std::string("'") + std::string(cmd) + "' returned error ").c_str(),
```
fwyzard: Potentially a dumb question: is the temporary string guaranteed to be valid ?

Isn't it safer to write
```c++
{
  std::string msg = std::string("'") + std::string(cmd) + "' returned error ";
  rtCheck<TApi, TThrow>(error, msg, file, line);
}
```
?

fwyzard: Also
```c++
'"'s + std::string(cmd) + "' returned error "s
```
would look nicer :-)

Can we add
```c++
using namespace std::literals;
```
inside the `namespace alpaka::uniform_cuda_hip::detail` namespace ?

psychocoderHPC:
please rebase against dev to get the fix from #2294 in

include/alpaka/core/UniformCudaHip.hpp
chillenzer: The lifetime of a temporary ends at the end of the expression, so practically it lives until the next `;`: "All temporary objects are destroyed as the last step in evaluating the [full-expression](https://en.cppreference.com/w/cpp/language/expressions#Full-expressions) that (lexically) contains the point where they were created" (from [cppreference](https://en.cppreference.com/w/cpp/language/lifetime)).

psychocoderHPC (approved):

PR #2261, opened 2024-05-01
Make all Examples use getValidWorkDivForKernel function [II]

mehmetyusufoglu:
Replaces the `getValidWorkDiv` function with `getValidWorkDivForKernel` which take into account the kernel in calculating workdiv.

- [x] This PR will be merged after #2251. Includes the changes of #2251.

This PR should be merged together with a Next PR: Change exec and task creation function. They will use KernelBundle as a single argument not KernelObject + Args... Otherwise kernel arguments would be used in kernel bundle and exec 2 times...

PR #2299, opened 2024-06-24
Make babelstream runnable and test for different Accs

mehmetyusufoglu:
babelstream benchmarking was creating errors and had unnecessary parts (because it was developed as an extention of [the test at Bristol Uni. ](https://github.com/UoB-HPC/BabelStream))

- Unnecesssary command line options are removed, 2 command line custom arguments are available

 * Run with Custom arguments:
  `./babelstream  --array-size=1280000 --number-runs=10`
 * Run with Catch2 arguments:
`  ./babelstream --success`
`./babelstream -r a.xml`
 * Run with Custom and catch2 arguments together:
`  ./babelstream  --success --array-size=1280000 --number-runs=10`
 * Run to list custom and catch2 arguments:
`  ./babelstream  --help`  (or  -h or  -?)

- Babelstream has 5 simple kernels for vector operations. **Add, Mul, Copy, Triad and Dot product**.  And there is an Init kernel. 

- RunTimes are measured by using std::chrono 

- RESULTS for `./babelstream --array-size=33554432 --number-runs=100` are below. They are equal ( or slightly better) compared to previous implementation which runs 100 times and uses minimum time as well.
 
```

AcceleratorType:AccCpuSerial<1,unsigned int>
NumberOfRuns:100
Precision:double
DataSize(items):33554432
DeviceName:13th Gen Intel(R) Core(TM) i7-1360P
WorkDivInit :{gridBlockExtent: (33554432), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (33554432), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (33554432), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (33554432), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (33554432), blockThreadExtent: (1), threadElemExtent: (1)}
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 InitKernel      12.626          0.042521 0.044909 0.045561  536.87 
 CopyKernel      21.185          0.025342 0.027276 0.025485  536.87 
 MultKernel      21.357          0.025138 0.026368 0.025286  536.87 
 AddKernel       25.272          0.031865 0.041926 0.032137  805.31 
 TriadKernel     25.174          0.03199  0.033468 0.032193  805.31 



AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:100
Precision:double
DataSize(items):33554432
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 InitKernel      62.451          0.0085967 0.0086461 0.0086267  536.87 
 CopyKernel      89.58           0.0059932 0.0060739 0.0060465  536.87 
 MultKernel      89.359          0.006008 0.0060836 0.0060491  536.87 
 AddKernel       90.551          0.0088934 0.0089586 0.008937  805.31 
 TriadKernel     90.61           0.0088876 0.0089629 0.0089369  805.31 
 DotKernel       93.256          0.0057569 0.0058079 0.0057916  536.87 

```
Full output file: 
 
[babelStreamOut.txt](https://github.com/user-attachments/files/16787245/babelStreamOut.txt)

**Old implementation values** for 4 kernels 100 runs in the following image: 
 
Screenshot from 2024-07-31 15-58-20


benchmarks/babelstream/src/benchmarkTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+    std::string dataTypeStr;
+    if(std::is_same<DataType, float>::value)
+    {
+        dataTypeStr = "single-precision";
+    }
+    else if(std::is_same<DataType, double>::value)
+    {
+        dataTypeStr = "double-precision";
+    }
+    std::cout << "Data type: " << typeid(DataType).name() << " " << dataTypeStr << std::endl;
```
SimeonEhrig: Do you know, if there is some reporting function to add the information as key value pair to the output? At the moment, it simply captures the stdout and write a string to the xml, which make it hard to parse.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<Catch2TestRun name="babelstream" rng-seed="1921823939" xml-format-version="2" catch2-version="3.3.2">
  <TestCase name="Babelstream Bechmark by Alpaka Kernel - Single P. - TestAccs1D - 0" tags="[benchmark-test]" filename="/home/simeon/projects/alpaka/benchmarks/babelstream/src/benchmarkTest.cpp" line="418">
    <BenchmarkResults name="CopyKernel" samples="100" resamples="100000" iterations="1" clockResolution="32.5898" estimatedDuration="1.62794e+07">
      <!-- All values in nano seconds -->
      <mean value="162549" lowerBound="162235" upperBound="163277" ci="0.95"/>
      <standardDeviation value="2305.52" lowerBound="1253.62" upperBound="4617.02" ci="0.95"/>
      <outliers variance="0.0750291" lowMild="0" lowSevere="0" highMild="1" highSevere="8"/>
    </BenchmarkResults>
    <BenchmarkResults name="AddKernel" samples="100" resamples="100000" iterations="1" clockResolution="32.5898" estimatedDuration="2.07714e+07">
      <!-- All values in nano seconds -->
      <mean value="208480" lowerBound="207810" upperBound="209855" ci="0.95"/>
      <standardDeviation value="4688.39" lowerBound="2825.43" upperBound="8760.11" ci="0.95"/>
      <outliers variance="0.161071" lowMild="0" lowSevere="0" highMild="1" highSevere="6"/>
    </BenchmarkResults>
    <BenchmarkResults name="MulKernel" samples="100" resamples="100000" iterations="1" clockResolution="32.5898" estimatedDuration="1.19507e+07">
      <!-- All values in nano seconds -->
      <mean value="122044" lowerBound="121594" upperBound="122670" ci="0.95"/>
      <standardDeviation value="2676.42" lowerBound="2061.12" upperBound="3957.46" ci="0.95"/>
      <outliers variance="0.151683" lowMild="0" lowSevere="0" highMild="1" highSevere="1"/>
    </BenchmarkResults>
    <BenchmarkResults name="TriadKernel" samples="100" resamples="100000" iterations="1" clockResolution="32.5898" estimatedDuration="2.11882e+07">
      <!-- All values in nano seconds -->
      <mean value="212046" lowerBound="211493" upperBound="212948" ci="0.95"/>
      <standardDeviation value="3544.51" lowerBound="2546.6" upperBound="5707.04" ci="0.95"/>
      <outliers variance="0.094408" lowMild="0" lowSevere="0" highMild="1" highSevere="9"/>
    </BenchmarkResults>
    <OverallResult success="true" skips="0">
      <StdOut>
========================
Time:2024-07-01 14:21:45

Using alpaka accelerator: AccCpuSerial&lt;1,unsigned int>
Data type: f single-precision
Data size: 1048576 elements
Work division: {gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
      </StdOut>
    </OverallResult>
  </TestCase>
```

benchmarks/babelstream/src/benchmarkTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+        }
+        else
+        {
+            static_assert(
+                std::is_floating_point_v<T> || std::is_integral_v<T>,
+                "FuzzyEqual<T> is only supported for integral or floating-point types.");
+        }
+    }
+
+    std::string getCurrentTimestamp();
```
SimeonEhrig: Why the forward declaration?

benchmarks/babelstream/src/benchmarkTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+    //! \tparam TAcc The accelerator environment to be executed on.
+    //! \tparam T The data type
+    //! \param acc The accelerator to be executed on.
+    //! \param a Pointer for vector a
+    //! \param c Pointer for result vector c
+    template<typename TAcc, typename T>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, T const* a, T* c) const
+    {
+        using Dim = alpaka::Dim<TAcc>;
+        if constexpr(Dim::value == 1)
```
SimeonEhrig: would be a static_assert() bettert? If the dimension is not 1, the kernel simply do nothing.

benchmarks/babelstream/src/benchmarkTest.cpp
```diff
@@ -0,0 +1,434 @@
...
+    // Verify the results
+    for(Idx i = 0; i < arraySize; i += 1000)
+    {
+        REQUIRE(FuzzyEqual(bufHostOutput[static_cast<Idx>(i)], static_cast<DataType>(3)));
+    }
+
+    // Since the kernel needs more than 1 threads per block only Gpu backends are used.
+    // A reference implementation can be found here:
+    // https://github.com/UoB-HPC/BabelStream/blob/main/src/hip/HIPStream.cpp
+    if constexpr(alpaka::accMatchesTags<Acc, alpaka::TagGpuCudaRt, alpaka::TagGpuHipRt, alpaka::TagGpuSyclIntel>)
```
SimeonEhrig: please use `alpaka::isMultiThreadAcc<Acc>` instead.

benchmarks/babelstream/src/benchmarkTest.cpp
mehmetyusufoglu: then wont the test fail because tests are triggered by different types? Let me try.

benchmarks/babelstream/src/benchmarkTest.cpp
mehmetyusufoglu: There was a warning at one compiler  "`warning: no previous declaration for ‘std::string getCurrentTimestamp()’ [-Wmissing-declarations]`" therefore i added a declaration. Alternatively, I can make the function `static` as well. Somehow one of the compilers forced for declaration.

benchmarks/babelstream/src/benchmarkTest.cpp
mehmetyusufoglu: ok, i had set the test type at the test. You are right.

benchmarks/babelstream/src/benchmarkTest.cpp
mehmetyusufoglu: I made the function static. Like functions in some other tests: e.g. `alpaka/test/unit/mem/buf/src/BufTest.cpp`

benchmarks/babelstream/src/benchmarkTest.cpp
mehmetyusufoglu: Yes I will use "INFO" macro of catch2. Good idea thanks. It turned out to be `-r json` also works for `catch2 3.5.2`. I am doing it for json, more readable.

mehmetyusufoglu: json reporter seem to not include benchmark results only records REQUIRE results etc. XML reporter prints benchmark results as expected.

mehmetyusufoglu: Only `WARN` has worked to log output other than using `cout`. I will use `WARN`.   Because others only workd when there is REQUIRE/CHECK and it fails.
```
TEST_CASE("TESTCASE tesing trial", "[testcase-test]")
{
    std::string dataType = "floating-point";
    std::string dataSize =   std::to_string(100000);
    int arraySize=1234;
    WARN(dataType);  // This will be included as custom XML in the report
    INFO("\n normal info Data type is" << dataType);
    UNSCOPED_INFO("unscoped Data type is" << dataType);
    INFO("info example");
    UNSCOPED_INFO("unscoped example");
    INFO("normal info array size is" << arraySize);
    UNSCOPED_INFO("unscoped array size is" << arraySize);
    INFO("normal info data size is" << dataSize);
    UNSCOPED_INFO("unscoped data size is" << dataSize);
    CAPTURE(dataSize);   
    CAPTURE(dataType);
  // !!!!
    CHECK(arraySize!=1234); // If this fails all INFO, capture and UNSCOPED_INFO prints. Otherwise only WARN prints
    SUCCEED("SUCCEED");  
}
```

mehmetyusufoglu: This will need additional operation to extract data from fields. May be I should use cout output directly and convert it to json by a script instead of dealing with this...
```
   <Warning filename="/home/yusufo81/projects/alpaka-dir/alpaka/benchmarks/babelstream/src/benchmarkTest.cpp" line="146">
      TimeStamp:2024-07-09 15:06:59
    </Warning>
    <Warning filename="/home/yusufo81/projects/alpaka-dir/alpaka/benchmarks/babelstream/src/benchmarkTest.cpp" line="147">
      DataSize:1048576
    </Warning>
    <Warning filename="/home/yusufo81/projects/alpaka-dir/alpaka/benchmarks/babelstream/src/benchmarkTest.cpp" line="148">
      DataSize:single-precision
    </Warning>
```

benchmarks/babelstream/CMakeLists.txt
```diff
@@ -19,9 +19,36 @@ if(NOT TARGET alpaka::alpaka)
...
+if(alpaka_CI)
+    # Only run for release builds since this is a benchmark
+    if(CMAKE_BUILD_TYPE STREQUAL "Release")
+        if(${CMAKE_SYSTEM_NAME} STREQUAL "Linux" OR ${CMAKE_SYSTEM_NAME} STREQUAL "Darwin")
+           # To prevent CI test timeout, the number of samples needs to be limited. In the catch2 benchmarking; the code is run samples x iterations times.
+           add_test(NAME ${_TARGET_NAME} COMMAND sh -c "${CMAKE_CURRENT_BINARY_DIR}/${_TARGET_NAME} ${_alpaka_TEST_OPTIONS}  --benchmark-samples 1 -r xml | tee BabelStreamBenchmarkResults.xml")
+        elseif(${CMAKE_SYSTEM_NAME} STREQUAL "Windows")  
+           # To prevent CI test timeout, the number of samples needs to be limited. In the catch2 benchmarking; the code is run samples x iterations times.
+           set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER benchmarks/babelstream)
+           add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} --benchmark-samples 1)
```
SimeonEhrig: Use this also for Linux and Darwin

mehmetyusufoglu: why ? Don't we need xml?

benchmarks/babelstream/CMakeLists.txt
```diff
@@ -19,9 +19,36 @@ if(NOT TARGET alpaka::alpaka)
...
+        endif()
```
SimeonEhrig:
```suggestion
       set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER benchmarks/babelstream)
       add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} --benchmark-samples 1 -r xml)
```

benchmarks/babelstream/CMakeLists.txt
mehmetyusufoglu: ok, done. ( XML output file is not generated currently, because CI is not a proper environment for running the benchmark to measure performance.  When the benchmark is run exclusively, XML file could be generated by changing cmake or while running the benchmark from command line)

benchmarks/babelstream/CMakeLists.txt
```diff
@@ -19,9 +19,30 @@ if(NOT TARGET alpaka::alpaka)
...
+target_link_libraries(
+    ${_TARGET_NAME}
+    PRIVATE common)
+
+#Run as a ctest
+if(alpaka_CI)
+    # Only run for release builds since this is a benchmark
+    if(CMAKE_BUILD_TYPE STREQUAL "Release")
+       set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER benchmarks/babelstream)
+       add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} --benchmark-samples 1 -r xml)
```
SimeonEhrig:
```suggestion
       add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME} --benchmark-samples 1 -r xml > /dev/null)
```

Otherwise the xml is displayed in the job output, which does not give use a benefit.

benchmarks/babelstream/src/babelStreamCommon.hpp
```diff
@@ -0,0 +1,144 @@
...
+            static_assert(
+                std::is_floating_point_v<T> || std::is_integral_v<T>,
+                "FuzzyEqual<T> is only supported for integral or floating-point types.");
+        }
+    }
+
+    //!
+    //! \brief getCurrentTimestamp gets the time-stamp and returns it as a string
+    //!
+    [[maybe_unused]] static std::string getCurrentTimestamp()
```
SimeonEhrig, line 130: Is this really required? Is see nothing, which should trigger the warning. Also several other places in the code, where `[[maybe_unused]]` is used.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -0,0 +1,344 @@
...
+    template<typename TAcc, typename T>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, T const* a, T const* b, T* c) const
+    {
+        const T scalar = static_cast<T>(scalarVal);
+        auto const i = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
+        c[i] = a[i] + scalar * b[i];
+    }
+};
+
+struct toXML
```
SimeonEhrig: looks unused

benchmarks/babelstream/CMakeLists.txt
mehmetyusufoglu: Even though it is not run exclusively it is still an indicator ?

benchmarks/babelstream/src/babelStreamCommon.hpp
mehmetyusufoglu: For some accelerator types the test is not called. I had compile errors. Therefore it was necessary.

benchmarks/babelstream/CMakeLists.txt
SimeonEhrig: That's true, but I expected that if something is going wrong with the CI, the babelstream executable returns a non 0 zero and die CI will fail. In this case we can remove temporary the `> /dev/null` to display the output.

Also I think it makes more sense, to create a custom job for this https://github.com/alpaka-group/alpaka/pull/2313 and checks the json output afterwards. Therefore the regular jobs should only check, if the benchmark is running with the XML reporter, but don't care about the output.

mehmetyusufoglu: ok, i committed your suggestion.

mehmetyusufoglu: tests started failing after adding >dev/null , local test also fails

SimeonEhrig: I opened a PR (#2322), which suppresses ctest output in general, if the test didn't failed. Therefore you do not need to redirect the output to `/dev/null`.

benchmarks/babelstream/CMakeLists.txt
```diff
@@ -19,9 +19,30 @@ if(NOT TARGET alpaka::alpaka)
...
+
+target_link_libraries(
+    ${_TARGET_NAME}
+    PRIVATE common)
+
+#Run as a ctest
+if(alpaka_CI)
+    # Only run for release builds since this is a benchmark
+    if(CMAKE_BUILD_TYPE STREQUAL "Release")
+       set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER benchmarks/babelstream)
```
SimeonEhrig: please move the line before `if(alpaka_CI)`. The `set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER benchmarks/babelstream)` should be done independent of if the test is executed in the CI or not.

benchmarks/babelstream/src/babelStreamDotKernelTest.cpp
```diff
@@ -0,0 +1,206 @@
...
+
+    auto bufHostOutput = alpaka::allocBuf<DataType, Idx>(devHost, arraySize);
+    auto bufAccOutput = alpaka::allocBuf<DataType, Idx>(devAcc, arraySize);
+    // Define the work division
+    using Vec = alpaka::Vec<Dim, Idx>;
+    auto const elementsPerThread = Vec::all(static_cast<Idx>(1));
+    auto const threadsPerGrid = Vec::all(arraySize);
+
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+    WorkDiv const workDiv = alpaka::getValidWorkDiv<Acc>(
```
psychocoderHPC: please use getValidWorkDivKErnel

psychocoderHPC (changes requested):
please rebase to use getValidWorkDivKErnel

benchmarks/babelstream/src/babelStreamDotKernelTest.cpp
mehmetyusufoglu: this kernel is not a part of Kokkos stream benchmarks, it is taken from "UoB-HPC/BabelStream" example when the babelstream example of alpaka is prepared initially. This kernel calculates block level dot product. 
https://github.com/UoB-HPC/BabelStream/blob/main/src/cuda/CUDAStream.cu

benchmarks/babelstream/CMakeLists.txt
mehmetyusufoglu: I removed the dev/null

benchmarks/babelstream/src/babelStreamDotKernelTest.cpp
psychocoderHPC: `getValidWorkDiv` is removed with https://github.com/alpaka-group/alpaka/pull/2261 so you can not use it anymore!

psychocoderHPC:
Please check that the new implementation has no performance degradation compared to the current dev branch.

psychocoderHPC (changes requested):
please provide with the new implementation the same output on std::out that we have in the old implementation, to ensure we have human readable output in the default bableStream format.

mehmetyusufoglu:
> please provide with the new implementation the same output on std::out that we have in the old implementation, to ensure we have human readable output in the default bableStream format.

It is human readable, when i print using cout at CI it is fine; but XML reporter of catch2 converts all cout outputs to a single XML field. To process that data  i had to reformat information as below so that it can be read as key:value pairs from that single field. May be I can add tabs/spaces to make it seen better.

```
  AcceleratorType:AccCpuSerial<1,unsigned int>
  TimeUnitForXMLReport:Nano Seconds
  KernelNames:InitKernel,CopyKernel,ScalingKernel,AddKernel,TriadKernel
  KernelDataUsageValues(KByte):8192,8192,8192,12288,12288
  KernelBandwidths(KBytes/ms):36341.9,36857.2,37809.7,38859.3,36311
```

mehmetyusufoglu:
> > Please check that the new implementation has no performance degradation compared to the current dev branch.
> 
> There is a decrease from 86-87 to 83 GB/sec. Now I am going to try without using BENCHMARK_ADVANCED (of catch2) and running both the same number of times ( ie 100).

This problem still persists, i will try more.

mehmetyusufoglu:
> > > Please check that the new implementation has no performance degradation compared to the current dev branch.
> > 
> > 
> > There is a decrease from 86-87 to 83 GB/sec. Now I am going to try without using BENCHMARK_ADVANCED (of catch2) and running both the same number of times ( ie 100).
> 
> This problem still persists, i will try more.

Ok, solved. Performance is the same as before this PR.  it was an issue of calculation unit. [Previous implementation was calculating bandwidth using 1MB=1E+6 bytes not pow(2,20) bytes.]

benchmarks/babelstream/src/babelStreamDotKernelTest.cpp
```diff
@@ -0,0 +1,225 @@
...
+    DataType const* sumPtr = std::data(bufHostSumPerBlock);
+    auto const result = std::reduce(sumPtr, sumPtr + gridBlockExtent, DataType{0});
+    // Since vector values are 1, dot product should be identical to arraySize
+    REQUIRE(FuzzyEqual(static_cast<DataType>(result), static_cast<DataType>(arraySize)));
+
+    //
+    // Calculate Measured Bandwidth and Print All Benchmark results
+    //
+    // Find the average of the durations array, currently not used, but might be needed.
+    // auto const avgDurationOfKernelExec = findAverage(times);
```
psychocoderHPC: dead code?

psychocoderHPC (changes requested):

benchmarks/babelstream/src/babelStreamCommon.hpp
```diff
@@ -0,0 +1,433 @@
+#pragma once
+#include <boost/tokenizer.hpp>
```
psychocoderHPC: I am not sure if this is part of boost compile time header. 
If not this will introduce additional dependencies for alpaka, at least benchmarking. 
Please remove the include and use best stl to split strings, I assume this is the reason why it is included.

psychocoderHPC (changes requested):

benchmarks/babelstream/src/babelStreamCommon.hpp
psychocoderHPC: The CI is maybe passing but the reason is only that all boost packages are installed. I tried it on our development system HAL

```
spack load openpmd-api@0.15.2 %gcc@12.2.0 \
    ^adios2@2.9.2 \
    ^hdf5@1.14.3 \
    ^openmpi@4.1.5 +atomics +cuda cuda_arch=80 \
    ^python@3.11.6 \
    ^py-numpy@1.26.2
spack load boost@1.83.0 %gcc@12.2.0
```
error:
```
 fatal error: boost/tokenizer.hpp: No such file or directory
```

mehmetyusufoglu: yes i used stl removed boost tokenizer after seeing the CI fail. Now removed the header too.

psychocoderHPC:
@mehmetyusufoglu  The CMake file for the benchmark is missing to include catch in case you compile without `-DBUILD_TESTING=ON` and only with `-Dalpaka_BUILD_BENCHMARKS=ON` compiling fails due to missing catch2 headers.

psychocoderHPC:
~~The output on the terminal should looks like the old output. ~~
~~This helps external users if they compare this bable stream against other upstream implementations.~~
~~The user should not search in the code if we use min/max or average for the results.~~

Sry my fold, the output is nearly equal, only MB is now GB/s

mehmetyusufoglu:
> The output on the terminal should looks like the old output. This helps external users if they compare this bable stream against other upstream implementations. The user should not search in the code if we use min/max or average for the results.

Yes it is the same output. Additionally, i can add that we used minimum time as an extra comment. ok.

psychocoderHPC:
> > The output on the terminal should looks like the old output. This helps external users if they compare this bable stream against other upstream implementations. The user should not search in the code if we use min/max or average for the results.
> 
> Yes it is the same output. Additionally, i can add that we used minimum time as an extra comment. ok.

I updated my cmment above.

> i can add that we used minimum time as an extra comment. ok.

Is this different compared too the original code? 
  - If not not need to add the comment
  - If yes, it should show us ethe same as the old code

mehmetyusufoglu:
> > > The output on the terminal should looks like the old output. This helps external users if they compare this bable stream against other upstream implementations. The user should not search in the code if we use min/max or average for the results.
> > 
> > 
> > Yes it is the same output. Additionally, i can add that we used minimum time as an extra comment. ok.
> 
> I updated my cmment above.
> 
> > i can add that we used minimum time as an extra comment. ok.
> 
> Is this different compared too the original code? If not no need to add the comment

No, same as original. It was not stated in the output. Therefore i did not mention.  i will not state in the output as well so not change the code.

benchmarks/babelstream/src/babelStreamCommon.hpp
psychocoderHPC: I tested it, the include can be removed because it is not used

mehmetyusufoglu: removed yes, i found CI error in the morning and removed boost tokenizer usage; but forgot tto remove header. now removed.

mehmetyusufoglu:
> @mehmetyusufoglu The CMake file for the benchmark is missing to include catch in case you compile without `-DBUILD_TESTING=ON` and only with `-Dalpaka_BUILD_BENCHMARKS=ON` compiling fails due to missing catch2 headers.

-  I am investigating but that was not creating a problem in CI. Here everything passed. https://gitlab.com/hzdr/crp/alpaka/-/commit/472bfed5d7045b6a94b04e9913b252ce1fd5e3dc
- Locally no problem observed.

mehmetyusufoglu:
> > > The output on the terminal should looks like the old output. This helps external users if they compare this bable stream against other upstream implementations. The user should not search in the code if we use min/max or average for the results.
> > 
> > 
> > Yes it is the same output. Additionally, i can add that we used minimum time as an extra comment. ok.
> 
> I updated my cmment above.
> 
> > i can add that we used minimum time as an extra comment. ok.
> 
> Is this different compared too the original code?
> 
>     * If not not need to add the comment
> 
>     * If yes, it should show us ethe same as the old code

It is the same with the previous code's output. Besides there are 4 additional fields as info: 1. Acc type, 2. data item size, 3. workdiv and 4. correct data-throughput for each kernel.

mehmetyusufoglu:
> > @mehmetyusufoglu The CMake file for the benchmark is missing to include catch in case you compile without `-DBUILD_TESTING=ON` and only with `-Dalpaka_BUILD_BENCHMARKS=ON` compiling fails due to missing catch2 headers.
> 
>     * I am investigating but that was not creating a problem in CI. Here everything passed. https://gitlab.com/hzdr/crp/alpaka/-/commit/472bfed5d7045b6a94b04e9913b252ce1fd5e3dc
> 
>     * Locally no problem observed.

All tests pass at gitlab and github. No problem.

psychocoderHPC:
@mehmetyusufoglu  please amend and force push this PR again, the CI is fixed

psychocoderHPC:
> -Dalpaka_BUILD_BENCHMARKS=ON

The problem still exists. As I wrote if you only enable benchmarking catch will not found if you not provide an installed catch version externally.

```
cmake ~/workspace/alpaka -Dalpaka_ACC_GPU_CUDA_ENABLE=ON -Dalpaka_ACC_GPU_CUDA_ONLY_MODE=ON -Dalpaka_BUILD_BENCHMARKS=ON
```

```
alpaka/benchmarks/babelstream/src/babelStreamMainTest.cpp:3:10: fatal error: catch2/catch_session.hpp: No such file or directory
    3 | #include "catch2/catch_session.hpp"
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [benchmarks/babelstream/CMakeFiles/babelstream.dir/build.make:77: benchmarks/babelstream/CMakeFiles/babelstream.dir/src/babelStreamMainTest.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:158: benchmarks/babelstream/CMakeFiles/babelstream.dir/all] Error 2

```

psychocoderHPC (changes requested):
https://github.com/alpaka-group/alpaka/pull/2299#issuecomment-2342848212 must be fixed before we can merge this PR. 
This error will not show up in the CI because we enable tests and benchmarking in the CI always together.

mehmetyusufoglu:
> [#2299 (comment)](https://github.com/alpaka-group/alpaka/pull/2299#issuecomment-2342848212) must be fixed before we can merge this PR. This error will not show up in the CI because we enable tests and benchmarking in the CI always together.

@SimeonEhrig  my latest commit would not solve the issue i think. What is your opinion?

SimeonEhrig:
> > [#2299 (comment)](https://github.com/alpaka-group/alpaka/pull/2299#issuecomment-2342848212) must be fixed before we can merge this PR. This error will not show up in the CI because we enable tests and benchmarking in the CI always together.
> 
> @SimeonEhrig my latest commit would not solve the issue i think. What is your opinion?

Something get broken. If I run `cmake -Dalpaka_BUILD_BENCHMARKS=ON` or `cmake -DBUILD_TESTING=ON` I get the following error message:

```bash
CMake Error at thirdParty/CMakeLists.txt:42 (find_package):
  Could not find a package configuration file provided by "Catch2" (requested
  version 3.5.2) with any of the following names:

    Catch2Config.cmake
    catch2-config.cmake

  Add the installation prefix of "Catch2" to CMAKE_PREFIX_PATH or set
  "Catch2_DIR" to a directory containing one of the above files.  If "Catch2"
  provides a separate development package or SDK, be sure it has been
  installed.

```

This CMake path should be used, if the CMake variable `alpaka_USE_INTERNAL_CATCH2` is `OFF`. But setting `-Dalpaka_USE_INTERNAL_CATCH2=ON` also does not solve the problem.

CMakeLists.txt
```diff
@@ -48,7 +48,9 @@ option(alpaka_INSTALL_TEST_HEADER "Install headers of the namespace alpaka::test
 
 include(CMakeDependentOption)
 cmake_dependent_option(alpaka_CHECK_HEADERS "Check all alpaka headers as part of the tests whether they can be compiled standalone." OFF BUILD_TESTING OFF)
+cmake_dependent_option(alpaka_CHECK_HEADERS "Check all alpaka headers as part of the benchmarks whether they can be compiled standalone." OFF alpaka_BUILD_BENCHMARKS OFF)
 cmake_dependent_option(alpaka_USE_INTERNAL_CATCH2 "Use internally shipped Catch2" ON BUILD_TESTING OFF)
+cmake_dependent_option(alpaka_USE_INTERNAL_CATCH2 "Use internally shipped Catch2" ON alpaka_BUILD_BENCHMARKS OFF)
```
psychocoderHPC: I think you must merge this line and the line above see https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

If you add two lines the last line will have the highest priority.

CMakeLists.txt
```diff
@@ -48,7 +48,9 @@ option(alpaka_INSTALL_TEST_HEADER "Install headers of the namespace alpaka::test
 
 include(CMakeDependentOption)
 cmake_dependent_option(alpaka_CHECK_HEADERS "Check all alpaka headers as part of the tests whether they can be compiled standalone." OFF BUILD_TESTING OFF)
+cmake_dependent_option(alpaka_CHECK_HEADERS "Check all alpaka headers as part of the benchmarks whether they can be compiled standalone." OFF alpaka_BUILD_BENCHMARKS OFF)
```
psychocoderHPC: I think you must merge this line and the line above see https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html

If you add two lines the last line will have the highest priority.

psychocoderHPC (changes requested):

CMakeLists.txt
mehmetyusufoglu: -  I have added an additional local variable  as` (alpaka_BUILD_BENCHMARKS  OR BUILD_TESTING)`.

But; Benchmarks were locally not compilable and i got the error, 
`/home/yusufo81/projects/alpaka-dir/alpaka/benchmarks/babelstream/src/babelStreamMainTest.cpp:5:10: fatal error: alpaka/alpaka.hpp: No such file or directory
    5 | #include <alpaka/alpaka.hpp>
      |          ^~~~~~~~~~~~~~~~~~~
`

-  to solve this  i had to change **the condition** to call `add_subdirectory("test/") `at line 167. But with this change, setting benchmark cmake variable became the same thing of setting `BUILD_TESTING`. It is a kind of work-around.
` \# Only build the tests if alpaka is the top-level project and BUILD_TESTING or alpaka_BUILD_BENCHMARKS is ON
\# Note: alpaka_BUILD_BENCHMARKS is added with an OR besides BUILD_TESTING below.
\# Because the benchmark which uses catch2 is not compilable locally without adding test directory as below
\# or without setting BUILD_TESTING cmake variable to ON. See PR2299.` 

```
if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND (BUILD_TESTING OR alpaka_BUILD_BENCHMARKS))
    add_subdirectory("test/")
endif()
```

CMakeLists.txt
```diff
@@ -153,8 +164,11 @@ if(alpaka_BUILD_BENCHMARKS)
     add_subdirectory("benchmarks/")
 endif()
 
-# Only build the tests if alpaka is the top-level project and BUILD_TESTING is ON
-if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND BUILD_TESTING)
+# Only build the tests if alpaka is the top-level project and BUILD_TESTING or alpaka_BUILD_BENCHMARKS is ON
+# Note: alpaka_BUILD_BENCHMARKS is added with an OR besides BUILD_TESTING below.
+# Because the benchmark which uses catch2 is not compilable locally without adding test directory as below
+# or without setting BUILD_TESTING cmake variable to ON. See PR2299.
+if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME AND (BUILD_TESTING OR alpaka_BUILD_BENCHMARKS))
```
psychocoderHPC: we do not want to build testing if benchmark only is enabled.

psychocoderHPC:
I will provide a fix after lunch!

psychocoderHPC (approved):

PR #1815, opened 2022-10-19
Update to Catch2 v3.1.1

j-stephan:
See title.

Edit: This command was used to set author and email for the "Third Party" user:

```
git commit --amend --author="Third Party <alpaka@hzdr.de>" --no-message
```

PR #2300, opened 2024-06-24
Update catch2 version to 3.5.2

mehmetyusufoglu:
Update catch2 version from 3.3.2 to **3.5.2**

The latest version 3.6.0 generated compile errors. Therefore `3.5.2` is selected which is the latest one that does not create compile errors. Compile error at unit and integration tests of Alpaka is related with changes in  `catch2/internal/catch_decomposer.hpp` file of catch2 and observed only for **nvcc11.0** and **nvcc11.1** compilers. (https://gitlab.com/hzdr/crp/alpaka/-/jobs/7248169798) It seems nvcc mixes up` __nv_bool` and `bool`.

A fix of the at catch2 would be as in the diff file attached. (SOlution on the alpaka side is possible but would not worth for now)
[diff.txt](https://github.com/user-attachments/files/16074711/diff.txt)

Currently reporting the bug and the possible solution to catch2 and using 3.5.2 instead of catch 3.6.0 is preferred.


SimeonEhrig:
Please change the authorship of the commit, which updates the source code of Catch2. Here you can find how: https://github.com/alpaka-group/alpaka/pull/1815

mehmetyusufoglu:
> Please change the authorship of the commit, which updates the source code of Catch2. Here you can find how: #1815

Thanks,  i changed the author of the commit ( although committed with my user-name).

SimeonEhrig (approved):

Issue #2248, opened 2024-03-13
std::is_trivially_copyable constraint on kernel type


jasminmohnke:
While #2198 clears the issue we had with the check on kernel arguments being trivially copyable as discussed in #2195, unfortunately I am running into a comparable issue with the check done on the kernel itself in

https://github.com/alpaka-group/alpaka/blob/a4142d3feb7686d803e1ec5f25d7b2278337f455/include/alpaka/kernel/Traits.hpp#L269-L271

If, for instance, a user kernel has a member that is of a type such as described previously (#2195), std::is_trivially_copyable<TKernelFnObj> evaluates to false as well. A trait similarly to the one you introduced for the kernel argument would be helpful to have for the kernel as well.

psychocoderHPC:
Thanks for opening the issue, we will have a look into it.

psychocoderHPC:
fixed with #2302

PR #2302, opened 2024-06-25
add trait `IsKernelTriviallyCopyable`

AuroraPerego:
following the request in #2248, this adds the same `trait` introduced previously for kernel arguments also for the kernel itself, with the same approach used in #2198.

### DIFF
--- a/include/alpaka/kernel/Traits.hpp
+++ b/include/alpaka/kernel/Traits.hpp
@@ -248,6 +248,33 @@ namespace alpaka
         }
     } // namespace detail
 
+    //! Check if the kernel type is trivially copyable
+    //!
+    //! \attention In case this trait is specialized for a user type the user should be sure that the result of calling
+    //! the copy constructor is equal to use memcpy to duplicate the object. An existing destructor should be free
+    //! of side effects.
+    //!
+    //! The default implementation is true for trivially copyable types (or for extended lambda expressions for CUDA).
+    //!
+    //! @tparam T type to check
+    //! @{
+    template<typename T, typename = void>
+    struct IsKernelTriviallyCopyable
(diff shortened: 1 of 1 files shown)


include/alpaka/kernel/Traits.hpp
```diff
@@ -248,6 +248,33 @@ namespace alpaka
...
+    //! @tparam T type to check
+    //! @{
+    template<typename T, typename = void>
+    struct IsKernelTriviallyCopyable
+#if BOOST_COMP_NVCC
+        : std::bool_constant<
+              std::is_trivially_copyable_v<T> || __nv_is_extended_device_lambda_closure_type(T)
+              || __nv_is_extended_host_device_lambda_closure_type(T)>
+#else
+        : std::bool_constant<std::is_trivially_copyable_v<T>>
```
fwyzard:
```suggestion
        : std::is_trivially_copyable<T>
```

fwyzard (approved):

AuroraPerego:
I've no idea why the tests failed with rocm 5.7 and passed with 5.1 to 5.6 and 6.0 :(
The one that fails is `mathTest`, with `Idx i: 634 computed : (0.890482,-7237.58) vs expected: (0.893934,-7237.59)`.

psychocoderHPC:
> I've no idea why the tests failed with rocm 5.7 and passed with 5.1 to 5.6 and 6.0 :( The one that fails is `mathTest`, with `Idx i: 634 computed : (0.890482,-7237.58) vs expected: (0.893934,-7237.59)`.

looks like a small rounding issue, I would ignore it for now because it is hard to find. COuld be that we are to strict with the precision.

Issue #2279, opened 2024-05-27
Support CUDA 12.4 and 12.5


SimeonEhrig:

ichinii:
after updating to cuda 12.5 i get following error message when running my alpaka program:

`  what():  /home/ich/projects/rayx/Extern/alpaka/include/alpaka/mem/buf/uniformCudaHip/Copy.hpp(143) 'TApi::setDevice(m_iDstDevice)' A previous API call (not this one) set the error  : 'cudaErrorUnsupportedPtxVersion': 'the provided PTX was compiled with an unsupported toolchain.'!`

It is not clear to me where this error comes from and if this might be a issue that a future alpaka version would fix. Do you maybe have a guess?

At this point I am asking myself if it is expected, that new cuda versions will likely make alpaka applications unable to function correctly? In other words: Should one be aware of the alpaka cuda support, before updating cuda on a system?

fwyzard:
what GPU are you compiling for ?

psychocoderHPC:
@ichinii, please post the cmake command used to configure and compile.

Do you set `-DCMAKE_CUDA_ARCHITECTURES=52` where 52 is the sm level of your target GPU.

psychocoderHPC:
https://forums.developer.nvidia.com/t/provided-ptx-was-compiled-with-an-unsupported-toolchain-error-using-cub/168292 is saying you have a driver and toolchain mismatch. 

- Do you compiled and executed the app on the same system?
- Do you have more than one CUDA SDK on the system?

Most most likely you missed to update you system driver.

ichinii:
oh i am sorry. this is not an alpaka related issue. 

i noticed the problem first with an alpaka application, but turns out the issue also occurs in a little test setup, without alpaka.
sorry that i did not test and provide additional information. i was guessing that the issue was common among alpaka users.

i tried it on arch linux and manjaro. they possibly provide incompatible cuda and driver versions. both are up to date on my system. **downgrading cuda to 12.2 worked for me**.

still I answer the questions:

> @ichinii, please post the cmake command used to configure and compile.
> 
> Do you set `-DCMAKE_CUDA_ARCHITECTURES=52` where 52 is the sm level of your target GPU.
```
cmake \
    -DCMAKE_CXX_COMPILER=g++-13 \
    -DCMAKE_CUDA_ARCHITECTURES=52 \
    -Dalpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE:BOOL=ON \
[...]

Issue #2259, opened 2024-04-23
compile error at global mem alloc when RDC is false


mehmetyusufoglu:
I faced a compile error at Develop branch:

```
 alpaka/test/unit/mem/view/src/ViewStaticAccMem.cpp(19): error: An inline __device__/__constant__/__managed__ variable must have internal linkage when the program is compiled in whole program mode (-rdc=false)
                              alpaka::DevGlobal<TAcc, Elem[3][2]> g_globalMemory2DUninitialized;
```

- Error is observed when alpaka_RELOCATABLE_DEVICE_CODE cmake variable is DEFAULT or OFF. 
- Error is not observed when alpaka_RELOCATABLE_DEVICE_CODE cmake variable is ON.

SimeonEhrig:
I checked the CI jobs again. There is a hand written job, which test RDC:
* https://github.com/alpaka-group/alpaka/blob/develop/script/gitlabci/job_cuda.yml
* current dev: https://gitlab.com/hzdr/crp/alpaka/-/jobs/6648385841

It tests nvcc 12.0 with gcc 12. Maybe your problem is a version specific bug. Which nvcc and host compiler version do you use?

mehmetyusufoglu:
> I checked the CI jobs again. There is a hand written job, which test RDC:
> 
>     * https://github.com/alpaka-group/alpaka/blob/develop/script/gitlabci/job_cuda.yml
> 
>     * current dev: https://gitlab.com/hzdr/crp/alpaka/-/jobs/6648385841
> 
> 
> It tests nvcc 12.0 with gcc 12. Maybe your problem is a version specific bug. Which nvcc and host compiler version do you use?

Error disapperars when the flag is ON. The CI checks the ON version not when the flag is OFF ?  My config is nvcc 12.4 , gcc 12.3

psychocoderHPC:
Maybe the problem is introduced in #2242 
Before the change, we always declared the variable `extern` too see: https://github.com/alpaka-group/alpaka/pull/2242/files#diff-2eddedd7c8abe84214af714d0e61060a85e1ef924b16bad0e7bb6fff54b250b8L25

@AuroraPerego Do you know why you removed the additional `extern` declaration of the variable?

AuroraPerego:
> Do you know why you removed the additional `extern` declaration of the variable?

With the new implementation the `extern` declaration was not working, I used the `inline` attribute instead

psychocoderHPC:
[...]

PR #2303, opened 2024-06-25
Fix the issue with device global variables with CUDA 12.4

AuroraPerego:
As reported in #2259, with CUDA 12.4 there is a compile time error due to device global variables, because `inline __device__` is not allowed by CUDA.

This removes the `inline` keyword and adds CUDA 12.4 to the CI to test it.

(I've added also CUDA 12.5 since it was mentioned in #2279)


AuroraPerego:
The error in the CI is in the build of CUDA 12.4 with GCC 13.1 (supported according to [the documentation](https://docs.nvidia.com/cuda/archive/12.4.0/cuda-installation-guide-linux/index.html#host-compiler-support-policy)):
```bash
$ cudafe++ --c++17 --gnu_version=130100 --display_error_number --orig_src_file_name "CMakeCUDACompilerId.cu" \
  --orig_src_path_name "/builds/hzdr/crp/alpaka/build/CMakeFiles/3.26.4/CompilerIdCUDA/CMakeCUDACompilerId.cu" \
  --allow_managed --m64 --parse_templates --gen_c_file_name "tmp/CMakeCUDACompilerId.cudafe1.cpp" \
  --stub_file_name "CMakeCUDACompilerId.cudafe1.stub.c" --gen_module_id_file \
  --module_id_file_name "tmp/CMakeCUDACompilerId.module_id" "tmp/CMakeCUDACompilerId.cpp4.ii"
/usr/include/x86_64-linux-gnu/bits/floatn-common.h(214): error: invalid combination of type specifiers
  typedef float _Float32;
                ^
... 
```
Looking at the discussion [here](https://forums.developer.nvidia.com/t/nvcc-12-3-with-gcc-13-2-not-working/271476/6), it may be related to how gcc was installed because in the version that works:
> `floatn-common.h` is modified and copied to `include-fixed` in the GCC installation folder. That adds a preprocessor condition on top of the typedef that removes it for GCC 13.
However, we were compiling GCC with `--disable-fixincludes` because we got bitten by that causing trouble on system updates (even with minor releases)
 
The lack of this `include-fixed` may be the reason for the error, but I've not found anything that says explicitly if this was enabled or disabled in the job log.

AuroraPerego:
with the following configuration:
```shell
$ gcc -v
Using built-in specs.
COLLECT_GCC=gcc
COLLECT_LTO_WRAPPER=/cvmfs/sft.cern.ch/lcg/releases/gcc/13.1.0-b3d18/x86_64-centos7/bin/../libexec/gcc/x86_64-pc-linux-gnu/13.1.0/lto-wrapper
Target: x86_64-pc-linux-gnu
Configured with: /workspace/build/contrib/gcc-13.1.0/src/gcc/13.1.0/configure --prefix=/workspace/lcgcmake-install/gcc/13.1.0/x86_64-centos7 --disable-nls --with-system-zlib --disable-multilib --enable-languages=c,c++,fortran,lto,go --with-gnu-ld --with-gnu-as --enable-__cxa_atexit --disable-libunwind-exceptions --enable-gnu-unique-object --enable-plugin --with-linker-hash-style=gnu --enable-linker-build-id --enable-libstdcxx-time=rt --enable-shared --disable-libgcj
Thread model: posix
Supported LTO compression algorithms: zlib
gcc version 13.1.0 (GCC) 

$ /usr/local/cuda/bin/nvcc --version
Copyright (c) 2005-2024 NVIDIA Corporation
Built on Thu_Mar_28_02:18:24_PDT_2024
Cuda compilation tools, release 12.4, V12.4.131
Build cuda_12.4.r12.4/compiler.34097967_0
```
all the tests compiled and passed.

SimeonEhrig:
I cannot reproduce it on my local system (HAL), either with C++ 17 or 20. I also update the CI (#2306, #2308) at the moment. Let's wait for the updates. Maybe out-dated software causes the problems.

include/alpaka/core/Common.hpp
```diff
@@ -137,7 +137,7 @@
     || BOOST_LANG_HIP)
 #    define ALPAKA_STATIC_ACC_MEM_GLOBAL                                                                              \
         template<typename TAcc>                                                                                       \
-        inline __device__
+        __device__
```
fwyzard: This breaks the semantic of global variables when relocatable device compilation is used, because it makes them effectively `static`, so separate files refer to separate instances of the global variable.

Now, the macro does say `static` in the name.

Should we make it `static` also in intent, and implement a separate one for global variables with external linkage ?  Or just fix this one to behave as we expect, despite then name ?

fwyzard: After some tests with and without RDC (relocatable device code) with both `nvcc` and `clang`, the most robust approach seems to be
  - for code compiled with the default flags, use `static`;
  - for code compiled with RDC (`nvcc -rdc`, `clang++ -fgpu-rdc`), use `inline`.

One can check for both with
```c++
#if defined __CUDACC_RDC__ || defined __CLANG_RDC__
#define ALPAKA_STATIC_ACC_MEM_GLOBAL template<typename TAcc> __device__ inline
#else
#define ALPAKA_STATIC_ACC_MEM_GLOBAL template<typename TAcc> __device__ static
#endif
```

This also works for `hipcc`, which uses the `-fgpu-rdc` flag and the `__CLANG_RDC__` macro.

@AuroraPerego could you rebase this PR and try to use the macro above ?

@SimeonEhrig do we have CI tests with and without RDC ?

AuroraPerego: Done, let's see how the tests go (I'm afraid the CI doesn't test the `-rdc` flag, but I've run the tests for CUDA 12.4 with gcc 13.1 using `alpaka_RELOCATABLE_DEVICE_CODE=ON` and they passed).
Did you manage to compile CUDA code with `clang` or is this expected to fail if tested?

SimeonEhrig: There is a custom job, who tests it: https://github.com/alpaka-group/alpaka/blob/690da679bd209212541a95bb4f8e8be7c486bb5a/script/gitlabci/job_cuda.yml#L4

It uses CUDA 12.0 and gcc 12. You can update it and/or add more tests to test nvcc + gcc combinations, where you know, that they behaves different. I think CUDA 12.0 + gcc 12 is the "default" case and CUDA 12.4  + gcc 13 a "special" case, If don't understand the problem wrong.

But in general, we should also add the `alpaka_RELOCATABLE_DEVICE_CODE` to the generator.

fwyzard: > Did you manage to compile CUDA code with `clang` or is this expected to fail if tested?

I got it to work, it seems `clang` does not support some multi-file cases that work with `nvcc`.
Probably not an issue with alpaka, because all code usually ends up in header files.

fwyzard:
We have an error with CUDA 12.4 and GCC 13.1, that seems unrelated to this PR:
```
  /usr/include/x86_64-linux-gnu/bits/floatn-common.h(214): error: invalid
  combination of type specifiers
    typedef float _Float32;
                  ^
  
  /usr/include/x86_64-linux-gnu/bits/floatn-common.h(251): error: invalid
  combination of type specifiers
    typedef double _Float64;
                   ^
  
  /usr/include/x86_64-linux-gnu/bits/floatn-common.h(268): error: invalid
  combination of type specifiers
    typedef double _Float32x;
                   ^
  
  /usr/include/x86_64-linux-gnu/bits/floatn-common.h(285): error: invalid
  combination of type specifiers
    typedef long double _Float64x;
                        ^
  
  4 errors detected in the compilation of "CMakeCUDACompilerId.cu".
```

AuroraPerego:
> We have an error with CUDA 12.4 and GCC 13.1, that seems unrelated to this PR:
> 
> ```
>   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(214): error: invalid
>   combination of type specifiers
>     typedef float _Float32;
>                   ^
>   
>   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(251): error: invalid
>   combination of type specifiers
>     typedef double _Float64;
>                    ^
>   
>   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(268): error: invalid
>   combination of type specifiers
>     typedef double _Float32x;
>                    ^
>   
>   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(285): error: invalid
>   combination of type specifiers
>     typedef long double _Float64x;
>                         ^
>   
>   4 errors detected in the compilation of "CMakeCUDACompilerId.cu".
> ```

yes, it's the same I saw the first time we ran the tests. Could be related to the way `gcc` is installed, I cannot reproduce it locally.

SimeonEhrig:
> > We have an error with CUDA 12.4 and GCC 13.1, that seems unrelated to this PR:
> > ```
> >   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(214): error: invalid
> >   combination of type specifiers
> >     typedef float _Float32;
> >                   ^
> >   
> >   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(251): error: invalid
> >   combination of type specifiers
> >     typedef double _Float64;
> >                    ^
> >   
> >   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(268): error: invalid
> >   combination of type specifiers
> >     typedef double _Float32x;
> >                    ^
> >   
> >   /usr/include/x86_64-linux-gnu/bits/floatn-common.h(285): error: invalid
> >   combination of type specifiers
> >     typedef long double _Float64x;
> >                         ^
> >   
> >   4 errors detected in the compilation of "CMakeCUDACompilerId.cu".
> > ```
> 
> yes, it's the same I saw the first time we ran the tests. Could be related to the way `gcc` is installed, I cannot reproduce it locally.

The path looks like it use the standard library of the host compiler which is GCC 9 in a Ubuntu 20.04 container. Can you please check with this application which stl your local environment choose: https://github.com/SimeonEhrig/MyCpp/tree/master/utils/dump_used_std_library

I will check what is select in the container after lunch.

AuroraPerego:
> The path looks like it use the standard library of the host compiler which is GCC 9 in a Ubuntu 20.04 container. Can you please check with this application which stl your local environment choose: https://github.com/SimeonEhrig/MyCpp/tree/master/utils/dump_used_std_library

```shell
$ ./build/nvcc-gcc/showLibCppImplementation 
host compiler: GCC
host compiler version: 13.1
device compiler: NVCC
device compiler version: 12.4.131

use libstdc++ (GNU GCC's standard library implementation)
version: 13
```

SimeonEhrig:
> > The path looks like it use the standard library of the host compiler which is GCC 9 in a Ubuntu 20.04 container. Can you please check with this application which stl your local environment choose: https://github.com/SimeonEhrig/MyCpp/tree/master/utils/dump_used_std_library
> 
> ```shell
> $ ./build/nvcc-gcc/showLibCppImplementation 
> host compiler: GCC
> host compiler version: 13.1
> device compiler: NVCC
> device compiler version: 12.4.131
> 
> use libstdc++ (GNU GCC's standard library implementation)
> version: 13
> ```

I could reproduce your problem during I tried to compile my util. The problem is caused by this line of code: https://github.com/alpaka-group/alpaka/blob/690da679bd209212541a95bb4f8e8be7c486bb5a/script/install_gcc.sh#L25

It changes the OS host compiler and therefore the compiler detection of CMake fail. I know a solution. We can set the compiler via cmake variables `CMAKE_C_COMPILER` and `CMAKE_CXX_COMPILER` which is actual the right way. You should never change the os host compiler. This causes several problems, e.g. kernel module compiling is not working anymore (I have some experience with it :-( ).

SimeonEhrig:
@AuroraPerego Looks like your bug is Ubuntu specific: https://forums.developer.nvidia.com/t/nvcc-12-3-with-gcc-13-2-not-working/271476/9

If I try to compile a really simple CUDA code on Ubuntu 20.04 with the CUDA 12.4 and g++ 13, I get the same error:

```bash
nvcc -ccbin /usr/bin/g++-13 main.cu -o main
/usr/include/x86_64-linux-gnu/bits/floatn-common.h(214): error: invalid combination of type specifiers
  typedef float _Float32;
                ^

/usr/include/x86_64-linux-gnu/bits/floatn-common.h(251): error: invalid combination of type specifiers
  typedef double _Float64;
                 ^

/usr/include/x86_64-linux-gnu/bits/floatn-common.h(268): error: invalid combination of type specifiers
  typedef double _Float32x;
                 ^

/usr/include/x86_64-linux-gnu/bits/floatn-common.h(285): error: invalid combination of type specifiers
  typedef long double _Float64x;
                      ^

4 errors detected in the compilation of "main.cu".
```
Therefore the problem is not caused by cmake. Compiling the same code with nvcc 12.5 is working. Therefore Nvidia fixed the bug. I can also compile the code with nvcc 12.4 and g++ 13, if I installed both with spack. If I understand the form post correctly, the problem is that Ubuntu does not use a common root folder where `include`, `bin` and so one are located in the same folder.

```
gcc-13
-> include
-> bin
-> ...
```

And nvcc 12.4 has problems to handle this and select the header files from a wrong g++ version. In case of spack, g++ uses a root folder and therefore it is working.

include/alpaka/core/Common.hpp
```diff
@@ -186,7 +192,7 @@
     || BOOST_LANG_HIP)
 #    define ALPAKA_STATIC_ACC_MEM_CONSTANT                                                                            \
         template<typename TAcc>                                                                                       \
-        inline __constant__
+        __constant__
```
fwyzard: Can you do the same as `__device__` also for `__constant__` ?

fwyzard:
> If I try to compile a really simple CUDA code on Ubuntu 20.04 with the CUDA 12.4 and g++ 13, I get the same error:

Is this a problem only with Ubuntu 20.04 ?

CUDA 12.4 and g++ 13 work for me on Ubuntu 22.04 (KDE Neon).

include/alpaka/core/Common.hpp
AuroraPerego: done!

AuroraPerego:
Let's see if this attempt to remove the GCC 13 - CUDA 12.4 - Ubuntu 20.04 combination works, otherwise I'll postpone the CI update to a follow-up PR and keep only the bug fix in this one.

fwyzard:
There is something wrong with `mdspan` and `icpx`:
```
In file included from /builds/hzdr/crp/alpaka/example/convolution2D/src/convolution2D.cpp:5:
In file included from /builds/hzdr/crp/alpaka/include/alpaka/alpaka.hpp:13:
In file included from /builds/hzdr/crp/alpaka/include/alpaka/acc/AccCpuOmp2Blocks.hpp:36:
In file included from /builds/hzdr/crp/alpaka/include/alpaka/dev/DevCpu.hpp:11:
In file included from /builds/hzdr/crp/alpaka/include/alpaka/mem/buf/Traits.hpp:10:
In file included from /builds/hzdr/crp/alpaka/include/alpaka/mem/view/Traits.hpp:27:
In file included from /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/mdspan:48:
In file included from /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/__p0009_bits/mdspan.hpp:20:
In file included from /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/__p0009_bits/layout_right.hpp:22:
/builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/__p0009_bits/layout_stride.hpp:35:9: fatal error: 'concepts' file not found
   35 | #include<concepts>
      |         ^~~~~~~~~~
```

However, I don't see how it can be related to this PR.

psychocoderHPC:
> There is something wrong with `mdspan` and `icpx`:
> 
> ```
> In file included from /builds/hzdr/crp/alpaka/example/convolution2D/src/convolution2D.cpp:5:
> In file included from /builds/hzdr/crp/alpaka/include/alpaka/alpaka.hpp:13:
> In file included from /builds/hzdr/crp/alpaka/include/alpaka/acc/AccCpuOmp2Blocks.hpp:36:
> In file included from /builds/hzdr/crp/alpaka/include/alpaka/dev/DevCpu.hpp:11:
> In file included from /builds/hzdr/crp/alpaka/include/alpaka/mem/buf/Traits.hpp:10:
> In file included from /builds/hzdr/crp/alpaka/include/alpaka/mem/view/Traits.hpp:27:
> In file included from /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/mdspan:48:
> In file included from /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/__p0009_bits/mdspan.hpp:20:
> In file included from /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/__p0009_bits/layout_right.hpp:22:
> /builds/hzdr/crp/alpaka/build/_deps/mdspan-src/include/experimental/__p0009_bits/layout_stride.hpp:35:9: fatal error: 'concepts' file not found
>    35 | #include<concepts>
>       |         ^~~~~~~~~~
> ```
> 
> However, I don't see how it can be related to this PR.

Let me check if we pinned the mdspan version. Maybe our tests using the latest dev version, what would be bad for CI testing.

psychocoderHPC:
@AuroraPerego Could you rebase this PR that we are sure you have all changes we did in the last weeks.

fwyzard (approved):

PR #2307, opened 2024-07-01
Add support for cooperative groups and grid synchronization

MichaelVarvarin:
Add 4 new functions ` createTaskCooperativeKernel(TWorkDiv const&, TKernelFnObj const& kernelFnObj, TArgs&&... args)` that creates a task to later launch cooperative kernel, a new type of kernel with support for synchronizing the whole grid (look into CUDA documentation on cooperative kernels), `execCooperative(TQueue& queue, TWorkDiv const& workDiv, TKernelFnObj const& kernelFnObj, TArgs&&... args)` that immediately executes cooperative kernel, `getMaxActiveBlocks(TDev const& device, TKernelFnObj const& kernelFnObj, alpaka::Vec<TDim, TIdx> const& blockThreadExtent, alpaka::Vec<TDim, TIdx> const& threadElemExtent, TArgs const&... args) -> int` that queries the maximum number of blocks supported for cooperative launch for the specified kernel and `syncGridThreads(acc)` a barrier for all threads on the grid, similar to `syncBlockThreads(acc)` . Also adds a new device prop `//m_cooperativeLaunch` that specifies, whether device support launch of cooperative kernels. Is queried at runtime for CUDA and HIP (for CUDA any GPU launched in the last 10 years should support it), is hard-coded to false for CpuSerial and to true for all else.


include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
```diff
@@ -41,7 +41,10 @@
 namespace alpaka
 {
     template<typename TApi, typename TAcc, typename TDim, typename TIdx, typename TKernelFnObj, typename... TArgs>
-    class TaskKernelGpuUniformCudaHipRt;
+    class TaskKernelNonCooperativeGpuUniformCudaHipRt;
```
fwyzard: I would keep the old name here

include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
```diff
@@ -41,7 +41,10 @@
...
+
+    template<typename TApi, typename TAcc, typename TDim, typename TIdx, typename TKernelFnObj, typename... TArgs>
+    class TaskKernelCooperativeGpuUniformCudaHipRt;
```
fwyzard: and use `CooperativeKernel` instead of `KernelCooperative` for the new functionality.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -143,52 +143,129 @@ namespace alpaka
         std::tuple<remove_restrict_t<std::decay_t<TArgs>>...> m_args;
     };
 
+    //! The GPU CUDA/HIP accelerator execution task for non-cooperative kernel.
+    template<typename TApi, typename TAcc, typename TDim, typename TIdx, typename TKernelFnObj, typename... TArgs>
+    class TaskKernelNonCooperativeGpuUniformCudaHipRt final
+        : public TaskKernelGpuUniformCudaHipRt<TApi, TAcc, TDim, TIdx, TKernelFnObj, TArgs...>
```
fwyzard: what are the advantages of creating a new class `TaskKernelNonCooperativeGpuUniformCudaHipRt` that inherits from `TaskKernelGpuUniformCudaHipRt` ?

MichaelVarvarin: To make it symmetrical with `TaskKernelCooperativeGpuUniformCudaHipRt`, which also inherits from `TaskKernelGpuUniformCudaHipRt`

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +468,315 @@ namespace alpaka
...
+                          << std::endl;
+#        endif
+                auto kernelName = alpaka::detail::
+                    gpuKernel<TKernelFnObj, TApi, TAcc, TDim, TIdx, remove_restrict_t<std::decay_t<TArgs>>...>;
+
+#        if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL
+                {
+                    // This checks if requested number of blocks is compliant with the maxima of the accelerator.
+                    int numBlocksPerSm = 0;
+                    cudaOccupancyMaxActiveBlocksPerMultiprocessor(
```
fwyzard: eventually, can you add an entry for this function in `include/alpaka/core/ApiCudaRt.hpp` and `include/alpaka/core/ApiHipRt.hpp` ?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +468,315 @@ namespace alpaka
...
+                    }
+                }
+
+#        endif
+
+#        if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
+                {
+                    // Log maximum number of blocks that can be requested by this kernel.
+                    int numBlocksPerSm = 0;
+                    cudaOccupancyMaxActiveBlocksPerMultiprocessor(
```
fwyzard: could you store the result and reuse it, instead of calling this twice ?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -390,6 +468,315 @@ namespace alpaka
...
+                // \NOTE: No const reference (const &) is allowed as the parameter type because the kernel launch
+                // language extension expects the arguments by value. This forces the type of a float argument given
+                // with std::forward to this function to be of type float instead of e.g. "float const & __ptr64"
+                // (MSVC). If not given by value, the kernel launch code does not copy the value but the pointer to the
+                // value location.
+                std::apply(
+                    [&](remove_restrict_t<std::decay_t<TArgs>> const&... args)
+                    {
+                        void const* kernelArgs[] = {&threadElemExtent, &task.m_kernelFnObj, &args...};
+                        cudaLaunchCooperativeKernel(
```
fwyzard: eventually, can you add an entry for this function in `include/alpaka/core/ApiCudaRt.hpp` and `include/alpaka/core/ApiHipRt.hpp` ?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
MichaelVarvarin: Of course, I thought it would be a little bit more clear, considering these are 2 different debug levels

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: But then it leads to a lot of code duplication, for no additional purpose :-(

Thinking about it, is inheritance the best approach here ?
Would you consider adding a template parameter to the task, _e.g._ `bool TCooperative`, and use it with `if constexpr` to distinguish how to submit the kernel ?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: I see.

Given that one is always a subset of the other, you could nest them:
```c++
#        if ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL 
    // here goes the initialisation of the variables common between full and minimal
    // and the code for minimal debug
    ...
#            if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
    // here goes the code for full debug
    ...
#            endif
#        endif
```

fwyzard:
Looks good so far :-)

One functionality that we will need is the possibility of querying the maximum number of blocks that can be used with a given kernel, so the user can store it and use it for launching the kernel.

MichaelVarvarin:
For some reason gridSync locks up, when compiled CUDA Clang, if numberOfBlocks > 2 * multiProcessorCount

example/helloWorldGridSync/src/helloWorldGridSync.cpp
```diff
@@ -0,0 +1,92 @@
...
+ */
+
+#include <alpaka/alpaka.hpp>
+
+#include <cstdint>
+#include <iostream>
+
+//! Hello world kernel, utilizing grid synchronization.
+//! Prints hello world from a thread, performs grid sync.
+//! and prints the sum of indixes of this thread and the opposite thread (the sums have to be the same).
```
mehmetyusufoglu, line 13: [nit] Could you explain what is the opposite thread here.

mehmetyusufoglu:
> Add support for cooperative groups and related functionality

Could you add more details to the PR definition

example/helloWorldGridSync/src/helloWorldGridSync.cpp
MichaelVarvarin: Thread, that has the same distance from the end of the grid dimension, as this from the start. So, if the IDs range from 0 to 9, these are 0 and 9, 1 and 8, 2 and 7 and so on.  Their sum is constant, so we can check, if grid sync was performed successfully

MichaelVarvarin:
> > Add support for cooperative groups and related functionality
> 
> Could you add more details to the PR definition

That depends on the desired scope of the PR, I've deliberately made it vague, so we can decide, when to merge it.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: As a way of example, I've done something similar in #2311 .

example/helloWorldGridSync/src/helloWorldGridSync.cpp
MichaelVarvarin: I deliberately made it a little vague, so we can decide later, at what amount of changes actually do a merge

MichaelVarvarin:
> > Add support for cooperative groups and related functionality
> 
> Could you add more details to the PR definition



I deliberately made it a little vague, so we can decide later, at what amount of changes actually do a merge

MichaelVarvarin:
> For some reason gridSync locks up, when compiled CUDA Clang, if numberOfBlocks > 2 * multiProcessorCount

This looks like an upstream issue, at least locally, when compiling with clang 17.0.6 and CUDA 12.1.1 and 12.5

fwyzard:
> > For some reason gridSync locks up, when compiled CUDA Clang, if numberOfBlocks > 2 * multiProcessorCount
> 
> This looks like an upstream issue, at least locally, when compiling with clang 17.0.6 and CUDA 12.1.1 and 12.5

Is it supposed to work?
What is the maximum number of concurrent blocks that can be used cooperatively with this kernel?

MichaelVarvarin:
> > > For some reason gridSync locks up, when compiled CUDA Clang, if numberOfBlocks > 2 * multiProcessorCount
> > 
> > 
> > This looks like an upstream issue, at least locally, when compiling with clang 17.0.6 and CUDA 12.1.1 and 12.5
> 
> Is it supposed to work? What is the maximum number of concurrent blocks that can be used cooperatively with this kernel?

Yes, it is, maximum number reported is 16 *  multiProcessorCount, and the kernel refuses to launch, on both nvcc and clang if this number is exceeded.

fwyzard:
> Add cooperative kernel launch and grid sync support for HIP

Nice 👍🏻

MichaelVarvarin:
> > Add cooperative kernel launch and grid sync support for HIP
> 
> Nice 👍🏻

Doesn't actually work, unfortunately, I will investigate it. Would be funny, if I find a second compiler bug
Upd. It does, it was a hardware issue.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -151,7 +151,9 @@ namespace alpaka
                         // m_sharedMemSizeBytes
                         static_cast<size_t>(AccCpuOmp2Blocks<TDim, TIdx>::staticAllocBytes()),
                         // m_globalMemSizeBytes
-                        getMemBytes(dev)};
+                        getMemBytes(dev),
+                        //m_cooperativeLaunch
+                        std::false_type()};
```
fwyzard: shouldn't this be simply
```suggestion
                        false};
```
?

test/unit/acc/src/AccDevPropsTest.cpp
```diff
@@ -54,7 +55,8 @@ TEST_CASE("AccDevProps.designated_initializers", "[acc]")
         .m_threadElemExtentMax = {60},
         .m_threadElemCountMax = 70,
         .m_sharedMemSizeBytes = 80,
-        .m_globalMemSizeBytes = 90};
+        .m_globalMemSizeBytes = 90
+        .m_cooperativeLaunch = std::false_type()};
```
fwyzard:
```suggestion
        .m_globalMemSizeBytes = 90,
        .m_cooperativeLaunch = false};
```

include/alpaka/alpaka.hpp
```diff
@@ -33,6 +33,9 @@
 #include "alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp"
 #include "alpaka/atomic/Op.hpp"
 #include "alpaka/atomic/Traits.hpp"
+// grid
+#include "alpaka/grid/GridSyncGpuCudaHip.hpp"
+#include "alpaka/grid/Traits.hpp"
```
fwyzard: Does it work if you move these three lines after "extent" ?
Everything else should be in alphabetical order.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -143,56 +150,94 @@ namespace alpaka
...
+            typename TKernelFnObj,
+            bool TCooperative,
+            typename... TArgs>
+        struct IdxType<TaskKernelGpuUniformCudaHipRt<TApi, TAcc, TDim, TIdx, TKernelFnObj, TCooperative, TArgs...>>
         {
             using type = TIdx;
         };
 
-        //! The CUDA/HIP kernel enqueue trait specialization.
+        //! The CUDA/HIP non-blocking kernel enqueue trait specialization.
```
fwyzard: Can you revert to the previous version of the comment ?
Whether the task is blocking or non-blocking now depends on the `TBlocking` parameter.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -253,6 +254,17 @@ namespace alpaka
...
+            void const* func,
+            dim3 gridDim,
+            dim3 blockDim,
+            void** args,
+            size_t sharedMem,
+            Stream_t stream)
+        {
+            return ::cudaLaunchCooperativeKernel(func, gridDim, blockDim, args, sharedMem, stream);
+        }
+
```
fwyzard, line 278: Could you change this to be templated on the `func` argument ?
```suggestion
        template <typename TFunc>
        static inline Error_t launchCooperativeKernel(
            TFunc func,
            dim3 gridDim,
            dim3 blockDim,
            void** args,
            size_t sharedMem,
            Stream_t stream)
        {
            return ::cudaLaunchCooperativeKernel(func, gridDim, blockDim, args, sharedMem, stream);
        }

```

Same for the HIP implementation.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -274,11 +362,27 @@ namespace alpaka
...
-                            blockDim,
-                            static_cast<std::size_t>(blockSharedMemDynSizeBytes),
-                            queue.getNativeHandle()>>>(threadElemExtent, task.m_kernelFnObj, args...);
+                        // checks whether to launch cooperative or non-cooperative kernel
+                        if constexpr(TCooperative)
+                        {
+                            void const* kernelArgs[] = {&threadElemExtent, &task.m_kernelFnObj, &args...};
+
+                            TApi::launchCooperativeKernel(
+                                reinterpret_cast<void*>(kernelName),
```
fwyzard: Can you check if the cast can be removed, after implementing https://github.com/alpaka-group/alpaka/pull/2307/files#r1714986927 ?

include/alpaka/core/ApiCudaRt.hpp
MichaelVarvarin: Does it really matter, if the underlying function wants void* anyway?

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
MichaelVarvarin: No, you now have to do the same cast inside of launchCooperativeKernel

SimeonEhrig:
@MichaelVarvarin  Sorry for removing the draft state. I thought I start the GitHub Action jobs. Not sure what was going wrong.

fwyzard:
I've squashed all commits and rebased the result on the current `develop` branch.

fwyzard:
> **[linux_clang-14_cuda-11.2_debug_analysis](https://github.com/alpaka-group/alpaka/actions/runs/18303563525/job/52115966210?pr=2307#logs)**

There is a bug in the CUDA cooperative group headers previous to CUDA 11.5.0, that tries to use Volta-only features unconditionally also for older architectures.

@SimeonEhrig since alpaka 2.0+ supports only CUDA 12.0 and later, could you update this test to CUDA 12 ?

fwyzard:
Ah, I see that @psychocoderHPC already opened https://github.com/alpaka-group/alpaka/pull/2505 for that.

fwyzard:
Unfortunately on my laptop the new CUDA cooperative group test hangs ... :(

fwyzard:
> Unfortunately on my laptop the new CUDA cooperative group test hangs ... :(

Looks like when a kernel is compiled with device debugging enabled (`nvcc -G`) then the GPU can run a much smaller number of concurrent blocks: I've seen 1 per SM on an L4 and L40, and 2 per SM on my laptop.

To be safe, I've added a check that if the alpaka kernel entry point is compiled with debugging enabled, the launch uses at most one block per SM.

psychocoderHPC:
> > Unfortunately on my laptop the new CUDA cooperative group test hangs ... :(
> 
> Looks like when a kernel is compiled with device debugging enabled (`nvcc -G`) then the GPU can run a much smaller number of concurrent blocks: I've seen 1 per SM on an L4 and L40, and 2 per SM on my laptop.

If you compile with debug the kernel is using most likely stack frames or spill registers. This can reduce the maximum number of thread blocks that can run within one wave. The same can happen if you use compile time arrays in the kernel and you use a dynamic index to iterate over them. This should be taken into account by the cuda methods you call for each kernel to get the maximum number of blocks in a wave for each kernel.

fwyzard:
> This should be taken into account by the cuda methods you call for each kernel to get the maximum number of blocks in a wave for each kernel.

The problem I'm seeing is that the CUDA API call returns the same maximum number of blocks for the debug version as for the non-debug version.

fwyzard:
Here is a reproducer.

### `test.cu`
```c++
#include <iostream>
#include <string>

#include <cuda_runtime.h>
#include <cooperative_groups.h>

#include "cuda_check.h"

__global__
void kernel() {
  if (threadIdx.x == 0)
    printf("block %3d before barrier\n", blockIdx.x);

  cooperative_groups::this_grid().sync();

  if (threadIdx.x == 0)
    printf("block %3d after barrier\n", blockIdx.x);
}

int main(void) {
  int dev = 0;
  int numThreads = 64;
  int numBlocks = 0;

  // query device properties
  cudaDeviceProp deviceProp;
  CUDA_CHECK(cudaGetDeviceProperties(&deviceProp, dev));
  std::cout << "number of multiprocessors: " << deviceProp.multiProcessorCount << '\n';

  int numBlocksPerSm = 0;
  CUDA_CHECK(cudaOccupancyMaxActiveBlocksPerMultiprocessor(&numBlocksPerSm, kernel, numThreads, 0));
  std::cout << "number of concurrent blocks per multiprocessors: " << numBlocksPerSm << '\n';

  numBlocks = deviceProp.multiProcessorCount * numBlocksPerSm;
  std::cout << "number of concurrent blocks: " << numBlocks << '\n';
  std::cout << "number of concurrent threads: " << numBlocks * numThreads << '\n';

  // launch a grid that can maximally fill the GPU, on the default stream with kernel arguments
  void *kernelArgs[] = {};
  dim3 dimBlock(numThreads, 1, 1);
  dim3 dimGrid(numBlocks, 1, 1);
  CUDA_CHECK(cudaLaunchCooperativeKernel(kernel, dimGrid, dimBlock, kernelArgs));
  CUDA_CHECK(cudaDeviceSynchronize());

  return 0;
}
```

Compiled with `-O3`:
```bash
$ /usr/local/cuda-12.9/bin/nvcc -arch=sm_86 test.cu -O3 -o test.o3
$ ./test.o3
number of multiprocessors: 20
number of concurrent blocks per multiprocessors: 16
number of concurrent blocks: 320
number of concurrent threads: 20480
block  13 before barrier
block   4 before barrier
...
block 304 before barrier
block 294 before barrier
block  65 after barrier
block   8 after barrier
...
block 203 after barrier
block 256 after barrier
```
and completes successfully.

Compiled with `-G`:
```bash
$ /usr/local/cuda-12.9/bin/nvcc -arch=sm_86 test.cu -G -o test.g
$ ./test.g
number of multiprocessors: 20
number of concurrent blocks per multiprocessors: 16
number of concurrent blocks: 320
number of concurrent threads: 20480
block  12 before barrier
block   0 before barrier
...
block  27 before barrier
block  18 before barrier
block  31 before barrier
```
and gets stuck here, after launching only 40 blocks (2 per SM).

PR #2309, opened 2024-07-04
remove docker run in GitHub Action Linux jobs

SimeonEhrig:
looks like relict from the Travis CI


SimeonEhrig:
Launching an extra docker container on GitHub Actions with ubuntu image makes sense, because the GitHub Ubuntu image is not the official image from docker hub. It already contains a lot of installed software and does not provide a clean environment. Therefore unexpected behavior at configuring the alpaka build is possible.

https://github.com/actions/runner-images/blob/main/images/ubuntu/Ubuntu2204-Readme.md

PR #2311, opened 2024-07-06
Update the CUDA/HIP kernel tasks

fwyzard:
Reduce code duplication in CUDA/HIP kernel launch: use a boolean template parameter to distinguish blocking (sync) and non-blocking (async) queues, instead of writing two almost identical copies of the enqueue function.

Add missing white space in enqueue log messages.

### DIFF
--- a/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
@@ -200,10 +200,10 @@ namespace alpaka
 #        if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
                 // std::size_t printfFifoSize;
                 // TApi::deviceGetLimit(&printfFifoSize, TApi::limitPrintfFifoSize);
-                // std::cout << __func__ << "INFO: printfFifoSize: " << printfFifoSize << std::endl;
+                // std::cout << __func__ << " INFO: printfFifoSize: " << printfFifoSize << std::endl;
                 // TApi::deviceSetLimit(TApi::limitPrintfFifoSize, printfFifoSize*10);
                 // TApi::deviceGetLimit(&printfFifoSize, TApi::limitPrintfFifoSize);
-                // std::cout << __func__ << "INFO: printfFifoSize: " << printfFifoSize << std::endl;
+                // std::cout << __func__ << " INFO: printfFifoSize: " << printfFifoSize << std::endl;
 #        endif
                 auto const gridBlockExtent = getWorkDiv<Grid, Blocks>(task);
                 auto const blockThreadExtent = getWorkDiv<Block, Threads>(task);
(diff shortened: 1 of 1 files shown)


include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -272,118 +281,17 @@ namespace alpaka
                     },
                     task.m_args);
 
-                if constexpr(ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL)
+                if constexpr(TBlocking || ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL)
```
mehmetyusufoglu, line 284: [nit] may be adding a aline explaining the TBlocking check could be good.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
```diff
@@ -178,14 +177,21 @@ namespace alpaka
...
+            typename TKernelFnObj,
+            typename... TArgs>
         struct Enqueue<
-            QueueUniformCudaHipRtNonBlocking<TApi>,
+            uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, TBlocking>,
             TaskKernelGpuUniformCudaHipRt<TApi, TAcc, TDim, TIdx, TKernelFnObj, TArgs...>>
         {
             ALPAKA_FN_HOST static auto enqueue(
-                QueueUniformCudaHipRtNonBlocking<TApi>& queue,
+                uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, TBlocking>& queue,
```
mehmetyusufoglu, line 194: [nit] expaining the new parameter in a comment line or doxygen line  would be usefull imho

fwyzard: Considering that every other parameter is not explained, I am not going to add an explanation for the new one in this PR.

include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
fwyzard: The comment two lines below is pretty explanatory: "Wait for the kernel execution to finish (etc.)"

Like the old code did not comment
```c++
// if debugging is enabled, wait
if constexpr(ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL)
  {
  ...
```
I think that adding a comment like
```c++
// if the queue is blocking or debugging is enabled, wait
if constexpr(TBlocking || ALPAKA_DEBUG >= ALPAKA_DEBUG_MINIMAL)
  {
  ...
```
is a bit redundant.

fwyzard: By the way, I'm considering a further change:
  - if the queue is blocking, synchronise and check the result of the synchronisation
  - if the queue is non-blocking, do not synchronise, and check only the result of the kernel submission

both when debugging is enabled or disabled.

However I would prefer to keep this PR one that does not introduce any changes in behaviour, and suggest this change in a separate PR.

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #2312, opened 2024-07-06
[RFC] Add support for empty work division grids

fwyzard:
A common pattern in the CMS software has been to fix the block size at compile time and determine the number of blocks at runtime based on the problem size, hat is the total number of elements to be processed.

On a few occasions, the number of elements turned out to be zero.
This pointed out a discrepancy between the CUDA/HIP and the CPU serial back-ends:
  - the CUDA/HIP runtimes fail to launch a kernel with an empty grid (that is, with zero blocks), and report an error;
  - the CPU serial back-end simple does not do any work.

To address this discrepancy, this PR proposed to support zero-sized grids in all back-ends, amending the definition of what constitutes a valid work division, and adding an explicit test for the 1D and 3D cases.

For CUDA/HIP back-ends, the library now simply does not submit the kernel launch at all.

---
**Note:** this change cannot break any existing code, since until now zero-sized grids could not be launched. However, once this feature is relied upon, users may inadvertently introduce bugs in their code if they rely on some operations to be done on every kernel launch, and do not realise that an empty grid does not result in a kernel not being launched at all.

An alternative approach could be to continue forbidding empty grids, and enforce it in the CPU back-end.

---
This PR depends on (and includes) #2311.

### DIFF
--- a/include/alpaka/workdiv/WorkDivHelpers.hpp
+++ b/include/alpaka/workdiv/WorkDivHelpers.hpp
@@ -473,11 +473,14 @@ namespace alpaka
             auto const blockThreadExtentMax = subVecEnd<Dim<TWorkDiv>>(accDevProps.m_blockThreadExtentMax);
             auto const threadElemExtentMax = subVecEnd<Dim<TWorkDiv>>(accDevProps.m_threadElemExtentMax);
 
-            for(typename Dim<TWorkDiv>::value_type i(0); i < Dim<TWorkDiv>::value; ++i)
+            for(typename Dim<TWorkDiv>::value_type i = 0; i < Dim<TWorkDiv>::value; ++i)
             {
-                // No extent is allowed to be zero or greater then the allowed maximum.
-                if((gridBlockExtent[i] < 1) || (blockThreadExtent[i] < 1) || (threadElemExtent[i] < 1)
-                   || (gridBlockExtentMax[i] < gridBlockExtent[i]) || (blockThreadExtentMax[i] < blockThreadExtent[i])
+                // The grid extent is allowed to be zero, indicating that no work should be done.
+                // The threads and elements extents are not allowed to be zero.
+                // No extent is allowed to be greater then the allowed maximum.
+                if((std::is_signed_v<TIdx> && gridBlockExtent[i] < 0) || (blockThreadExtent[i] < 1)
+                   || (threadElemExtent[i] < 1) || (gridBlockExtentMax[i] < gridBlockExtent[i])
+                   || (blockThreadExtentMax[i] < blockThreadExtent[i])
                    || (threadElemExtentMax[i] < threadElemExtent[i]))
                 {
                     return false;
--- a/test/unit/workDiv/src/WorkDivEmptyGrid.cpp
+++ b/test/unit/workDiv/src/WorkDivEmptyGrid.cpp
@@ -1,0 +1,73 @@
+/* Copyright 2024 Andrea Bocci
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#include <alpaka/acc/AccCpuOmp2Blocks.hpp>
+#include <alpaka/acc/AccCpuOmp2Threads.hpp>
+#include <alpaka/acc/AccCpuSerial.hpp>
+#include <alpaka/acc/AccCpuTbbBlocks.hpp>
+#include <alpaka/acc/AccDevProps.hpp>
+#include <alpaka/acc/AccGpuUniformCudaHipRt.hpp>
+#include <alpaka/core/Assert.hpp>
+#include <alpaka/idx/Traits.hpp>
+#include <alpaka/kernel/KernelBundle.hpp>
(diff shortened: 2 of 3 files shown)


include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -473,11 +473,14 @@ namespace alpaka
             auto const blockThreadExtentMax = subVecEnd<Dim<TWorkDiv>>(accDevProps.m_blockThreadExtentMax);
             auto const threadElemExtentMax = subVecEnd<Dim<TWorkDiv>>(accDevProps.m_threadElemExtentMax);
 
-            for(typename Dim<TWorkDiv>::value_type i(0); i < Dim<TWorkDiv>::value; ++i)
+            for(typename Dim<TWorkDiv>::value_type i = 0; i < Dim<TWorkDiv>::value; ++i)
             {
-                // No extent is allowed to be zero or greater then the allowed maximum.
-                if((gridBlockExtent[i] < 1) || (blockThreadExtent[i] < 1) || (threadElemExtent[i] < 1)
-                   || (gridBlockExtentMax[i] < gridBlockExtent[i]) || (blockThreadExtentMax[i] < blockThreadExtent[i])
+                // The grid extent is allowed to be zero, indicating that no work should be done.
```
mehmetyusufoglu, line 478:
```suggestion
                // The grid-block extent is allowed to be zero (or zero-vector), indicating that no work should be done.
```

test/unit/workDiv/src/WorkDivEmptyGrid.cpp
```diff
@@ -0,0 +1,73 @@
...
+    using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+
+    const Platform platform{};
+    const Device dev = alpaka::getDevByIdx(platform, 0);
+    Queue queue = alpaka::Queue<Device, alpaka::Blocking>{dev};
+
+    WorkDiv empty_grid{{0u, 0u, 0u}, {1u, 1u, 1u}, {1u, 1u, 1u}};
+    CHECK_NOTHROW(alpaka::exec<Acc>(queue, empty_grid, TestKernelAbort{}));
+
+    CHECK_NOTHROW(alpaka::wait(queue));
```
mehmetyusufoglu, line 72: [question] since the queue is blocking it blocks the host and the program will not end before the end of kernel execution. Hence this CHECK_NOTHROW will never be tested?

include/alpaka/workdiv/WorkDivHelpers.hpp
```diff
@@ -473,11 +473,14 @@ namespace alpaka
...
+                // The threads and elements extents are not allowed to be zero.
+                // No extent is allowed to be greater then the allowed maximum.
+                if((std::is_signed_v<TIdx> && gridBlockExtent[i] < 0) || (blockThreadExtent[i] < 1)
```
mehmetyusufoglu, line 481:
```suggestion
                      if(! (blockThreadExtent[i] >= 0) || (blockThreadExtent[i] < 1)
```

test/unit/workDiv/src/WorkDivEmptyGrid.cpp
fwyzard: I'm not sure I understand the question.

Are you asking
1. if `alpaka::wait(queue)` is redundant ? or
2. if, even if it throws, `CHECK_NOTHROW` will not catch it ?

If the first: maybe, it depends if the implementation of `alpaka::exec` for a blocking queue is guaranteed to catch all errors, or not. An extra check will not harm.

If the second: `alpaka::wait()` may throw is the runtime is in error; `CHECK_NOTHROW` should catch that.

include/alpaka/workdiv/WorkDivHelpers.hpp
mehmetyusufoglu, line 481:
```suggestion
                if constexpr((std::is_signed_v<TIdx> && gridBlockExtent[i] < 0) || (blockThreadExtent[i] < 1)
```
to ensure that is checked in compile time and avoid silly compile warnings?

fwyzard: I don't think you can do that.

test/unit/workDiv/src/WorkDivEmptyGrid.cpp
mehmetyusufoglu: Last line alpaka::wait(queue)  does nothing since alpaka::exec is already blocking the host. But i see extra check will not harm.

fwyzard:
After thinking some more about it, I am more concern about

> **Note:** this change cannot break any existing code, since until now zero-sized grids could not be launched. However, once this feature is relied upon, users may inadvertently introduce bugs in their code if they rely on some operations to be done on every kernel launch, and do not realise that an empty grid does not result in a kernel not being launched at all.

than about the simplification from being able to launch an "empty" grid.

I will try to implement the alternative:

> An alternative approach could be to continue forbidding empty grids, and enforce it in the CPU back-end.

PR #2314, opened 2024-07-16
Let the SYCL queue implement `ConceptCurrentThreadWaitFor`, `ConceptGetDev` and `ConceptQueue`

AuroraPerego:
`QueueGenericSyclBase` was not implementing the concepts used by `isQueue` and `isDevice`. As a result, `isQueue<QueueGenericSyclBase<TDev, TBlocking>>` evaluated to `false`.
This PR fixes it by implementing the concepts for the queue and the device as done for `QueueUniformCudaHipRt` for CUDA/HIP.

### DIFF
--- a/include/alpaka/dev/Traits.hpp
+++ b/include/alpaka/dev/Traits.hpp
@@ -61,7 +61,7 @@ namespace alpaka
 
     //! True if TDev is a device, i.e. if it implements the ConceptDev concept.
     template<typename TDev>
-    inline constexpr bool isDevice = concepts::ImplementsConcept<ConceptDev, TDev>::value;
+    inline constexpr bool isDevice = concepts::ImplementsConcept<ConceptDev, std::decay_t<TDev>>::value;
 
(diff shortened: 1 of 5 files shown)


fwyzard (dismissed):

AuroraPerego:
Thanks to Andrea suggestion I've added `std::remove_cv_t<>`, otherwise a `const QueueType` (or `const DevType`) would have resulted in not being a queue (or a device).
I've also added two tests, one for `isQueue` and one for `isDevice`.

SimeonEhrig:
> Thanks to Andrea suggestion I've added `std::remove_cv_t<>`, otherwise a `const QueueType` (or `const DevType`) would have resulted in not being a queue (or a device).
> I've also added two tests, one for `isQueue` and one for `isDevice`.

Normally I use `std::decay` for this because it also removes references (and it is a no-brainer ;-) ).  I would prefer it, if there is no reason against it.

SimeonEhrig (approved):

PR #2308, opened 2024-07-02
set alpaka_CXX_STANDARD in the job generator

SimeonEhrig:
The CI job generator did not set the alpaka c++ standard via environment variable. Let's see what explods :sunglasses:

PR #2315, opened 2024-07-16
add bash variable guards in CI install scripts

SimeonEhrig:
Refactor bash CI scripts and add guards to guaranty, that every configuration environment variable is used in the cmake configuration.

The PR is motivated by PR #2308, where the `alpaka_CXX_STANDARD` environment variable was not set in the cmake configuration string.

Additional changes:
- Set empty default variables if CUDA, HIP or SYCL backend is not enabled to avoid unbound variables.
- Set CMake default configuration variables to avoid unbound variables.
- Set alpaka_DEBUG to 0 if not set. In an upcoming PR, the job generator should set more jobs to `alpaka_DEBUG=2`.
- Remove quotes for many CMake configuration arguments, to make CMake output more readable. If argument values contains white spaces, quotes are not avoidable.


mehmetyusufoglu:
> _ No description provided. _

Could you please add at least a line of explanation if you don't mind.

psychocoderHPC (approved):

PR #2316, opened 2024-07-17
[WIP] Add CloverLeaf benchmark

mehmetyusufoglu:
A compute bound benchmark depends on A hydrodynamics problem to solve the compressible Euler equations in 3D, using an explicit, second-order method. (Currently it is 3D but could be 2D as well)

- mdspan is used
- Still WIP, time and bandwith measurements will be added after result verification

- [ ] rebase against develop after #2349 is merged


mehmetyusufoglu:
Not going to be implemented

PR #2317, opened 2024-07-18
Matrix  Multiplication example using MdSpan

mehmetyusufoglu:
This matrix multiplication example is  using `mdspan` data structure for passing the data to/from kernel and for filling the data in the host conveniently. 

By using mdspan passing data size and pitch values to the kernel for each 2d array is not needed. Filling the host data is not difficult as well.

If these workshop examples are found to be useful for the users of Alpaka in general; this PR can be merged as well. 

- [x]  must be squashed before merge


example/matrixAddWithMdspan/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+#
+# Copyright 2023 Erik Zenker, Benjamin Worpitz, Jan Stephan
```
SimeonEhrig: Needs to be changed

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
```diff
@@ -0,0 +1,166 @@
+/* Copyright 2023 Mehmet Yusufoglu, Bernhard Manfred Gruber, René Widera, Andrea Bocci, Simeon Ehrig
```
SimeonEhrig: Needs to be changed

example/matrixMulWithMdspan/CMakeLists.txt
```diff
@@ -0,0 +1,53 @@
+#
+# Copyright 2023 Erik Zenker, Benjamin Worpitz, Jan Stephan
```
SimeonEhrig: Needs to be modiefied

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
```diff
@@ -0,0 +1,178 @@
+/* Copyright 2023 Mehmet Yusufoglu, Bernhard Manfred Gruber, René Widera, Andrea Bocci, Simeon Ehrig
```
SimeonEhrig: Needs to be modiefied

SimeonEhrig (dismissed):

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
```diff
@@ -0,0 +1,166 @@
...
+        extent2D,
+        Vec::ones(),
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
+
+    // Execute the kernel
+    alpaka::exec<Acc>(queue, workDiv, MatrixAddKernel{}, mdDevA, mdDevB, mdDevC);
+
+    // Wait for the kernel to finish
+    alpaka::wait(queue);
```
fwyzard: This `wait` is not needed.

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
```diff
@@ -0,0 +1,166 @@
+/* Copyright 2023 Mehmet Yusufoglu, Bernhard Manfred Gruber, René Widera, Andrea Bocci, Simeon Ehrig
+ * SPDX-License-Identifier: ISC
+ */
+
+#include <alpaka/alpaka.hpp>
+// Needed for running example for all backends available; one by one
+#include <alpaka/example/ExecuteForEachAccTag.hpp>
+
+#include <experimental/mdspan>
```
fwyzard: should this be wrapped in a check on the C++ standard, so if we are using c++23 we take `<mdspan>` and otherise if we are using the version that comes with alpaka we take `"experimental/mdspan"` ?

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
```diff
@@ -0,0 +1,166 @@
...
+struct MatrixAddKernel
+{
+    //! \tparam TAcc Accelerator type
+    //! \tparam MdSpan The type of the multidimensional span (mdspan)
+    //! \param acc Accelerator
+    //! \param A First input matrix
+    //! \param B Second input matrix
+    //! \param C Output matrix where the result of A + B will be stored
+    template<typename TAcc, typename MdSpan>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, MdSpan A, MdSpan B, MdSpan C) const
```
fwyzard: One thing I don't like here is that there is no check that `MdSpan` is actually an `mdspan` object, nor that it has 2 dimensions.

fwyzard: But I'm fine keeping it undefined for the time being, if we plan to restrict it with concepts once we move to c++20.

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
```diff
@@ -0,0 +1,178 @@
...
+{
+    //! \tparam TAcc Accelerator type
+    //! \tparam MdSpan The type of the multidimensional span (mdspan)
+    //! \param acc Accelerator
+    //! \param A First input matrix
+    //! \param B Second input matrix
+    //! \param C Output matrix where the result of A * B will be stored
+    //! \param K The shared dimension between A and B
+    template<typename TAcc, typename MdSpan>
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, MdSpan A, MdSpan B, MdSpan C, Idx K) const
```
fwyzard: shouldn't `K` be derived from `A` and `B`, with a check that the dimensions match ?

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
```diff
@@ -0,0 +1,178 @@
...
+        Vec::ones(),
+        false,
+        alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
+
+    // Execute the kernel
+    alpaka::exec<Acc>(queue, workDiv, MatrixMulKernel{}, mdDevA, mdDevB, mdDevC, K);
+
+    // Wait for the kernel to finish
+    alpaka::wait(queue);
+
```
fwyzard: Not needed.

fwyzard:
> If these workshop examples are found to be useful for the users of Alpaka in general; this PR can be merged as well.

I like the idea of having examples with `mdspan`, though I would prefer to see if the types can be used more explicitly.

mehmetyusufoglu:
> > If these workshop examples are found to be useful for the users of Alpaka in general; this PR can be merged as well.
> 
> I like the idea of having examples with `mdspan`, though I would prefer to see if the types can be used more explicitly.

By "Explicitly" you mean while passing to the kernel and instead of `auto` in the line  `auto mdHostA = alpaka::experimental::getMdSpan(bufHostA);` using mdspan ? I will try, but `getMdSpan` does a lot of work it seems :
```
 ALPAKA_FN_HOST static auto getMdSpan(TView& view)
                {
                    constexpr auto dim = Dim<TView>::value;
                    using Element = Elem<TView>;
                    auto extents = detail::makeExtents(view, std::make_index_sequence<dim>{});
                    auto* ptr = reinterpret_cast<std::byte*>(getPtrNative(view));
                    auto const strides = toArray(getPitchesInBytes(view));
                    layout_stride::mapping<decltype(extents)> m{extents, strides};
                    return std::experimental::mdspan<Element, decltype(extents), layout_stride, detail::ByteIndexedAccessor<Element>>{
                        ptr,
                        m};
                }
```

fwyzard:
> By "Explicitly" you mean while passing to the kernel and instead of `auto` in the line `auto mdHostA = alpaka::experimental::getMdSpan(bufHostA);` using mdspan ?

No, I only meant that instead of
```c++
template<typename TAcc, typename MdSpan>
ALPAKA_FN_ACC void operator()(TAcc const& acc, MdSpan A, MdSpan B, MdSpan C) const { ... }
```
I would prefer something like
```c++
template<typename TAcc,
    typename T,
    typename Extents,
    typename LayoutPolicy,
    typename AccessorPolicy>
ALPAKA_FN_ACC void operator()(TAcc const& acc,
    mdspan<T, Extents, LayoutPolicy, AccessorPolicy> A,
    mdspan<T, Extents, LayoutPolicy, AccessorPolicy> B,
    mdspan<T, Extents, LayoutPolicy, AccessorPolicy> C) const
{ ... }
```

However, I acknowledge that this is a lot more verbose, and maybe easier to write once we enable c++20 and concepts.

example/matrixAddWithMdspan/CMakeLists.txt
mehmetyusufoglu: ok changed to 2024 and SPDX-License-Identifier: MPL-2.0

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
mehmetyusufoglu: ok changed to 2024 and SPDX-License-Identifier: MPL-2.0

example/matrixMulWithMdspan/CMakeLists.txt
mehmetyusufoglu: ok changed to 2024 and SPDX-License-Identifier: MPL-2.0

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: ok changed to 2024 and SPDX-License-Identifier: MPL-2.0

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
mehmetyusufoglu: I have added a type check. Hopefully users will not find it complicated.

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: I will keep this and added a comment saying "if the queue is blocking this is not needed."

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: Good catch. Thanks done.

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
```diff
@@ -0,0 +1,182 @@
...
+//! isMdspan simply checks if a type is of type std::experimental::mdspan or not
+//! Primary template for is_mdspan (defaults to false)
+template<typename T>
+struct isMdspan : std::false_type
+{
+};
+
+//! Specialization for mdspan with four template arguments
+template<typename ElementType, typename Extents, typename LayoutPolicy, typename AccessorPolicy>
+struct isMdspan<std::experimental::mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy>> : std::true_type
```
fwyzard: can you rename this to `IsMdspan` ?

and add
```c++
template <typename T>
inline constexpr bool is_mdspan = IsMdspan<T>::value;
```
?

mehmetyusufoglu:
> > By "Explicitly" you mean while passing to the kernel and instead of `auto` in the line `auto mdHostA = alpaka::experimental::getMdSpan(bufHostA);` using mdspan ?
> 
> No, I only meant that instead of
> 
> ```c++
> template<typename TAcc, typename MdSpan>
> ALPAKA_FN_ACC void operator()(TAcc const& acc, MdSpan A, MdSpan B, MdSpan C) const { ... }
> ```
> 
> I would prefer something like
> 
> ```c++
> template<typename TAcc,
>     typename T,
>     typename Extents,
>     typename LayoutPolicy,
>     typename AccessorPolicy>
> ALPAKA_FN_ACC void operator()(TAcc const& acc,
>     mdspan<T, Extents, LayoutPolicy, AccessorPolicy> A,
>     mdspan<T, Extents, LayoutPolicy, AccessorPolicy> B,
>     mdspan<T, Extents, LayoutPolicy, AccessorPolicy> C) const
> { ... }
> ```
> 
> However, I acknowledge that this is a lot more verbose, and maybe easier to write once we enable c++20 and concepts.

Your suggestion is good and  much expressive, but for simplicity i pass TMdSpan then check whether it is exactly  std::experimental::mdspan.  

 ```
template<typename TAcc, typename TMdSpan>
    ALPAKA_FN_ACC void operator()(TAcc const& acc, TMdSpan A, TMdSpan B, TMdSpan C) const
    {
        // compile time check
        static_assert(isMdspan<TMdSpan>::value, "The type TMdSpan should be an std mdspan");

        auto const i = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[0];
        auto const j = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc)[1];

        if(i < A.extent(0) && j < A.extent(1))
        {
            C(i, j) = A(i, j) + B(i, j);
        }
    }
```

File: example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp (hunk 9 above)
Comment (fwyzard) [reply example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp]: No, this `wait` is never needed, also is the queue is _not blocking_.

Comment (fwyzard) [reply example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp]: The `wait` below (after the `memcpy`) is indeed needed, and it should always be there.

**Comment by fwyzard** [comment]
> for simplicity i pass TMdSpan then check whether it is exactly std::experimental::mdspan.

Good idea, that's a good compromise.

File: example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp (hunk 6 above)
Comment (SimeonEhrig) [reply example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp]: I disagree because the version shipped with alpaka is different to the standard implementation in an important point. The alpaka version uses the `operator()` [1] to access elements. The standard implementation uses the `operator[]` because since C++ 23 more than one argument is allowed. Therefore we would need to add a preporcessor if-else or wrapper function each time the memory is accessed.

[1] The experimental implementation provides also the `operator[]` if we use C++ 23. But we also support C++17 and 20.

File: example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp (hunk 7 above)
Comment (SimeonEhrig) [reply example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp]: You can also easily check the dimension with the static function `rank()`.

```c++
        static_assert(MdSpan::rank() == 2);
```

example/matrixAddWithMdspan/CMakeLists.txt
SimeonEhrig: Don't forget the author name

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
fwyzard: Mhm, I see.

On the other hand, `alpaka` (the library) does not _use_ the `mdspan` objects, it only returns them.

I think that, if somebody uses `alpaka` in a c++23 environment, they should get the official `mdspan`, not the experimental version (like we do for `boost::atomic_ref` vs `std::atomic_ref`).

Can we `#if __cplusplus` the examples to use `()` for c++17/c++20, and `[]` for c++23 ?

SimeonEhrig: I agree, but I think this can be a little bit more work.

The implementation we use creates a defines `MDSPAN_USE_BRACKET_OPERATOR`. Than it looks like this: https://github.com/SimeonEhrig/MyCpp/blob/28d20632657697533184f95c883739718f8397ad/features/23/mdspan/hello_mspan/main.cpp#L44
I expect, that the define will be not created by the standard library and we need to create it by our self.

Another point is, that our implementation annotates the functions with `__host__` and `__device__` to make it compatible to CUDA and HIP. At the moment, Nvidia does not provide a stl implementation and I'm not sure, if AMD does it: https://en.cppreference.com/w/cpp/compiler_support#cpp23
But Clang provides already an implementation. Therefore, in the worst case wit detect the stl implementation of Clang, which is maybe not annotated  and tries to compile it with the CUDA or HIP backend. It this case, we need a workaround selection mechanism, if the mdspan implementation is not annotated.

Long story short, I agree that we should be forward compatible but I think we need to solve this in an extra PR.

fwyzard: > Long story short, I agree that we should be forward compatible but I think we need to solve this in an extra PR.

OK for me.

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: I meant the last `wait` (after the `memcpy`), if the queue is blocking the host is blocked by `memcpy`  hence no need to `wait` after `memcpy`?  (and the queue is blocking-queue `using Queue = alpaka::Queue<Acc, alpaka::Blocking>;`)

fwyzard: Yes.

If a queue is non-blocking, it is necessary to do `alpaka::wait(queue)` after the `memcpy` and before accessing the data on the host.

If a queue is blocking, it is not necessary to do `alpaka::wait(queue)`.
However, calling `alpaka::wait(queue)` should be cheap, though not completely free.
So maybe we can keep it anyway ?

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
mehmetyusufoglu: done, thanks.

example/matrixAddWithMdspan/CMakeLists.txt
mehmetyusufoglu: done, thanks.

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
mehmetyusufoglu: done,  thanks.

example/matrixAddWithMdspan/src/matrixAddMdSpan.cpp
mehmetyusufoglu: ok, i will open a related issue not to forget, or @SimeonEhrig  could you open since you know the details...

SimeonEhrig: I opened it already: #2319

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: ok, I used `wait` only after `memcpy` statements and added no comments... Although the `queue` is blocking, as you said it is not expensive.

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
```diff
@@ -0,0 +1,202 @@
...
+    alpaka::memcpy(queue, bufDevB, bufHostB);
+    alpaka::wait(queue);
+
+    // Create mdspan views for device buffers using alpaka::experimental::getMdSpan
+    auto mdDevA = alpaka::experimental::getMdSpan(bufDevA);
+    auto mdDevB = alpaka::experimental::getMdSpan(bufDevB);
+    auto mdDevC = alpaka::experimental::getMdSpan(bufDevC);
+
+    //  Let alpaka calculate good block and grid sizes given our full problem extent.
+    auto const workDiv = alpaka::getValidWorkDiv<Acc>(
```
psychocoderHPC: please use getValidWorkDivKernel

psychocoderHPC (changes requested):
please rebase to use getValidWorkDivKErnel

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: rebased, done, thanks.

psychocoderHPC:
offline discussed with @mehmetyusufoglu : we will remove the add example because it is 99.9% equal to the multiplication and IMO the code duplication is not required.

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
```diff
@@ -0,0 +1,206 @@
...
+            {
+                success = false;
+                break;
+            }
+        }
+    }
+
+    std::cout << "Multiplication of matrices of size " << M << "x" << K << " and " << K << "x" << N << " using mdspan "
+              << (success ? "succeeded" : "failed") << "!" << std::endl;
+    return EXIT_SUCCESS;
```
psychocoderHPC, line 194: if `success` is false you should end with `EXIT_FAILURE`

psychocoderHPC (changes requested):

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
mehmetyusufoglu: ok, added. Thanks a lot. (Tested  by `ctest`  for both true-positive, true-negative cases. ) Other examples like convolution1d,2d,2dwithMdSpan are correct in this sense.

psychocoderHPC (approved):

PR #2318, opened 2024-07-21
Use KernelBundle in exec and task create [KB-III]

mehmetyusufoglu:
Add an overload to `exec` and change `createTaskKernel` functions so that they will take the kernel and arguments as a single argument of type `KernelBundle` .  Once `KernelBundle` instance is created to be used in `GetValidWorkDivForKernel`; using it again for `exec` (and in createTaskKernel) is needed because it already bundles the arguments with the kernel object hence exec does not need seperate kernel arguments.

This is the **3rd PR** of `KernelBundle` related tasks. Previous ones were #2251 (merged) and #2261.

- `CreateTaskKernel` has a KernelBundle argument
- `isKernelArgumentTriviallyCopyable` is moved to `KernelBundle` since it is not used anywhere else and better to check there. Because KernelBundle is used if the kernel is executed somehow; either before exec or inside the overloaded exec.
-  kernel return type check traits changed; now a `KernelBundle` argument can be used directly to check the kernel return type
-  Debug log about kernel object starting with `#if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL ...` moved into the `KernelBundle`. 

- [x] This PR will be merged after #2261.
- [x]  must be squashed before merge

### DIFF
--- a/example/convolution2D/src/convolution2D.cpp
+++ b/example/convolution2D/src/convolution2D.cpp
@@ -320,18 +320,7 @@ auto example(TAccTag const&) -> int
     auto const workDiv = alpaka::getValidWorkDivForKernel<DevAcc>(devAcc, bundeledKernel, extent, Vec::ones());
 
     // Run the kernel
-    alpaka::exec<DevAcc>(
-        queueAcc,
-        workDiv,
-        convolutionKernel2D,
-        std::data(bufInputAcc),
-        std::data(outputDeviceMemory),
-        matrixWidth,
-        matrixHeight,
-        std::data(bufFilterAcc),
-        filterWidth,
-        intputWidthAllocated,
-        filterWidthAllocated);
+    alpaka::exec<DevAcc>(queueAcc, workDiv, bundeledKernel);
 
     // Allocate memory on host to receive the resulting matrix as an array
     auto resultGpuHost = alpaka::allocBuf<DataType, Idx>(devHost, extent1D);
--- a/example/heatEquation/src/heatEquation.cpp
+++ b/example/heatEquation/src/heatEquation.cpp
@@ -146,8 +146,10 @@ auto example(TAccTag const&) -> int
 
     for(uint32_t step = 0; step < numTimeSteps; step++)
     {
+        auto const& tempBundeledKernel = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+
         // Compute next values
-        alpaka::exec<Acc>(queue, workDiv, heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+        alpaka::exec<Acc>(queue, workDiv, tempBundeledKernel);
 
         // We assume the boundary conditions are constant and so these values
         // do not need to be updated.
--- a/example/helloWorld/src/helloWorld.cpp
+++ b/example/helloWorld/src/helloWorld.cpp
@@ -149,11 +149,18 @@ auto example(TAccTag const&) -> int
     // The queue can be blocking or non-blocking
     // depending on the chosen queue type (see type definitions above).
     // Here it is synchronous which means that the kernel is directly executed.
-    alpaka::exec<Acc>(
-        queue,
-        workDiv,
-        helloWorldKernel
-        /* put kernel arguments here */);
+    alpaka::exec<Acc>(queue, workDiv, bundeledKernel);
+
+    // Alternative way for kernel execution (commented out deliberately)
+    // exec function can be called by directly using kernel instance and kernel arguments without bundling kernel in a
+    // KernelBundle. This might be usefull if getValidWorkDivForKernel is not needed before kernel execution and
+    // workdiv is determined directly instantiating the WorkDivMembers class with predetermined values.
+    // alpaka::exec<Acc>(
+    //     queue,
+    //     workDiv,
+    //     helloWorldKernel,
+    //     /* put kernel arguments here*/);
+
     alpaka::wait(queue);
 
     return EXIT_SUCCESS;
--- a/example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
+++ b/example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
@@ -114,7 +114,9 @@ void naiveCudaStyle(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nNaive CUDA style processing - each thread processes one data point:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << "alpaka element layer not used\n";
-    alpaka::exec<TAcc>(queue, workDiv, NaiveCudaStyleKernel{}, std::data(bufAcc), n);
+
+    NaiveCudaStyleKernel naiveCudaStyleKernel;
+    alpaka::exec<TAcc>(queue, workDiv, naiveCudaStyleKernel, std::data(bufAcc), n);
     testResult(queue, bufAcc);
 }
 
@@ -177,7 +179,11 @@ void gridStridedLoop(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nGrid strided loop processing - fixed number of threads and blocks:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << "alpaka element layer not used\n";
-    alpaka::exec<TAcc>(queue, workDiv, GridStridedLoopKernel{}, std::data(bufAcc), n);
+
+    GridStridedLoopKernel gridStridedLoopKernel;
+    auto bufAccPtr = std::data(bufAcc);
+    auto const& bundeledKernel = alpaka::KernelBundle(gridStridedLoopKernel, bufAccPtr, n);
+    alpaka::exec<TAcc>(queue, workDiv, bundeledKernel);
     testResult(queue, bufAcc);
 }
 
@@ -252,7 +258,11 @@ void chunkedGridStridedLoop(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nChunked grid strided loop processing - fixed number of threads and blocks:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << elementsPerThread << " alpaka elements per thread\n";
-    alpaka::exec<TAcc>(queue, workDiv, ChunkedGridStridedLoopKernel{}, std::data(bufAcc), n);
+
+    ChunkedGridStridedLoopKernel chunkedGridStridedLoopKernel;
+    auto const& bundeledKernel = alpaka::KernelBundle(chunkedGridStridedLoopKernel, std::data(bufAcc), n);
+    alpaka::exec<TAcc>(queue, workDiv, bundeledKernel);
+
     testResult(queue, bufAcc);
 }
 
@@ -318,7 +328,11 @@ void naiveOpenMPStyle(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nNaive OpenMP style processing - each thread processes a single consecutive range of elements:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << "alpaka element layer not used\n";
-    alpaka::exec<TAcc>(queue, workDiv, NaiveOpenMPStyleKernel{}, std::data(bufAcc), n);
+
+    NaiveOpenMPStyleKernel naiveOpenMPStyleKernel;
+    auto const& bundeledKernel = alpaka::KernelBundle(naiveOpenMPStyleKernel, std::data(bufAcc), n);
+
+    alpaka::exec<TAcc>(queue, workDiv, bundeledKernel);
     testResult(queue, bufAcc);
 }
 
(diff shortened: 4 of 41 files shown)


psychocoderHPC:
This PR can be rebased against the dev branch.

psychocoderHPC:
@mehmetyusufoglu please rebase this PR

mehmetyusufoglu:
> @mehmetyusufoglu please rebase this PR

done

fwyzard (changes requested):
I disagree to changing all examples to use the kernel bundle.

It is a completely unrealistic approach, because kernel arguments and work division are _expected_ to change on every invocation - making the creation of the bundle an extra complication.

mehmetyusufoglu:
> I disagree to changing all examples to use the kernel bundle.
> 
> It is a completely unrealistic approach, because kernel arguments and work division are _expected_ to change on every invocation - making the creation of the bundle an extra complication.

- We need kernel arguments in 2 places in the code 1. `getValidWorkDivForKernel` and 2.`exec`. Therefore we needed `KernelBundle` not to give kernel arguments 2 times to 2 different functions as arguments.
- 3 examples (parallel loop patterns, reduce, tag spec)  don't use getValidWorkDivForKernel; they use directly their specific work-div and give the kernel arguments in `exec` without using KernelBundle.    
- As a solution: I can make some more examples  use directly determined workdiv and call exec with normal kernel arguments and write explicitly in examples (at least the basic ones vectorAdd, matmul... which users usually check initially) saying there are 2 alternatives etc. ?
- Changing documentation according to your comment above might also help?

fwyzard:
> * We need kernel arguments in 2 places in the code 1. `getValidWorkDivForKernel` and 2.`exec`. Therefore we needed `KernelBundle` not to give kernel arguments 2 times to 2 different functions as arguments.

My point is that in production code I would not use `getValidWorkDivForKernel` each time I call a kernel:
  - in part because the call to `cudaFuncGetAttributes` is expensive, so I would use it at most once;
  - in part because I would prefer to let the developers pick an optimal work division after benchmarking the code (OK, I know, this is a tough ask...).

So the reason that we need the kernel arguments in two places does not really apply any more.

If you change all (or even most) of the example to use kernel bundles, it gives the impression that that is the "right" way to launch a kernel with alpaka - and I'm still not convinced that this is the message I would like to pass.

fwyzard:
> * As a solution: I can make some more examples  use directly determined workdiv and call exec with normal kernel arguments and write explicitly in examples (at least the basic ones vectorAdd, matmul... which users usually check initially) saying there are 2 alternatives etc. ?
> 
> * Changing documentation according to your comment above might also help?

Yes, I agree that these two would help.

include/alpaka/kernel/Traits.hpp
```diff
@@ -236,145 +273,127 @@ namespace alpaka
...
+//! \tparam TArgs Kernel function object argument types as a parameter pack.
 //! \param queue The queue to enqueue the view copy task into.
 //! \param workDiv The index domain work division.
 //! \param kernelFnObj The kernel function object which should be executed.
 //! \param args,... The kernel invocation arguments.
 #if BOOST_COMP_CLANG
 #    pragma clang diagnostic pop
 #endif
-    template<typename TAcc, typename TQueue, typename TWorkDiv, typename TKernelFnObj, typename... TArgs>
-    ALPAKA_FN_HOST auto exec(TQueue& queue, TWorkDiv const& workDiv, TKernelFnObj const& kernelFnObj, TArgs&&... args)
```
psychocoderHPC, line 375: There should be no need to change `&&` to value type

psychocoderHPC (changes requested):

include/alpaka/kernel/Traits.hpp
mehmetyusufoglu: done, thanks.

mehmetyusufoglu:
> This PR can be rebased against the dev branch.

It is up to date.

mehmetyusufoglu:
> > * As a solution: I can make some more examples  use directly determined workdiv and call exec with normal kernel arguments and write explicitly in examples (at least the basic ones vectorAdd, matmul... which users usually check initially) saying there are 2 alternatives etc. ?
> > * Changing documentation according to your comment above might also help?
> 
> Yes, I agree that these two would help.

I have added a comment to the _HElloWorld_ example and removed `KernelBundle` usage if `GetValidWorkDivForKernel` is not used. I will prepare a new PR adressing this request after that one.

mehmetyusufoglu:
@fwyzard 6 tests under integ directory is changed:

1. testTemplate.cpp: getValidWorkDivForKernel is used therefore bundle is needed
2. axpy.cpp: getValidWorkDivForKernel is used
3. mandelbrot.cpp: getValidWorkDivForKernel is used
4. matmul.cpp: getValidWorkDivForKernel is used
5. seperableCalculation/main.cpp: getValidWorkDivForKernel is used
6. sharedMem.cpp
The function `measureTaskRunTimeMs` takes a queue and a task. KernelBundle instance is created to be used in `CreateTaskKernel`. Although the instance(`bundeledKernel2`) is not used for finding workdiv. 

```
// Create the kernel execution task using the real data.
    auto const& bundeledKernel2 = alpaka::KernelBundle(kernel, std::data(blockRetValsAcc));
    auto const taskKernel = alpaka::createTaskKernel<Acc>(workDiv, bundeledKernel2);

    // Profile the kernel execution.
    std::cout << "Execution time: " << alpaka::test::integ::measureTaskRunTimeMs(queue, taskKernel) << " ms"
              << std::endl;
```

example/heatEquation/src/heatEquation.cpp
```diff
@@ -146,8 +146,10 @@ auto example(TAccTag const&) -> int
 
     for(uint32_t step = 0; step < numTimeSteps; step++)
     {
+        auto const& tempBundeledKernel = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+
         // Compute next values
-        alpaka::exec<Acc>(queue, workDiv, heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+        alpaka::exec<Acc>(queue, workDiv, tempBundeledKernel);
```
psychocoderHPC, line 152: `tempBundeledKernel` -> `tmpKernelBundle`

`temp` is only used in the windows world^^

example/convolution2D/src/convolution2D.cpp
```diff
@@ -320,18 +320,7 @@ auto example(TAccTag const&) -> int
...
-        convolutionKernel2D,
-        std::data(bufInputAcc),
-        std::data(outputDeviceMemory),
-        matrixWidth,
-        matrixHeight,
-        std::data(bufFilterAcc),
-        filterWidth,
-        intputWidthAllocated,
-        filterWidthAllocated);
+    alpaka::exec<DevAcc>(queueAcc, workDiv, bundeledKernel);
```
psychocoderHPC, line 323: please rename all `bundeledKernel` to `kernelBundle`

psychocoderHPC: `bundeled` is a typo, please check for other places in this PR too.

example/helloWorld/src/helloWorld.cpp
```diff
@@ -149,11 +149,18 @@ auto example(TAccTag const&) -> int
...
-        workDiv,
-        helloWorldKernel
-        /* put kernel arguments here */);
+    alpaka::exec<Acc>(queue, workDiv, bundeledKernel);
+
+    // Alternative way for kernel execution (commented out deliberately)
+    // exec function can be called by directly using kernel instance and kernel arguments without bundling kernel in a
+    // KernelBundle. This might be usefull if getValidWorkDivForKernel is not needed before kernel execution and
+    // workdiv is determined directly instantiating the WorkDivMembers class with predetermined values.
+    // alpaka::exec<Acc>(
```
psychocoderHPC, line 158: This code is not tested and I call this dead code because it will become outdated soon.
If we would like to keep it we should think about calling the kernel twice!

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
```diff
@@ -114,7 +114,9 @@ void naiveCudaStyle(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nNaive CUDA style processing - each thread processes one data point:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << "alpaka element layer not used\n";
-    alpaka::exec<TAcc>(queue, workDiv, NaiveCudaStyleKernel{}, std::data(bufAcc), n);
+
+    NaiveCudaStyleKernel naiveCudaStyleKernel;
+    alpaka::exec<TAcc>(queue, workDiv, naiveCudaStyleKernel, std::data(bufAcc), n);
```
psychocoderHPC, line 119: please revert the change and keep 

`alpaka::exec<TAcc>(queue, workDiv, NaiveCudaStyleKernel{}, std::data(bufAcc), n);`
No need to change this line.

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
```diff
@@ -177,7 +179,11 @@ void gridStridedLoop(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nGrid strided loop processing - fixed number of threads and blocks:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << "alpaka element layer not used\n";
-    alpaka::exec<TAcc>(queue, workDiv, GridStridedLoopKernel{}, std::data(bufAcc), n);
+
+    GridStridedLoopKernel gridStridedLoopKernel;
```
psychocoderHPC, line 183: please revert this change, there is no need to change this. The old interface should work too.

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
```diff
@@ -252,7 +258,11 @@ void chunkedGridStridedLoop(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nChunked grid strided loop processing - fixed number of threads and blocks:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << elementsPerThread << " alpaka elements per thread\n";
-    alpaka::exec<TAcc>(queue, workDiv, ChunkedGridStridedLoopKernel{}, std::data(bufAcc), n);
+
+    ChunkedGridStridedLoopKernel chunkedGridStridedLoopKernel;
```
psychocoderHPC, line 262: no need to change this code, please revert!

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
```diff
@@ -318,7 +328,11 @@ void naiveOpenMPStyle(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nNaive OpenMP style processing - each thread processes a single consecutive range of elements:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << "alpaka element layer not used\n";
-    alpaka::exec<TAcc>(queue, workDiv, NaiveOpenMPStyleKernel{}, std::data(bufAcc), n);
+
+    NaiveOpenMPStyleKernel naiveOpenMPStyleKernel;
+    auto const& bundeledKernel = alpaka::KernelBundle(naiveOpenMPStyleKernel, std::data(bufAcc), n);
```
psychocoderHPC, line 333: No need to change the code please revert

example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
```diff
@@ -396,7 +410,11 @@ void openMPSimdStyle(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     std::cout << "\nOpenMP SIMD style processing - each thread processes a single consecutive range of elements:\n";
     std::cout << "   " << blocksPerGrid << " blocks, " << threadsPerBlock << " threads per block, "
               << elementsPerThread << " alpaka elements per thread\n";
-    alpaka::exec<TAcc>(queue, workDiv, OpenMPSimdStyleKernel{}, std::data(bufAcc), n);
+
+    OpenMPSimdStyleKernel openMPSimdStyleKernel;
+    auto const& bundeledKernel = alpaka::KernelBundle(openMPSimdStyleKernel, std::data(bufAcc), n);
```
psychocoderHPC, line 415: please revert.

example/tagSpecialization/src/tagSpecialization.cpp
```diff
@@ -109,10 +109,11 @@ auto example(TAccTag const&) -> int
         static_cast<std::size_t>(1),
         static_cast<std::size_t>(1)};
 
+    WrapperKernel wrapperKernel;
     // Run the wrapper kernel, which calls the actual specialized
     // Specializing the entry kernel with tags is not possible. Therefore pre processor guards are required, see
     // kernelSpecialization example.
-    alpaka::exec<Acc>(queue, workDiv, WrapperKernel{});
+    alpaka::exec<Acc>(queue, workDiv, wrapperKernel);
```
psychocoderHPC, line 116: please change, this change is not required

include/alpaka/kernel/KernelBundle.hpp
```diff
@@ -30,9 +63,21 @@ namespace alpaka
             : m_kernelFn(std::move(kernelFn))
             , m_args(std::forward<TArgs>(args)...)
         {
+#if BOOST_COMP_NVCC
```
psychocoderHPC, line 66: The asserts here are fine but why due you sneaked these into this PR? They are ortogonal to all other changes here and we need to remove the ckecks for cuda where we did it in the past.

psychocoderHPC (changes requested):
I will take over this PR and finish it.

It is mixing to many things together and therefore is hard to review.

include/alpaka/kernel/Traits.hpp
```diff
@@ -236,145 +273,127 @@ namespace alpaka
...
             std::is_same_v<Idx<std::decay_t<TWorkDiv>>, Idx<TAcc>>,
             "The idx type of TAcc and the idx type of TWorkDiv have to be identical!");
-
-#if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
-        std::cout << __func__ << " workDiv: " << workDiv << ", kernelFnObj: " << core::demangled<decltype(kernelFnObj)>
-                  << std::endl;
-#endif
-        return trait::CreateTaskKernel<TAcc, TWorkDiv, TKernelFnObj, TArgs...>::createTaskKernel(
+        KernelBundle<TKernelFn, TArgs...> const& kernelBundle
+            = alpaka::KernelBundle(kernelFnObj, std::forward(args)...);
```
psychocoderHPC, line 343: must be `std::forward<TArgs>(args)`

psychocoderHPC:
first PR is open: #2339

- in the new PR `isKernelArgumentTriviallyCopyable` is not moved to kernel bundle because it is calling a trait the user could change and traits should stay in the trait file.

psychocoderHPC:
this PR is replaced by #2349

PR #2320, opened 2024-07-22
alpaka CI: set CMAKE_CXX_COMPILER instead using update-alternatives

SimeonEhrig:
Changing the OS host compiler has dangerous side effects and is not common on productive systems. Therefore our CI should not do it. Instead the CI should use  `CMAKE_CXX_COMPILER` to set the compiler.

- Instead changing the default host compiler, set the compiler via CMAKE_CXX_COMPILER argument at CMake configure.
- Rename environment variable CXX to ALPAKA_CI_CXX, because CXX can cause unexpected side effects.
- The environment variable CMAKE_CXX_COMPILER is created in the scripts depending on the variable ALPAKA_CI_CXX.
- Remove the environment variable CC completely because we do not compile C code.
- Set also CMAKE_CUDA_COMPILER as CMake configure argument manually.

Set `CMAKE_CUDA_COMPILER` as CMake argument was also necessary because of the changed behavior of setting the C++ compiler.

### DIFF
--- a/script/gitlabci/job_base.yml
+++ b/script/gitlabci/job_base.yml
@@ -18,13 +18,12 @@
   image: ubuntu:${ALPAKA_CI_UBUNTU_VER}
   variables:
     ALPAKA_CI_UBUNTU_VER: "20.04"
-    CC: gcc
-    CXX: g++
+    ALPAKA_CI_CXX: g++
     alpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE: "ON"
     alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE: "ON"
     alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE: "OFF"
     alpaka_ACC_GPU_CUDA_ENABLE: "ON"
-    CMAKE_CUDA_COMPILER: nvcc
+    ALPAKA_CI_CUDA_COMPILER: nvcc
     ALPAKA_CI_STDLIB: libstdc++
     # CI contains a Quadro P5000 (sm_61)
     CMAKE_CUDA_ARCHITECTURES: "61"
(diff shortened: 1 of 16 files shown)


script/docker_ci.sh
```diff
@@ -12,8 +12,8 @@ ALPAKA_CI_BOOST_BRANCH="boost-${ALPAKA_BOOST_VERSION}"
 
 # runtime and compile time options
 ALPAKA_DOCKER_ENV_LIST=()
-ALPAKA_DOCKER_ENV_LIST+=("--env" "CC=${CC}")
-ALPAKA_DOCKER_ENV_LIST+=("--env" "CXX=${CXX}")
+ALPAKA_DOCKER_ENV_LIST+=("--env" "ALPAKA_CI_CC=${ALPAKA_CI_CC}")
```
psychocoderHPC: setting the C compiler could be skipped we do not use the C compiler

script/gitlabci/job_base.yml
```diff
@@ -18,8 +18,8 @@
   image: ubuntu:${ALPAKA_CI_UBUNTU_VER}
   variables:
     ALPAKA_CI_UBUNTU_VER: "20.04"
-    CC: gcc
-    CXX: g++
+    ALPAKA_CI_CC: gcc
```
psychocoderHPC: setting the C compiler should not be required an will automaticlally fall back to the system C compiler, alpaka is not compiling C code.

psychocoderHPC (changes requested):

script/docker_ci.sh
SimeonEhrig: The boost install script is the only one, which use the variable `ALPAKA_CI_CC`. I can check, if we can use the `ALPAKA_CI_CXX` instead.

SimeonEhrig: I use the `ALPAKA_CI_CXX` variable to select the C compiler in the `install_boost.sh` script. Therefore I removed the variable `ALPAKA_CI_CC` and also `CC`.

psychocoderHPC (approved):

SimeonEhrig:
@psychocoderHPC ready to merge

PR #2324, opened 2024-07-24
fix missing C++20 STL for ICPX in the CI

### LINKED ISSUE #2321: CI error: ICPX does not compile with C++ 20 and mdspan
The CI does not select a STL which supports C++ 20 if the Intel ICPX is used. Therefore mdspan cannot be compiled because it uses C++ 20 concepts. The job is only sometimes generated because of bugs and randomness in the job generator script.

Without mdspan, alpaka compiles because it does not use any C++ 20 STL feature.

SimeonEhrig:
fix #2321

### DIFF
--- a/script/install_oneapi.sh
+++ b/script/install_oneapi.sh
@@ -41,6 +41,15 @@ then
     set +eu
     source /opt/intel/oneapi/setvars.sh
     set -eu
+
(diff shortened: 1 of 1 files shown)


psychocoderHPC (approved):

psychocoderHPC:
I will merge this fix with my admin power to fix the current broken dev branch.

PR #2327, opened 2024-07-28
Suppress GCC warning about casting a function to void*

fwyzard:
Suppress the `-Wconditionally-supported` GCC warning about

> casting between pointer-to-function and pointer-to-object is conditionally-supported

This is required to be supported by POSIX, and used extensively by dlsym(3).


fwyzard:
> ```
>   Idx i: 994 computed : (1812.72,-0.4457) vs expected: (1812.72,-0.443972)
> ```

Is this a real error that needs to be looked into (the [full log](https://github.com/user-attachments/files/16425202/linux_hipcc5.5_cmake3.26.4_boost1.78.0_ubuntu20.04_cxx20_mdspan_Debug.log) is attached) ?

psychocoderHPC (approved):

Issue #2328, opened 2024-07-28
`hostOnlyAPITest` fails


fwyzard:
I've building the alpaka tests with gcc 13 (tested also 12) and CUDA 12.5 (tested also 12.4) on Ubuntu 22.04, with
```bash
cmake /home/fwyzard/src/alpaka-group/alpaka \
  -G 'Unix Makefiles' \
  --log-level=VERBOSE \
  -Dalpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE='OFF' \
  -Dalpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE='OFF' \
  -Dalpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE='ON' \
  -Dalpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE='OFF' \
  -Dalpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE='OFF' \
  -Dalpaka_ACC_CPU_DISABLE_ATOMIC_REF=OFF \
  -Dalpaka_ACC_GPU_CUDA_ENABLE='ON' \
  -Dalpaka_ACC_GPU_CUDA_ONLY_MODE='OFF' \
  -Dalpaka_ACC_GPU_HIP_ENABLE='OFF' \
  -Dalpaka_ACC_GPU_HIP_ONLY_MODE='OFF' \
  -Dalpaka_ACC_SYCL_ENABLE=OFF \
  -Dalpaka_ASSERT_ACC_ENABLE=ON \
  -Dalpaka_BLOCK_SHARED_DYN_MEMBER_ALLOC_KIB=47 \
  -Dalpaka_BUILD_BENCHMARKS=ON \
  -Dalpaka_BUILD_EXAMPLES=ON \
  -Dalpaka_CXX_STANDARD=17 \
  -Dalpaka_DEBUG=2 \
  -Dalpaka_DISABLE_VENDOR_RNG=ON \
  -Dalpaka_ENABLE_WERROR=OFF \
  -Dalpaka_FAST_MATH=DEFAULT \
  -Dalpaka_FTZ=DEFAULT \
  -Dalpaka_INSTALL_TEST_HEADER=OFF \
  -Dalpaka_RELOCATABLE_DEVICE_CODE=DEFAULT \
  -Dalpaka_USE_MDSPAN=OFF \
  -DBUILD_TESTING=ON \
[...]

PR #2329, opened 2024-07-29
Link `libcudart` even when `libcurand` is not used

fwyzard:
Link the CUDA runtime library (`libcudart.so`) when compiling host code that uses CUDA, also when the cuRAND library (`libcurand.so`) is not used.


fwyzard:
Fixes #2328

psychocoderHPC (approved):

PR #2331, opened 2024-07-30
Rewrite the `getValidWorkDivForKernel` 1D and 2D tests

fwyzard:
Fix the `getValidWorkDivForKernel` tests for the CUDA backend and merge it with the other parallel backends.

Fix the `getValidWorkDivForKernel` tests for the SYCL CPU backend and move it to the parallel backends.

Rewrite the `getValidWorkDivForKernel` tests: test more valid and invalid work divisions, merge the serial and parallel backends, and use traits to run tests specific to the serial backends.

### DIFF
--- a/test/unit/workDiv/src/WorkDivForKernelTest.cpp
+++ b/test/unit/workDiv/src/WorkDivForKernelTest.cpp
@@ -69,7 +69,7 @@ struct TestKernelWithManyRegisters
         double sum = var0 + var1 + var2 + var3 + var4 + var5 + var6 + var7 + var8 + var9 + var10 + var11 + var12
                      + var13 + var14 + var15 + var16 + var17 + var18 + var19 + var20 + var21 + var22 + var23 + var24
                      + var25 + var26 + var27 + var28 + var29 + var30 + var31 + var32 + var33 + var34 + var35;
-        printf("The sum is %5.2f, the argument is %lu ", sum, val);
+        printf("The sum is %5.2f, the argument is %lu\n", sum, val);
     }
 };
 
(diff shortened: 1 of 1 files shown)


fwyzard:
@mehmetyusufoglu can you double check the tests ?

mehmetyusufoglu:
> @mehmetyusufoglu can you double check the tests ?

Ok, doing. Thanks.

mehmetyusufoglu:
> @mehmetyusufoglu can you double check the tests ?

Locally everything worked, i am trying to understand which accelerator is generating the error. I will 

> @mehmetyusufoglu can you double check the tests ?

Locally everything worked for me as below, I pushed a copy branch to test on CI with logs, i will test on HAL as well. 
```
/projects/alpaka-dir/alpaka/build/test/unit/workDiv$ ./workDivTest 
Randomness seeded to: 3752370894
- Acc: AccCpuSerial<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 536870911), blockThreadExtent: (2, 1), threadElemExtent: (1, 1)}
- Acc: AccCpuThreads<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 4329604), blockThreadExtent: (2, 124), threadElemExtent: (1, 1)}
- Acc: AccCpuTbbBlocks<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 536870911), blockThreadExtent: (2, 1), threadElemExtent: (1, 1)}
- Acc: AccCpuOmp2Blocks<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 536870911), blockThreadExtent: (2, 1), threadElemExtent: (1, 1)}
- Acc: AccCpuOmp2Threads<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 35791394), blockThreadExtent: (2, 15), threadElemExtent: (1, 1)}
- Acc: AccGpuCudaRt<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 729444), blockThreadExtent: (2, 736), threadElemExtent: (1, 1)}
All tests passed (425 assertions in 85 test cases)
```

test/unit/workDiv/src/WorkDivForKernelTest.cpp
```diff
@@ -164,97 +161,68 @@ TEMPLATE_LIST_TEST_CASE("getValidWorkDivForKernel.2D", "[workDivKernel]", TestAc
...
+
+    // Check that using the maximum number of threads per block is valid.
+    auto const validWorkDiv
+        = WorkDiv{Vec{8, threadsPerGridTestValue / threadsPerBlock / 8}, Vec{1, threadsPerBlock}, Vec{1, 1}};
+    CHECK(alpaka::isValidWorkDivKernel<Acc>(dev, kernelBundle, validWorkDiv));
+
+    // Check that using too many threads per block is not valid.
+    auto const invalidWorkDiv
+        = WorkDiv{Vec{8, threadsPerGridTestValue / threadsPerBlock / 8}, Vec{2, threadsPerBlock}, Vec{1, 1}};
+    CHECK(not alpaka::isValidWorkDivKernel<Acc>(dev, kernelBundle, invalidWorkDiv));
```
mehmetyusufoglu: My initial observation is that; the check assumes  `2 * threadsPerBlock` will be more than allowed block size. But `threadsPerBlock` is not the max block size, it is a value found by getValidWorkDiv, why  `2 * threadsPerBlock`  should be more than allowed block-size limit for all accelerators?. Anyway I am trying to understand which accelerator is that. Thanks.

mehmetyusufoglu: Seems to be happening at AccCpuOmp2Threads. Below is valid because AccCpuOmp2Threads allows 64 threads. Test expect an invalid workdiv but it is valid. Which is in my opinion normal result. We should not test for AccCpuOmp2Threads. [TEST at HAL]
```
- Acc: AccCpuOmp2Threads<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 22369621), blockThreadExtent: (2, 24), threadElemExtent: (1, 1)}
threadsPerBlock:24, threadsPerBlockLimit:64
```

mehmetyusufoglu: https://github.com/alpaka-group/alpaka/actions/runs/10170404404/job/28129487668?pr=2333

fwyzard: My idea was that
  - `threadsPerGridTestValue` is the maximum number of threads per grid (`props.m_blockThreadCountMax * props.m_gridBlockCountMax`)
  - `Vec{8, threadsPerGridTestValue / threadsPerBlock / 8}` × `Vec{2, threadsPerBlock}` should give a total of `2 * threadsPerGridTestValue / threadsPerBlock * threadsPerBlock / 8 * 8` which I expected to be bigger than `threadsPerGridTestValue`

fwyzard: That is, I don't expect to exceed the block size, but the total grid size.

mehmetyusufoglu: I checked the old `isValidWorkDiv` and `isValidWorkDivForKernel`; what you described is never checked in alpaka it seems.
Both use `accDevProps.m_gridBlockExtentMax[i] `and `accDevProps.m_blockThreadExtentMax[i] `values.  They check each extent of extent-array one by one. Secondly they check grid-block and block-thread total values. `accDevProps.m_gridBlockCountMax < gridBlockExtent.prod() `. But there is no  max-threads-per-grid in Device Properties. 

I can add the following check
  `If(gridBlockExtent.prod()*blockThreadExtent.prod() > accDevProps.m_gridBlockCountMax * accDevProps.m_BlockThreadCountMax) return false;
`

fwyzard: Ah, I see !

Let me update the PR to test what we actually check in Alpaka first.

Before adding the new check we should check if it actually fails on the various backends, like CUDA.

mehmetyusufoglu: I added it but accDevProps of Omp2 threads is too large.

- Acc: AccCpuOmp2Threads<2,unsigned int>, invalidWorkDiv: {gridBlockExtent: (8, 22369621), blockThreadExtent: (2, 24), threadElemExtent: (1, 1)}
threadsPerBlock:24, threadsPerBlockLimit:64
calculation: if(178956968 *48 > 4294967295 * 64) return false;
check:4294967168 > 4294967232 returns true because accDevProps.m_gridBlockCountMax is too large for Ompthreads;

fwyzard: OK, also on the CpuSerial backend, it seems that the maximum grid size is not validated properly.

`props.m_gridBlockCountMax` is `2147483647` (the maximum value for a signed 32-bit integer), but launching a kernel with `4294967294` (twice that) blocks still works:
```
[+] QueueGenericThreadsBlocking
[-] QueueGenericThreadsBlocking
createTaskKernel workDiv: {gridBlockExtent: (4294967294), blockThreadExtent: (1), threadElemExtent: (1)}, kernelFnObj: TestKernelWithManyRegisters
[+] operator()
operator() blockSharedMemDynSizeBytes: 0 B
The sum is 1492.00, the argument is 200
The sum is 1492.00, the argument is 200
The sum is 1492.00, the argument is 200
The sum is 1492.00, the argument is 200
The sum is 1492.00, the argument is 200
The sum is 1492.00, the argument is 200
The sum is 1492.00, the argument is 200
...
```

fwyzard: For the time being I'll update the test to check only the block size, not the grid size.

We can follow up on the grid size check on https://github.com/alpaka-group/alpaka/issues/2334 .

mehmetyusufoglu: Great info, thanks.

psychocoderHPC (approved):

PR #2335, opened 2024-07-31
Fix `isValidWorkDivKernel` to use the correct device

fwyzard:
Currently `isValidWorkDivKernel` does not use the device passed as an argument to determine the kernel properties: instead, it creates a new device taking the first one from the accelerator platform.

This can have two problems:
  - the devices in the accelerator platform may have different properties; for example, one can have two different NVIDIA GPUs in the same machine (_e.g._ a V100 and an L4) that lead to  different kernel attributes; checking the attributes from one GPU and running on the other may give unexpected results.
  - some platforms (_e.g._ SYCL) have stateful device objects, so it's better not to create a new one every time.

The fix is to use the device that is passed as an argument to query the kernel attributes.

### DIFF
--- a/include/alpaka/workdiv/WorkDivHelpers.hpp
+++ b/include/alpaka/workdiv/WorkDivHelpers.hpp
@@ -430,31 +430,25 @@ namespace alpaka
 
     //! \tparam TDim The dimensionality of the accelerator device properties.
     //! \tparam TIdx The idx type of the accelerator device properties.
-    //! \tparam TKernelBundle The type of the bundle of kernel and the arguments. Kernel is used to get number of
-    //! threads per block, this number could be less than or equal to the number of threads per block according to
-    //! device properties.
     //! \tparam TWorkDiv The type of the work division.
     //! \param accDevProps The maxima for the work division.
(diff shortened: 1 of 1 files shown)


fwyzard:
@mehmetyusufoglu can you review this PR ?

fwyzard:
By the way, since we have `getValidWorkDivForKernel`, would it make sense to rename `isValidWorkDivKernel` to `isValidWorkDivForKernel` ?

mehmetyusufoglu:
Yes it is renamed. The old one remains.

31 Tem 2024 Çar 10:22 tarihinde Andrea Bocci ***@***.***>
şunu yazdı:

mehmetyusufoglu:
> @mehmetyusufoglu can you review this PR ?

Yes, you are right GetFunctionAttributes and GetDeviceProperties should use the same device.

psychocoderHPC:
> By the way, since we have `getValidWorkDivForKernel`, would it make sense to rename `isValidWorkDivKernel` to `isValidWorkDivForKernel` ?

IMO good idea, currently we can do such changes be cause the interface is very new anyway.

psychocoderHPC (approved):

PR #2337, opened 2024-07-31
Implement device functions to simplify writing kernel code

fwyzard:
Implement device functions to simplify writing kernel code:
  - `oncePerGrid(acc)`;
  - `oncePerBlock(acc)`;
  - `uniformElements(acc, size)` and `uniformElements(acc, begin, end)`;
  - `uniformGroups(acc, size)` and `uniformGroupElements(acc, group, size)`;
  - `independentGroups(acc, groups)` and `independentGroupElements(acc, group_size)`;
  - variations of these functions along different dimensions.

Implement tests for the most common functions.


include/alpaka/exec/Once.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+    /* oncePerGrid
+     *
+     * `oncePerGrid(acc)` returns true for a single thread within the kernel execution grid.
+     *
+     * Usually the condition is true for block 0 and thread 0, but these indices should not be relied upon.
+     */
+
+    template<typename TAcc, typename = std::enable_if_t<isAccelerator<TAcc>>>
+    ALPAKA_FN_ACC inline constexpr bool oncePerGrid(TAcc const& acc)
```
mehmetyusufoglu, line 26: Maybe?
```suggestion
    ALPAKA_FN_ACC inline constexpr bool trueOncePerGrid(TAcc const& acc)
```

test/unit/exec/src/IndependentElements.cpp
```diff
@@ -0,0 +1,204 @@
...
+#    pragma clang diagnostic ignored "-Wexit-time-destructors"
+#endif
+// Global Host object used by all tests.
+using Host = alpaka::DevCpu;
+static Host host = alpaka::getDevByIdx(alpaka::PlatformCpu{}, 0);
+#if BOOST_COMP_CLANG
+#    pragma clang diagnostic pop
+#endif
+
+/* Add the group id to te value of each element in the group.
```
mehmetyusufoglu:
```suggestion
/* Add the group id to the value of each element in the group.
```

test/unit/exec/src/Once.cpp
```diff
@@ -0,0 +1,137 @@
...
+            ALPAKA_CHECK(*status, *value == 0);
+
+            alpaka::atomicAdd(acc, value, 1, alpaka::hierarchy::Grids{});
+
+            ALPAKA_CHECK(*status, *value == 1);
+        }
+    }
+};
+
+// The host and deivce compiler and libraries may use a different representation for boolean values. Reasonably enough,
```
mehmetyusufoglu:
```suggestion
// The host and device compiler and libraries may use a different representation for boolean values. Reasonably enough,
```

test/unit/exec/src/IndependentElements.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+    // Initialise the accelerator platform.
+    Platform platform{};
+
+    // Random number generator with a gaussian distribution.
+    std::random_device rd{};
+    std::default_random_engine engine{rd()};
+
+    // Uniform distribution.
+    std::uniform_int_distribution<Idx> random_size{100, 201};
```
mehmetyusufoglu, line 93:
```suggestion
    std::uniform_int_distribution<Idx> randomSize{100, 201};
```

test/unit/exec/src/IndependentElements.cpp
```diff
@@ -0,0 +1,204 @@
...
+    }
+
+    // Tolerance.
+    constexpr float epsilon = 0.000001f;
+
+    // Buffer size.
+    const Idx size = indices_h[groups];
+
+    // Allocate the input and output host buffer in pinned memory accessible by the Platform devices.
+    auto in_h = alpaka::allocMappedBuf<float, Idx>(host, platform, size);
```
mehmetyusufoglu, line 115: Camel Case variable naming?
```suggestion
    auto inH = alpaka::allocMappedBuf<float, Idx>(host, platform, size);
```

include/alpaka/exec/Once.hpp
fwyzard: No.

test/unit/exec/src/IndependentElements.cpp
fwyzard: Will fix, thanks.

test/unit/exec/src/Once.cpp
fwyzard: Will fix, thanks.

test/unit/exec/src/IndependentElements.cpp
fwyzard: I don't particularly like this naming style.
I follow it for user-facing function names, like in the rest of alpaka, but I don't like to use it for variable names.

mehmetyusufoglu: Yes, i also don't like actually.

fwyzard:
let's see if the 23rd time's a charm...

SimeonEhrig:
I check your code locally. The problem is the `AccCpuThreads` backend. On our workstation with 32 Core Epyc and Clang 17 debug build, the test already runs 100s. I think the annotation of the thread sanitizer also add a big overhead and the CI runners are weaker than our workstation.

I suggest to disable the tests with `std::thread` backend or if it make sense to test the thread backend, decrease the problem size.

fwyzard:
> I suggest to disable the tests with `std::thread` backend or if it make sense to test the thread backend, decrease the problem size.

I agree... I've tried reducing the block size in all tests to 32 threads (or elements) per block.

fwyzard:
Closed in favour of https://github.com/alpaka-group/alpaka/pull/2369 that includes further fixes and clean up.

PR #2339, opened 2024-08-01
change `CreateTaskKernel` interface

psychocoderHPC:
Use `KernelBundle` for task kernel creation.

This is the first part of the split PR #2318

PR #2340, opened 2024-08-01
refactor examples/tests + additional `alpaka::exec()` interface

psychocoderHPC:
# new interface
- add new interface `exec(queue, workDiv, kernelBundle)`
- add new interface `createKernelTask(workDiv, kernelBundle)`

We observed a tiny overhead when we start kernel with 100 arguments or an 1000 byte array as argument, see https://github.com/alpaka-group/alpaka/pull/2339#issuecomment-2268504326
Overhead is <1ms for `10000` kernel invocations.

# refactoring
- rename `bundeledKernelFill` into `kernelBundle`
- use the kernel bundle everywhere where we already have this object for the valid work division creation

This PR replaces #2339. Due to performance issues observed in #2339 the `TaskKernel` objects will keep the kernel instances and arguments separately and will not use `KernelBundle`

changes coming from PR https://github.com/alpaka-group/alpaka/pull/2318

### DIFF
--- a/example/convolution1D/src/convolution1D.cpp
+++ b/example/convolution1D/src/convolution1D.cpp
@@ -140,7 +140,7 @@ auto example(TAccTag const&) -> int
     DataType* nativeInputDeviceMemory = std::data(inputDeviceMemory);
     DataType* nativeOutputDeviceMemory = std::data(outputDeviceMemory);
 
-    auto const& bundeledKernel = alpaka::KernelBundle(
+    auto const& kernelBundle = alpaka::KernelBundle(
         convolutionKernel,
         nativeInputDeviceMemory,
         nativeFilterDeviceMemory,
@@ -150,17 +150,9 @@ auto example(TAccTag const&) -> int
 
     // Let alpaka calculate good block and grid sizes given our full problem extent
     auto const workDiv
-        = alpaka::getValidWorkDivForKernel<DevAcc>(devAcc, bundeledKernel, threadsPerGrid, elementsPerThread);
+        = alpaka::getValidWorkDivForKernel<DevAcc>(devAcc, kernelBundle, threadsPerGrid, elementsPerThread);
     // Run the kernel
-    alpaka::exec<DevAcc>(
-        queue,
-        workDiv,
-        convolutionKernel,
-        nativeInputDeviceMemory,
-        nativeFilterDeviceMemory,
-        nativeOutputDeviceMemory,
-        inputSize,
-        filterSize);
+    alpaka::exec<DevAcc>(queue, workDiv, kernelBundle);
 
     // Allocate memory on host
     auto resultGpuHost = alpaka::allocBuf<DataType, Idx>(devHost, inputSize);
--- a/example/heatEquation/src/heatEquation.cpp
+++ b/example/heatEquation/src/heatEquation.cpp
@@ -134,9 +134,9 @@ auto example(TAccTag const&) -> int
 
     HeatEquationKernel heatEqKernel;
 
-    auto const& bundeledKernel = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+    auto const& kernelBundle = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
     // Let alpaka calculate good block and grid sizes given our full problem extent
-    auto const workDiv = alpaka::getValidWorkDivForKernel<Acc>(devAcc, bundeledKernel, extent, elemPerThread);
+    auto const workDiv = alpaka::getValidWorkDivForKernel<Acc>(devAcc, kernelBundle, extent, elemPerThread);
 
     // Copy host -> device
     alpaka::memcpy(queue, uCurrBufAcc, uCurrBufHost);
--- a/example/helloWorld/src/helloWorld.cpp
+++ b/example/helloWorld/src/helloWorld.cpp
@@ -133,12 +133,11 @@ auto example(TAccTag const&) -> int
     // Kernels can be everything that is trivially copyable, has a
     // callable operator() and takes the accelerator as first
     // argument. So a kernel can be a class or struct, a lambda, etc.
-    HelloWorldKernel helloWorldKernel;
+    alpaka::KernelBundle kernelBundle = HelloWorldKernel{};
 
-    auto const& bundeledKernel = alpaka::KernelBundle(helloWorldKernel);
     // Let alpaka calculate good block and grid sizes given our full problem extent
     auto const workDiv
-        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, bundeledKernel, threadsPerGrid, elementsPerThread);
+        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, kernelBundle, threadsPerGrid, elementsPerThread);
 
     // Run the kernel
     //
@@ -149,11 +148,7 @@ auto example(TAccTag const&) -> int
     // The queue can be blocking or non-blocking
     // depending on the chosen queue type (see type definitions above).
     // Here it is synchronous which means that the kernel is directly executed.
-    alpaka::exec<Acc>(
-        queue,
-        workDiv,
-        helloWorldKernel
-        /* put kernel arguments here */);
+    alpaka::exec<Acc>(queue, workDiv, kernelBundle);
     alpaka::wait(queue);
 
     return EXIT_SUCCESS;
--- a/example/randomCells2D/src/randomCells2D.cpp
+++ b/example/randomCells2D/src/randomCells2D.cpp
@@ -202,15 +202,16 @@ auto example(TAccTag const&) -> int
 
     auto pitchBufAccRandV = alpaka::getPitchesInBytes(bufAccRandV)[0];
 
-    auto const& bundeledKernelInitRandom
+    auto const& kernelBundleInitRandom
         = alpaka::KernelBundle(initRandomKernel, extent, ptrBufAccRandS, pitchBufAccRandS);
     // Let alpaka calculate good block and grid sizes given our full problem extent
     auto const workDivInitRandom
-        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, bundeledKernelInitRandom, extent, Vec(perThreadY, perThreadX));
+        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, kernelBundleInitRandom, extent, Vec(perThreadY, perThreadX));
 
-    alpaka::exec<Acc>(queue, workDivInitRandom, initRandomKernel, extent, ptrBufAccRandS, pitchBufAccRandS);
+    alpaka::exec<Acc>(queue, workDivInitRandom, kernelBundleInitRandom);
     alpaka::wait(queue);
 
+    // execute the same kernel with different pointers
     alpaka::exec<Acc>(queue, workDivInitRandom, initRandomKernel, extent, ptrBufAccRandV, pitchBufAccRandV);
     alpaka::wait(queue);
 
@@ -230,7 +231,7 @@ auto example(TAccTag const&) -> int
     alpaka::memcpy(queue, bufAccS, bufHostS);
     RunTimestepKernelSingle runTimestepKernelSingle;
 
-    auto const& bundeledKernelRuntimeStep = alpaka::KernelBundle(
+    auto const& kernelBundleRuntimeStep = alpaka::KernelBundle(
         runTimestepKernelSingle,
         extent,
         ptrBufAccRandS,
@@ -239,21 +240,10 @@ auto example(TAccTag const&) -> int
         pitchBufAccS);
 
     // Let alpaka calculate good block and grid sizes given our full problem extent
-    auto const workDivRuntimeStep = alpaka::getValidWorkDivForKernel<Acc>(
-        devAcc,
-        bundeledKernelRuntimeStep,
-        extent,
-        Vec(perThreadY, perThreadX));
+    auto const workDivRuntimeStep
+        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, kernelBundleRuntimeStep, extent, Vec(perThreadY, perThreadX));
 
-    alpaka::exec<Acc>(
-        queue,
-        workDivRuntimeStep,
-        runTimestepKernelSingle,
-        extent,
-        ptrBufAccRandS,
-        ptrBufAccS,
-        pitchBufAccRandS,
-        pitchBufAccS);
+    alpaka::exec<Acc>(queue, workDivRuntimeStep, kernelBundleRuntimeStep);
     alpaka::memcpy(queue, bufHostS, bufAccS);
 
     auto pitchBufAccV = alpaka::getPitchesInBytes(bufAccV)[0];
--- a/example/randomStrategies/src/randomStrategies.cpp
+++ b/example/randomStrategies/src/randomStrategies.cpp
@@ -247,7 +247,7 @@ void runStrategy(Box<TAccTag>& box)
     // the initial parameters solely from the thread index
 
 
-    auto const& bundeledKernel = alpaka::KernelBundle(
+    auto const& kernelBundle = alpaka::KernelBundle(
         initRandomKernel,
         box.extentRand,
         ptrBufAccRand,
@@ -256,21 +256,14 @@ void runStrategy(Box<TAccTag>& box)
     // Let alpaka calculate good block and grid sizes given our full problem extent
     auto const workDivRand = alpaka::getValidWorkDivForKernel<typename Box<TAccTag>::Acc>(
         alpaka::getDevByIdx(box.accPlatform, 0),
-        bundeledKernel,
+        kernelBundle,
         box.extentRand,
         typename Box<TAccTag>::Vec(typename Box<TAccTag>::Idx{1}),
         false,
         alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
 
 
-    alpaka::exec<typename Box<TAccTag>::Acc>(
-        box.queue,
-        workDivRand,
-        initRandomKernel,
-        box.extentRand,
-        ptrBufAccRand,
-        static_cast<unsigned>(
-            box.extentResult[0] / box.extentRand[0])); // == NUM_ROLLS; amount of work to be performed by each thread
+    alpaka::exec<typename Box<TAccTag>::Acc>(box.queue, workDivRand, kernelBundle);
 
     alpaka::wait(box.queue);
 
@@ -291,27 +284,20 @@ void runStrategy(Box<TAccTag>& box)
     alpaka::memcpy(box.queue, box.bufAccResult, box.bufHostResult);
     FillKernel fillKernel;
 
-    auto const& bundeledKernelFill
-        = alpaka::KernelBundle(fillKernel, box.extentResult, ptrBufAccRand, ptrBufAccResult);
+    auto const& kernelBundleFill = alpaka::KernelBundle(fillKernel, box.extentResult, ptrBufAccRand, ptrBufAccResult);
 
     // Let alpaka calculate good block and grid sizes given our full problem extent
     auto const workdivResult = alpaka::getValidWorkDivForKernel<typename Box<TAccTag>::Acc>(
         alpaka::getDevByIdx(box.accPlatform, 0),
-        bundeledKernelFill,
+        kernelBundleFill,
         box.extentResult,
-        typename Box<TAccTag>::Vec(static_cast<typename Box<TAccTag>::Idx>(
-            NUM_ROLLS)), // One thread per "point"; each performs NUM_ROLLS "rolls"
+        // One thread per "point"; each performs NUM_ROLLS "rolls"
+        typename Box<TAccTag>::Vec(static_cast<typename Box<TAccTag>::Idx>(NUM_ROLLS)),
         false,
         alpaka::GridBlockExtentSubDivRestrictions::Unrestricted);
 
 
-    alpaka::exec<typename Box<TAccTag>::Acc>(
-        box.queue,
-        workdivResult,
-        fillKernel,
-        box.extentResult,
-        ptrBufAccRand,
-        ptrBufAccResult);
+    alpaka::exec<typename Box<TAccTag>::Acc>(box.queue, workdivResult, kernelBundleFill);
     alpaka::memcpy(box.queue, box.bufHostResult, box.bufAccResult);
     alpaka::wait(box.queue);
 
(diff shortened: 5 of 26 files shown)


mehmetyusufoglu:
Great, thanks. I know it is draft, just to add, the only issue is `isKernelArgumentTriviallyCopyable` check. [It was moved into  KernelBundle class at #2318  since it is not used anywhere else and better to check there. Because KernelBundle is used if the kernel is executed somehow; either before exec or inside the overloaded exec.]

include/alpaka/kernel/Traits.hpp
```diff
@@ -375,6 +362,30 @@ namespace alpaka
     ALPAKA_FN_HOST auto exec(TQueue& queue, TWorkDiv const& workDiv, TKernelFnObj const& kernelFnObj, TArgs&&... args)
         -> void
     {
-        enqueue(queue, createTaskKernel<TAcc>(workDiv, kernelFnObj, std::forward<TArgs>(args)...));
+        auto const& bundeledKernel = alpaka::KernelBundle(kernelFnObj, std::forward<TArgs>(args)...);
```
mehmetyusufoglu: typo at variable name `bundeledKernel`, `kernelBundle`

include/alpaka/kernel/Traits.hpp
```diff
@@ -328,42 +332,25 @@ namespace alpaka
...
 
 #if BOOST_COMP_CLANG
 #    pragma clang diagnostic push
 #    pragma clang diagnostic ignored                                                                                  \
         "-Wdocumentation" // clang does not support the syntax for variadic template arguments "args,..."
 #endif
-//! Executes the given kernel in the given queue.
-//!
+    //! Executes the given kernel in the given queue.
+    //!
```
mehmetyusufoglu, line 348: [nit] additional line not needed.

psychocoderHPC:
> Great, thanks. I know it is draft, just to add, the only issue is `isKernelArgumentTriviallyCopyable` check. [It was moved into KernelBundle class at #2318 since it is not used anywhere else and better to check there. Because KernelBundle is used if the kernel is executed somehow; either before exec or inside the overloaded exec.]

Thanks for pointing it out. I found it in #2339 too and reverted it. We can not move this trait away because this will mean it is hard to find for the user. Letzt have a look in a follow-up PR if it is solvable.

psychocoderHPC:
@fwyzard Could you have a look into this PR

example/heatEquation/src/heatEquation.cpp
```diff
@@ -146,8 +146,10 @@ auto example(TAccTag const&) -> int
 
     for(uint32_t step = 0; step < numTimeSteps; step++)
     {
+        auto const& tmpKernelBundle = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+
         // Compute next values
-        alpaka::exec<Acc>(queue, workDiv, heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+        alpaka::exec<Acc>(queue, workDiv, tmpKernelBundle);
```
fwyzard: This is one example where I would prefer _not to_ introduce the extra kernel bundle, since it makes the code longer.

example/helloWorld/src/helloWorld.cpp
```diff
@@ -133,12 +133,11 @@ auto example(TAccTag const&) -> int
     // Kernels can be everything that is trivially copyable, has a
     // callable operator() and takes the accelerator as first
     // argument. So a kernel can be a class or struct, a lambda, etc.
-    HelloWorldKernel helloWorldKernel;
+    alpaka::KernelBundle kernelBundle = HelloWorldKernel{};
```
fwyzard, line 136: Shouldn't this be
```c++
alpaka::KernelBundle kernelBundle = alpaka::KernelBundle(HelloWorldKernel{});
```
?

fwyzard: Ah, the explicit constructor call is not needed, because it is not marked as `explicit` ?

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -202,16 +202,17 @@ auto example(TAccTag const&) -> int
...
     auto const workDivInitRandom
-        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, bundeledKernelInitRandom, extent, Vec(perThreadY, perThreadX));
+        = alpaka::getValidWorkDivForKernel<Acc>(devAcc, kernelBundleInitRandom, extent, Vec(perThreadY, perThreadX));
 
     alpaka::exec<Acc>(queue, workDivInitRandom, initRandomKernel, extent, ptrBufAccRandS, pitchBufAccRandS);
     alpaka::wait(queue);
 
-    alpaka::exec<Acc>(queue, workDivInitRandom, initRandomKernel, extent, ptrBufAccRandV, pitchBufAccRandV);
+    // execute the same kernel with different pointers
+    alpaka::exec<Acc>(queue, workDivInitRandom, kernelBundleInitRandom);
```
fwyzard: Isn't the previous call the one that should be replace with the `kernelBundle` ?
This one uses different arguments.

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -266,11 +266,7 @@ void runStrategy(Box<TAccTag>& box)
     alpaka::exec<typename Box<TAccTag>::Acc>(
         box.queue,
         workDivRand,
-        initRandomKernel,
-        box.extentRand,
-        ptrBufAccRand,
-        static_cast<unsigned>(
-            box.extentResult[0] / box.extentRand[0])); // == NUM_ROLLS; amount of work to be performed by each thread
+        kernelBundle); // == NUM_ROLLS; amount of work to be performed by each thread
```
fwyzard: I think the comment should be moved to line 254.

include/alpaka/kernel/KernelBundle.hpp
```diff
@@ -32,7 +32,6 @@ namespace alpaka
         {
         }
 
-    private:
```
fwyzard, line 35: Why no longer `private` ?

psychocoderHPC: We access the member from outside. Writing getter would be possible but IMO not strictly required

example/heatEquation/src/heatEquation.cpp
psychocoderHPC: It uses `getValidWorkDivForKernel` therefore the bundle is required anyway and it is already in the mainline. This changing the kernel call to avoid calling the kernel with different arguments as used in `getValidWorkDivForKernel`

example/helloWorld/src/helloWorld.cpp
psychocoderHPC: `auto kernelBundle = alpaka::KernelBundle(HelloWorldKernel{});`

will be the long why to write the same line of code I did. The constructor is no marked `explicit` but all types will be deducted because the second parameter of the constructor is a variadic template.

example/randomStrategies/src/randomStrategies.cpp
psychocoderHPC: I will remove this comment here. We have this comment on the line where `NUM_ROLLS` is used, I do not know why the comment is there. The coresponding lines have nothing to do with `NUM_ROLLS`

example/randomCells2D/src/randomCells2D.cpp
psychocoderHPC: well catched, thats a bug I introduced.

psychocoderHPC:
@fwyzard I updated the PR.

test/integ/sharedMem/src/sharedMem.cpp
```diff
@@ -160,8 +161,11 @@ TEMPLATE_LIST_TEST_CASE("sharedMem", "[sharedMem]", TestAccs)
     auto blockRetValsAcc = alpaka::allocBuf<Val, Idx>(devAcc, resultElemCount);
     alpaka::memcpy(queue, blockRetValsAcc, blockRetVals, resultElemCount);
 
-    // Create the kernel execution task.
-    auto const taskKernel = alpaka::createTaskKernel<Acc>(workDiv, kernel, std::data(blockRetValsAcc));
+
+    // Create the kernel execution task using the real data.
+    auto const& kernelBundle2 = alpaka::KernelBundle(kernel, std::data(blockRetValsAcc));
```
fwyzard: I think this could be simply
```suggestion
    auto const kernelBundle2 = alpaka::KernelBundle(kernel, std::data(blockRetValsAcc));
```
since it's a constructor call ?

fwyzard: However, I don't like that the bundle is created only to execute the kernel.

This is another example where I don't like forcing the use of the bundles.

test/integ/separableCompilation/src/main.cpp
```diff
@@ -111,23 +111,21 @@ TEMPLATE_LIST_TEST_CASE("separableCompilation", "[separableCompilation]", TestAc
     alpaka::memcpy(queueAcc, memBufAccA, memBufHostA);
     alpaka::memcpy(queueAcc, memBufAccB, memBufHostB);
 
-    auto const& bundeledKernel
-        = alpaka::KernelBundle(kernel, memBufAccA.data(), memBufAccB.data(), memBufAccC.data(), numElements);
+    auto const& kernelBundle = alpaka::KernelBundle(
+        kernel,
+        std::data(memBufAccA),
```
fwyzard, line 116: I do like `x.data()` more than `std::data(x)` (it's shorter, and the latter does not gain anything), but I guess it's a matter of taste.

include/alpaka/test/KernelExecutionFixture.hpp
```diff
@@ -69,21 +69,22 @@ namespace alpaka::test
...
 
 
-            auto bundeledKernel = alpaka::KernelBundle<TKernelFnObj, decltype(getPtrNative(bufAccResult)), TArgs...>(
-                kernelFnObj,
-                getPtrNative(bufAccResult),
-                std::forward<TArgs>(args)...);
+            auto const& kernelBundle
+                = alpaka::KernelBundle<TKernelFnObj, decltype(getPtrNative(bufAccResult)), TArgs...>(
+                    kernelFnObj,
+                    getPtrNative(bufAccResult),
```
fwyzard, line 75:
```suggestion
                    bufAccResult.data(),
```
or
```suggestion
                    std::data(bufAccResult),
```
?

example/heatEquation/src/heatEquation.cpp
fwyzard: It uses `getValidWorkDivForKernel` with one bundle, but it does not use _that_ bundle to launch the kernel, it creates a new one _only_ to launch the kernel with different arguments.

There is no reason to create and use the second bundle.

This is exactly the use case where one should _not_ use a bundle, and call `exec `directly with the arguments.

example/randomCells2D/src/randomCells2D.cpp
```diff
@@ -230,7 +231,7 @@ auto example(TAccTag const&) -> int
     alpaka::memcpy(queue, bufAccS, bufHostS);
     RunTimestepKernelSingle runTimestepKernelSingle;
 
-    auto const& bundeledKernelRuntimeStep = alpaka::KernelBundle(
+    auto const& kernelBundleRuntimeStep = alpaka::KernelBundle(
```
fwyzard, line 234:
```suggestion
    auto const kernelBundleRuntimeStep = alpaka::KernelBundle(
```
?

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -247,7 +247,7 @@ void runStrategy(Box<TAccTag>& box)
     // the initial parameters solely from the thread index
 
 
-    auto const& bundeledKernel = alpaka::KernelBundle(
+    auto const& kernelBundle = alpaka::KernelBundle(
```
fwyzard, line 250:
```suggestion
    auto const kernelBundle = alpaka::KernelBundle(
```

example/randomStrategies/src/randomStrategies.cpp
```diff
@@ -291,27 +284,20 @@ void runStrategy(Box<TAccTag>& box)
     alpaka::memcpy(box.queue, box.bufAccResult, box.bufHostResult);
     FillKernel fillKernel;
 
-    auto const& bundeledKernelFill
-        = alpaka::KernelBundle(fillKernel, box.extentResult, ptrBufAccRand, ptrBufAccResult);
+    auto const& kernelBundleFill = alpaka::KernelBundle(fillKernel, box.extentResult, ptrBufAccRand, ptrBufAccResult);
```
fwyzard, line 287:
```suggestion
    auto const kernelBundleFill = alpaka::KernelBundle(fillKernel, box.extentResult, ptrBufAccRand, ptrBufAccResult);
```

example/convolution1D/src/convolution1D.cpp
```diff
@@ -140,7 +140,7 @@ auto example(TAccTag const&) -> int
     DataType* nativeInputDeviceMemory = std::data(inputDeviceMemory);
     DataType* nativeOutputDeviceMemory = std::data(outputDeviceMemory);
 
-    auto const& bundeledKernel = alpaka::KernelBundle(
+    auto const& kernelBundle = alpaka::KernelBundle(
```
fwyzard, line 143:
```suggestion
    auto const kernelBundle = alpaka::KernelBundle(
```

fwyzard:
hi @psychocoderHPC, I've updated my benchmark to compare the performance of `alpaka::exec` with a kernel with explicit arguments, and the performance of `alpaka::exec` with the corresponding kernel bundle: 
[benchmark_2340.tar.gz](https://github.com/user-attachments/files/16530776/benchmark_2340.tar.gz)

The `Makefile` is simpler since this PR does not change the behaviour of `alpaka::exec` :-)

Running the two benchmarks on my laptop, this time pinning all jobs to a single core, I get:

### kernel.cc
```bash
$ taskset -c 15 ./kernel.cpu >& /dev/null; for N in {0..99}; do taskset -c 15 ./kernel.cpu | tail -n1; done
...
10000 launches took KernelOneArg:   1.16 ms     KernelStruct:   2.07 ms KernelArray:  12.67 ms
10000 launches took KernelOneArg:   1.26 ms     KernelStruct:   2.06 ms KernelArray:  12.92 ms
10000 launches took KernelOneArg:   1.14 ms     KernelStruct:   2.06 ms KernelArray:  11.32 ms
```

### bundle.cc
```bash
$ taskset -c 15 ./bundle.cpu >& /dev/null; for N in {0..99}; do taskset -c 15 ./bundle.cpu | tail -n1; done
...
10000 launches took KernelOneArg:   1.17 ms     KernelStruct:   3.30 ms KernelArray:  13.10 ms
10000 launches took KernelOneArg:   1.17 ms     KernelStruct:   3.30 ms KernelArray:  13.24 ms
10000 launches took KernelOneArg:   1.21 ms     KernelStruct:   2.10 ms KernelArray:  14.93 ms
```

So, using a kernel bundles is slower by a small but consistent amount.

I will try to understand where the slowdown comes from - but that can be fixed in a follow-up PR.

fwyzard:
By the way, I noticed another issue with the `alpaka::KernelBundle`: it is possible to construct a `KernelBundle` object with a kernel and a set or arguments that are incompatible with that kernel.

The compiler will complain only when one tries to use the bundle, but not when it is constructed.

example/randomCells2D/src/randomCells2D.cpp
psychocoderHPC: We have it already in the mainline. It is valid because `const&` is extending the live time of the temporary. 
I am no fan of using it but since @mehmetyusufoglu did it already everywhere I kept it.

psychocoderHPC:
> By the way, I noticed another issue with the `alpaka::KernelBundle`: it is possible to construct a `KernelBundle` object with a kernel and a set or arguments that are incompatible with that kernel.
> 
> The compiler will complain only when one tries to use the bundle, but not when it is constructed.

This can not avoided because we do not have the accelerator at this point to do the signature verification. This is done in `TaskKernel`

example/randomCells2D/src/randomCells2D.cpp
fwyzard: I see... OK, we can fix it later :-)

test/integ/sharedMem/src/sharedMem.cpp
psychocoderHPC: But it is required for `getValidWorkDivKernel`. We kept therefore the old interface but if you use `getValidWorkDivKernel` you should use the bundle for the kernel call to avoid signature mismatch which can end with a kernel crash due to invalid work extents.

fwyzard:
> > By the way, I noticed another issue with the `alpaka::KernelBundle`: it is possible to construct a `KernelBundle` object with a kernel and a set or arguments that are incompatible with that kernel.
> > The compiler will complain only when one tries to use the bundle, but not when it is constructed.
> 
> This can not avoided because we do not have the accelerator at this point to do the signature verification. This is done in `TaskKernel`

I see... and we need the accelerator because the kernel is templated on the accelerator type 🤔

test/integ/separableCompilation/src/main.cpp
psychocoderHPC: yes this is matter of tast and was in the code before too.

test/integ/sharedMem/src/sharedMem.cpp
fwyzard: I understand - but we are not using `kernelBundle2` with `getValidWorkDivKernel`.

psychocoderHPC:
> > > By the way, I noticed another issue with the `alpaka::KernelBundle`: it is possible to construct a `KernelBundle` object with a kernel and a set or arguments that are incompatible with that kernel.
> > > The compiler will complain only when one tries to use the bundle, but not when it is constructed.
> > 
> > 
> > This can not avoided because we do not have the accelerator at this point to do the signature verification. This is done in `TaskKernel`
> 
> I see... and we need the accelerator because the kernel is templated on the accelerator type 🤔

Yes that's the problem. In the first drafts we carried the bundle with the accelerator but in general, this is also not nice if you would like to re-use the kernel for a different accelerator.

include/alpaka/test/KernelExecutionFixture.hpp
psychocoderHPC: I moved only code without changing it.

psychocoderHPC:
@fwyzard Is there a real show stopper for merging this code. IMO your comments are mostly style or best practices.
Let us talk about this in a developer meeting and discuss how we would like to handle `.data()`, `getNativePtr()` and `auto const & = KernelBundle` and apply changes in an orthogonal PR to the full code base.

fwyzard:
I would like to revert these two changes, where the kernel bundle is used unnecessarily:

### `example/heatEquation/src/heatEquation.cpp`
```diff
+        auto const& tmpKernelBundle = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+
         // Compute next values
-        alpaka::exec<Acc>(queue, workDiv, heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
+        alpaka::exec<Acc>(queue, workDiv, tmpKernelBundle);
```

### `test/integ/sharedMem/src/sharedMem.cpp`
```diff
-    // Create the kernel execution task.
-    auto const taskKernel = alpaka::createTaskKernel<Acc>(workDiv, kernel, std::data(blockRetValsAcc));
+
+    // Create the kernel execution task using the real data.
+    auto const& kernelBundle2 = alpaka::KernelBundle(kernel, std::data(blockRetValsAcc));
+    auto const taskKernel = alpaka::createTaskKernel<Acc>(workDiv, kernelBundle2);
+
```

The rest is OK.

test/integ/sharedMem/src/sharedMem.cpp
psychocoderHPC: fair point. In prinziple the example is broken because the workdiv is assumed to work for the second kernel too. I will change this line back

psychocoderHPC:
> I would like to revert these two changes, where the kernel bundle is used unnecessarily:
> 
> ### `example/heatEquation/src/heatEquation.cpp`
> ```diff
> +        auto const& tmpKernelBundle = alpaka::KernelBundle(heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
> +
>          // Compute next values
> -        alpaka::exec<Acc>(queue, workDiv, heatEqKernel, pCurrAcc, pNextAcc, numNodesX, dx, dt);
> +        alpaka::exec<Acc>(queue, workDiv, tmpKernelBundle);
> ```
> 
> ### `test/integ/sharedMem/src/sharedMem.cpp`
> ```diff
> -    // Create the kernel execution task.
> -    auto const taskKernel = alpaka::createTaskKernel<Acc>(workDiv, kernel, std::data(blockRetValsAcc));
> +
> +    // Create the kernel execution task using the real data.
> +    auto const& kernelBundle2 = alpaka::KernelBundle(kernel, std::data(blockRetValsAcc));
> +    auto const taskKernel = alpaka::createTaskKernel<Acc>(workDiv, kernelBundle2);
> +
> ```
> 
> The rest is OK.

done

fwyzard (approved):

fwyzard:
By the way, after reading the code for `alpaka::exec`, `alpaka::createTaskKernel` and `alpaka::KernelBundle`, I think that the extra copy of the kernel arguments is unavoidable when using the bundle.

With the current implementation, constructing the `KernelBundle` requires copying the arguments into the bundle itself.

Changing the bundle to store only a `const&` to the argument would avoid the copy, but it would not be safe: the arguments could go out of scope, or be modified, after the bundle is created but before the kernel is launched.

Changing the bundle to "move" the arguments when launching the kernel may not be feasible, may not be safe (if we reuse the bundle for multiple launches), it wouldn't help anyway for types like large structs.

One possible (future) improvement could be to change the way the CUDA kernels are launched to use `cudaLaunchKernel` and use the bundle to store the temporary arguments, instead of making an extra copy. However, this helps only for the CUDA backend (where the launch cost is anyway much larger that the cost of the extra copy), not for the various CPU backends.

psychocoderHPC:
@fwyzard I have set this PR to draft that it not get merged. There is an alternative to the kernel bundle. We could also change the `getValidWorkDivKernel` function signature to avoid the kernel bundle completly. 
This is before the release still possible.

fwyzard:
I'm actually fine to merge the PR, as long as the alternative of not using the `KernelBundle` is kept.

On the other hand, we could have a look at what a kernel graph node looks like in CUDA graphs, and see if can adapt the `KernalBundle` to match ?

psychocoderHPC:
this PR is closed because we switched back to old interfaces where kernel bundle is not used anymore https://github.com/alpaka-group/alpaka/pull/2349

PR #2342, opened 2024-08-03
Make alpaka follow `CMAKE_CUDA_RUNTIME_LIBRARY`

fwyzard:
CMake uses [`CMAKE_CUDA_RUNTIME_LIBRARY`](https://cmake.org/cmake/help/latest/variable/CMAKE_CUDA_RUNTIME_LIBRARY.html) to link the static (`libcudart_static.a`) or shared (`libcudart.so`) CUDA runtime library.

This PR makes alpaka follow the same variable, and applies it also to the cuRAND libraries.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -543,11 +543,19 @@ if(alpaka_ACC_GPU_CUDA_ENABLE)
         endif()
 
         # Link the CUDA Runtime library
-        target_link_libraries(alpaka INTERFACE CUDA::cudart)
+        if(CMAKE_CUDA_RUNTIME_LIBRARY STREQUAL "Shared")
+            target_link_libraries(alpaka INTERFACE CUDA::cudart)
+        else()
+            target_link_libraries(alpaka INTERFACE CUDA::cudart_static)
+        endif()
 
         if(NOT alpaka_DISABLE_VENDOR_RNG)
             # Use cuRAND random number generators
-            target_link_libraries(alpaka INTERFACE CUDA::curand)
+            if(CMAKE_CUDA_RUNTIME_LIBRARY STREQUAL "Shared")
+                target_link_libraries(alpaka INTERFACE CUDA::curand)
+            else()
+                target_link_libraries(alpaka INTERFACE CUDA::curand_static)
+            endif()
         endif()
     else()
         message(FATAL_ERROR "Optional alpaka dependency CUDA could not be found!")


psychocoderHPC (approved):

cmake/alpakaCommon.cmake
```diff
@@ -543,11 +543,19 @@ if(alpaka_ACC_GPU_CUDA_ENABLE)
         endif()
 
         # Link the CUDA Runtime library
-        target_link_libraries(alpaka INTERFACE CUDA::cudart)
+        if(CMAKE_CUDA_RUNTIME_LIBRARY STREQUAL "Shared")
```
psychocoderHPC, line 546: By default `CMAKE_CUDA_RUNTIME_LIBRARY` is not set so this if condation will be false and we link static.
In all old alpaka versions we linked always against `cudart`. 
@fwyzard was the change of the default behavior of alpaka intended?

I run into issue with our main code PIConGPU when compiling with clang as CUDA compiler.

If there is no strong reason for the change of the default behaviour I could open a PR with 

```
if(NOT DEFINED CMAKE_CUDA_RUNTIME_LIBRARY) 
     set(CMAKE_CUDA_RUNTIME_LIBRARY "Shared") 
 endif() 
```

Which will keep the default behaviour of alpaka but still provide the possibility of static linking.

fwyzard: No, I did not mean to change the default behaviour.

However, the default behaviour before #2329 and this PR was kind of tricky.

If configured with `alpaka_DISABLE_VENDOR_RNG=OFF`:
  - the CUDA binaries (_e.g._ `axpyTest`) are linked with both `libcudart.so` and `-lcudart_static`; I think the shared library took precedence because it was coming first on the command line;
  - the CUDA host-only applications (_e.g._ `hostOnlyAPITest`) were linked with `libcudart.so`.

If configured with `alpaka_DISABLE_VENDOR_RNG=ON`:
  - the CUDA binaries (_e.g._ `axpyTest`) were linked with `cudart_static`;
  - the CUDA host-only applications (_e.g._ `hostOnlyAPITest`) were not linked with either library, and in fact failed to link.

fwyzard: If the default should be "shared" (I agree it should be), let me invert the logic of the checks:
```cmake
        if(CMAKE_CUDA_RUNTIME_LIBRARY STREQUAL "Static")
            target_link_libraries(alpaka INTERFACE CUDA::cudart_static)
        else()
            target_link_libraries(alpaka INTERFACE CUDA::cudart)
        endif()
```

fwyzard: See https://github.com/alpaka-group/alpaka/pull/2348 .

PR #2347, opened 2024-08-06
fix `getFunctionAttributes` for the SYCL backend

AuroraPerego:
`workDivTest` on the SYCL CPU and GPU backend was failing with the message:
```
getFunctionAttributes function is not specialised for the given arguments.
```
This PR fixes this error (not sure, but I assume the reason why the current specialization does not work may be that `AccCpuSycl` and `AccGpuSyclIntel` inherit from `AccGenericSycl` instead of being simple `using` as for the CUDA/HIP case).

### DIFF
--- a/include/alpaka/kernel/TaskKernelGpuSyclIntel.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuSyclIntel.hpp
@@ -16,6 +16,37 @@ namespace alpaka
     template<typename TDim, typename TIdx, typename TKernelFnObj, typename... TArgs>
     using TaskKernelGpuSyclIntel
         = TaskKernelGenericSycl<AccGpuSyclIntel<TDim, TIdx>, TDim, TIdx, TKernelFnObj, TArgs...>;
+
+    namespace trait
+    {
+        //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDev The device type.
+        //! \tparam TDim The dimensionality of the accelerator device properties.
+        //! \tparam TIdx The idx type of the accelerator device properties.
+        //! \tparam TKernelFn Kernel function object type.
+        //! \tparam TArgs Kernel function object argument types as a parameter pack.
+        template<typename TDev, typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
+        struct FunctionAttributes<AccGpuSyclIntel<TDim, TIdx>, TDev, KernelBundle<TKernelFn, TArgs...>>
+        {
+            //! \param dev The device instance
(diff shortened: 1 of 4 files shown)


include/alpaka/kernel/TaskKernelGpuSyclIntel.hpp
```diff
@@ -16,6 +16,37 @@ namespace alpaka
...
+
+    namespace trait
+    {
+        //! \brief Specialisation of the class template FunctionAttributes
+        //! \tparam TDev The device type.
+        //! \tparam TDim The dimensionality of the accelerator device properties.
+        //! \tparam TIdx The idx type of the accelerator device properties.
+        //! \tparam TKernelFn Kernel function object type.
+        //! \tparam TArgs Kernel function object argument types as a parameter pack.
+        template<typename TDev, typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
```
psychocoderHPC, line 28: could we not keep the implementation in `TaskKernelGenericSycl` and simple inherit from that? THis will avoid the code duplication.

```C++
template<typename TDev, typename TDim, typename TIdx, typename TKernelFn, typename... TArgs>
        struct FunctionAttributes<AccGpuSyclIntel<TDim, TIdx>, TDev, KernelBundle<TKernelFn, TArgs...>>
struct FunctionAttributes<AccGpuSyclIntel<TDim, TIdx>, TDev, KernelBundle<TKernelFn, TArgs...>> :  FunctionAttributes<AccGenericSycl<TDim, TIdx>, TDev, KernelBundle<TKernelFn, TArgs...>>{};
```

psychocoderHPC (dismissed):
This version is fine but maybe my suggestion will be more explicit that both backends are equal.

include/alpaka/kernel/TaskKernelGpuSyclIntel.hpp
AuroraPerego: I did not manage to make it work, I think the problem is the line:
```cpp
auto const& props = alpaka::getAccDevProps<AccGenericSycl<TDim, TIdx>>(dev);
```
where `AccGenericSycl` is used explicitly (this makes the two implementations different).

fwyzard:
@psychocoderHPC I think that, overall, the SYCL implementation needs to be refactored to reduce code duplication across all back-ends.
However I would prefer that is done in a separate PR, as it's going to touch a lot more files.

fwyzard:
@AuroraPerego can you make the same fix also in `TaskKernelFpgaSyclIntel.hpp` ?

psychocoderHPC (approved):

PR #2365, opened 2024-08-19
Add 2D heat equation example

ikbuibui:
Adds a 2D heat equation example for the upcoming alpaka workshop.
This example is supposed to show how alpaka should be used, best practices, etc. So comments and reviews are welcome and appreciated from all. 
This PR is a WIP but please review and comment in the state you find it in.


include/alpaka/wait/Traits.hpp
```diff
@@ -17,11 +17,11 @@ namespace alpaka
     namespace trait
     {
         //! The thread wait trait.
-        template<typename TAwaited, typename TSfinae = void>
```
psychocoderHPC: this should not be part of the PR and is not required. The sfinae param ,should not be removed.

example/heatEquation2D/src/BoundaryKernel.hpp
```diff
@@ -0,0 +1,98 @@
...
+//!                 u(x, y, t)  | t = t_current
+//! \param chunkSize
+//! \param pitch
+//! \param dx step in x
+//! \param dy step in y
+//! \param dt step in t
+
+struct BoundaryKernel
+{
+    template<typename TAcc, typename T_Chunk>
```
mehmetyusufoglu:
```suggestion
    template<typename TAcc, typename TChunk>
```

example/heatEquation2D/CMakeLists.txt
```diff
@@ -0,0 +1,61 @@
...
+
+project(${_TARGET_NAME} LANGUAGES CXX)
+
+
+################################################################################
+# PNGwriter
+################################################################################
+
+# find PNGwriter installation
+find_package(PNGwriter 0.7.0 REQUIRED CONFIG)
```
psychocoderHPC: if we set this to require this PR will not pass the CI.
Should we make this optional and guard the IO with `#ifdef`?

psychocoderHPC (changes requested):

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+template<size_t T_SharedMemSize1D>
+struct StencilKernel
+{
+    template<typename TAcc, typename TDim, typename TIdx>
+    ALPAKA_FN_ACC auto operator()(
+        TAcc const& acc,
+        double const* const uCurrBuf,
+        double* const uNextBuf,
+        alpaka::Vec<TDim, TIdx> const chunkSize,
+        alpaka::Vec<TDim, TIdx> const pitch,
```
mehmetyusufoglu: Can `pitch` be a struct consisting 2 variables; a scalar `typeSize` and a scalar `pitch` value.

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+        alpaka::syncBlockThreads(acc);
+
+        // Each kernel executes one element
+        double const r_x = dt / (dx * dx);
+        double const r_y = dt / (dy * dy);
+
+        // go over only core cells
+        for(auto i = threadIdx1D; i < chunkSize.prod(); i += numThreadsPerBlock)
+        {
+            auto idx2d = alpaka::mapIdx<2>(alpaka::Vec(i), chunkSize);
```
mehmetyusufoglu:
```suggestion
            auto idx2D = alpaka::mapIdx<2>(alpaka::Vec(i), chunkSize);
```

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+            auto idx2d = alpaka::mapIdx<2>(alpaka::Vec(i), chunkSize + halo);
+            idx2d = idx2d + blockStartIdx;
+            auto elem = getElementPtr(uCurrBuf, idx2d, pitch);
+            sdata[i] = *elem;
+        }
+
+        alpaka::syncBlockThreads(acc);
+
+        // Each kernel executes one element
+        double const r_x = dt / (dx * dx);
```
mehmetyusufoglu:
```suggestion
        double const rX = dt / (dx * dx);
```

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+        double const r_y = dt / (dy * dy);
```
mehmetyusufoglu:
```suggestion
        double const rY = dt / (dy * dy);
```

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#pragma once
+
+#include <alpaka/alpaka.hpp>
+
+#include <cmath>
+
+//! Exact solution to the test problem
+//! u_t(x, y, t) = u_xx(x, t) + u_yy(y, t), x in [0, 1], y in [0, 1], t in [0, T]
+//! u(0, t) = u(1, t) = 0
```
mehmetyusufoglu, line 13:
```suggestion
//! u(0,y, t) = u(1, y, t) = 0  0 ≤ y ≤ 1, t ≥ 0,
```

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,76 @@
...
+//! u(x, 0) = sin(pi * x)
```
mehmetyusufoglu, line 14: //! initial conditions
//! u(x,y,0) = sin (pi*x )*sin(pi*y), (x, y ) ∈ [0,1]x[0,1],

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -0,0 +1,194 @@
...
+    BufAcc uNextBufAcc{alpaka::allocBuf<double, Idx>(devAcc, extent)};
+    BufAcc uCurrBufAcc{alpaka::allocBuf<double, Idx>(devAcc, extent)};
+
+    // Set buffer to initial conditions
+    initalizeBuffer(uBufHost, extent, dx, dy);
+
+    // Select queue
+    using QueueProperty = alpaka::NonBlocking;
+    using QueueAcc = alpaka::Queue<Acc, QueueProperty>;
+    QueueAcc queue1{devAcc};
```
mehmetyusufoglu: May be betters names can be found for queue1 and queue2.

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+//! Initialize the buffer to the analytical solution at t=0
+//!
+//! \param buffer buffer holding the solution at tMax
+//! \param extent extents of the buffer
+//! \param dx
+//! \param dy
+
+template<typename T_Buffer, typename T_Extent>
```
mehmetyusufoglu:
```suggestion
template<typename TBuffer, typename TExtent>
```

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,76 @@
...
+
+#include <alpaka/alpaka.hpp>
+
+#include <cmath>
+
+//! Exact solution to the test problem
+//! u_t(x, y, t) = u_xx(x, t) + u_yy(y, t), x in [0, 1], y in [0, 1], t in [0, T]
+//! u(0, t) = u(1, t) = 0
+//! u(x, 0) = sin(pi * x)
+//! u(0, y) = sin(pi * y)
```
mehmetyusufoglu: I think initial and boundary conditions are not clear  in this problem definition.

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+        {
+            auto idx2d = alpaka::mapIdx<2>(alpaka::Vec(i), chunkSize);
+            idx2d = idx2d + alpaka::Vec<TDim, TIdx>{1, 1}; // offset for halo above and to the left
+            auto localIdx1D = alpaka::mapIdx<1>(idx2d, chunkSize + halo)[0u];
+
+
+            auto bufIdx = idx2d + blockStartIdx;
+            auto elem = getElementPtr(uNextBuf, bufIdx, pitch);
+
+            *elem = sdata[localIdx1D] * (1.0 - 2.0 * r_x - 2.0 * r_y) + sdata[localIdx1D - 1] * r_x
```
mehmetyusufoglu: Dividing the equation could be useful?

```
auto const left = sdata[localIdx1D - 1];
auto const right = sdata[localIdx1D + 1];
auto const down = sdata[localIdx1D-chunkSize[1] - 2];  // by the way, why -2
auto const up = sdata[localIdx1D+chunkSize[1] +2];  //   or why +2 ?
```

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+//! \tparam T_SharedMemSize1D size of the shared memory box
+//!
+//! Solving equation u_t(x, t) = u_xx(x, t) + u_yy(y, t) using a simple explicit scheme with
+//! forward difference in t and second-order central difference in x and y
+//!
+//! \param uCurrBuf Current buffer with grid values of u for each x and the current value of t:
+//!                 u(x, y, t) | t = t_current
+//! \param uNextBuf resulting grid values of u for each x and the next value of t:
+//!              u(x, y, t) | t = t_current + dt
+//! \param chunkSize
```
mehmetyusufoglu: Defining chunkSize and pitch is necessary

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -0,0 +1,194 @@
...
+        (extent[0] - haloSize[0] - 1) / chunkSize[0] + 1,
+        (extent[1] - haloSize[1] - 1) / chunkSize[1] + 1};
+
+    static_assert(
+        (extent[0] - haloSize[0]) % chunkSize[0] == 0 && (extent[1] - haloSize[1]) % chunkSize[1] == 0,
+        "Domain must be divisible by chunk size");
+
+    auto const pitchCurrAcc{alpaka::getPitchesInBytes(uCurrBufAcc)};
+
+    alpaka::WorkDivMembers<Dim, Idx> workDiv_manual{numChunks, threadsPerBlock, elemPerThread};
```
mehmetyusufoglu: [nit] We can print work division here. It will be informative for the users.
std::cout << " WorkDivision: " << workDivManual << std::endl;

example/heatEquation2D/CMakeLists.txt
ikbuibui: Yes, we can make this optional and gaurd with ifdefs. We can then also treat openPMD the same way as I assume it will have the same problem

include/alpaka/wait/Traits.hpp
ikbuibui: You are right, this sneaked in from some other branch by mistake. On a side note I think those sfinae params really are unused

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+#include <alpaka/alpaka.hpp>
+
+//! alpaka version of explicit finite-difference 2D heat equation solver
+//!
+//! \tparam T_SharedMemSize1D size of the shared memory box
+//!
+//! Solving equation u_t(x, t) = u_xx(x, t) + u_yy(y, t) using a simple explicit scheme with
+//! forward difference in t and second-order central difference in x and y
+//!
+//! \param uCurrBuf Current buffer with grid values of u for each x and the current value of t:
```
mehmetyusufoglu:
```suggestion
//! \param uCurrBuf Current buffer with grid values of u for each x,y pair and the current value of t:
```

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+//!                 u(x, y, t) | t = t_current
+//! \param uNextBuf resulting grid values of u for each x and the next value of t:
```
mehmetyusufoglu:
```suggestion
//! \param uNextBuf resulting grid values of u for each x,y pair and the next value of t:
```

example/heatEquation2D/src/analyticalSolution.hpp
mehmetyusufoglu: Screenshot from 2024-08-20 13-47-25

example/heatEquation2D/src/BoundaryKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/analyticalSolution.hpp
ikbuibui: We do not have Dirichlet or Neumann boundaries. We have an analytical solution to the equation with a special type of boundary condition not inspired from physics. This example is chosen because it is easy to verify the correctness of the finite difference stencil kernel and not because the physics is of any particular interest.
I have removed the outdated boundaries copied over from the 1D example

example/heatEquation2D/src/analyticalSolution.hpp
ikbuibui: Done

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Changed the +2 and -2 to the size of the halo. 
Splitting can be done, I will open this up to the opinion of others.

example/heatEquation2D/src/analyticalSolution.hpp
mehmetyusufoglu: Yes I see ( `u(0, t) = u(1, t) = 0` can not be initial condition for a 2D problem. )

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -0,0 +1,194 @@
...
+        alpaka::wait(queue1);
+        std::swap(uNextBufAcc, uCurrBufAcc);
+    }
+
+    // Copy device -> host
+    alpaka::memcpy(queue1, uBufHost, uCurrBufAcc);
+    alpaka::wait(queue1);
+
+    // Validate
+    auto const [resultCorrect, maxError] = validateSolution(uBufHost, extent, dx, dy, tMax);
```
chillenzer:
```suggestion
    auto const [resultIsCorrect, maxError] = validateSolution(uBufHost, extent, dx, dy, tMax);
```
means that the if condition reads like fluent English.

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -0,0 +1,194 @@
...
+    //  \code{.cpp}
+    //  auto tag = TagCpuSerial;
+    //  return example(tag);
+    //  \endcode
+    //
+    // valid tags:
+    //   TagCpuSerial, TagGpuHipRt, TagGpuCudaRt, TagCpuOmp2Blocks, TagCpuTbbBlocks,
+    //   TagCpuOmp2Threads, TagCpuSycl, TagCpuTbbBlocks, TagCpuThreads,
+    //   TagFpgaSyclIntel, TagGenericSycl, TagGpuSyclIntel
+    return alpaka::executeForEachAccTag([=](auto const& tag) { return example(tag); });
```
chillenzer, line 216:
```suggestion
    return alpaka::executeForEachAccTag([](auto const& tag) { return example(tag); });
```
We're not capturing anything, are we?

example/heatEquation2D/src/helpers.hpp
```diff
@@ -0,0 +1,19 @@
+/* Copyright 2020 Tapish Narwal
+ * SPDX-License-Identifier: ISC
+ */
+
+#pragma once
+
+#include <alpaka/alpaka.hpp>
+
+template<typename T, typename TDim, typename TIdx>
+ALPAKA_FN_ACC T* getElementPtr(T* ptr, alpaka::Vec<TDim, TIdx> idx, alpaka::Vec<TDim, TIdx> pitch)
```
chillenzer:
```suggestion
template<typename T>
ALPAKA_FN_ACC T* getElementPtr(T* ptr, auto idx, auto pitch)
```
Which C++ are we on? Can we do this already?

example/heatEquation2D/src/helpers.hpp
```diff
@@ -0,0 +1,19 @@
...
+{
+    return reinterpret_cast<T*>(reinterpret_cast<std::byte*>(ptr) + idx[0] * pitch[0] + idx[1] * pitch[1]);
+}
+
+template<typename T, typename TDim, typename TIdx>
+ALPAKA_FN_ACC T const* getElementPtr(T const* ptr, alpaka::Vec<TDim, TIdx> idx, alpaka::Vec<TDim, TIdx> pitch)
+{
+    return reinterpret_cast<T const*>(reinterpret_cast<std::byte const*>(ptr) + idx[0] * pitch[0] + idx[1] * pitch[1]);
+}
```
chillenzer: This can be merged with the one above if you use something like
```C++
template <typename T, typename U>
using const_like = std::conditional_t<std::is_const_v<T>, U const, U>;
```
and then `const_like<T, std::byte>*` for the cast.

example/heatEquation2D/src/StencilKernel.hpp
chillenzer: Splitting doesn't sound half bad. Alternatively, you could define the indices upfront, e.g., `auto const left = localIdx1D - 1` and so on.

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,76 @@
...
+//!
+//! \param x value of x
+//! \param x value of y
+//! \param t value of t
+
+ALPAKA_FN_HOST_ACC auto exactSolution(double const x, double const y, double const t) -> double
```
chillenzer, line 17:
```suggestion
ALPAKA_FN_HOST_ACC auto exactSolution(double const t, double const x, double const y) -> double
```
is what I would have written. That is also spiritually closer to the data layout isn't it? (I'm aware that we only hold slices for two t at any given time but it's still somehow the "slowest index".

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,76 @@
...
+{
+    constexpr double pi = 3.141592653589793238462643383279502884;
```
chillenzer: Now I wonder if you typed this by hand. XD

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Yes it is possible but I dont see a reason for it. alpaka::getPitchesInBytes already returns a alpaka Vec containing those variables, so we can just keep this object. It handles cases of different dimensions and is just as convenient as a struct would be.

chillenzer (commented):
Just a few minor things skimming through.

example/heatEquation2D/src/helpers.hpp
fwyzard: But it is not the same: the latter does not restrict the type to be a `Vec`, nor does it require the two types to be the same.

example/heatEquation2D/src/analyticalSolution.hpp
fwyzard: `M_PI` anyone ?

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+        TAcc const& acc,
+        double const* const uCurrBuf,
+        double* const uNextBuf,
+        alpaka::Vec<TDim, TIdx> const chunkSize,
+        alpaka::Vec<TDim, TIdx> const pitch,
+        double const dx,
+        double const dy,
+        double const dt) const -> void
+    {
+        auto& sdata(alpaka::declareSharedVar<double[T_SharedMemSize1D], __COUNTER__>(acc));
```
fwyzard:
```suggestion
        auto& sdata = alpaka::declareSharedVar<double[T_SharedMemSize1D], __COUNTER__>(acc);
```

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -0,0 +1,199 @@
...
+    }
+
+    // Initialize host-buffer
+    // This buffer will hold the current values (used for the next step)
+    auto uBufHost = alpaka::allocBuf<double, Idx>(devHost, extent);
+
+    // Accelerator buffer
+    using BufAcc = alpaka::Buf<Acc, double, Dim, Idx>;
+    BufAcc uNextBufAcc{alpaka::allocBuf<double, Idx>(devAcc, extent)};
+    BufAcc uCurrBufAcc{alpaka::allocBuf<double, Idx>(devAcc, extent)};
```
fwyzard: why `auto` on line 79 and an explicit type here ?
```suggestion
    auto uNextBufAcc = alpaka::allocBuf<double, Idx>(devAcc, extent);
    auto uCurrBufAcc = alpaka::allocBuf<double, Idx>(devAcc, extent);
```

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,73 @@
...
+
+//! Initialize the buffer to the analytical solution at t=0
+//!
+//! \param buffer buffer holding the solution at tMax
+//! \param extent extents of the buffer
+//! \param dx
+//! \param dy
+
+template<typename TBuffer, typename TExtent>
+auto initalizeBuffer(TBuffer& buffer, TExtent const& extent, double const dx, double const dy) -> void
```
fwyzard: The `extent` could be taken from the `buffer`.

If the `extent` is kept explicit, there should be a check that it is smaller or equal to the size of the `buffer`.

example/heatEquation2D/src/helpers.hpp
chillenzer: Yes, of course. But it is more pleasant for the eye and we know the precise context, this will be called from. I would argue that a beginner not familiar with alpaka cares less about repeating the type information twice (thrice even if you count the `TDim`, `TIdx` template arguments) and more about handling a manageable amount of information in their head. A double reinterpret cast wrapping a manual index calculation in the function body seems to me like enough to digest for a beginner.

fwyzard:
Just a thought: why not use the `mdspan` accessor for the 2D buffers ?

example/heatEquation2D/src/heatEquation2D.cpp
ikbuibui: Done

example/heatEquation2D/src/analyticalSolution.hpp
ikbuibui: Changed to take extents from buffer

example/heatEquation2D/src/heatEquation2D.cpp
ikbuibui: Changed to auto, it was a copy paste relic from the 1D heat equation example

example/heatEquation2D/src/helpers.hpp
ikbuibui: Still using C++17, so no we can't do this anyway

example/heatEquation2D/src/helpers.hpp
ikbuibui: Good point, renamed metafunction to ``const_match`` and merged the two ``getElementPtr``

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/analyticalSolution.hpp
ikbuibui: The value was taken from M_PIf64. I thought that macro was kind of ugly, so I defined the constexpr variable here. I can switch to M_PI if you prefer that.

fwyzard: Well, I guess that's a different point of view :-D

You can use `alpaka::math::constants::pi` from `alpaka/math/Traits.hpp`.

ikbuibui:
> Just a thought: why not use the `mdspan` accessor for the 2D buffers ?

This is a good point, and can be incorporated.
Currently after a discussion with Rene we were of the opinion that since pitches are a frequent tripping point when dealing with 2d buffers and people in the workshop may not be very familiar with GPU programming, it might be good to not hide that behind `mdspan`. 
On the other hand, making the example simpler to understand is a good thing, so I am still not sure which way to go. I will bring this up in the next alpaka meeting for a small discussion.

mehmetyusufoglu:
> > Just a thought: why not use the `mdspan` accessor for the 2D buffers ?
> 
> This is a good point, and can be incorporated. Currently after a discussion with Rene we were of the opinion that since pitches are a frequent tripping point when dealing with 2d buffers and people in the workshop may not be very familiar with GPU programming, it might be good to not hide that behind `mdspan`. On the other hand, making the example simpler to understand is a good thing, so I am still not sure which way to go. I will bring this up in the next alpaka meeting for a small discussion.

mdspan can be a second step. After showing the code with pitch / stride we can show mdspan.

example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -0,0 +1,198 @@
...
+
+    auto const maxThreadsPerBlock = alpaka::getAccDevProps<Acc>(devAcc).m_blockThreadExtentMax;
+    auto const threadsPerBlock = maxThreadsPerBlock.prod() < chunkSize.prod() ? maxThreadsPerBlock : chunkSize;
+
+    constexpr alpaka::Vec<Dim, Idx> numChunks{
+        (extent[0] - haloSize[0] - 1) / chunkSize[0] + 1,
+        (extent[1] - haloSize[1] - 1) / chunkSize[1] + 1};
+
+    static_assert(
+        (extent[0] - haloSize[0]) % chunkSize[0] == 0 && (extent[1] - haloSize[1]) % chunkSize[1] == 0,
```
mehmetyusufoglu: checking `numNodes/chunkSize` would be enough.

example/heatEquation2D/src/BoundaryKernel.hpp
```diff
@@ -0,0 +1,98 @@
...
+        auto const blockStartIdx = gridBlockIdx * chunkSize;
+
+        // apply boundary conditions
+        // top row
+        if(gridBlockIdx[0] == 0)
+        {
+            auto const globalIdx = blockStartIdx + alpaka::Vec<Dim, Idx>(0, 1);
+            for(auto i = threadIdx1D; i < chunkSize[1]; i += numThreadsPerBlock)
+            {
+                auto idx2D = globalIdx + alpaka::Vec<Dim, Idx>(0, i);
```
mehmetyusufoglu, line 59: a comment is needed for this line

example/heatEquation2D/src/StencilKernel.hpp
mehmetyusufoglu: ok

example/heatEquation2D/src/analyticalSolution.hpp
```diff
@@ -0,0 +1,74 @@
...
+
+//! Initialize the buffer to the analytical solution at t=0
+//!
+//! \param buffer buffer holding the solution at tMax
+//! \param extent extents of the buffer
+//! \param dx
+//! \param dy
+
+template<typename TBuffer>
+auto initalizeBuffer(TBuffer& buffer, double const dx, double const dy) -> void
```
mehmetyusufoglu, line 60: Double arguments are passed already pass-by-value, what is the meaning of `const` here?

fwyzard: While I'm not a fan of declaring by-value parameters as `const`, what it does is tell the compiler that the function body is not supposed to modify the value of the parameter.

It's like code like
```c++
{
  const int value = some_function();
  printf("value is %d", value);
}
```

The `const` is not really useful, but it doesn't harm.

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+        // Get extents(dimensions)
+        auto const gridBlockExtent = alpaka::getWorkDiv<alpaka::Grid, alpaka::Blocks>(acc);
+        auto const blockThreadExtent = alpaka::getWorkDiv<alpaka::Block, alpaka::Threads>(acc);
+        auto const numThreadsPerBlock = blockThreadExtent.prod();
+
+        // Get indexes
+        auto const gridBlockIdx = alpaka::getIdx<alpaka::Grid, alpaka::Blocks>(acc);
+        auto const blockThreadIdx = alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc);
+        auto const threadIdx1D = alpaka::mapIdx<1>(blockThreadIdx, blockThreadExtent)[0u];
+        auto const blockStartIdx = gridBlockIdx * chunkSize;
```
mehmetyusufoglu: [maybe] to make it shorter and less frightening :) ? 
```suggestion
         // Get the total number of threads per block
        auto const numThreadsPerBlock = alpaka::getWorkDiv<alpaka::Block, alpaka::Threads>(acc).prod();
        // 1D thread index within the block
        auto const threadIdx1D = alpaka::mapIdx<1>(alpaka::getIdx<alpaka::Block, alpaka::Threads>(acc), 
                                                   alpaka::getWorkDiv<alpaka::Block, alpaka::Threads>(acc))[0u];
        // Block's starting index in the grid
        auto const blockStartIdx = alpaka::getIdx<alpaka::Grid, alpaka::Blocks>(acc) * chunkSize;
```

example/heatEquation2D/src/StencilKernel.hpp
```diff
@@ -0,0 +1,87 @@
...
+
+        constexpr alpaka::Vec<TDim, TIdx> halo{2, 2};
+
+        for(auto i = threadIdx1D; i < T_SharedMemSize1D; i += numThreadsPerBlock)
+        {
+            auto idx2d = alpaka::mapIdx<2>(alpaka::Vec<alpaka::DimInt<1u>, TIdx>(i), chunkSize + halo);
+            idx2d = idx2d + blockStartIdx;
+            auto elem = getElementPtr(uCurrBuf, idx2d, pitch);
+            sdata[i] = *elem;
+        }
```
mehmetyusufoglu, line 63: [maybe] just to make it look shorter
```suggestion
          // Load the grid values from global memory to shared memory
        for (auto i = threadIdx1D; i < T_SharedMemSize1D; i += numThreadsPerBlock)
        {
            auto idx2D = alpaka::mapIdx<2>(alpaka::Vec<alpaka::DimInt<1u>, TIdx>(i), chunkSize + halo) + blockStartIdx;
            sharedData[i] = *getElementPtr(uCurrBuf, idx2D, pitch);
        }
```

example/heatEquation2D/src/BoundaryKernel.hpp
```diff
@@ -0,0 +1,98 @@
...
+        {
+            auto const globalIdx = blockStartIdx + alpaka::Vec<Dim, Idx>{1, chunkSize[1] + 1};
+
+            for(auto i = threadIdx1D; i < chunkSize[0]; i += numThreadsPerBlock)
+            {
+                auto idx2D = globalIdx + alpaka::Vec<Dim, Idx>(i, 0);
+                auto elem = getElementPtr(uBuf, idx2D, pitch);
+                *elem = exactSolution(idx2D[1] * dx, idx2D[0] * dy, step * dt);
+            }
+        }
```
mehmetyusufoglu: [ nit]  this lambda simplified the code a lot. i tried seems working.
```suggestion
             // Lambda function to apply boundary conditions
        auto applyBoundary = [&](auto const& globalIdxStart, auto const length, bool isRow) {
            for (auto i = threadIdx1D; i < length; i += numThreadsPerBlock)
            {
                alpaka::Vec<Dim, Idx> idx2D = globalIdxStart + (isRow ? alpaka::Vec<Dim, Idx>{0, i} : alpaka::Vec<Dim, Idx>{i, 0});
                auto elem = getElementPtr(uBuf, idx2D, pitch);
                *elem = exactSolution(idx2D[1] * dx, idx2D[0] * dy, step * dt);
            }
        };

               // Apply boundary conditions for the top row
        if (gridBlockIdx[0] == 0) {
            applyBoundary(blockStartIdx + alpaka::Vec<Dim, Idx>{0, 1}, chunkSize[1], true);
        }

               // Apply boundary conditions for the bottom row
        if (gridBlockIdx[0] == gridBlockExtent[0] - 1) {
            applyBoundary(blockStartIdx + alpaka::Vec<Dim, Idx>{chunkSize[0] + 1, 1}, chunkSize[1], true);
        }

               // Apply boundary conditions for the left column
        if (gridBlockIdx[1] == 0) {
            applyBoundary(blockStartIdx + alpaka::Vec<Dim, Idx>{1, 0}, chunkSize[0], false);
        }

               // Apply boundary conditions for the right column
        if (gridBlockIdx[1] == gridBlockExtent[1] - 1) {
            applyBoundary(blockStartIdx + alpaka::Vec<Dim, Idx>{1, chunkSize[1] + 1}, chunkSize[0], false);
        }
```

example/heatEquation2D/src/heatEquation2D.cpp
ikbuibui: No we aren't. This is a change which can be made across all alpaka examples.

example/heatEquation2D/src/heatEquation2D.cpp
ikbuibui: Renamed to computeQueue and dumpQueue based on how the queue tasks are distributed

example/heatEquation2D/src/StencilKernel.hpp
ikbuibui: Done

example/heatEquation2D/src/heatEquation2D.cpp
ikbuibui: changed

example/heatEquation2D/src/BoundaryKernel.hpp
psychocoderHPC: @ikbuibui Could you take this suggestion in and rebase and squash your code. 32 commits is a lot for this example.

ikbuibui: done

psychocoderHPC (approved):

PR #2368, opened 2024-08-21
Update to oneAPI 2024.2

fwyzard:


script/job_generator/versions.py
```diff
@@ -29,7 +29,7 @@
         "12.5",
     ],
     HIPCC: ["5.1", "5.2", "5.3", "5.4", "5.5", "5.6", "5.7", "6.0"],
-    ICPX: ["2024.0"],
+    ICPX: ["2024.2"],
```
fwyzard, line 32: @SimeonEhrig should we test only the latest version, or all of `2024.0`, `2024.1`, `2024.2` ?

I think testing and supporting only the latest one should be ok, but let me know what you think.

SimeonEhrig: Until there is no good reason, we should only test the latest release. It saves a lot of work. We do the same for CUDA on Windows, because there is no user uses an older version.

Maybe in the future, if we use OneAPI cluster, it is required to support older versions.

SimeonEhrig:
@fwyzard Can you please push again. Looks like the CI trigger didn't worked.

psychocoderHPC (approved):

psychocoderHPC:
@fwyzard please amend and force push this PR again, the CI is fixed

PR #2369, opened 2024-08-21
Implement device functions to simplify writing kernel code [v2]

fwyzard:
Implement device functions to simplify writing kernel code:
  - `oncePerGrid(acc)`;
  - `oncePerBlock(acc)`;
  - `uniformElements(acc, size)` and `uniformElements(acc, begin, end)`;
  - `uniformGroups(acc, size)` and `uniformGroupElements(acc, group, size)`;
  - `independentGroups(acc, groups)` and `independentGroupElements(acc, group_size)`;
  - variations of these functions along different dimensions.

Implement tests for the most common functions.

PR #2372, opened 2024-08-26
Kernel loops examples

fwyzard:
This PR uses the device functions introduced in #2369 to simplify some of the examples.

### DIFF
--- a/include/alpaka/exec/IndependentElements.hpp
+++ b/include/alpaka/exec/IndependentElements.hpp
@@ -4,6 +4,7 @@
 #include "alpaka/idx/Accessors.hpp"
 
 #include <algorithm>
+#include <ciso646> // workaround for MSVC in c++17 mode - TODO: remove once we move to c++20
 #include <cstddef>
 #include <type_traits>
 
--- a/example/vectorAdd/src/vectorAdd.cpp
+++ b/example/vectorAdd/src/vectorAdd.cpp
@@ -1,5 +1,5 @@
-/* Copyright 2023 Benjamin Worpitz, Matthias Werner, Bernhard Manfred Gruber, Jan Stephan, Luca Ferragina,
- *                Aurora Perego
+/* Copyright 2024 Benjamin Worpitz, Matthias Werner, Bernhard Manfred Gruber, Jan Stephan, Luca Ferragina,
+ *                Aurora Perego, Andrea Bocci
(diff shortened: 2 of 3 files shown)


include/alpaka/exec/IndependentElements.hpp
```diff
@@ -4,6 +4,7 @@
 #include "alpaka/idx/Accessors.hpp"
 
 #include <algorithm>
+#include <ciso646>
```
psychocoderHPC: I found that the inclusion of this header has no affect in C++ because C++ itself defines these operants.
@fwyzard Why do you added these header?

https://cplusplus.com/reference/ciso646/

psychocoderHPC (changes requested):

include/alpaka/exec/IndependentElements.hpp
fwyzard: Because MSVC is a stupid compiler and _does not_ define the keywords by default: https://learn.microsoft.com/en-us/cpp/build/reference/microsoft-extensions-to-c-and-cpp?view=msvc-170&redirectedfrom=MSDN#iso646h-not-enabled (and `/Ze` is the default).

fwyzard: OK, MSVC _does_ define the keywords in c++20 mode: https://godbolt.org/z/eMre9PMvT .

So we can remove them again once we switch to c++20.

SimeonEhrig: Please add the comment `// TODO: can be removed if C++ 20 is the minimum requirement`
This allows us to use `grep` for cleaning up the code, when we move to C++20.

SimeonEhrig (approved):

SimeonEhrig:
@psychocoderHPC Your change request blocks the merge.

fwyzard:
@SimeonEhrig are there some other tests that you would like to see migrated to the new loops ?

SimeonEhrig:
> @SimeonEhrig are there some other tests that you would like to see migrated to the new loops ?

I have to check the examples. But would be cool if we migrate a multi dimensional kernel.

psychocoderHPC (approved):

psychocoderHPC:
> @psychocoderHPC Your change request blocks the merge.

you should be able to dismiss my review.

PR #2374, opened 2024-08-28
refactoring of the SYCL backend

AuroraPerego:
This refactoring aims to reduce code duplication and ease the process of adding new SYCL backends (e.g. `GpuSyclNvidia` and `GpuSyclAmd`).

main changes:
- removed the inheritance for the SYCL `Acc`, changed to `using`
- templated everything on the `Tag`, device selectors included, and moved some common methods / structs in the `*GenericSycl.hpp` files

### DIFF
--- a/include/alpaka/acc/AccCpuSycl.hpp
+++ b/include/alpaka/acc/AccCpuSycl.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2023 Jan Stephan, Luca Ferragina, Andrea Bocci
+/* Copyright 2024 Jan Stephan, Luca Ferragina, Andrea Bocci, Aurora Perego
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -7,9 +7,6 @@
 #include "alpaka/acc/AccGenericSycl.hpp"
 #include "alpaka/acc/Tag.hpp"
 #include "alpaka/core/Sycl.hpp"
-
-#include <string>
-#include <utility>
 
 #if defined(ALPAKA_ACC_SYCL_ENABLED) && defined(ALPAKA_SYCL_ONEAPI_CPU)
 
--- a/include/alpaka/acc/AccCpuSycl.hpp
+++ b/include/alpaka/acc/AccCpuSycl.hpp
@@ -7,7 +7,6 @@
 #include "alpaka/acc/AccGenericSycl.hpp"
 #include "alpaka/acc/Tag.hpp"
 #include "alpaka/core/Sycl.hpp"
-#include "alpaka/platform/PlatformCpuSycl.hpp"
 
 #include <string>
 #include <utility>
@@ -20,7 +19,7 @@ namespace alpaka
     //!
     //! This accelerator allows parallel kernel execution on a oneAPI-capable CPU target device.
     template<typename TDim, typename TIdx>
-    using AccCpuSycl = AccGenericSycl<detail::SyclCpuSelector, TDim, TIdx>;
+    using AccCpuSycl = AccGenericSycl<TagCpuSycl, TDim, TIdx>;
 
     namespace trait
     {
--- a/include/alpaka/acc/AccCpuSycl.hpp
+++ b/include/alpaka/acc/AccCpuSycl.hpp
@@ -6,16 +6,8 @@
 
 #include "alpaka/acc/AccGenericSycl.hpp"
 #include "alpaka/acc/Tag.hpp"
-#include "alpaka/core/Concepts.hpp"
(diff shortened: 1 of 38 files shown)


psychocoderHPC:
@AuroraPerego could you please perform locally `git commit --amend  -a` and force push again to re-trigger the CI.
Simple restarting does not work, I am not sure if some caches avoid a clean re-test.

fwyzard:
At some point, could you also squash the last two commits into the third one ?

include/alpaka/acc/AccCpuSycl.hpp
fwyzard, line 1: @AuroraPerego I think you should update the copyright notices on this and the other files you modified.

include/alpaka/acc/AccCpuSycl.hpp
```diff
@@ -6,16 +6,7 @@
...
-#include "alpaka/dev/DevCpuSycl.hpp"
-#include "alpaka/dev/Traits.hpp"
-#include "alpaka/kernel/TaskKernelCpuSycl.hpp"
-#include "alpaka/kernel/Traits.hpp"
-#include "alpaka/platform/PlatformCpuSycl.hpp"
-#include "alpaka/platform/Traits.hpp"
-#include "alpaka/vec/Vec.hpp"
 
 #include <string>
 #include <utility>
```
fwyzard: I think these two headers can be removed (also in the other accelerator specialisations).
```suggestion
```

include/alpaka/platform/PlatformCpuSycl.hpp
```diff
@@ -18,27 +18,20 @@ namespace alpaka
...
-        struct SyclCpuSelector
+        template<>
+        struct SYCLDeviceSelector<TagCpuSycl>
         {
             auto operator()(sycl::device const& dev) const -> int
             {
                 return dev.is_cpu() ? 1 : -1;
             }
+
+            static constexpr char name[] = "CpuSycl";
```
fwyzard: If this is only used in the accelerator specialisations, should it be moved there ?
Or, if it is used anywhere else, maybe it should be moved to the platform classes ?

include/alpaka/mem/buf/BufGenericSycl.hpp
```diff
@@ -197,38 +195,40 @@ namespace alpaka::trait
...
+            DevCpu const& host,
+            PlatformGenericSycl<TTag> const& platform,
+            TExtent const& extent) -> BufCpu<TElem, TDim, TIdx>
         {
             ALPAKA_DEBUG_MINIMAL_LOG_SCOPE;
 
-            // Allocate SYCL page-locked memory on the host, mapped into the TPlatform address space and
-            // accessible to all devices in the TPlatform.
+            // Allocate SYCL page-locked memory on the host, mapped into the PlatformGenericSycl address space and
+            // accessible to all devices in the PlatformGenericSycl.
```
fwyzard: I would write more generically "platform" instead of "PlatformGenericSycl":
```suggestion
            // Allocate SYCL page-locked memory on the host, mapped into the SYCL platform's address space and
            // accessible to all devices in the SYCL platform.
```

include/alpaka/platform/PlatformCpuSycl.hpp
AuroraPerego: It's used only [here](https://github.com/AuroraPerego/alpaka/blob/SYCLrefactoring/include/alpaka/acc/AccGenericSycl.hpp#L167), so I can try to move it.

fwyzard (changes requested):
See the other comments.

include/alpaka/acc/AccGenericSycl.hpp
```diff
@@ -164,8 +164,8 @@ namespace alpaka::trait
     {
         static auto getAccName() -> std::string
         {
-            return std::string("Acc") + detail::SYCLDeviceSelector<TTag>::name + "<" + std::to_string(TDim::value)
-                   + "," + core::demangled<TIdx> + ">";
+            return std::string("Acc") + core::demangled<TTag>.substr(11) + "<" + std::to_string(TDim::value) + ","
```
fwyzard: 😭

fwyzard:
```suggestion
            return std::string("Acc") + core::demangled<TTag>.substr(__builtin_strlen("alpaka::Tag")) + "<" + std::to_string(TDim::value) + ","
```
at least is self-documenting :-p

fwyzard (approved):
Note to whoever merges this PR: the commits should be squashed into a single one.

PR #2375, opened 2024-08-28
enable support for mapped memory allocation for the SYCL backend

AuroraPerego:
The `allocMappedBuf` method was already there, but when calling `allocMappedBufIfSupported` it was not used because the trait `hasMappedBufSupport` has never been specialized to be `true`.
This led to spotting a bug in `Buffer.hpp` used in `mathTest`, where the accelerator platform was used to allocate memory on the host (never seen before because CUDA/HIP and the CPU backend don't use the `Platform` to do the allocation).

### DIFF
--- a/test/unit/math/src/Buffer.hpp
+++ b/test/unit/math/src/Buffer.hpp
@@ -38,12 +38,10 @@ namespace mathtest
         using PlatformAcc = alpaka::Platform<DevAcc>;
         using BufAcc = alpaka::Buf<DevAcc, TData, Dim, Idx>;
 
-        PlatformHost platformHost;
         DevHost devHost;
 
         BufHost hostBuffer;
         BufAcc devBuffer;
-        PlatformAcc platformAcc;
 
         // Native pointer to access buffer.
         TData* const pHostBuffer;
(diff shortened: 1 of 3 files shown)


test/unit/math/src/Buffer.hpp
```diff
@@ -56,7 +56,7 @@ namespace mathtest
         // Constructor needs to initialize all Buffer.
         Buffer(DevAcc const& devAcc)
             : devHost{alpaka::getDevByIdx(platformHost, 0)}
-            , hostBuffer{alpaka::allocMappedBufIfSupported<TData, Idx>(devHost, platformAcc, Tcapacity)}
+            , hostBuffer{alpaka::allocMappedBufIfSupported<TData, Idx>(devHost, platformHost, Tcapacity)}
```
fwyzard: why this change ?

fwyzard: `allocMappedBufIfSupported` is supposed to allocate host memory, which is visible by the devices on the given platform if the mapped allocation is supported.

Calling `allocMappedBufIfSupported` with the host platform will allocate standard host memory.

AuroraPerego: The code crashed during this allocation and I thought that was the reason, looking into it a bit more I've located the crash in the line `auto ctx = platform.syclContext();` (strange). I'll go on with the investigation of the crash and try to fix it.

AuroraPerego: `platformAcc` was not initialized correctly, I've passed it to the constructor of `Buffer`. I've done the same for `platformHost` to be safe.

psychocoderHPC:
@AuroraPerego could you please perform locally `git commit --amend  -a` and force push again to re-trigger the CI.
Simple restarting does not work, I am not sure if some caches avoid a clean re-test.

fwyzard (approved):

PR #2376, opened 2024-08-29
Mark the deduction guide for vector as ALPAKA_FN_HOST_ACC

ikbuibui:
This is a fix for some compilers which cant otherwise deduce the type on device
PR also changes an example to show usage of this deduction on device code

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -592,7 +592,8 @@ namespace alpaka
     };
 
     template<typename TFirstIndex, typename... TRestIndices>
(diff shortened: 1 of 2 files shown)


include/alpaka/vec/Vec.hpp
```diff
@@ -592,7 +592,8 @@ namespace alpaka
     };
 
     template<typename TFirstIndex, typename... TRestIndices>
-    Vec(TFirstIndex&&, TRestIndices&&...) -> Vec<DimInt<1 + sizeof...(TRestIndices)>, std::decay_t<TFirstIndex>>;
+    ALPAKA_FN_HOST_ACC Vec(TFirstIndex&&, TRestIndices&&...)
```
fwyzard, line 595: It should also be marked as `constexpr`, but IIUC that requires c++20.

fwyzard (approved):

PR #2381, opened 2024-09-12
Add support for new backends and library versions

fwyzard:
Add support for:
  - CUDA 12.6
  - ROCm 6.1, 6.2
  - CMake 3.27-3.30
  - Boost 1.83-1.86


fwyzard:
Note that HIP/ROCm 6.2 with `-O0` has a segmentation fault in the compiler while building one of the tests.
I've tried to disable this combination in the CI, but I don't know if we should document it anywhere ?

fwyzard:
@SimeonEhrig I'm worried that these changes may substantially increase the workload on the CI ?

psychocoderHPC (approved):

PR #2383, opened 2024-09-16
reduce Uniform/Independent-Elements iterator register footprint

### LINKED ISSUE #2382: reduce register footprint for the new iterators
https://github.com/alpaka-group/alpaka/pull/2377#issuecomment-2348309816 showed that the register footprint increased strongly when switching examples to the new iteration schema.
The reason is that the iterator state is huge and increases the register footprint.

This issue should evaluate possible optimizations to solve this issue.

psychocoderHPC:
reduce register footprint

fix #2382

Rewrite Uniform/Independent-Elements iterator to reduce the register footprint.

- avoid multiple returns within a function
- reduce the iterator state size by one element

Change bufferCopy example, use the element layer for CPU accelerators.

The original code before using the UniformElements in the bufferCopy example required 38 and 56 registers.
With this PR we requires 64 and 71 registers but this can not be compared to the old develop branch because the old implementation ignores the element layer.
Compared to the current development branch where [80 and 85 registers](https://github.com/alpaka-group/alpaka/pull/2377#issuecomment-2348309816) are required this refactoring is a huge improvement and will increase the occupancy in real word usage.

The increment of the iterator have now always the cost of two integral value increments and two comparisons where the comparisons will be transformed into [comparison and selection instruction](https://docs.nvidia.com/cuda/parallel-thread-execution/index.html#comparison-and-selection-instructions) on GPU devices therefore there is no branching overhead.

```
ptxas info    : 219055 bytes gmem, 72 bytes cmem[3]
ptxas info    : Compiling entry function '_ZN6alpaka6detail9gpuKernelI17PrintBufferKernelNS_9ApiCudaRtENS_22AccGpuUniformCudaHipRtIS3_St17integral_constantImLm3EEmEES6_mJNSt12experimental6mdspanIjNS8_7extentsImJLm18446744073709551615ELm18446744073709551615ELm18446744073709551615EEEENS8_13layout_strideENS_12experimental6traits6detail19ByteIndexedAccessorIjEEEEEEEvNS_3VecIT2_T3_EET_DpT4_' for 'sm_52'
ptxas info    : Function properties for _ZN6alpaka6detail9gpuKernelI17PrintBufferKernelNS_9ApiCudaRtENS_22AccGpuUniformCudaHipRtIS3_St17integral_constantImLm3EEmEES6_mJNSt12experimental6mdspanIjNS8_7extentsImJLm18446744073709551615ELm18446744073709551615ELm18446744073709551615EEEENS8_13layout_strideENS_12experimental6traits6detail19ByteIndexedAccessorIjEEEEEEEvNS_3VecIT2_T3_EET_DpT4_
    32 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 64 registers, 408 bytes cmem[0]
ptxas info    : Compiling entry function '_ZN6alpaka6detail9gpuKernelI16TestBufferKernelNS_9ApiCudaRtENS_22AccGpuUniformCudaHipRtIS3_St17integral_constantImLm3EEmEES6_mJNSt12experimental6mdspanIjNS8_7extentsImJLm18446744073709551615ELm18446744073709551615ELm18446744073709551615EEEENS8_13layout_strideENS_12experimental6traits6detail19ByteIndexedAccessorIjEEEEEEEvNS_3VecIT2_T3_EET_DpT4_' for 'sm_52'
ptxas info    : Function properties for _ZN6alpaka6detail9gpuKernelI16TestBufferKernelNS_9ApiCudaRtENS_22AccGpuUniformCudaHipRtIS3_St17integral_constantImLm3EEmEES6_mJNSt12experimental6mdspanIjNS8_7extentsImJLm18446744073709551615ELm18446744073709551615ELm18446744073709551615EEEENS8_13layout_strideENS_12experimental6traits6detail19ByteIndexedAccessorIjEEEEEEEvNS_3VecIT2_T3_EET_DpT4_
    0 bytes stack frame, 0 bytes spill stores, 0 bytes spill loads
ptxas info    : Used 71 registers, 408 bytes cmem[0]
```

### DIFF
--- a/include/alpaka/exec/UniformElements.hpp
+++ b/include/alpaka/exec/UniformElements.hpp
@@ -130,11 +130,12 @@ namespace alpaka
 
                 ALPAKA_FN_ACC inline const_iterator(Idx elements, Idx stride, Idx extent, Idx first)
                     : elements_{elements}
-                    , stride_{stride}
+                    ,
+                    // we need to reduce the stride by on element range because index_ is later increased with each
+                    // increment
+                    stride_{stride - elements}
                     , extent_{extent}
-                    , first_{std::min(first, extent)}
-                    , index_{first_}
-                    , range_{std::min(first + elements, extent)}
+                    , index_{std::min(first, extent)}
                 {
                 }
 
@@ -148,21 +149,16 @@ namespace alpaka
                 ALPAKA_FN_ACC inline const_iterator& operator++()
                 {
                     // increment the index along the elements processed by the current thread
+                    ++indexElem_;
                     ++index_;
-                    if(index_ < range_)
-                        return *this;
+                    if(indexElem_ >= elements_)
+                    {
+                        indexElem_ = Idx{0};
+                        index_ += stride_;
+                    }
+                    if(index_ >= extent_)
+                        index_ = extent_;
 
-                    // increment the thread index with the grid stride
-                    first_ += stride_;
-                    index_ = first_;
-                    range_ = std::min(first_ + elements_, extent_);
-                    if(index_ < extent_)
-                        return *this;
-
-                    // the iterator has reached or passed the end of the extent, clamp it to the extent
-                    first_ = extent_;
-                    index_ = extent_;
-                    range_ = extent_;
                     return *this;
                 }
 
@@ -176,7 +172,7 @@ namespace alpaka
 
                 ALPAKA_FN_ACC inline bool operator==(const_iterator const& other) const
                 {
-                    return (index_ == other.index_) and (first_ == other.first_);
+                    return (*(*this) == *other);
                 }
 
                 ALPAKA_FN_ACC inline bool operator!=(const_iterator const& other) const
@@ -190,9 +186,8 @@ namespace alpaka
                 Idx stride_;
                 Idx extent_;
                 // modified by the pre/post-increment operator
-                Idx first_;
                 Idx index_;
-                Idx range_;
+                Idx indexElem_ = {0};
             };
 
         private:
--- a/include/alpaka/exec/UniformElements.hpp
+++ b/include/alpaka/exec/UniformElements.hpp
@@ -153,7 +153,7 @@ namespace alpaka
                     ++index_;
                     if(indexElem_ >= elements_)
                     {
-                        indexElem_ = Idx{0};
+                        indexElem_ = 0;
                         index_ += stride_;
                     }
                     if(index_ >= extent_)
@@ -187,7 +187,7 @@ namespace alpaka
                 Idx extent_;
                 // modified by the pre/post-increment operator
                 Idx index_;
-                Idx indexElem_ = {0};
+                Idx indexElem_ = 0;
             };
 
         private:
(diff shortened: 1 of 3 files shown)


psychocoderHPC:
IndependentElements iterator register footprint in the test goes down from 28 registers to 22.

psychocoderHPC:
@fwyzard the implementation for `const_iterator` in IndependentElements and UniformElements is 100% identical, should we in a follow up PR provide the implementation once and avoid code duplication?

psychocoderHPC:
@fwyzard CI passed, ready for review

include/alpaka/exec/UniformElements.hpp
```diff
@@ -190,9 +186,8 @@ namespace alpaka
                 Idx stride_;
                 Idx extent_;
                 // modified by the pre/post-increment operator
-                Idx first_;
                 Idx index_;
-                Idx range_;
+                Idx indexElem_ = {0};
```
fwyzard:
```suggestion
                Idx indexElem_ = 0;
```

include/alpaka/exec/UniformElements.hpp
```diff
@@ -148,21 +149,16 @@ namespace alpaka
...
-
-                    // increment the thread index with the grid stride
-                    first_ += stride_;
-                    index_ = first_;
-                    range_ = std::min(first_ + elements_, extent_);
-                    if(index_ < extent_)
-                        return *this;
+                    if(indexElem_ >= elements_)
+                    {
+                        indexElem_ = Idx{0};
```
fwyzard:
```suggestion
                        indexElem_ = 0;
```

fwyzard (commented):
The logic of `UniformElements` looks correct.

Does the code work if you use `0` instead of `Idx{0}`, or does the compiler uselessly complain ?

fwyzard:
What do you think would be the impact on the register usage and performance if the iterator kept a pointer to the `UniformElementsAlong` object instead of storing `elements_`, `stride_` and `extent_` ?

fwyzard:
> @fwyzard the implementation for `const_iterator` in IndependentElements and UniformElements is 100% identical, should we in a follow up PR provide the implementation once and avoid code duplication?

Sure.

Currently they need to be `friend` of their respective classes, but maybe that can be passed as a template argument ?

psychocoderHPC:
> What do you think would be the impact on the register usage and performance if the iterator kept a pointer to the `UniformElementsAlong` object instead of storing `elements_`, `stride_` and `extent_` ?

This will most likely not help and in case you create the iterator within a function it is very easy to produce illegal memory access because the main object is already destroyed.
Never the less coping a constant value is not the problem, the compile is optimizing this out. I also played with keeping the ACC in the iterator as const reference (this is safe from alpaka side) and query the number of elements each time because they are stored in the accelerator. This is not decreasing the memory footprint.

include/alpaka/exec/UniformElements.hpp
psychocoderHPC: I changed it and I hope that there will be no sign warning.
Local test not throw warning, let's see what the CI is saying.

fwyzard:
> This will most likely not help 

It doesn't.

> and in case you create the iterator within a function it is very easy to produce illegal memory access because the main object is already destroyed.

Well, using an iterator after the container it iterates over has been destroyed is definitely a programming error - the same applies to all containers, like a `vector`, _etc_.

`UniformElementsND` already uses this approach to avoid copying a large number of dimensions.

fwyzard (approved):

psychocoderHPC:
ready to be merged

PR #2385, opened 2024-09-17
Explicitly call alpaka::detail to achieve SYCL compilation

MichaelVarvarin:


psychocoderHPC:
This bugfix is required because we have a detail namespace in `alpaka::trait` and `alpaka`. 
All usages from within the namespace `alpaka::trait` will look into `alpaka::trait::detail` if we do a lookup to e.g. `detail::Foo`

psychocoderHPC (approved):

psychocoderHPC (approved):

PR #2387, opened 2024-09-18
support Clang 18 and 19

SimeonEhrig:
hopefully there is nothing to do because @fwyzard already fixed several Clang 18 and 19 problems during PR #2368

PR #2388, opened 2024-09-18
implement markdown generator for compiler backend compatibility table

SimeonEhrig:
The Python script generates the Markdown table for compiler back-end compatibility from a json config.

Solves the problem, that at least a 4K display is required to change the table :-(
Can be combined with the scripts of `job_generator.py` later to automatically generates some of the columns.

~~Should be merged after #2387 to solve the merge conflict.~~

Removed Clang 18 and 19. Looks like #2387 requires some work. Therefore merge this PR first.


SimeonEhrig:
Stopped the CI because it does not test the script.

fwyzard (approved):

fwyzard:
@SimeonEhrig is this ready to go ?

SimeonEhrig:
Yes. The generator is ready and I replaced the table with the generated version. The content is the same like before. Only the white space and so matches now.

psychocoderHPC (approved):

PR #2391, opened 2024-09-23
Single thread acc throw for invalid workdiv fix

ikbuibui:
For a single thread acc if the WorkDiv number of threads per block is more than one, throw a runtime error
We will now throw in create task kernel. The current approach of throwing when the task kernel is executed doesn't work. ``CallbackThread::startWorkerThread``, tries to handle the  exception, but it needs to be dealt with the thread that creates it.

### DIFF
--- a/include/alpaka/acc/AccCpuOmp2Blocks.hpp
+++ b/include/alpaka/acc/AccCpuOmp2Blocks.hpp
@@ -189,6 +189,13 @@ namespace alpaka
                 TKernelFnObj const& kernelFnObj,
                 TArgs&&... args)
             {
+                if(workDiv.m_blockThreadExtent.prod() != static_cast<TIdx>(1u))
+                {
+                    throw std::runtime_error(
+                        "The given work division is not valid for a single thread Acc: "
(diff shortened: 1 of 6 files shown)


fwyzard (dismissed):

fwyzard:
LGTM

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -189,6 +189,13 @@ namespace alpaka
                 TKernelFnObj const& kernelFnObj,
                 TArgs&&... args)
             {
+                if(workDiv.m_blockThreadExtent.prod() != static_cast<TIdx>(1u))
```
psychocoderHPC, line 192: please remove the checks in the `operator()` no need to keep the old checks if we can not catch the exception.

psychocoderHPC (changes requested):

include/alpaka/acc/AccCpuOmp2Blocks.hpp
ikbuibui: Fixed

psychocoderHPC (approved):

PR #2267, opened 2024-05-06
implement AccIsEnabled

SimeonEhrig:
The value of AccIsEnabled indicates whether an accelerator is enabled for a specific tag.

PR #2392, opened 2024-09-23
fix cray clang compiler errors

psychocoderHPC:
- fix cray clang compile issue
  ```
  include/alpaka/acc/TagAccIsEnabled.hpp:34:66: error: template template argument has different template parameters than its corresponding template template parameter
   34 |     using EnabledAccTags = alpaka::meta::Filter<AccTags, alpaka::AccIsEnabled>;
      |    
  ```
- fix cray clang OpenMP linker issues, missing flag `-fopenmp`
- remove clang workaround we applied in https://github.com/alpaka-group/alpaka/pull/2267

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -194,13 +194,6 @@ else()
         endif()
     endif()
 
-    # C++17 relaxed template template argument matching is disabled by default until Clang 19
-    # https://github.com/llvm/llvm-project/commit/b86e0992bfa6
-    # https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#150
-    # for example, is required to create alpaka::EnabledAccTags
-    # the feature is implemented since Clang 4
-    alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<AND:$<CXX_COMPILER_ID:Clang,AppleClang,IntelLLVM>>:SHELL:-frelaxed-template-template-args>")
-
     # Add debug optimization levels. CMake doesn't do this by default.
     # Note that -Og is the recommended gcc optimization level for debug mode but is equivalent to -O1 for clang (and its derivates).
     alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<AND:$<CONFIG:Debug>,$<CXX_COMPILER_ID:GNU>,$<COMPILE_LANGUAGE:CXX>>:SHELL:-Og>"
@@ -610,13 +603,6 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
             alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<COMPILE_LANGUAGE:CXX>:-D__HIP_PLATFORM_HCC__>")
         endif()
 
-        # C++17 relaxed template template argument matching is disabled by default until Clang 19
-        # https://github.com/llvm/llvm-project/commit/b86e0992bfa6
-        # https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#150
-        # for example, is required to create alpaka::EnabledAccTags
-        # TODO(SimeonEhrig): restict HIP version, if first HIP version is release using Clang 19
-        alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<COMPILE_LANGUAGE:HIP>:SHELL:-frelaxed-template-template-args>")
-
         alpaka_compiler_option(HIP_KEEP_FILES "Keep all intermediate files that are generated during internal compilation steps 'CMakeFiles/<targetname>.dir'" OFF)
         if(alpaka_HIP_KEEP_FILES)
             alpaka_set_compiler_options(HOST_DEVICE target alpaka "$<$<COMPILE_LANGUAGE:HIP>:SHELL:-save-temps>")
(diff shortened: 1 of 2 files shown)


cmake/alpakaCommon.cmake
```diff
@@ -357,6 +357,12 @@ if(alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE OR alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE)
     else()
         find_package(OpenMP REQUIRED COMPONENTS CXX)
         target_link_libraries(alpaka INTERFACE OpenMP::OpenMP_CXX)
+        # shown with CMake 3.29 and cray clang 17
+        # workaround: cmake is missing to add '-fopenmp' to the linker flags
+        if(CMAKE_CXX_COMPILER_ID STREQUAL "CrayClang")
+            target_link_libraries(alpaka INTERFACE -fopenmp)
+        endif()
+        find_package(OpenMP COMPONENTS CXX)
```
ikbuibui:
```suggestion

```
find_package is already on line#358

SimeonEhrig:
Adding the Cray Compiler Compiler ID here in the CMake should solve the problem:

https://github.com/alpaka-group/alpaka/blob/5c5a690f48aeedc6b8651e5b5f486767c12f2be8/cmake/alpakaCommon.cmake#L202

I removed this line from my CMake, the `Filter.hpp` does not compile.

```bash
$ cmake --build -j -t kernelSpecialization
Building CXX object example/kernelSpecialization/CMakeFiles/kernelSpecialization.dir/src/kernelSpecialization.cpp.o
In file included from /home/simeon/projects/alpaka/example/kernelSpecialization/src/kernelSpecialization.cpp:5:
In file included from /home/simeon/projects/alpaka/include/alpaka/alpaka.hpp:26:
/home/simeon/projects/alpaka/include/alpaka/acc/TagAccIsEnabled.hpp:34:66: error: template template argument has different template parameters than its corresponding template template parameter
   34 |     using EnabledAccTags = alpaka::meta::Filter<AccTags, alpaka::AccIsEnabled>;
      |                                                                  ^
/home/simeon/projects/alpaka/include/alpaka/acc/TagAccIsEnabled.hpp:23:5: note: too many template parameters in template template argument
   23 |     template<typename TTag, typename = void>
      |     ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/home/simeon/projects/alpaka/include/alpaka/meta/Filter.hpp:42:30: note: previous template template parameter is here
   42 |     template<typename TList, template<typename> class TPred>
      |                              ^~~~~~~~~~~~~~~~~~
In file included from /home/simeon/projects/alpaka/example/kernelSpecialization/src/kernelSpecialization.cpp:6:
/home/simeon/projects/alpaka/include/alpaka/example/ExecuteForEachAccTag.hpp:25:97: error: no member named 'EnabledAccTags' in namespace 'alpaka'
   25 |         return std::apply([=](auto const&... tags) { return (callable(tags) || ...); }, alpaka::EnabledAccTags{});
      |                                                                                         ~~~~~~~~^
/home/simeon/projects/alpaka/example/kernelSpecialization/src/kernelSpecialization.cpp:109:12: error: cannot initialize return object of type 'int' with an rvalue of type 'void'
  109 |     return alpaka::executeForEachAccTag([=](auto const& tag) { return example(tag); });
      |            ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
3 errors generated.
gmake[3]: *** [example/kernelSpecialization/CMakeFiles/kernelSpecialization.dir/build.make:76: example/kernelSpecialization/CMakeFiles/kernelSpecialization.dir/src/kernelSpecialization.cpp.o] Error 1
gmake[2]: *** [CMakeFiles/Makefile2:2248: example/kernelSpecialization/CMakeFiles/kernelSpecialization.dir/all] Error 2
gmake[1]: *** [CMakeFiles/Makefile2:2255: example/kernelSpecialization/CMakeFiles/kernelSpecialization.dir/rule] Error 2
gmake: *** [Makefile:673: kernelSpecialization] Error 2
```

If I apply your changes, it compiles.

cmake/alpakaCommon.cmake
psychocoderHPC: fixed

psychocoderHPC:
@SimeonEhrig I removed the cmake workaround.
The advantage is that stand alone header does not need additional flags and CMS does not need to handle this case in there build suite too.

fwyzard:
I have no idea what these changes are about :-p

SimeonEhrig:
> I have no idea what these changes are about :-p

Up to Clang 19, the compiler does not support relaxed template template args by default, which is a C++17 feature: https://www.open-std.org/jtc1/sc22/wg21/docs/cwg_defects.html#150
Therefore all Clang based compilers needs to set the compiler flag `-frelaxed-template-template-arg` to enable the feature. I forgot to add the Cray Clang compiler ID to the CMake. Therefore the compiler cannot compile the code.

René's code change simply avoids to use the feature. Therefore the flag does not need to be set anymore.

fwyzard:
Thanks for the explanation. I let you approve the PR :-)

SimeonEhrig (approved):

PR #2399, opened 2024-09-27
change documentation for Buf and mdspan

mehmetyusufoglu:
Simple additions to cheetsheet
- to solve an ambiguity in `alpaka::Buf` part
- to show setting and getting items of `mdspan`

### DIFF
--- a/docs/source/basic/cheatsheet.rst
+++ b/docs/source/basic/cheatsheet.rst
@@ -1,6 +1,10 @@
 Cheatsheet
 ==========
 
+.. raw:: pdf
+
+   Spacer 0,1
+
 .. only:: html
 
    Download pdf version :download:`here <../../cheatsheet/cheatsheet.pdf>`
@@ -17,10 +21,6 @@ General
 
      #include <alpaka/alpaka.hpp>
      using namespace alpaka;
-
-.. raw:: pdf
-
-   Spacer 0,5
 
 Accelerator, Platform and Device
 --------------------------------
@@ -61,8 +61,7 @@ Create a queue for a device
   Property:
      .. code-block:: c++
 
-	Blocking
-	NonBlocking
+	Blocking, NonBlocking
 
 Put a task for execution
   .. code-block:: c++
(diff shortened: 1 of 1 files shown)


docs/source/basic/cheatsheet.rst
```diff
@@ -109,7 +109,11 @@ Allocate a buffer in host memory
   .. code-block:: c++
 
      // Use alpaka vector as a static array for the extents
-     Vec<Dim, Idx> extent = value;
+     // If Dim is DimInt<1>;
+     Vec<Dim, Idx> extent = scalarIdxValue;
```
fwyzard: to me this would be more clear as
```suggestion
     Vec<Dim, Idx> extent = value;
```
or
```suggestion
     Vec<DimInt<1>, Idx> extent = value;
```

docs/source/basic/cheatsheet.rst
```diff
@@ -109,7 +109,11 @@ Allocate a buffer in host memory
...
+     // If Dim is DimInt<2>;
+     Vec<Dim, Idx> extent = {scalarIdxValue1,scalarIdxValue2};
```
fwyzard: to me this would be more clear as
```suggestion
     Vec<Dim, Idx> extent = {value1, value2};
```
or
```suggestion
     Vec<DimInt<2>, Idx> extent = {value1, value2};
```

mehmetyusufoglu:
@fwyzard  your suggestions (first options) are added, thanks.

psychocoderHPC:
I will try to reduce the cheatsheet changes that our cheatsheet is fitting to two pdf pages again

psychocoderHPC:
we should add more content to make the cheatsheet three pages or we should take care to keep two pdf sides. The printed version looks bad in case we have only a few lines on an empty page

mehmetyusufoglu:
> we should add more content to make the cheatsheet three pages or we should take care to keep two pdf sides. The printed version looks bad in case we have only a few lines on an empty page

I see, thanks. I think cheat-sheet doesn't have to be one sheet, it is used online by people as a shorthand and concise description.

PR #2403, opened 2024-09-30
implement alpaka::concepts::Tag

SimeonEhrig:
First alpaka C++20 concept :-) I think there are several points to discuss.

- [x] merge PR #2402 before

Implement `alpaka::concept::Tag` which checks the requirements for an alpaka tag.
Remove support Clang 10, 11 and 12 because they are not fully supports C++20 concepts.


include/alpaka/acc/Tag.hpp
```diff
@@ -5,12 +5,18 @@
...
 #include <type_traits>
 
+namespace alpaka
+{
+    struct ConceptTag;
+} // namespace alpaka
+
 #define CREATE_ACC_TAG(tag_name)                                                                                      \
-    struct tag_name                                                                                                   \
+    struct tag_name : public concepts::Implements<alpaka::ConceptTag, tag_name>                                       \
```
SimeonEhrig: I need to introduce the `ConceptTag` to make the class unique. Otherwise the only property to identify a alpaka tag is the helper static class function `get_name`, which is not unique.

Maybe we should think about to rename `alpaka::concepts` because it confuses now, than we have C++20 concepts.

SimeonEhrig:
@alpaka-group/alpaka-maintainers I was thinking about naming concepts. Here it is easy because `alpaka::Tag` is unique and not used yet. Maybe the name looks like a base class where the actual tags inherintant, but I think this fine.

I the case of `alpaka::Queue` this is different. We have already the class `alpaka::Queue`:
https://github.com/alpaka-group/alpaka/blob/e20236d86be9a8093fa2efae44eaf2db79dc38db/example/helloWorld/src/helloWorld.cpp#L74

In the case of template class deduction this can confuse:
```c++
// use the class alpaka::Queue, deducts the argument from the function call
alpaka::Queue queue = create_queue();

// use the concept alpaka::Queue
alpaka::Queue auto queue2 = create_queue();
``` 
The C++ standard makes clear, where classes and where concepts are used. But without the context, like in the documentation, it is hard to distinguish.

Possible solutions:
- let it as it
- add a prefix like for templates: `alpaka::CTag, alpaka::CAcc, alpaka::CQueue`
- used the alpaka namespace `alpaka::concept::Tag, alpaka::concept::Acc, alpaka::concept::Queue`

What do you think?

fwyzard:
What is the syntax for using concepts ?

SimeonEhrig:
> What is the syntax for using concepts ?

Have a look in this commit: https://github.com/alpaka-group/alpaka/pull/2403/commits/37c62ed8b2e7f977da7025e2fda6e12cd8466e5e

Also interesting is, what is the error message. Try to call the functions `specialize_tag<NoTag1>()`  and `specialize_tag<NoTag2>()` in the `AccTagTest.cpp` and see what the compiler is telling you.

SimeonEhrig:
You can also write something like this:

```c++
template<typename T>
   requires std::integral<T>
T gcd(T a, T b) {
   if( b == 0 ) return a;
   else return gcd(b, a % b);
}
```
But I prefer the version, which I used because it uses less boilerplate code and looks more intuitive.

include/alpaka/acc/Tag.hpp
SimeonEhrig: I checked what `alpaka::concepts::Implements` is actual doing and if we can replace it with C++20 concepts.
The answer is no, because it implements a type hierarchy like in Julia with type traits. It is possible to do it with C++ concepts but nothing which is already provided by the STL.

Therefore I suggest to rename `alpaka::concepts` to `alpaka::type_hierarchy` and also the corresponding function and structs to avoid confusions with C++20 concepts.

SimeonEhrig: In an offline discussion with @fwyzard he suggested `alpaka::interfaces`. I like the idea, because actual this mechanism imitates traits from other languages but traits has already a meaning in alpaka. Therefore interface is a good solution because it sounds like java interfaces which is pretty similar.

psychocoderHPC:
> Possible solutions:
> 
> * let it as it
> * add a prefix like for templates: `alpaka::CTag, alpaka::CAcc, alpaka::CQueue`
> * used the alpaka namespace `alpaka::concept::Tag, alpaka::concept::Acc, alpaka::concept::Queue`
> 
> What do you think?

I am a friend of namespaces and not mangling all into names. `C*` is also very often used for classes. This is what I learned long time ago. Namespaces will clearly separate implementations from concepts.

fwyzard:
> You can also write something like this:
> 
> ```c++
> template<typename T>
>    requires std::integral<T>
> ```

No, I agree that 
```c++
template<std::integral T>
```

looks cleaner.

fwyzard:
> used the alpaka namespace `alpaka::concept::Tag, alpaka::concept::Acc, alpaka::concept::Queue`

👍🏻

fwyzard:
To avoid confusion, we could rename the current namespace and classes called "Concept" to "Interface" ?

SimeonEhrig:
> > You can also write something like this:
> > ```c++
> > template<typename T>
> >    requires std::integral<T>
> > ```
> 
> No, I agree that
> 
> ```c++
> template<std::integral T>
> ```
> 
> looks cleaner.

I agree with you. And I was thinking about, when you need this `requires` in general. The answer is, if you have more than one template argument in the concept.

```c++
template<typename T, typename U>
concept foo = requires { requires std::is_same_v<T,U>; };

template<typename T, typename U>
void bar() requires foo<T,U> {}
```

But if we can do it, we should use the first syntax.

SimeonEhrig:
> To avoid confusion, we could rename the current namespace and classes called "Concept" to "Interface" ?

I agree with the idea. So let's rename the current `alpaka::concepts` to `alpaka::inteface` and reuse the namespace `alpaka::concepts` for C++20 concepts.

By the way. All our namespaces are singular. But we cannot use `alpaka::concept` because `concept` is a key word.

include/alpaka/acc/TagAccIsEnabled.hpp
```diff
@@ -20,12 +21,12 @@ namespace alpaka
 {
     //! \brief check if the accelerator is enabled for a given tag
     //! \tparam TTag alpaka tag type
-    template<typename TTag, typename = void>
+    template<alpaka::concepts::Tag TTag, typename = void>
```
psychocoderHPC: IMO `alpaka::` can be removed because we are already in the namespace alpaka.

psychocoderHPC (changes requested):
`alpaka::concepts::Tag` should be `concepts::Tag`. 
Except in the tests if we are not in the alpaka namespace.

psychocoderHPC (approved):

Issue #2405, opened 2024-10-01
remove boost demangle


psychocoderHPC:
https://github.com/alpaka-group/alpaka/blob/e20236d86be9a8093fa2efae44eaf2db79dc38db/include/alpaka/core/DemangleTypeNames.hpp#L18-L19

We aim to remove boost completely. With clang 15+ and gcc 11+ we can remove this code with the following code.
These versions support `std::source_location`.

```C++
#include <source_location>
#include <string>
#include <string_view>
    /// \file
    /// use source_location to derive the demangled type name
    /// based on:
    /// https://www.reddit.com/r/cpp/comments/lfi6jt/finally_a_possibly_portable_way_to_convert_types/?utm_source=share&utm_medium=web3x&utm_name=web3xcss&utm_term=1&utm_content=share_button

    struct DummyType
    {
    };

    template<typename T>
    inline auto EmbedTypeIntoSignature()
    {
        return std::string_view{std::source_location::current().function_name()};
    }

    template<typename T>
    struct Demangled
    {
        static auto name()
        {
            constexpr size_t testSignatureLength = sizeof("DummyType") - 1;
            auto const DummySignature = EmbedTypeIntoSignature<DummyType>();
            // count char's until the type name starts
            auto const startPosition = DummySignature.find("DummyType");
            // count char's after the type information by removing type name information and pre information
            auto const tailLength = DummySignature.size() - startPosition - testSignatureLength;
            auto const EmbeddingSignature = EmbedTypeIntoSignature<T>();
            auto const typeLength = EmbeddingSignature.size() - startPosition - tailLength;
            return EmbeddingSignature.substr(startPosition, typeLength);
        }
    };

    template<typename T>
    auto demangled = Demangled<T>::name();
```

fwyzard:
Can it be `constexpr` ?

fwyzard:
In general, I do not share the "remove boost at all costs" goal, so to me the new implementation needs to be at least as good as the old one, or possibly better.

SimeonEhrig:
`std::source_location::current()` is actual pretty well. I think it is more a question if we want remove the compilers, which does not support `std::source_location::current()`.

SimeonEhrig:
@fwyzard If you want, you can test it here: https://godbolt.org/z/P6MsrKTKv

Update: a more advanced example with namespace: https://godbolt.org/z/159Gs313h

SimeonEhrig:
> Can it be `constexpr` ?

[...]

PR #2409, opened 2024-10-02
remove boost as mandatory dependency

### LINKED ISSUE #481: Remove Core Boost Dependency

psychocoderHPC:
fix https://github.com/alpaka-group/alpaka/issues/481

Make boost an optional dependency.
`alpaka::core::demangled<TYPE>` returns a mangled name in case boost is not available. https://github.com/alpaka-group/alpaka/issues/2405 can solve this but should not be mixed into this PR.
This PR removes the mixing of alpakas own defines e.g `BOOST_LANG_HIP` which looked like boost but was exposed by alpaka under BOOST names.

This PR is a base for discussion at the next developer meeting. In a separate PR we could reduce the number of tested boost versions. We should also run tests without boost at all.


SimeonEhrig:
@psychocoderHPC Can you please split up the PR. At least in two PR: remove `boost::atomic_ref` and the compiler detection.

At the moment it is nearly impossible to review it and I also saw, that you didn't modify the CI, which is definitively useful.

SimeonEhrig:
I think also a third PR with the demangling function names is useful.

psychocoderHPC:
Yes, I can do this after we discuss it in general.

SimeonEhrig:
> Yes, I can do this after we discuss it in general.

You got me. I didn't read the description until the end :grimacing:

SimeonEhrig:
@psychocoderHPC Can you please short summarize the main changes?

If I'm not wrong the following boost features are used:
- preprocessor defines for compiler detection
- `boost::atomic_ref` if `std::atomic_ref` is not avaible
- demangle type names

Did I missed something?

psychocoderHPC:
> @psychocoderHPC Can you please short summarize the main changes?
> 
> If I'm not wrong the following boost features are used:
> 
> * preprocessor defines for compiler detection
> * `boost::atomic_ref` if `std::atomic_ref` is not avaible
> * demangle type names
> 
> Did I missed something?

no thats a good summery

psychocoderHPC:
solved with #2503

PR #2410, opened 2024-10-03
Rename namespace alpaka::concepts to alpaka::inteface

SimeonEhrig:
Rename the alpaka namespace `alpaka::concepts` to `alpaka::interface` because it can be easily mixed up with C++20 concepts. But `alpaka::concepts` functionality does not implement C++20 with C++17. Instead it allows to implement a type hierarchy allows to check for it. That's a similar concept like Java interfaces. Therefore we name it interface now.


SimeonEhrig:
@fwyzard Can you please have a look on it. I will also talk to  @psychocoderHPC afterwards.

psychocoderHPC:
I am fine with this change, this makes the way free for concepts.

include/alpaka/dev/Traits.hpp
```diff
@@ -61,13 +61,13 @@ namespace alpaka
 
     //! True if TDev is a device, i.e. if it implements the ConceptDev concept.
```
psychocoderHPC, line 62: please check the documentation. I will merge it but we should change it in a follow up PR.

psychocoderHPC (approved):

include/alpaka/dev/Traits.hpp
SimeonEhrig: That's something which I realized during the work at the `alpaka::concepts::Tag` PR. Each class which uses `alpaka::interface` needs to define a class, where it use it for the inheritance hierarchy. I didn't changed it in this PR, because it does not affect the functionality of the `alpaka::interface` functions.

I can rename all the classes in an extra PR. Also they don't follow an specific name scheme. For `alpaka::concepts::Tag` I use the following Interface class:

https://github.com/alpaka-group/alpaka/blob/85c330296d4d3567af56e1961503df1e293cbd1c/include/alpaka/acc/Tag.hpp#L15

Maybe I can overtake the name scheme for all other classes.

PR #2411, opened 2024-10-07
Prints Acc tags of tags-tuple at runtime

mehmetyusufoglu:
Examples was exiting silently if there are no enabled backends. This issue is reported by the assistants of the october2024 workshop.

 This PR creates a function that prints the Acc-tags of the enabled backends at runtime. If there is no backends, the user is informed.

```
./heatEquation2D
Check enabled accelerator tags:
No Tags!
```

```
./heatEquation2D 
Check enabled accelerator tags:
Tags: TagCpuSerial,TagCpuThreads
Using alpaka accelerator: AccCpuSerial<2,unsigned int>
Execution results correct!
Using alpaka accelerator: AccCpuThreads<2,unsigned int>
Execution results correct!
```

### DIFF
--- a/example/heatEquation2D/src/heatEquation2D.cpp
+++ b/example/heatEquation2D/src/heatEquation2D.cpp
@@ -204,6 +204,8 @@ auto example(TAccTag const&) -> int
 
 auto main() -> int
 {
+    std::cout << "Check enabled accelerator tags:" << std::endl;
+    alpaka::printTagNames<alpaka::EnabledAccTags>();
     // Execute the example once for each enabled accelerator.
     // If you would like to execute it for a single accelerator only you can use the following code.
     //  \code{.cpp}
--- a/include/alpaka/acc/Tag.hpp
+++ b/include/alpaka/acc/Tag.hpp
@@ -7,6 +7,7 @@
 #include "alpaka/core/BoostPredef.hpp"
 
 #include <iostream>
+#include <tuple>
 #include <type_traits>
 
 #define CREATE_ACC_TAG(tag_name)                                                                                      \
@@ -69,4 +70,28 @@ namespace alpaka
         alpaka::TagFpgaSyclIntel,
         alpaka::TagGpuSyclIntel>;
 
+    //!  \brief Function to print the names of each tag in the given tuple of tags
+    //!  \tparam TTuple is the type of the tuple of tags
+    template<typename TTuple>
+    void printTagNames()
+    {
+        // Check if the tuple is empty using std::tuple_size_v
+        if(std::tuple_size_v<TTuple> == 0)
+        {
+            std::cout << "No Tags!";
+        }
+        else
+        {
+            std::cout << "Tags: ";
+            // Print tags with comma in between
+            std::apply(
+                [](auto... args)
+                {
+                    auto index = std::tuple_size_v<TTuple>;
+                    ((std::cout << args.get_name() << (--index > 0u ? "," : "")), ...);
+                },
+                TTuple{});
+        }
+        std::cout << std::endl;
+    }
 } // namespace alpaka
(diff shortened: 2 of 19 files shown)


example/heatEquation2D/src/heatEquation2D.cpp
```diff
@@ -215,5 +215,7 @@ auto main() -> int
     //   TagCpuSerial, TagGpuHipRt, TagGpuCudaRt, TagCpuOmp2Blocks, TagCpuTbbBlocks,
     //   TagCpuOmp2Threads, TagCpuSycl, TagCpuTbbBlocks, TagCpuThreads,
     //   TagFpgaSyclIntel, TagGenericSycl, TagGpuSyclIntel
+
+    alpaka::printAccTags<alpaka::EnabledAccTags>();
```
psychocoderHPC: Please add this to all examples, this behaviour is not only affecting the heat equation

include/alpaka/acc/Tag.hpp
```diff
@@ -69,4 +70,21 @@ namespace alpaka
         alpaka::TagFpgaSyclIntel,
         alpaka::TagGpuSyclIntel>;
 
+    //!  \brief Function to print the names of each tag in the given tuple of acc tags
+    //!  \tparam TTuple is the type of the tuple of tags
+    template<typename TTuple>
+    void printAccTags()
+    {
+        // Check if the tuple is empty using std::tuple_size
+        if(std::tuple_size<TTuple>{} == 0)
```
psychocoderHPC: `std::size_size_v<TTuple>`

include/alpaka/acc/Tag.hpp
```diff
@@ -69,4 +70,21 @@ namespace alpaka
...
+        {
+            std::cout << "No Acc Tags!";
+        }
+        else
+        {
+            std::cout << "Acc Tags: ";
+            std::apply([](auto... args) { ((std::cout << args.get_name() << " "), ...); }, TTuple{});
```
psychocoderHPC: I suggest something like this following code (Not tested) to have a comma seperated list.

```C++
std::apply(
  [](auto... args) { 
    auto i = std::size_size_v<TTuple>;
    ((std::cout << args.get_name() << (--i > 0u ? "," : "")), ...);
  }, 
  TTuple{}
);

```

psychocoderHPC (changes requested):

include/alpaka/acc/Tag.hpp
mehmetyusufoglu: ok, good idea

include/alpaka/acc/Tag.hpp
```diff
@@ -69,4 +70,21 @@ namespace alpaka
         alpaka::TagFpgaSyclIntel,
         alpaka::TagGpuSyclIntel>;
 
+    //!  \brief Function to print the names of each tag in the given tuple of acc tags
+    //!  \tparam TTuple is the type of the tuple of tags
+    template<typename TTuple>
+    void printAccTags()
```
SimeonEhrig:
```suggestion
    void printTagNames()
```

Please avoid the term Acc where it is possible. This can easily cause confusion.

include/alpaka/acc/Tag.hpp
mehmetyusufoglu: done using std::tuple_size_v<TTuple>

include/alpaka/acc/Tag.hpp
mehmetyusufoglu: alternatives exists using `make_index_sequence` but i preferred this since it is still cleaner. thanks.

SimeonEhrig (approved):

psychocoderHPC (approved):

psychocoderHPC:
I updated already which CI jobs are required. This PR was created before the last merge where clang10-12 got removed, therefore I merge it with my admin power.

PR #2415, opened 2024-10-23
add static function dim() to alpaka::Vec

SimeonEhrig:
Easy way to get the dim of a alpaka::Vec.

### DIFF
--- a/include/alpaka/vec/Vec.hpp
+++ b/include/alpaka/vec/Vec.hpp
@@ -368,6 +368,12 @@ namespace alpaka
             return (*this)[I];
         }
 
+        //! \return The number of dimensions of the vector.
+        [[nodiscard]] ALPAKA_FN_HOST_ACC static consteval auto dim() noexcept -> TVal
+        {
+            return TDim::value;
+        }
+
         //! \return The element-wise sum of two vectors.
         ALPAKA_NO_HOST_ACC_WARNING
         ALPAKA_FN_HOST_ACC friend constexpr auto operator+(Vec const& p, Vec const& q) -> Vec
--- a/test/unit/vec/src/VecTest.cpp
+++ b/test/unit/vec/src/VecTest.cpp
@@ -177,7 +177,14 @@ TEST_CASE("basicVecTraits", "[vec]")
     }
 
     {
-        constexpr alpaka::Vec<Dim, Idx> vec3(static_cast<Idx>(47u), static_cast<Idx>(8u), static_cast<Idx>(3u));
+        using Vec3 = alpaka::Vec<Dim, Idx>;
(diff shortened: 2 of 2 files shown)


mehmetyusufoglu:
I opened another PR  (#2416)  without noticing this is already opened, I closed the other one but I think the name should be size insead of dim.

SimeonEhrig:
My explanation from the other PR, why we should not name the function `size()`.

We can compare `alpaka::Vec` with [std::extents](https://en.cppreference.com/w/cpp/container/mdspan/extents). `std::extents` has no explicit `size` but `rank`, what is the same like your `Dim`. If we use `std::extents` with `std::mdspan`, we have the function `size()` which returns the production of the values of each `rank` in the `std::extents`. See this example:

```c++
std::vector data(100, 0);
std::mdspan md(data.data(), std::extents{10, 10});

// output: 100
std::cout << md.size() << "\n";

// output:
// rank 0: 10
// rank 1: 10
for(auto i = 0; i < md.rank(); ++i){
    std::cout << "rank " << i << ": " << md.extent(i) << "\n"; 
}
```

Also the special case `alpaka::Vec<int, alpaka::DimInt<1>>` is the same like `std::vector<int>`. And in the case of `std::vector` `size()` will return the value of the one dimension.

Therefore is clear what is `size()` is doing. We can discuss, if we want to name the function `rank()` or `dim()` something else but `size()` is already used for a different function. For alpaka::Vec it is:

```c++
alpaka::Vec<int, alpaka::DimInt<1>> vec = {10};
std::cout << vec[0] << "\n"; // outputs the size 10
std::cout << vec.size() << "\n"; // outputs the size 10

alpaka::Vec<int, alpaka::DimInt<1>> vec2 = {10, 2};
std::cout << vec.size() << "\n"; // outputs the size 20
``

File: include/alpaka/vec/Vec.hpp [hunk 1]
```diff
@@ -368,6 +368,12 @@ namespace alpaka
             return (*this)[I];
         }
 
+        //! \return The number of dimensions of the vector.
+        [[nodiscard]] ALPAKA_FN_HOST_ACC static consteval auto dim() noexcept -> TVal
```
j-stephan, line 372: Didn't we agree on `dims` (or similar) here?

mehmetyusufoglu: We agreed with the Simeon's suggession, dim.

SimeonEhrig: I talked with @psychocoderHPC . Because it returns only a single number and `std::mdspan` use `rank()` and not `ranks()` , singular makes more sense.

psychocoderHPC (approved):

PR #2420, opened 2024-11-07
fix bablestream benchmark

mehmetyusufoglu:
1.  Some  of the 5 kernels of Babelstream-benchmark were not connected to each other, with this change if one of them changed somehow and fails; error is cached in the last result. ( Since we don't check after each kernel run, this is needed to make sure all kernels are connected.)
2.  Using arrays in a different order in calling different kernels might affect the performance (although not observed) due to caching, therefore using same arrays for the same kernels (as in the original [babelstream of UoB](https://github.com/UoB-HPC/BabelStream/blob/main/src/main.cpp)) in the kernel call sequence is also done by above change.
3. An optional kernel is added, `NStream`. This can be run separately alone. 
4. One of the 5 kernels of babelstream, `the triad kernel,` was optionally being run alone [in the original code by UoB](https://github.com/UoB-HPC/BabelStream/blob/main/src/main.cpp). This option is also added.

 This PR is an extension of previous PR: [#2299](https://github.com/alpaka-group/alpaka/pull/2299)
 
**New parameters and kernel calls with specific arrays in the kernel call sequence to avoid cache usage differences:**

A = 0.1 B= 0.2 C= 0.0 scalar = 0.4
C = A // copy
B = scalar * C // mult
C = A + B  // add
A = B + scalar * C  // triad
Missing optional kernel NStream is added 

**Update November 25:** ALL 5 Babelstream Kernels (copy add mul dot triad) run for all backends.

RESULTS
```
Randomness seeded to: 3331482523
Kernels: Init, Copy, Mul, Add, Triad, Dot Kernels


AcceleratorType:AccCpuSerial<1,unsigned int>
NumberOfRuns:2
Precision:single
DataSize(items):262144
DeviceName:13th Gen Intel(R) Core(TM) i7-1360P
WorkDivInit :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.00188882
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       23.732          0.00013255 0.00013255 0.00013255 3.1457 
 CopyKernel      23.401          8.9618e-05 8.9618e-05 8.9618e-05 2.0972 
 DotKernel       0.96404         0.0021754 0.0021754 0.0021754 2.0972 
 InitKernel      0.96656         0.0032545 0.0032545 0.0032545 3.1457 
 MultKernel      21.214          9.8855e-05 9.8855e-05 9.8855e-05 2.0972 
 TriadKernel     24.391          0.00012897 0.00012897 0.00012897 3.1457 

Kernels: Init, Copy, Mul, Add, Triad, Dot Kernels


AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:2
Precision:single
DataSize(items):262144
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.00317802
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       72.282          4.352e-05 4.352e-05 4.352e-05 3.1457 
 CopyKernel      77.294          2.7132e-05 2.7132e-05 2.7132e-05 2.0972 
 DotKernel       30.834          6.8014e-05 6.8014e-05 6.8014e-05 2.0972 
 InitKernel      11.531          0.00027281 0.00027281 0.00027281 3.1457 
 MultKernel      64.534          3.2497e-05 3.2497e-05 3.2497e-05 2.0972 
 TriadKernel     74.74           4.2089e-05 4.2089e-05 4.2089e-05 3.1457 

Kernels: Init, Copy, Mul, Add, Triad, Dot Kernels


AcceleratorType:AccCpuSerial<1,unsigned int>
NumberOfRuns:2
Precision:double
DataSize(items):262144
DeviceName:13th Gen Intel(R) Core(TM) i7-1360P
WorkDivInit :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (262144), blockThreadExtent: (1), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.00384645
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       16.979          0.00037054 0.00037054 0.00037054 6.2915 
 CopyKernel      15.052          0.00027866 0.00027866 0.00027866 4.1943 
 DotKernel       1.208           0.0034721 0.0034721 0.0034721 4.1943 
 InitKernel      1.7377          0.0036205 0.0036205 0.0036205 6.2915 
 MultKernel      15.806          0.00026537 0.00026537 0.00026537 4.1943 
 TriadKernel     17.379          0.000362 0.000362 0.000362 6.2915 

Kernels: Init, Copy, Mul, Add, Triad, Dot Kernels


AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:2
Precision:double
DataSize(items):262144
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.00536517
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       77.046          8.1658e-05 8.1658e-05 8.1658e-05 6.2915 
 CopyKernel      65.102          6.4427e-05 6.4427e-05 6.4427e-05 4.1943 
 DotKernel       33.015          0.00012704 0.00012704 0.00012704 4.1943 
 InitKernel      64.109          9.8137e-05 9.8137e-05 9.8137e-05 6.2915 
 MultKernel      74.419          5.6361e-05 5.6361e-05 5.6361e-05 4.1943 
 TriadKernel     77.08           8.1622e-05 8.1622e-05 8.1622e-05 6.2915 

All tests passed (16 assertions in 4 test cases)

RUN FOR BENCHMARKING

./babelstream --array-size=33554432 --number-runs=100
Array size set to: 33554432
Number of runs provided: 100
Randomness seeded to: 1929971841
Kernels: Init, Copy, Mul, Add, Triad, Dot Kernels


AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:100
Precision:single
DataSize(items):33554432
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.311451
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       90.574          0.0044456 0.0044824 0.0044742 402.65 
 CopyKernel      90.291          0.002973 0.0030168 0.0030108 268.44 
 DotKernel       92.872          0.0028904 0.0029612 0.002928 268.44 
 InitKernel      92.054          0.0043741 0.0043741 0.0043741 402.65 
 MultKernel      90.464          0.0029673 0.0030106 0.0030042 268.44 
 TriadKernel     90.758          0.0044366 0.0044949 0.0044788 402.65 

Kernels: Init, Copy, Mul, Add, Triad, Dot Kernels


AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:100
Precision:double
DataSize(items):33554432
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.589239
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       90.494          0.008899 0.008962 0.0089358 805.31 
 CopyKernel      88.791          0.0060464 0.0061075 0.0060708 536.87 
 DotKernel       93.117          0.0057656 0.0058216 0.005797 536.87 
 InitKernel      89.423          0.0090056 0.0090056 0.0090056 805.31 
 MultKernel      89.072          0.0060274 0.0061036 0.0060612 536.87 
 TriadKernel     90.582          0.0088903 0.0089994 0.0089574 805.31 

All tests passed (8 assertions in 2 test cases)

```

### DIFF
--- a/benchmarks/babelstream/src/babelStreamCommon.hpp
+++ b/benchmarks/babelstream/src/babelStreamCommon.hpp
@@ -3,11 +3,11 @@
 #include <algorithm>
 #include <chrono>
 #include <cmath>
-#include <ctime>
 #include <iomanip>
 #include <iostream>
 #include <limits>
 #include <map>
+#include <memory>
 #include <numeric>
 #include <sstream>
 #include <stdexcept>
@@ -22,61 +22,112 @@ namespace
     // According to tests, 2^25 or larger values are needed for proper benchmarking:
     // ./babelstream --array-size=33554432 --number-runs=100
     // To prevent timeouts in CI, a smaller default value is used.
-    [[maybe_unused]] auto arraySizeMain = 1024 * 1024;
+    [[maybe_unused]] auto arraySizeMain = 1024 * 256;
 
     // Minimum array size to be used.
-    [[maybe_unused]] constexpr auto minArrSize = 1024 * 128;
+    [[maybe_unused]] constexpr auto minArrSize = 1024 * 256;
 
     // Scalar value for Mul and Triad kernel parameters.
-    [[maybe_unused]] constexpr auto scalarVal = 2.0f;
+    [[maybe_unused]] constexpr double scalarVal = 0.4;
 
     // Block thread extent for DotKernel test work division parameters.
     [[maybe_unused]] constexpr auto blockThreadExtentMain = 1024;
+    [[maybe_unused]] constexpr auto dotGridBlockExtent = 256;
 
     // Number of runs for each kernel, can be changed by command line arguments.
     // At least 100 runs are recommended for good benchmarking.
     // To prevent timeouts in CI, a small value is used.
     [[maybe_unused]] auto numberOfRuns = 2;
 
-    // Data input value for babelstream.
-    [[maybe_unused]] constexpr auto valA = 1.0f;
+    // Data input values for babelstream.
+    [[maybe_unused]] constexpr double initA = 0.1;
+    [[maybe_unused]] constexpr double initB = 0.2;
+    // Change this if triad kernel is going to be run alone
+    [[maybe_unused]] constexpr double initC = 0.0;
+
+    //! Values corresponding to the command line argument run-kernels
+    enum class KernelsToRun
+    {
+        All, // init, add, copy, mul, triad, dot
+        Triad, // only init and triad
+        NStream // only init and nstream
+    };
+
+    // Define the variable showing the kernel(s) being run
+    [[maybe_unused]] KernelsToRun kernelsToBeExecuted{KernelsToRun::All};
 
     //! handleCustomArguments Gets custom cmd line arguments from the all arguments.
     //! Namely gets --array-size=1234 and --number-runs=1234 and keeps the others which are
     //! command line args for Catch2 session.
     [[maybe_unused]] static void handleCustomArguments(int& argc, char* argv[])
     {
-        std::vector<char*> newArgv;
-        newArgv.push_back(argv[0]); // Keep the program name
+        std::vector<char*> newArgv({argv[0]}); // keep program name
 
         for(int i = 1; i < argc; ++i)
         {
             std::string arg = argv[i];
             if(arg.rfind("--array-size=", 0) == 0)
             {
-                auto const arrSize = std::stoi(arg.substr(13)); // Convert to integer
-                if(arrSize > minArrSize)
+                try
                 {
-                    arraySizeMain = arrSize;
-                    std::cout << "Array size provided(items): " << arraySizeMain << std::endl;
+                    // Convert argument to integer
+                    auto arrSize = std::stoi(arg.substr(13));
+                    if(arrSize > minArrSize)
+                    {
+                        arraySizeMain = arrSize;
+                        std::cout << "Array size set to: " << arraySizeMain << std::endl;
+                    }
+                    else
+                    {
+                        std::cout << "Array size too small. Must be at least " << minArrSize
+                                  << ", using default: " << arraySizeMain << std::endl;
+                    }
                 }
-                else
+                catch(std::invalid_argument const&)
                 {
-                    std::cout << "Too small array size given. Must be at least " << minArrSize << std::endl;
-                    std::cout << "Using default array size(number of items): " << arraySizeMain << std::endl;
+                    std::cerr << "Invalid array size argument: " << arg << ". Default value used." << std::endl;
                 }
             }
             else if(arg.rfind("--number-runs=", 0) == 0)
             {
-                auto const numRuns = std::stoi(arg.substr(14)); // Convert to integer
-                if(numRuns > 0)
+                try
                 {
-                    numberOfRuns = numRuns;
-                    std::cout << "Number of runs provided: " << numberOfRuns << std::endl;
+                    // Convert argument to integer
+                    auto const numRuns = std::stoi(arg.substr(14));
+                    if(numRuns > 0)
+                    {
+                        numberOfRuns = numRuns;
+                        std::cout << "Number of runs provided: " << numberOfRuns << std::endl;
+                    }
+                    else
+                    {
+                        std::cout << "Using default number of runs: " << numberOfRuns << std::endl;
+                    }
                 }
-                else
+                catch(std::invalid_argument const&)
                 {
-                    std::cout << "Using default number of runs: " << numberOfRuns << std::endl;
+                    std::cerr << "Invalid number of runs argument: " << arg << " . Default value used." << std::endl;
+                }
+            }
+            else if(arg.rfind("--run-kernels=", 0) == 0)
+            {
+                // Get argument to determine which kernels will be run
+                auto const kernelsString = arg.substr(14);
+                if(kernelsString == "nstream")
+                {
+                    std::cout << "Only nstream kernel will be executed." << std::endl;
+                    kernelsToBeExecuted = KernelsToRun::NStream;
+                }
+                else if(kernelsString == "triad")
+                {
+                    kernelsToBeExecuted = KernelsToRun::Triad;
+                    std::cout << "Only triad kernel will be executed." << std::endl;
+                }
+                else if(kernelsString == "all")
+                {
+                    // The variable kernelsToBeExecuted default value is  "all";
+                    kernelsToBeExecuted = KernelsToRun::All;
+                    std::cout << "All 5 babelstream kernels are going to be executed." << std::endl;
                 }
             }
             else
@@ -87,7 +138,11 @@ namespace
             if(arg.rfind("-?", 0) == 0 || arg.rfind("--help", 0) == 0 || arg.rfind("-h", 0) == 0)
             {
                 std::cout << "Usage of custom arguments (arguments which are not Catch2):  --array-size=33554432 and "
-                             "--number-runs=100"
+                             "--number-runs=100\n"
+                          << std::endl;
+                std::cout << "If you want to run only nstream kernel or triad kernel use --run-kernels=nstream or  "
+                             "--run-kernels=triad. Otherwise all 5 standard kernels will be executed. Copy, Mul, Add, "
+                             "Triad and Dot kernel."
                           << std::endl;
             }
         }
@@ -98,6 +153,12 @@ namespace
         {
             argv[i] = newArgv[static_cast<size_t>(i)];
         }
+
+        // Array size must a multiple of
+        if(arraySizeMain % blockThreadExtentMain != 0)
+            throw std::runtime_error(
+                "Array size is " + std::to_string(arraySizeMain) + ". It must be a multiple of block-size "
+                + std::to_string(blockThreadExtentMain));
     }
 
     //! FuzzyEqual compares two floating-point or integral type values.
@@ -111,7 +172,7 @@ namespace
     {
         if constexpr(std::is_floating_point_v<T>)
         {
-            return std::fabs(a - b) < std::numeric_limits<T>::epsilon() * static_cast<T>(100.0);
+            return std::fabs(a - b) < (std::numeric_limits<T>::epsilon() * static_cast<T>(100.0));
         }
         else if constexpr(std::is_integral_v<T>)
         {
@@ -213,12 +274,14 @@ namespace
         NumRuns,
         DataSize,
         DataType,
+        CopyTimeFromAccToHost,
         WorkDivInit,
         WorkDivCopy,
         WorkDivAdd,
         WorkDivTriad,
         WorkDivMult,
         WorkDivDot,
+        WorkDivNStream,
         DeviceName,
         TimeUnit,
         KernelNames,
@@ -251,6 +314,8 @@ namespace
             return "DataSize(items)";
         case BMInfoDataType::DataType:
             return "Precision";
+        case BMInfoDataType::CopyTimeFromAccToHost:
+            return "AccToHost Memcpy Time(sec)";
         case BMInfoDataType::DeviceName:
             return "DeviceName";
         case BMInfoDataType::TimeUnit:
@@ -279,6 +344,8 @@ namespace
             return "WorkDivMult ";
         case BMInfoDataType::WorkDivDot:
             return "WorkDivDot  ";
+        case BMInfoDataType::WorkDivNStream:
+            return "WorkDivNStream";
         default:
             return "";
         }
@@ -314,11 +381,159 @@ namespace
         return bytesReadWriteGB / static_cast<double>(runTimeSeconds);
     }
 
+    //! \brief calculateBabelstreamExpectedResults Fins the expected values by doing the same operations with the
+    //! initial values \param expectedA Array item value expected (all array values are equal in original babelstream)
+    //! \param expectedB Array item value expected (all array values are equal in original babelstream)
+    //! \param expectedC Array item value expected (all array values are equal in original babelstream)
+    template<typename T>
+    [[maybe_unused]] static void calculateBabelstreamExpectedResults(T& expectedA, T& expectedB, T& expectedC)
+    {
+        //
+        // All items of arrays are the same, therefore an expected value is for the whole array
+        for(auto i = 0; i < numberOfRuns; i++)
+        {
+            if(kernelsToBeExecuted == KernelsToRun::All)
+            {
+                expectedC = expectedA;
+                expectedB = static_cast<T>(scalarVal) * expectedC;
+                expectedC = expectedA + expectedB;
+                expectedA = expectedB + static_cast<T>(scalarVal) * expectedC;
+            }
+            else if(kernelsToBeExecuted == KernelsToRun::Triad)
+            {
+                expectedA = expectedB + static_cast<T>(scalarVal) * expectedC;
+            }
+            else if(kernelsToBeExecuted == KernelsToRun::NStream)
+            {
+                // each run changes the result
+                expectedA += expectedB + static_cast<T>(scalarVal) * expectedC;
+            }
+        }
+    }
+
+    /**
+     * /brief The RuntimeResults class bundles the kernel runtime data in a map
+     * The keys of the map are kernel names the values of the map are KernelRunData struct pointers
+     */
+    class RuntimeResults
+    {
+        struct KernelRunData
+        {
+            std::vector<double> timingsSuccessiveRuns; // Stores execution timings of successive runs
+            double byteReadWriteMB{0}; // Bytes read/write in MB
+            double bandwidthKernel{0}; // Bandwidth of kernel
+            double minExecTime{0}; // Minimum execution time
+            double maxExecTime{0}; // Maximum execution time
+            double avgExecTime{0}; // Average execution time
+        };
+
+    public:
+        // Map from kernelName (string) to a unique_ptr for KernelRunData
+        // Using unique_ptr for automatic memory management
+        std::map<std::string, std::unique_ptr<KernelRunData>> kernelToRundataMap;
+
+        // Function to initialize the byteReadWriteMB field for each kernel
+        template<typename DataType>
+        void initializeByteReadWrite(size_t arraySize)
+        {
+            // Define kernel names and their throughput values based on the provided array size
+            std::map<std::string, double> throughputValues
+                = {{"InitKernel", getDataThroughput<DataType>(3u, static_cast<unsigned>(arraySize))},
+                   {"CopyKernel", getDataThroughput<DataType>(2u, static_cast<unsigned>(arraySize))},
+                   {"MultKernel", getDataThroughput<DataType>(2u, static_cast<unsigned>(arraySize))},
+                   {"AddKernel", getDataThroughput<DataType>(3u, static_cast<unsigned>(arraySize))},
+                   {"TriadKernel", getDataThroughput<DataType>(3u, static_cast<unsigned>(arraySize))},
+                   {"DotKernel", getDataThroughput<DataType>(2u, static_cast<unsigned>(arraySize))},
+                   {"NStreamKernel", getDataThroughput<DataType>(2u, static_cast<unsigned>(arraySize))}};
+
+            // Populate each KernelRunData entry in kernelToRundataMap
+            for(auto const& [kernelName, throughput] : throughputValues)
+            {
+                // Check if the kernel name exists in the map
+                if(kernelToRundataMap.find(kernelName) != kernelToRundataMap.end())
+                {
+                    // Set the byteReadWriteMB field in the corresponding KernelRunData
+                    kernelToRundataMap[kernelName]->byteReadWriteMB = throughput;
+                }
+            }
+        }
+
+        //! /brief calculateBandwidthsForKernels Function calculates bandwidth for each kernel and update execution
+        //! times Fills the fields of bandwidth, execution min-time, execution max-time and execution avg-time
+        template<typename DataType>
+        void calculateBandwidthsForKernels()
+        {
+            for(auto const& [kernelName, kernelData] : kernelToRundataMap)
+            {
+                // Calculate min and max execution times from recorded vector of times for the kernel named kernelName
+                auto const minmaxPair = findMinMax(kernelData->timingsSuccessiveRuns);
+                kernelData->minExecTime = minmaxPair.first;
+                kernelData->maxExecTime = minmaxPair.second;
+
+                // Calculate average execution time
+                kernelData->avgExecTime = findAverage(kernelData->timingsSuccessiveRuns);
+
+                // Calculate bandwidth based on byteReadWriteMB and min execution time
+                kernelData->bandwidthKernel = calculateBandwidth(kernelData->byteReadWriteMB, minmaxPair.first);
+            }
+        }
+
+        //! /brief  Get item from each struct in the map and make a vector
+        //! /tparam Func is the accessor function to access to a specific field
+        template<typename Func>
+        std::vector<double> getItemFromStructs(Func accessor) const noexcept
+        {
+            std::vector<double> results;
+            for(auto const& [key, dataStruct] : kernelToRundataMap)
+            {
+                results.push_back(accessor(dataStruct.get())); // Dereference unique_ptr with .get()
+            }
+            return results;
+        }
+
+        // Functions to retrieve specific fields as vectors for all kernels
+        std::vector<double> getBandwidthKernelVec() const noexcept
+        {
+            return getItemFromStructs([](KernelRunData const* dataStruct) { return dataStruct->bandwidthKernel; });
+        }
+
+        std::vector<double> getThroughputKernelArray() const noexcept
+        {
+            return getItemFromStructs([](KernelRunData const* dataStruct) { return dataStruct->byteReadWriteMB; });
+        }
+
+        std::vector<double> getAvgExecTimeKernelArray() const noexcept
+        {
+            return getItemFromStructs([](KernelRunData const* dataStruct) { return dataStruct->avgExecTime; });
+        }
+
+        std::vector<double> getMinExecTimeKernelArray() const noexcept
+        {
+            return getItemFromStructs([](KernelRunData const* dataStruct) { return dataStruct->minExecTime; });
+        }
+
+        std::vector<double> getMaxExecTimeKernelArray() const noexcept
+        {
+            return getItemFromStructs([](KernelRunData const* dataStruct) { return dataStruct->maxExecTime; });
+        }
+
+        // Function to add a kernelName-timesForRuns pair by storing a unique_ptr to KernelRunData
+        void addKernelTimingsVec(std::string const& kernelName) noexcept
+        {
+            // Use make_unique to create a new KernelRunData object and store it in the map
+            kernelToRundataMap[kernelName] = std::make_unique<KernelRunData>();
+        }
+    };
+
     //! MetaData class to store and serialize benchmark information.
     //! \details The MetaData class includes a single map to keep all benchmark information and provides serialization
     //! methods for generating output.
-    class MetaData
+    class BenchmarkMetaData
     {
+    private:
+        // Information type to string. String can be comma separated values.
+        std::map<BMInfoDataType, std::string> metaDataMap;
+
     public:
         //! setItem  Sets an item in the metadata map.
         //! \tparam T The type of the value to store.
@@ -353,29 +568,32 @@ namespace
         {
             std::stringstream ss;
             // define lambda to add values to a string stream created already
-            auto addItemValue = [&, this](BMInfoDataType item) {
-                ss << "\n" << typeToTypeStr(item) << ":" << metaDataMap.at(item);
+            auto addItemValueToSS = [&, this](BMInfoDataType item)
+            {
+                if(metaDataMap.count(item) != 0)
+                    ss << "\n" << typeToTypeStr(item) << ":" << metaDataMap.at(item);
             };
 
-            // Initially chose some data to serialize
+            // Initially choose some data to serialize from the meta-data map to add to string stream
             ss << "\n";
-            addItemValue(BMInfoDataType::AcceleratorType);
-            addItemValue(BMInfoDataType::NumRuns);
-            addItemValue(BMInfoDataType::DataType);
-            addItemValue(BMInfoDataType::DataSize);
-            addItemValue(BMInfoDataType::DeviceName);
-            addItemValue(BMInfoDataType::WorkDivInit);
-            addItemValue(BMInfoDataType::WorkDivCopy);
-            addItemValue(BMInfoDataType::WorkDivMult);
-            addItemValue(BMInfoDataType::WorkDivAdd);
-            addItemValue(BMInfoDataType::WorkDivTriad);
-            if(metaDataMap.count(BMInfoDataType::WorkDivDot) != 0)
-                addItemValue(BMInfoDataType::WorkDivDot);
+            addItemValueToSS(BMInfoDataType::AcceleratorType);
+            addItemValueToSS(BMInfoDataType::NumRuns);
+            addItemValueToSS(BMInfoDataType::DataType);
+            addItemValueToSS(BMInfoDataType::DataSize);
+            addItemValueToSS(BMInfoDataType::DeviceName);
+            addItemValueToSS(BMInfoDataType::WorkDivInit);
+            addItemValueToSS(BMInfoDataType::WorkDivCopy);
+            addItemValueToSS(BMInfoDataType::WorkDivMult);
+            addItemValueToSS(BMInfoDataType::WorkDivAdd);
+            addItemValueToSS(BMInfoDataType::WorkDivTriad);
+            addItemValueToSS(BMInfoDataType::WorkDivDot);
+            addItemValueToSS(BMInfoDataType::WorkDivNStream);
+            addItemValueToSS(BMInfoDataType::CopyTimeFromAccToHost);
 
+            // if the item is a string with delimited values get the item then the value at the index
             auto getItemFromStrList = [this](BMInfoDataType item, int index) -> std::string
             {
                 std::string const str = metaDataMap.at(item);
-
                 if(index < 1)
                 {
                     throw std::invalid_argument("Index must be 1 or greater.");
@@ -405,7 +623,7 @@ namespace
                 throw std::out_of_range("Index out of range");
             };
 
-            // Prepare Table
+            // Prepare Table Display
             // Table column names
             ss << std::endl;
             ss << std::left << std::setw(15) << typeToTypeStr(BMInfoDataType::KernelNames) << " " << std::left
@@ -421,7 +639,7 @@ namespace
             // Table rows. Print test results for each kernel line by line
             for(auto i = 1; i <= numberOfKernels; i++)
             {
-                // Print the row for the kernel i
+                // Print the row for the kernel i.
                 ss << " " << std::left << std::setw(15) << getItemFromStrList(BMInfoDataType::KernelNames, i) << " ";
                 ss << std::left << std::setw(15) << getItemFromStrList(BMInfoDataType::KernelBandwidths, i) << " ";
                 ss << std::left << std::setw(8) << getItemFromStrList(BMInfoDataType::KernelMinTimes, i) << " ";
@@ -433,8 +651,6 @@ namespace
 
             return ss.str();
         }
-
-    private:
-        std::map<BMInfoDataType, std::string> metaDataMap;
     };
+
 } // namespace
--- a/benchmarks/babelstream/src/babelStreamMainTest.cpp
+++ b/benchmarks/babelstream/src/babelStreamMainTest.cpp
@@ -9,24 +9,32 @@
 #include <catch2/catch_template_test_macros.hpp>
 #include <catch2/catch_test_macros.hpp>
 
+#include <algorithm>
+#include <iterator>
 #include <string>
 
 /**
- * Babelstream benchmarking example. Babelstream has 5 kernels. Add, Multiply, Copy, Triad and Dot.
- * Babelstream is a memory-bound benchmark since the main operation in the kernels has high Code Balance (bytes/FLOP)
- * value. For example c[i] = a[i] + b[i]; has 2 reads 1 writes and has one FLOP operation. For double precision each
- * read-write is 8 bytes. Hence Code Balance (3*8 / 1) = 24 bytes/FLOP.
+ * Babelstream benchmarking example. Babelstream has 5 kernels. Add, Multiply, Copy, Triad and Dot. NStream is
+ * optional. Init kernel is run before 5 standard kernel sequence. Babelstream is a memory-bound benchmark since the
+ * main operation in the kernels has high Code Balance (bytes/FLOP) value. For example c[i] = a[i] + b[i]; has 2 reads
+ * 1 writes and has one FLOP operation. For double precision each read-write is 8 bytes. Hence Code Balance (3*8 / 1) =
+ * 24 bytes/FLOP.
  *
  * Some implementations and the documents are accessible through https://github.com/UoB-HPC
  *
  * Can be run with custom arguments as well as catch2 arguments
- * Run with Custom arguments:
+ * Run with Custom arguments and for kernels: init, copy, mul, add, triad (and dot kernel if a multi-thread acc
+ * available):
  * ./babelstream --array-size=33554432 --number-runs=100
- * Runt with default array size and num runs:
+ * Run with Custom arguments and select from 3 kernel groups: all, triad, nstream
+ * ./babelstream --array-size=33554432 --number-runs=100 --run-kernels=triad (only triad kernel)
+ * ./babelstream --array-size=33554432 --number-runs=100 --run-kernels=nstream (only nstream kernel)
+ * ./babelstream --array-size=33554432 --number-runs=100 --run-kernels=all (default case. Add, Multiply, Copy, Triad
+ * and Dot) Run with default array size and num runs:
  * ./babelstream
- * Run with Catch2 arguments and defaul arrary size and num runs:
+ * Run with Catch2 arguments and default array size and num runs:
  * ./babelstream --success
- * ./babelstream -r a.xml
+ * ./babelstream -r xml
  * Run with Custom and catch2 arguments together:
  * ./babelstream  --success --array-size=1280000 --number-runs=10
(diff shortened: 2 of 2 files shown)


psychocoderHPC:
@mehmetyusufoglu Can you please check if CPU is working too.

mehmetyusufoglu:
It certainly works for CPUs, The CPU results file is attached. (But i am trying to run dot kernel for CPUs, which is originally using 256 blocks and 1024 threads per block. Because i saw in a recent paper it is run at a CPU)
[CPU_RUNS.txt](https://github.com/user-attachments/files/17716890/CPU_RUNS.txt)

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -312,136 +345,252 @@ void testKernels()
...
+            }
+        }
+
+        metaData.setItem(BMInfoDataType::WorkDivInit, workDivInit);
+        metaData.setItem(BMInfoDataType::WorkDivCopy, workDivCopy);
+        metaData.setItem(BMInfoDataType::WorkDivAdd, workDivAdd);
+        metaData.setItem(BMInfoDataType::WorkDivMult, workDivMult);
+        metaData.setItem(BMInfoDataType::WorkDivTriad, workDivTriad);
+
+    } // for all 5 kernels and dot if accelerator is suitable
```
psychocoderHPC: I do not see the verification if all benchmarks will be executed. Please use the verification provided upstream this will ensure all is correct without doing a verification independent for all combinations.

https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L513

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -312,136 +345,252 @@ void testKernels()
...
+                        alpaka::getPtrNative(bufAccSumPerBlock),
+                        static_cast<alpaka::Idx<Acc>>(arraySize));
+                },
+                "DotKernel");
+
+            alpaka::memcpy(queue, bufHostSumPerBlock, bufAccSumPerBlock, gridBlockExtent);
+            alpaka::wait(queue);
 
+            DataType const* sumPtr = std::data(bufHostSumPerBlock);
+            float const result = std::reduce(sumPtr, sumPtr + gridBlockExtent, 0.0f);
```
psychocoderHPC: This and the memcpy has be be part of the measurement, because that's our fault that we not execute the full reduction on device. 
To have a fair comparison the allocation of the result must be part of `measureKernelExec` too but the cuda upstream implementation is cheating here too so I assume this is fine to be allocated outside of `measureKernelExec`

psychocoderHPC (changes requested):

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -312,136 +345,252 @@ void testKernels()
...
+            { alpaka::exec<Acc>(queue, workDivAdd, AddKernel(), bufAccInputAPtr, bufAccInputBPtr, bufAccOutputCPtr); },
+            "AddKernel");
+    }
 
-    // Test the Triad-kernel. Calculate C=A+scalar*B where B=scalar*A.
-    measureKernelExec(
-        [&]()
-        { alpaka::exec<Acc>(queue, workDivTriad, TriadKernel(), bufAccInputAPtr, bufAccInputBPtr, bufAccOutputCPtr); },
-        "TriadKernel");
+    if(kernelsToBeExecuted == KernelsToRun::All || kernelsToBeExecuted == KernelsToRun::Triad)
```
chillenzer:
```suggestion
    else if(kernelsToBeExecuted == KernelsToRun::Triad)
```
following https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L532

benchmarks/babelstream/src/babelStreamCommon.hpp
```diff
@@ -28,55 +28,105 @@ namespace
     [[maybe_unused]] constexpr auto minArrSize = 1024 * 128;
 
     // Scalar value for Mul and Triad kernel parameters.
-    [[maybe_unused]] constexpr auto scalarVal = 2.0f;
+    [[maybe_unused]] constexpr double scalarVal = 0.4;
 
     // Block thread extent for DotKernel test work division parameters.
     [[maybe_unused]] constexpr auto blockThreadExtentMain = 1024;
+    [[maybe_unused]] constexpr auto dotGridBlockExtent = 256;
```
psychocoderHPC, line 35: feel free to add a extent for CPUs too to support CPU dot execution required for the verification.

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: The lines 419 to 440 verifies init and 4 kernels. Dot kernel is verified separately. I guess you mean reorganization of verification so that it will be cleaner, like in the link?

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: In [the original code](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.cu), they don't do full reduction. Kernel calculates dot product at the block level but does not sum.


```
template <class T>
__global__ void dot_kernel(const T * a, const T * b, T * sum, int array_size)
{
  __shared__ T tb_sum[TBSIZE];
  int i = blockDim.x * blockIdx.x + threadIdx.x;
  const size_t local_i = threadIdx.x;
  tb_sum[local_i] = {};
  for (; i < array_size; i += blockDim.x*gridDim.x)
    tb_sum[local_i] += a[i] * b[i];

  for (int offset = blockDim.x / 2; offset > 0; offset /= 2)
  { __syncthreads();
     if (local_i < offset)
     { tb_sum[local_i] += tb_sum[local_i+offset];
     }
  }
  if (local_i == 0)
    sum[blockIdx.x] = tb_sum[local_i];
}
```

mehmetyusufoglu: ok, I saw these are also part of measurement. Thanks.

mehmetyusufoglu: done, thanks.  [reduce is taken into measure]

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
-    std::vector<std::string> kernelLabels;
-    // Vector for collecting successive run-times of a single kernel in benchmark macro
-    std::vector<double> times;
+    auto const workDivNStream = alpaka::getValidWorkDiv(
+        kernelCfg,
+        devAcc,
+        NstreamKernel(),
+        bufAccInputAPtr,
+        bufAccInputBPtr,
+        bufAccOutputCPtr);
```
chillenzer, line 322: In the original babelstream cuda implementation, the (equivalent of a) workdiv is a compile-time constant, see [here](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.cu#L136) for a kernel call and [here](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.h#L18) and [here](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L57) for the corresponding definitions.

I think it should be a conscious decision if we want to do it differently (because we might argue that that's the proper way to do it in alpaka), and if so, we should document that choice somewhere. Of course, this applies to all instances of using `getValidWorkDiv`.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
+    // Init kernel
+    alpaka::exec<Acc>(
+        queue,
+        workDivInit,
+        InitKernel(),
+        bufAccInputAPtr,
+        bufAccInputBPtr,
+        bufAccOutputCPtr,
+        static_cast<DataType>(initA),
+        static_cast<DataType>(initB));
```
chillenzer: [In the original](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L309), this call is already timed.

chillenzer: Why is there no initialisation value given for `C`?

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -92,14 +102,14 @@ struct MultKernel
     //! \tparam TAcc The accelerator environment to be executed on.
     //! \tparam T The data type
     //! \param acc The accelerator to be executed on.
-    //! \param a Pointer for vector a
+    //! \param c Pointer for vector c
     //! \param b Pointer for result vector b
     template<typename TAcc, typename T>
-    ALPAKA_FN_ACC void operator()(TAcc const& acc, T* const a, T* b) const
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, T* const c, T* b) const
```
chillenzer: These two arguments are interchanged as compared to [upstream](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.cu#L182). The change is consistent with the swapped arguments in the call (so everything is correct here) but for easier readability we might still want to swap this.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
+                        alpaka::wait(queue);
+                        DataType const* sumPtr = std::data(bufHostSumPerBlock);
+                        resultDot = std::reduce(sumPtr, sumPtr + gridBlockExtent, 0.0f);
+                    },
+                    "DotKernel");
+                // Add workdiv to the list of workdivs to print later
+                metaData.setItem(BMInfoDataType::WorkDivDot, workDivDot);
+            }
+        }
+        else if(kernelsToBeExecuted == KernelsToRun::NStream)
```
chillenzer: Why is there an `else` here? It reads slightly inconsistently and produces a coupling between the two conditions that is not present for the other conditions.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
+                             accMatchesTags<TAcc, alpaka::TagGpuCudaRt, alpaka::TagGpuHipRt, alpaka::TagGpuSyclIntel>)
+            {
+                using WorkDiv = alpaka::WorkDivMembers<Dim, Idx>;
+                // Threads per block for Dot kernel
+                constexpr Idx blockThreadExtent = blockThreadExtentMain;
+                // Blocks per grid for Dot kernel
+                Idx const gridBlockExtent = static_cast<Idx>(dotGridBlockExtent);
+                // Vector of sums of each block
+                auto bufAccSumPerBlock = alpaka::allocBuf<DataType, Idx>(devAcc, gridBlockExtent);
+                auto bufHostSumPerBlock = alpaka::allocBuf<DataType, Idx>(devHost, gridBlockExtent);
```
chillenzer: I see why you are using a different workdiv here but it feels a bit inconsistent with the use of `getValidWorkDiv` above. Is there a rationale that under which these two can be subsumed?

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
+                [&]() {
+                    alpaka::exec<Acc>(
+                        queue,
+                        workDivTriad,
+                        TriadKernel(),
+                        bufAccInputAPtr,
+                        bufAccInputBPtr,
+                        bufAccOutputCPtr);
+                },
+                "TriadKernel");
```
chillenzer, line 461: Technically speaking, the upstream version runs [the triad-only version](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L155) of the code differently. The run-all version measures each run individually, the triad-only version measures the time for the whole loop. Admittedly, that is [fixed in the development branch](https://github.com/UoB-HPC/BabelStream/pull/186) already, so we are practically ahead of our time here.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
+        }
+        alpaka::wait(queue);
+
+    } // End of main loop which runs the kernels many times
+
+
+    // Copy arrays back to host since the execution of kernels except dot kernel finished
     alpaka::memcpy(queue, bufHostOutputC, bufAccOutputC, arraySize);
     alpaka::memcpy(queue, bufHostOutputB, bufAccInputB, arraySize);
     alpaka::memcpy(queue, bufHostOutputA, bufAccInputA, arraySize);
```
chillenzer: These get timed [int the original version](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L338).

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
+            sumErrC += std::fabs(bufHostOutputC[static_cast<Idx>(i)] - expectedC);
+            sumErrB += std::fabs(bufHostOutputB[static_cast<Idx>(i)] - expectedB);
+            sumErrA += std::fabs(bufHostOutputA[static_cast<Idx>(i)] - expectedA);
+        }
+
+        // Normalize and compare sum of the errors
+        // Use a different equality check if floating point errors exceed precision of FuzzyEqual function
+        REQUIRE(FuzzyEqual(sumErrC / static_cast<DataType>(arraySize) / expectedC, static_cast<DataType>(0.0)));
+        REQUIRE(FuzzyEqual(sumErrB / static_cast<DataType>(arraySize) / expectedB, static_cast<DataType>(0.0)));
+        REQUIRE(FuzzyEqual(sumErrA / static_cast<DataType>(arraySize) / expectedA, static_cast<DataType>(0.0)));
```
chillenzer: This is a plausible but not identical failure criterion compared to [upstream](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L545): Our version considers the relative error, upstream considers the absolute error. That can be quite a different thing depending on the precise numbers we're running. We probably wouldn't notice for sufficiently large values because any failure would likely blow up to much larger errors when iterated often enough. Also, we're not really interested in the small accumulating round-off errors but more in wrongly implemented kernels which would typically have different orders of magnitude in their errors. But in an unfortunate parameter regime, this might make a difference...

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,210 +296,328 @@ void testKernels()
...
-        metaData.setItem(BMInfoDataType::WorkDivDot, workDivDot);
-    }
+        // Verify Dot kernel with specific blocksize which is larger than 1
+        if constexpr(alpaka::accMatchesTags<TAcc, alpaka::TagGpuCudaRt, alpaka::TagGpuHipRt, alpaka::TagGpuSyclIntel>)
+        {
+            float const expectedSum = static_cast<float>(static_cast<DataType>(arraySize) * expectedA * expectedB);
 
+            //  Dot product should be identical to arraySize*initA*initB
+            //  Use a different equality check if floating point errors exceed precision of FuzzyEqual function
+            REQUIRE(FuzzyEqual((resultDot - expectedSum) / expectedSum, 0.0f));
```
chillenzer: Just FYI: In this case, taking the relative error is also done in the upstream.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -57,13 +66,14 @@ struct InitKernel
...
     template<typename TAcc, typename T>
-    ALPAKA_FN_ACC void operator()(TAcc const& acc, T* a, T* b, T* c, T initA) const
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, T* a, T* b, T* c, T initialA, T initialB) const
     {
         auto const [i] = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
-        a[i] = initA;
-        b[i] = static_cast<T>(0.0);
+        a[i] = initialA;
+        b[i] = initialB;
         c[i] = static_cast<T>(0.0);
```
chillenzer: Why is this a compile-time constant? Compilers might do funny optimisations here if we give them too much of a hint.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -57,13 +66,14 @@ struct InitKernel
...
     //! \param acc The accelerator to be executed on.
     //! \param a Pointer for vector a
-    //! \param initA the value to set all items in the vector
+    //! \param initialA the value to set all items in the vector a
+    //! \param initialB the value to set all items in the vector b
     template<typename TAcc, typename T>
-    ALPAKA_FN_ACC void operator()(TAcc const& acc, T* a, T* b, T* c, T initA) const
+    ALPAKA_FN_ACC void operator()(TAcc const& acc, T* a, T* b, T* c, T initialA, T initialB) const
     {
         auto const [i] = alpaka::getIdx<alpaka::Grid, alpaka::Threads>(acc);
```
chillenzer, line 73: Very nice trick! It never occured to me that I could just unpack the members of a vector immediately. Much nicer than an ugly `[0]` or `.x()` at the end.

chillenzer (changes requested):
Hi, thanks for your update. I've mostly checked for compliance with upstream babelstream as that was apparently a major point of discussion. But I want to first applaud you for that nice `auto [i] = getIdx...;` trick. Very nice indeed!

It's mostly small things I've found some of which we might actively decide to do differently, e.g., not measuring some timings of the infrastructure-ish calls. There was one section that GH didn't allow me to comment on, so I'll put that in here:
Technically speaking, the dot kernel does a slightly different thing by accumulating the `threadSum`s in registers and only storing them once into shared memory. [Our version](https://github.com/alpaka-group/alpaka/blob/cf19b90df511106939c383e2943f5b78bcb618b4/benchmarks/babelstream/src/babelStreamMainTest.cpp#L187) vs. [upstream version](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.cu#L256). Likely to be optimised away by the compiler because both versions leave full flexibility concerning memory ordering here but we can't be sure I believe.

Also, just for my information: Why is `tbSum` a reference in that same kernel? It very much looks like it must be dangling but if this compiles and runs correctly it apparently isn't?

benchmarks/babelstream/src/babelStreamMainTest.cpp
psychocoderHPC: Since alpaka can run on different hardware it is fine to calculate the workdiv and not use hard-coded values. 
Hardcoding per hardware/ backend is possible too but not nice.

benchmarks/babelstream/src/babelStreamMainTest.cpp
psychocoderHPC: This is allowed for a benchmark, we know a good parameter for a specific hardware and hard coded it.

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: copy kernel sets the items of C by using value of A.

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: Ok, changed, makes sense.

mehmetyusufoglu: The original code is using 10E-8 for double and float types. I used 100*std::epsilon which is much stricter for double precision, therefore i preferred relative error but absolute error calculation is also in the epsilon limit i defined, thanks.

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: There is a code repetition in the original code, both `run_triad `and `run_all` calls the triad kernel. Here, i call the same code piece for both cases.
 https://github.com/alpaka-group/alpaka/blob/7a8b2050bad28f0c6624c0413f7793d5a871049b/benchmarks/babelstream/src/main.cpp#L108

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: It is not one of babelstream kernels but ok i am implementing.

mehmetyusufoglu:
> Hi, thanks for your update. I've mostly checked for compliance with upstream babelstream as that was apparently a major point of discussion. But I want to first applaud you for that nice `auto [i] = getIdx...;` trick. Very nice indeed!
> 
> It's mostly small things I've found some of which we might actively decide to do differently, e.g., not measuring some timings of the infrastructure-ish calls. There was one section that GH didn't allow me to comment on, so I'll put that in here: Technically speaking, the dot kernel does a slightly different thing by accumulating the `threadSum`s in registers and only storing them once into shared memory. [Our version](https://github.com/alpaka-group/alpaka/blob/cf19b90df511106939c383e2943f5b78bcb618b4/benchmarks/babelstream/src/babelStreamMainTest.cpp#L187) vs. [upstream version](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.cu#L256). Likely to be optimised away by the compiler because both versions leave full flexibility concerning memory ordering here but we can't be sure I believe.
> 
> Also, just for my information: Why is `tbSum` a reference in that same kernel? It very much looks like it must be dangling but if this compiles and runs correctly it apparently isn't?

 tbSum is reference because  the function return type is `-> T&` and returns a dereferenced value `return *data`;

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: Won't make a difference i think, now changed and a new argument is used for C

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: Yes it is exactly the same.

mehmetyusufoglu:
> Hi, thanks for your update. I've mostly checked for compliance with upstream babelstream as that was apparently a major point of discussion. But I want to first applaud you for that nice `auto [i] = getIdx...;` trick. Very nice indeed!
> 
> It's mostly small things I've found some of which we might actively decide to do differently, e.g., not measuring some timings of the infrastructure-ish calls. There was one section that GH didn't allow me to comment on, so I'll put that in here: Technically speaking, the dot kernel does a slightly different thing by accumulating the `threadSum`s in registers and only storing them once into shared memory. [Our version](https://github.com/alpaka-group/alpaka/blob/cf19b90df511106939c383e2943f5b78bcb618b4/benchmarks/babelstream/src/babelStreamMainTest.cpp#L187) vs. [upstream version](https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/cuda/CUDAStream.cu#L256). Likely to be optimised away by the compiler because both versions leave full flexibility concerning memory ordering here but we can't be sure I believe.


Yes, this was the choice at the first implementation at our repo, i used directly like in the cuda implementation now. Checking the performance.

benchmarks/babelstream/src/babelStreamMainTest.cpp
chillenzer: Okay. Resolved.

benchmarks/babelstream/src/babelStreamMainTest.cpp
chillenzer: Yes, I agree that it makes moderate sense at best but it might be interesting information and it brings us closer to upstream. Your announced change is not yet found in the PR.

chillenzer: Yes, but the init kernel is now doing a different thing. And the compiler might make further unwanted optimisations.

benchmarks/babelstream/src/babelStreamMainTest.cpp
chillenzer: Okay. Resolved.

benchmarks/babelstream/src/babelStreamMainTest.cpp
chillenzer: Resolved.

benchmarks/babelstream/src/babelStreamMainTest.cpp
chillenzer: Thanks!

chillenzer: I agree that it probably won't.

chillenzer:
> tbSum is reference because the function return type is `-> T&` and returns a dereferenced value `return *data`;

Thanks for the explanation! That makes sense.

Concerning the `reduce` implementation, I had an offline discussion with @psychocoderHPC: The concept to benchmark here is any implementation of a reduction based on alpaka. In that sense, we are not required to follow the reference implementation precisely. Not hammering on shared memory with every thread is probably a worthwhile change.

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: ok, implemented. Thanks.

```
AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:2
Precision:double
DataSize(items):1048576
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       87.201          0.0002886 0.0002886 0.0002886 25.166 
 CopyKernel      79.096          0.00021211 0.00021211 0.00021211 16.777 
 DotKernel       74.74           0.00022447 0.00022447 0.00022447 16.777 
 InitKernel      87.865          0.00028641 0.00028641 0.00028641 25.166 
 MultKernel      85.107          0.00019713 0.00019713 0.00019713 16.777 
 TriadKernel     87.046          0.00028911 0.00028911 0.00028911 25.166 
```

benchmarks/babelstream/src/babelStreamMainTest.cpp
mehmetyusufoglu: Read time (actually copy time from Acc to Host for 3 arrays ) has been added to the output display as **AccToHost Memcpy Time(sec).** Thanks.

```
AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:100
Precision:double
DataSize(items):33554432
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (32768), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.570856
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       90.665          0.0088822 0.0089985 0.0089376 805.31 
 CopyKernel      89.087          0.0060264 0.0061119 0.0060773 536.87 
 DotKernel       93.055          0.0057694 0.0058486 0.0058113 536.87 
 InitKernel      84.437          0.0095374 0.0095374 0.0095374 805.31 
 MultKernel      89.35           0.0060086 0.0060852 0.0060568 536.87 
 TriadKernel     90.222          0.0089258 0.0090338 0.0089565 805.31 

```

benchmarks/babelstream/src/babelStreamCommon.hpp
mehmetyusufoglu: additional test case is added for blocksize = 1 for Dot Kernel. Hence Dot kernel is tested for CPU backend and for blocksize =1.

mehmetyusufoglu:
> @mehmetyusufoglu Can you please check if CPU is working too.

Randomness seeded to: 2905169299
Kernels: Init, Copy, Mul, Add, Triad Kernels (and Dot kernel, if acc is multi-thread per block.)


AcceleratorType:AccCpuSerial<1,unsigned int>
NumberOfRuns:2
Precision:single
DataSize(items):1048576
DeviceName:13th Gen Intel(R) Core(TM) i7-1360P
WorkDivInit :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.0107734
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       0.026882        0.46807  0.46807  0.46807  12.583 
 CopyKernel      0.019559        0.42889  0.42889  0.42889  8.3886 
 InitKernel      0.029203        0.43088  0.43088  0.43088  12.583 
 MultKernel      0.019739        0.42498  0.42498  0.42498  8.3886 
 TriadKernel     0.025445        0.49452  0.49452  0.49452  12.583 

Kernels: Init, Copy, Mul, Add, Triad Kernels (and Dot kernel, if acc is multi-thread per block.)


AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:2
Precision:single
DataSize(items):1048576
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.0135214
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       80.998          0.00015535 0.00015535 0.00015535 12.583 
 CopyKernel      69.936          0.00011995 0.00011995 0.00011995 8.3886 
 DotKernel       48.621          0.00017253 0.00017253 0.00017253 8.3886 
 InitKernel      51.814          0.00024285 0.00024285 0.00024285 12.583 
 MultKernel      76.158          0.00011015 0.00011015 0.00011015 8.3886 
 TriadKernel     81.478          0.00015443 0.00015443 0.00015443 12.583 

Kernels: Init, Copy, Mul, Add, Triad Kernels (and Dot kernel, if acc is multi-thread per block.)


AcceleratorType:AccCpuSerial<1,unsigned int>
NumberOfRuns:2
Precision:double
DataSize(items):1048576
DeviceName:13th Gen Intel(R) Core(TM) i7-1360P
WorkDivInit :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (1048576), blockThreadExtent: (1), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.0151765
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       0.059712        0.42146  0.42146  0.42146  25.166 
 CopyKernel      0.042238        0.39721  0.39721  0.39721  16.777 
 InitKernel      0.03913         0.64314  0.64314  0.64314  25.166 
 MultKernel      0.04646         0.36111  0.36111  0.36111  16.777 
 TriadKernel     0.062699        0.40138  0.40138  0.40138  25.166 

Kernels: Init, Copy, Mul, Add, Triad Kernels (and Dot kernel, if acc is multi-thread per block.)


AcceleratorType:AccGpuCudaRt<1,unsigned int>
NumberOfRuns:2
Precision:double
DataSize(items):1048576
DeviceName:NVIDIA RTX A500 Laptop GPU
WorkDivInit :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivCopy :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivMult :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivAdd  :{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivTriad:{gridBlockExtent: (1024), blockThreadExtent: (1024), threadElemExtent: (1)}
WorkDivDot  :{gridBlockExtent: (256), blockThreadExtent: (1024), threadElemExtent: (1)}
AccToHost Memcpy Time(sec):0.0173797
Kernels         Bandwidths(GB/s) MinTime(s) MaxTime(s) AvgTime(s) DataUsage(MB) 
 AddKernel       87.21           0.00028857 0.00028857 0.00028857 25.166 
 CopyKernel      80.442          0.00020856 0.00020856 0.00020856 16.777 
 DotKernel       73.262          0.000229 0.000229 0.000229 16.777 
 InitKernel      85.267          0.00029514 0.00029514 0.00029514 25.166 
 MultKernel      85.196          0.00019693 0.00019693 0.00019693 16.777 
 TriadKernel     87.512          0.00028757 0.00028757 0.00028757 25.166 

All tests passed (18 assertions in 4 test cases)

mehmetyusufoglu:
> > tbSum is reference because the function return type is `-> T&` and returns a dereferenced value `return *data`;
> 
> Thanks for the explanation! That makes sense.
> 
> Concerning the `reduce` implementation, I had an offline discussion with @psychocoderHPC: The concept to benchmark here is any implementation of a reduction based on alpaka. In that sense, we are not required to follow the reference implementation precisely. Not hammering on shared memory with every thread is probably a worthwhile change.

Ok I reverted it back. (Yes accessing shared memory at each thread many times is not needed at such case)

benchmarks/babelstream/src/babelStreamCommon.hpp
mehmetyusufoglu: Ok, now all backends are used and tested. [ `AccCpuThreads` backend is very slow, but passed the pipeline. ]

mehmetyusufoglu:
Recent update by 25th Nov: All backends are run and results are controlled for all of them. 

 [That was not preferred due to CI runtime but now it takes less than the longest runtime example which is heat-equation.]

psychocoderHPC (approved):

benchmarks/babelstream/src/babelStreamCommon.hpp
```diff
@@ -22,61 +22,112 @@ namespace
     // According to tests, 2^25 or larger values are needed for proper benchmarking:
     // ./babelstream --array-size=33554432 --number-runs=100
     // To prevent timeouts in CI, a smaller default value is used.
-    [[maybe_unused]] auto arraySizeMain = 1024 * 1024;
+    [[maybe_unused]] auto arraySizeMain = 1024 * 256;
```
psychocoderHPC, line 25: is this an intended change? The original code is using https://github.com/UoB-HPC/BabelStream/blob/2f00dfb7f8b7cfe8c53d20d5c770bccbf8673440/src/main.cpp#L57C18-L57C26

PR #2422, opened 2024-11-13
Check in CMake if the compiler supports the minimum required C++ standard

### LINKED ISSUE #2414: Add compiler check for the supported C++ standard in CMake
In a offline discussion with @franzpoeschel he told me, that there are some confusion because alpaka does not compile with C++17 anymore. The reason is because the dev branch only supports C++20 and newer but the latest release (1.2) supports C++17.

Therefore an improvement would be to check the supported C++ standard of the compiler in CMake and display a message that the compiler does not support C++20 and maybe the message also to upgrade the compiler or use alpaka 1.2.

SimeonEhrig:
fix #2414

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -57,6 +57,20 @@ function(alpaka_set_compiler_options scope type name)
     endif()
 endfunction()
 
+# Check if compiler supports required C++ standard.
+#
+# language - can be CXX, HIP or CUDA
+# min_cxx_standard - C++ standard which is the minimum requirement
+function(checkCompilerCXXSupport language min_cxx_standard)
+    string(TOUPPER "${language}" language_upper_case)
+    string(TOLOWER "${language}" language_lower_case)
+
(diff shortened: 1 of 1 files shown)


cmake/alpakaCommon.cmake
```diff
@@ -57,6 +57,20 @@ function(alpaka_set_compiler_options scope type name)
...
+#
+# language - can be CXX, HIP or CUDA
+# min_cxx_standard - C++ standard which is the minimum requirement
+function(checkCompilerCXXSupport language min_cxx_standard)
+    string(TOUPPER "${language}" language_upper_case)
+    string(TOLOWER "${language}" language_lower_case)
+
+    if(NOT "${language_lower_case}_std_${min_cxx_standard}" IN_LIST CMAKE_${language_upper_case}_COMPILE_FEATURES)
+        message(FATAL_ERROR "The ${language_upper_case} compiler does not support C++ ${min_cxx_standard}. \
+        Please upgrade your compiler our use alpaka 1.2 which supports C++17.")
```
mehmetyusufoglu:
```suggestion
        Please upgrade your compiler or use alpaka 1.2 which supports C++17.")
```

psychocoderHPC (approved):

PR #2426, opened 2024-11-21
Fix template arguments of SYCL buffer specializations

AuroraPerego:
`BufCpuSycl`, `BufFpgaSyclIntel`, and `BufGpuSyclIntel` were still templated on the platform instead of being templated on the Tag.
We are not using them anywhere in the alpaka tests, that's why this issue has not been spotted before. 
I changed `memBufTest` to use them explicitly to avoid similar issues in the future, let me know if it's ok.

### DIFF
--- a/test/unit/mem/buf/src/BufTest.cpp
+++ b/test/unit/mem/buf/src/BufTest.cpp
@@ -15,6 +15,58 @@
 #include <numeric>
 #include <type_traits>
 
+namespace buftest
+{
+    template<typename TDim, typename TDev, typename TElem, typename TIdx, typename TExtent>
+    auto allocBuf(TDev dev, TExtent extent)
+    {
+        return alpaka::allocBuf<TElem, TIdx>(dev, extent);
+    }
+
+    template<typename TDim, typename TElem, typename TIdx, typename TExtent>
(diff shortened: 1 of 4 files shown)


test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -15,6 +15,54 @@
...
+        return alpaka::allocBuf<TElem, TIdx>(dev, extent);
+    }
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+    template<typename TDim, typename TElem, typename TIdx, typename TExtent>
+    auto allocBuf(alpaka::DevCudaRt dev, TExtent extent) -> alpaka::BufCudaRt<TElem, TDim, TIdx>
+    {
+        return alpaka::allocBuf<TElem, TIdx>(dev, extent);
+    }
+#elif defined(ALPAKA_ACC_GPU_HIP_ENABLED)
```
fwyzard: I would change 
```suggestion
#endif

#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)
```

so if a test enables multiple back-ends they can all be tested.

AuroraPerego: Ok, done!

fwyzard:
Looks good to me 👍🏻

fwyzard (approved):

PR #2430, opened 2024-11-22
An example using cuBLAS library in alpaka

mehmetyusufoglu:
This example uses cuBLAS library for matrix multiplication by using allocated alpaka buffers and alpaka queue. Another example is using rocBLAS library.

Cmake file is still to be changed for CI to not fail for other backends.

``` // Get the native CUDA stream from Alpaka queue
    auto alpakaStream = alpaka::getNativeHandle(queue);

    // cuBLAS setup from alpaka stream
    cublasHandle_t cublasHandle;
    cublasCreate(&cublasHandle);
    cublasSetStream(cublasHandle, alpakaStream);

    // Perform matrix multiplication: C = A * B
    float alpha = 1.0f, beta = 0.0f; // Set beta to 0.0f to overwrite C
    cublasSgemm(
        cublasHandle,
        CUBLAS_OP_N,
        CUBLAS_OP_N, // No transpose for A and B
        M,
        N,
        K, // Dimensions: C = A * B
        &alpha,
        alpaka::getPtrNative(bufDevA), M ...
    );
    alpaka::wait(queue); // Wait for multiplication to complete```

### DIFF
--- a/example/useCuBLASInAlpaka/CMakeLists.txt
+++ b/example/useCuBLASInAlpaka/CMakeLists.txt
@@ -1,0 +1,58 @@
+#
+# Copyright 2023 Benjamin Worpitz, Jan Stephan
+# SPDX-License-Identifier: ISC
+#
+
+################################################################################
+# Required CMake version.
+
+cmake_minimum_required(VERSION 3.25)
+
+set_property(GLOBAL PROPERTY USE_FOLDERS ON)
+
+################################################################################
+# Project.
+
+set(_TARGET_NAME useCuBLASInAlpaka)
+
+project(${_TARGET_NAME} LANGUAGES CXX)
+
+# Check if the cmake variables to see if the Acc option is Cuda Only.
+if(NOT alpaka_ACC_GPU_CUDA_ONLY_MODE)
+    # Print a warning and skip target creation
+    message(WARNING "Skipping build of 'useCuBLASInAlpaka' because alpaka_ACC_GPU_CUDA_ONLY_MODE is not enabled.")
+    return()
+endif()
+
+# Add cuBLAS library
+find_package(CUDA REQUIRED)
+set(CUDA_LIBRARIES ${CUDA_LIBRARIES} cublas)
+
+#-------------------------------------------------------------------------------
+# Find alpaka.
+
+if(NOT TARGET alpaka::alpaka)
+    option(alpaka_USE_SOURCE_TREE "Use alpaka's source tree instead of an alpaka installation" OFF)
+
+    if(alpaka_USE_SOURCE_TREE)
+        # Don't build the examples recursively
+        set(alpaka_BUILD_EXAMPLES OFF)
+        add_subdirectory("${CMAKE_CURRENT_LIST_DIR}/../.." "${CMAKE_BINARY_DIR}/alpaka")
+    else()
+        find_package(alpaka REQUIRED)
+    endif()
+endif()
+
+#-------------------------------------------------------------------------------
+# Add executable.
+
+alpaka_add_executable(
+    ${_TARGET_NAME}
+    src/useCuBLASInAlpaka.cpp)
+target_link_libraries(
+    ${_TARGET_NAME}
+    PUBLIC alpaka::alpaka ${CUDA_LIBRARIES})
+
+set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER example)
+#set_target_properties(${_TARGET_NAME}  PROPERTIES CUDA_STANDARD 17)
+add_test(NAME ${_TARGET_NAME} COMMAND ${_TARGET_NAME})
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: example/useBLASInAlpaka/src/useBLASInAlpaka.cpp [hunk 1]
```diff
@@ -0,0 +1,167 @@
...
+    float alpha = 1.0f, beta = 0.0f; // Set beta to 0.0f to overwrite C
+    cublasSgemm(
+        cublasHandle,
+        CUBLAS_OP_N,
+        CUBLAS_OP_N, // No transpose for A and B
+        M,
+        N,
+        K, // Dimensions: C = A * B
+        &alpha,
+        alpaka::getPtrNative(bufDevA),
```
psychocoderHPC: better use `std::data()` instead of `alpaka::getPtrNative()`

psychocoderHPC (changes requested):

example/useBLASInAlpaka/src/useBLASInAlpaka.cpp
```diff
@@ -0,0 +1,167 @@
...
+{
+    using Dim1D = alpaka::DimInt<1>;
+
+    // Define matrix dimensions, A is MxK and B is KxN
+    Idx const M = 4; // Rows in A and C
+    Idx const N = 2; // Columns in B and C
+    Idx const K = 3; // Columns in A and rows in B
+
+    // Define device and queue
+    using Acc = alpaka::AccGpuCudaRt<Dim1D, Idx>;
```
psychocoderHPC: Could you please use the CUDA tag and derive the ACC from the tag? THis will reduce the work as soon as we refactor the accelerators.

psychocoderHPC (changes requested):

example/useBLASInAlpaka/src/useBLASInAlpaka.cpp
mehmetyusufoglu: ok, done, thanks.

example/useBLASInAlpaka/src/useBLASInAlpaka.cpp
mehmetyusufoglu: I used the standard tags like other examples, but prevented the configuration of this example at cmake if ACC_CUDA_ONLY cmake variable is not set.

mehmetyusufoglu: Anyway i used cuda tag as  you suggested. This example could have a direct main rather than using ExampleTags since only will run with single backend.

`using Acc = alpaka::TagToAcc<alpaka::TagGpuCudaRt, Dim1D, Idx>;`

example/useCuBLASInAlpaka/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+
+alpaka_add_executable(
+    ${_TARGET_NAME}
+    src/useCuBLASInAlpaka.cpp)
+target_link_libraries(
+    ${_TARGET_NAME}
+    PUBLIC alpaka::alpaka ${CUDA_LIBRARIES})
+
+set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER example)
+set_target_properties(${_TARGET_NAME}  PROPERTIES CUDA_STANDARD 14)
```
mehmetyusufoglu, line 57: @SimeonEhrig  is this the correct selection? I am not sure.

example/useCuBLASInAlpaka/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+
+alpaka_add_executable(
+    ${_TARGET_NAME}
+    src/useCuBLASInAlpaka.cpp)
+target_link_libraries(
+    ${_TARGET_NAME}
+    PUBLIC alpaka::alpaka ${CUDA_LIBRARIES})
+
+set_target_properties(${_TARGET_NAME} PROPERTIES FOLDER example)
+set_target_properties(${_TARGET_NAME}  PROPERTIES CUDA_STANDARD 17)
```
mehmetyusufoglu: @SimeonEhrig is this the correct selection? Since we use Cuda 12.0 and later CUDA_STANDARD 17 should be selected?

This property is initialized by the value of the CMAKE_CUDA_STANDARD variable:CMAKE_CUDA_STANDARD

example/useBLASInAlpaka/src/useBLASInAlpaka.cpp
SimeonEhrig: @psychocoderHPC I agree with @mehmetyusufoglu . It does not make sense to use the same template, like in the other examples. The code can be only used with the CUDA backend. Therefore we need no complicated iteration over the enabled tags.

example/useCuBLASInAlpaka/CMakeLists.txt
SimeonEhrig: The `CUDA_STANDARD` property is already set via the alpaka common cmake.

mehmetyusufoglu:
This PR is closed because the changes is added to  #2433  since 2 PR's will share the same directory in `examples` directory. Rather than being 2 separate directories in `examples`.

PR #2433, opened 2024-11-28
an example using rocBLAS and cuBLAS in alpaka code

mehmetyusufoglu:
Uses rocBLAS `rocblas_sgemm` general matrix multiplication function at an alpaka stream and with alpaka buffers. 

Comment on Dec11th: changes of #2430 is added to this PR since that example and the example here will share the same directory  under `examples`


psychocoderHPC:
IMO we should provide a subfolder in the example folder and add the rocm and cuda examples there in.

suggested name: nativeAPI

example/useBLASInAlpaka/useRocBLASInAlpaka/src/useRocBLASInAlpaka.cpp
```diff
@@ -0,0 +1,175 @@
...
+    // Associate the HIP stream with the rocBLAS handle
+    status = rocblas_set_stream(rocblasHandle, rocStream);
+    if(status != rocblas_status_success)
+    {
+        std::cerr << "rocblas_set_stream failed with status: " << status << std::endl;
+        rocblas_destroy_handle(rocblasHandle);
+        return EXIT_FAILURE;
+    }
+    // Perform matrix multiplication: C = alpha * A * B + beta * C
+    float alpha = 1.0f, beta = 0.0f;
```
psychocoderHPC, line 104: please choose a value different to 1.0 and 0.0 for the Constance.

psychocoderHPC (changes requested):

example/useBLASInAlpaka/useRocBLASInAlpaka/src/useRocBLASInAlpaka.cpp
mehmetyusufoglu: I have implemented matrix multiplication, therefore beta is 0.0. Either i can add a comment saying "beta is 0.0 because this is pure multiplication" or chose the `beta` nonzero and say that this is a linear operation (not multiplication) inside the comments and outputs.

psychocoderHPC: Ahh ok, in that case the comment should be adjusted and say that alpaha is 1 and beta zero because you a pure matrix multiplication is performed.

mehmetyusufoglu: done, thanks.

psychocoderHPC (approved):

PR #2438, opened 2024-12-11
Windows CI: use preinstalled CMake instead installing specific version

SimeonEhrig:
Uninstalling the pre installed CMake version is not possible.

https://github.com/actions/runner-images/blob/main/images/windows/Windows2022-Readme.md#tools

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -115,7 +115,7 @@ jobs:
           container: ubuntu:20.04
         - name: windows_cl-2022_debug_analysis
           os: windows-2022
-          env: {ALPAKA_CI_CXX: cl.exe,  ALPAKA_CI_CL_VER: 2022,                                   CMAKE_BUILD_TYPE: Debug,   ALPAKA_BOOST_VERSION: 1.78.0, ALPAKA_CI_CMAKE_VER: 3.25.3,                     ALPAKA_CI_ANALYSIS: ON, alpaka_DEBUG: 2}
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -115,7 +115,7 @@ jobs:
           container: ubuntu:20.04
         - name: windows_cl-2022_debug_analysis
           os: windows-2022
-          env: {ALPAKA_CI_CXX: cl.exe,  ALPAKA_CI_CL_VER: 2022,                                   CMAKE_BUILD_TYPE: Debug,   ALPAKA_BOOST_VERSION: 1.78.0, ALPAKA_CI_CMAKE_VER: 3.25.3,                     ALPAKA_CI_ANALYSIS: ON, alpaka_DEBUG: 2}
+          env: {ALPAKA_CI_CXX: cl.exe,  ALPAKA_CI_CL_VER: 2022,                                   CMAKE_BUILD_TYPE: Debug,   ALPAKA_BOOST_VERSION: 1.78.0, ALPAKA_CI_CMAKE_VER: 3,                          ALPAKA_CI_ANALYSIS: ON, alpaka_DEBUG: 2}
```
mehmetyusufoglu, line 118: Would "3" work for any 3.x installed or it means directly 3.0 ?

psychocoderHPC: I assume the package manager we use to install cmake does not supports minor versions and will use something modern.
Lets wait for a response from @SimeonEhrig

SimeonEhrig: `3` means only the major version is specified. At the moment CMake 3.31 is preinstalled. If MS is updating the image we get a another version of CMake 3 (probably something newer than 3.31).
I cannot uninstall the preinstalled version. Therefore I cannot decide what we use and I thought CMake 4 will be not released soon the the version `3` shows that we only get CMake 3.

psychocoderHPC (approved):

SimeonEhrig:
I opened an issue, that the problem will be fixed: https://github.com/actions/runner-images/issues/11174
But to be honest, I don't believe that it will be fixed. MS never fixed one of my bug reports :-(

PR #2439, opened 2024-12-11
add CI job, which verifies that the README.md contains the generated compiler support table

SimeonEhrig:
The script `generate_supported_compilers.py` generates the compiler support Markdown table for the README.md from a config file. The script was introduced because the markdown table is huge and hard to modify.

Therefore the CI job checks if the README.md contains the markdown table, which is generated from the script. This covers the case, if somebody is changing directly the `README.md` instead modifying the `supported_compilers.json`, rendering a new table and copy the output in the `README.md`.

### DIFF
--- a/.github/workflows/verify-readme.yml
+++ b/.github/workflows/verify-readme.yml
@@ -1,0 +1,14 @@
+# Copyright 2023 Bernhard Manfred Gruber
+# SPDX-License-Identifier: MPL-2.0
+
+name: Check compiler support table in README.md
+on: [push, pull_request]
+
+jobs:
+  single-header:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+    - name: verify that the compiler support table is up to date
+      run: |
+        ./script/readme_generator/generate_supported_compilers.py --verify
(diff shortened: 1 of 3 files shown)


SimeonEhrig:
@AuroraPerego Can you please check, if the error message is helpful: https://github.com/alpaka-group/alpaka/actions/runs/12274502625/job/34247545316?pr=2439

AuroraPerego:
> @AuroraPerego Can you please check, if the error message is helpful: https://github.com/alpaka-group/alpaka/actions/runs/12274502625/job/34247545316?pr=2439

yes, it does not tell directly which number in that line is the wrong one but the instructions on how to fix it are clear, I think it's fine, thanks!

SimeonEhrig:
> > @AuroraPerego Can you please check, if the error message is helpful: https://github.com/alpaka-group/alpaka/actions/runs/12274502625/job/34247545316?pr=2439
> 
> yes, it does not tell directly which number in that line is the wrong one but the instructions on how to fix it are clear, I think it's fine, thanks!

Thant's good :-) 
I can only display, which line number in the generated table was not found in the README.md. The problem is, I have no anchor point in README.md. Therefore I need to find the table. If line 0 of the table in the README.md does not match the line 0 of the generated table, I can only assume where the is table starting.
In the end you have the same problem like in git. Sometimes your diffs looks pretty strange because git can also only assume what you actual changed.

But nevertheless, in the end you will replace the whole table and normally you know what you changed in the README.md. Therefore the line number should be not necessary.

psychocoderHPC (approved):

PR #2456, opened 2025-01-23
Fix `parallelLoopPatterns`  for the Intel FPGA SYCL back-end

### LINKED ISSUE #2450: `parallelLoopPatterns` fails on the Intel FPGA emulator SYCL backend
The test tries to run kernels with 128 blocks and 64×2²⁰ (67108864) threads.

This results in 8×2³⁰ (8796093022208) threads, which causes at least two problems:
  - the compiler assumes that kernel-related queries (and possibly indices) fit in an `int32_t` variable, and so are limited to `INT_MAX`;
  - the examples use `uint32_t` indices, which overflow for values over 2³²-1.

The assumption of the compiler can be relaxed with the [`-fno-sycl-id-queries-fit-in-int` flag](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2025-0/fsycl-id-queries-fit-in-int.html), but I don't think we should do that by default in alpaka.

The examples could be fixed to use `uint64_t` (or plain `int64_t`) indices.

Or, the example could clamp the number of blocks or the number of threads per block so the product fits in `INT_MAX`.

fwyzard:
Extend `parallelLoopPatterns` to use 64-bit indices to support more than 2³² elements.

Add a dedicated fix for SYCL back-ends, that assume that kernel-related queries (and possibly indices) are limited to `INT_MAX`, unless the `-fno-sycl-id-queries-fit-in-int` compiler flag is used.

### DIFF
--- a/example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
+++ b/example/parallelLoopPatterns/src/parallelLoopPatterns.cpp
@@ -171,7 +171,12 @@ void gridStridedLoop(TDev& dev, TQueue& queue, TBufAcc& bufAcc)
     // and fixed number of blocks tied to hardware parameters.
     // alpaka element layer is not used in this pattern.
     uint64_t const threadsPerBlock = maxThreadsPerBlock;
+#if defined(__SYCL_ID_QUERIES_FIT_IN_INT__) && __SYCL_ID_QUERIES_FIT_IN_INT__
+    // the compiler assumes that SYCL query parameters and maximum grid*block size fits in INT_MAX (2**31-1)
+    uint64_t const blocksPerGrid = std::min(deviceProperties.m_multiProcessorCount, INT_MAX / threadsPerBlock);
+#else
     uint64_t const blocksPerGrid = deviceProperties.m_multiProcessorCount;
+#endif
     uint64_t const elementsPerThread = 1u;
     auto const workDiv = WorkDiv{blocksPerGrid, threadsPerBlock, elementsPerThread};
     std::cout << "\nGrid strided loop processing - fixed number of threads and blocks:\n";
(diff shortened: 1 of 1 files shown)


fwyzard:
I could avoid the new `Vec` constructor if somebody manages to update `parallelLoopPatterns` to build without it - though I think it may be more generally useful to have it.

fwyzard:
We could also make the `__SYCL_ID_QUERIES_FIT_IN_INT__` check more general, and introduce a new property in `include/alpaka/acc/AccDevProps.hpp` with the maximum number of threads per grid.

fwyzard:
Fixes #2450.

include/alpaka/vec/Vec.hpp
```diff
@@ -79,6 +79,16 @@ namespace alpaka
         }
 #endif
 
+        template<
+            typename UDim,
+            typename UVal,
+            typename
+            = std::enable_if_t<UDim::value == TDim::value && std::is_nothrow_convertible_v<std::decay_t<UVal>, TVal>>>
```
psychocoderHPC: I am not sure about this. In alpaka we avoid everywhere implicit casts (if we have it it is IMO a bug).
This constructor opens implicit casts and we are able to convert 64bit values to 32bit values.
This could result in hidden issues due to silent loss of precision.

If I am wrong please correct me!

psychocoderHPC (changes requested):
I request changes but currently do not expect that this is changed. I would like to talk about possible side effects if it is correct what I wrote.

include/alpaka/vec/Vec.hpp
fwyzard: Mhm, yes, this allows an implicit cast from `Vec<Dim3, uint64_t>` to `Vec<Dim3, uint32_t>`, in the same way that one can implicitly cast from `uint64_t` to `uint32_t`.

In genera I'm not fond of all the explicit casts that we need to use with alpaka, but if you don't like to relax this constraint, ok.

I will have to figure out how to fix the test without this.

fwyzard: OK, it wasn't complicated :)

fwyzard: By the way, maybe we could still add a conversion constructor, with some requirements:
  - if the destination type is larger than the source type (_e.g._ from `int32_t` to `int64_t`), there is not runtime check
  - otherwise, we explicitly check at runtime that the values can be stored in the destination type without loss of precision

psychocoderHPC:
@fwyzard could you please push again to trigger the CI. In case someone is assigned the hzdr bot is crashing an ignoring the job.
Before you was assigned and therefore the job not reported the status.

fwyzard:
> In case someone is assigned the hzdr bot is crashing an ignoring the job.

Ah !

psychocoderHPC (approved):

PR #2459, opened 2025-01-24
Fix babelstream for SYCL backends

fwyzard:
Rely on `alpaka::isMultiThreadAcc` instead of hand-picking the accelerator types

### DIFF
--- a/benchmarks/babelstream/src/babelStreamMainTest.cpp
+++ b/benchmarks/babelstream/src/babelStreamMainTest.cpp
@@ -176,6 +176,14 @@ struct DotKernel
     template<typename TAcc, typename T>
     ALPAKA_FN_ACC void operator()(TAcc const& acc, T const* a, T const* b, T* sum, alpaka::Idx<TAcc> arraySize) const
     {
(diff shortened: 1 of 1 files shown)


mehmetyusufoglu:
At first alpaka-babelstream version; Dot Kernel was not being run for CPU backends, it was not necessary for the Benchmark. Then in the second version CPU backend cases added for consistency. They can completely be removed i think. Creating "shared memory" at CPU is not meaningful.

fwyzard:
#### [macos_xcode-15.4_release](https://github.com/alpaka-group/alpaka/actions/runs/12949886112/job/36133750834?pr=2459#logs)
> libc++abi: terminating due to uncaught exception of type std::__1::system_error: mutex lock failed: Invalid argument

I don't have a way to reproduce and debug this.
The closest I can try is to use clang++ 15 with libc++, and that worked for me.

mehmetyusufoglu:
@fwyzard  should i create a separate PR for the babelstream including this change?

fwyzard:
Superseded by #2467 .

PR #2460, opened 2025-01-26
Debug #2459: Fix babelstream for SYCL backends

fwyzard:


mehmetyusufoglu:
i will try to run failing set-up from docker.

mehmetyusufoglu:
> i will try to run failing set-up from docker.

Since MacOsx it is not possible to run it using Docker.

PR #2462, opened 2025-01-26
[1.3] Gitlab CI: enable test for the SYCL CPU backend

fwyzard:


fwyzard:
> /builds/hzdr/crp/alpaka/test/unit/vec/src/VecTest.cpp:479:1: error: passing no argument for the '...' parameter of a variadic macro is a C++20 extension [-Werror,-Wc++20-extensions]

Mhm, looks like we cannot enable this for the 1.3.0 branch then.

SimeonEhrig:
> > /builds/hzdr/crp/alpaka/test/unit/vec/src/VecTest.cpp:479:1: error: passing no argument for the '...' parameter of a variadic macro is a C++20 extension [-Werror,-Wc++20-extensions]
> 
> Mhm, looks like we cannot enable this for the 1.3.0 branch then.

That's strange. Before we could compile it with ICPX 2024.2 and C++17. If I'm not wrong, you only changed the OneAPI version and did not install the Open-CL Runtime anymore.

SimeonEhrig:
@fwyzard I spent some time with the issue. The update to OneAPI 2025 causes the problem. The new compiler throws this warning, which becomes an error because of our build configuration. There is already a catch2 issue for it: https://github.com/catchorg/Catch2/issues/2910
So ether we wait for the fix and update or we suppress the warning for targets using catch2.

fwyzard:
> So ether we wait for the fix and update or we suppress the warning for targets using catch2.

I see.

I would propose to suppress the warning, until the fix is merged in CMake or if we use a version without the fix.

SimeonEhrig:
> I would propose to suppress the warning, until the fix is merged in CMake or if we use a version without the fix.

I agree. It's working with warning and the warning will be gone, when we move to C++20. I will work on a CMake fix for our alpaka CMake.

SimeonEhrig:
@fwyzard Can you please apply the patch on your branch. This suppress the warning and should fix the problem.
[0001-suppress-c-20-extension-warning-if-ICPX-2025-and-C-1.zip](https://github.com/user-attachments/files/18584816/0001-suppress-c-20-extension-warning-if-ICPX-2025-and-C-1.zip)

fwyzard:
@SimeonEhrig thank you for the patch.

Do you know if clang 19 would have the same issue ?

SimeonEhrig:
> @SimeonEhrig thank you for the patch.
> 
> Do you know if clang 19 would have the same issue ?

~I think so. I can test it later, if you don't have a clang 19.~
Clang 19 has the same problem.

SimeonEhrig:
@fwyzard I extended the patch for clang 19. You need to revert the patch.
[0001-suppress-c-20-extension-warning-if-ICPX-2025-and-C-1.zip](https://github.com/user-attachments/files/18587087/0001-suppress-c-20-extension-warning-if-ICPX-2025-and-C-1.zip)

fwyzard:
> @fwyzard I extended the patch for clang 19. You need to revert the patch. [0001-suppress-c-20-extension-warning-if-ICPX-2025-and-C-1.zip](https://github.com/user-attachments/files/18587087/0001-suppress-c-20-extension-warning-if-ICPX-2025-and-C-1.zip)

@SimeonEhrig thank you for the update.

SimeonEhrig:
@fwyzard Now trigger the wrap-test bug also here: https://github.com/alpaka-group/alpaka/pull/2464#issuecomment-2620822173

fwyzard:
Yes... unfortunately that looks to be a real issue: https://github.com/alpaka-group/alpaka/pull/2464#issuecomment-2621667902 .

SimeonEhrig (approved):

SimeonEhrig:
I merge after #2469 is merged.

fwyzard:
(rebased to fix the conflicts)

PR #2464, opened 2025-01-27
Debug #2459: Fix babelstream for SYCL backends

fwyzard:
This PR includes
  - #2142
  - #2459

to check the cumulative impact of both PRs.

PR #2467, opened 2025-01-29
Limit babelstream DotKernel block size for multithreaded CPUs

mehmetyusufoglu:
Although CPU backends are not necessary for the DotKernel of the Benchmark, we were running them. New SYCL backend CI tests failed, therefore this change is made. 

See:
https://github.com/alpaka-group/alpaka/pull/2462
https://github.com/alpaka-group/alpaka/pull/2464

### DIFF
--- a/benchmarks/babelstream/src/babelStreamMainTest.cpp
+++ b/benchmarks/babelstream/src/babelStreamMainTest.cpp
@@ -325,14 +325,38 @@ void testKernels()
     // Lambda to create and return work division for dot kernel
     auto getWorkDivForDotKernel = [&]<typename AccType>() -> alpaka::WorkDivMembers<Dim, Idx>
     {
-        // Use babelstream standard work division for multi-threaded backends
-        if constexpr(alpaka::
-                         accMatchesTags<AccType, alpaka::TagGpuCudaRt, alpaka::TagGpuHipRt, alpaka::TagGpuSyclIntel>)
+        // Use babelstream benchmark standard work division for multi-threaded backends
+        // For GPUs the benchmark expects a fixed workdiv: {256,1024,1}
+        // For multi-threaded CPUs, which are not part of the benchmark, the code below will limit the blocksize with
+        // the limit of the backend.
+        if constexpr(alpaka::isMultiThreadAcc<AccType>)
         {
-            return alpaka::WorkDivMembers{
+            auto const kernelFunctionAttributes = alpaka::getFunctionAttributes<Acc>(
+                devAcc,
+                DotKernel(),
+                bufAccInputAPtr,
+                bufAccInputBPtr,
+                bufAccOutputCPtr, // this is used here a kind of dummy
+                static_cast<alpaka::Idx<AccType>>(arraySize));
+
+            // Get the maxThreadPerBlock
+            auto const maxThreadsPerBlock = kernelFunctionAttributes.maxThreadsPerBlock;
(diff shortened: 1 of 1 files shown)


benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -325,14 +325,42 @@ void testKernels()
     // Lambda to create and return work division for dot kernel
     auto getWorkDivForDotKernel = [&]<typename AccType>() -> alpaka::WorkDivMembers<Dim, Idx>
     {
-        // Use babelstream standard work division for multi-threaded backends
-        if constexpr(alpaka::
-                         accMatchesTags<AccType, alpaka::TagGpuCudaRt, alpaka::TagGpuHipRt, alpaka::TagGpuSyclIntel>)
+        // Use babelstream benchmark standard work division for multi-threaded backends
+        // For Gpu's the benchmark expects a fixed workdiv: {256,1024,1}
```
fwyzard:
```suggestion
        // For GPUs the benchmark expects a fixed workdiv: {256,1024,1}
```

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -325,14 +325,42 @@ void testKernels()
...
+                bufAccInputAPtr,
+                bufAccInputBPtr,
+                bufAccOutputCPtr, // this is used here a kind of dummy
+                static_cast<alpaka::Idx<AccType>>(arraySize));
+
+            // Get the maxThreadPerBlock
+            auto const maxThreadsPerBlock = kernelFunctionAttributes.maxThreadsPerBlock;
+            // Threads per block is 1024 for benchmark, if the system does not allow use the max value
+            auto threadsPerBlock
+                = maxThreadsPerBlock < blockThreadExtentMain ? maxThreadsPerBlock : blockThreadExtentMain;
```
fwyzard:
```suggestion
            auto threadsPerBlock = std::min(maxThreadsPerBlock, blockThreadExtentMain);
```

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -325,14 +325,42 @@ void testKernels()
...
+
+            // Reduce operation at dot-kernel needs even block size
+            if(threadsPerBlock != 1 && threadsPerBlock % 2 != 0)
+            {
+                threadsPerBlock -= 1;
+            }
```
fwyzard: given the comment that

> dot-kernel needs even block size

I'm wondering if `threadsPerBlock == 1` is a valid value ?

fwyzard: If `threadsPerBlock == 1` is not valid and should be rounded to `2`, this could be simplified as
```suggestion
            // Reduce operation at dot-kernel needs even block size
            threadsPerBlock += (threadsPerBlock % 2);
```
or
```suggestion
            // Reduce operation at dot-kernel needs even block size
            threadsPerBlock = (threadsPerBlock + 1) / 2 * 2;
```
whatever you like better.

fwyzard (changes requested):

fwyzard:
@mehmetyusufoglu can you change the title and make this a "real" pull request ?
From my point of view it can be merged by itself.

fwyzard (approved):

fwyzard:
@psychocoderHPC the changes are fine for you ?

PR #2468, opened 2025-01-29
Disable test not supported by SYCL

fwyzard:
According to the documentation of the **SYCL Group functions** [1](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:group-functions) and **SYCL Group algorithms library** [2](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:algorithms), calling `sycl::all_of_group` with some threads having exited the kernel is _undefined behaviour_. This was confirmed during a support meeting with Intel.

oneAPI has an extension to support non-uniform (sub)groups: [sycl_ext_oneapi_non_uniform_groups](https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_oneapi_non_uniform_groups.asciidoc).
The implementation of the SYCL warp functions should be extended using it; until then, we disable the failing test when using a SYCL back-end and instead print an informational message.

### DIFF
--- a/test/unit/warp/src/Activemask.cpp
+++ b/test/unit/warp/src/Activemask.cpp
@@ -65,6 +65,9 @@ struct alpaka::trait::WarpSize<ActivemaskMultipleThreadWarpTestKernel<TWarpSize>
 TEMPLATE_LIST_TEST_CASE("activemask", "[warp]", alpaka::test::TestAccs)
 {
     using Acc = TestType;
+    using Dim = alpaka::Dim<Acc>;
+    using Idx = alpaka::Idx<Acc>;
+
     if constexpr(alpaka::accMatchesTags<
                      Acc,
                      alpaka::TagCpuSycl,
@@ -72,14 +75,10 @@ TEMPLATE_LIST_TEST_CASE("activemask", "[warp]", alpaka::test::TestAccs)
                      alpaka::TagFpgaSyclIntel,
                      alpaka::TagGenericSycl>)
     {
-        std::cout << "Test disabled for SYCL\n";
-        return;
+        WARN("Test disabled for SYCL");
     }
     else
     {
-        using Dim = alpaka::Dim<Acc>;
-        using Idx = alpaka::Idx<Acc>;
-
         auto const platform = alpaka::Platform<Acc>{};
         auto const dev = alpaka::getDevByIdx(platform, 0);
         auto const warpExtents = alpaka::getWarpSizes(dev);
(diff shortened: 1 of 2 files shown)


test/unit/warp/src/Activemask.cpp
fwyzard, line 1: Note that most of these changes are simply due to a different indentation.

test/unit/warp/src/All.cpp
fwyzard, line 1: Note that most of these changes are simply due to a different indentation.

test/unit/warp/src/All.cpp
```diff
@@ -70,46 +70,59 @@ TEMPLATE_LIST_TEST_CASE("all", "[warp]", alpaka::test::TestAccs)
...
+                     alpaka::TagGenericSycl>)
     {
-        auto const scalar = Dim::value == 0 || warpExtent == 1;
-        if(scalar)
-        {
-            alpaka::test::KernelExecutionFixture<Acc> fixture(alpaka::Vec<Dim, Idx>::all(4));
-            REQUIRE(fixture(AllSingleThreadWarpTestKernel{}));
-        }
-        else
+        std::cout << "Test disabled for SYCL\n";
```
fwyzard: Would it be better to use
```suggestion
        WARN("Test disabled for SYCL");
```
instead ?

It prints a much bigger message:
```
-------------------------------------------------------------------------------
all - alpaka::test::TestAccs - 2
-------------------------------------------------------------------------------
/home/fwyzard/src/alpaka-group/alpaka/test/unit/warp/src/All.cpp:67
...............................................................................

/home/fwyzard/src/alpaka-group/alpaka/test/unit/warp/src/All.cpp:80: warning:
  Test disabled for SYCL

```
instead of simply
```
Test disabled for SYCL
```

SimeonEhrig: `WARN` is better because it shows, which test is disabled. Maybe in a future PR, we can enable that warnings will be displayed if `ctest` is executed :thinking: At the moment, if `ctest` is executed, you will not see, that the test is skipped.

SimeonEhrig:
@fwyzard Change the output function to `WARN` and the PR is ready for merging. Did you open an issue about fixing the test?

test/unit/warp/src/All.cpp
fwyzard: OK, I'll make the change.

fwyzard:
> @fwyzard Change the output function to `WARN` and the PR is ready for merging. 

Done

> Did you open an issue about fixing the test?

No, I haven't.

SimeonEhrig:
> No, I haven't.

If you don't want to solve in near time, it would be good if you open one with the description of this PR. Personally I prefer open issues for existing problem than looking in comments of closed pull requests.

SimeonEhrig (approved):

fwyzard:
> it would be good if you open one with the description of this PR

#2470

PR #2471, opened 2025-01-30
Gitlab CI: enable the SYCL FPGA emulation backend

fwyzard:
Install the oneAPI FPGA support libraries and enable the FPGA back-end in emulator mode.

### DIFF
--- a/script/job_generator/job_modifier.py
+++ b/script/job_generator/job_modifier.py
@@ -152,9 +152,14 @@ def add_sycl_fpga_jobs(job_matrix: List[Dict[str, Tuple[str, str]]]) -> List[Dic
 
     for job in job_matrix:
         if ALPAKA_ACC_SYCL_ENABLE in job:
-            for enabled_device in (SYCL_CPU, SYCL_FPGA):
+            extended_job = job.copy()
+            extended_job[SYCL_DEVICE] = (SYCL_CPU, ON)
+            extended_job_matrix.append(extended_job)
+            # TODO: a bug in the blockSharedSharingTest disallow testing the
+            # SYCL FPGA backend in debug mode
+            if job[BUILD_TYPE][VERSION] == CMAKE_RELEASE:
                 extended_job = job.copy()
-                extended_job[SYCL_DEVICE] = (enabled_device, ON)
+                extended_job[SYCL_DEVICE] = (SYCL_FPGA, ON)
                 extended_job_matrix.append(extended_job)
         else:
             job[SYCL_DEVICE] = ("", "")


script/before_install.sh
```diff
@@ -115,6 +115,10 @@ if [ "${alpaka_ACC_SYCL_ENABLE}" == "OFF" ]; then
     echo_yellow "<DEFAULT: SYCL environment variables for disabled backend>"
     export alpaka_SYCL_ONEAPI_CPU=${alpaka_SYCL_ONEAPI_CPU:=""}
     export alpaka_SYCL_ONEAPI_CPU_ISA=${alpaka_SYCL_ONEAPI_CPU_ISA:=""}
+    export alpaka_SYCL_ONEAPI_FPGA=${alpaka_SYCL_ONEAPI_FPGA:=""}
```
SimeonEhrig: Can you please add a `TODO` that this should be split in a separate Intel SYCL CPU and FPGA job in future with the next commit.

script/job_generator/generate_job_yaml.py
```diff
@@ -391,6 +391,10 @@ def job_variables(job: Dict[str, Tuple[str, str]]) -> Dict[str, str]:
         variables["ALPAKA_CI_ONEAPI_VERSION"] = job[DEVICE_COMPILER][VERSION]
         variables["alpaka_SYCL_ONEAPI_CPU"] = "ON"
         variables["alpaka_SYCL_ONEAPI_CPU_ISA"] = "avx2"
+        variables["alpaka_SYCL_ONEAPI_FPGA"] = "ON"
```
SimeonEhrig: Can you please add a `TODO` that this should be split in a separate Intel SYCL CPU and FPGA job in future with the next commit.

script/before_install.sh
fwyzard: Sure, I can do that, but do you prefer to merge it like this with a TODO, or to actually split it ?

SimeonEhrig: Like here: https://github.com/alpaka-group/alpaka/blob/eca2cd543ac0753b755be94f3f4cfbbac4f8591e/script/job_generator/generate_job_yaml.py#L287-L290

The name is optional in general but in this case you can use my one. I use grep to find `TODO`s.

fwyzard: Sure, I understand that :-)

What I meant to ask is: are we OK to merge this PR as-is (plus the TODO comment), given that it breaks the SYCL CPU workflows ?

SimeonEhrig: No, please not. I think you missed my question in the mattermost chat yesterday. It is not possible to compile the CPU and FPGA backend in you job. If this is true, I will start to open an new PR and extend the job generator that it can generate separate jobs for SYCL CPU and FPGA backend.

fwyzard: I agree we should not merge this PR as-is, without splitting the CPU and FPGA workflows (or finding another solution).

Then I don't understand the reason for the TODO: by the time we _do_ merge it, the split will have been done.

SimeonEhrig: > Then I don't understand the reason for the TODO: by the time we do merge it, the split will have been done.

I was not sure, what is the way to go. You did not directly answer on my question on mattermost and you wrote that you asked Intel. So I was not sure how to continue and I thought better document it, before forget it.

SimeonEhrig:
@fwyzard I implemented generating different jobs for targeting SYCL CPU and SYCL GPU in this PR #2474

fwyzard:
Merged as part of #2474.

PR #2472, opened 2025-02-01
Implement a c++20 constexpr type demangler

### LINKED ISSUE #2405: remove boost demangle

fwyzard:
Replace the run-time type demangler based on Boost with a constexpr version based on embedding the type name in the compiler generated function name.

Based on https://www.reddit.com/r/cpp/comments/lfi6jt/finally_a_possibly_portable_way_to_convert_types/ and https://rodusek.com/posts/2021/03/09/getting-an-unmangled-type-name-at-compile-time/ .


include/alpaka/core/DemangleTypeNames.hpp
```diff
@@ -1,23 +1,90 @@
...
-#    pragma clang diagnostic ignored "-Wmissing-variable-declarations"
-#endif
+
+    // constexpr template string_view that contains the demangled type name.
     template<typename T>
-    inline const std::string demangled = boost::core::demangle(typeid(T).name());
-#if BOOST_COMP_CLANG
-#    pragma clang diagnostic pop
-#endif
+    inline constexpr std::string demangled = std::string(demangle<T>());
```
fwyzard: This should be rewritten using an `std::string_view` rather than an `std::string` - but it requires changes in many other places, I'll take care in a second moment.

fwyzard:
Looks like event with `-std=c++20` many of the compilers we support did not provide `std::format` :-(

fwyzard:
I really _hate_ clang pointless warnings about perfectly valid C++ being treated as errors:
> ```
> include/alpaka/acc/AccCpuSerial.hpp:162:42: error: declaration requires an exit-time destructor [-Werror,-Wexit-time-destructors]
> ```

fwyzard:
> Looks like event with `-std=c++20` many of the compilers we support did not provide `std::format` :-(

Same for `std::source_location`:

> 2025-02-02T12:02:57.1072996Z /__w/alpaka/alpaka/include/alpaka/core/DemangleTypeNames.hpp:14:10: fatal error: 'source_location' file not found
> 2025-02-02T12:02:57.1073952Z #include <source_location>
> 2025-02-02T12:02:57.1074334Z          ^~~~~~~~~~~~~~~~~

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -162,7 +165,22 @@ namespace alpaka
...
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+                using namespace std::literals;
+                static std::string const accName =
+#    ifdef __cpp_lib_format
+                    std::format("AccCpuOmp2Blocks<{},{}>", TDim::value, core::demangled<TIdx>);
+#    else
+                    "AccCpuOmp2Blocks<"s + std::to_string(TDim::value) + ","s + std::string(core::demangled<TIdx>)
```
SimeonEhrig, line 177: Is `std::stringstream` maybe a better solution?

fwyzard: I can give it a try.

SimeonEhrig: I don't know, if `std::stringstream` has a performance disadvantage compare to `std::format`. I need to check it. But if not, I would suggest it.

SimeonEhrig: I was a little bit searching. `std::format` should be faster by factors and concatenating `std::string` should be faster than `std::stringstream`. The benefit of `std::stringstream` is the robustness for different data types.  

- https://mariusbancila.ro/blog/2023/09/12/formatting-text-in-c-the-old-and-the-new-ways/
- https://stackoverflow.com/questions/14741144/stdstringstream-vs-stdstring-for-concatenating-many-strings
- https://lemire.me/blog/2023/10/19/for-processing-strings-streams-in-c-can-be-slow/

fwyzard: This should work:
```c++
                static std::string const accName =
#    ifdef __cpp_lib_format
                    std::format("AccCpuOmp2Blocks<{},{}>", TDim::value, core::demangled<TIdx>);
#    else
                    (std::stringstream() << "AccCpuOmp2Blocks<" << TDim::value << "," << core::demangled<TIdx> << ">").str();
#    endif
```

Do you prefer to keep the optional use of `std::format`, or go with `std::stringstream` in any case ?

SimeonEhrig: Who requires `pop warning` pagma? Is it for `std::format`? 
I think we should keep the `std::format` because it combines simplicity with performance and a certain time, we can remove the `std::stringstream` solution.

fwyzard: > Who requires `pop warning` pragma? Is it for `std::format`?

No, it's the `static` data member that I introduced to avoid regenerating the name each time the method is called.

fwyzard: > I think we should keep the `std::format` because it combines simplicity with performance and a certain time, we can remove the `std::stringstream` solution.

OK, will do it like that.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
```diff
@@ -162,7 +166,21 @@ namespace alpaka
         {
             ALPAKA_FN_HOST static auto getAccName() -> std::string
```
SimeonEhrig, line 166: The PR headline mentioned a `constexpr` type demangler. Is it possible to make the function `constexpr` or `consteval`? In this case the performance of `std::format` and `std::stringstream` does not matter.

fwyzard: I didn't find a way.

The performance doesn't matter much anyway, because the name is computed only once, the first time it's used.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
fwyzard: By the way, another possibility would be to use the external [{fmt}](https://github.com/fmtlib/fmt) library for the compilers that do not support `std::format` yet.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
SimeonEhrig: At the moment I play a little bit around with it but it is not easy that compiler really optimize.
But if the performance does not matter, we simply keep the simple implementation and optimize it, if become an bottleneck.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
SimeonEhrig: I also thought about this idea, but than the PR does not make sense anymore, because the PR should remove a dependency and not replace it.

fwyzard: I don't mind having boost as a dependency, so for me improvement is the `constexpr` functionality.

However, unlike boost I think {fmt} is header only, and we could include it directly in alpaka as a third party library like catch and mdspan.

Anyway, it's just an idea, and in case we can follow up in a later PR.

SimeonEhrig: > I don't mind having boost as a dependency, so for me improvement is the `constexpr` functionality.
> 
Ah okay. I thought this PR is pulling out a part of this PR #2409. But this was a misunderstanding from my side (to much different stuff in the morning ;-) ).

> However, unlike boost I think {fmt} is header only, and we could include it directly in alpaka as a third party library like catch and mdspan.
> 
> Anyway, it's just an idea, and in case we can follow up in a later PR.

Fair enough. We need only to support one specific version, which we can ship via sub directory or via CMake. We can discuss this tomorrow in the VC.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
SimeonEhrig: Okay, I found a major problem, if we want to make it `constexpr`. C++23 is required for static variables in `constexpr` functions.

fwyzard: Yes, it's the same issue I had to work around [here](https://github.com/fwyzard/alpaka/blob/fe19cdadf930d0adbd50448a99af08ae6fcba23c/include/alpaka/core/DemangleTypeNames.hpp#L75-L78):
```c++
    // Store the demangled type name as a null-terminated array of bytes.
    // Note: this could be a function-static constexpr variable in c++23.
    template<typename T>
    inline constexpr auto storage = demangleAsArray<T>();
```

Maybe we can use a `consteval` function and store it as a `static constexpr` data member.

psychocoderHPC:
The fallback source location implementation for demange equal to https://github.com/alpaka-group/alpaka/issues/2405?

include/alpaka/acc/AccCpuOmp2Blocks.hpp
SimeonEhrig: During the VC today we talked about the subject. It is not performance critical. Therefore we can keep it as it is.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
SimeonEhrig: During the VC we discussed, if you should use the fmt library. @psychocoderHPC is against it, because the advantage of using it in a debug function is not big enough for the extra work caused by maintaining a dependency.

include/alpaka/acc/AccGpuUniformCudaHipRt.hpp
```diff
@@ -224,8 +228,22 @@ namespace alpaka
...
-                       + core::demangled<TIdx> + ">";
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic push
+#        pragma clang diagnostic ignored "-Wexit-time-destructors"
+#    endif
+                static std::string const accName =
+#    ifdef __cpp_lib_format
+                    std::format("AccGpu{}Rt<{},{}>", TApi::name, TDim::value, core::demangled<TIdx>);
+#    else
+                    (std::ostringstream()
```
psychocoderHPC: To workaround the nvcc error here 

```
error: class "std::basic_ostream<char, std::char_traits<char>>" has no member "str"
                          .str();
                           ^
```

I would try if `std::string(<ostringstream>)` works

include/alpaka/core/DemangleTypeNames.hpp
```diff
@@ -1,23 +1,96 @@
...
+        storage[length] = '\0';
+
+        return storage;
+    }
 
-#include <boost/core/demangle.hpp>
+    // Store the demangled type name as a null-terminated array of bytes.
+    // Note: this could be a function-static constexpr variable in c++23.
+    template<typename T>
+    inline constexpr auto storage = demangleAsArray<T>();
```
psychocoderHPC: [ci error](https://gitlab.com/hzdr/crp/alpaka/-/jobs/9023006168#L2060)

```
/builds/hzdr/crp/alpaka/include/alpaka/core/DemangleTypeNames.hpp(78): error: call to consteval function "<unnamed>::demangleAsArray<T>() [with T=uint32_t]" did not produce a valid constant expression
/builds/hzdr/crp/alpaka/include/alpaka/core/DemangleTypeNames.hpp(72): note #2751-D: access to uninitialized object
```

Maybe using `inline ALPAKA_FN_HOST constexpr auto storage = demangleAsArray<T>();` solve the issue

psychocoderHPC (changes requested):

include/alpaka/acc/AccCpuOmp2Blocks.hpp
fwyzard: Unfortunately gcc 10 does not like the `ostringstream`-based syntax (it works with gcc 11 and later, and all clang+ versions I tried).

I'll revert to
```c++
... = "AccCpuOmp2Blocks<"s + std::to_string(TDim::value) + ","s + std::string(core::demangled<TIdx>);
```

include/alpaka/core/DemangleTypeNames.hpp
```diff
@@ -1,23 +1,97 @@
...
+        // Use a fundamental type to detect the prefix and suffix, because MSVC includes the class, struct, or union
+        // keyword in the type signature.
+        using TestPatternType = double;
+
+        constexpr std::string_view testPatternName("double");
+
+        // Embed the demangled type name into a function signature, then extract it and return it as a null-terminated
+        // array of bytes.
+        template<typename T>
+        consteval auto demangleAsArray()
```
psychocoderHPC, line 57: the ci for nvcc 12.0 says:

```
/builds/hzdr/crp/alpaka/include/alpaka/core/DemangleTypeNames.hpp(82): error: call to consteval function "alpaka::core::detail::demangleAsArray<T>() [with T=uint32_t]" did not produce a valid constant expression
/builds/hzdr/crp/alpaka/include/alpaka/core/DemangleTypeNames.hpp(76): note #2751-D: access to uninitialized object
          detected during:
            instantiation of "const auto alpaka::core::detail::storage [with T=uint32_t]" 
(88): here
            instantiation of "std::string_view alpaka::core::detail::demangle<T>() [with T=uint32_t]" 
```

fwyzard: Yes, I can reproduce it locally with NVCC 12.0 and 12.1.
It appears to be fixed in NVCC 12.2 and later.

fwyzard: I can try to make it not `constexpr`/`consteval` if CUDA is too old ...

fwyzard: Ah, no need, I found a simpler fix: if I initialise the array the compiler seems happy.

include/alpaka/acc/AccCpuOmp2Blocks.hpp
fwyzard: I've opened #2504 to remind us to look into this later on.

fwyzard:
OK, NVCC 12.0 seems happy, now let's see if I fixed SYCL as well...

fwyzard:
ICPX seems happy as well :-)

psychocoderHPC (approved):

include/alpaka/acc/AccCpuSerial.hpp
```diff
@@ -155,7 +158,21 @@ namespace alpaka
...
+                static std::string const accName =
+#    ifdef __cpp_lib_format
+                    std::format("AccCpuSerial<{},{}>", TDim::value, core::demangled<TIdx>);
+#    else
+                    "AccCpuSerial<"s + std::to_string(TDim::value) + ","s + std::string(core::demangled<TIdx>) + ">"s;
+#    endif
+                return accName;
+#    if BOOST_COMP_CLANG
+#        pragma clang diagnostic pop
+#    endif
```
slava77, line 175: (I'm not sure which of the branches/variants we get in the CMSSW build) it looks like this change now prints a special character
a call `std::cout << " Running for Acc = " << alpaka::getAccName<ALPAKA_ACCELERATOR_NAMESPACE::Acc3D>() << std::endl;`

was returning
`Running for Acc = AccCpuSerial<3,unsigned int>` before and
`Running for Acc = AccCpuSerial<3,unsigned int�>` now

fwyzard: Yes, this is fixed in the `develop` branch by https://github.com/alpaka-group/alpaka/pull/2571 .

PR #2475, opened 2025-02-05
Add ConstBuf

AntonReinhard:
I've started implementing a constant buffer as discussed in #1729 
This constant buffer type can be constructed by converting implicitly from a normal/non-const buffer type (to allow passing to functions for example), or explicitly through a `makeConstBuf` trait. The constant version then only allows retrieving constant pointers, which prevents mutation of the underlying data of the buffer. The buffer also shares ownership, so one could create a buffer, fill it with data, then convert the buffer to a constant buffer when no more mutation is necessary.

For now I have a version drafted for the CPU buffer only and would like some feedback.


test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    // check that Constant buffer can't be converted to non-const buffer
+    STATIC_REQUIRE_FALSE(std::convertible_to<TCBuf, TBuf>);
+    STATIC_REQUIRE(std::convertible_to<TCBuf, TCBuf>);
+    STATIC_REQUIRE(std::convertible_to<TBuf, TBuf>);
+    STATIC_REQUIRE(std::convertible_to<TBuf, TCBuf>);
+
+    // check that Constant buffer cannot give a non const native ptr
+    // STATIC_REQUIRE(buftest::onlyConstNativePtr<TCBuf>);
+    // getPtrNative(c_buf)* = 0.f;  // <- this does not compile, as desired
+    // however, the static require above doesn't work and seems to get float* instead of float const*
```
AntonReinhard: I don't understand what is going wrong here, if anyone has an idea I'd appreciate some help. The concept `onlyConstNativePtr` concept is defined above lines 68-71.

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    template<typename TDim, typename TElem, typename TIdx, typename TExtent>
+    auto allocConstBuf(alpaka::DevCpu dev, TExtent extent) -> alpaka::ConstBufCpu<TElem, TDim, TIdx>
+    {
+        return alpaka::makeConstBuf(alpaka::allocBuf<TElem, TIdx>(dev, extent));
+    }
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
+    template<typename TDim, typename TElem, typename TIdx, typename TExtent>
+    auto allocBuf(alpaka::DevCudaRt dev, TExtent extent) -> alpaka::ConstBufCudaRt<TElem, TDim, TIdx>
```
fwyzard: Shouldn't this be `allocConstBuf` ?
Same for the functions below.

AntonReinhard: Yes, but I haven't implemented the Buffer for the other Accelerators yet anyway. I wanted to make a draft with the CPU Buffer first before I go and do it for all the other devices too.

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+    auto allocBuf(alpaka::DevCpuSycl dev, TExtent extent) -> alpaka::ConstBufCpuSycl<TElem, TDim, TIdx>
+    {
+        return alpaka::makeConstBuf(alpaka::allocBuf<TElem, TIdx>(dev, extent));
+    }
+#endif
+#if defined(ALPAKA_ACC_SYCL_ENABLED) and defined(ALPAKA_SYCL_ONEAPI_GPU)
+    template<typename TDim, typename TElem, typename TIdx, typename TExtent>
+    auto allocBuf(alpaka::DevGpuSyclIntel dev, TExtent extent) -> alpaka::ConstBufGpuSyclIntel<TElem, TDim, TIdx>
+    {
+        return alpaka::makeConstBuf(alpaka::allocBuf<TElem, TIdx>(dev, extent));
```
psychocoderHPC: I have currently problems following the PR and I am not sure anymore what this PR is trying to solve, maybe we should discuss this in our developer meeting.
I thought the `ConstBuff` should convert an exiting buffer into a nonmutable buffer, which makes IMO sense.
In this line we create a constant buffer that is not mutable, I am asking myself how to fill  nonmutable buffer with data :thinking:

test/unit/mem/buf/src/ConstBufTest.cpp
mehmetyusufoglu: Is it possible to use more direct type checking like the ones below . There are decltype and auto variables it is not easy to follow. 
```

    static_assert(
        std::is_same_v<decltype(alpaka::getPtrNative(buftest::allocConstBuf<Dim, Elem, Idx>(dev, extent))), Elem const*>,
        "getPtrNative for buftest::allocConstBuf<Dim, Elem, Idx>(dev, extent) must return Elem const*");

    static_assert(
        std::is_same_v<decltype(alpaka::getPtrNative(alpaka::allocBuf<Elem, Idx>(dev, extent))), Elem const*>,
        "getPtrNative for alpaka::allocBuf must return Elem const*");

    static_assert(
        !std::is_same_v<decltype(alpaka::getPtrNative(buftest::allocConstBuf<Dim, Elem, Idx>(dev, extent))), Elem *>,
        "getPtrNative for buftest::allocConstBuf<Dim, Elem, Idx>(dev, extent) must NOT return Elem *");
```

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: As I also explained in the meeting, this `allocBuf` function is only used for testing where I want to check that it is impossible to mute the ConstBuf's buffer. Therefore I don't care what data is inside.
The `makeConstBuf` function is simply a convenient explicit cast, creating a constant buffer from the mutable buffer. It just calls the copy constructor, but calling the copy constructor yourself would be tedious because you'd need to also provide all the correct template arguments for the resulting `ConstBuf`.

test/unit/mem/buf/src/ConstBufTest.cpp
SimeonEhrig: You need to remove the `const` modifier of `ConstBufCpu` type:
```c++
    template<typename TBuf>
    concept onlyConstNativePtr = requires(TBuf t) {
        {
            alpaka::getPtrNative(t)
        } -> std::same_as<alpaka::Elem<std::remove_const_t<TBuf>> const*>;
    };
```

`alpaka::Elem` removes only [volatile](https://github.com/alpaka-group/alpaka/blob/1793a0760e57bd3cca2b71170507ae1c26b5c0dd/include/alpaka/elem/Traits.hpp#L21) and you didn't specialized the type trait `ElemType` for the const version. But I have no idea, if this is design or a bug.

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    // check that Constant buffer can't be converted to non-const buffer
+    STATIC_REQUIRE_FALSE(std::convertible_to<TCBuf, TBuf>);
+    STATIC_REQUIRE(std::convertible_to<TCBuf, TCBuf>);
+    STATIC_REQUIRE(std::convertible_to<TBuf, TBuf>);
+    STATIC_REQUIRE(std::convertible_to<TBuf, TCBuf>);
+
+    // check that Constant buffer cannot give a non const native ptr
+    // STATIC_REQUIRE(buftest::onlyConstNativePtr<TCBuf>);
+    // getPtrNative(c_buf)* = 0.f;  // <- this does not compile, as desired
```
SimeonEhrig:
```suggestion
    // *getPtrNative(c_buf) = 0.f;  // <- this does not compile, as desired
```

Typo and syntax error.

include/alpaka/mem/buf/cpu/ConstBufCpu.hpp
```diff
@@ -0,0 +1,129 @@
...
+#include <functional>
+#include <memory>
+#include <type_traits>
+#include <utility>
+
+namespace alpaka
+{
+    //! The constant CPU memory buffer.
+    template<typename TElem, typename TDim, typename TIdx>
+    class ConstBufCpu : public internal::ViewAccessOps<ConstBufCpu<TElem, TDim, TIdx>>
```
SimeonEhrig, line 35: Like @fwyzard suggest, maybe it is a better idea to inherit from `BufCpu` and only specialize the [GetPtrNative](https://github.com/AntonReinhard/alpaka/blob/59d22208e7caa1a5c779e66556e9e99f8463e798/include/alpaka/mem/buf/BufCpu.hpp#L153) trait to save code.

AntonReinhard:
I'm redrafting this to try to prevent too much code duplication.

AntonReinhard:
In order to make my solution truly work generically, I'd have to change the implementation of the `BufGenericSycl` and `BufUniformCudaHipRt` and make them work like the `BufCpu`, i.e., hide it away in an extra Impl class. The pointers (currently `m_spBufCpuImpl` for the CPU impl, `m_spMem` for Sycl and CUDA/hip) are public fields in the classes, so if I change these, they will potentially break compatability. I'm not sure why these are public in the first place though.

Is that a problem?

SimeonEhrig:
@AntonReinhard Where is the `makeConstBuf` function? In the unit test, I found preprocessor cascade, which is not suitable for the user API. If I should review an intermediate state, the `makeConstBuf` function is still useful as entry point for the review.

AntonReinhard:
> @AntonReinhard Where is the `makeConstBuf` function? In the unit test, I found preprocessor cascade, which is not suitable for the user API. If I should review an intermediate state, the `makeConstBuf` function is still useful as entry point for the review.

I don't currently have a `makeConstBuf` anymore. I could add it again if it is deemed more user friendly. Currently, you simply cast an existing buffer into the respective `ConstBuf` version. The implementations are the constructors in the `ConstBuf<CPU/SYCL/CUDA/HIP>` files.

SimeonEhrig:
> > @AntonReinhard Where is the `makeConstBuf` function? In the unit test, I found preprocessor cascade, which is not suitable for the user API. If I should review an intermediate state, the `makeConstBuf` function is still useful as entry point for the review.
> 
> I don't currently have a `makeConstBuf` anymore. I could add it again if it is deemed more user friendly. Currently, you simply cast an existing buffer into the respective `ConstBuf` version. The implementations are the constructors in the `ConstBuf<CPU/SYCL/CUDA/HIP>` files.

Well, it is a little bit hard to review a functionally, if there is no code to demonstrate how to use it :grimacing: But I can play around with casting the buffer.
Actual I'm a little bit surprised, that you don't wrote an test and an example which uses the user API. If I start implement a function, I write the user API in a early stage to check, if the API is doable and my approach is working like thought. But if you get it manage it without it, it is fine. I think it is only uncommon.

AntonReinhard:
> > > @AntonReinhard Where is the `makeConstBuf` function? In the unit test, I found preprocessor cascade, which is not suitable for the user API. If I should review an intermediate state, the `makeConstBuf` function is still useful as entry point for the review.
> > 
> > 
> > I don't currently have a `makeConstBuf` anymore. I could add it again if it is deemed more user friendly. Currently, you simply cast an existing buffer into the respective `ConstBuf` version. The implementations are the constructors in the `ConstBuf<CPU/SYCL/CUDA/HIP>` files.
> 
> Well, it is a little bit hard to review a functionally, if there is no code to demonstrate how to use it 😬 But I can play around with casting the buffer. Actual I'm a little bit surprised, that you don't wrote an test and an example which uses the user API. If I start implement a function, I write the user API in a early stage to check, if the API is doable and my approach is working like thought. But if you get it manage it without it, it is fine. I think it is only uncommon.

The test in `ConstBufTest.cpp` does test the user API. It allocates a "normal" mutable buffer using the `alpaka::allocBuf` trait, then converts it immediately to the constant version of it. 

I should add more tests for this functionality, maybe I can find another test that would benefit from being switched to a `ConstBuf` usage somewhere, to have a more real use case.

AntonReinhard:
I added the `makeConstBuf` function and trait back. It now simply takes a buffer (mutable or constant) and returns a constant one with the same memory set (and no copy performed), but only read access allowed.
I used this in the matmul test, where the A and B matrices don't need write access within the kernel. They are now created, then filled with their contents, and then made into constant versions, which are then used for the rest of the test.

AntonReinhard:
I also split the buffer definitions and implementations in multiple files and tried to make it as consistent as possible across the different implementations. For every implementation (cpu, sycl, uniformcudahip) there are now (here given as example for the CPU ones):
- CpuBufImpl.hpp: The definition of the class storing the dimension, extent, etc. of the actual buffer.
- MutBufCpu.hpp: The definition of the class for a mutable buffer.
- ConstBufCpu.hpp: The definition of the class for a constant buffer.
- BufCpu.hpp: The main file importing the `MutBuf` and `ConstBuf` versions, their traits, `Copy.hpp` and `Set.hpp`, and making an alias from the `MutBufCpu` template to the `BufCpu` name in the same style as before this PR.
- traits/ConstBufCpuTraits.hpp: Implementation of the traits concerning the constant buffer.
- traits/MutBufCpuTraits.hpp: Implementation of the traits concerning the mutable buffer.
- specialization/*.hpp: Specializations of the constant and mutable buffer names for specific devices, in the case of sycl and uniformcudahip.

include/alpaka/mem/buf/Traits.hpp
```diff
@@ -189,4 +194,16 @@ namespace alpaka
...
+    //!
+    //! \tparam TBuf The type of the original buffer.
+    //! \param buf The original buffer.
+    //! \return The transformed buffer with only read-access allowed.
+    template<typename TBuf>
+    ALPAKA_FN_HOST auto makeConstBuf(TBuf const& buf)
+    {
+        return trait::MakeConstBuf<TBuf>::makeConstBuf(buf);
+    }
+
```
SimeonEhrig, line 216: Is it also possible to implement an `alpaka::ConstBuf` like `alpaka::Buf` it is implemented in this file?

https://github.com/alpaka-group/alpaka/blob/d3286539ba4a3037267b1cfc86e6cb300b4ff4af/include/alpaka/mem/buf/Traits.hpp#L56-L57

The definition is used in the examples, like here
https://github.com/alpaka-group/alpaka/blob/e906b1c0c449ebab02b45db6475246c972251aad/example/monteCarloIntegration/src/monteCarloIntegration.cpp#L97-L98

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,111 @@
...
+
+namespace buftest
+{
+    template<typename TDim, typename TElem, typename TIdx, typename TExtent>
+    auto allocConstBuf(alpaka::DevCpu dev, TExtent extent) -> alpaka::ConstBufCpu<TElem, TDim, TIdx>
+    {
+        return alpaka::makeConstBuf(alpaka::allocBuf<TElem, TIdx>(dev, extent));
+    }
+
+#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
```
SimeonEhrig: I don't think, we need the `#if`-else cascade anymore. If a backend does not implement the required traits, we get a compiler error.
You should test against the generic interface. The test automatically iterate over all enable backends. The rest is done by the CI.

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,111 @@
+/* Copyright 2025 Anton Reinhard
```
SimeonEhrig, line 1: Idea for an test: Can you please write a test, which creates an `alpaka::Buf`, creates `alpaka::ConstBuf` from it and the `alpaka::Buf` should run of scope. So the `alpaka::ConstBuf` should still point to valid memory and `alpaka::Buf` should not delete it.

include/alpaka/mem/buf/cpu/CpuBuf.hpp
fwyzard, line 1: The alias is called `BufCpu`, so the file should be called `BufCpu.hpp`, not `CpuBuf.hpp`.

SimeonEhrig:
We need a VC. I don't understand your implementation. The type of `alpaka::ConstBufCpu` is understandable:
`alpaka::ConstBufCpu<unsigned int, std::integral_constant<unsigned long, 1ul>, unsigned long>`

And you explained me, that `alpaka::MutBufCpu` "inherits" from `alpaka::ConstBufCpu` but I don't the mechanism. This is the type of `alpaka::MutBufCpu`:
`alpaka::MutBufCpu<alpaka::ConstBufCpu, alpaka::detail::BufCpuImpl<unsigned int, std::integral_constant<unsigned long, 1ul>, unsigned long>, alpaka::DevCpu, unsigned int, std::integral_constant<unsigned long, 1ul>, unsigned long>`

include/alpaka/mem/buf/cpu/CpuBuf.hpp
```diff
@@ -0,0 +1,17 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#include "alpaka/mem/buf/cpu/ConstBufCpu.hpp"
+#include "alpaka/mem/buf/cpu/MutBufCpu.hpp"
+
+namespace alpaka
+{
+    template<typename TElem, typename TDim, typename TIdx>
+    using BufCpu = MutBufCpu<ConstBufCpu, alpaka::detail::BufCpuImpl<TElem, TDim, TIdx>, DevCpu, TElem, TDim, TIdx>;
```
fwyzard: In fact, what is the point of having a `MutBufCpu` class and a `BufCpu` alias ?
Why not simply have a `BufCpu` class ?

include/alpaka/mem/buf/cpu/CpuBufImpl.hpp
fwyzard, line 1: `BufCpuImpl` name vs `CpuBufImpl.hpp`

fwyzard (changes requested):
Please do not introduce new names like `MutBufUniformCudaHipRt` and `MutBufCpu`, and then aliases for the old names.

include/alpaka/mem/buf/cpu/CpuBuf.hpp
AntonReinhard: The point was to hide the additional template arguments I introduced from users of the class. But I can try to get rid of them by deriving them inside the class directly, I think it should be possible.

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: That's a good idea :+1:

include/alpaka/mem/buf/Traits.hpp
AntonReinhard: That should be easy to add :+1:

include/alpaka/mem/buf/cpu/CpuBuf.hpp
AntonReinhard: There were some inconsistencies in the naming of classes and files beforehand, I mostly tried to keep the inconsistencies "consistently". I'll try to rename things to make them make more sense.

include/alpaka/mem/buf/cpu/CpuBuf.hpp
fwyzard: Mhm, I see.

SimeonEhrig: We should discuss this via VC or directly. I still think, we can make it easier. For me, I don't understand why we need the template parameter `ConstBufCpu` **and** `alpaka::detail::BufCpuImpl` **and** `DevCpu`.

I think, if I understand the code, we can find a more understandable solution, maybe by inheritance, maybe with traits and concepts or something else.

AntonReinhard: Yes I already noticed they are not needed anymore at all, it was leftover from my previous more generic attempt at an implementation. Please wait a bit until I can revise what I have, then we can talk about that.

include/alpaka/mem/buf/cpu/CpuBufImpl.hpp
AntonReinhard: Done in ef24409ef8de95123204c9be82b4d48a568eebea

include/alpaka/mem/buf/cpu/CpuBuf.hpp
AntonReinhard: Please take a look at ef24409ef8de95123204c9be82b4d48a568eebea

include/alpaka/mem/buf/cpu/CpuBuf.hpp
AntonReinhard: Done in ef24409ef8de95123204c9be82b4d48a568eebea

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: Done in 501d2b1a765f26e65c785584fe25cdb321152fcb

include/alpaka/mem/buf/Traits.hpp
AntonReinhard: Done in ba70daa61920643357f1fe0a930eb2e3720067eb

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: Actually, the ConstBufTest already does this.
```cpp
template<typename TDev, typename TDim, typename TElem, typename TIdx, typename TExtent>
auto allocConstBuf(TDev dev, TExtent extent) -> alpaka::ConstBuf<TDev, TElem, TDim, TIdx>
{
    return alpaka::makeConstBuf(alpaka::allocBuf<TElem, TIdx>(dev, extent));
}
```
This function creates a normal buffer, then creates a new constant buffer from it and returns only that. The temporary originally allocated buffer goes out of scope there.

AntonReinhard:
> We need a VC. I don't understand your implementation. The type of `alpaka::ConstBufCpu` is understandable: `alpaka::ConstBufCpu<unsigned int, std::integral_constant<unsigned long, 1ul>, unsigned long>`

In ef24409ef8de95123204c9be82b4d48a568eebea, I removed the extra template arguments from the mutable buffer types, so they should look roughly the same as before the PR. They were leftovers from my previous attempt at an implementation that I reworked.

> And you explained me, that `alpaka::MutBufCpu` "inherits" from `alpaka::ConstBufCpu` but I don't the mechanism. This is the type of `alpaka::MutBufCpu`: `alpaka::MutBufCpu<alpaka::ConstBufCpu, alpaka::detail::BufCpuImpl<unsigned int, std::integral_constant<unsigned long, 1ul>, unsigned long>, alpaka::DevCpu, unsigned int, std::integral_constant<unsigned long, 1ul>, unsigned long>`

The mutable buffer types don't actually inherit from the const buffer types anymore. They only use the constant buffers' implementations of all the traits by casting themselves to a const buffer and then calling the trait on that.
For example:
```cpp
    //! The BufCpu width get trait specialization.
    template<typename TElem, typename TDim, typename TIdx>
    struct GetExtents<BufCpu<TElem, TDim, TIdx>>
    {
        ALPAKA_FN_HOST auto operator()(BufCpu<TElem, TDim, TIdx> const& buf)
        {
            return GetExtents<ConstBufCpu<TElem, TDim, TIdx>>{}(ConstBufCpu<TElem, TDim, TIdx>{buf});
        } //                  ^ Calling the const buf trait     ^ Casting to a const buf
    };
```

AntonReinhard:
Now that I think about it, the casts might not be able to be compiled away by the compiler because they copy a `shared_ptr` internally. Maybe they should not cast and instead just copy the implementation as code...

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,74 @@
...
+    concept onlyConstNativePtr = requires(TBuf t) {
+        {
+            alpaka::getPtrNative(t)
+        } -> std::same_as<alpaka::Elem<TBuf> const*>;
+    };
+
+} // namespace buftest
+
+template<typename TAcc>
+static auto testConstBuffer(alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> const& extent) -> void
```
SimeonEhrig, line 51: Can you please add also an test, which checks the return type of the access operator of a const and non const buffer. Therefor you need to allocate an 1D buffer.

```c++
static_assert(std::is_same_v<decltype(buf[0]), int&>);
```

SimeonEhrig:
The current implementation looks much better. Please add the test case, which I suggested. Else @fwyzard will do the final review.

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: I added it in dde2ed4f0e99a39d865aeca36d627b1f9b842052

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    Queue queue{dev};
+
+    // create local buffer filled with all ones
+    auto const platformHost = alpaka::PlatformCpu{};
+    auto const devHost = alpaka::getDevByIdx(platformHost, 0);
+    std::vector<Elem> dataHost(static_cast<std::size_t>(getExtentProduct(extent)), static_cast<Elem>(1));
+    auto bufHost = alpaka::createView(devHost, dataHost.data(), extent);
+
+    auto copiedBack = alpaka::allocBuf<Elem, Idx>(devHost, extent);
```
fwyzard, line 115: why `copiedBack` ?
if I understand correctly the test, the data is always on CPU:
  - in `dataHost`
  - in `bufHost`
  - in `copiedBack`

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+    STATIC_REQUIRE(buftest::onlyConstNativePtr<TCBuf>);
+    // *getPtrNative(c_buf) = 0.f;  // <- this does not compile, as desired
+
+    // check return types of the buffers
+    STATIC_REQUIRE(std::is_same_v<decltype(buf[0]), Elem&>);
+    STATIC_REQUIRE(std::is_same_v<decltype(c_buf[0]), Elem const&>);
+}
+
+template<typename TAcc>
+static auto testConstBufLifetime(alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> const& extent) -> void
```
fwyzard, line 92: I don't think this test actually does any tests of the buffers lifetime.
Maybe the implementation should be different ?
Or maybe just the name ?

test/unit/mem/buf/src/ConstBufTest.cpp
```diff
@@ -0,0 +1,136 @@
...
+        auto const platformHost = alpaka::PlatformCpu{};
+        auto const devHost = alpaka::getDevByIdx(platformHost, 0);
+        std::vector<TElem> dataHost(static_cast<std::size_t>(getExtentProduct(extent)), static_cast<TElem>(1));
+        auto bufHost = alpaka::createView(devHost, dataHost.data(), extent);
+        QueueAcc queueAcc{dev};
+        alpaka::memcpy(queueAcc, buf, bufHost);
+        alpaka::wait(queueAcc);
+
+        // make the buffer constant and return only that
+        auto const constBuf = alpaka::makeConstBuf(buf);
```
fwyzard: Would `alpaka::makeConstBuf(std::move(buf))` work ?
Or does it break the type deduction in the trait ?

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: In the next line, I `memcpy` the data from the `c_buf`, which lives on the accelerator, back to the CPU to be able to compare it with the expected results.

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: In `initConstBuf`, a normal buffer is allocated, initialized, then converted to a constant buffer. The originally allocated buffer then runs out of scope. So the intention is that this tests that the returned constant buffer still owns the memory and it didn't get deleted when the original non-constant buffer runs out of scope.

include/alpaka/mem/buf/cpu/ConstBufCpu.hpp
```diff
@@ -0,0 +1,58 @@
...
+        template<typename TExtent, typename Deleter>
+        ALPAKA_FN_HOST ConstBufCpu(DevCpu const& dev, TElem* pMem, Deleter deleter, TExtent const& extent)
+            : m_spBufImpl{
+                std::make_shared<detail::BufCpuImpl<TElem, TDim, TIdx>>(dev, pMem, std::move(deleter), extent)}
+        {
+        }
+
+        ALPAKA_FN_HOST ConstBufCpu(BufCpu<TElem, TDim, TIdx> const& buf) : m_spBufImpl{buf.m_spBufImpl}
+        {
+        }
```
fwyzard, line 47: Can you add an overload
```c++
        ALPAKA_FN_HOST ConstBufCpu(BufCpu<TElem, TDim, TIdx> && buf) : m_spBufImpl{std::move(buf.m_spBufImpl)} {}
```
?

fwyzard: (also for the other buffer types)

fwyzard (changes requested):

test/unit/mem/buf/src/ConstBufTest.cpp
AntonReinhard: It seems to work :+1:

test/unit/mem/buf/src/ConstBufTest.cpp
fwyzard: Ah, OK, it's kind of hidden form the test itself, but I see what you mean.

include/alpaka/mem/buf/uniformCudaHip/traits/ConstBufUniformCudaHipRtTraits.hpp
```diff
@@ -0,0 +1,205 @@
...
+            TElem* memPtr = nullptr;
+            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::hostMalloc(
+                reinterpret_cast<void**>(&memPtr),
+                sizeof(TElem) * static_cast<std::size_t>(getExtentProduct(extent)),
+                TApi::hostMallocMapped | TApi::hostMallocPortable));
+            auto deleter = [](TElem* ptr) { ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::hostFree(ptr)); };
+
+            return BufCpu<TElem, TDim, TIdx>(host, memPtr, std::move(deleter), extent);
+        }
+    };
```
fwyzard: Move the allocation to the non-const buffer traits ?

include/alpaka/mem/buf/sycl/traits/ConstBufGenericSyclTraits.hpp
```diff
@@ -0,0 +1,143 @@
+/* Copyright 2024 Jan Stephan, Luca Ferragina, Aurora Perego, Andrea Bocci
```
fwyzard: Add your name ?

include/alpaka/mem/buf/uniformCudaHip/BufUniformCudaHipRt.hpp
```diff
@@ -0,0 +1,54 @@
...
+    public:
+        std::shared_ptr<TBufImpl> m_spBufImpl;
+    };
+
+} // namespace alpaka
+
+#    include "alpaka/mem/buf/uniformCudaHip/Copy.hpp"
+#    include "alpaka/mem/buf/uniformCudaHip/Set.hpp"
+#    include "alpaka/mem/buf/uniformCudaHip/traits/BufUniformCudaHipRtTraits.hpp"
+#    include "alpaka/mem/buf/uniformCudaHip/traits/ConstBufUniformCudaHipRtTraits.hpp"
```
fwyzard, line 52: I find it confusing that we have these includes in the mutable buffer's header, and not in the const buffer's header.
(I understand that the latter includes the former, but it seems more of an accident than a strict requirement).

Do they need to be include after the definition of `BufUniformCudaHipRt` (but not after the definition of `ConstBufUniformCudaHipRt`) ?

~~Would it work to move them at the end of `BufUniformCudaHipRtImpl.hpp` ?~~

No, that doesn't seem the right place either...

include/alpaka/mem/buf/sycl/traits/ConstBufGenericSyclTraits.hpp
AntonReinhard: Done in fbb8325e3379c28a5da44a557eccaea75d8dc32d

include/alpaka/mem/buf/uniformCudaHip/traits/ConstBufUniformCudaHipRtTraits.hpp
AntonReinhard: Forgot this one... Done in fbb8325e3379c28a5da44a557eccaea75d8dc32d

include/alpaka/mem/buf/cpu/ConstBufCpu.hpp
AntonReinhard: Added in fbb8325e3379c28a5da44a557eccaea75d8dc32d

include/alpaka/mem/buf/uniformCudaHip/traits/ConstBufUniformCudaHipRtTraits.hpp
```diff
@@ -0,0 +1,180 @@
...
+    template<typename TApi, typename TDim>
+    struct HasAsyncBufSupport<TDim, DevUniformCudaHipRt<TApi>> : std::bool_constant<TDim::value <= 1>
+    {
+    };
+
+    //! The pinned/mapped memory allocation capability trait specialization.
+    template<typename TApi>
+    struct HasMappedBufSupport<PlatformUniformCudaHipRt<TApi>> : public std::true_type
+    {
+    };
```
fwyzard: Sorry, I missed them earlier: these should also go to the mutable buffer, since they describe what allocation methods are supported.

include/alpaka/mem/buf/uniformCudaHip/BufUniformCudaHipRt.hpp
fwyzard: OK, if there is no better place, I guess they can stay here.

fwyzard:
OK, I have two last general comments...

Comment #1: can you make sure that one can call `alpaka::makeConstBuf(std::move(buf))` and have the move operation propagate to all implementations, so that there is no increment/decrement of the shared pointer counter ?

Comment #2: I'm still convinced that this could be implemented making `Buf` inherit from `ConstBuf`, which could avoid the cast and the temporary when a `Buf` is passed to a function that takes a `ConstBuf const&` - but we can go ahead and merge this PR, and explore this in a follow up.

include/alpaka/mem/buf/uniformCudaHip/traits/ConstBufUniformCudaHipRtTraits.hpp
AntonReinhard: Moved in 78cf911ebb9e15e8c45de565f46abb7fff0cf3af

SimeonEhrig:
There is a concept to check, if a object is moveable. https://en.cppreference.com/w/cpp/concepts/movable
```c++
// e.g.
STATIC_REQUIRED(std::moveable<ConstBuf<...>>);
```

AntonReinhard:
> There is a concept to check, if a object is moveable. https://en.cppreference.com/w/cpp/concepts/movable
> 
> ```c++
> // e.g.
> STATIC_REQUIRED(std::moveable<ConstBuf<...>>);
> ```

Thanks for that hint, I added a test for this as well.

Otherwise I also agree with @fwyzard 's comment 2, I would prefer to merge this now and I can look at the inheritance possibility again in a separate PR.

SimeonEhrig (approved):

fwyzard (approved):

PR #2476, opened 2025-02-07
Add option for installation

chillenzer:
Allow to install alpaka as part of another project if configured and export its targets into that installation.

Also fixes a bug: `target_include_directories` was given an absolute path (`_alpaka_INCLUDE_DIRECTORY` is absolute) to the source directory which meant that dependents always used the headers directly from the source directory if they were using the `alpaka::alpaka` target directly. That was never a problem if dependents were applications that didn't expose alpaka headers themselves or the final application to be built used alpaka itself (leading to basically the same situation). But if you try to install a library L that depends on alpaka and a user of the library tries to use L from the installation without using alpaka directly, it needs to find alpaka's headers and they are located in the source directory (which might have been removed already), not in the installation directory.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,7 +162,7 @@ endif()
 ################################################################################
 # Installation.
 
-option(alpaka_INSTALL "" OFF)
+option(alpaka_INSTALL "install alpaka even if used as project dependency" OFF)
 # Do not install if alpaka is used as a CMake subdirectory
 if(CMAKE_PROJECT_NAME STREQUAL PROJECT_NAME OR alpaka_INSTALL)
     include(CMakePackageConfigHelpers)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -162,8 +162,9 @@ endif()
 ################################################################################
 # Installation.
 
+option(alpaka_INSTALL "" OFF)
 # Do not install if alpaka is used as a CMake subdirectory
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -162,8 +162,9 @@ endif()
 ################################################################################
 # Installation.
 
+option(alpaka_INSTALL "" OFF)
```
psychocoderHPC: please add a short doc instead of `""` e.g.

`allow installation of alpaka if used as project dependency`

psychocoderHPC: please add this option additionally to our documentation: https://github.com/alpaka-group/alpaka/blob/develop/docs/source/advanced/cmake.rst

psychocoderHPC (changes requested):

CMakeLists.txt
chillenzer: Both done.

psychocoderHPC (approved):

chillenzer:
Jobs are failing with 
```
Error: retrieving gpg key timed out.
Cleaning up project directory and file based variables 00:01
ERROR: Job failed: exit code 1
```
Please, restart!

chillenzer:
@psychocoderHPC We're down to two failing runners: https://gitlab.com/hzdr/crp/alpaka/-/pipelines/1664050946/failures Please restart!

psychocoderHPC (approved):

PR #2436, opened 2024-12-05
add math functions for complex type for the oneAPI backend

AuroraPerego:
The oneAPI backend did not have the specialization of math functions on the device for complex types.
I moved (to avoid code duplication) the specializations done for CUDA/HIP to `Complex.hpp` to allow the oneAPI backend to use them as well.

PR #2479, opened 2025-02-17
Make complex math functions host-device

chillenzer:
This is the immediate fix requested in [here](https://github.com/alpaka-group/alpaka/pull/2436/files/cbf73204f60d33e65f58a29f42b9138ed89387be#r1956328783) for the following problem: Compiling in `alpaka_ACC_GPU_CUDA_ONLY_MODE=ON` resolves `ALPAKA_FN_ACC` into `__device__` (and not `__host__ __device__`). However, in some situations the compiler has problems deducing the correct return type (for return type `auto`) if it only ever knows the device code and doesn't generate host code. This results in weird compilation errors along the lines of `error: invalid use of void expression` for functions that clearly return a non-`void` value. In our case, the complex math functions are the culprit.

CUDA-only mode is currently not run in CI (presumably because some examples explicitly need `AccCpuSerial`). @SimeonEhrig, could we run CUDA-only mode in CI?

CAUTION: Running `test/unit/math/mathTest` in CUDA-only mode still fails after this fix with the following error message:
```
include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp(334) 'TApi::funcGetAttributes(&funcAttrs, reinterpret_cast<void const*>(kernelName))' returned error  : 'cudaErrorInvalidDeviceFunction': 'invalid device function'!

[...]

/test/unit/math/src/mathLambda.cpp:60: FAILED:
due to a fatal error condition:
  testing acc:alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, std::
  integral_constant<unsigned long, 1ul>, unsigned long> data type:double
  functor:__nv_hdl_wrapper_t<false, true, false, __nv_dl_tag<void
  (LambdaMathTestTemplate<alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt,
  std::integral_constant<unsigned long, 1ul>, unsigned long>, double>::*)()
  const, &(void LambdaMathTestTemplate<alpaka::AccGpuUniformCudaHipRt<alpaka::
  ApiCudaRt, std::integral_constant<unsigned long, 1ul>, unsigned long>,
  double>::operator()<std::tuple<mathtest::OpAbs, mathtest::OpAcos, mathtest::
  OpAcosh, mathtest::OpArg, mathtest::OpAsin, mathtest::OpAsinh, mathtest::
  OpAtan, mathtest::OpAtanh, mathtest::OpCbrt, mathtest::OpCeil, mathtest::
  OpCos, mathtest::OpCosh, mathtest::OpErf, mathtest::OpExp, mathtest::OpFloor,
  mathtest::OpLog, mathtest::OpLog2, mathtest::OpLog10, mathtest::OpRound,
  mathtest::OpRsqrt, mathtest::OpSin, mathtest::OpSinh, mathtest::OpSqrt,
  mathtest::OpTan, mathtest::OpTanh, mathtest::OpTrunc, mathtest::OpIsnan,
  mathtest::OpIsinf, mathtest::OpIsfinite> >() const), 1u>, void (mathtest::
  ArgsItem<double, (mathtest::Arity)1> const&, alpaka::AccGpuUniformCudaHipRt
  <alpaka::ApiCudaRt, std::integral_constant<unsigned long, 1ul>, unsigned
  long> const&)> seed:2721678720
  SIGILL - Illegal instruction signal
```
(running in debug mode generates the SIGILL which seems to be somehow hidden otherwise). This disappears by disabling CUDA-only mode (still activating only a single accelerator). So apparently there's something more going on here. `cuda-gdb` reveals that it's while `getValidWorkDiv` failing to get function properties. This is likely related, potentially because I haven't found all instances of this happening. We need to dig into this further but I would still welcome merging this partial fix already because it's needed for PIConGPU.

As @psychocoderHPC is on vacation, would you, @AuroraPerego or someone else, do the honours please?

### DIFF
--- a/include/alpaka/math/Complex.hpp
+++ b/include/alpaka/math/Complex.hpp
@@ -593,7 +593,7 @@ namespace alpaka
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            ALPAKA_FN_ACC auto operator()(TCtx const& ctx, Complex<T> const& arg)
+            ALPAKA_FN_HOST_ACC auto operator()(TCtx const& ctx, Complex<T> const& arg)
             {
                 return sqrt(ctx, arg.real() * arg.real() + arg.imag() * arg.imag());
             }
@@ -605,7 +605,7 @@ namespace alpaka
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            ALPAKA_FN_ACC auto operator()(TCtx const& ctx, Complex<T> const& arg)
+            ALPAKA_FN_HOST_ACC auto operator()(TCtx const& ctx, Complex<T> const& arg)
             {
                 // This holds everywhere, including the branch cuts: acos(z) = -i * ln(z + i * sqrt(1 - z^2))
                 return Complex<T>{static_cast<T>(0.0), static_cast<T>(-1.0)}
@@ -623,7 +623,7 @@ namespace alpaka
         {
             //! Take context as original (accelerator) type, since we call other math functions
             template<typename TCtx>
-            ALPAKA_FN_ACC auto operator()(TCtx const& ctx, Complex<T> const& arg)
+            ALPAKA_FN_HOST_ACC auto operator()(TCtx const& ctx, Complex<T> const& arg)
             {
                 // acos(z) = ln(z + sqrt(z-1) * sqrt(z+1))
                 return log(ctx, arg + sqrt(ctx, arg - static_cast<T>(1.0)) * sqrt(ctx, arg + static_cast<T>(1.0)));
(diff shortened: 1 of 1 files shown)


SimeonEhrig:
> CUDA-only mode is currently not run in CI (presumably because some examples explicitly need `AccCpuSerial`). @SimeonEhrig, could we run CUDA-only mode in CI?

Do we need to test different CUDA versions or is one fine?

chillenzer:
> Do we need to test different CUDA versions or is one fine?

I think a single one would be enough. It might come a time when `nvcc` learns to handle this correctly at which point we'd need to care about the version more precisely. But a single check would be a good start to build upon. We can still add more later as required.

AuroraPerego (approved):

PR #2480, opened 2025-02-18
Add CUDA-only mode job to the CI

SimeonEhrig:
related to this problem: https://github.com/alpaka-group/alpaka/pull/2479#issue-2857604635

### DIFF
--- a/example/bufferCopy/src/bufferCopy.cpp
+++ b/example/bufferCopy/src/bufferCopy.cpp
@@ -9,6 +9,8 @@
 #include <cstdint>
 #include <iostream>
 
+#ifdef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
+
 //! Prints all elements of the buffer.
 struct PrintBufferKernel
 {
@@ -254,3 +256,12 @@ auto main() -> int
     //   TagFpgaSyclIntel, TagGenericSycl, TagGpuSyclIntel
     return alpaka::executeForEachAccTag([=](auto const& tag) { return example(tag); });
 }
+
+#else
+
+int main()
+{
+    return EXIT_SUCCESS;
+}
+
+#endif
--- a/example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
+++ b/example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
@@ -31,7 +31,7 @@ inline constexpr bool is_mdspan = IsMdspan<T>::value;
 // Index type
 using Idx = std::size_t;
 // Set data type
-using DataType = float;
+using DataType = std::uint64_t;
 
 /**
  * @brief Kernel for performing multiplication of two 2D matrices. Each element is computed by a different thread.
@@ -165,17 +165,19 @@ auto example(TAccTag const&) -> int
     auto mdHostC = alpaka::experimental::getMdSpan(bufHostC);
     for(Idx i = 0; i < M; ++i)
     {
-        for(Idx j = 0; j < N; ++j)
+        if(success)
         {
-            DataType expectedValue = 0.0f;
-            for(Idx k = 0; k < K; ++k)
+            for(Idx j = 0; j < N; ++j)
             {
-                expectedValue += mdHostA(i, k) * mdHostB(k, j);
-            }
-            if(mdHostC(i, j) != expectedValue)
-            {
-                success = false;
-                break;
+                DataType expectedValue = 0.0f;
+                for(Idx k = 0; k < K; ++k)
+                {
+                    expectedValue += mdHostA(i, k) * mdHostB(k, j);
+                }
+                if(mdHostC(i, j) != expectedValue)
+                {
+                    success = false;
+                }
             }
         }
     }
--- a/example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
+++ b/example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
@@ -102,7 +102,7 @@ auto example(TAccTag const&) -> int
     Idx const K = 1024;
 
     // Define device and queue
-    using Acc = alpaka::AccCpuSerial<Dim, Idx>;
+    using Acc = alpaka::TagToAcc<TAccTag, Dim, Idx>;
     using Queue = alpaka::Queue<Acc, alpaka::Blocking>;
     using Vec = alpaka::Vec<Dim, Idx>;
 
--- a/test/integ/cudaOnly/src/cudaNativeFunctions.cpp
+++ b/test/integ/cudaOnly/src/cudaNativeFunctions.cpp
@@ -14,7 +14,7 @@
 #        pragma clang diagnostic push
 #        pragma clang diagnostic ignored "-Wmissing-prototypes"
 #    endif
-__device__ auto userDefinedThreadFence() -> void
+__device__ static auto userDefinedThreadFence() -> void
 {
     __threadfence();
 }
--- a/example/counterBasedRng/src/counterBasedRng.cpp
+++ b/example/counterBasedRng/src/counterBasedRng.cpp
@@ -11,6 +11,8 @@
 #include <random>
 #include <typeinfo>
 
+#ifdef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
+
 //! A kernel that fills an array with pseudo-random numbers.
 class CounterBasedRngKernel
 {
@@ -241,3 +243,12 @@ auto main() -> int
     //   TagFpgaSyclIntel, TagGenericSycl, TagGpuSyclIntel
     return alpaka::executeForEachAccTag([=](auto const& tag) { return example(tag); });
 }
+
+#else
+
+int main()
+{
+    return EXIT_SUCCESS;
+}
+
+#endif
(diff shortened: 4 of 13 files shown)


fwyzard:
Is the error

> ```
> /builds/hzdr/crp/alpaka/example/useBLASInAlpaka/useCuBLASInAlpaka/src/useCuBLASInAlpaka.cpp:11:10: fatal error: cublas_v2.h: No such file or directory
>    11 | #include <cublas_v2.h>
>       |          ^~~~~~~~~~~~~
> ```

actually related to this PR ?

SimeonEhrig:
> Is the error
> 
> > ```
> > /builds/hzdr/crp/alpaka/example/useBLASInAlpaka/useCuBLASInAlpaka/src/useCuBLASInAlpaka.cpp:11:10: fatal error: cublas_v2.h: No such file or directory
> >    11 | #include <cublas_v2.h>
> >       |          ^~~~~~~~~~~~~
> > ```
> 
> actually related to this PR ?

No, the error was introduced in another PR. Looks like the CMake is wrong. But without CI job, would could not detect it. I will fix it later.

SimeonEhrig:
> > Is the error
> > > ```
> > > /builds/hzdr/crp/alpaka/example/useBLASInAlpaka/useCuBLASInAlpaka/src/useCuBLASInAlpaka.cpp:11:10: fatal error: cublas_v2.h: No such file or directory
> > >    11 | #include <cublas_v2.h>
> > >       |          ^~~~~~~~~~~~~
> > > ```
> > 
> > 
> > actually related to this PR ?
> 
> No, the error was introduced in another PR. Looks like the CMake is wrong. But without CI job, would could not detect it. I will fix it later.

I was wrong. The problem is the container. cuBlas is not installed. Now I have to check why. I thought it is part of the basic installation of the CUDA SDK.

fwyzard:
> I was wrong. The problem is the container. cuBlas is not installed. Now I have to check why. I thought it is part of the basic installation of the CUDA SDK.

But, how do the other CUDA CI tests work, then ?

SimeonEhrig:
> > I was wrong. The problem is the container. cuBlas is not installed. Now I have to check why. I thought it is part of the basic installation of the CUDA SDK.
> 
> But, how do the other CUDA CI tests work, then ?

Other tests does not require the cuBlas library and until yet we didn't test this example, because we had no cuda only test.

fwyzard:
> Other tests does not require the cuBlas library

Ah !

fwyzard:
Same error as https://github.com/alpaka-group/alpaka/pull/2503#issuecomment-2967341691 :-/

SimeonEhrig:
@chillenzer Looks like that the bug from PR #2479 still exist: https://gitlab.com/hzdr/crp/alpaka/-/jobs/10341320002
Not we get a better error message.

That is a cmake config, which you can use to reproduce it locally.

```bash
cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_CXX_COMPILER=g++ -Dalpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE=OFF -Dalpaka_ACC_GPU_CUDA_ONLY_MODE=ON -Dalpaka_ENABLE_WERROR=ON -Dalpaka_ACC_GPU_CUDA_ENABLE=ON -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON -Dalpaka_USE_MDSPAN=FETCH -Dalpaka_BUILD_BENCHMARKS=ON -Dalpaka_CHECK_HEADERS=ON
```

example/bufferCopy/src/bufferCopy.cpp
```diff
@@ -9,6 +9,8 @@
 #include <cstdint>
 #include <iostream>
 
+#ifdef ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLED
```
mehmetyusufoglu, line 12: Only expected to run for this backend ?

SimeonEhrig: No, if I don't define it wrong the serial backend needs to be enabled and all other backends can be.

mehmetyusufoglu: ok, that means it is not compiled for cuda-only.

fwyzard:
Why the change in data type ?

example/bufferCopy/src/bufferCopy.cpp
SimeonEhrig: That's correct. The examples requires an CPU accelerator for the host, to enqueue tasks. Therefore they cannot work in CUDA only mode, where all accelerators except the CUDA is disabled.

chillenzer:
> @chillenzer Looks like that the bug from PR #2479 still exist: https://gitlab.com/hzdr/crp/alpaka/-/jobs/10341320002 Not we get a better error message.
> 
> That is a cmake config, which you can use to reproduce it locally.
> 
> ```shell
> cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_CXX_COMPILER=g++ -Dalpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE=OFF -Dalpaka_ACC_GPU_CUDA_ONLY_MODE=ON -Dalpaka_ENABLE_WERROR=ON -Dalpaka_ACC_GPU_CUDA_ENABLE=ON -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON -Dalpaka_USE_MDSPAN=FETCH -Dalpaka_BUILD_BENCHMARKS=ON -Dalpaka_CHECK_HEADERS=ON
> ```

Ah, I knew that there was something. If this is urgent, someone else should pick it up please. This would land on a very full to-do list currently and is not highest priority.

SimeonEhrig:
> Why the change in data type ?

I get summation errors if I compare serial implementation and a parallelized implementation because of the low precession of `float`. The delta is about 10e7. I could spend some time to solve the problem or simply change the data type which does not have the problem. I think for the example it is fine, because it should demonstrate memory access with a well know algorithm.

mehmetyusufoglu:
Changing ALPAKA_FN_HOST_ACC to ALPAKA_FN_ACC at LambdaMathTestTemplate seems solved the problem. I will create a PR

mehmetyusufoglu:
> > @chillenzer Looks like that the bug from PR #2479 still exist: https://gitlab.com/hzdr/crp/alpaka/-/jobs/10341320002 Not we get a better error message.
> > That is a cmake config, which you can use to reproduce it locally.
> > ```shell
> > cmake -DCMAKE_BUILD_TYPE=RELEASE -DCMAKE_CXX_COMPILER=g++ -Dalpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE=OFF -Dalpaka_ACC_GPU_CUDA_ONLY_MODE=ON -Dalpaka_ENABLE_WERROR=ON -Dalpaka_ACC_GPU_CUDA_ENABLE=ON -Dalpaka_BUILD_EXAMPLES=ON -DBUILD_TESTING=ON -Dalpaka_USE_MDSPAN=FETCH -Dalpaka_BUILD_BENCHMARKS=ON -Dalpaka_CHECK_HEADERS=ON
> > ```
> 
> Ah, I knew that there was something. If this is urgent, someone else should pick it up please. This would land on a very full to-do list currently and is not highest priority.

I am checking.

SimeonEhrig:
> Changing ALPAKA_FN_HOST_ACC to ALPAKA_FN_ACC at LambdaMathTestTemplate seems solved the problem. I will create a PR

I tested it locally. Looks good. Thanks for the fix.

psychocoderHPC (approved):

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
```diff
@@ -165,17 +165,19 @@ auto example(TAccTag const&) -> int
...
-            DataType expectedValue = 0.0f;
-            for(Idx k = 0; k < K; ++k)
-            {
-                expectedValue += mdHostA(i, k) * mdHostB(k, j);
-            }
-            if(mdHostC(i, j) != expectedValue)
+            for(Idx j = 0; j < N; ++j)
             {
-                success = false;
-                break;
```
fwyzard, line 178: why remove the `break` ?

test/integ/cudaOnly/src/cudaNativeFunctions.cpp
```diff
@@ -14,7 +14,7 @@
 #        pragma clang diagnostic push
 #        pragma clang diagnostic ignored "-Wmissing-prototypes"
 #    endif
-__device__ auto userDefinedThreadFence() -> void
+__device__ static auto userDefinedThreadFence() -> void
```
fwyzard, line 17: just curious - why `static` ?

example/matrixMulWithMdspan/src/matrixMulMdSpan.cpp
SimeonEhrig: I thought it is not working, because we are in double loop and we exit only the inner loop. But this was a mistake on my side. We are only in a single loop. The new implementation does the same like the old one. Maybe a little bit more inefficient.

test/integ/cudaOnly/src/cudaNativeFunctions.cpp
SimeonEhrig: There was the warning, that `userDefinedThreadFence` needs to be defined. I think it means a definition in a header. I used static to change visibility to hidden.

fwyzard: Mhm, that's what the `pragma clang diagnostic` right above were supposed to silence.

SimeonEhrig: GCC throw the exception

Issue #2466, opened 2025-01-29
`alpaka::math::isfinite(...)` does not work in "fast math" compilation mode


fwyzard:
The math functions like `isfinite`, `isnan`, `isinf` that rely on their `std` counterparts for CPU code fail when the code is built with `-Ofast`, `-ffast-math` or `-ffinite-math`.
This is documents on the GCC manual [1](https://gcc.gnu.org/onlinedocs/gcc/Other-Builtins.html) and on cppreference [2](https://en.cppreference.com/w/cpp/numeric/math/isinf), but may still be a source of issues.

With `-O3`:
```
$ ./mathTest 
Randomness seeded to: 1147509038
===============================================================================
All tests passed (268120 assertions in 290 test cases)
```

With `-Ofast`:
```
$ ./mathTest 
Randomness seeded to: 3034605778
...
/home/fwyzard/src/alpaka-group/alpaka/test/unit/math/src/TestTemplate.hpp:145: FAILED:
  REQUIRE( isApproxEqual(results(i), std_result) )
with expansion:
  false
with messages:
  testing acc:alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, std::
  integral_constant<unsigned long, 1ul>, unsigned long> data type:double
  functor:mathtest::OpIsnan seed:2757112749
  Operator: OpIsnan
  Type: double
  The args buffer: 
  capacity: 1000
  0: [ 0 ]
  1: [ nan ]
  2: [ nan ]
  3: [ inf ]
  4: [ -inf ]
  5: [ -664.7846836187099 ]
  6: [ 56.67392124865512 ]
...
  Idx i: 1 computed : 1 vs expected: 0
...

===============================================================================
test cases:    290 |    284 passed | 6 failed
assertions: 262136 | 262130 passed | 6 failed
```

The six failures are for `OpIsnan`, `OpIsinf`, `OpIsfinite`, times two for `float` and `double`.

The full log is attached: [mathTest.log](https://github.com/user-attachments/files/18583190/mathTest.log) .

fwyzard:
In CMS code we use checks based on the IEEE float and double representation: [FWCore/Utilities/interface/isFinite.h](https://github.com/cms-sw/cmssw/blob/master/FWCore/Utilities/interface/isFinite.h) .

Those are pretty old, so we could improve them for c++20 (`std::bit_cast`, _etc_) and adapt them for alpaka.

mehmetyusufoglu:
[...]

PR #2482, opened 2025-03-13
alpaka::math custom implementations for CPU backend

mehmetyusufoglu:
Custom implementations are provided for 3 math functions, since the math functions like` isfinite, isnan, isinf `that rely on their std counterparts for CPU code fail when the code is built with `-Ofast, -ffast-math` or `-ffinite-math`.
For detailed definition:
 See #2466 

- [x]  Wait for the solution of #2486 
- [x] Wait for #2489

### DIFF
--- a/include/alpaka/math/MathStdLib.hpp
+++ b/include/alpaka/math/MathStdLib.hpp
@@ -8,6 +8,10 @@
 #include "alpaka/core/Decay.hpp"
 #include "alpaka/math/Traits.hpp"
 
+#include <bit>
+#include <cstdint>
+#include <type_traits>
+
 namespace alpaka::math
 {
     //! The standard library abs, implementation covered by the general template.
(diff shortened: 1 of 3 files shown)


include/alpaka/math/MathStdLib.hpp
```diff
@@ -294,6 +294,90 @@ namespace alpaka::math
...
+        //! Custom IEEE 754 bitwise implementation of isfinite
+        template<typename TArg>
+        struct Isfinite<IsfiniteStdLib, TArg, std::enable_if_t<std::is_floating_point_v<TArg>>>
+        {
+            auto operator()(IsfiniteStdLib const& /* ctx */, TArg const& arg) -> bool
+            {
+                if constexpr(std::is_same_v<TArg, float>)
+                {
+                    constexpr std::uint32_t expMask = 0x7F80'0000;
+                    std::uint32_t bits = std::bit_cast<std::uint32_t>(arg);
```
psychocoderHPC, line 312: include `<bit>` is missing

psychocoderHPC (changes requested):

mehmetyusufoglu:
This PR waits for the solution of issue #2486, since CI compilation using Clang13 fails since std::bit_cast is not supported by Clang13 and libc++-13.

mehmetyusufoglu:
@psychocoderHPC  this can be merged i think.

psychocoderHPC (approved):

PR #2490, opened 2025-04-02
Implement creation of view for generic contiguous range 

sbaldu:
This PR generalizes the construction of host views from standard containers to generic contiguous ranges using concepts.


AuroraPerego:
There is a failing workflow with CUDA 12.5 and GCC 10, is it expected after #2489?

mehmetyusufoglu:
> There is a failing workflow with CUDA 12.5 and GCC 10, is it expected after #2489?

GCC 10 is removed because it does not support some C++20 features like bit header. Please rebase the branch.

fwyzard:
@sbaldu can you rebase your branch to the current `develop` branch, in order to pick up the update to the CI configuration ?

include/alpaka/mem/view/ViewSeqRange.hpp
```diff
@@ -0,0 +1,93 @@
...
+{
+
+    namespace detail
+    {
+        template<typename T>
+        concept ContiguousContainer = requires(T t) {
+            t.size();
+            t.data();
+            requires std::ranges::contiguous_range<T>;
+        };
```
AuroraPerego, line 21:
```suggestion
        template<typename T>
        concept ContiguousContainer =
            std::ranges::contiguous_range<T> &&
            requires(T t) {
                t.size();
                t.data();
            };
```
can you check if something like this helps with the error you are getting about the constraint depending on itself?

sbaldu: Same error

fwyzard:
OK, my interpretation of the error is:
  - in `ViewSeqRange.hpp`, `ContiguousContainer<T>` requires `t.data()`
  - for alpala buffers and views, `t.data()` is implemented inheriting from `ViewAccessOps<T>` _if_ `isView<T>`
  - `isView<T>` "requires" `getPtrNative(t)`
  - in `ViewSeqRange.hpp`, `getPtrNative(t)` requires `ContiguousContainer<T>`

So we end up with a circular dependency of `ContiguousContainer<T>` on itself :-/

fwyzard:
More simply put:
  - alpaka implements `t.data()` calling `getPtrNative(t)`
  - your code implements `getPtrNative(t)` calling `t.data()`

🤷🏻

PR #2492, opened 2025-04-07
Implement cache of device properties

sbaldu:
This PR caches the device properties once they have been accessed so as to reduce overheads for multiple consecutive calls.

### DIFF
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -6,7 +6,7 @@
 #pragma once
 
 #include "alpaka/dev/Traits.hpp"
-#include "alpaka/dev/common/QueueRegistry.hpp"
+#include "alpaka/dev/common/DevGenericImpl.hpp"
 #include "alpaka/dev/cpu/SysInfo.hpp"
 #include "alpaka/mem/buf/Traits.hpp"
 #include "alpaka/platform/Traits.hpp"
@@ -55,7 +55,7 @@ namespace alpaka
     namespace cpu::detail
     {
         //! The CPU device implementation.
-        using DevCpuImpl = alpaka::detail::QueueRegistry<cpu::ICpuQueue>;
+        using DevCpuImpl = alpaka::detail::DevGenericImpl<cpu::ICpuQueue>;
     } // namespace cpu::detail
 
     //! The CPU device handle.
--- a/include/alpaka/dev/DevCpu.hpp
+++ b/include/alpaka/dev/DevCpu.hpp
@@ -6,6 +6,7 @@
 #pragma once
 
 #include "alpaka/dev/Traits.hpp"
+#include "alpaka/dev/common/DeviceProperties.hpp"
 #include "alpaka/dev/common/QueueRegistry.hpp"
 #include "alpaka/dev/cpu/SysInfo.hpp"
 #include "alpaka/mem/buf/Traits.hpp"
@@ -57,7 +58,9 @@ namespace alpaka
         friend struct trait::GetDevByIdx<PlatformCpu>;
 
     protected:
-        DevCpu() : m_spDevCpuImpl(std::make_shared<cpu::detail::DevCpuImpl>())
+        DevCpu()
+            : m_spDevCpuImpl(std::make_shared<cpu::detail::DevCpuImpl>())
+            , m_deviceProperties(std::make_shared<alpaka::DeviceProperties>())
         {
         }
 
@@ -89,8 +92,15 @@ namespace alpaka
             return 0;
         }
 
+        friend struct trait::GetName<DevCpu>;
+        friend struct trait::GetMemBytes<DevCpu>;
+        friend struct trait::GetFreeMemBytes<DevCpu>;
+        friend struct trait::GetWarpSizes<DevCpu>;
+        friend struct trait::GetPreferredWarpSize<DevCpu>;
+
     private:
         std::shared_ptr<cpu::detail::DevCpuImpl> m_spDevCpuImpl;
+        std::shared_ptr<alpaka::DeviceProperties> m_deviceProperties;
     };
 
     namespace trait
@@ -99,9 +109,13 @@ namespace alpaka
         template<>
         struct GetName<DevCpu>
         {
-            ALPAKA_FN_HOST static auto getName(DevCpu const& /* dev */) -> std::string
+            ALPAKA_FN_HOST static auto getName(DevCpu const& dev) -> std::string
             {
-                return cpu::detail::getCpuName();
+                if(!dev.m_deviceProperties->name.has_value())
+                {
+                    dev.m_deviceProperties->name = cpu::detail::getCpuName();
+                }
+                return dev.m_deviceProperties->name.value();
             }
         };
 
@@ -109,9 +123,13 @@ namespace alpaka
         template<>
         struct GetMemBytes<DevCpu>
         {
-            ALPAKA_FN_HOST static auto getMemBytes(DevCpu const& /* dev */) -> std::size_t
+            ALPAKA_FN_HOST static auto getMemBytes(DevCpu const& dev) -> std::size_t
             {
-                return cpu::detail::getTotalGlobalMemSizeBytes();
+                if(!dev.m_deviceProperties->totalGlobalMem.has_value())
+                {
+                    dev.m_deviceProperties->totalGlobalMem = cpu::detail::getTotalGlobalMemSizeBytes();
+                }
+                return dev.m_deviceProperties->totalGlobalMem.value();
             }
         };
 
@@ -119,9 +137,13 @@ namespace alpaka
         template<>
         struct GetFreeMemBytes<DevCpu>
         {
-            ALPAKA_FN_HOST static auto getFreeMemBytes(DevCpu const& /* dev */) -> std::size_t
+            ALPAKA_FN_HOST static auto getFreeMemBytes(DevCpu const& dev) -> std::size_t
             {
-                return cpu::detail::getFreeGlobalMemSizeBytes();
+                if(!dev.m_deviceProperties->freeGlobalMem.has_value())
+                {
+                    dev.m_deviceProperties->freeGlobalMem = cpu::detail::getFreeGlobalMemSizeBytes();
+                }
+                return dev.m_deviceProperties->freeGlobalMem.value();
             }
         };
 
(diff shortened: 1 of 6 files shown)


include/alpaka/dev/DevCpu.hpp
```diff
@@ -89,8 +93,16 @@ namespace alpaka
...
+        friend struct trait::GetName<DevCpu>;
+        friend struct trait::GetMemBytes<DevCpu>;
+        friend struct trait::GetFreeMemBytes<DevCpu>;
+        friend struct trait::GetWarpSizes<DevCpu>;
+        friend struct trait::GetPreferredWarpSize<DevCpu>;
+
     private:
         std::shared_ptr<cpu::detail::DevCpuImpl> m_spDevCpuImpl;
+        std::shared_ptr<alpaka::DeviceProperties> m_deviceProperties;
+        std::shared_ptr<std::mutex> m_mutex;
```
fwyzard: Can you move these into the `cpu::detail::DevCpuImpl` class ?

sbaldu: Yes, done

SimeonEhrig:
@sbaldu The AMD CI runner has driver problems. I talk to the support. Don't care about the failed HIP runtime tests for the moment.

sbaldu:
> @sbaldu The AMD CI runner has driver problems. I talk to the support. Don't care about the failed HIP runtime tests for the moment.

Ok thanks. I see that the rest of the tests are passing.

SimeonEhrig:
> > @sbaldu The AMD CI runner has driver problems. I talk to the support. Don't care about the failed HIP runtime tests for the moment.
> 
> Ok thanks. I see that the rest of the tests are passing.

The problem still exist. I'm on it. It takes a little bit longer, because the administrator is in vacations and I'm working together with another administrator which overtakes it temporary.

SimeonEhrig:
@sbaldu Sorry, I did miss take. Your code has a bug. I can reproduce it locally.
It looks like a well know bug in the CI but the CI is working. I mixed up this PR with your other PR #2490 and therefore I thought your changes cannot break the tests. But the changes of the code of this PR looks like, that they can produce a dead look.

sbaldu:
> @sbaldu Sorry, I did miss take. Your code has a bug. I can reproduce it locally. It looks like a well know bug in the CI but the CI is working. I mixed up this PR with your other PR #2490 and therefore I thought your changes cannot break the tests. But the changes of the code of this PR looks like, that they can produce a dead look.

Ok thanks for the update. I'll fix it asap.

include/alpaka/dev/DevCpu.hpp
```diff
@@ -99,19 +105,35 @@ namespace alpaka
...
-            ALPAKA_FN_HOST static auto getMemBytes(DevCpu const& /* dev */) -> std::size_t
+            ALPAKA_FN_HOST static auto getMemBytes(DevCpu const& dev) -> std::size_t
             {
-                return cpu::detail::getTotalGlobalMemSizeBytes();
+                auto& totalGlobalMem = dev.m_spDevCpuImpl->deviceProperties().totalGlobalMem;
+                {
+                    std::lock_guard<std::mutex> lock(dev.m_spDevCpuImpl->mutex());
+                    if(!totalGlobalMem.has_value())
+                    {
+                        totalGlobalMem = cpu::detail::getTotalGlobalMemSizeBytes();
```
psychocoderHPC: suggestion to avoid any locks.
```C++
    template<>
    struct GetMemBytes<DevCpu>
    {
        ALPAKA_FN_HOST static auto getMemBytes(DevCpu const& dev) -> std::size_t
        {

            static auto totalGlobalMem = cpu::detail::getTotalGlobalMemSizeBytes();
            return totalGlobalMem.value();
        }
    };
```

psychocoderHPC (dismissed):

include/alpaka/dev/DevCpu.hpp
fwyzard: Yes, it can use a `static` local variable for the CPU case, because we assume that there is a single CPU device anyway.
For the other devices we still need to use a lock, or maybe an [`std::once_flag`](https://en.cppreference.com/w/cpp/thread/once_flag) with [`std::call_once`](https://en.cppreference.com/w/cpp/thread/call_once) that is a bit more lightweight.

sbaldu: Following on what @fwyzard said, with the static variable solution wouldn't we have problems if a user uses CPUs as accelerators and has more than one CPU? Is such a use case possible?
For the other backends I've updated using `std::call_once` as suggested.

fwyzard: At least for the moment alpaka assumes that there exist only a single CPU device, even on a machine with multiple sockets and many cores, and even when it is used by multiple backends (serial, tbb, openmp, etc).

So the `static` approach is fine for now.

sbaldu: Ok I see, thank you

psychocoderHPC: note: I missed that this is a per device property. I aggry with @fwyzard that as long as alpaka only supports a single complete CPU it is fine to use `static`

include/alpaka/dev/DevCpu.hpp
```diff
@@ -99,19 +105,36 @@ namespace alpaka
...
             {
-                return cpu::detail::getCpuName();
+                auto& devProperties = dev.m_spDevCpuImpl->deviceProperties();
+                std::call_once(
+                    dev.m_spDevCpuImpl->onceFlag(),
+                    [&]() noexcept
+                    {
+                        devProperties->name = cpu::detail::getCpuName();
+                        devProperties->totalGlobalMem = cpu::detail::getTotalGlobalMemSizeBytes();
+                    });
```
fwyzard: here, and for the other backends: I would suggest to wrap the common code use to read all device properties in a function, and call it from multiple places, instead of duplicating the code.

include/alpaka/dev/DevGenericSycl.hpp
```diff
@@ -112,11 +113,23 @@ namespace alpaka
                 return m_context;
             }
 
+            std::once_flag& onceFlag()
+            {
+                return m_onceFlag;
+            }
```
fwyzard: While I'm not a fan of that syntax, we should be consistent with the rest of the code.
Can you change the signature of the function to
```diff
-            std::once_flag& onceFlag()
+            auto onceFlag() -> std::once_flag&
```
?

sbaldu: Of course, done

fwyzard:
Now I'm looking at parts of alpaka I never read before :-D

I understand why you added the properties and flag to the `QueueRegistry` class - the CPU and CUDA/HIP devices have a `shared_ptr` to an instance of that class.

From a semantic (?) point of view, something called a `QueueRegistry` does not seem to be the right place for it :-/

@psychocoderHPC what do you think ?
I would rather avoid adding a second shared state, so reusing the `QueueRegistry` seems to make sense.
Maybe we could rename it to something more in line with the added features ?
What does this code look like in alpaka 3 ?

fwyzard:
Instead, from a more technical point of view, I would prefer to move more of the common code to a reusable function, to avoid repetitions.

Could you try to move the `setDeviceProperties` directly in the class that owns the `once_flag` and the `DeviceProperties`, at least for the SYCL case ?

For the CPU and CUDA/HIP cases it seems more complicated.

include/alpaka/dev/DevUniformCudaHipRt.hpp
```diff
@@ -108,18 +116,58 @@ namespace alpaka
 
     namespace trait
     {
+
+        template<typename TApi>
+        static inline void setDeviceProperties(
```
fwyzard: This should not be `static`:
```suggestion
        inline void setDeviceProperties(
```

include/alpaka/dev/DevGenericSycl.hpp
```diff
@@ -154,14 +167,51 @@ namespace alpaka
 
     namespace trait
     {
+
+        template<concepts::Tag TTag>
+        static inline void setDeviceProperties(
```
fwyzard:
```suggestion
        inline void setDeviceProperties(
```

include/alpaka/dev/common/QueueRegistry.hpp
```diff
@@ -5,14 +5,21 @@
...
 #include <memory>
 #include <mutex>
+#include <optional>
 
 namespace alpaka::detail
 {
+
+    inline void setDeviceProperties(int iDevice, alpaka::DeviceProperties& devProperties);
+    template<typename TApi>
+    inline void setDeviceProperties(int iDevice, alpaka::DeviceProperties& devProperties);
```
fwyzard: I wonder if this could be declared as
```
    template<typename TDev>
    inline void setDeviceProperties(TDev const& device, alpaka::DeviceProperties& devProperties);
```
and then specialised for the `DevCpu` and `DevUniformCudaHipRt<TApi>` cases ?

That is, passing the device as an argument instead of the index.

include/alpaka/dev/DevGenericSycl.hpp
```diff
@@ -112,11 +113,41 @@ namespace alpaka
...
             }
 
+            auto deviceProperties() -> std::optional<alpaka::DeviceProperties>&
+            {
+                std::call_once(
+                    m_onceFlag,
+                    [&]()
+                    {
+                        m_deviceProperties = std::make_optional<alpaka::DeviceProperties>();
+                        auto const device = this->get_device();
```
fwyzard: can this be a `const&`
```suggestion
                        auto const& device = this->get_device();
```
?

sbaldu:
ok now all the tests seem to be passing.
@SimeonEhrig @psychocoderHPC I renamed the `QueueRegistry` as `DevGenericImpl`, which is similar to the naming used for SYCL, is that ok?

include/alpaka/dev/common/DevGenericImpl.hpp
```diff
@@ -5,19 +5,33 @@
...
+    template<typename TApi>
+    class DevUniformCudaHipRt;
+} // namespace alpaka
 
 namespace alpaka::detail
 {
+    inline void setDeviceProperties(alpaka::DevCpu const&, alpaka::DeviceProperties&);
+    template<typename TApi>
+    inline void setDeviceProperties(alpaka::DevUniformCudaHipRt<TApi> const&, alpaka::DeviceProperties&);
+
```
fwyzard: Why do you need to declare these here ?

AuroraPerego (approved):

PR #2493, opened 2025-04-10
[DO NOT MERGE] trigger AMD CI

SimeonEhrig:
Only for testing the CI


SimeonEhrig:
Was not CI error. Found the correct reason for the failing CI.

PR #2498, opened 2025-05-13
Implement support for explicit GPU kernel names

fwyzard:
Introduce two macros `ALPAKA_KERNEL_SCOPED_NAME(KERNEL, NAMESPACE, NAME)` and `ALPAKA_KERNEL_NAME(KERNEL, NAME)` to define explicit names for GPU kernels.

A GPU kernel normally looks like `alpaka::detail::gpuKernel<Kernel, Acc, ...>`.
Using the macro
```c++
ALPAKA_KERNEL_NAME(Kernel, kernel)
```
makes it appear as `kernel<Acc, ...>`.

The macros can only be used in the global namespace. Using them in another namespaces causes a static assertion.

Add a test for the new functionality.

### DIFF
--- a/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
+++ b/include/alpaka/kernel/TaskKernelGpuUniformCudaHipRt.hpp
@@ -78,6 +78,14 @@ namespace alpaka
 #        if BOOST_COMP_CLANG
 #            pragma clang diagnostic pop
 #        endif
+
+        template<typename TKernelFnObj, typename TAcc, typename... TArgs>
+        inline void (*kernelName)(
+            Vec<Dim<TAcc>, Idx<TAcc>> const,
+            TKernelFnObj const,
+            remove_restrict_t<std::decay_t<TArgs>>...)
+            = gpuKernel<TKernelFnObj, TAcc, TArgs...>;
+
     } // namespace detail
 
     namespace uniform_cuda_hip
@@ -248,7 +256,7 @@ namespace alpaka
 #        endif
 
                 auto kernelName
-                    = alpaka::detail::gpuKernel<TKernelFnObj, TAcc, remove_restrict_t<std::decay_t<TArgs>>...>;
+                    = alpaka::detail::kernelName<TKernelFnObj, TAcc, remove_restrict_t<std::decay_t<TArgs>>...>;
 
 #        if ALPAKA_DEBUG >= ALPAKA_DEBUG_FULL
                 // Log the function attributes.
@@ -317,7 +325,7 @@ namespace alpaka
                 [[maybe_unused]] TKernelFn const& kernelFn,
                 [[maybe_unused]] TArgs&&... args) -> alpaka::KernelFunctionAttributes
             {
-                auto kernelName = alpaka::detail::gpuKernel<
+                auto kernelName = alpaka::detail::kernelName<
                     TKernelFn,
                     AccGpuUniformCudaHipRt<TApi, TDim, TIdx>,
                     remove_restrict_t<std::decay_t<TArgs>>...>;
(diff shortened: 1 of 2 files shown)


SimeonEhrig:
I assume the PR should improve the output of compiler errors and profiler output?

fwyzard:
(rebased to fix the conflicts)

fwyzard:
> I assume the PR should improve the output of compiler errors and profiler output?

Yes, the idea is that one can give explicit names to the kernels, and these will show up in tools like a CUDA GDB, Nsight Systems, _etc._

SimeonEhrig:
> Yes, the idea is that one can give explicit names to the kernels, and these will show up in tools like a CUDA GDB, Nsight Systems, _etc._

Okay, I will check it during the day. I had a similar idea, but with an compile time string as template parameter to avoid macros. The problem is, that not each tool can display it as string. Some tools display something like `(char)75, (char)101, ...` instead `Kernelname`.

SimeonEhrig:
> @fwyzard For clarification, the feature is optional? 
> For example a kernel is not automatically renamed if the kernel is passed to `exec` function.

Yes, correct.

> Also we "annotate" an kernel via global macro. We do not create a new object/type from our kernel object, which is renamed and pass it to the `exec` function.

Yes, correct.

The macro defines a dedicated function that is used to launch the alpaka kernel, because form the CUDA point of view, this function _is_ the kernel.

SimeonEhrig:
Example with `ncu`.

Default kernel without name:
```bash
 void alpaka::detail::gpuKernel<DefaultKernel, alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, std::integral_constant<unsigned long, (unsigned long)1>, long>, bool *, int>(alpaka::Vec<alpaka::trait::DimType<T2, void>::type, alpaka::trait::IdxType<T2, void>::type>, T1, T3...) (1, 1, 1)x(1, 1, 1), Context 1, Stream 14, Device 0, CC 8.0
    Section: GPU Speed Of Light Throughput
    ----------------------- ------------- ------------
    Metric Name               Metric Unit Metric Value
    ----------------------- ------------- ------------
    DRAM Frequency          cycle/nsecond         1.19
    SM Frequency            cycle/usecond       909.41
    Elapsed Cycles                  cycle        2,591
    Memory Throughput                   %         0.55
    DRAM Throughput                     %         0.12
    Duration                      usecond         2.85
    L1/TEX Cache Throughput             %        28.46
    L2 Cache Throughput                 %         0.55
    SM Active Cycles                cycle        14.05
    Compute (SM) Throughput             %         0.00
    ----------------------- ------------- ------------
```

Kernel with name:
```bash
  void kernelWithName<alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, std::integral_constant<unsigned long, (unsigned long)3>, long>, bool *, int>(alpaka::Vec<alpaka::trait::DimType<T1, void>::type, alpaka::trait::IdxType<T1, void>::type>, KernelWithNam
e, T2...) (1, 1, 1)x(1, 1, 1), Context 1, Stream 42, Device 0, CC 8.0
    Section: GPU Speed Of Light Throughput
    ----------------------- ------------- ------------
    Metric Name               Metric Unit Metric Value
    ----------------------- ------------- ------------
    DRAM Frequency          cycle/nsecond         1.20
    SM Frequency            cycle/usecond       914.38
    Elapsed Cycles                  cycle        2,928
    Memory Throughput                   %         0.49
    DRAM Throughput                     %         0.10
    Duration                      usecond         3.20
    L1/TEX Cache Throughput             %        19.91
    L2 Cache Throughput                 %         0.49
    SM Active Cycles                cycle        20.09
    Compute (SM) Throughput             %         0.00
    ----------------------- ------------- ------------
```


Kernel with name and scope:
```bash
  void scope::kernelWithName<alpaka::AccGpuUniformCudaHipRt<alpaka::ApiCudaRt, std::integral_constant<unsigned long, (unsigned long)1>, long>, bool *, int>(alpaka::Vec<alpaka::trait::DimType<T1, void>::type, alpaka::trait::IdxType<T1, void>::type>, KernelWithScopedName, T2...) (1, 1, 1)x(1, 1, 1), Context 1, Stream 62, Device 0, CC 8.0
    Section: GPU Speed Of Light Throughput
    ----------------------- ------------- ------------
    Metric Name               Metric Unit Metric Value
    ----------------------- ------------- ------------
    DRAM Frequency          cycle/nsecond         1.19                                                                                                                                                                                                             SM Frequency            cycle/usecond       913.20
    Elapsed Cycles                  cycle        2,923
    Memory Throughput                   %         0.49
    DRAM Throughput                     %         0.10                                                                                                                                                                                                             Duration                      usecond         3.20
    L1/TEX Cache Throughput             %        19.88
    L2 Cache Throughput                 %         0.49
    SM Active Cycles                cycle        20.12
    Compute (SM) Throughput             %         0.00
    ----------------------- ------------- ------------
```

SimeonEhrig (approved):

SimeonEhrig:
Takes me a moment to understand the mechanism. You use a function pointer to point to the generic implementation of the wrapper kernel, which has a generic name. The macro specialize the wrapper kernel and give the `__globale__` function a good name. With the template specialization rule, the function pointer points to the name function wrapper. Cool idea.

fwyzard:
Thanks :-)

Initially I hoped to make the `__global__` function wrapper a method of the "kernel" `struct` itself - but CUDA allows only free functions to be `__global__` :-/

PR #2503, opened 2025-06-10
remove `BoostPredef.hpp` by `Config.hpp`

psychocoderHPC:
With this PR boost becomes an optional dependency.

- [x] rebase after #2472 because it is included

- use `ALPAKA_*` definitions instead of `BOOST_*`
- remove code parts for not supported INTEL compiler
- remove `BOOST_OS_UNIX` usage by `ALPAKA_OS_LINUX`
- remove `BOOST_COMP_MSVC_EMULATED` since alpaka only supports MSVC
  under windows
- only load boost as dependency if `std::atomic_ref` is not found.

This PR is **not** removing boost for now from the CI. This will be part of an independent PR, we need to enable boost as dependency


include/alpaka/core/Config.hpp
```diff
@@ -0,0 +1,230 @@
...
+#    else
+#        define ALPAKA_LANG_HIP ALPAKA_VERSION_NUMBER_NOT_AVAILABLE
+#    endif
+#endif
+
+// CUDA
+#if !defined(ALPAKA_LANG_CUDA)
+#    if defined(__CUDACC__) || defined(__CUDA__)
+#        include <cuda.h>
+// HIP doesn't give us a patch level for the last entry, just a gitdate
```
fwyzard:
```suggestion
// CUDA doesn't give us a patch level for the last entry, just zero.
```

psychocoderHPC:
@fwyzard this PR can be reviewed

include/alpaka/core/Config.hpp
```diff
@@ -0,0 +1,230 @@
...
+
+#pragma once
+
+#ifdef __INTEL_COMPILER
+#    warning                                                                                                          \
+        "The Intel Classic compiler (icpc) is no longer supported. Please upgrade to the Intel LLVM compiler (ipcx)."
+#endif
+
+#define ALPAKA_VERSION_NUMBER(major, minor, patch)                                                                    \
+    ((((major) % 1000) * 100'000'000) + (((minor) % 1000) * 100000) + ((patch) % 100000))
```
fwyzard: This will overflow if `major` is 22 or greater and is not explicitly marked as `unsigned int` or a `long`.
Would it make sense to force the promotion to `long` or `unsigned long` ?
_e.g._
```suggestion
    ((((major) % 1000ul) * 100'000'000ul) + (((minor) % 1000ul) * 100'000ul) + ((patch) % 100'000ul))```

File: include/alpaka/core/Config.hpp [hunk 3]
```diff
@@ -0,0 +1,230 @@
...
+// ARM
+#if !defined(ALPAKA_ARCH_ARM)
+#    if defined(__ARM_ARCH) || defined(__arm__) || defined(__arm64)
+#        define ALPAKA_ARCH_ARM 1
+#    else
+#        define ALPAKA_ARCH_ARM 0
+#    endif
+#endif
+
+// ARM
```
fwyzard:
```suggestion
// NVIDIA GPUs
```

include/alpaka/core/Config.hpp
```diff
@@ -0,0 +1,230 @@
...
+#if !defined(ALPAKA_ARCH_PTX)
+#    if defined(__CUDA_ARCH__)
+#        define ALPAKA_ARCH_PTX 1
+#    else
+#        define ALPAKA_ARCH_PTX 0
+#    endif
+#endif
```
fwyzard, line 100: Could you add an equivalent flag for AMD GPUs ?

include/alpaka/core/Config.hpp
psychocoderHPC: good catch, I will use you suggestion

include/alpaka/core/Config.hpp
psychocoderHPC: done ALPAKA_ARCH_HSA

psychocoderHPC:
@fwyzard I pushed the changes

include/alpaka/core/Config.hpp
fwyzard: 👍🏻

fwyzard (approved):
The changes look good to me.

fwyzard:
... though the CI seems to disagree 🤷🏻

fwyzard:
The error
```
/usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/source_location(53): error: identifier "__builtin_source_location" is undefined
      using __builtin_ret_type = decltype(__builtin_source_location());
```
seems pretty much unrelated to this PR, and likely related to the new demangler.

But then it's weird that the tests pass in #2472 🤔

SimeonEhrig:
> The error
> 
> ```
> /usr/bin/../lib/gcc/x86_64-linux-gnu/11/../../../../include/c++/11/source_location(53): error: identifier "__builtin_source_location" is undefined
>       using __builtin_ret_type = decltype(__builtin_source_location());
> ```
> 
> seems pretty much unrelated to this PR, and likely related to the new demangler.
> 
> But then it's weird that the tests pass in #2472 🤔

The explanation is, that the PR #2472 didn't generated this test job, which failed. This is a know bug, that pair-wise generator uses an random number generator. Unfortunately, I have no idea where and how to fix the seed. This is actual weird, that the generator can be hidden in an open-source project.

In both cases, there is a combination of nvcc 12.1 with llvm 15. There are some different test parameters. I think we should first check the build type. The job which passed uses a release build. The job which fails is a debug build.
- passed: https://gitlab.com/hzdr/crp/alpaka/-/jobs/10308430091
- failed: https://gitlab.com/hzdr/crp/alpaka/-/jobs/10334046953

psychocoderHPC:
rebased against #2508 to fix the ci error

fwyzard (approved):

SimeonEhrig (approved):

PR #2505, opened 2025-06-10
CI: remove CUDA 11.2 from gihub action

psychocoderHPC:
Use CUDA 12.1 with clang 18 instead of CUDA 11.2.
CUDA 12.1 is the latest version supported by clang 18. Clang 18 is the first compiler supports CUDA 12+.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ concurrency:
 # alpaka_ACC_ANY_BT_OMP5_ENABLE                 : {ON, OFF}
 #   [ON] OMP_NUM_THREADS                        : {1, 2, 3, 4}
 # alpaka_ACC_GPU_CUDA_ENABLE                    : {ON, OFF}
-#   [ON] ALPAKA_CI_CUDA_VERSION                 : {11.2, 11.3, 11.4, 11.5, 11.6, 11.7, 11.8, 12.0, 12.1, 12.2, 12.3, 12.4, 12.5, 12.6}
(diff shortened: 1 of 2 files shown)


SimeonEhrig:
How was this actually working? CUDA 11.x does not support C++20?

psychocoderHPC:
> How was this actually working? CUDA 11.x does not support C++20?

nvcc <12 does not support C++20 but all CUDA code is compiled with clang and clang supports C++20 therefore it works.

SimeonEhrig:
Clang 18 does not support CUDA 12.3. It only supports 12.1: https://github.com/alpaka-group/bashi/blob/a41fdfbcbf003cce70c077e8d36ea5f2a74ab424/src/bashi/versions.py#L282 (Clang 18 is not part of the list, because it supports the same CUDA versions like Clang 17).

psychocoderHPC:
Mhhh PIConGPU is saying something else https://github.com/ComputationalRadiationPhysics/picongpu/blob/041cb1bef683d936785aea9cb07ca9a302e16e47/share/ci/n_wise_generator.py#L149, I will test if it works with 12.1

SimeonEhrig:
Looks like Clang-CUDA Debug builds are broken in general. Try this workaround for the CMake: https://github.com/llvm/llvm-project/issues/58491#issuecomment-2513654931

psychocoderHPC:
@SimeonEhrig I forgot ro rebase to habe the ulimit fix in. Will push the pr again asap

SimeonEhrig (approved):

PR #2507, opened 2025-06-11
Install cuBLAS and add CUDA-only mode job to the CI

fwyzard:
Follow up to #2480, related to https://github.com/alpaka-group/alpaka/pull/2479#issue-2857604635.

### DIFF
--- a/script/gitlabci/job_cuda.yml
+++ b/script/gitlabci/job_cuda.yml
@@ -33,3 +33,30 @@ linux_nvcc12.0_gcc11_release_extended_lambda_off_compile_only:
     alpaka_CUDA_SHOW_REGISTER: "OFF"
     alpaka_CUDA_KEEP_FILES: "OFF"
     alpaka_CUDA_EXPT_EXTENDED_LAMBDA: "OFF"
+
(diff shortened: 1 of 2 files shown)


SimeonEhrig:
This will not work with this image. CUDA is already pre-installed. You need to move the install instructions outside the `if agc-manager` state.
https://github.com/alpaka-group/alpaka/blob/64a7f3509bf729a007685b163cc4947f389ebbf8/script/install_cuda.sh#L25

SimeonEhrig:
Please also add the line `CI_FILTER: linux_nvcc12.6_gcc13_release_cuda_only_mode` to your commit message to run only the cuda-only job.

see here: https://github.com/alpaka-group/alpaka/pull/2480/commits/7b927d95430227cb2f57eec0e98d48cc31ff1f6a

fwyzard:
I actually removed it on purpose, to check that the change would not break the other jobs.

SimeonEhrig:
> I actually removed it on purpose, to check that the change would not break the other jobs.

Other strategy. If the CI pass, we don't need to run again.

SimeonEhrig:
The code was merged with PR #2480

PR #2508, opened 2025-06-13
fix compile when including `std::source_location`

psychocoderHPC:
fix CI compile issue with nvcc + clang15 as host compiler + MDSpan

```
error: `include/c++/11/source_location(53): error: identifier "__builtin_source_location" is undefined`
```

### DIFF
--- a/include/alpaka/core/DemangleTypeNames.hpp
+++ b/include/alpaka/core/DemangleTypeNames.hpp
@@ -11,8 +11,15 @@
 #pragma once
 
 #include <array>
-#ifdef __cpp_lib_source_location
-#    include <source_location>
+/** The following compilers does not use std::source_location therefore we do not include the header.
+ * Including the header source_location would lead into compiler issues under the following conditions.
+ *  - nvcc is used with clang 15 as host compiler and MDSpan is used too.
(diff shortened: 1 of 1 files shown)


include/alpaka/core/DemangleTypeNames.hpp
```diff
@@ -11,8 +11,15 @@
 #pragma once
 
 #include <array>
-#ifdef __cpp_lib_source_location
-#    include <source_location>
+/** The following compilers does not use std::source_loaction therefore we do not include the header.
```
mehmetyusufoglu:
```suggestion
/** The following compilers does not use std::source_location therefore we do not include the header.
```

fwyzard: OK ...
In fact, basically we are never using `std::source_location::current().function_name()`, just `__PRETTY_FUNCTION__` or `__FUNCSIG__` :-D

fwyzard (dismissed):

include/alpaka/core/DemangleTypeNames.hpp
psychocoderHPC: fixed the typo.

Yes only for unknown compilers we would fall back to std::source_location.

fwyzard (approved):

PR #2511, opened 2025-06-14
Fix `ALPAKA_THROW_ACC` for CUDA/HIP-only mode

fwyzard:
In CUDA-only or HIP-only mode, `ALPAKA_FN_ACC` functions are `__device__` functions, and do not support having a throw visible even during host compilation.

### DIFF
--- a/include/alpaka/core/RuntimeMacros.hpp
+++ b/include/alpaka/core/RuntimeMacros.hpp
@@ -7,18 +7,20 @@
 // Implementation details
 #include "alpaka/core/Sycl.hpp"
 
+#include <stdexcept>
+
 //! ALPAKA_THROW_ACC either aborts(terminating the program and creating a core dump) or throws std::runtime_error
-//! depending on the Acc. The std::runtime_error exception can be catched in the catch block.
+//! depending on the Acc. The std::runtime_error exception can be caught in the catch block.
 //!
-//! For CUDA __trap function is used which triggers std::runtime_error but can be catched during wait not exec.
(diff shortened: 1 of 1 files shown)


mehmetyusufoglu:
is this expected to prevent "host" compile error due to ALPAKA_THROW_ACC in case of Only-Cuda or Only-Hip backend? But we already don't get any error without this change? May be this condition will never work since it is at the end of the if else clauses. ( Cuda or Hip-Only necessitates cuda or hip backend cmake flags as well )

fwyzard:
We don't get any errors in the CI because we don't test in HIP-only mode.

In the default mode kernels are `__host__ __device__`, and the throw is allowed (at compile time).

mehmetyusufoglu:
Ok I tested at a computer(HAL) with amd/rocm/hip, **without this PR** there is a **compile error:**

```
/home/yusufo81/alpaka/test/unit/runtime/src/KernelThrow.cpp:25:13: error: cannot use 'throw' in __device__ function
            ALPAKA_THROW_ACC("Exception thrown by the kernel.");
            ^
/home/yusufo81/alpaka/include/alpaka/core/RuntimeMacros.hpp:50:13: note: expanded from macro 'ALPAKA_THROW_ACC'
            throw std::runtime_error(MSG);      
``` 
       
**with this PR:**  **No compile error** and no runtime assert; i expected.

```
Randomness seeded to: 690123303
test cases: 2 | 2 passed
assertions: - none
```

psychocoderHPC (approved):

PR #2516, opened 2025-06-17
alpaka kernels for matrix multiplication and addition

Akasxh:


fwyzard:
Hi Akash,
the example you propose to add looks like a more comprehensive version of the one at `example/useBLASInAlpaka/useCuBLASInAlpaka/src/useCuBLASInAlpaka.cpp`.

If you would like to make it more interesting, you could try to extend it so that
  - for the CUDA backend (`#if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)`), it uses cuBLAS
  - for the HIP/ROCm backend (`#if defined(ALPAKA_ACC_GPU_HIP_ENABLED)`), it uses rocBLAS
  - for the CPU backend(s) it uses OpenBLAS (or some other implementation, I'm really not familiar with it!)
  
?

Akasxh:
Yes, I'll make something more specific.

I'm looking for making few kernels such that it would work on all of them as you said (#ifdefine...)

SimeonEhrig:
Maybe we can combine it with some kine of API design example. Something like

```c++
template<typename TAcc, typename TQueueProperty, typename TBuffer, typename TElem /* = getElementTypeFromBuffer<TBuffer */>
void gemm(alpaka::Queue<TAcc, TQueueProperty> queue, TBuffer A, TBuffer B, TBuffer C, TElem alpha, TElem beta){
// generic alpaka kernel
}

template<typename TQueueProperty, typename TBuffer, typename TElem /* = getElementTypeFromBuffer<TBuffer */>
void gemm(alpaka::Queue<alpaka::AccGpuCudaRt, TQueueProperty> queue, TBuffer A, TBuffer B, TBuffer C, TElem alpha, TElem beta){
// use cuBlas
}
#endif
```

Would be pretty cool, if we have an example that shows how alpaka can be used for libraries and with specialization. This gives use also feedback, if we do API changes/extensions.

psychocoderHPC:
IMO it would make sense to extent the already exiting example https://github.com/alpaka-group/alpaka/tree/develop/example/useBLASInAlpaka and extent it, or add a second `*.cpp` file which is the code of the first plus additional work to demonstrate more integration.
This would have the advantage that we have not a exploding number of examples with a lot of repititions.

Akasxh:
Yes I am trying to do something similar. 

Also while I'm at it could you give me a more concise objective at this as I'm new to it.

Thanks

psychocoderHPC:
An objective should be oriented toward real-world use cases you try to solve. 
What is your use case, why you looked specifically into mixing native CUDA/HIP with alpaka?

Akasxh:
My use case as such is making ALPAKA kernels that can be implemented in CUDA, ROCm and SYCL for code generation for CERNs ROOT TMVA's SOFIE ( System for Optimised Fast Inference code Emit )

https://github.com/root-project/root/blob/master/tmva%2Fsofie%2FREADME.md

While doing so I'm essentially designing ALPAKA Kernels for all operations that are used in ML models like conv2d, relu, etc.

fwyzard:
@SimeonEhrig the CI failures look similar to what you observed in the CUDA-only tests in #2480:
> ```
> /builds/hzdr/crp/alpaka/example/matrixMulAdd/src/matrixMulAdd.cpp:12:10: fatal error: 'cublas_v2.h' file not found
> #include <cublas_v2.h>
>          ^~~~~~~~~~~~~
> 1 error generated.
> ```

Does the new test need some special care? Or maybe, does the CI need some further changes?

SimeonEhrig:
> @SimeonEhrig the CI failures look similar to what you observed in the CUDA-only tests in #2480:
> 
> > ```
> > /builds/hzdr/crp/alpaka/example/matrixMulAdd/src/matrixMulAdd.cpp:12:10: fatal error: 'cublas_v2.h' file not found
> > #include <cublas_v2.h>
> >          ^~~~~~~~~~~~~
> > 1 error generated.
> > ```
> 
> Does the new test need some special care? Or maybe, does the CI need some further changes?

There was a bug in the CUDA installer, which forgets to install cublas. We have also version 4.0 contains in the registry. I'm not sure if the problem still exist in 4.0 but this is my roadmap.
1. release alpaka 2.0
2. update to container 4.0 
3. if does not solve the problem, build container 4.0 again
4. if does not solve the problem, create container 4.1 which solves the problem (container 4.1 also removes pre installed boost)

fwyzard:
@Akasxh here are some suggestions for your project
  - as @psychocoderHPC suggested, move your example to `example/useBLASInAlpaka/`, and see if you can merge it with the existing one; if your example includes all features form the existing one, you can also just replace it/
  - in parallel, add an implementation for each back-end: CPU-only, HIP, SYCL.
  - once you have all implementations working, try to merge them: have a single `main()` that calls one back-end or the other using `#if defined ...` or `if constexpr (...)`
  - finally, wrap the GEMM calls into a generic `alpaka::example::sgemm()` function of functor, as @SimeonEhrig suggested, so that `main()` can just call `sgemm<Acc>(queue, ...)` and the function will dispatch to the correct back-end.

If you would like to have a chat about it, we can have a meeting on Zoom.
I'll send you the details in private.

example/matrixMulAdd/src/matrixMulAdd.cpp
```diff
@@ -0,0 +1,146 @@
...
+    auto const devHost = alpaka::getDevByIdx(platformHost, 0);
+    auto const platformAcc = alpaka::Platform<Acc>{};
+    auto const devAcc = alpaka::getDevByIdx(platformAcc, 0);
+
+    Queue queue(devAcc);
+
+    // 2x2 matrices stored in column major order
+    auto bufHostA = alpaka::allocBuf<Data, Idx>(devHost, 4u);
+    auto bufHostB = alpaka::allocBuf<Data, Idx>(devHost, 4u);
+    auto bufHostC1 = alpaka::allocBuf<Data, Idx>(devHost, 4u);
```
mehmetyusufoglu, line 49: 4 item array. Hence Kernel sums all items and only one thread is used ?

Akasxh: Sorry for the late reply, yes you are right also this in line 88:

`    alpaka::KernelCfg<Acc> cfgAdd{alpaka::Vec<Dim1D, Idx>(1), Idx{1}};
`

PR #2521, opened 2025-06-19
[1.3] Fix bablestream benchmark

fwyzard:
Backport #2420 and #2467.

Make kernels depend each other, use original variables and access order.
Limit babelstream DotKernel block size for multithreaded CPUs.

### DIFF
--- a/benchmarks/babelstream/src/babelStreamMainTest.cpp
+++ b/benchmarks/babelstream/src/babelStreamMainTest.cpp
@@ -1,32 +1,46 @@
-
 #include "babelStreamCommon.hpp"
-#include "catch2/catch_session.hpp"
 
 #include <alpaka/alpaka.hpp>
 #include <alpaka/test/acc/TestAccs.hpp>
 
 #include <catch2/benchmark/catch_benchmark.hpp>
+#include <catch2/catch_session.hpp>
 #include <catch2/catch_template_test_macros.hpp>
 #include <catch2/catch_test_macros.hpp>
 
+#include <algorithm>
+#include <chrono>
+#include <cmath>
+#include <cstdint>
+#include <iostream>
+#include <iterator>
+#include <numeric>
 #include <string>
+#include <type_traits>
+#include <vector>
 
 /**
- * Babelstream benchmarking example. Babelstream has 5 kernels. Add, Multiply, Copy, Triad and Dot.
- * Babelstream is a memory-bound benchmark since the main operation in the kernels has high Code Balance (bytes/FLOP)
- * value. For example c[i] = a[i] + b[i]; has 2 reads 1 writes and has one FLOP operation. For double precision each
- * read-write is 8 bytes. Hence Code Balance (3*8 / 1) = 24 bytes/FLOP.
+ * Babelstream benchmarking example. Babelstream has 5 kernels. Add, Multiply, Copy, Triad and Dot. NStream is
+ * optional. Init kernel is run before 5 standard kernel sequence. Babelstream is a memory-bound benchmark since the
+ * main operation in the kernels has high Code Balance (bytes/FLOP) value. For example c[i] = a[i] + b[i]; has 2 reads
+ * 1 writes and has one FLOP operation. For double precision each read-write is 8 bytes. Hence Code Balance (3*8 / 1) =
+ * 24 bytes/FLOP.
  *
  * Some implementations and the documents are accessible through https://github.com/UoB-HPC
  *
  * Can be run with custom arguments as well as catch2 arguments
- * Run with Custom arguments:
+ * Run with Custom arguments and for kernels: init, copy, mul, add, triad (and dot kernel if a multi-thread acc
+ * available):
  * ./babelstream --array-size=33554432 --number-runs=100
- * Runt with default array size and num runs:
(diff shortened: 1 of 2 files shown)


fwyzard:
I had to make a small change to replace the use of a template lambda (that is c++20-only) with a regular `if constexpr`.

benchmarks/babelstream/src/babelStreamMainTest.cpp
```diff
@@ -267,41 +313,82 @@ void testKernels()
...
+        kernelCfg,
+        devAcc,
+        NstreamKernel(),
+        bufAccInputAPtr,
+        bufAccInputBPtr,
+        bufAccOutputCPtr);
+
+
+    // Work division for Dot Kernel
+    alpaka::WorkDivMembers<Dim, Idx> workDivDot{Vec::all(1), Vec::all(1), Vec::all(1)};
```
psychocoderHPC, line 331: IN the current alpaka 2.0 this code part is different and e.g. on multi-thread accelerator, takes care that the the number of threads is multiple of two.
Maybe this is the part you mentioned in the total that you adjusted it.
This part could be the reason of the segfault with sycl.

```C++
// Lambda to create and return work division for dot kernel
    auto getWorkDivForDotKernel = [&]<typename AccType>() -> alpaka::WorkDivMembers<Dim, Idx>
    {
        // Use babelstream benchmark standard work division for multi-threaded backends
        // For GPUs the benchmark expects a fixed workdiv: {256,1024,1}
        // For multi-threaded CPUs, which are not part of the benchmark, the code below will limit the blocksize with
        // the limit of the backend.
        if constexpr(alpaka::isMultiThreadAcc<AccType>)
        {
            auto const kernelFunctionAttributes = alpaka::getFunctionAttributes<Acc>(
                devAcc,
                DotKernel(),
                bufAccInputAPtr,
                bufAccInputBPtr,
                bufAccOutputCPtr, // this is used here a kind of dummy
                static_cast<alpaka::Idx<AccType>>(arraySize));

            // Get the maxThreadPerBlock
            auto const maxThreadsPerBlock = kernelFunctionAttributes.maxThreadsPerBlock;
            // Threads per block is 1024 for benchmark, if the system does not allow use the max value
            auto threadsPerBlock = std::min(maxThreadsPerBlock, blockThreadExtentMain);

            // Reduce operation at dot-kernel needs even block size
            threadsPerBlock = (threadsPerBlock + 1) / 2 * 2;

            // Dot kernel is only used for benchmarking of GPU backends; and Work division is fixed for benchmark:
            // 256,1024,1. https://github.com/UoB-HPC/BabelStream/blob/main/src/cuda/CUDAStream.cu Hence blocksize
            // should be 1024 for GPU backends. But for multi-threaded CPUs; this code would also run and in that case
            // the blocksize would be less.
            auto workDiv = alpaka::WorkDivMembers{
                Vec::all(static_cast<alpaka::Idx<AccType>>(dotGridBlockExtent)),
                Vec::all(static_cast<alpaka::Idx<AccType>>(threadsPerBlock)),
                Vec::all(1)};

            return workDiv;
        }
        else
        {
            // Work division for single-threaded backends
            // Since block size is 1, the elements per grid is dotGridBlockExtent * blockThreadExtentMain
            alpaka::KernelCfg<AccType> const kernelCfgDot
                = {Vec::all(dotGridBlockExtent * blockThreadExtentMain), elementsPerThread};

            return alpaka::getValidWorkDiv(
                kernelCfgDot,
                devAcc,
                DotKernel(),
                bufAccInputAPtr,
                bufAccInputBPtr,
                bufAccOutputCPtr, // this is used here a kind of dummy
                static_cast<alpaka::Idx<AccType>>(arraySize));
        }
    };
```

fwyzard: Yes, template lambdas require c++20, so we cannot use them in the 1.3 branch.
The code should be equivalent though - it just does `if constexpr` directly instead of wrapping that inside a lambda and calling it.

fwyzard:
I'm mildly confused about how the CI managed to build and run the babelstream test for SYCL... I've been backporting fixes for a while and I needed the last couple to get it to even compile on my machine :-(

And now I do see the same (?) segfault, but I don't understand the reason 🤷🏻

SimeonEhrig:
> I'm mildly confused about how the CI managed to build and run the babelstream test for SYCL... I've been backporting fixes for a while and I needed the last couple to get it to even compile on my machine :-(
> 
> And now I do see the same (?) segfault, but I don't understand the reason 🤷🏻

The only explanation is, that the build parameters slightly different, such build type or if it is using mdspan. Also the used std library can be a problem. The Clang compiler cannot select different libstdc++ version, if they are installed via apt. Therefore our CI uses the libstdc++ version of the host compiler.

benchmarks/babelstream/src/babelStreamMainTest.cpp
fwyzard: Bah, sorry, I misread your earlier comment... yes, it looks like I was missing the second fix.
I've re-applied it, let's see if it helps: on my local machine I still get a crash, but not it happens _after_ all Catch2 tests have succeeded 🤷🏻

PR #2525, opened 2025-06-23
[1.3] `alpaka::math` custom implementations for CPU backend

fwyzard:
Backport of #2482 to the develop-1.3 branch.

Custom implementations are provided for 3 math functions, since the math functions like `isfinite`, `isnan`, `isinf` that rely on their std counterparts for CPU code fail when the code is built with `-Ofast`, `-ffast-math` or `-ffinite-math`.

For detailed definition see #2466.

Note: #2482 uses `std::bit_cast` that requires c++20. In this PR that is replaced by a non-`constexpr` implementation based on `memcpy`.

### DIFF
--- a/include/alpaka/core/BitCast.hpp
+++ b/include/alpaka/core/BitCast.hpp
@@ -1,0 +1,24 @@
+/* Copyright 2025 Andrea Bocci
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include <cstring>
+#include <type_traits>
+
+namespace alpaka::core
+{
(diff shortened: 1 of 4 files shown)


fwyzard:
@psychocoderHPC, the original fix in #2482 relies on `std::bit_cast`, which is in c++20.
Should I try to implement a workaround (likely using `std::memcpy` and hoping the compiler will optimise it away) or drop this fix from 1.3.0 ?

psychocoderHPC (approved):

fwyzard:
Looks like this is finally going to pass all CI tests 🎉

PR #2531, opened 2025-06-24
Implement creation of CUDA/HIP Queue with external Stream

sbaldu:
This PR implements the possibility of creating a CUDA/HIP Queue from an externally defined Stream.
To do this, the boolean member `isOwning` was added to the Queue class, which turns off the Stream destruction if the Queue was "not owning".
It also adds two overloads for the `cuda/hipGetDevice`, which were missing from the runtime APIs.

### DIFF
--- a/docs/source/dev/backends.rst
+++ b/docs/source/dev/backends.rst
@@ -240,9 +240,9 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +---------------------------------+-----------------------------------------------------------------------+
     | cudaDeviceSynchronize           | void alpaka::wait(device)                                             |
     +---------------------------------+-----------------------------------------------------------------------+
-    | cudaGetDevice                   | n/a (no current device)                                               |
+    | cudaGetDevice                   | alpaka::ApiCudaRt::getDevice()                                        |
     +---------------------------------+-----------------------------------------------------------------------+
-    | cudaGetDeviceCount              | std::size_t alpaka::getDevCount< TPlatform >()                         |
+    | cudaGetDeviceCount              | std::size_t alpaka::getDevCount< TPlatform >()                        |
     +---------------------------------+-----------------------------------------------------------------------+
     | cudaGetDeviceFlags              | --                                                                    |
     +---------------------------------+-----------------------------------------------------------------------+
--- a/docs/source/dev/backends.rst
+++ b/docs/source/dev/backends.rst
@@ -240,7 +240,7 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +---------------------------------+-----------------------------------------------------------------------+
     | cudaDeviceSynchronize           | void alpaka::wait(device)                                             |
     +---------------------------------+-----------------------------------------------------------------------+
-    | cudaGetDevice                   | alpaka::ApiCudaRt::getDevice()                                        |
+    | cudaGetDevice                   | n/a (no current device)                                               |
     +---------------------------------+-----------------------------------------------------------------------+
     | cudaGetDeviceCount              | std::size_t alpaka::getDevCount< TPlatform >()                        |
     +---------------------------------+-----------------------------------------------------------------------+
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -194,12 +194,7 @@ namespace alpaka
 #    endif
         }
 
-        static inline Error_t getDevice(int* device)
-        {
-            return ::cudaGetDevice(device);
-        }
-
-        static inline int getDevice()
+        static inline int getCurrentDevice()
         {
             int device;
             auto error = ::cudaGetDevice(&device);
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -7,8 +7,6 @@
 #include "alpaka/core/Config.hpp"
 #include "alpaka/core/UniformCudaHip.hpp"
 
-#include <stdexcept>
-
 #ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
 #    include <cuda_runtime_api.h>
 
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "alpaka/core/Config.hpp"
+#include "alpaka/core/UniformCudaHip.hpp"
 
 #include <stdexcept>
 
@@ -197,9 +198,8 @@ namespace alpaka
         static inline int getCurrentDevice()
         {
             int device;
-            auto error = ::cudaGetDevice(&device);
-            if(error != cudaSuccess)
-                throw std::runtime_error(cudaGetErrorString(error));
+            using TApi = alpaka::ApiCudaRt;
+            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(::cudaGetDevice(&device));
 
             return device;
         }
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -192,6 +192,11 @@ namespace alpaka
 #    endif
         }
 
+        static inline Error_t getDevice(int* device)
+        {
+            return ::cudaGetDevice(device);
+        }
+
         static inline Error_t getDeviceCount(int* count)
         {
             return ::cudaGetDeviceCount(count);
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -6,6 +6,8 @@
 
 #include "alpaka/core/Config.hpp"
 
+#include <stdexcept>
+
 #ifdef ALPAKA_ACC_GPU_CUDA_ENABLED
 #    include <cuda_runtime_api.h>
 
@@ -200,7 +202,9 @@ namespace alpaka
         static inline int getDevice()
         {
             int device;
-            ::cudaGetDevice(&device);
+            auto error = ::cudaGetDevice(&device);
+            if(error != cudaSuccess)
+                throw std::runtime_error(cudaGetErrorString(error));
 
             return device;
         }
--- a/include/alpaka/core/ApiCudaRt.hpp
+++ b/include/alpaka/core/ApiCudaRt.hpp
@@ -197,6 +197,14 @@ namespace alpaka
             return ::cudaGetDevice(device);
         }
 
+        static inline int getDevice()
+        {
+            int device;
+            ::cudaGetDevice(&device);
+
+            return device;
+        }
+
         static inline Error_t getDeviceCount(int* count)
         {
             return ::cudaGetDeviceCount(count);
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -5,6 +5,7 @@
 #pragma once
 
 #include "alpaka/core/Config.hpp"
+#include "alpaka/core/UniformCudaHip.hpp"
 
 #include <stdexcept>
 
@@ -222,9 +223,7 @@ namespace alpaka
         static inline int getCurrentDevice()
         {
             int device;
-            auto error = ::hipGetDevice(&device);
-            if(error != hipSuccess)
-                throw std::runtime_error(hipGetErrorString(error));
+            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(::hipGetDevice(&device));
 
             return device;
         }
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -6,6 +6,8 @@
 
 #include "alpaka/core/Config.hpp"
 
+#include <stdexcept>
+
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
 
 #    include <hip/hip_runtime_api.h>
@@ -225,7 +227,9 @@ namespace alpaka
         static inline int getDevice()
         {
             int device;
-            ::hipGetDevice(&device);
+            auto error = ::hipGetDevice(&device);
+            if(error != hipSuccess)
+                throw std::runtime_error(hipGetErrorString(error));
 
             return device;
         }
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -217,6 +217,11 @@ namespace alpaka
 #    endif
         }
 
+        static inline Error_t getDevice(int* device)
+        {
+            return ::hipGetDevice(device);
+        }
+
         static inline Error_t getDeviceCount(int* count)
         {
             return ::hipGetDeviceCount(count);
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -222,6 +222,14 @@ namespace alpaka
             return ::hipGetDevice(device);
         }
 
+        static inline int getDevice()
+        {
+            int device;
+            ::hipGetDevice(*&evice);
+
+            return device;
+        }
+
         static inline Error_t getDeviceCount(int* count)
         {
             return ::hipGetDeviceCount(count);
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -223,6 +223,7 @@ namespace alpaka
         static inline int getCurrentDevice()
         {
             int device;
+            using TApi = alpaka::ApiHipRt;
             ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(::hipGetDevice(&device));
 
             return device;
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -225,7 +225,7 @@ namespace alpaka
         static inline int getDevice()
         {
             int device;
-            ::hipGetDevice(*&evice);
+            ::hipGetDevice(&device);
 
             return device;
         }
--- a/include/alpaka/core/ApiHipRt.hpp
+++ b/include/alpaka/core/ApiHipRt.hpp
@@ -7,8 +7,6 @@
 #include "alpaka/core/Config.hpp"
 #include "alpaka/core/UniformCudaHip.hpp"
 
-#include <stdexcept>
-
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
 
 #    include <hip/hip_runtime_api.h>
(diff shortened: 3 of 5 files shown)


sbaldu:
I saw that in the table in `docs/source/dev/backends.rst` there is the table of the cuda/hip API functions and their equivalents in alpaka. Should I add `getDevice` there?

SimeonEhrig:
> I saw that in the table in `docs/source/dev/backends.rst` there is the table of the cuda/hip API functions and their equivalents in alpaka. Should I add `getDevice` there?

Yes, please.

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -71,8 +85,11 @@ namespace alpaka
                 // Make sure all pending async work is finished before destroying the stream to guarantee determinism.
                 // This would not be necessary for plain CUDA/HIP operations, but we can have host functions in the
                 // stream, which reference this queue instance and its CallbackThread. Make sure they are done.
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamSynchronize(m_UniformCudaHipQueue));
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamDestroy(m_UniformCudaHipQueue));
+                if(m_isOwning)
```
mehmetyusufoglu, line 88: could you please add the explanation above if

sbaldu: Of course

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -81,11 +98,13 @@ namespace alpaka
             }
 
         public:
+            int m_deviceId;
```
fwyzard: Unused ?

test/unit/queue/src/QueueTest.cpp
```diff
@@ -278,3 +280,21 @@ TEMPLATE_LIST_TEST_CASE("isQueue", "[queue]", alpaka::test::TestQueues)
...
+#    elif defined(ALPAKA_ACC_GPU_HIP_ENABLED)
+using Api = alpaka::ApiHipRt;
+#    endif
+
+TEMPLATE_LIST_TEST_CASE("constructQueueWithStream", "[queue]", alpaka::test::TestQueues)
+{
+    Api::Stream_t stream;
+    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(Api::streamCreate(&stream));
+    auto queue = alpaka::QueueUniformCudaHipRtBlocking<Api>(stream);
+    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(Api::streamDestroy(stream));
```
fwyzard:
```suggestion
    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(Api::streamDestroy(stream));
```
?

I think we only use `ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT` in destructors.

docs/source/dev/backends.rst
```diff
@@ -240,9 +240,9 @@ The following tables list the functions available in the `CUDA Runtime API <http
     +---------------------------------+-----------------------------------------------------------------------+
     | cudaDeviceSynchronize           | void alpaka::wait(device)                                             |
     +---------------------------------+-----------------------------------------------------------------------+
-    | cudaGetDevice                   | n/a (no current device)                                               |
+    | cudaGetDevice                   | alpaka::ApiCudaRt::getDevice()                                        |
```
fwyzard: I'm not sure we should document this as part of the public interface.

SimeonEhrig: Good point, but on the other side we documented at least one internal, see `createCudaStream`.
https://alpaka.readthedocs.io/en/stable/dev/backends.html#cuda-runtime-api

fwyzard: Mhm, I see only
image
which doesn't have any alpaka internals ?

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -192,6 +194,21 @@ namespace alpaka
 #    endif
         }
 
+        static inline Error_t getDevice(int* device)
+        {
+            return ::cudaGetDevice(device);
+        }
```
fwyzard: This variant does not seem to be used, I would suggest to drop it.

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -192,6 +194,21 @@ namespace alpaka
...
+
+        static inline int getDevice()
+        {
+            int device;
+            auto error = ::cudaGetDevice(&device);
+            if(error != cudaSuccess)
+                throw std::runtime_error(cudaGetErrorString(error));
+
+            return device;
+        }
```
fwyzard: And rename this as `getCurrentDevice()`.

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -71,8 +85,11 @@ namespace alpaka
                 // Make sure all pending async work is finished before destroying the stream to guarantee determinism.
                 // This would not be necessary for plain CUDA/HIP operations, but we can have host functions in the
                 // stream, which reference this queue instance and its CallbackThread. Make sure they are done.
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamSynchronize(m_UniformCudaHipQueue));
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamDestroy(m_UniformCudaHipQueue));
+                if(m_isOwning)
+                {
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamSynchronize(m_UniformCudaHipQueue));
+                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamDestroy(m_UniformCudaHipQueue));
+                }
```
fwyzard, line 92: I agree that, if the queue is not owned, it should not be destroyed in the destructor.
However, maybe it should still be synchronised ?
@sbaldu I guess it mostly depends on how you plan to use the non-owning queues.

```suggestion
                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamSynchronize(m_UniformCudaHipQueue));
                if(m_isOwning)
                {
                    ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamDestroy(m_UniformCudaHipQueue));
                }
```

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -220,10 +245,11 @@ namespace alpaka
...
             {
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(TApi::launchHostFunc(
-                    queue.getNativeHandle(),
-                    uniformCudaHipRtHostFunc,
-                    new HostFuncData{*queue.m_spQueueImpl, task}));
+                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(
+                    TApi::launchHostFunc(
+                        queue.getNativeHandle(),
+                        uniformCudaHipRtHostFunc,
+                        new HostFuncData{*queue.m_spQueueImpl, task}));
```
fwyzard: This block seems to be unchanged ?

fwyzard:
Other than the comments I left on the code, the main concern I have is that the new interface is only implemented for CUDA and HIP queues.

Any ideas if and how it could be implemented for CPU and SYCL ?

docs/source/dev/backends.rst
SimeonEhrig: I would say `alpaka::QueueCudaRtBlocking` is an internal. 
But the documentation is outdated in general and has no clear guideline. Therefore I'm also fine to remove the `cudaGetDevice` entry.

fwyzard: > I would say `alpaka::QueueCudaRtBlocking` is an internal.

Ah... that's actually what we use explicitly in the CMS code :-D
Well, `QueueCudaRtNonBlocking`, but still.

What is the more "public" interface you suggest we should use ?

SimeonEhrig: Okay, I was not aware what is `QueueCudaRtNonBlocking` rely. I thought this is an alias for `alpaka::Queue<alpaka::AccGpuCudaRt, alpaka::NonBlocking>`. So let's remove `alpaka::ApiCudaRt::getDevice()`.

Original I thought documentation is good in general but I didn't thought about that it could be internal API, which should be not in the public documentation.

sbaldu: Ok I will remove it

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -192,6 +194,16 @@ namespace alpaka
 #    endif
         }
 
+        static inline int getCurrentDevice()
+        {
+            int device;
+            auto error = ::cudaGetDevice(&device);
+            if(error != cudaSuccess)
+                throw std::runtime_error(cudaGetErrorString(error));
```
fwyzard: Instead of explicitly throwing an exception, would it be feasible to reuse the functions and macros from `include/alpaka/core/UniformCudaHip.hpp` ?

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -1,6 +1,5 @@
-/* Copyright 2022 Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci, Bernhard Manfred Gruber,
- * Antonio Di Pilato
- * SPDX-License-Identifier: MPL-2.0
+/* Copyright 2022 Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci,
```
fwyzard:
```suggestion
/* Copyright 2025 Benjamin Worpitz, Matthias Werner, René Widera, Andrea Bocci,
```
!

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -25,221 +24,229 @@
...
-        {
-            [[nodiscard]] static auto getNativeHandle(
-                uniform_cuda_hip::detail::QueueUniformCudaHipRt<TApi, TBlocking> const& queue)
-            {
-                return queue.getNativeHandle();
-            }
-        };
-    } // namespace trait
+namespace alpaka {
+template <typename TApi> class EventUniformCudaHipRt;
```
fwyzard: The indentation looks weird ?

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
sbaldu: I guess it could make sense in both ways.
In my mind the idea is that I have a CUDA/HIP framework, with one or more streams, and I want to launch some work from a library implemented in alpaka, so I pass the stream, from which the Queue is constructed. In that case I think it makes sense to expect the developer to synchronize the stream themselves.
Alternatively, we could say that it would be the third library's duty to synchronize before finishing its job.

fwyzard: OK, so your use case is something like
```c++
void submit(cudaStream_t stream, ...) {
  alpaka::QueueCudaRtNonBlocking queue(stream);
  // call some alpaka-based library using queue
}
```
?

Then I agree it makes sense for the `queue` _not_ to synchronise, otherwise `submit()` would become a blocking function.

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -222,9 +223,7 @@ namespace alpaka
         static inline int getCurrentDevice()
         {
             int device;
-            auto error = ::hipGetDevice(&device);
-            if(error != hipSuccess)
-                throw std::runtime_error(hipGetErrorString(error));
+            ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK(::hipGetDevice(&device));
```
fwyzard, line 225: I guess this too needs an `using TApi = ...` ?

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -161,8 +186,8 @@ namespace alpaka
 
         //! The CUDA/HIP RT queue thread wait trait specialization.
         //!
-        //! Blocks execution of the calling thread until the queue has finished processing all previously requested
-        //! tasks (kernels, data copies, ...)
+        //! Blocks execution of the calling thread until the queue has finished
+        //! processing all previously requested tasks (kernels, data copies, ...)
```
fwyzard: Please revert this change.

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -198,11 +223,13 @@ namespace alpaka
 
             struct HostFuncData
             {
-                // We don't need to keep the queue alive, because in it's dtor it will synchronize with the CUDA/HIP
-                // stream and wait until all host functions and the CallbackThread are done. It's actually an error to
-                // copy the queue into the host function. Destroying it here would call CUDA/HIP APIs from the host
-                // function. Passing it further to the Callback thread, would make the Callback thread hold a task
-                // containing the queue with the CallbackThread itself. Destroying the task if no other queue instance
+                // We don't need to keep the queue alive, because in it's dtor it will
```
fwyzard: please revert this change

include/alpaka/core/ApiHipRt.hpp
sbaldu: Yes thank you

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
sbaldu: Done

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
sbaldu: Done

include/alpaka/core/ApiCudaRt.hpp
```diff
@@ -5,6 +5,9 @@
 #pragma once
 
 #include "alpaka/core/Config.hpp"
+#include "alpaka/core/UniformCudaHip.hpp"
+
+#include <stdexcept>
```
fwyzard: this can be removed ?

include/alpaka/core/ApiHipRt.hpp
```diff
@@ -5,6 +5,9 @@
 #pragma once
 
 #include "alpaka/core/Config.hpp"
+#include "alpaka/core/UniformCudaHip.hpp"
+
+#include <stdexcept>
```
fwyzard: same ?

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -43,36 +47,52 @@ namespace alpaka
...
+                // - [cuda/hip]StreamNonBlocking: Specifies that work running in the created
+                // queue may run concurrently with work in queue 0 (the NULL queue),
+                //   and that the created queue should perform no implicit synchronization
+                //   with queue 0.
                 // Create the queue on the current device.
-                // NOTE: [cuda/hip]StreamNonBlocking is required to match the semantic implemented in the alpaka
-                // CPU queue. It would be too much work to implement implicit default queue synchronization on CPU.
+                // NOTE: [cuda/hip]StreamNonBlocking is required to match the semantic
+                // implemented in the alpaka CPU queue. It would be too much work to
+                // implement implicit default queue synchronization on CPU.
```
fwyzard: can you revert this changes to the comments as well ?

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
```diff
@@ -43,36 +47,52 @@ namespace alpaka
...
 
-                // Make sure all pending async work is finished before destroying the stream to guarantee determinism.
-                // This would not be necessary for plain CUDA/HIP operations, but we can have host functions in the
-                // stream, which reference this queue instance and its CallbackThread. Make sure they are done.
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamSynchronize(m_UniformCudaHipQueue));
-                ALPAKA_UNIFORM_CUDA_HIP_RT_CHECK_NOEXCEPT(TApi::streamDestroy(m_UniformCudaHipQueue));
+                // Make sure all pending async work is finished before destroying the stream
+                // to guarantee determinism. This would not be necessary for plain CUDA/HIP
+                // operations, but we can have host functions in the stream, which reference
+                // this queue instance and its CallbackThread. Make sure they are done.
```
fwyzard: And these.

include/alpaka/core/ApiCudaRt.hpp
sbaldu: I can remove `stdexcept`, but `UniformCudaHip.hpp` I think I need for the runtime check macro

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
sbaldu: Sure

include/alpaka/core/ApiCudaRt.hpp
fwyzard: yes, I only meant `stdexcept`

sbaldu: Done, thanks

include/alpaka/core/ApiHipRt.hpp
sbaldu: Done

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
sbaldu: Done

include/alpaka/queue/cuda-hip/QueueUniformCudaHipRt.hpp
sbaldu: Done

fwyzard (approved):

fwyzard (approved):

SimeonEhrig (approved):

sbaldu:
@psychocoderHPC It looks like the CI is passing now, so it should be ok.

psychocoderHPC (approved):

PR #2537, opened 2025-06-29
Use `__builtin_assume_aligned()` to promise the correct alignment

fwyzard:
Use `__builtin_assume_aligned()` where needed to guarantee the correct alignment to the compiler, instead of disabling the `-Wcast-align` warning.

Note: the same effect could be achieved with an intermediate cast to `uintptr_t`. In principle `__builtin_assume_aligned()` should be preferred because the compiler might actually make use of the information.

### DIFF
--- a/test/unit/mem/view/src/ViewSubViewTest.cpp
+++ b/test/unit/mem/view/src/ViewSubViewTest.cpp
@@ -16,12 +16,6 @@
 #include <numeric>
 #include <type_traits>
 
-#if ALPAKA_COMP_GNUC
-#    pragma GCC diagnostic push
-#    pragma GCC diagnostic ignored                                                                                    \
-        "-Wcast-align" // "cast from 'std::uint8_t*' to 'Elem*' increases required alignment of target type"
-#endif
-
 namespace alpaka::test
 {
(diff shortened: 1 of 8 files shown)


test/unit/mem/view/src/ViewSubViewTest.cpp
```diff
@@ -48,7 +42,7 @@ namespace alpaka::test
             auto viewPtrNative = reinterpret_cast<std::uint8_t*>(alpaka::getPtrNative(buf));
             if constexpr(TDim::value > 0)
                 viewPtrNative += (offsetView * pitchBuf).sum();
-            REQUIRE(reinterpret_cast<TElem*>(viewPtrNative) == alpaka::getPtrNative(view));
+            REQUIRE(viewPtrNative == reinterpret_cast<std::uint8_t const*>(alpaka::getPtrNative(view)));
```
mehmetyusufoglu, line 45: is there a possibility that _getPtrNative_ returns const?

fwyzard: yes, for a `ViewConst` or any kind of `View<const T>`

psychocoderHPC (approved):

PR #2544, opened 2025-07-03
Implement `alpaka::transform`

### LINKED ISSUE #2542: Implement `alpaka::transform` similar to `std::tramsform`
@felicepantaleo and @sbaldu have suggested that alpaka could implement an `alpaka::transform()` metafunction to apply a transformation to all elements in a range or buffer.

After a quick chat, the interface could be along the lines of
```c++
alpaka::transform<Tag>(
    queue,
    first_in,
    last_in,
    first_out,
    [] ALPAKA_FN_ACC(float x) { return do_something_with(x); });
```
and
```c++
alpaka::transform<Tag>(
    queue,
    input_buf,
    output_buf,
    [] ALPAKA_FN_ACC(float x) { return do_something_with(x); });
```

@psychocoderHPC what is your opinion ?
@felicepantaleo @sbaldu do you want to expand on the reason for this ?

I have a prototype implementation, so I can open a PR if we agree to move forward with it.

fwyzard:
- [x] rebase against #2566 before merging

Implement two version of `alpaka::transform`.

The iterator range version

    alpaka::transform<Tag>(queue, first_in, last_in, first_out, op)

applies asynchronously the given function `op` to the elements of the semi-open input range `[first_in, last_in)`, and stores the result in an output range starting from `first_out`, using the accelerator back-end identified by `Tag`.

The buffer version

    alpaka::transform<Tag>(queue, out, op, in)

applies asynchronously the given function `op` to the elements of the input buffer `in` and stores the result in the corresponding elements of the output buffer `out`, using the accelerator back-end identified by `Tag`.

The function `op` should accept either a single argument (of the iterator's or buffer's element type) or two arguments (an accelerator and an element), and return a value of the element type.

Note that the range version is always one-dimensional, while the buffer version uses the dimensionality of the buffer.

### DIFF
--- a/include/alpaka/algo/Transform.hpp
+++ b/include/alpaka/algo/Transform.hpp
@@ -1,0 +1,193 @@
+/* Copyright 2025 Andrea Bocci, Simeon Ehrig
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/acc/Tag.hpp"
+#include "alpaka/dev/Traits.hpp"
+#include "alpaka/dim/DimIntegralConst.hpp"
+#include "alpaka/dim/Traits.hpp"
+#include "alpaka/elem/Traits.hpp"
+#include "alpaka/exec/UniformElements.hpp"
+#include "alpaka/extent/Traits.hpp"
+#include "alpaka/idx/Traits.hpp"
+#include "alpaka/kernel/Traits.hpp"
+#include "alpaka/mem/view/Traits.hpp"
+#include "alpaka/vec/Vec.hpp"
+#include "alpaka/workdiv/WorkDivHelpers.hpp"
+
+#include <iterator>
+#include <type_traits>
+
+namespace alpaka
+{
+
+    namespace detail
+    {
+
+        template<typename TFn>
+        struct TransformKernel
+        {
+            TFn fn;
+
+            template<typename TAcc, typename T>
+            ALPAKA_FN_ACC void operator()(TAcc const& acc, T const* in_ptr, T* out_ptr, alpaka::Idx<TAcc> size) const
+            {
+                static_assert(std::is_invocable_r_v<T, TFn, T> or std::is_invocable_r_v<T, TFn, TAcc const&, T>);
+
+                static_assert(alpaka::Dim<TAcc>::value == 1u);
+                using Idx = alpaka::Idx<TAcc>;
+
+                for(Idx i : alpaka::uniformElements(acc, size))
+                {
+                    if constexpr(std::is_invocable_r_v<T, TFn, T>)
+                    {
+                        // std::is_invocable_r_v<T, TFn, T>
+                        out_ptr[i] = fn(in_ptr[i]);
+                    }
+                    else
+                    {
+                        // std::is_invocable_r_v<T, TFn, TAcc const&, T>
+                        out_ptr[i] = fn(acc, in_ptr[i]);
+                    }
+                }
+            }
+        };
+
+        template<typename TFn>
+        struct TransformKernelND
+        {
+            TFn fn;
+
+            template<typename TAcc, typename T>
+            ALPAKA_FN_ACC void operator()(
+                TAcc const& acc,
+                T const* in_ptr,
+                alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> in_pithces,
+                T* out_ptr,
+                alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> out_pitches,
+                alpaka::Vec<alpaka::Dim<TAcc>, alpaka::Idx<TAcc>> in_size) const
+            {
+                static_assert(std::is_invocable_r_v<T, TFn, T> or std::is_invocable_r_v<T, TFn, TAcc const&, T>);
+
+                using Dim = alpaka::Dim<TAcc>;
+                using Idx = alpaka::Idx<TAcc>;
+                using Vec = alpaka::Vec<Dim, Idx>;
+
+                for(Vec idx : alpaka::uniformElementsND(acc, in_size))
+                {
+                    auto p_in = reinterpret_cast<T const*>(
+                        reinterpret_cast<uintptr_t>(in_ptr) + static_cast<uintptr_t>((idx * in_pithces).sum()));
+                    auto p_out = reinterpret_cast<T*>(
+                        reinterpret_cast<uintptr_t>(out_ptr) + static_cast<uintptr_t>((idx * out_pitches).sum()));
+                    if constexpr(std::is_invocable_r_v<T, TFn, T>)
+                    {
+                        // std::is_invocable_r_v<T, TFn, T>
+                        *p_out = fn(*p_in);
+                    }
+                    else
+                    {
+                        // std::is_invocable_r_v<T, TFn, TAcc const&, T>
+                        *p_out = fn(acc, *p_in);
+                    }
+                }
+            }
+        };
+
+    } // namespace detail
+
+    /*
+     * Applies asynchronously the given function `fn` to the elements of the input range starting at `in`,
+     * and stores the result in the semi-open output range [`out_begin`,`out_end`), using the accelerator
+     * back-end identified by `Tag`.
+     */
+    template<alpaka::concepts::Tag TTag, typename TQueue, typename T, typename TFn>
+    void transform(TQueue& queue, T* out_begin, T* out_end, TFn&& fn, T* in)
+    {
+        using Idx = typename std::iterator_traits<T*>::difference_type;
+        using Acc1D = alpaka::TagToAcc<TTag, alpaka::DimInt<1>, Idx>;
+
+        static_assert(
+            std::is_invocable_r_v<T, TFn, T> or std::is_invocable_r_v<T, TFn, Acc1D const&, T>,
+            "TFn must accept either one argument (of type T) or two arguments (an accelerator and an argument of type "
+            "T), and return a value of type T.");
+
+        Idx size = std::distance(out_begin, out_end);
+        detail::TransformKernel<TFn> kernel{fn};
+
+        // Find a valid work division. This could be further optimised.
+        auto const config
+            = alpaka::KernelCfg<Acc1D>{size, Idx{1}, false, alpaka::GridBlockExtentSubDivRestrictions::Unrestricted};
+        auto const grid = alpaka::getValidWorkDiv(config, alpaka::getDev(queue), kernel, in, out_begin, size);
+
+        // Apply the fn function to all elements of the input range.
+        alpaka::exec<Acc1D>(queue, grid, kernel, in, out_begin, size);
+    }
+
+    /*
+     * Applies asynchronously the given function `fn` to the elements of the input buffer `in`,
+     * and stores the result in the corresponding elements of the output buffer `out`,
+     * using the accelerator back-end identified by `Tag`.
+     */
+    template<alpaka::concepts::Tag TTag, typename TQueue, typename TBuf, typename TFn, typename TConstBuf>
+    void transform(TQueue& queue, TBuf& out, TFn&& fn, TConstBuf const& in)
+    {
+        // Check that the input and output buffers have compatible types.
+        using Idx = alpaka::Idx<TConstBuf>;
+        static_assert(
+            std::is_same_v<alpaka::Idx<TBuf>, Idx>,
+            "The input and output buffers must have the same index type.");
+        using Dim = alpaka::Dim<TConstBuf>;
+        static_assert(
+            std::is_same_v<alpaka::Dim<TBuf>, Dim>,
+            "The input and output buffers must have the same dimension.");
+        using In = std::remove_const_t<alpaka::Elem<TConstBuf>>;
+        using Out = alpaka::Elem<TBuf>;
+        using Vec = alpaka::Vec<Dim, Idx>;
+        using Acc = alpaka::TagToAcc<TTag, Dim, Idx>;
+
+        static_assert(
+            std::is_invocable_r_v<Out, TFn, In const> or std::is_invocable_r_v<Out, TFn, Acc const&, In const>,
+            "TFn must accept either one argument (of the buffer's element type) or two arguments (an accelerator and "
+            "the element type), and return a value of the buffer's element type.");
+
+        // Check that the input and output buffers have the same size.
+        Vec size = alpaka::getExtents(in);
+        assert(alpaka::getExtents(out) == size and "The input and output buffers must have the same extents.");
+
+        // Pass details of the input and output buffers to the kernel:
+        //   - address of the first elements
+        //   - pitches (in bytes) along all dimensions
+        //   - number of elements along all dimensions
+        detail::TransformKernelND<TFn> kernel{fn};
+
+        // Find a valid work division. This could be further optimised.
+        auto const config = alpaka::KernelCfg<Acc>{
+            size,
+            Vec::ones(),
+            false,
+            alpaka::GridBlockExtentSubDivRestrictions::Unrestricted};
+        auto const grid = alpaka::getValidWorkDiv(
+            config,
+            alpaka::getDev(queue),
+            kernel,
+            in.data(),
+            alpaka::getPitchesInBytes(in),
+            out.data(),
+            alpaka::getPitchesInBytes(out),
+            size);
+
+        // Apply the fn function to all elements of the input buffer.
+        alpaka::exec<Acc>(
+            queue,
+            grid,
+            kernel,
+            in.data(),
+            alpaka::getPitchesInBytes(in),
+            out.data(),
+            alpaka::getPitchesInBytes(out),
+            size);
+    }
+
+} // namespace alpaka
--- a/test/unit/algo/src/Transform.cpp
+++ b/test/unit/algo/src/Transform.cpp
@@ -1,0 +1,157 @@
+/* Copyright 2025 Andrea Bocci
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#include <alpaka/alpaka.hpp>
+#include <alpaka/test/Extent.hpp>
+#include <alpaka/test/acc/TestAccs.hpp>
+#include <alpaka/test/queue/Queue.hpp>
+
+#include <catch2/catch_template_test_macros.hpp>
+#include <catch2/catch_test_macros.hpp>
+#include <catch2/matchers/catch_matchers_floating_point.hpp>
+
+#include <cmath>
+#include <random>
+
+// A functor that takes a single argument of type float and returns a value of type float
+struct Duplicate
+{
+    template<typename T>
+    ALPAKA_FN_HOST_ACC T operator()(T value) const
+    {
+        return value * 2;
+    }
+};
+
+// A functor that takes two arguments (an accelerator object and a float) and returns a value of type float
+struct Sin
+{
+    template<typename TAcc, typename T>
+    ALPAKA_FN_HOST_ACC T operator()(TAcc const& acc, T value) const
+    {
+        return alpaka::math::sin(acc, value);
+    }
+};
+
+using TestAccs1D = alpaka::test::EnabledAccs<alpaka::DimInt<1u>, uint32_t>;
+
+TEMPLATE_LIST_TEST_CASE("transform_pointers", "[algo]", TestAccs1D)
+{
+    using Acc = TestType;
+    using Tag = alpaka::AccToTag<Acc>;
+    using Platform = alpaka::Platform<Acc>;
+    using Device = alpaka::Dev<Acc>;
+    using Queue = alpaka::Queue<Device, alpaka::NonBlocking>;
+
+    using Idx = alpaka::Idx<Acc>;
+    // using Dim = alpaka::Dim<Acc>;
+    // using Vec = alpaka::Vec<Dim, Idx>;
+
+    const alpaka::PlatformCpu host_platform{};
+    auto const host = alpaka::getDevByIdx(host_platform, 0u);
+
+    const Platform platform{};
+    const Device device = alpaka::getDevByIdx(platform, 0);
+    Queue queue{device};
+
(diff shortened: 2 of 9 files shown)


SimeonEhrig:
I will have a look on this PR later. I have a lot of ideas about the `alpaka::transform` because [vikunja](https://github.com/alpaka-group/vikunja) already implemented it.

fwyzard:
@SimeonEhrig I got this error from the CI
```
/builds/hzdr/crp/alpaka/include/alpaka/algo/Transform.hpp(34): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag

/builds/hzdr/crp/alpaka/include/alpaka/algo/Transform.hpp(83): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag
```

Should I work around it ?
Or do we require `--extended-lambda` in alpaka ?

SimeonEhrig:
> @SimeonEhrig I got this error from the CI
> 
> ```
> /builds/hzdr/crp/alpaka/include/alpaka/algo/Transform.hpp(34): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag
> 
> /builds/hzdr/crp/alpaka/include/alpaka/algo/Transform.hpp(83): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag
> ```
> 
> Should I work around it ? Or do we require `--extended-lambda` in alpaka ?

The CI disable it by default. For mdspan it is required, therefore we enable it. I think it is disabled by default, because we never required it. I think the main idea was to test without an experimental feature, if possible. At the time of the decision, it was experimental.

https://github.com/alpaka-group/alpaka/blob/967f9ace50765f02276fbf7586129646968502f6/script/job_generator/generate_job_yaml.py#L344-L360

I know during development of vikunja, there was a limitation of lambdas, which disable the usage. But maybe this was fixed in a later version (I tested it at least with CUDA 11.x). I think we can enable it by default. So set the default to `ON` and remove the mdspan if statement.

By the, here is documentation about the CUDA extended lambda behavior: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#extended-lambdas

fwyzard:
OK, it's a bit more verbose, but I can rework the implementation and test of `transform` without the extended lambdas.

fwyzard:
OK, I've implemented the optional `acc` argument to the functor.

fwyzard:
OK, the implementation should be complete.

I'm not saying that we should just go ahead and merge it - if these algorithms are already implemented in [vikunja](https://github.com/alpaka-group/vikunja), I'm fine to just use them there!

psychocoderHPC:
> alpaka::transform<Tag>(queue, in, out, op)



> @SimeonEhrig I got this error from the CI
> 
> ```
> /builds/hzdr/crp/alpaka/include/alpaka/algo/Transform.hpp(34): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag
> 
> /builds/hzdr/crp/alpaka/include/alpaka/algo/Transform.hpp(83): error: __host__ or __device__ annotation on lambda requires --extended-lambda nvcc flag
> ```
> 
> Should I work around it ? Or do we require `--extended-lambda` in alpaka ?

The problem are generic lambdas. If you use a CPU and GPU accelerator than this kind of errors happen. If you use CUDA only than you can use `[] __device__(){}` but you must know all types and you are not allowed to use `auto` for arguments.
For alpaka3 I do not support therefore lambdas defined on the host side. It is currently not possible to write heterogeneous code.
If you specialize all lambda arguments and not use an alpaka ACC as first argument e.g. `[] ALPAKA_FN_HOST_ACC(int,int){}`  than host lambdas are no problem.

fwyzard:
Yes, I noticed. 

The way we use alpaka with a dedicated namespace for each "back end", we can fully specify the accelerator type, so lambdas can still be useful.

include/alpaka/algo/Transform.hpp
```diff
@@ -0,0 +1,192 @@
...
+
+    } // namespace detail
+
+    /*
+     * Applies asynchronously the given function `fn` to the elements of the semi-open input range
+     * `[first_in, last_in)`, and stores the result in an output range starting from `first_out`,
+     * using the accelerator back-end identified by `Tag`.
+     */
+    template<typename TTag, typename TQueue, typename T, typename TFn>
+    void transform(TQueue& queue, T const* first, T const* last, T* out, TFn&& fn)
```
SimeonEhrig: I would avoid an API with raw memory pointers, until there is a good reason for it. There are some disadvantages of raw pointers. Mainly a raw pointer cannot store additional information like is the memory 1D or MD, is the memory continuous, there are pitches and so one. Therefore the user needs care about, that the raw pointer fulfill all requirements, which you defined in the doc string.

Instead we should support only one API, which also reduce the maintenance work, which takes only `alpaka::Buffer` and `alpaka::Views` and we use temporary `alpaka::view` to wrap raw pointers and container types e.g.:

```c++
alpaka::transform(
  queue,
  alpaka::view{p_in_begin, p_in_end},
  alpaka::view{p_out_begin, p_out_end},
  [](){}
);
```
In the case of if our original memory are raw pointers, this looks a little bit verbose. But if we use an container type like `std::array` and `std::vector`, it looks more user friendly. And this approach allows us to add more information to the memory without the need to change all the APIs. For example, at some point the related device of the memory is interesting.

```c++
alpaka::CPU = cpu{};
std::array a_in = {0,1,2};
std::vector b_out(3,0);

alpaka::transform(
  queue,
  alpaka::view{a_in, cpu}
  alpaka::view{b_out, cpu},
  [](){}
);
```

include/alpaka/algo/Transform.hpp
```diff
@@ -0,0 +1,192 @@
...
+        };
+
+    } // namespace detail
+
+    /*
+     * Applies asynchronously the given function `fn` to the elements of the semi-open input range
+     * `[first_in, last_in)`, and stores the result in an output range starting from `first_out`,
+     * using the accelerator back-end identified by `Tag`.
+     */
+    template<typename TTag, typename TQueue, typename T, typename TFn>
```
SimeonEhrig:
```suggestion
    template<alpaka::concepts::Tag TTag, typename TQueue, typename T, typename TFn>
```

include/alpaka/algo/Transform.hpp
```diff
@@ -0,0 +1,192 @@
...
+        // Apply the fn function to all elements of the input range.
+        alpaka::exec<Acc1D>(queue, grid, kernel, first, out, size);
+    }
+
+    /*
+     * Applies asynchronously the given function `fn` to the elements of the input buffer `in`,
+     * and stores the result in the corresponding elements of the output buffer `out`,
+     * using the accelerator back-end identified by `Tag`.
+     */
+    template<typename TTag, typename TQueue, typename TConstBuf, typename TBuf, typename TFn>
```
SimeonEhrig:
```suggestion
    template<alpaka::concepts::Tag TTag, typename TQueue, typename TConstBuf, typename TBuf, typename TFn>
```

A concept for `alpaka::Buf` and `alpaka::ConstBuf` would be also nice, but does not exist. I think this is something for an upcoming PR.

include/alpaka/algo/Transform.hpp
```diff
@@ -0,0 +1,192 @@
...
+        static_assert(
+            std::is_same_v<std::remove_const_t<alpaka::Elem<TConstBuf>>, Elem>,
+            "The input and output buffers must have the same element type.");
+        using Vec = alpaka::Vec<Dim, Idx>;
+        using Acc = alpaka::TagToAcc<TTag, Dim, Idx>;
+
+        static_assert(
+            std::is_invocable_r_v<Elem, TFn, Elem> or std::is_invocable_r_v<Elem, TFn, Acc const&, Elem>,
+            "TFn must accept either one argument (of the buffer's element type) or two arguments (an accelerator and "
+            "the element type), and return a value of the buffer's element type.");
```
SimeonEhrig: Let me try to move this to a concept :-)

include/alpaka/algo/Transform.hpp
```diff
@@ -0,0 +1,192 @@
...
+            std::is_same_v<alpaka::Idx<TBuf>, Idx>,
+            "The input and output buffers must have the same index type.");
+        using Dim = alpaka::Dim<TConstBuf>;
+        static_assert(
+            std::is_same_v<alpaka::Dim<TBuf>, Dim>,
+            "The input and output buffers must have the same dimension.");
+        using Elem = alpaka::Elem<TBuf>;
+        static_assert(
+            std::is_same_v<std::remove_const_t<alpaka::Elem<TConstBuf>>, Elem>,
+            "The input and output buffers must have the same element type.");
```
SimeonEhrig: Is there a reason for it? `std::transform` does not require it and I see no reason why we should require it.

include/alpaka/algo/Transform.hpp
fwyzard: I'm fine with this idea.

In fact, if one uses a container like `std::array` or `std::vector`, they should already be usable as alpaka Views, right ?

I would maybe make something like `alpaka::cpu` or `alpaka::host()` available centrally, so it's easier for people to use - but that's a different topic.

fwyzard: @sbaldu @felicepantaleo since you were requesting the `alpaka::transform` feature, what do you think about this proposal to restrict it to buffers and views ?

include/alpaka/algo/Transform.hpp
fwyzard: Nope, just a wrong assumption on my part. I'll remove it.

include/alpaka/algo/Transform.hpp
SimeonEhrig: > In fact, if one uses a container like `std::array` or `std::vector`, they should already be usable as alpaka Views, right ?

Yes, `std::array` and `std::vector` are already supported. And if this [PR](https://github.com/alpaka-group/alpaka/pull/2490) is merged, more types are possible.

> I would maybe make something like `alpaka::cpu` or `alpaka::host()` available centrally, so it's easier for people to use - but that's a different topic.

Agree.

include/alpaka/algo/Transform.hpp
fwyzard: Done, thanks.

include/alpaka/algo/Transform.hpp
fwyzard: Removed, thanks.

include/alpaka/algo/Transform.hpp
SimeonEhrig: That's strange. There is only the concept `std::invocable` which equivalent to `std::is_invocable_v` but there is no concept equivalent to `std::is_invocable_r_v`. Nevertheless, you should not check this on the host side. The Acc type on host side is a different type compare to the device side. @psychocoderHPC told me, that it is not possible to construct kernel acc type on host side.

fwyzard: I'm not sure I follow.

I know that we cannot construct an _object_ of type `Acc` on the host, but shouldn't the _type_ be fully defined anyway ?

psychocoderHPC:
I like that you added a n-dimensional transform to alpaka this can simplify basic things a lot.
For documentation reason, I would like to point out that I decided after internal discussions with other developers for alpaka3 not to follow the STL interface.

The interface in alpaka3 the following to support easy writing n-ary functors without the need for a zip iterator/container.

```
void transform(queue, executor,  auto&& out,  auto&& fn, auto&&... in);
```

full interface: https://github.com/alpaka-group/alpaka3/blob/e92dab24ac85d487f521f55f3f5b5ae47d8ec6cd/include/alpaka/onHost/algo/transform.hpp#L49-L69
test examples: https://github.com/alpaka-group/alpaka3/blob/e92dab24ac85d487f521f55f3f5b5ae47d8ec6cd/tests/unit/algorithm/transform.cpp#L99

include/alpaka/algo/Transform.hpp
SimeonEhrig: In the past @psychocoderHPC  told me it. I want to implement a fancy operator, which allows to transform an value before accessing it. Therefore I need the Acc type on host, to define the class.

I didn't checked the code yet, if we get same type on host and device or an slightly different but a wrote a small kernel to check this.

```c++

template<typename THostAcc>
struct CheckAccObject
{
    template<typename TAcc>
    ALPAKA_FN_ACC auto operator()(TAcc const& acc) const -> void
    {
        static_assert(std::is_same_v<TAcc, THostAcc>);
    }
};

int main(){
  // ...
    CheckAccObject<Acc> checkKernel;

    alpaka::exec<Acc>(
        queue,
        workDiv,
        checkKernel
        /* put kernel arguments here */);
    alpaka::wait(queue);
}
```

The kernel is fine for all backends. Maybe @psychocoderHPC has a comment on it.

SimeonEhrig:
> I like that you added a n-dimensional transform to alpaka this can simplify basic things a lot. For documentation reason, I would like to point out that I decided after internal discussions with other developers for alpaka3 not to follow the STL interface.
> 
> The interface in alpaka3 the following to support easy writing n-ary functors without the need for a zip iterator/container.
> 
> ```
> void transform(queue, executor,  auto&& out,  auto&& fn, auto&&... in);
> ```
> 
> full interface: https://github.com/alpaka-group/alpaka3/blob/e92dab24ac85d487f521f55f3f5b5ae47d8ec6cd/include/alpaka/onHost/algo/transform.hpp#L49-L69 test examples: https://github.com/alpaka-group/alpaka3/blob/e92dab24ac85d487f521f55f3f5b5ae47d8ec6cd/tests/unit/algorithm/transform.cpp#L99

Is there a reason, why the functor is between output and input memory? Personally I would group the memory arguments.

```c++
void transform(queue, executor, auto&& fn, auto&& out, auto&&... in);
```

psychocoderHPC:
sry wrote it without copiing the signature is

```C++
void transform(queue, executor, auto&& out, auto&& fn,  auto&&... in);
```

SimeonEhrig:
> sry wrote it without copiing the signature is
> 
> ```c++
> void transform(queue, executor, auto&& out, auto&& fn,  auto&&... in);
> ```

You wrote the signature and I suggested to reorder the arguments to cluster the output and input memory.

psychocoderHPC:
@SimeonEhrig  Sry misread it.

The reason for the order is that it follows the reading `out = fn(in...)` what follows the interface for kernel execution, where we have `fn, args...`. This allows the user to see which parameters are passed into the functor. `out` is a parameter the user will not touch in the functor because it is never forwarded to the user.

`transformReduce(queue, executor,neutralElement,outPtr,reduceFn,transformFn,in...)` follows this order too. [see](https://github.com/alpaka-group/alpaka3/blob/50e4c2ff35e497722061f62f95b4d68b85472598/include/alpaka/onHost/algo/transformReduce.hpp#L57-L64)

fwyzard:
During the VC we agreed to go with `void transform(queue, executor, auto&& out, auto&& fn,  auto&&... in);`

fwyzard:
I've updated the buffer interface to match what we discussed:
```c++
alpaka::transform<Tag>(queue, out, op, in);
```
so that it can later be extended to multiple inputs:
```c++
alpaka::transform<Tag>(queue, out, op, in1, in2 ...);
```

@sbaldu how would you like the STL-interface to look like ?

fwyzard:
After discussing offline with @sbaldu we picked
```c++
alpaka::transform<Tag>(queue, out_begin, out_end, op, in);
```

that could later be extended to
```c++
alpaka::transform<Tag>(queue, out_begin, out_end, op, in1, in2, ...);
```

fwyzard:
So, GCC and NVCC consider the `transform` kernel to be trivially copyable, while clang does not.

@psychocoderHPC do you have any suggestions to fix this ?

fwyzard:
OK, with the latest changes it should work also on clang.

fwyzard:
To be rebased after https://github.com/alpaka-group/alpaka/pull/2565 is merged.

fwyzard:
@SimeonEhrig the log says only

> The following tests FAILED:
>	 27 - algoTest (Failed)

Is there a way to re-run it and get the full output ?

fwyzard:
Turns out that _was_ the whole output :-)

SimeonEhrig:
@sbaldu  A little bit off-topic, but relevant for my current on alpaka3. Would the following solution also work for you, instead direct passing the pointers to `alpaka::transform`?

```c++
int * out_begin = //...;
int * out_end = //...;

alpaka::transform<Tag>(queue, alpaka::view{out_begin, out_end}, op, in);
```

fwyzard:
What is `alpaka::view` here ?

Is it the same as the "view" that provides access to memory (on some device) with a buffer-like interface ?

Or something like an `std::pair` that provides `begin()`/`end()`, kind of like the range interface in the STL ?

SimeonEhrig:
> What is `alpaka::view` here ?
> 
> Is it the same as the "view" that provides access to memory (on some device) with a buffer-like interface ?
> 
> Or something like an `std::pair` that provides `begin()`/`end()`, kind of like the range interface in the STL ?

Good point. I'm not talking about the `alpaka::view` of alpaka mainline. It's a hypothetical `alpaka3::view`. A `alpaka3::view` should be a buffer-like interface. 
I also think we need at least an additional device as parameter to construct the view. I think that's makes more clear.

```c++
int * out_begin = //...;
int * out_end = //...;

alpaka::transform<Tag>(queue, alpaka3::view{queue.dev(), out_begin, out_end}, op, in);
```

fwyzard:
Ah, I see what you mean.

Then that

> ```c++
> int * out_begin = //...;
> int * out_end = //...;
> 
> alpaka::transform<Tag>(queue, alpaka3::view{queue.dev(), out_begin, out_end}, op, in);
> ```

is basically the buffer-based approach:
```c++
alpaka::transform<Tag>(queue, out, op, in)
```

except that you pass one view/buffer and one pointer.

I find this mix to be a bit confusing though. Wouldn't it be better to simply pass the buffer/views ?

fwyzard:
In general, I think the preferred approach using alpaka is to pass buffer/views for all parameters.

The pointer-based interface is there to simplify the code if it doesn't use alpaka for the memory management.

SimeonEhrig:
My idea is that you can create a temporary `alpaka3::view` with a begin and end pointer in place or construct a `alpaka::view3` variable with the begin and end pointer and pass it to `alpaka::transform`.

There are two benefits of the approach

1. Remove the need to overload the algorithm (e.g. `alpaka::transform`) with a buffer version and a begin/end-pointer version. We need only to implement a `alpaka3::view` constructor, which takes the begin and end-pointer.
2. We can add additional compile and runtime information to the pointer. Add the moment, the algorithm simply expected some pointer properties, like the memory is located on the same device like the output buffer. But we cannot ask for it.

test/unit/algo/src/Transform.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+    // Check the correctness of the results.
+    for(Idx i = 0; i < size; ++i)
+    {
+        float expected = Duplicate{}(sample[i]);
+        REQUIRE_THAT(result[i], Catch::Matchers::WithinULP(expected, 4));
+    }
+}
+
+TEMPLATE_LIST_TEST_CASE("transform_buffer", "[algo]", alpaka::test::TestAccs)
```
SimeonEhrig, line 99: If I'm not wrong, the test should test 1D, 2D and 3D memory because of the dimension of the Acc type. If you need to change the code (e.g. @psychocoderHPC found something), can please add this information in the test name or as comment. But is is not mandatory would trigger the CI pipeline :grimacing:

SimeonEhrig:
@psychocoderHPC Can you please also do a review. From my side, it looks good but pitch calculation of multi dimensional memory is not easy. A second review would be helpful.

SimeonEhrig:
> My idea is that you can create a temporary `alpaka3::view` with a begin and end pointer in place or construct a `alpaka::view3` variable with the begin and end pointer and pass it to `alpaka::transform`.
> 
> There are two benefits of the approach
> 
>     1. Remove the need to overload the algorithm (e.g. `alpaka::transform`) with a buffer version and a begin/end-pointer version. We need only to implement a `alpaka3::view` constructor, which takes the begin and end-pointer.
> 
>     2. We can add additional compile and runtime information to the pointer. Add the moment, the algorithm simply expected some pointer properties, like the memory is located on the same device like the output buffer. But we cannot ask for it.

We discussed it in the VC today. For alpaka mainline, we keep the pointer overload. For alpaka3, we will only support memory with the view interface.

test/unit/algo/src/Transform.cpp
fwyzard: @SimeonEhrig sorry but I did not understand what is the issue here ?

SimeonEhrig: The problem is, I asked myself if the test is testing multi dimensional memory. In the end, I checked out your code and added an `Info()` call to display the dimension of the `Acc` type  to verify that multi dimensional memory is tested. To avoid this work for somebody who want's to understand the test without knowing that `alpaka::test::TestAccs` contains `Acc`s, with different dimensionality, a comment would help.

fwyzard: Ehm, OK, I don't disagree - but we already use
```c++
TEMPLATE_LIST_TEST_CASE("test", "[test]", alpaka::test::TestAccs)
```
in more than 80 tests in alpaka, and none of the other has a comment 🤷🏻

SimeonEhrig: That's true. I mentioned it, because this was the only part of the PR, where I was not sure by reading the code and getting the check marks from the CI. Therefore I really need to modify the code and compile it again. And I know, there are many tests, which it was not mentioned. Therefore I wrote that it is not mandatory, only a improvement.

psychocoderHPC:
@fwyzard you can rebase against the develop branch, #2566 is merged

fwyzard:
Thanks.

In fact, maybe I don't need to do another `git rebase` if GitHub is already happy ?

fwyzard:
@psychocoderHPC if you can approve the PR, I can try to merge it - and rebase it if that fails.

SimeonEhrig (approved):

PR #2547, opened 2025-07-08
Implement concept `alpaka::concepts::Acc`

fwyzard:
Implement concept `alpaka::concepts::Acc`.
Update code to use concepts and `requires` clauses instead of SFINAE and `std::enable_if`.

### DIFF
--- a/include/alpaka/acc/Traits.hpp
+++ b/include/alpaka/acc/Traits.hpp
@@ -11,7 +11,6 @@
 #include "alpaka/dev/Traits.hpp"
 #include "alpaka/dim/Traits.hpp"
 #include "alpaka/idx/Traits.hpp"
-#include "alpaka/kernel/Traits.hpp"
 #include "alpaka/platform/Traits.hpp"
 #include "alpaka/queue/Traits.hpp"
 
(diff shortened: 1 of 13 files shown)


SimeonEhrig (approved):

SimeonEhrig:
Nice work. I like how concepts make the code more readable :-)

include/alpaka/acc/Traits.hpp
```diff
@@ -21,13 +20,15 @@
...
 namespace alpaka
 {
-    struct ConceptAcc
+    struct InterfaceAcc
     {
     };
 
-    //! True if TAcc is an accelerator, i.e. if it implements the ConceptAcc concept.
-    template<typename TAcc>
-    inline constexpr bool isAccelerator = interface::ImplementsInterface<ConceptAcc, TAcc>::value;
```
fwyzard, line 30: should we keep (add back) `isAccelerstor`, in case any users' code uses it ?
Maybe with a deprecation warning ?

SimeonEhrig: That's a good point. I didn't thought about this. Yes please bring it back with a deprecation warning.

fwyzard: Eh, me neither, I only thought about it this morning. 

Will do!

fwyzard: Done in https://github.com/alpaka-group/alpaka/pull/2549 .

PR #2551, opened 2025-07-09
remove boost from regular CI jobs

SimeonEhrig:
Install boost only, if C++ 20 `std::atomic_ref` is not available.

I had to duplicate the C++ test code for `std::atomic_ref` because the original code fails to compile. For the script it is easier, if the code compiles and returns a status code, depending if the feature is available or not.


fwyzard:
If you prefer to avoid compiling and running a test program, you could use something like
```bash
echo '#include <atomic>' | $CXX -x c++ -std=c++20 -dM -E - | grep -q __cpp_lib_atomic_ref
```

(except that I do not know the incantation for MSVC)

fwyzard (dismissed):

SimeonEhrig:
> If you prefer to avoid compiling and running a test program, you could use something like
> 
> ```shell
> echo '#include <atomic>' | $CXX -x c++ -std=c++20 -dM -E - | grep -q __cpp_lib_atomic_ref
> ```
> 
> (except that I do not know the incantation for MSVC)

I agree, if this is working we can use this. It's less work. But I'm afraid if we don't use CMake, I have to implement at least two different solutions for Windows and Linux/Mac. So let's test it.

SimeonEhrig:
> > If you prefer to avoid compiling and running a test program, you could use something like
> > ```shell
> > echo '#include <atomic>' | $CXX -x c++ -std=c++20 -dM -E - | grep -q __cpp_lib_atomic_ref
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (except that I do not know the incantation for MSVC)
> 
> I agree, if this is working we can use this. It's less work. But I'm afraid if we don't use CMake, I have to implement at least two different solutions for Windows and Linux/Mac. So let's test it.

@fwyzard I tried it but I cannot handle Windows. Nether finding the cl.exe nor setting the correct compiler flags.

fwyzard:
Playing on godbolt, the flags should be something like 
```
cl.exe test.cc /std:c++20 /EP /d1PP
```

No idea how to _find_ cl.exe, though.

SimeonEhrig:
> Playing on godbolt, the flags should be something like
> 
> ```
> cl.exe test.cc /std:c++20 /EP /d1PP
> ```
> 
> No idea how to _find_ cl.exe, though.

To be honest, I don't want to maintain an extra solution for a system, where I have no access. Therefore I'm fine to have an "big" CMake project, which cares about compiler detection and setting the flags, instead a lightweight solution where I have to do this by myself.

psychocoderHPC (approved):

PR #2556, opened 2025-07-29
SIMD implementation and examples

mehmetyusufoglu:
**Alpaka SIMD implementation**

A Two-Level SIMD Architecture
**Level 1**: std::experimental::simd  

- Uses the standardized C++ SIMD library if available
- Detects and uses the SIMD instructions (SSE2, AVX2, AVX512)
- Disabled automatically in mixed CPU+GPU builds due to NVCC limitations

**Level 2:** Array-Based Fallback Backend
- Fallback implementation when std::experimental::simd is unavailable, or not supported
- Uses std::array with compile-time loop unrolling
-  `Cpu-gpu  mixed backend`  and` cpu-serial backend with clang + libc++ builds` use this fallback

Key Features
 SIMD width determined depending on the accelerator type at compile-time

**APIs**
**2 different APIs:**  Low-level explicit SIMD types/operations and high-level SimdAlgo forEach abstraction.

Three SIMD Examples
**vectorAddSIMD**: Has 2 kernels. Shows both `explicit SIMD type/operator` usage and `SimdAlgo forEach` approaches
**reduceRGBSIMD**: Real-world RGB-to-grayscale conversion with 8.8x-10.7x speedups
**sumOfSquaresSIMD**: Demonstration of SIMD usage (reduction) 

Performance Results
 
Screenshot from 2025-08-18 13-57-47

Screenshot from 2025-08-18 14-48-48

**TEST1:   RUN of EXAMPLE with FALLBACK Simd-Policy:**
 (Cmake fallback simd-policy setting: ALPAKA_FORCE_DISABLE_STD_SIMD    ON)

```
./example/examplesUsingSIMD/reduceRGBSIMD/reduceRGBSIMD 
Elements: 33554432
Accelerator: AccCpuSerial<1,long unsigned int>
Testing packed RGBA SIMD processing...

=== SIMD System Information ===
Compiler: GCC 13.3
Standard Library: libstdc++ 13
SIMD Instructions: AVX2
SIMD Policy: Fallback
Accelerator Setup: CPU only (Serial)
Compiler Environment: libstdc++ stdlib | AVX2 instructions | Release build
================================
simdWidth for packed RGBA (uint32_t) is 8
[WorkDiv CPU-Serial] numElements=33554432 simdWidth=8 packCount=4194304 gridBlocks=524288 threadsPerBlock=1 logicalThreads=524288 hardwareThreads=1
Device buffer aligned: YES
Scalar: 524288 threads (524288 blocks × 1 threads) simdWidth=8
SIMD  : 524288 threads (524288 blocks × 1 threads) simdWidth=8 (in-place)

--- Performance Summary ---
scalar_time_ms = 137.032
simd_time_ms   = 15.647
speedup        = 8.758 (scalar/simd)
simd policy    = Fallback
---------------------------
```

**TEST2: RUN with default Simd-Policy (std::experimental::simd)**
```

./example/examplesUsingSIMD/reduceRGBSIMD/reduceRGBSIMD
Elements: 33554432
Accelerator: AccCpuSerial<1,long unsigned int>
Testing packed RGBA SIMD processing...

=== SIMD System Information ===
Compiler: GCC 13.3
Standard Library: libstdc++ 13
SIMD Instructions: AVX2
SIMD Policy: std::experimental::simd
Accelerator Setup: CPU only (Serial)
Compiler Environment: libstdc++ stdlib | AVX2 instructions | Release build
================================
simdWidth for packed RGBA (uint32_t) is 8
[WorkDiv CPU-Serial] numElements=33554432 simdWidth=8 packCount=4194304 gridBlocks=524288 threadsPerBlock=1 logicalThreads=524288 hardwareThreads=1
Device buffer aligned: YES
Scalar: 524288 threads (524288 blocks × 1 threads) simdWidth=8
SIMD  : 524288 threads (524288 blocks × 1 threads) simdWidth=8 (in-place)

--- Performance Summary ---
scalar_time_ms = 184.336
simd_time_ms   = 14.527
speedup        = 12.689 (scalar/simd)
simd policy    = std::experimental::simd
---------------------------
```


SimeonEhrig:
Did you changed anything on the job generator? Looks like only reformating.

example/reduceRGBbySIMD/src/reduceRGBbySIMD.cpp
```diff
@@ -0,0 +1,324 @@
...
+    alpaka::memcpy(queue, devicePackedRGB, packedRGB);
+
+    float scalarPackedTime = 0.0f;
+    float simdPackedTime = 0.0f;
+
+    // Test 1: Scalar Packed RGBA (FAIR COMPARISON - same data format, same algorithm)
+    {
+        std::cout << "Scalar: " << workDiv << " threads, " << simdWidth << " elements each" << std::endl;
+
+        auto start = std::chrono::high_resolution_clock::now();
```
psychocoderHPC: `wait()` is missing it could be you messure the memcopy if you use a sync queue.

mehmetyusufoglu:
> Did you changed anything on the job generator? Looks like only reformating.

I removed, just reformatting.

CMakeLists.txt
```diff
@@ -36,6 +36,7 @@ set_property(GLOBAL PROPERTY USE_FOLDERS ON)
 
 option(alpaka_BUILD_EXAMPLES "Build the examples" OFF)
 option(alpaka_BUILD_BENCHMARKS "Build the benchmarks" OFF)
+option(ALPAKA_FORCE_DISABLE_STD_SIMD "Force use of fallback PortableSimd implementation instead of std::experimental::simd" OFF)
```
psychocoderHPC: please rename to alpaka_USE_STD_SIMD set the value to off to use our portable simd type.
If you have disable in the name you have a double negation if you set it to off. This is making things complicated.
btw all variable start always with a small alpaka

CMakeLists.txt
```diff
@@ -95,6 +96,12 @@ if(alpaka_BUILD_EXAMPLES OR alpaka_BUILD_BENCHMARKS OR BUILD_TESTING)
 endif()
 include(${_alpaka_ROOT_DIR}/cmake/alpakaCommon.cmake)
 
+# Propagate force-disable SIMD flag if requested
+if(ALPAKA_FORCE_DISABLE_STD_SIMD)
+  add_compile_definitions(ALPAKA_FORCE_DISABLE_STD_SIMD=1)
```
psychocoderHPC: take care that you have somewhere in the code a macro which is falling back to a usefull default in case the define is not set. This is e.g. the case in standalone headers where cmake is not used.

psychocoderHPC (changes requested):
I checked only the beginning, this is not a complete review

CMakeLists.txt
mehmetyusufoglu: ok, changed to alpaka_USE_STD_SIMD and corrected code accordingly . Default is ON.

CMakeLists.txt
mehmetyusufoglu: oki will add this into the code; to make it available even cmake is not used.
 `#ifndef ALPAKA_USE_STD_SIMD #define ALPAKA_USE_STD_SIMD 1 #endif `
 
 There is also internal variable ALPAKA_SIMD_USE_STD_EXPERIMENTAL, i will change its name sincethey are very similar.

mehmetyusufoglu: Ok i removed internal variable ALPAKA_SIMD_USE_STD_EXPERIMENTAL from the code. If ALPAKA_USE_STD_SIMD is ON at the cmake-configuration AND alpaka is technically available by internal code macro ALPAKA_SIMD_ENV_STD_EXP_SIMD_AVAILABLE; then std::simd will be used.

docs/source/advanced/simd.rst
```diff
@@ -0,0 +1,80 @@
...
+
+Provide portable SIMD operations on CPU (vectorized) and GPU (scalar, width=1) with unified API.
+
+Required Build Flags
+--------------------
+
+Always use:
+
+* ``-O3`` (Release build)
+* ``-march=native`` (CPU-specific optimizations)
```
fwyzard, line 15: To make the resulting binaries a bit more generic, rather than `-march=native` you could use the x86-64 [micro-architectures](https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels):
  - `-march=x86-64-v2` to enable only SSE3/SSE4;
  - `-march=x86-64-v3` to enable also AVX/AVX2 and FMA;
  - `-march=x86-64-v4` to enable also AVX512.

example/CMakeLists.txt
fwyzard, line 1: I guess the new tests should be added in alphabetical order, like to existing ones ?

mehmetyusufoglu: ok, doing.

docs/source/advanced/simd.rst
mehmetyusufoglu: Ok added, thanks i did not know. (Although i could ONLY test at `clang` and `g++`, i also added corresponding flags for other compilers.)

example/CMakeLists.txt
mehmetyusufoglu: done, thanks.

PR #2558, opened 2025-08-01
fix windows operation system detection

### LINKED ISSUE #2555: Windows not detected when compiling with msvc
In `include/alpaka/core/Config.hpp` windows should be detected by following code, but none of the tested preprocessor definitions are defined by msvc
```cpp
// WINDOWS
#if !defined(ALPAKA_OS_WINDOWS)
#    if defined(__WIN32__) || defined(__MINGW32__) || defined(WIN32)
#        define ALPAKA_OS_WINDOWS 1
#    else
#        define ALPAKA_OS_WINDOWS 0
#    endif
#endif
```
adding a test for `_WIN32` fixes the issue for me.

---

Tested on:
```
OS: Windows 11 64bit
Compiler: msvc 17.14.9 (July 2025)
```

---

AFAI researched, msvc defines `_WIN32`, but it doesnt define any of the tested preprocessor definitions.
Further, according to [this article](https://web.archive.org/web/20140625123925/http://nadeausoftware.com/articles/2012/01/c_c_tip_how_use_compiler_predefined_macros_detect_operating_system), all compilers define `_WIN32` on windows, thus checking for only `_WIN32` would be sufficient.

---

I can create a PR if this is confirmed

psychocoderHPC:
fix #2555

### DIFF
--- a/include/alpaka/core/Config.hpp
+++ b/include/alpaka/core/Config.hpp
@@ -26,7 +26,7 @@
 
 // WINDOWS
 #if !defined(ALPAKA_OS_WINDOWS)
-#    if defined(__WIN32__) || defined(__MINGW32__) || defined(WIN32)
+#    if defined(_WIN64) || defined(__MINGW64__)
 #        define ALPAKA_OS_WINDOWS 1
 #    else
 #        define ALPAKA_OS_WINDOWS 0


include/alpaka/core/Config.hpp
```diff
@@ -26,7 +26,7 @@
 
 // WINDOWS
 #if !defined(ALPAKA_OS_WINDOWS)
-#    if defined(__WIN32__) || defined(__MINGW32__) || defined(WIN32)
+#    if defined(_WIN64) || defined(__MINGW64__)
```
SimeonEhrig, line 29: Why did you use `_WIN64` instead `_WIN32`. Is there a reason why we should not support 32Bit?

psychocoderHPC: alpaka is not supporting 32bit therefore I choosed _WIN64

If we would support 32bit we need to test it because e.g. size_t will be 32bit too. Long time ago we decided not to support 32bit.

SimeonEhrig (approved):

PR #2560, opened 2025-08-13
Port CI job-generator to using bashi

SimeonEhrig:
[bashi](https://github.com/SimeonEhrig/bashi) is our new combination library which generates the CI jobs. It will replace the [alpaka-job-matrix-library](https://github.com/alpaka-group/alpaka-job-matrix-library).

# Review Infos

@psychocoderHPC @chillenzer Some information for the review. The code is more or less a rewrite of the `job-generator` in the folder `script/job_generator`. When the review is done, I will delete the old code and rename the folder `job-generator2` to `job-generator`. At the moment, I keep the new folder to make the diff much more readable.

# Known Problems

- [x] README.md is missing
- [x] ~the job shuffle does not work~ -> Shuffle works. For the first moment, it looks like shuffle does not work, because the jobs inside a waves are sorted by name. But that's not a problem, because the are executed in parallel. The jobs are not global sorted, which means for example all GCC jobs runs a the first and then all Clang jobs follow and so one.
- [x] writing custom jobs like [here](https://github.com/alpaka-group/alpaka/pull/2560/changes#diff-57083f4603182a6537327ffefedbdc5dceb9f9ae24a4d15d741db707839f2c41) is much to verbose. There is a helper function in bashi in tests, which I could use. But I need to publish the functionality.
- [x] `bashi` is released on pypi -> https://pypi.org/project/bashi-ci/


script/job_generator2/src/alpaka_bashi/versions.py
```diff
@@ -0,0 +1,92 @@
...
+    # Relationships can be easily extended. The following example assumes that bashi has already
+    # defined the relationship for Clang-CUDA 7 up to 17 and the CUDA SDK. The relationship is to be
+    # extended up to Clang-CUDA 22.
+    #
+    # clang_cuda_max_cuda_version = CLANG_CUDA_MAX_CUDA_VERSION + [
+    #    ClangCudaSDKSupport("18", "12.3"),
+    #    ClangCudaSDKSupport("22", "13.0"),
+    # ]
+    #
+    # bashi.VersionRelation(clang_cuda_max_cuda_version=clang_cuda_max_cuda_version)
```
SimeonEhrig, line 85: At the moment, the example is a copy of the real code. So as bashi supports Clang 22, the real code will be removed.

script/before_install.sh
```diff
@@ -115,13 +115,15 @@ if [ "${alpaka_ACC_SYCL_ENABLE}" == "OFF" ]; then
...
+    export alpaka_SYCL_ONEAPI_GPU_DEVICES=${alpaka_SYCL_ONEAPI_GPU_DEVICES:=""}
     export alpaka_SYCL_ONEAPI_FPGA=${alpaka_SYCL_ONEAPI_FPGA:=""}
     export alpaka_SYCL_ONEAPI_FPGA_MODE=${alpaka_SYCL_ONEAPI_FPGA_MODE:=""}
     export alpaka_SYCL_ONEAPI_FPGA_BOARD=${alpaka_SYCL_ONEAPI_FPGA_BOARD:=""}
     export alpaka_SYCL_ONEAPI_FPGA_BSP=${alpaka_SYCL_ONEAPI_FPGA_BSP:=""}
 else
-    if !( [ "$alpaka_SYCL_ONEAPI_CPU" == "ON" ] || [ "$alpaka_SYCL_ONEAPI_FPGA" == "ON" ] ); then
-        echo_red "ERROR: the SYCL CPU or FPGA device needs to enabled"
+    if !( [ "$alpaka_SYCL_ONEAPI_CPU" == "ON" ] || [ "$alpaka_SYCL_ONEAPI_GPU" == "ON" ] || [ "$alpaka_SYCL_ONEAPI_FPGA" == "ON" ] ); then
+        echo_red "ERROR: the SYCL CPU, GPU or FPGA device needs to enabled"
```
chillenzer:
```suggestion
        echo_red "ERROR: the SYCL CPU, GPU or FPGA device needs to be enabled"
```

script/before_install.sh
```diff
@@ -130,10 +132,26 @@ else
...
+    if [ "$alpaka_SYCL_ONEAPI_CPU" == "ON" ] && [ "$alpaka_SYCL_ONEAPI_GPU" == "ON" ]; then
+        echo_red "ERROR: the SYCL CPU or GPU device cannot be enabled at the same time"
+        exit 1
+    fi
+
+    if [ "$alpaka_SYCL_ONEAPI_GPU" == "ON" ] && [ "$alpaka_SYCL_ONEAPI_FPGA" == "ON" ]; then
+        echo_red "ERROR: the SYCL GPU or FPGA device cannot be enabled at the same time"
+        exit 1
+    fi
+
```
chillenzer, line 144: Why does this script have such checks? Don't we have full control over what variables are set at this point? Double-checking a list of combinations by hand here, seems brittle and intense to maintain. Also above, of course...

script/before_install.sh
```diff
@@ -142,6 +160,12 @@ else
     fi
 fi
 
+#-------------------------------------------------------------------------------
+# Install paths
+export ALPAKA_CI_CMAKE_DIR=$HOME/cmake
+export ALPAKA_CI_CUDA_DIR=$HOME/cuda
+export ALPAKA_CI_HIP_ROOT_DIR=$HOME/hip
```
chillenzer: Why do you set these? I can't find a place where those are used.

script/job_generator2/pyproject.toml
```diff
@@ -0,0 +1,38 @@
...
+[project]
+name = "alpaka-bashi"
+maintainers = [
+    {name = "Simeon Ehrig", email = "s.ehrig@hzdr.de"}, 
+]
+authors = [
+    {name = "Simeon Ehrig", email = "s.ehrig@hzdr.de"}, 
+]
+readme = "README.md"
+requires-python = ">=3.10"
```
chillenzer, line 14: 3.10 is EOL in October. You might want to already switch to 3.11.

script/job_generator2/tests/test_alpaka_filter.py
```diff
@@ -0,0 +1,430 @@
...
+                (ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE, ON),
+                (HOST_COMPILER, GCC, 11),
+                (ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE, ON),
+                (ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE, ON),
+                (ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE, ON),
+                (DEVICE_COMPILER, GCC, 11),
+                (ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE, ON),
+            ],
+        ]:
+            self.assertTrue(alpaka_filter_typechecked(parse_param_value_tuples(row)), f"{row}")
```
chillenzer: If you provide a custom failure message, I'd suggest a little more than just the data.

chillenzer: `for ...: assert ...` can hide valuable information becuase it will stop at the first failure prohibiting conclusions like "oh, all but this one combination pass... there must be one particular code path to check". `unittest` is not very well equipped for this kind of parametrized test (`pytest` is [much better in that respect](https://docs.pytest.org/en/7.1.x/example/parametrize.html)) but I'd suggest to at least use a [`subTest`](https://docs.python.org/3/library/unittest.html#distinguishing-test-iterations-using-subtests) for such situations.

script/job_generator2/tests/test_alpaka_filter.py
```diff
@@ -0,0 +1,430 @@
...
+            [
+                (ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE, ON),
+                (HOST_COMPILER, GCC, 11),
+                (ALPAKA_ACC_CPU_B_SEQ_T_THREADS_ENABLE, ON),
+                (ALPAKA_ACC_CPU_B_OMP2_T_SEQ_ENABLE, ON),
+                (ALPAKA_ACC_CPU_B_SEQ_T_OMP2_ENABLE, ON),
+                (DEVICE_COMPILER, GCC, 11),
+                (ALPAKA_ACC_CPU_B_TBB_T_SEQ_ENABLE, ON),
+            ],
+        ]:
```
chillenzer: Please extract this into a named constant `VALID_BACKEND_COMBINATIONS` outside of this function. (Of course, apply everywhere!)

Also, do all of these combinations test different logic/code paths? If not, I'd prefer to reduce the number to a manageable amount. Tests should be about logic not about data.

script/job_generator2/tests/test_alpaka_filter.py
```diff
@@ -0,0 +1,430 @@
...
+        runtime_info = {
+            RT_HOST_COMPILER_CUDA_SUPPORT: alpaka_bashi.runtime_info.get_rt_func_host_compiler_supports_cuda(
+                input_versions=input_versions, version_relation=get_alpaka_version_relation()
+            )
+        }
+
+        rt_host_compiler_cuda_support = cast(
+            alpaka_bashi.runtime_info.HostCompilerSupportsCuda,
+            runtime_info[RT_HOST_COMPILER_CUDA_SUPPORT],
+        )
```
chillenzer: I don't see the reason to cast here. Why don't you tighten the annotation on `get_rt_func_host_compiler_supports_cuda`?

script/job_generator2/tests/test_alpaka_filter.py
```diff
@@ -0,0 +1,430 @@
...
+@typechecked
+def alpaka_filter_typechecked(
+    row: ParameterValueTuple,
+    output: Optional[IO[str]] = None,
+    runtime_info: Dict[str, Callable[..., bool]] | None = None,
+) -> bool:
+    """Type-checked version of SoftwareDependencyFilter()(). Type checking has a big performance
+    cost, which is why the non type-checked version is used for the pairwise generator.
+    """
+    return alpaka_bashi.filter.AlpakaFilter(output=output, runtime_infos=runtime_info)(row)
```
chillenzer: I don't quite see why you would want to use this. If your tests below verify that `AlpakaFilter` has all the semantics that you want, you don't have to care what types it handles.

script/job_generator2/tests/test_find_latest_cuda_sdk_minor_versions.py
```diff
@@ -0,0 +1,92 @@
...
+            ],
+        ]:
+            comb_list.append(parse_param_value_tuples(row))
+
+        expected_result: Dict[ValueName, List[ValueVersion]] = {}
+        expected_result[GCC] = sorted([packaging.version.parse(str(ver)) for ver in [12.1, 12.8]])
+        expected_result[CLANG] = sorted(
+            [packaging.version.parse(str(ver)) for ver in [11.1, 11.8, 12.1, 12.4]]
+        )
+        expected_result[CLANG_CUDA] = sorted([packaging.version.parse(str(ver)) for ver in [12.2]])
```
chillenzer:
```suggestion
        expected_result: Dict[ValueName, List[ValueVersion]] = {
        GCC: sorted([packaging.version.parse(str(ver)) for ver in [12.1, 12.8]])
        CLANG: sorted(
            [packaging.version.parse(str(ver)) for ver in [11.1, 11.8, 12.1, 12.4]]
        )
        CLANG_CUDA: sorted([packaging.version.parse(str(ver)) for ver in [12.2]])
        }
```

script/job_generator2/tests/test_find_latest_cuda_sdk_minor_versions.py
```diff
@@ -0,0 +1,92 @@
...
+                (DEVICE_COMPILER, NVCC, 11.6),
+                (ALPAKA_ACC_GPU_CUDA_ENABLE, 11.6),
+            ],
+            [
+                (HOST_COMPILER, CLANG_CUDA, 18),
+                (DEVICE_COMPILER, CLANG_CUDA, 18),
+                (ALPAKA_ACC_GPU_CUDA_ENABLE, 12.2),
+            ],
+        ]:
+            comb_list.append(parse_param_value_tuples(row))
```
chillenzer, line 71: Anti-pattern: `for...: my_list.append(...)` (I'll put this here where it is admittedly not most relevant. Please apply everywhere.)

Reasoning:
- Not declarative (when there's something to declare): The (at least) 3 lines of code 
```python
my_list = []
for ...:
    my_list.append(...)
```
do not structurally imply the intention of initialising a container. They are a prescription that only happens to result in this. The single line
```python
my_list = [...]
```
is structurally an assignment, so the reader can be certain that the purpose of this code is to initialise a list (except for cases for serious misuse of list comprehensions).

- Fragile state handling: It might well be that the loop gets interrupted half-way through. In that case, we'd end up with a half-filled list (that we might already hold references to somewhere else). There's no way to guarantee a consistent state of `my_list`. In the declarative approach, `my_list` only exists after the full list comprehension has passed.

- Maintainability: It's all to easy to sneak in other statements into such a loop. After a few iterations, you'll have a blown up loop the original intention of which is almost completely lost. You might even introduce control flow `if... else...` and potentially miss to have the `append` in every branch.

script/job_generator2/tests/utils.py
```diff
@@ -0,0 +1,168 @@
...
+                )
+
+        expected_val_pairs.append(
+            bashi.utils.create_parameter_value_pair(
+                regular_entry_pair[0][0],
+                regular_entry_pair[0][1],
+                regular_entry_pair[0][2],
+                regular_entry_pair[1][0],
+                regular_entry_pair[1][1],
+                regular_entry_pair[1][2],
```
chillenzer:
```suggestion
                *regular_entry_pair[0], *regular_entry_pair[1]
```
or even (`from itertools import chain`)
```suggestion
                *chain(*regular_entry_pair)
```

script/job_generator2/tests/utils.py
```diff
@@ -0,0 +1,168 @@
...
+                if len(input_entry) != 2:
+                    raise ValueError(
+                        f"input_list[{pair_number}][{entry_number}] {input_entry}\n"
+                        "The tuple needs to contain two entries:"
+                        "\n(<parameter>, <value-version>)"
+                    )
+                default_input_entry = cast(DefaultParsableParameterValue, input_entry)
+                regular_entry_pair.append(
+                    (default_input_entry[0], default_input_entry[0], default_input_entry[1])
+                )
```
chillenzer, line 138: I think this inline error handling severely hinders readability. Just consider the code without it

```suggestion
    expected_val_pairs: List[bashi.ParameterValuePair] = []
    for pair_number, input_pair in enumerate(input_list):
        regular_entry_pair: List[RegularParsableParameterValue] = []
        for entry_number, input_entry in enumerate(input_pair):
            if input_entry[0] in (bashi.HOST_COMPILER, bashi.DEVICE_COMPILER):
                compiler_input_entry = cast(CompilerParsableParameterValue, input_entry)
                regular_entry_pair.append(compiler_input_entry)
            else:
                default_input_entry = cast(DefaultParsableParameterValue, input_entry)
                regular_entry_pair.append(
                    (default_input_entry[0], default_input_entry[0], default_input_entry[1])
                )
```
(and don't get me started on list comprehensions, casts, etc.)...

I'd probably argue that in a `tests/utils.py` file, it's perfectly justify-able to not have perfect error handling as you don't have external users that could do completely unexpected things to your code but would not dare to look beneath the covers. But if you want to keep error handling, I'd suggest that you extract it into a separate function that you can call before the actual functional code.

script/job_generator2/README.md
```diff
@@ -0,0 +1,96 @@
+# Job-Generator
+
+The `job-generator` creates all jobs for GitLab CI.
+It takes multiple compiler and software versions as input and generates a test matrix by [pairwise combination](https://en.wikipedia.org/wiki/All-pairs_testing).
+When you run the `job-generator` process, the output is a valid GitLab CI YAML file for CI.
+
+The `job-generator` uses the [bashi](https://github.com/alpaka-group/bashi/) library, which provides the functions for creating combinations.
+`bashi` contains most of the filter rules specific to alpaka. The `job-generator` primarily handles the generation of the GitLab CI YAML file based on the values of a combination.
```
chillenzer, line 8: Is `bashi` specific to `alpaka` already? I would have thought that only `alpaka_bashi` in this repository is specific.

script/job_generator2/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+# Install
+
+The `job-generator` is a Python package and is very easy to install.
+
+```bash
+# create an environment
+# other tools, like micromamba can be used instead
+python3 -m venv .env
+source .env/bin/activate
```
chillenzer: I'd probably leave out instructions for environment management completely:
```suggestion
# create a virtual environment with your favourite tool, then
```
Users familiar with venvs will know what to do. Users unfamiliar with those would run the above commands blindly and then wonder why they can't access other packages that they might have installed in another environment.

script/job_generator2/README.md
```diff
@@ -0,0 +1,96 @@
...
+# Development
+
+If you want to modify the `job-generator` code, you can install it in an editable version.
+This means that the `job-generator` command will run your modified code.
+
+```bash
+# create an environment
+# other tools, like micromamba can be used instead
+python3 -m venv .env
+source .env/bin/activate
```
chillenzer: Same here.

script/job_generator2/README.md
```diff
@@ -0,0 +1,96 @@
...
+The CI job will fail.
+Therefore, a new software relationship must be added to the [get_alpaka_version_relation() function in the version.py file](./src/alpaka_bashi/versions.py), which restricts NVCC 13.1 to GCC 15.
+
+## Python Exception: covertable.exceptions.InvalidCondition
+
+Sometimes, when running the `job-generator`, an exception of type `covertable.exceptions.InvalidCondition` may be triggered when a new software version is added.
+This can happen if a filter rule or a software rule is missing.
+`bashi` provides a [best practices guide](https://github.com/alpaka-group/bashi/blob/main/docs/rules.md) for resolving the issue.
+
+**Attention:** Since `job-generator` is based on `bashi`, different tool calls are required.
```
chillenzer, line 81: I don't understand this sentence. Please expand.

script/job_generator2/README.md
```diff
@@ -0,0 +1,96 @@
...
+
+- To display which combinations the `job-generator` accepts and which it does not, run `job-generator --debug-print=args`. Green combinations pass and red ones are invalid. Run `job-generator --help` for more information.
+- Use `alpaka-validate` instead of `bashi-validate`. `alpaka-validate` uses the `bashi` filters as well as the alpaka-specific filter.
+
+## Update the Verification
+
+In most cases, the verification functions adapt to the new software version, and verification passes automatically. However, sometimes existing verification rules need to be adjusted or new ones added. The verification rules are defined in the file [verify.py](./src/alpaka_bashi/verify.py).
```
chillenzer, line 88: Does this mean that there is a duplication of data?

script/job_generator2/tests/runtime_info/test_rt_func_host_compiler_supports_cuda.py
```diff
@@ -0,0 +1,68 @@
...
+            alpaka_bashi.runtime_info.get_rt_func_host_compiler_supports_cuda(
+                input_versions=input_versions, version_relation=get_alpaka_version_relation()
+            ),
+        )
+        self.assertEqual(
+            host_compiler_support_cuda.get_max_version(GCC), packaging.version.parse("12")
+        )
+        self.assertEqual(
+            host_compiler_support_cuda.get_max_version(CLANG), packaging.version.parse("16")
+        )
```
chillenzer, line 68: Don't we have the same test in `test_alpaka_filter.py`?

script/job_generator2/tests/backend_filter/test_no_hip_compiler_backends.py
```diff
@@ -0,0 +1,76 @@
...
+                        ALPAKA_ACC_GPU_CUDA_ENABLE,
+                    ],
+                ),
+            ],
+        ]:
+
+            self.assertTrue(
+                alpaka_bashi.filter.no_hip_compiler_backends(backend_list),
+                backend_list,
+            )
```
chillenzer: This test is in the `test_no_hip_...` file but references `cuda` a lot. That doesn't sound right...

script/job_generator2/tests/backend_filter/test_get_valid_backend_combinations.py
```diff
@@ -0,0 +1,135 @@
...
+import alpaka_bashi.filter
+import bashi
+
+
+class TestAlpakaFilterValidGetBackendCombination(unittest.TestCase):
+    def test_serial_cpu_is_always_on(self):
+        input_row = parse_param_value_tuples([(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE, ON)])
+        self.assertGreater(
+            len(alpaka_bashi.filter.get_valid_compiler_backend_combinations(input_row)), 0
+        )
```
chillenzer: How does the name of the test relate to the semantics? If I explicitly set serial backend to `ON`, it is on by definition and not "always on". Or do I have a misunderstanding here?

chillenzer (changes requested):
First round, focused on tests.

script/job_generator2/tests/backend_filter/test_get_valid_backend_combinations.py
SimeonEhrig: Our test suite requires that the CMake flag `-Dalpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE=ON` is set and therefore in each test case the serial cpu is available.

script/job_generator2/tests/backend_filter/test_no_hip_compiler_backends.py
SimeonEhrig: I think, I placed it here because I was writing rule for a HIP backend. Nevertheless, I can solve the problem in a easy way. The function is not used, therefore I can remove it. I think it was a left over and non of the linters blamed it, because it is used in the tests 😅

script/job_generator2/tests/runtime_info/test_rt_func_host_compiler_supports_cuda.py
SimeonEhrig: Your are right. They are slightly different, but I can merge it.

script/job_generator2/tests/utils.py
SimeonEhrig: I understand your idea, but this code was already pretty helpful, when I used an AI to refactor my code.

script/before_install.sh
SimeonEhrig: There are two reasons. First it is also possible to write hand written jobs, e.g. for sanitizers.

Second there is a piratical reason. The CI bash scripts are a mess. It's pretty hard to find out, which variables needs to be set and which not. In the past we had the problem, that some variables was missing and therefore the CI jobs didn't test, which was expected. I implemented this guards for the "script black box", to be sure that the expected stuff is tested. This was my solution to solve the problem in reasonable time.

script/before_install.sh
SimeonEhrig: Do you mean all 3 variables or only the variable `ALPAKA_CI_HIP_ROOT_DIR`? I'm not sure, if GitHub shows only the diff block. I removed `ALPAKA_CI_HIP_ROOT_DIR`, because it was only checked, if the variable was set.

`ALPAKA_CI_CMAKE_DIR` is used in the `install_cmake.sh` and `install_boost.sh`.
`ALPAKA_CI_CUDA_DIR` is used in the `install_cuda.sh`.

script/job_generator2/pyproject.toml
SimeonEhrig: This is related to [bashi](https://github.com/alpaka-group/bashi/blob/main/pyproject.toml). Normally I remove the old version when the new version is released. The CI is using Python 3.13. ~and I will open an PR in  bashi to support Python 3.14.~

Bashi now supports from 3.10 until 3.14. I don't want to increase the minimum Python version without reason, because I know people will use the system Python version of it's old Linux LTS.

script/job_generator2/README.md
SimeonEhrig: I think I need to rephrase it. bashi provides a rule set for alpaka-based projects (e.g. alpaka, alpaka3, PIConGPU and HASEonGPU), which forbids parameter-value combinations which are technically possible. The `job-generator` of the alpaka project itself puts rules on top, to reduce test workload, e.g. test only a subset of technical possible backend combinations.

script/job_generator2/README.md
SimeonEhrig: **Attention:** This guide refers to the `example.py` file and the `bashi-validate` tool. If you are applying this guide to the `job-generator` package, you must use the `job-generator` application instead of `example.py` and `alpaka-validate` instead of `bashi-validate`.

Better?

script/job_generator2/README.md
SimeonEhrig: If you mean with data, the different filter rules yes. I didn't find a better to verify my result, than describe in two different ways, how to calculate the expected result and cross check it with each other.

The problem is, the number of expected and unexpected results is so big, I manually say this is correct or this is not correct. Therefore I used a generative verification algorithm to prove my result. And to be correct, I don't know if my result is correct. With this approach I simply increased the chance, that it is correct.

SimeonEhrig:
@chillenzer I started with the "low-hanging fruits" of your review. I will also work on the other comments at the next days.

script/job_generator2/tests/test_alpaka_filter.py
SimeonEhrig: That's the error message:

```
.....................................F......
======================================================================
FAIL: test_valid_backend_combinations_a1 (test_alpaka_filter.TestAlpakaFilter.test_valid_backend_combinations_a1)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/simeon/projects/alpaka/script/job_generator2/tests/test_alpaka_filter.py", line 64, in test_valid_backend_combinations_a1
    self.assertTrue(alpaka_filter_typechecked(parse_param_value_tuples(row)), f"{row}")
    ~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: False is not true : [('alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE', '1.0.0'), ('host_compiler', 'gcc', 11), ('alpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE', '1.0.0'), ('alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE', '1.0.0'), ('alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE', '1.0.0'), ('device_compiler', 'gcc', 11), ('alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE', '0.0.0')]

----------------------------------------------------------------------
Ran 44 tests in 0.014s
```

I can repeat in the error message the information which are already there:
- what does it check? `test_valid_backend_combinations_a1` if the filter rule a1 is passed
- with which in input? ` [('alpaka_ACC_CPU_B_SEQ_T_SEQ_ENABLE', '1.0.0'), ('host_compiler', 'gcc', 11), ('alpaka_ACC_CPU_B_SEQ_T_THREADS_ENABLE', '1.0.0'), ('alpaka_ACC_CPU_B_OMP2_T_SEQ_ENABLE', '1.0.0'), ('alpaka_ACC_CPU_B_SEQ_T_OMP2_ENABLE', '1.0.0'), ('device_compiler', 'gcc', 11), ('alpaka_ACC_CPU_B_TBB_T_SEQ_ENABLE', '0.0.0')]`
- should or should not pass? `AssertionError: False is not true`

script/before_install.sh
SimeonEhrig: Clarified in a offline discussion. Only `ALPAKA_CI_HIP_ROOT_DIR` is not used.

script/job_generator2/pyproject.toml
SimeonEhrig: In a offline discussion. It's okay. The CI uses a more modern Python version and the minimum version is removed when the version is not supported anymore.

script/job_generator2/tests/test_alpaka_filter.py
SimeonEhrig: Clarified in an offline discussion: The output is fine.

script/job_generator2/tests/backend_filter/test_get_valid_backend_combinations.py
SimeonEhrig: Clarified in an offline discussion: The test does not make sense. I changed to test `(ALPAKA_ACC_CPU_B_SEQ_T_SEQ_ENABLE, OFF)` does not generate a combination.

script/job_generator2/tests/test_alpaka_filter.py
SimeonEhrig: I removed it.

script/job_generator2/tests/utils.py
SimeonEhrig: I removed all checks except the check for number of elements depending on the value of the first parameter.

script/job_generator2/README.md
SimeonEhrig: @chillenzer I overworked it. Please read again.

script/job_generator2/README.md
SimeonEhrig: @chillenzer I overworked it. Please read again.

SimeonEhrig:
@chillenzer I did all your requests.

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+    parser.add_argument(
+        "--no-image-check",
+        action="store_false",
+        help="Disable registry check for existing Docker image.",
+    )
+
+    parser.add_argument(
+        "--debug-print",
+        type=str,
+        choices=FILTER_MODE_ARGS.keys(),
```
chillenzer: It's cleaner and shorter to just define
```python
class FilterDebugMode(Enum):
    OFF="off"
    ...
```
and then you can just do
```suggestion
        type=bashi.FilterDebugMode,
```

This will correctly fail for invalid values and return the result in the correct type directly. `FILTER_MODE_ARGS` can be remove, of course.

chillenzer (commented):
I'll just dump a batch whenever I turn away from this again, so here's some further partial review.

script/job_generator/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+        comb_list = alpaka_bashi.filter_combinations(comb_list, job_filter_name)
+        print(f"number of filtered combinations: {len(comb_list)}", file=sys.stderr)
+
+    if args.print_combinations:
+        for c in comb_list:
+            bashi.print_row_nice(c)
+        sys.exit(0)
+
+    # shuffle jobs to increase the chance to run different compiler in the first wave
+    random.Random(42).shuffle(comb_list)
```
chillenzer, line 230: Blocking: That's shuffling with a fixed seed. It will always yield exactly the same output given the same input. Is that really what you want here?

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+                    image_check=args.no_image_check,
+                    stages=True,
+                    wave_sizes=wave_sizes,
+                )
+
+            if len(jobs) == 0:
+                jobs = alpaka_bashi.get_dummy_job_yaml(pipeline_name)
+
+            with open(output_path, "w", encoding="utf-8") as output_file:
+                alpaka_bashi.write_job_yaml(jobs, output_file)
```
chillenzer: Not blocking: Please extract functions for the different branches.

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+    # to stdout.
+    # We split up the pipelines and merge again, because in the meantime reorder operations can be
+    # applied on the different pipelines.
+    # By the way, it also automatically sort the jobs by pipeline.
+    if not args.split_pipeline:
+        single_pipeline: bashi.CombinationList = []
+        for pipeline in pipelines.values():
+            single_pipeline += pipeline
+        jobs = alpaka_bashi.get_job_yaml(
+            combination_list=single_pipeline,
```
chillenzer: Not blocking:
```suggestion
        jobs = alpaka_bashi.get_job_yaml(
            combination_list=list(chain(*pipelines.values())),
```

chillenzer (commented):
More partial review.

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+            if striped_line.strip().startswith(commit_message_filter_prefix):
+                return striped_line[len(commit_message_filter_prefix) :].strip()
+
+    if args.filter:
+        return args.filter
+
+    return ""
+
+
+# pylint: disable=too-many-locals
```
chillenzer: Why would you do this? I totally agree that there are "too many locals". I'd prefer to split this up into multiple functions but this is not blocking.

script/job_generator/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+
+    param_matrix: bashi.ParameterValueMatrix = bashi.get_parameter_value_matrix(
+        software_versions=software_versions
+    )
+
+    version_relation = alpaka_bashi.get_alpaka_version_relation()
+
+    alpaka_filter = alpaka_bashi.AlpakaFilter()
+    runtime_infos = bashi.get_runtime_infos(param_matrix, version_relation)
+    runtime_infos |= alpaka_bashi.get_runtime_infos(software_versions, version_relation)
```
chillenzer, line 199: Why do these two have different interfaces? Isn't that confusing?

script/job_generator2/src/alpaka_bashi/combination_modifier/execution_type.py
```diff
@@ -0,0 +1,176 @@
...
+    # make CUDA version unique
+    for compiler_name in versions:
+        versions[compiler_name] = list(set(versions[compiler_name]))
+
+    return versions
+
+
+@typechecked
+def execution_type_cuda_backend(combination_list: bashi.CombinationList):
+    """Annotate for each hipcc version one job as runtime. The rest is compile time."""
```
chillenzer:
```suggestion
    """Annotate for each cuda version one job as runtime. The rest is compile time."""
```

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+        parameter_value_matrix=param_matrix,
+        runtime_infos=runtime_infos,
+        custom_filter=alpaka_filter,
+        version_relation=version_relation,
+        # change me to display which combinations passed and did not pass the filter chain
+        debug_print=FILTER_MODE_ARGS[args.debug_print],
+    )
+    print(f"number of combinations: {len(comb_list)}", file=sys.stderr)
+
+    alpaka_bashi.add_combinations_parameters(comb_list)
```
chillenzer: Not blocking: I would prefer returning the new list instead of in-place modification. In addition to the typical reasons for functional-style programming, to me it was not quite clear if this was supposed to be a modification of the combination list or the change of some persistent internal state somewhere else (based on the combination list).

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+def setup_row_printer() -> None:
+    """Set extra configurations for the bashi.print_row_nice() function"""
+    bashi.add_print_row_nice_parameter_alias(alpaka_bashi.BUILD_TYPE, "buildType")
+    bashi.add_print_row_nice_parameter_alias(alpaka_bashi.JOB_EXECUTION_TYPE, "jobType")
+
+    for val_name, aliases in alpaka_bashi.get_version_aliases().items():
+        bashi.add_print_row_nice_version_alias(val_name, aliases)
+
+
+def get_filter(args: argparse.Namespace) -> str:
```
chillenzer:
```suggestion
def get_filter_name(args: argparse.Namespace) -> str:
```

script/job_generator/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+    wave_sizes = {
+        alpaka_bashi.CI_PIPELINE_COMPILE_ONLY_VER: 20,
+        alpaka_bashi.CI_PIPELINE_RUNTIME_CPU_VER: 20,
+    }
+
+    # If the pipelines are not splitted and therefore written to different files, write everything
+    # to stdout.
+    # We split up the pipelines and merge again, because in the meantime reorder operations can be
+    # applied on the different pipelines.
+    # By the way, it also automatically sort the jobs by pipeline.
```
chillenzer, line 238: Why is there a distinction in the control flow here? "Print to stdout" is the same as "having a single file to write to which happens to be stdout". I think this could be simplified significantly.

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+    # If the pipelines are not splitted and therefore written to different files, write everything
+    # to stdout.
+    # We split up the pipelines and merge again, because in the meantime reorder operations can be
+    # applied on the different pipelines.
+    # By the way, it also automatically sort the jobs by pipeline.
+    if not args.split_pipeline:
+        single_pipeline: bashi.CombinationList = []
+        for pipeline in pipelines.values():
+            single_pipeline += pipeline
+        jobs = alpaka_bashi.get_job_yaml(
```
chillenzer:
```suggestion
        jobs = alpaka_bashi.get_job_description(
```
or `definition` or the like. `yaml` reads to me like implying a data type (e.g., `str` encoding a `yaml`-based serialisation). Performing `|=` on it blew my mind and I had to look up what the function actually did.

script/job_generator2/src/apps/validate.py
```diff
@@ -0,0 +1,60 @@
...
+
+
+def main() -> None:
+    """Entry point function."""
+
+    validator = bashiValidate.Validator(version_relation=alpaka_bashi.get_alpaka_version_relation())
+    validator.parser.add_argument(
+        "--missing-parameters",
+        action="store_true",
+        help="Display all missing parameters, which was not set via application argument",
```
chillenzer:
```suggestion
        help="Display all missing parameters not set via application argument",
```

chillenzer (commented):
Next batch.

script/job_generator2/src/apps/job_generator.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+# pylint: disable=too-many-locals
+def main() -> None:
+    """Entry point"""
+    args = get_args()
+
+    setup_row_printer()
+
+    software_versions = alpaka_bashi.get_alpaka_version()
```
chillenzer:
```suggestion
    software_versions = alpaka_bashi.get_software_versions_for_alpaka()
```
(The other one is extremely misleading.)

script/job_generator2/src/alpaka_bashi/versions.py
```diff
@@ -0,0 +1,92 @@
...
+    """Return dict of all compiler and software versions, which should be used as input for the
+    combination generator.
+
+    Raises:
+        RuntimeError: If no valid Clang-CUDA versions exist.
+
+    Returns:
+        Dict[str, List[Union[str, int, float]]]: List of compiler and software versions.
+    """
+    alpaka_version = ALPAKA_VERSIONS.copy()
```
chillenzer:
```suggestion
    from copy import deepcopy
    alpaka_version = deepcopy(ALPAKA_VERSIONS)
```
This is safer (although I don't have immediate reason to assume that there's a logic flaw resulting from this): Simple copy will do a copy of the (mutable) dictionary you are working with but it will have the same (mutable!) list instances as content. So, the form of assignment you are doing in the following lines is safe
```python
my_dict[key] = new_value
```
because this mutates the dictionary itself but
```python
my_dict[key].append(new_value)
```
would still result in a change of the original instance.

Admittedly, the safest way would be to store data in immutable data structures, e.g., `namedtuple` and/or `tuple` but I don't know about your type constraints.

script/job_generator/src/alpaka_bashi/versions.py
```diff
@@ -0,0 +1,92 @@
...
+        RuntimeError: If no valid Clang-CUDA versions exist.
+
+    Returns:
+        Dict[str, List[Union[str, int, float]]]: List of compiler and software versions.
+    """
+    alpaka_version = ALPAKA_VERSIONS.copy()
+
+    clang_cuda_versions = _get_clang_cuda_versions()
+    if len(clang_cuda_versions) == 0:
+        raise RuntimeError("Alpaka custom filter does not work without Clang-CUDA version.")
```
chillenzer, line 66: Why's that?

script/job_generator2/src/alpaka_bashi/versions.py
```diff
@@ -0,0 +1,92 @@
...
+
+    clang_cuda_versions = _get_clang_cuda_versions()
+    if len(clang_cuda_versions) == 0:
+        raise RuntimeError("Alpaka custom filter does not work without Clang-CUDA version.")
+    alpaka_version[CLANG_CUDA] = clang_cuda_versions
+
+    return alpaka_version
```
chillenzer:
```suggestion
    return ALPAKA_VERSIONS | {CLANG_CUDA: _get_clang_cuda_versions()}
```
could be a very simple implementation to the same effect (except for `RuntimeError`).

chillenzer (commented):
Next batch.

script/job_generator/src/apps/job_generator.py
SimeonEhrig: Yes, we want to have the same ordering for each run. Generally, we structure things so that as many different configurations as possible (mainly different compilers) are tested at the beginning, allowing us to find errors as quickly as possible.

However, if a job fails, we want to keep the order of the jobs being tested. We test in waves. This means that N out of M jobs are always tested, and only if all jobs in a wave are successful is the next wave started (waves are required for fair use of the CI resources). If the order weren’t fixed, it could happen that you write and push a fix, and this time the failed job is in the second wave. In the first wave, there is another job that fails due to a different bug. In this case, you have only two options: fix the other bug as well, even though you don’t know if the first fix works, or keep force-pushing until the desired code is in the first wave.

script/job_generator/src/apps/job_generator.py
SimeonEhrig: The different interfaces are problem driven. `bashi.get_runtime_infos` requires the `param_matrix`. At the moment, I use the default function from bashi to create the `param_matrix` from the `software_versions` but there are use cases, where a project specific function will be used instead.

I could recreate the `software_versions` from the `param_matrix` with some coding effort but I don't see good cost-benefit-ration for this.

I could only add some unused parameters to unify the interfaces but I'm not sure if this is a benefit. Therefore I would keep it as it is because it is to much work for to few benefit.

script/job_generator/src/apps/job_generator.py
SimeonEhrig: After discussion, it is not easy to unify both code paths.

chillenzer (approved):
We've decided that the trade-off between having this mainline (with potentially minor details wrong or missing) vs. having this lying around as a PR is in favour of merging.

SimeonEhrig:
Please do a `squash and merge`. The commit history is not well defined.

.gitlab-ci.yml
```diff
@@ -22,22 +21,25 @@ variables:
 generate:
   stage: generator
   # update alpine version to use newer python version
-  image: alpine:3.18
+  image: python:3.13-alpine
```
psychocoderHPC, line 24: I am wondering why you downgraded the python version.

psychocoderHPC (approved):
I will merge it, if needed the python version can be increased again in a sperate PR.

.gitlab-ci.yml
SimeonEhrig: You missed that I changed the image. First I used an alpine image, where I installed python3 (alpine 18 provides 3.11). Now I use an image, where Python is preinstalled which bases on alpine. Therefore I updated from Python 3.11 to 3.13.

PR #2561, opened 2025-08-22
fix version macros

psychocoderHPC:
- fix and test version macros

The bug was e.g., with the icpx compiler version number, where the major version is the year, but the code silently cut `2025` to `25`.
The tests then showed that `ALPAKA_YYYYMM_TO_VERSION()` is wrong too.


psychocoderHPC:
@alpaka-group/alpaka-maintainers please review and merge

SimeonEhrig (approved):

PR #2562, opened 2025-08-26
Separate `ViewAccessOps` for host and device views

sbaldu:
Currently all types of Buffers and Views inherit from `ViewAccessOps`, and this provides accessors `operator*`, `operator[]`, etc even for device buffers, although using those methods would result in a seg fault.
In this PR I separated `ViewAccessOps` in a `HostViewAccessor` and `DeviceViewAccessor`, and a matryoshka template `ViewAccessorType` that selects the correct accessor depending on the device type. 

Note: I also rewrote the `isView` trait as a concept, since it was marked as TODO.
FYI @fwyzard

### DIFF
--- a/include/alpaka/mem/view/ViewAccessOps.hpp
+++ b/include/alpaka/mem/view/ViewAccessOps.hpp
@@ -4,6 +4,8 @@
 
 #pragma once
 
+#include "alpaka/acc/Tag.hpp"
+#include "alpaka/dev/DevGenericSycl.hpp"
 #include "alpaka/extent/Traits.hpp"
 #include "alpaka/mem/view/Traits.hpp"
 
@@ -16,8 +18,10 @@
 namespace alpaka
 {
 
-    struct DevCpu;
-    struct DevCpuSycl;
+    class DevCpu;
+
+    template<concepts::Tag TTag>
+    class DevGenericSycl;
 
 } // namespace alpaka
 
@@ -28,9 +32,15 @@ namespace alpaka::internal
     concept ViewType = requires {
         typename Idx<TView>;
         typename Dim<TView>;
-        { getPtrNative(std::declval<TView>()) };
-        { getPitchesInBytes(std::declval<TView>()) };
-        { getExtents(std::declval<TView>()) };
+        {
+            getPtrNative(std::declval<TView>())
+        };
+        {
+            getPitchesInBytes(std::declval<TView>())
+        };
+        {
+            getExtents(std::declval<TView>())
+        };
     };
 
     template<ViewType TView>
@@ -188,7 +198,7 @@ namespace alpaka::internal
     };
 
     template<>
-    struct ViewAccessor<alpaka::DevCpuSycl>
+    struct ViewAccessor<alpaka::DevGenericSycl<alpaka::TagCpuSycl>>
     {
         template<ViewType TView>
         using AccessorType = HostViewAccessor<TView>;
--- a/include/alpaka/mem/view/ViewAccessOps.hpp
+++ b/include/alpaka/mem/view/ViewAccessOps.hpp
@@ -4,7 +4,7 @@
 
 #pragma once
 
-#include "alpaka/dim/Traits.hpp"
+#include "alpaka/dev/DevCpu.hpp"
 #include "alpaka/extent/Traits.hpp"
 #include "alpaka/mem/view/Traits.hpp"
 
@@ -172,20 +172,18 @@ namespace alpaka::internal
         }
     };
 
-    struct DevCpu;
-
     template<typename TDev>
     struct ViewAccessor
     {
         template<ViewType TView>
-        using AccessorType = HostViewAccessor<TView>;
+        using AccessorType = DeviceViewAccessor<TView>;
     };
 
     template<>
     struct ViewAccessor<DevCpu>
     {
         template<ViewType TView>
-        using AccessorType = DeviceViewAccessor<TView>;
+        using AccessorType = HostViewAccessor<TView>;
     };
 
     template<typename TDev, ViewType TView>
--- a/include/alpaka/mem/view/ViewAccessOps.hpp
+++ b/include/alpaka/mem/view/ViewAccessOps.hpp
@@ -4,7 +4,6 @@
 
 #pragma once
 
-#include "alpaka/dev/DevCpu.hpp"
 #include "alpaka/extent/Traits.hpp"
 #include "alpaka/mem/view/Traits.hpp"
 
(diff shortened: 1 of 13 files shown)


fwyzard:
Instead of adding a new template parameter for the `Device`, can you extract it from the buffer, _e.g._ `alpaka::Dev<Buffer>` ?

include/alpaka/mem/view/ViewConst.hpp
```diff
@@ -18,7 +18,7 @@ namespace alpaka
     //! A non-modifiable wrapper around a view. This view acts as the wrapped view, but the underlying data is only
     //! exposed const-qualified.
     template<typename TView>
-    struct ViewConst : internal::ViewAccessOps<ViewConst<TView>>
+    struct ViewConst : internal::ViewAccessorType<decltype(alpaka::getDev(std::declval<TView>())), ViewConst<TView>>
```
fwyzard:
```suggestion
    struct ViewConst : internal::ViewAccessorType<alpaka::Dev<TView>, ViewConst<TView>>
```

fwyzard:
Ah, buffers for the CPU SYCL device (`using DevCpuSycl = DevGenericSycl<TagCpuSycl>;`) should also be made accessible on the host.

sbaldu:
> Ah, buffers for the CPU SYCL device (`using DevCpuSycl = DevGenericSycl<TagCpuSycl>;`) should also be made accessible on the host.

Ah you're right. It should be sufficient to just add a new specialization.

include/alpaka/mem/view/ViewAccessOps.hpp
```diff
@@ -14,28 +13,59 @@
 #include <type_traits>
 #include <utility>
 
+namespace alpaka
+{
+
+    class DevCpu;
+    class DevCpuSycl;
```
fwyzard: `DevCpuSycl` is a type alias, not a class:
```c++
using DevCpuSycl = DevGenericSycl<TagCpuSycl>;
```

What's the best way to forward-declare it ?

fwyzard: Does it work to `#include "alpaka/dev/DevCpuSycl.hpp"` instead of doing a forward declaration ?

include/alpaka/mem/view/ViewConst.hpp
```diff
@@ -18,7 +18,8 @@ namespace alpaka
     //! A non-modifiable wrapper around a view. This view acts as the wrapped view, but the underlying data is only
     //! exposed const-qualified.
     template<typename TView>
-    struct ViewConst : internal::ViewAccessOps<ViewConst<TView>>
+    // struct ViewConst : internal::ViewAccessorType<decltype(alpaka::getDev(std::declval<TView>())), ViewConst<TView>>
```
fwyzard: Please remove the commented-out line ?

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -275,12 +275,16 @@ static auto testBufferAccessorAdaptor(
     auto const base = reinterpret_cast<uintptr_t>(std::data(buf));
     auto const expected = base + static_cast<uintptr_t>((pitch * index).sum());
     INFO("element " << index << " expected at offset " << expected - base);
+#if not defined(ALPAKA_ACC_GPU_CUDA_ENABLED) and not defined(ALPAKA_ACC_GPU_HIP_ENABLED)                              \
+    and not defined(ALPAKA_SYCL_ONEAPI_GPU) and not defined(ALPAKA_SYCL_ONEAPI_FPGA)
```
fwyzard: This will skip the test for all backends, including the CPU ones, if any of the GPU backends is enabled.

Instead of using a preprocessor check, could you use an `if constexpr` based on the tag of the accelerator, or the type of the platform ?

sbaldu: Thanks, I am now checking the platform

include/alpaka/mem/view/ViewConst.hpp
sbaldu: Thanks

include/alpaka/mem/view/ViewAccessOps.hpp
sbaldu: I am now forward declaring `DevGenericSycl`, instead of `DevCpuSycl`. I don't think including the header would work because there is an `#ifdef` that defines it only if the SYCL flags are passed

fwyzard: The `#ifdef`s are there on purpose :-)

test/unit/mem/buf/src/BufTest.cpp
fwyzard: I think checking the Tag should be more robust: if the corresponding backend is not enabled, the Platform may not be defined at all - while the Tags should always be defined.

include/alpaka/mem/view/ViewAccessOps.hpp
```diff
@@ -14,28 +15,56 @@
...
+        using value_type = Elem<TView>;
+        using pointer = value_type*;
+        using const_pointer = value_type const*;
+        using reference = value_type&;
+        using const_reference = value_type const&;
+        using Idx = alpaka::Idx<TView>;
+        using Dim = alpaka::Dim<TView>;
 
+    public:
+        ALPAKA_FN_HOST auto data() -> pointer
```
fwyzard: Can you mark this one as `[[nodiscard]]` too ?

fwyzard: ...and the non-`const` one for `HostViewAccessor` too, I don't know why it wasn't marked before 🤷🏻

test/unit/mem/buf/src/BufTest.cpp
```diff
@@ -15,6 +15,14 @@
 #include <numeric>
 #include <type_traits>
 
+namespace alpaka
+{
+
+    template<concepts::Tag TTag>
+    struct PlatformGenericSycl;
+
+} // namespace alpaka
```
fwyzard: Why is this needed ?

fwyzard: (here and for the other tests)

include/alpaka/mem/view/ViewAccessOps.hpp
sbaldu: Done

test/unit/mem/buf/src/BufTest.cpp
sbaldu: Thanks, now it's not needed because I am using tags for SYCL. Removed it.

fwyzard:
Thanks @sbaldu looks good.
I'll wait for the tests to pass and merge it 👍🏻

fwyzard (approved):

PR #2564, opened 2025-09-11
Do not crash if a SYCL platform is invalid or empty

### LINKED ISSUE #2554: SYCL platforms throw an execption if there are no devices
Instantiating a SYCL platform for which there are no devices fails with an exception like
```
terminate called after throwing an instance of 'sycl::_V1::exception'
  what():  No device of requested type available. Please check https://software.intel.com/content/www/us/en/develop/articles/intel-oneapi-dpcpp-system-requirements.html
Aborted (core dumped)
```

alpaka should catch the exception and return an empty platform.

fwyzard:


fwyzard:
Looks like `clang-format` does not know how to handle function try blocks :-(

@chillenzer @SimeonEhrig do you have any suggestions for the formatting ?

chillenzer:
You're not the first to stumble upon this:
- https://github.com/llvm/llvm-project/issues/58987
- https://github.com/llvm/llvm-project/issues/58183
- https://github.com/llvm/llvm-project/issues/20989

It's also fixed: https://github.com/llvm/llvm-project/pull/95878

But we're on an ancient version of `clang-format` and I'm not allowed to update...

fwyzard:
I see, thanks.

Indeed `clang-format-22` handles it correctly 🤷🏻

fwyzard:
OK, my code was wrong anyway, so I can use an approach that clang-format can handle...

psychocoderHPC (approved):

PR #2565, opened 2025-09-12
Implement `alpaka::allocAsyncBuf()` for SYCL devices

fwyzard:
The behaviour is emulated using regular SYCL device allocations and a delayed deallocation, as SYCL does not provide asynchronous, queue-ordered memory allocation/deallocation primitives.

The allocation uses `sycl::malloc_device()` and is always immediate and blocking.

The deallocation depends on the properties of the queue:
  * if the queue is blocking, all operations submitted when the buffer goes out of scope should already have completed; free the memory immediately, and wait for the free operation to complete;
  * if the queue is non-blocking, enqueue the call to `sycl::free()` in a host_task.

### DIFF
--- a/include/alpaka/mem/buf/sycl/traits/BufGenericSyclTraits.hpp
+++ b/include/alpaka/mem/buf/sycl/traits/BufGenericSyclTraits.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2025 Anton Reinhard
+/* Copyright 2025 Jan Stephan, Luca Ferragina, Aurora Perego, Andrea Bocci, Anton Reinhard
  * SPDX-License-Identifier: MPL-2.0
  */
 
@@ -6,6 +6,7 @@
 
 #include "alpaka/mem/buf/Traits.hpp"
 #include "alpaka/mem/buf/sycl/BufGenericSycl.hpp"
+#include "alpaka/queue/sycl/QueueGenericSyclBase.hpp"
 
 #ifdef ALPAKA_ACC_SYCL_ENABLED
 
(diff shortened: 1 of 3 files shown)


include/alpaka/mem/buf/sycl/traits/BufGenericSyclTraits.hpp
```diff
@@ -195,10 +196,78 @@ namespace alpaka::trait
...
+            TElem* memPtr = sycl::malloc_device<TElem>(static_cast<std::size_t>(getExtentProduct(extent)), q);
+            auto deleter = [q](TElem* ptr) mutable
+            {
+                if constexpr(TBlocking)
+                {
+                    // If the queue is blocking, all operations submitted when the buffer goes out of scope should
+                    // always have completed. Free the memory immediately, and wait for the free operation to complete.
+                    // From SYCL 4.8.3.6: Note: Whether free is blocking or non-blocking is unspecified. Applications
+                    // should not rely on free for synchronization, nor assume that free cannot cause deadlocks.
+                    sycl::free(ptr, q);
```
psychocoderHPC, line 255: Is it not possible to add a `q.wait();` before the free call to ensure that all buffer operations are finished, that the user is not required to provide the buffer in an unused state?

fwyzard: If the queue is blocking, each operation should always complete before the host executes the next operation.
So by the time the destructor of the buffer is called, any operations submitted to the same queue should have already completed, and an explicit `q.wait()` should not be necessary.

psychocoderHPC: But what is if you use the buffer within multiple threads with a blocking queue. In this case it could still be that another thread submitted an algorithm into the queue which is using a view to the buffer. With the wait call before we can guarantee that no work (also submitted by other threads) is in the queue and the buffers pointer can safely freed.

fwyzard: Mhm, I would argue that's a race condition that the user has to avoid ?

---

Let me check if I understood correctly, the scenario you have in mind is something like this ?

```c++
QueueGpuSyclIntelNonBlocking queue(device);
{
  auto buffer = alpaka::allocAsyncBuf<float>(queue);
  std::async(std::launch::async, do_something, queue, buffer);
  // buffer goes out of scope, but do_something(queue, buffer) may have enqueued some operations that are not yet ready
}
```

Then even if `buffer` goes out of the local scope, the copy of `buffer` in `do_something` keeps a reference, and the buffer would not be destroyed until `do_something` completes.
If all the operations on the buffer are done via `queue` they are all be enqueued before `buffer` can go out of scope and the destructor is called, and so they should all complete before the call to `sycl::free`.

---

If the scenario is different, for example if `do_something()` _does not_ hold a reference to `buffer`:
```c++
QueueGpuSyclIntelNonBlocking queue(device);
{
  auto buffer = alpaka::allocAsyncBuf<float>(queue);
  std::async(std::launch::async, do_something, queue, buffer.data());
}
```
or it doesn't use `queue` to submit other operations:
```c++
QueueGpuSyclIntelNonBlocking queue(device);
{
  auto buffer = alpaka::allocAsyncBuf<float>(queue);
  std::async(std::launch::async, do_something, buffer);
}
```
then there may be a race condition between `buffer` being destroyed and the operations on it having completed.

But I think that calling `q.wait()` before `sycl::free` wouldn't add any guarantee: the other thread(s) may still enqueue a new operation in-between the call to `q.wait()` and `sycl::free` (or in the first case, even after the call to `sycl::free`).

fwyzard: ---
Anyway for a blocking queue calling `q.wait()` should basically be free... so if you come up with a possible scenario where it would help, adding it should be cheap.

psychocoderHPC: You are right, it was a wrong workflow in my mind. In case another thread is enqueing something into a blocking queue the thread is anyway waiting until the operation is finished and the destructor of the buffer will not be called.

So all fine with this part of the code.

psychocoderHPC (approved):

PR #2566, opened 2025-09-14
Extend `alpaka::allocAsyncBuf()` to N-dimensional buffers on CUDA/HIP devices

fwyzard:
CUDA and HIP do not provide an asynchronous equivalent of `cuda`/`hipMallocPitch` or `cuda`/`hipMalloc3D`, and do not document the how the pitch is determined by those functions.

However, on all tested platforms, the alignment used for the pitch matches the value of the `textureAlignment` device attribute: 512 bytes on NVIDIA GPUs, 256 bytes on AMD GPUs.

The asynchronous allocation of N-dimensional buffers can be implemented in alpaka using `cuda`/`hipMallocAsync` and handling the pitch based on the `textureAlignment` value (which is currently hard-coded).

### DIFF
--- a/include/alpaka/mem/buf/uniformCudaHip/traits/BufUniformCudaHipRtTraits.hpp
+++ b/include/alpaka/mem/buf/uniformCudaHip/traits/BufUniformCudaHipRtTraits.hpp
@@ -7,6 +7,7 @@
 #include "alpaka/mem/buf/Traits.hpp"
 #include "alpaka/mem/buf/cpu/BufCpu.hpp"
 #include "alpaka/mem/buf/uniformCudaHip/BufUniformCudaHipRt.hpp"
+#include "alpaka/mem/view/Traits.hpp"
 
 #if defined(ALPAKA_ACC_GPU_CUDA_ENABLED) || defined(ALPAKA_ACC_GPU_HIP_ENABLED)
 
@@ -256,7 +257,7 @@ namespace alpaka::trait
 
     //! The CUDA/HIP stream-ordered memory allocation capability trait specialization.
     template<typename TApi, typename TDim>
-    struct HasAsyncBufSupport<TDim, DevUniformCudaHipRt<TApi>> : std::bool_constant<TDim::value <= 1>
+    struct HasAsyncBufSupport<TDim, DevUniformCudaHipRt<TApi>> : std::true_type
     {
     };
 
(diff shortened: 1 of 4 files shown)


fwyzard:
@psychocoderHPC I've imeplemented N-dimensional allocations using the same alignment that CUDA and HIP use natively in `cuda`/`hipMallocPitch`.

If we pefer we could instead decide to use a smaller alignment, for example 128 bytes, or make it dependent on the device warp size, _etc_.

psychocoderHPC:
> @psychocoderHPC I've imeplemented N-dimensional allocations using the same alignment that CUDA and HIP use natively in `cuda`/`hipMallocPitch`.
> 
> If we pefer we could instead decide to use a smaller alignment, for example 128 bytes, or make it dependent on the device warp size, _etc_.

I have done the same in alpaka3 because I need the per line alignment at compile time. For NVIDIA and AMD GPUs I use currently hard coded 128-byte because I was not 100% sure if the alignment is guaranteed to be larger.

https://github.com/alpaka-group/alpaka3/blob/66b93eaa35055cf84c5e79db6d4abf17d4680eb8/include/alpaka/api/unifiedCudaHip/Queue.hpp#L632-L638

128byte equals 32 therads (one warp) times 4 byte what should always be guaranteed, else we would not be able to tune for fast coalesced memory access.

fwyzard:
We also use 128 bytes in the CMS software for the alignment of all GPU memory buffers.

Let me know if you prefer that I can change this to be 128 bytes everywhere, irrespective of the CUDA and ROCm texture alignment.

psychocoderHPC:
Maybe we should change it to 128 byte. In any cases, I know assuming 128byte will not limit the performance.

Note, in alpaka3 I wrote a test case which validates the alignment, the alignment is carried with the MdSpan object, therefore it is easy to validate it for the CPU too. In mainline alpaka this could be more complicated because, depending on the supported SIMD length, the alignment differs.

https://github.com/alpaka-group/alpaka3/blob/66b93eaa35055cf84c5e79db6d4abf17d4680eb8/tests/unit/mem/alloc.cpp#L157-L220

fwyzard:
> In mainline alpaka this could be more complicated because, depending on the supported SIMD length, the alignment differs.

In the CMS software we pass `-DALPAKA_DEFAULT_HOST_MEMORY_ALIGNMENT=128` to the compiler, in order to always have an alignment of (at least) 128 bytes for all alpaka memory allocations.

include/alpaka/mem/buf/uniformCudaHip/traits/BufUniformCudaHipRtTraits.hpp
```diff
@@ -256,43 +257,74 @@ namespace alpaka::trait
...
+                std::size_t const width = static_cast<std::size_t>(extent.back()) * sizeof(TElem);
+                // On all tested NVIDIA and AMD GPUs the alignment used for pitched allocations is the same value
+                // reported by the textureAlignment device property (512 bytes on NVIDA GPUs, 256 bytes on AMD GPUs).
+                // This was tested on: NVIDIA Tesla T4, A100, L40S, H100, and RTX 3050 Ti Laptop GPUs,
+                // and on AMD Radeon Pro WX 9100, Radeon Pro W7800/W7900, Instinct MI250X, and Instinct MI300X.
+                // However, it is expected that an alignment of 128 bytes (32 threads per warp times 4 bytes per float
+                // or int) should be sufficient to achieve coalesced memory accesses, and would reduce the amount of
+                // wasted memory.
+                constexpr std::size_t alignment = 128;
+                pitch = (width + alignment - 1) / alignment * alignment;
```
psychocoderHPC, line 294: should be 

```C++
pitch = core::divCeil(width,alignment)*alignment;
```

SimeonEhrig:
@fwyzard The code is implicit tested by PR #2544 ? I only found existing tests, where the  `alpaka::allocAsyncBuf()` is tested with 1D.

include/alpaka/mem/view/Traits.hpp
```diff
@@ -44,6 +45,31 @@ namespace alpaka
...
+            constexpr auto dim = TIdx{TDim::value};
+            if constexpr(dim > 0)
+                pitchBytes.back() = static_cast<TIdx>(sizeof(TElem));
+            if constexpr(dim > 1)
+            {
+                if(pitch == 0)
+                    pitchBytes[TDim::value - 2] = extent.back() * pitchBytes.back();
+                else
+                    pitchBytes[TDim::value - 2] = static_cast<TIdx>(
+                        (static_cast<std::size_t>(extent.back() * pitchBytes.back()) + pitch - 1) / pitch * pitch);
```
psychocoderHPC, line 65: should be

```C++
static_cast<TIdx>(core::divCeil(static_cast<std::size_t>(extent.back() * pitchBytes.back()),pitch) * pitch);
```

psychocoderHPC (changes requested):

include/alpaka/mem/buf/uniformCudaHip/traits/BufUniformCudaHipRtTraits.hpp
fwyzard: Could be, but the current version is also just fine.

Maybe if there was a "core" function to to the rounding to the next multiple, without the need to multiply again by the divisor.

But I'm not going to add one in this PR...

fwyzard:
> @fwyzard The code is implicit tested by PR #2544 ? I only found existing tests, where the `alpaka::allocAsyncBuf()` is tested with 1D.

Doesn't `test/unit/mem/buf/src/BufTest.cpp`
```c++
TEMPLATE_LIST_TEST_CASE("memBufAsyncBasicTest", "[memBuf]", alpaka::test::TestAccs)
{   
    using Acc = TestType;
    using Dim = alpaka::Dim<Acc>;
    using Idx = alpaka::Idx<Acc>;
    
    if constexpr(alpaka::hasAsyncBufSupport<alpaka::Dev<Acc>, Dim>)
    {
        testAsyncBufferMutable<Acc>(alpaka::test::extentBuf<Dim, Idx>);
    }
    else
    {
        INFO("Stream-ordered memory buffers are not supported in this configuration.");
    }
}
```
run the test for any dimensions ?

SimeonEhrig:
> > @fwyzard The code is implicit tested by PR #2544 ? I only found existing tests, where the `alpaka::allocAsyncBuf()` is tested with 1D.
> 
> Doesn't `test/unit/mem/buf/src/BufTest.cpp`
> 
> ```c++
> TEMPLATE_LIST_TEST_CASE("memBufAsyncBasicTest", "[memBuf]", alpaka::test::TestAccs)
> {   
>     using Acc = TestType;
>     using Dim = alpaka::Dim<Acc>;
>     using Idx = alpaka::Idx<Acc>;
>     
>     if constexpr(alpaka::hasAsyncBufSupport<alpaka::Dev<Acc>, Dim>)
>     {
>         testAsyncBufferMutable<Acc>(alpaka::test::extentBuf<Dim, Idx>);
>     }
>     else
>     {
>         INFO("Stream-ordered memory buffers are not supported in this configuration.");
>     }
> }
> ```
> 
> run the test for any dimensions ?

Yes the test runs multi dimensional. I didn't found it. Thanks for clarification.

psychocoderHPC (approved):

PR #2570, opened 2025-09-29
Implement compile-time warp size, when available

fwyzard:
Still to do:
  - [x] add tests
  - [ ] ~~define the warp size at compile time for SYCL kernels, when specialised~~

### DIFF
--- a/include/alpaka/warp/Traits.hpp
+++ b/include/alpaka/warp/Traits.hpp
@@ -23,6 +23,14 @@ namespace alpaka::warp
         template<typename TWarp, typename TSfinae = void>
         struct GetSize;
 
+        //! The compile-time warp size trait.
+        template<typename TWarp, typename TSfinae = void>
+        struct GetSizeCompileTime;
+
+        //! The warp size upper-limit trait.
+        template<typename TWarp, typename TSfinae = void>
+        struct GetSizeUpperLimit;
+
         //! The all warp vote trait.
         template<typename TWarp, typename TSfinae = void>
         struct All;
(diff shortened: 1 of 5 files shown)


AuroraPerego:
can `https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/kernel/SyclSubgroupSize.hpp` be merged with it for the SYCL part?
SYCL also has the issue that the warp sizes available are more than one..

fwyzard:
> can `https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/kernel/SyclSubgroupSize.hpp` be merged with it for the SYCL part? 

No, exactly because this:

> SYCL also has the issue that the warp sizes available are more than one..

And also because SYCL could be JIT'ted, and then there cannot be a known warp size at compile time.

fwyzard:
The idea for SYCL is to reuse the `alpaka::trait::warpSize<TKernelFnObj, TAcc>` trait, for the kernels that require a well-defined warp size.

---

Is that even possible ?
We cannot query the kernel type in a generic function.

---

One way could be to encode the warp size (requested by the kernel at compile time) in the type of the accelerator.

This however is going to require substantial changes, and is better suited for a separate PR.

fwyzard:
I've added a second function to return the maximum possible warp size.
This should help support devices where the warp size is not defined at compile time (SYCL, mostly) for the cases where a compile-time constant is needed to allocate an array or some other storage space.

include/alpaka/warp/Traits.hpp
```diff
@@ -68,12 +76,37 @@ namespace alpaka::warp
         return trait::GetSize<ImplementationBase>::getSize(warp);
     }
 
+    //! If the warp size is available as a compile-time constant returns its value; otherwise returns 0.
+    //!
+    //! \tparam TWarp The warp implementation type.
+    //! \param warp The warp implementation.
+    ALPAKA_NO_HOST_ACC_WARNING
+    template<typename TWarp>
+    ALPAKA_FN_ACC constexpr auto getSizeCompileTime() -> std::int32_t
```
fwyzard, line 84: @psychocoderHPC I've tried some variations on
```c++
constexpr auto getSizeCompileTime(TWarp const& /*warp*/) -> std::int32_t
```
so the user code could call
```c++
alpaka::warp::getSizeUpperLimit(acc);
```
instead of
```c++
alpaka::warp::getSizeUpperLimit<TAcc>();
```
but all my attempts failed with errors saying that `acc` cannot be used in a constant expression.

Would you have any ideas how to make it work ?

psychocoderHPC: This is how I handle it in alpaka3 to solve the same problem. Within the implementation, you must call the typed version. This gives the user a interface without explicit types and solves the issue.


https://github.com/alpaka-group/alpaka3/blob/f43b3b427ad56958c2a2792f840672d052d1e3de/include/alpaka/interface.hpp#L22-L29

psychocoderHPC: this are the coresponding internal methods

https://github.com/alpaka-group/alpaka3/blob/f43b3b427ad56958c2a2792f840672d052d1e3de/include/alpaka/internal/interface.hpp#L50-L65

fwyzard: I've tried to use
```c++
template<typename TWarp>
ALPAKA_FN_ACC constexpr auto getSizeCompileTime(TWarp const& /*warp*/) -> std::int32_t
```
and call it via
```c++
constexpr std::int32_t constexprSize = alpaka::warp::getSizeCompileTime(acc);
```
but I get:

```
.../test/unit/warp/src/GetSize.cpp(31): error: expression must have a constant value
constexpr std::int32_t constexprSize = alpaka::warp::getSizeCompileTime(acc);
                                       ^
.../test/unit/warp/src/GetSize.cpp(31): note #2689-D: the value of parameter "acc" (declared at line 20) cannot be used as a constant
constexpr std::int32_t constexprSize = alpaka::warp::getSizeCompileTime(acc);
                                                                        ^
```

fwyzard:
@psychocoderHPC does this approach look reasonable ?
would you have suggestion for better naming ?

fwyzard:
The last commit adds:
  - using `ALPAKA_DEFAULT_AMD_WAVEFRONT_SIZE` for unknown AMD GPUs, or failing with an `#error`
  - mark the functions as `__device__`

fwyzard:
clang does not allow `__builtin_unreachable()` in a `constexpr` function.

psychocoderHPC (approved):

PR #2571, opened 2025-10-07
Do not include the null terminaton character in the demangled `string_view`

fwyzard:


fwyzard:
The `string_view` returned by `alpaka::demangled<T>` includes a trailing null character.
This breaks use cases where the `string_view` is embedded in a larger `string`.

fwyzard:
I can reproduce the failures with ROCm, and they seem due to https://github.com/ROCm/ROCm/issues/4751 .

If I increase the stack size from 8 MB to 20 MB (`ulimit -s 20000`) the tests complete correctly.

SimeonEhrig (approved):

PR #2574, opened 2025-10-17
Extend the CI to test CUDA 12.8 and 12.9

fwyzard:
Also update the older CUDA version to the latest point releases.

PR #2581, opened 2025-11-27
Add memory orderings and use them for fences

ikbuibui:
Adds memory ordering tags.
Defines a trait to get the default memory orders for fences for each backend.
Adds ability for the user to optionally specify a memory ordering for mem fences.


ikbuibui:
I dont define a `consume` memory ordering at all, since from what I know it is not well defined (broken) and deprecated in C++26 anyways and everyone just implements it as `acquire` to be safe and correct.
https://en.cppreference.com/w/cpp/atomic/memory_order.html#Release-Consume_ordering

ikbuibui:
The implementation for HIP comes from [here](https://github.com/ROCm/clr/blob/037b8cdd66f07692ce6ce7ec05a0fafce4728ac0/hipamd/include/hip/amd_detail/amd_device_functions.h#L680)

ikbuibui:
CUDA memory orders are introduced with this PR but these are available only from [CUDA 12.8](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#atomic-functions), so #2574 would be nice to test this

ikbuibui:
Inspected PTX and confirmed that this works for CUDA>=12.8.
With `alpaka::mem_fence(acc, alpaka::mem_order::acq_rel, alpaka::memory_scope::Block{});`. we get a `fence.acq_rel.cta;` instead of a `fence.sc.cta;` we have otherwise

include/alpaka/mem/order/MemoryOrder.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+#include <concepts>
+
+namespace alpaka
+{
+    namespace mem_order
+    {
+        /**
+         * The user requested memory order may be converted to a stronger memory order guarantee if the backend does
+         * not support the requested memory ordering
```
fwyzard, line 16: >     * The user requested memory order may be converted to a stronger memory order guarantee if the backend does
>     * not support the requested memory ordering

Though in practice this is never the case ?

include/alpaka/mem/order/MemoryOrderHip.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * SPDX-License-Identifier: MPL-2.0
+ */
+
+#pragma once
+
+#include "alpaka/mem/order/MemoryOrder.hpp"
+
+#include <concepts>
+
+#ifdef ALPAKA_ACC_GPU_HIP_ENABLED
```
fwyzard, line 11: ~~Looking at the HIP documentation, I think this may require ROCm 6.4.0 (based on clang 19).~~
~~ROCm 6.4.0 mentions `__builtin_amdgcn_fence` in [the docs](https://rocm.docs.amd.com/projects/llvm-project/en/docs-6.4.0/LLVM/clang/html/LanguageExtensions.html#builtin-amdgcn-fence), while ROCm 6.3.3 [does not](https://rocm.docs.amd.com/projects/llvm-project/en/docs-6.3.3/LLVM/clang/html/LanguageExtensions.html).~~

include/alpaka/mem/order/MemoryOrder.hpp
```diff
@@ -0,0 +1,55 @@
...
+
+        struct MemoryOrderTag
+        {
+        };
+
+        struct SeqCst : MemoryOrderTag
+        {
+        };
+
+        struct AcqRel : MemoryOrderTag
```
fwyzard, line 28: Just an idea, what if we spelled the full `AcquireRelease` and `SequentialConsistency` ?

include/alpaka/mem/order/MemoryOrderStl.hpp
fwyzard, line 1: How about `MemoryOrderStl.h` (_etc_) ?

include/alpaka/mem/order/MemoryOrderHip.hpp
fwyzard: However looking at the HIP code, it was available much earlier, looks like it was simply undocumented.

fwyzard:
I'm wondering if using an `enum` for the memory spaces wouldn't be simpler than using tags and a concept ?

include/alpaka/mem/order/MemoryOrder.hpp
fwyzard: Mhm, I see that it is done at the implementation level, not in the backend-specific tags ?

include/alpaka/mem/fence/MemFenceOmp2Order.hpp
```diff
@@ -0,0 +1,76 @@
+/* Copyright 2022 Jan Stephan, Bernhard Manfred Gruber, Andrea Bocci, Tapish Narwal
```
fwyzard: I think this file is all your doing :-) ?

fwyzard:
Now I'm wondering what happens if one tries to use an acquire-release or sequentially-consistent thread-level fence to order with a block-level fence ?

I don't know _why_ one would do it, but in principle they could.

However this could silently fail in some backends where either thread-level or block-level fences are "skipped" ?

include/alpaka/mem/fence/MemFenceUniformCudaHipBuiltIn.hpp
```diff
@@ -27,37 +31,184 @@ namespace alpaka
...
 #        endif
 
-    namespace trait
+
+    namespace detail
     {
-        template<>
-        struct MemFence<MemFenceUniformCudaHipBuiltIn, memory_scope::Block>
+        template<alpaka::MemoryOrder TMemOrder>
+        [[maybe_unused]] static constexpr __device__ void cuda_ptx_fence_device([[maybe_unused]] TMemOrder order)
```
fwyzard, line 41: why not use `cuda::atomic::atomic_thread_fence(...)` ?

fwyzard:
@ikbuibui overall this looks very good !

Two questions:
  - what do you think is missing (since it is in draft mode) ?
  - I didn't check myself; do we have a reasonable set of tests for the fence operations ?   if not, is it something you would be interested / willing to implement ?

include/alpaka/mem/order/MemoryOrder.hpp
ikbuibui: Yes, it is done at the implementation level as it depends on details of the backend.

include/alpaka/mem/order/MemoryOrder.hpp
ikbuibui: I'm open to it with no particular preference. I followed the STL naming since it was concise. If you prefer the full names let me know and I'll update them.

include/alpaka/mem/fence/MemFenceUniformCudaHipBuiltIn.hpp
ikbuibui: That would have been ideal, but I was trying to avoid requiring `libcu++`

fwyzard: I wouldn't insist on using it, but if it's available, does the job, and does not create licensing or installation problems (all details to be discussed) maybe we should consider making use of it ?

ikbuibui:
> I'm wondering if using an `enum` for the memory spaces wouldn't be simpler than using tags and a concept ?

Yes, that is also absolutely doable, and maybe simpler. I was comfortable doing it with tags so thats what I used :) 

>     * what do you think is missing (since it is in draft mode) ? 
>     * I didn't check myself; do we have a reasonable set of tests for the fence operations ?   if not, is it something you would be interested / willing to implement ?

A combined answer to both was that I still want to look into the tests and haven't done so. In any case I'm not sure if there will be any reasonable way to test if the memory orderings are working as intended other than inspecting the generated code. If I don't come up with anything, I'll mark the PR as ready in the first week of next year. 

> Now I'm wondering what happens if one tries to use an acquire-release or sequentially-consistent thread-level fence to order with a block-level fence ?
> 
> However this could silently fail in some backends where either thread-level or block-level fences are "skipped" ?

This is a good question. I'll think about this a bit, but in any case this problem existed before this PR as well.

include/alpaka/mem/order/MemoryOrderStl.hpp
ikbuibui: Renamed.

include/alpaka/mem/fence/MemFenceOmp2Order.hpp
ikbuibui: Whoops. Too much copy-pasting.

ikbuibui:
Marking as ready for review. We can discuss the potential optional inclusion of `licu++` and the way this can be approached in the developer meeting. 
We have a minimal test for our memory barrier which is fine, in principle it is possible to add some more tests for the barrier, and also to maybe have a test for ordering (i'm not 100% sure about this though as maybe this test only checks if sequential consistency is satisfied or not).
However, how extensive our test suite should be is an open question from my end. In principle we can get away with less testing since we are mostly just doing abstraction.

ikbuibui:
Actually I would feel better if we merge this after #2574, since I have a funny feeling about how I have used the `ALPAKA_ARCH_PTX` macro for guarding against the host compiler from looking into places where it shouldn't (espeically for clang cuda given my work with #2602). I will set this to draft again.

ikbuibui:
Sorry about the delay in the PR. I got sidetracked with other work. 
Changes are all in the last commit. 
- I added a way to conditionally include `<cuda/atomic>` and I define a macro if it is. 
- For a device fence in the PTX fallback I was using a system fence. I switched this to a gpu level fence. 
- I removed the use of the compiler builtin `__nv_atomic_thread_fence` as it was unneeded complexity (Using it required `CUDA > 12.8` and a check if the builtin is available for the compute capability used (via a macro defined by NVCC) . I preferred to keep the inline PTX fence as the fallback. This inbuilt can be looked into again in the future, when we drop SM versions older than 70 or require CUDA 12.8 as a minimum, when it becomes viable to use without a lot of fallbacks being required.  

I did some manual tests for some particular conditions as well. The fence works correctly with
- `<cuda/atomic>` if it is available
- `clang20` as CUDA compiler.

ikbuibui:
Ready for review.

SimeonEhrig (approved):

fwyzard:
@ikbuibui are there any tests or examples about how to use this ?

ikbuibui:
@fwyzard There are no tests since I found no easy way to verify correctness except inspecting the generated code. However your comment does make me think that we should nonetheless add some tests/examples using this feature to ensure that it is working (at least compiling correctly, even if robust correctness checks are slightly difficult).
I will discuss with @psychocoderHPC tomorrow and open a PR with a either a test or an example which showcases the usage.
For current reference, it is to be used like this
 `alpaka::mem_fence(acc, alpaka::mem_order::acquire, alpaka::memory_scope::Device{});`

fwyzard:
Thanks.

At least we should have some tests that the code compiles and runs :-)

I don't have a suggestion how to check that a barrier is actually working or not, but if you can think of something it would of course be very good to have it.

PR #1566, opened 2022-01-10
Avoid locking for atomic operation on simple types

fwyzard:
This pull request is an attempt to solve the issue #1452. It is an updated and rebased version of #1453.

For Trivially Copyable types, aligned with `boost::atomic_ref<T>::required_alignment`, I specialized atomic operations to avoid locking. In the case in which atomics are often used to increase a counter, this may be beneficial.

For all feedback, discussion, evaluation of the performance,_etc._, please see #1453.

PR #2582, opened 2025-11-27
fix `atomicDec()` for CPU

psychocoderHPC:
Atomic decrement implemented

```C++
((old >= value) ? 0 : static_cast<T>(old - 1))
```

instead of

```C++
(old == static_cast<T>(0) || old > value) ? value : static_cast<T>(old - static_cast<T>(1))
```

This bug was introduced with alpaka 0.9.0 in PR #1566, and it showed up in alpaka3 while porting test #1772, which we never merged into the mainline.

For reference, here is the link to the CUDA definition of `atomicDec` https://docs.nvidia.com/cuda/cuda-c-programming-guide/#atomicdec

### DIFF
--- a/include/alpaka/atomic/AtomicAtomicRef.hpp
+++ b/include/alpaka/atomic/AtomicAtomicRef.hpp
@@ -162,10 +162,10 @@ namespace alpaka
                 isSupportedByAtomicAtomicRef<T>();
                 alpaka::detail::atomic_ref<T> ref(*addr);
                 T old = ref;
-                T result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                T result = (old == static_cast<T>(0) || old > value) ? value : (old - static_cast<T>(1));
                 while(!ref.compare_exchange_weak(old, result))
                 {
-                    result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                    result = (old == static_cast<T>(0) || old > value) ? value : (old - static_cast<T>(1));
                 }
                 return old;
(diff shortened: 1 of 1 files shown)


ikbuibui (approved):

include/alpaka/atomic/AtomicAtomicRef.hpp
```diff
@@ -162,10 +162,11 @@ namespace alpaka
                 isSupportedByAtomicAtomicRef<T>();
                 alpaka::detail::atomic_ref<T> ref(*addr);
                 T old = ref;
-                T result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                T result = (old == static_cast<T>(0) || old > value) ? value : static_cast<T>(old - static_cast<T>(1));
                 while(!ref.compare_exchange_weak(old, result))
                 {
-                    result = ((old >= value) ? 0 : static_cast<T>(old - 1));
+                    result
+                        = (old == static_cast<T>(0) || old > value) ? value : static_cast<T>(old - static_cast<T>(1));
```
AuroraPerego: is the `static_cast<T>` around `old - static_cast<T>(1)` necessary? The type should already be `T`

psychocoderHPC: no it is one cast to much

psychocoderHPC: fixed

AuroraPerego (approved):

Issue #2553, opened 2025-07-27
Implement `alpaka::Dev<Tag>`


fwyzard:
To get the device type corresponding to a given `Tag` we have to do something like
```c++
using Device = alpaka::Dev<alpaka::TagToAcc<Tag, Dim1D, uint32_t>>;
```

It would be cleaner to implement `alpaka::Dev<Tag>`, so we could do simply
```c++
using Device = alpaka::Dev<Tag>;
```

sbaldu:
I think this can be closed after #2584.

PR #2584, opened 2025-11-28
Specialize `DevType` for accelerator Tags

sbaldu:
This PR closes issue #2553 buy specializing the `alpaka::Dev` struct for Tag types, thus making it less verbose to obtain the Device type from the accelerator Tag.

### DIFF
--- a/include/alpaka/acc/TagAccIsEnabled.hpp
+++ b/include/alpaka/acc/TagAccIsEnabled.hpp
@@ -34,4 +34,15 @@ namespace alpaka
     //! list of all tags where the related accelerator is enabled
     using EnabledAccTags = alpaka::meta::Filter<AccTags, alpaka::AccIsEnabled>;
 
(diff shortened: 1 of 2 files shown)


sbaldu:
I noticed that in `TagAccIsEnabled.hpp` the file header is missing. Should I add it? In which case, who should I add as authors of that file?

AuroraPerego:
> I noticed that in `TagAccIsEnabled.hpp` the file header is missing. Should I add it? In which case, who should I add as authors of that file?

Given the commit history, it's Simeon

sbaldu:
> > I noticed that in `TagAccIsEnabled.hpp` the file header is missing. Should I add it? In which case, who should I add as authors of that file?
> 
> Given the commit history, it's Simeon

Done, thanks

AuroraPerego (approved):

PR #2073, opened 2023-08-10
[SYCL] Adopt the experimental free function extension

fwyzard:
Access the current nd_item via a free function instead of storing it as a data member of the various SYCL accelerator base classes.

PR #2585, opened 2025-11-28
add warp mask functions

AuroraPerego:
Now that oneAPI 2025.3 is out and the bug related to the opportunistic group has been fixed, I've rebased the branch and updated it. Main changes:
- warp functions now support passing the mask
- `activemask` in HIP has now been implemented 
- I've defined the macro `HIP_ENABLE_WARP_SYNC_BUILTINS` directly in the code to enable sync functions in HIP starting from 6.2 to be independent from the way the code is compiled (`cmake`, `Makefile`.. )
- I've enabled the tests for SYCL

Note that using oneAPI < 2025.3 most likely will not work (but we won't notice since we are not running tests on the Intel GPU).
I don't know if it's ok to merge it or if it's best to move to 2025.3 first, but in the meantime we can discuss the implementation.

### DIFF
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -13,7 +13,7 @@
 #endif
 
 #if ALPAKA_LANG_HIP
-#    if HIP_VERSION >= 60200000 && HIP_VERSION < 70000000
+#    if HIP_VERSION >= 60'200'000 && HIP_VERSION < 70'000'000
 #        define HIP_ENABLE_WARP_SYNC_BUILTINS
 #    endif
 // HIP defines some keywords like __forceinline__ in header files.
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -13,6 +13,9 @@
 #endif
 
 #if ALPAKA_LANG_HIP
+#    if HIP_VERSION >= 60200000 && HIP_VERSION < 70000000
+#        define HIP_ENABLE_WARP_SYNC_BUILTINS
+#    endif
 // HIP defines some keywords like __forceinline__ in header files.
 #    include <hip/hip_runtime.h>
 #endif
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -13,7 +13,7 @@
 #endif
 
 #if ALPAKA_LANG_HIP
-#    if HIP_VERSION >= 60'200'000 && HIP_VERSION < 70'000'000
+#    if ALPAKA_COMP_HIP >= ALPAKA_VERSION_NUMBER(6, 2, 0) && ALPAKA_COMP_HIP < ALPAKA_VERSION_NUMBER(7, 0, 0)
 #        define HIP_ENABLE_WARP_SYNC_BUILTINS
 #    endif
 // HIP defines some keywords like __forceinline__ in header files.
--- a/include/alpaka/core/CudaHipCommon.hpp
+++ b/include/alpaka/core/CudaHipCommon.hpp
@@ -23,6 +23,9 @@
 #    endif
 
 #    ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+#        if HIP_VERSION >= 60200000 && HIP_VERSION < 70000000
+#            define HIP_ENABLE_WARP_SYNC_BUILTINS
+#        endif
 #        include <hip/hip_runtime.h>
 #    endif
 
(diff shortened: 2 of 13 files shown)


include/alpaka/core/Common.hpp
```diff
@@ -13,6 +13,9 @@
 #endif
 
 #if ALPAKA_LANG_HIP
+#    if HIP_VERSION >= 60'200'000 && HIP_VERSION < 70'000'000
```
psychocoderHPC: plese use `ALPAKA_COMP_HIP` https://github.com/alpaka-group/alpaka/blob/8648cc6e3c0366618566c7b5d25052c04d6d8732/include/alpaka/core/Config.hpp#L122C14-L130

should be
```C++
#if ALPAKA_COMP_HIP >= ALPAKA_VESRION(6,2) && ALPAKA_COMP_HIP < ALPAKA_VERSION(7,0)
```

include/alpaka/core/CudaHipCommon.hpp
```diff
@@ -23,6 +23,9 @@
 #    endif
 
 #    ifdef ALPAKA_ACC_GPU_HIP_ENABLED
+#        if HIP_VERSION >= 60'200'000 && HIP_VERSION < 70'000'000
```
psychocoderHPC: pleasde use `ALPAKA_COMP_HIP`

include/alpaka/test/event/EventHostManualTrigger.hpp
```diff
@@ -476,6 +476,9 @@ namespace alpaka::trait
 
 #ifdef ALPAKA_ACC_GPU_HIP_ENABLED
 
+#    if HIP_VERSION >= 60'200'000 && HIP_VERSION < 70'000'000
```
psychocoderHPC: `ALPAKA_COMP_HIP`

psychocoderHPC (changes requested):

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -120,9 +109,9 @@ namespace alpaka::warp::trait
         // FIXME This should be std::uint64_t on AMD GCN architectures and on CPU,
         // but the former is not targeted in alpaka and CPU case is not supported in SYCL yet.
         // Restrict to warpSize <= 32 for now.
-        static auto ballot(warp::WarpGenericSycl<TDim> const& warp, std::int32_t predicate) -> std::uint32_t
+        static auto ballot(warp::WarpGenericSycl<TDim> const& /*warp*/, std::int32_t predicate) -> std::uint32_t
         {
-            auto const sub_group = warp.m_item_warp.get_sub_group();
+            auto sub_group = sycl::ext::oneapi::this_work_item::get_sub_group();
```
fwyzard, line 114: ages ago I remember that `sycl::ext::oneapi::this_work_item` had worse performance than using the work item passed to the lambda.

Maybe it has improved, but can we make that change in a separate PR ?

Or here the difference is not relevant ?

AuroraPerego: You already did the [PR](https://github.com/alpaka-group/alpaka/pull/2073) 😃 
According to the comments there, the performance was similar, but we wanted to wait because the feature just came out. Indeed, now they are not in the `experimental` namespace anymore. Do you want to go ahead with your PR, and then I will rebase on it? Or do you prefer to do it in another way?

fwyzard: Eh, but that was two years ago!
If there is no impact on the performance, let's go ahead with _this_ PR as is, and we can revisit my old one later...

psychocoderHPC:
@AuroraPerego If this PR should go into the release 2.1 you should get the CI green until the latest 11.12.2025 that we have at least one day to prepare the release.

AuroraPerego:
> @AuroraPerego If this PR should go into the release 2.1 you should get the CI green until the latest 11.12.2025 that we have at least one day to prepare the release.

I'm not sure I'll be able to fix it for Thursday, do not consider it for the release :( 
For oneAPI, either we don't add this feature now, or I put the `#if`/`#else` for the different versions, because the code in this PR does not work with versions < 2025.3 
For HIP, I have to understand why it does not compile with versions <6.2

psychocoderHPC:
> > @AuroraPerego If this PR should go into the release 2.1 you should get the CI green until the latest 11.12.2025 that we have at least one day to prepare the release.
> 
> I'm not sure I'll be able to fix it for Thursday, do not consider it for the release :( For oneAPI, either we don't add this feature now, or I put the `#if`/`#else` for the different versions, because the code in this PR does not work with versions < 2025.3 For HIP, I have to understand why it does not compile with versions <6.2

I am fine with moving this to the next release. This should keep the current code more stable.

psychocoderHPC (approved):

PR #2595, opened 2025-12-16
Add runtime check of extents for memory operations

### LINKED ISSUE #2491: The destination view and the extent are required to have compatible index types!
Trying to use `alpaka::memcpy` from an `std::vector<float>` to an `ViewPlainPtr<DevUniformCudaHipRt<ApiCudaRt>, float, std::integral_constant<long unsigned int, 1>, unsigned int>`

fails with

> `The destination view and the extent are required to have compatible index types!`

because the `vector` uses `unsigned long` as index type, while the device view uses `unsigned int`.

Rather than a compile time failure, if the destination index type is not large enough to hold all possible values of the source index type, can we defer to a runtime check that the size actually fits in the destination index type ?

sbaldu:
This PR addresses issue #2491 and removes the `static_assert`s that imposed compile-time checks on the index types of the views and extents in memory operations, substituting them with runtime checks on the actual extents.

### DIFF
--- a/include/alpaka/mem/view/Traits.hpp
+++ b/include/alpaka/mem/view/Traits.hpp
@@ -1,5 +1,5 @@
 /* Copyright 2024 Axel Hübl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan, Bernhard Manfred Gruber,
- *                Aurora Perego
+ *                Aurora Perego, Simone Balducci
  * SPDX-License-Identifier: MPL-2.0
  */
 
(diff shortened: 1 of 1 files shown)


include/alpaka/mem/view/Traits.hpp
```diff
@@ -240,9 +240,8 @@ namespace alpaka
         static_assert(
             Dim<TView>::value == Dim<TExtent>::value,
             "The view and the extent are required to have the same dimensionality!");
-        static_assert(
-            meta::IsIntegralSuperset<Idx<TView>, Idx<TExtent>>::value,
-            "The view and the extent must have compatible index types!");
+
+        assert((extent <= getExtents(view)).all() && "The memset extent must not be larger than the view extents!");
```
AuroraPerego:
```suggestion
        assert((extent <= getExtents(view)).all() && "The memset extent must not be larger than the view extent!");
```
and also for the others below

sbaldu: done

AuroraPerego (approved):

SimeonEhrig (approved):

PR #2602, opened 2025-12-19
Memory ordered atomics

### LINKED ISSUE #2123: Support atomics with memory orders
AFAIU the atomics that alpaka exhibt a relaxed memory order. That is, performing an atomic operation in alpaka does neither synchronize with other atomic operatons, nor give any guarantee on the visibility of preceding/subsequent ordinary loads and stores. We effectively assume a weakly-ordered memory model. Any synchchronization of loads and stores requires the use of explicit memory fences in addition to relaxed atomic operations. This is not how C++ does atomics since C++11.

In C++11 (and C11, Rust, etc.) atomic operations can be configured on their effect on the visible order of memory operations. This configuraiton is the [memory order](https://en.cppreference.com/w/cpp/atomic/memory_order). CUDA's libc++ started to offer such configurable atomics a while back, see for example the definition of [atomic_ref](https://nvidia.github.io/libcudacxx/extended_api/synchronization_primitives/atomic_ref.html). I have not found a similar feature in HIP yet. SYCL has those: https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_memory_ordering

For the efficient implementation of certain synchronization primitives, e.g. locks, or certain lock-free data structures, e.g. lock-free queues, atomics with various memory orders are a necessity. To that end, I wonder when and how those should be supported in alpaka. I am mostly leaning towards a class type similar to `std::atomic_ref`, but I am open to suggestions.

### LINKED ISSUE #2172: atomic_ref based atomics are too strong
The CPU atomic implementation using `std::atomic_ref` use a sequentially consistent memory ordering, which is a stronger guarantee than their CUDA counterparts, which are weakly ordered and always require explicit fences. Therefore, the CPU atomics should also be weakened to a relaxed memory order, potentially improving performance on CPUs.

ikbuibui:
Add support for memory ordered atomics. Closes #2123 
Sets the old behaviour of relaxed atomics as the default. 
Also makes atomic ref ordering relaxed by default. Closes #2172 

We emulate ordered atomics using fences when ordered atomics available natively.
We fall back to stronger guarantees if a backend does not support the requested order.
For CUDA in particular we do this according to [this resource](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/index.html#atomics-application-binary-interface).

- [x] Needs #2581 to be merged before this, as it depends on the fence implementation to emulate atomic orderings.
- [ ] Need to check and re-enable atomic min and max for HIP

### DIFF
--- a/include/alpaka/atomic/AtomicAtomicRef.hpp
+++ b/include/alpaka/atomic/AtomicAtomicRef.hpp
@@ -10,6 +10,8 @@
 
 #    include "alpaka/atomic/Traits.hpp"
 #    include "alpaka/core/Config.hpp"
+#    include "alpaka/mem/order/MemOrderStl.hpp"
+#    include "alpaka/mem/order/MemoryOrder.hpp"
 
 #    include <array>
 #    include <atomic>
@@ -17,6 +19,8 @@
 
 #    ifndef ALPAKA_DISABLE_ATOMIC_ATOMICREF
 #        ifndef ALPAKA_HAS_STD_ATOMIC_REF
+#            include "alpaka/mem/order/MemOrderBoost.hpp"
+
 #            include <boost/atomic.hpp>
 #        endif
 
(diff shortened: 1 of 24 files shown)


SimeonEhrig:
It was implemented in alpaka 3.0 and will be not back ported.

ikbuibui:
*It will be. Not done in alpaka3 yet

SimeonEhrig:
> *It will be. Not done in alpaka3 yet

Sorry, was a misunderstanding.

PR #2605, opened 2026-01-16
Update the reference `mdspan` library to 2025.12.11

fwyzard:
The latest tagged version, 0.6.0, has issues building with `hipcc` and the `clang -x cu`  GPU compilers. As of 2023.06.09 these issues were resolved.

More recent versions reworked the namespace used for `mdspan`, so we can directly define it in the `alpaka::experimental` namespace, without touching the `std` namespace.

The version as of 2025.12.11 passes all tests, with minimal adjustments for SYCL:
  - we need to not expose the macro implementation of `printf`;
  - we need a workaround for the `[[no_unique_address]]` attribute that breaks the FPGA compiler.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -909,9 +909,10 @@ if (alpaka_USE_MDSPAN STREQUAL "SYSTEM")
 elseif (alpaka_USE_MDSPAN STREQUAL "FETCH")
     include(FetchContent)
     FetchContent_Declare(
+        # kokkos/mdspan as of 2025.12.11
         mdspan
         GIT_REPOSITORY https://github.com/kokkos/mdspan.git
-        GIT_TAG 973ef6415a6396e5f0a55cb4c99afd1d1d541681
+        GIT_TAG bcfcc9ea8fc5390b99261a4d8450c3f2fc18a7f2
     )
     # we don't use FetchContent_MakeAvailable(mdspan) since it would also install mdspan
     # see also: https://stackoverflow.com/questions/65527126/how-to-disable-installation-a-fetchcontent-dependency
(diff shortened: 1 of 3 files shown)


fwyzard:
About the FPGA issue: the minimal test that reproduces the error is:
```c++
TEMPLATE_LIST_TEST_CASE("submdspan", "[memView]", alpaka::test::TestAccs)
{
    using TElem = int;

    using Acc = TestType;
    using Dev = alpaka::Dev<Acc>;
    using Dim = alpaka::Dim<Acc>;
    using Idx = alpaka::Idx<Acc>;

    auto const platform = alpaka::Platform<Dev>{};
    auto const dev = alpaka::getDevByIdx(platform, 0);
    auto queue = alpaka::Queue<Dev, alpaka::Blocking>(dev);

    auto const extent = alpaka::Vec{2, 3}; // 2 rows, 3 cols
    auto buf = alpaka::allocBuf<TElem, int>(dev, extent);
    alpaka::test::iotaFillView(queue, buf);

    auto const mds = alpaka::experimental::getMdSpan(buf);
    alpaka::test::KernelExecutionFixture<Acc> fixture(Vec::ones());
    CHECK(fixture(
        [=] ALPAKA_FN_ACC(Acc const&, bool* success) noexcept
        {
            auto elem_1_2 = alpaka::experimental::submdspan(mds, 1, 2);
            ALPAKA_CHECK(*success, elem_1_2() == 5);
        }));
}
```

However, the equivalent minimal test written with native SYCL does not seem to have any problems:
```c++
int main() {
    using TElem = int;

    constexpr std::size_t rows = 2;
    constexpr std::size_t cols = 3;
    constexpr std::size_t size = rows * cols;

    // SYCL queue on the FPGA emulator
    sycl::queue q(sycl::ext::intel::fpga_emulator_selector_v);

    // USM shared memory allocation
    std::shared_ptr<TElem> buf(sycl::malloc_shared<TElem>(size, q), [q](void* ptr){sycl::free(ptr, q);});
    std::iota(buf.get(), buf.get() + size, 0);

    // wrap the memory in an mdspan
    using extents_t = std::extents<std::size_t, rows, cols>;
    using mdspan_t = std::mdspan<TElem, extents_t, std::layout_right>;
    mdspan_t mds(buf.get());

    std::shared_ptr<bool> success(sycl::malloc_shared<bool>(1, q), [q](void* ptr){sycl::free(ptr, q);});
    *success = true;

    q.single_task([=, s = success.get()]() {
        auto elem_1_2 = std::submdspan(mds, 1, 2);
        if (elem_1_2() != 5) { *s = false; }
    }).wait();

    assert(*success);

    return 0;
}
```

psychocoderHPC (approved):

PR #2606, opened 2026-01-20
update Readme supported compiler matrix

psychocoderHPC:
Since at least #2574 the compiler matrix is outdated.

I had a few problems seeing which version of clang-cuda and nvcc-clang are tested or supported.
Somehow, we need to take the global job filter rules into account + the filter rules within alpaka.
I checked now the jobs generated in alpaka in the CI and updated the readme matrix generator based on what we are running with an eye on the filter rules too.
This means maybe there are more versions supported, but it is hard to keep all filter rules in mind to update the JSON file.


fwyzard:
@psychocoderHPC I trust you on the update of what is being tested :-)

fwyzard (approved):

PR #2573, opened 2025-10-14
Extend the CI to test clang 20, ROCm 6.3, 6.4 and 7.0

fwyzard:

PR #2609, opened 2026-01-26
CMake mark supported HIP versions 6.0 - 7.0

psychocoderHPC:
With https://github.com/alpaka-group/alpaka/pull/2573 but also before where we switched to C++20 we forgot to update in CMake the supported HIP versions.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -549,8 +549,8 @@ if(alpaka_ACC_GPU_HIP_ENABLE)
         enable_language(HIP)
         find_package(hip REQUIRED)
 
-        set(_alpaka_HIP_MIN_VER 5.1)
-        set(_alpaka_HIP_MAX_VER 6.2)
+        set(_alpaka_HIP_MIN_VER 6.0)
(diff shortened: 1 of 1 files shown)


fwyzard:
I don't know much about CMake... does
```
set(_alpaka_HIP_MAX_VER 7.0)
```
mean that 7.0 is supported, or not ?

fwyzard (approved):

psychocoderHPC:
> I don't know much about CMake... does
> 
> ```
> set(_alpaka_HIP_MAX_VER 7.0)
> ```
> 
> mean that 7.0 is supported, or not ?

The test is written by us

```
if(${_hip_MAJOR_MINOR_VERSION} VERSION_LESS ${_alpaka_HIP_MIN_VER} OR ${_hip_MAJOR_MINOR_VERSION} VERSION_GREATER ${_alpaka_HIP_MAX_VER})
```

The range is [min;max], so 7.0 is included. The test is not a hard enforcement; it is only notifying the user that we have not tested the newer version, but the user is still free to use newer versions.

PR #2610, opened 2026-01-27
Do not use `std::min` in kernels

fwyzard:
`std::min` is not availble in CUDA code unless the flag `--expt-relaxed-constexpr` is used.

### DIFF
--- a/include/alpaka/exec/UniformElements.hpp
+++ b/include/alpaka/exec/UniformElements.hpp
@@ -5,15 +5,21 @@
 #include "alpaka/exec/ElementIndex.hpp"
 #include "alpaka/idx/Accessors.hpp"
 
-#include <algorithm>
+#include <concepts>
 #include <cstddef>
-#include <type_traits>
 
 namespace alpaka
 {
 
     namespace detail
     {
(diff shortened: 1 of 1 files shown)


psychocoderHPC:
Is the flag `--expt-relaxed-constexpr` not a required flag when using CUDA? IMO we are using many constexpr std library features within the kernel.

see: https://github.com/alpaka-group/alpaka/blob/db66388aa4ebd6ab8127a69b9a53c0f140b669cb/cmake/alpakaCommon.cmake#L454-L455

fwyzard:
If it's a hard requirement, can we add somewhere a check that `__CUDACC_RELAXED_CONSTEXPR__` is defined ?

psychocoderHPC:
> If it's a hard requirement, can we add somewhere a check that `__CUDACC_RELAXED_CONSTEXPR__` is defined ?

This sounds like a good idea. I suggest to add the static assert guarded with `ALPAKA_COMP_NVCC` to the file https://github.com/alpaka-group/alpaka/blob/develop/include/alpaka/core/ApiCudaRt.hpp

PR #2616, opened 2026-03-11
Feature view creation from queue

sbaldu:
This PR introduces the possibility of creating a View by passing a Queue.
The interface of the `createView` overloads is now templated on a `detail::DeviceProvider` concept, representing an object from which it's possible to obtain the Device (i.e. a Device itself or a Queue).

### DIFF
--- a/include/alpaka/mem/view/Traits.hpp
+++ b/include/alpaka/mem/view/Traits.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2024 Axel Hübl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan, Bernhard Manfred Gruber,
+/* Copyright 2026 Axel Hübl, Benjamin Worpitz, Matthias Werner, Andrea Bocci, Jan Stephan, Bernhard Manfred Gruber,
  *                Aurora Perego, Simone Balducci
  * SPDX-License-Identifier: MPL-2.0
  */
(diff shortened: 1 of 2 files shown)


include/alpaka/mem/view/Traits.hpp
```diff
@@ -491,6 +491,23 @@ namespace alpaka
                 os << rowSuffix;
             }
         };
+
+        template<typename T>
+        concept DeviceProvider = alpaka::isDevice<T> || alpaka::isQueue<T>;
```
SimeonEhrig:
```suggestion
        namespace concepts {
          template<typename T>
          concept DeviceProvider = alpaka::isDevice<T> || alpaka::isQueue<T>;
        }
```

SimeonEhrig:
Looks good. Only put the concept in the `alpaka::concepts` namespace.

include/alpaka/mem/view/Traits.hpp
sbaldu: Done

SimeonEhrig (approved):

include/alpaka/mem/view/Traits.hpp
```diff
@@ -491,6 +497,20 @@ namespace alpaka
                 os << rowSuffix;
             }
         };
+
```
fwyzard, line 500: does `alpaga::getDev` work with a device as an argument?
if it does, or if that can be added, the concept can be simplified to "anything that can be passed to getDev".

fwyzard: and this function could simply be `getDev`.

Issue #2615, opened 2026-02-23
Introdue a type `alpaka::warp::mask_t`


fwyzard:
The type of the value returned by warp operations like `ballot()` and `activemask()` is different for different back-ends.

It would help to write generic code if alpaka would provide an alias for this type, for example
```c++
namespace alpaka::warp {
  template <typename TWarp>
  using mask_t = ...;
}
```

fwyzard:
Note that HIP uses `uint64_t` (actually `unsigned long long`) independently of the GPU wavefront size:
```c++
unsigned long long __ballot(int predicate);
unsigned long long __activemask();
```

fwyzard:
Note that oneAPI uses a `struct sub_group_mask` with an implementation-defined size, independent from the backend.
The current code internally uses an `uint64_t`.

sbaldu:
I think this was closed by #2617?

PR #2617, opened 2026-03-13
Implement wrapper type for warp masks

sbaldu:
This PR addresses issue #2615 from @fwyzard, implementing a wrapper type `alpaka::warp::Mask` around the type returned by `activeMask` and `ballot`, providing the same interface for all the backends.

### DIFF
--- a/include/alpaka/warp/WarpGenericSycl.hpp
+++ b/include/alpaka/warp/WarpGenericSycl.hpp
@@ -1,4 +1,4 @@
-/* Copyright 2023 Jan Stephan, Luca Ferragina, Andrea Bocci, Aurora Perego
+/* Copyright 2026 Jan Stephan, Luca Ferragina, Andrea Bocci, Aurora Perego, Simone Balducci
  * SPDX-License-Identifier: MPL-2.0
  *
  * The implementations of Shfl::shfl(), ShflUp::shfl_up(), ShflDown::shfl_down() and ShflXor::shfl_xor() are derived
@@ -26,6 +26,8 @@ namespace alpaka::warp
     class WarpGenericSycl : public interface::Implements<alpaka::warp::ConceptWarp, WarpGenericSycl<TDim>>
     {
     public:
+        using mask_type = std::uint32_t;
+
         WarpGenericSycl(sycl::nd_item<TDim::value> my_item) : m_item_warp{my_item}
         {
         }
@@ -36,6 +38,7 @@ namespace alpaka::warp
 
 namespace alpaka::warp::trait
 {
+
     template<typename TDim>
     struct GetSize<warp::WarpGenericSycl<TDim>>
     {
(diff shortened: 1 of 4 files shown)


include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -159,13 +166,7 @@ namespace alpaka::warp
...
             static __device__ auto ballot(
                 [[maybe_unused]] warp::WarpUniformCudaHipBuiltIn const& warp,
-                std::int32_t predicate)
-            // return type is required by the compiler
-#        if defined(ALPAKA_ACC_GPU_CUDA_ENABLED)
-                -> std::uint32_t
-#        else
-                -> std::uint64_t
-#        endif
+                std::int32_t predicate) -> MaskType<WarpUniformCudaHipBuiltIn>::type
```
fwyzard: can this be

```suggestion
                std::int32_t predicate) -> Mask<WarpUniformCudaHipBuiltIn>
```
?

include/alpaka/warp/WarpUniformCudaHipBuiltIn.hpp
```diff
@@ -31,6 +32,16 @@ namespace alpaka::warp
 
     namespace trait
     {
+        template<>
```
fwyzard: I am mildly concerned that this may introduce an ODR violation.
Do you think the warp type could be templated on TApi, and that could be used to distinguish the warp size ?

fwyzard:
Can you make the mask type a nested type of the warp itself ?

Being able to write `Warp::mask_t` is simpler than `alpaka::warp::Mask<Warp>`.

fwyzard:
Also, could you update the tests and example to make use of this type ?

include/alpaka/warp/WarpGenericSycl.hpp
```diff
@@ -36,6 +36,13 @@ namespace alpaka::warp
 
 namespace alpaka::warp::trait
 {
+
+    template<typename TDim>
+    struct MaskType<WarpGenericSycl<TDim>>
+    {
+        using type = std::uint32_t;
```
psychocoderHPC: The mask type depends on the architecture you are targeting. If you target AMD it is 64bit because the warp/wave size is 64. The reason why it is 64 is that the OneApi plugin is using ROCm as connector to AMD hardware and in ROCm the mask size is devied as 64bit data type, IMO even if the warp size is 32.

psychocoderHPC: In alpaka3 we depend on the `warp` size but in alpaka mainline the warp size is not compile-time.
Maybe we can specialize the type on the accelerator type.

psychocoderHPC (dismissed):

include/alpaka/warp/WarpGenericSycl.hpp
fwyzard: Yes, oneAPI uses `uint64_t` as the underlying type for the mask:
```c++
struct sub_group_mask {
  friend class sycl::detail::Builder;
  using BitsType = uint64_t;
...
```

As you suggest we could use different mask types for different oneAPI back-ends:
  - 32 bit for NVIDIA GPUs
  - 64 bit for AMD GPUs
  - 32 bit for Intel GPUs (?)
  - 64 bit for CPUs (though last time I checked a subgroup size of 64 was broken, but it's been a while)
  - no idea for Intel/Altera FPGAs, probably 32 bit

or just 64 bits for all oneAPI back-ends if it's simpler and does not have any significant impact on the performance.

psychocoderHPC: IMO the version which uses different sizes per device type makes sense. Using always 64bit means one register overhead on systems with 32 as warp size.

SimeonEhrig:
@sbaldu There is a `typename` missing in the code: https://gitlab.com/hzdr/crp/alpaka/-/jobs/13914947238

fwyzard (approved):

fwyzard (dismissed):

SimeonEhrig (approved):

PR #2620, opened 2026-03-17
fix broken example convolution 2D

psychocoderHPC:
Remove kernel with shared memory.
  - the keernel was creating shared memory with the size of the global mesh
  - shared memory was not fully initilized if more than one thread block was used
  - The global 2D convolution kernel was not used and therefore dead code

Switch to the global convolution kernel.


psychocoderHPC:
Will be discussed in the developer meeting.

SimeonEhrig (approved):

PR #2626, opened 2026-06-13
Add support for CUDA 13.3

fwyzard:


SimeonEhrig:
This is a local bug, which I also found in the bashi PR: https://gitlab.com/hzdr/crp/alpaka/-/jobs/14840776512
It forget to add the SM 8.0 for the A100.

I can check, if we can easily overtake the fix from the bashi port: https://github.com/alpaka-group/alpaka/pull/2560/changes/4660c53483388859b0da2de08fe2d7e9e839af0a

script/install_cuda.sh
```diff
@@ -44,6 +44,9 @@ else
         elif [[ "$(cat /etc/os-release)" == *"24.04"* ]]
         then
             ALPAKA_CUDA_DISTRO=ubuntu2404
+        elif [[ "$(cat /etc/os-release)" == *"26.04"* ]]
+        then
+            ALPAKA_CUDA_DISTRO=ubuntu2604
```
SimeonEhrig, line 49: Why did you add Ubuntu 26.04? At the moment our base images uses Ubuntu 22.04 and 24.04.

fwyzard: No real reason, just for completeness since I was editing that file

SimeonEhrig: Okay. That was one of my two assumptions. The other was, that I need to provide a new base image soon ;-)

fwyzard:
replaced by #2627

PR #2630, opened 2026-07-31
fix error when passing `std::function` to `enqueue` with SYCL

m-fila:
Changing the SYCL queue's `enqueue` to use perfect forwarding instead of accepting arguments by const ref.

With oneAPI 2025, passing a host callback in form of `std::function` to `enqueue` method gives a compilation error. This is because oneAPI internally accepts the argument by `std::function&&`:

- Passing a non-`std::function` callable by const ref works because it is converted to a temporary `std::function`.
- Passing a `const std::function&` does not work, since it can't be converted to rvalue `std::function` and gives following error

```
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/handler.hpp:1878:5: error: no matching member function for call to 'SetHostTask'
 1878 |     SetHostTask(std::move(Func));
      |     ^~~~~~~~~~~
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/handler.hpp:2100:5: note: in instantiation of function template specialization 'sycl::handler::host_task_impl<const std::function<void ()> &>' requested here
 2100 |     host_task_impl(Func);
      |     ^
/__w/acts/acts/build/_deps/alpaka-src/include/alpaka/queue/sycl/QueueGenericSyclBase.hpp:153:41: note: in instantiation of function template specialization 'sycl::handler::host_task<const std::function<void ()> &>' requested here
  153 |                                     cgh.host_task(task);
      |                                         ^
/__w/acts/acts/build/_deps/alpaka-src/include/alpaka/queue/sycl/QueueGenericSyclBase.hpp:248:47: note: in instantiation of function template specialization 'alpaka::detail::QueueGenericSyclImpl::enqueue<false, std::function<void ()>>' requested here
  248 |                 queue.m_spQueueImpl->template enqueue<TBlocking>(task);
      |                                               ^
/__w/acts/acts/build/_deps/alpaka-src/include/alpaka/queue/Traits.hpp:49:54: note: in instantiation of member function 'alpaka::trait::Enqueue<alpaka::detail::QueueGenericSyclBase<alpaka::TagGpuSyclIntel, false>, std::function<void ()>>::enqueue' requested here
   49 |         trait::Enqueue<TQueue, std::decay_t<TTask>>::enqueue(queue, std::forward<TTask>(task));
      |                                                      ^
/__w/acts/acts/Traccc/device/alpaka/src/utils/queue.cpp:60:13: note: in instantiation of function template specialization 'alpaka::enqueue<alpaka::detail::QueueGenericSyclBase<alpaka::TagGpuSyclIntel, false>, std::function<void ()>>' requested here
   60 |   ::alpaka::enqueue(*m_impl->m_queue, std::move(func));
      |             ^
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/handler.hpp:1863:8: note: candidate function not viable: 1st argument ('typename std::remove_reference<const function<void ()> &>::type' (aka 'const std::function<void ()>')) would lose const qualifier
 1863 |   void SetHostTask(std::function<void()> &&Func);
      |        ^           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/opt/intel/oneapi/compiler/2025.0/bin/compiler/../../include/sycl/handler.hpp:1864:8: note: candidate function not viable: no known conversion from 'function<void ()>' to 'function<void (interop_handle)>' for 1st argument
 1864 |   void SetHostTask(std::function<void(interop_handle)> &&Func);
      |        ^           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~

```

As I propose, perfect forwarding preserves the task's value categor, allowing `std::function` rvalues to reach `host_task` without being converted to const ref and avoiding the compilation error.

This change makes the SYCL backend's enqueue implementation differ from the other accelerator backends, which still accept tasks by const ref. Those backends are unaffected because they internally don't require forwarding it as an rvalue.

### DIFF
--- a/include/alpaka/queue/sycl/QueueGenericSyclBase.hpp
+++ b/include/alpaka/queue/sycl/QueueGenericSyclBase.hpp
@@ -127,30 +127,31 @@ namespace alpaka
             }
 
             template<bool TBlocking, typename TTask>
-            auto enqueue(TTask const& task) -> void
+            auto enqueue(TTask&& task) -> void
             {
                 {
+                    using TaskType = std::decay_t<TTask>;
                     std::lock_guard<std::shared_mutex> lock{m_mutex};
 
                     clean_dependencies();
 
                     // Execute task
-                    if constexpr(is_sycl_task<TTask> && !is_sycl_kernel<TTask>) // Copy / Fill
+                    if constexpr(is_sycl_task<TaskType> && !is_sycl_kernel<TaskType>) // Copy / Fill
                     {
                         m_last_event = task(m_queue, m_dependencies); // Will call queue.{copy, fill} internally
                     }
                     else
                     {
                         m_last_event = m_queue.submit(
-                            [this, &task](sycl::handler& cgh)
+                            [this, captured_task = std::forward<TTask>(task)](sycl::handler& cgh) mutable
                             {
                                 if(!m_dependencies.empty())
                                     cgh.depends_on(m_dependencies);
 
-                                if constexpr(is_sycl_kernel<TTask>) // Kernel
-                                    task(cgh); // Will call cgh.parallel_for internally
+                                if constexpr(is_sycl_kernel<TaskType>) // Kernel
+                                    captured_task(cgh); // Will call cgh.parallel_for internally
                                 else // Host
-                                    cgh.host_task(task);
+                                    cgh.host_task(std::move(captured_task));
                             });
                     }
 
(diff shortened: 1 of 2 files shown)


SimeonEhrig:
Actual we try to keep the interface consistent between the backends. That's still true, until somebody uses `std::function`. I think for this case it is fine. Please do the following 2 taks:
1. Document in doc string of the SYCL implementation of `enqueue`, that this is a backend specific optimization and using a const ref is the portable way.
2. Rebease the PR. I checked the jobs. It looks like every failed job bases on a runner problem.

SimeonEhrig (approved):

PR #2645, opened 2026-08-28
update references to correct values and add DOIs

PrometheusPi:
This PR:
 - adds DOIs for the two papers
 - fixes some errors and adds some extra information to these papers

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ consider citing us accordingly in your derived work and publications:
 % colocated with the
 %   "2017 ISC High Performance Conference"
 %   in Frankfurt, Germany
-@inproceedings{MathesP3MA2017,
+@inproceedings{MatthesP3MA2017,
   author    = {{Matthes}, A. and {Widera}, R. and {Zenker}, E. and {Worpitz}, B. and
                {Huebl}, A. and {Bussmann}, M.},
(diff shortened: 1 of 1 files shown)


PrometheusPi:
@psychocoderHPC The CI seems to have an error - could you restart? (Should not be related to this PR.)

README.md
```diff
@@ -164,10 +164,14 @@ consider citing us accordingly in your derived work and publications:
   archivePrefix = "arXiv",
   eprint    = {1706.10086},
   keywords  = {Computer Science - Distributed, Parallel, and Cluster Computing},
-  day       = {30},
-  month     = {Jun},
+  day       = {20},
```
psychocoderHPC, line 167: We need to clarify if we should remove the arXiv link and which date is the correct.
For the first paper 30.Jun was the date of arXiv publication.

psychocoderHPC (changes requested):
We need to clarify if we should remove the arXiv link and which date is the correct.
For the first paper 30.Jun was the date of arXiv publication.

README.md
PrometheusPi: If we state it is peer-reviewed, the date of the journal publication needs to be stated. **But** we also should keep the arXiv link, to provide an option to access a pre-version without paywall.

psychocoderHPC:
This pr does not change any code, therefore I will merge it without the CI.

PR #2646, opened 2026-08-31
fix unknown `activeMask` with HIP 6.3.0

psychocoderHPC:
The warp mask feature was introduced with HIP 6.3.1 see https://rocm.docs.amd.com/en/docs-6.3.1/about/release-notes.html#hip-6-3-1

The bug was introduced with #2585 where we check for 6.3 instead of 6.3.1.

```
warp/WarpUniformCudaHipBuiltIn.hpp:120:24: error: use of undeclared identifier '__activemask'; did you mean 'Activemask'?
  120 |                 return __activemask();
      |                        ^
```

Found with picongpu where we still use HIP 6.3.0 https://gitlab.com/hzdr/crp/picongpu/-/jobs/16151713214

### DIFF
--- a/include/alpaka/core/Common.hpp
+++ b/include/alpaka/core/Common.hpp
@@ -13,7 +13,7 @@
 #endif
 
 #if ALPAKA_LANG_HIP
-#    if ALPAKA_COMP_HIP >= ALPAKA_VERSION_NUMBER(6, 2, 0) && ALPAKA_COMP_HIP < ALPAKA_VERSION_NUMBER(7, 0, 0)
+#    if ALPAKA_COMP_HIP >= ALPAKA_VERSION_NUMBER(6, 3, 1) && ALPAKA_COMP_HIP < ALPAKA_VERSION_NUMBER(7, 0, 0)
 #        define HIP_ENABLE_WARP_SYNC_BUILTINS
 #    endif
 // HIP defines some keywords like __forceinline__ in header files.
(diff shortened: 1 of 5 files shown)


fwyzard:
@psychocoderHPC `__activemask()` & friends are available from ROCm 6.2.0, but they are disabled by default and can be enabled with `#define HIP_ENABLE_WARP_SYNC_BUILTINS`.

They are enabled by default from ROCm 7.0.0, unless one defines `HIP_DISABLE_WARP_SYNC_BUILTINS`.

psychocoderHPC:
> @psychocoderHPC `__activemask()` & friends are available from ROCm 6.2.0, but they are disabled by default and can be enabled with `#define HIP_ENABLE_WARP_SYNC_BUILTINS`.
> 
> They are enabled by default from ROCm 7.0.0, unless one defines `HIP_DISABLE_WARP_SYNC_BUILTINS`.

We are setting `HIP_ENABLE_WARP_SYNC_BUILTINS` in alpaka https://github.com/psychocoderHPC/alpaka/blob/caf8a6daa36a1e2df670e35cf3844a9269ff0e56/include/alpaka/core/Common.hpp#L17, but testing with 6.3.0 shows that this is not working.
In alpaka we test 6.3.2 but PIConGPU is using 6.3.0.

psychocoderHPC:
You are right, it is in the headers of 6.2.0 release https://github.com/ROCm/clr/blob/dd7f9576620569fc8847cef767efcc1f4bab9d1a/hipamd/include/hip/amd_detail/amd_warp_sync_functions.h#L32
I will check again why it is not working in the test

psychocoderHPC:
Ok I found the issue, the problem was that we had one place in PIConGPU that included the hip headers directly and there we have not set the define required to activate the mask support for HIP.

PR #2648, opened 2026-09-02
fix special job

### LINKED ISSUE #2647: extended lambda flag is wrong set
In the special test `linux_special_nvcc12.5_gcc13_release_extended_lambda_off_compile_only` where `alpaka_CUDA_EXPT_EXTENDED_LAMBDA` should be off it is enabled.

It is `export alpaka_CUDA_EXPT_EXTENDED_LAMBDA=ON` [@see](https://gitlab.com/hzdr/crp/alpaka/-/jobs/16231762301/viewer#L55)

psychocoderHPC:
fix #2647

The name of the job should be `cuda only` and not `extended lambda off`, because the test where the lambdas are disabled, we have a special template.

The first commit removes `script/job_generator` from pre-commit because the rules of black, mypy-linter and pylint-linter collide with the alpaka coding styles we agreed on.

Note: The first commit is also in #2649 but the merge order is independent.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -36,10 +36,10 @@ repos:
   rev: v0.2.1
   hooks:
     - id: ruff
-      types_or: [ python, pyi, jupyter ]
-      # The ignores in here are chosen to conform with the currently
-      # existing code and not motivated any other way.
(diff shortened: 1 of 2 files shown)


SimeonEhrig (approved):
Yes, that is a copy-paste error.

PR #2649, opened 2026-09-02
Extend pre-commit to mach CI tests

psychocoderHPC:
- remove script/job_generator from pre-commit ruff
- add black python style test to pre-commit
- run pre-commit on the full codebase


psychocoderHPC:
@SimeonEhrig rebased

SimeonEhrig (approved):

SimeonEhrig (changes requested):
Please remove the black formatter job. The job is redundant. And by the way, I have no idea why the job didn't covered this files.

https://github.com/alpaka-group/alpaka/blob/856778c770439eca5f8ee7f008406aee4d74836c/.github/workflows/job-generator.yml#L4-L13

m-fila:
Does it make sense to have both ruff-format and black in the pre-commit? Is there a reason no to run pre-commit in the CI so the checks are the same?

SimeonEhrig:
> Does it make sense to have both ruff-format and black in the pre-commit?

No, they produce different results. I know it, because I use black since years and this summer I moved to the Zed editor, where ruff-format is the default editor. Therefor we keep black until there is a good reason to move to ruff-format.

> Is there a reason no to run pre-commit in the CI so the checks are the same?

I'm not sure, what you mean with this sentence.

m-fila:
> No, they produce different results. I know it, because I use black since years and this summer I moved to the Zed editor, where ruff-format is the default editor. Therefor we keep black until there is a good reason to move to ruff-format.

I mean why have two instead of just going with one. In propsed setup the pre-commit runs both one after the other

> I'm not sure, what you mean with this sentence.

I mean, from what I understood you get different checks in the CI and from running `pre-commit` locally. It's not uncommon to run something like `pre-commit run --all-files --show-diff-on-failure` in the CI instead of more jobs trying to replicate what it does

SimeonEhrig:
> I mean why have two instead of just going with one. In propsed setup the pre-commit runs both one after the other

You are right. That's indeed wrong. We should remove the ruff-formatter.

> I mean, from what I understood you get different checks in the CI and from running pre-commit locally. It's not uncommon to run something like pre-commit run --all-files --show-diff-on-failure in the CI instead of more jobs trying to replicate what it does

Make sense.

psychocoderHPC:
I updated the PR, I pushed new commits on top but can remove the first if needed.

- black is removed from gihub ci and pre-commit -> ruff is used as before the PR
- add pre-commit test job to the github actions
- changed python max line length to 120 to fit the settings from pre-commit
- apply pre-commit to all files in the code

psychocoderHPC:
@SimeonEhrig do you knwo why mypy and pylint is failing. could it be that this tests was added after the job generator was in the code base. Than it is clear because only the changed files will be checked.
If so I do not know how to fix it, except by suppressing the errors.

SimeonEhrig:
> I updated the PR, I pushed new commits on top but can remove the first if needed.
> 
>     * black is removed from gihub ci and pre-commit -> ruff is used as before the PR
> 
>     * add pre-commit test job to the github actions
> 
>     * changed python max line length to 120 to fit the settings from pre-commit
> 
>     * apply pre-commit to all files in the code

@psychocoderHPC  That's a mistake. ruff-formatt was never used before. It was only in the pre-commit configuration, but we never executed. The python formatter is black. Please change it and I will have a look on mypy and pylint afterwards.

psychocoderHPC:
> > I updated the PR, I pushed new commits on top but can remove the first if needed.
> > ```
> > * black is removed from gihub ci and pre-commit -> ruff is used as before the PR
> > 
> > * add pre-commit test job to the github actions
> > 
> > * changed python max line length to 120 to fit the settings from pre-commit
> > 
> > * apply pre-commit to all files in the code
> > ```
> 
> @psychocoderHPC That's a mistake. ruff-formatt was never used before. It was only in the pre-commit configuration, but we never executed. The python formatter is black. Please change it and I will have a look on mypy and pylint afterwards.

OK:
-  I removed now ruff from pre commit and the CI.
- added blake to pre-commit
- run pre-commit on all files (all python files using now blake, not only the job generator)
- added CI job to execute pre-commit

script/job_generator/src/alpaka_bashi/ci_yaml/names.py
```diff
@@ -43,7 +43,9 @@ def get_job_suffix(combination: bashi.Combination) -> str:
                 if combination[software].version == ON_VER:
                     version_str += "_mdspan"
             else:
-                version_str += f"_{combination[software].name}{str(combination[software].version)}"
+                version_str += (
+                    f"_{combination[software].name}{str(combination[software].version)}"
+                )
```
SimeonEhrig: It looks like, the pre-commit ignores the configuration in the [pyproject.toml](https://github.com/alpaka-group/alpaka/blob/5bd939558b6d05098f38e1af92f8ef16d5018614/script/job_generator/pyproject.toml#L37). The original line has a length of 100 chars.

Maybe you should set line length in the pre-commit config.

psychocoderHPC: Yes it is not taking it into account, pre-commit is used now for the full python code in alpaka, thats what ruff did before. With the PR that introduced bashi you pulled a different style for parts of the alpaka phython script. I can ot follow anymore what we should do, I think it was overlooked that there was already an formater for python in alpaka so we should deside to use only one and use the same style everywhere.

Let us have a short offline discussion for this to avoid updating this PR again and again.

psychocoderHPC:
@SimeonEhrig I update the PR, it is using now blacke with 100 as line length in pre-commit

SimeonEhrig (approved):

PR #2650, opened 2026-09-03
workaround github clone issues

psychocoderHPC:
The CI has issues to fecth external dependencies due to Github limitations, this PR is applying a knwon workaround.

https://github.com/orgs/community/discussions/206581

This is only a temporary workaround until we switch to using an access key.


psychocoderHPC:
~~We have no chicken eg issue, our linter using different rules and therefore I can not apply this fix a separate PR.
I will add it additionally to #2649.~~

I need to disable pre-commit on my local system so it doesn't fix the python file I changed.

AuroraPerego (approved):

PR #2653, opened 2026-09-09
update mdspan

SimeonEhrig:
- solve a compiler bug for CUDA 13.2
- Mdspan uses cuda::std::numeric_limits now. The cuda backend include mdspan by default if enable that the class is defined and in the correct namespace.

### DIFF
--- a/cmake/alpakaCommon.cmake
+++ b/cmake/alpakaCommon.cmake
@@ -904,10 +904,10 @@ if (alpaka_USE_MDSPAN STREQUAL "SYSTEM")
 elseif (alpaka_USE_MDSPAN STREQUAL "FETCH")
     include(FetchContent)
     FetchContent_Declare(
-        # kokkos/mdspan as of 2025.12.11
+        # kokkos/mdspan as of 2026.08.19
         mdspan
         GIT_REPOSITORY https://github.com/kokkos/mdspan.git
-        GIT_TAG bcfcc9ea8fc5390b99261a4d8450c3f2fc18a7f2
+        GIT_TAG ef0adc1ca08b6eda520a7fd3fcfd8125662518e7
     )
     # we don't use FetchContent_MakeAvailable(mdspan) since it would also install mdspan
     # see also: https://stackoverflow.com/questions/65527126/how-to-disable-installation-a-fetchcontent-dependency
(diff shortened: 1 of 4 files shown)


fwyzard:
Do we have any checks in c++23 mode with a compiler that supports `std::mdspan` instead of using the Kokkos version?

fwyzard:
CUDA 13.4.1 was released yesterday, with support for GCC 16, which includes support for `std::mdspan` in the standard library.

cmake/alpakaCommon.cmake
```diff
@@ -907,7 +907,7 @@ elseif (alpaka_USE_MDSPAN STREQUAL "FETCH")
         # kokkos/mdspan as of 2025.12.11
```
fwyzard:
```suggestion
        # kokkos/mdspan as of 2026.08.19
```
is the date for the commit below

fwyzard:
Unfortunately CUDA 13.4.1 does not work with GCC 16 in c++23 mode (see ), so we cannot easily test if `std::mdspan` would work, and anyway we cannot use it.

So I would delay any concerns about the implementation of `std::mdspan` in gcc 16 in c++23 mode to a later time 🤷🏻‍♂️

fwyzard (dismissed):
Just squash my commit into the rest.

fwyzard:
Sorry, I'm not sure what I did... hopefully everything is back OK now.

fwyzard (approved):

SimeonEhrig:
@fwyzard You changes looks fine.

I only was a little bit struggling with the new header, because of the using namespace and the missing import guard. But if approved. Therefore I assume it is valid ;-)

And if I read the code correctly, we should support `std::mdspan` from the CUDA SDK, if it compiles.

fwyzard:
@SimeonEhrig do you mean this comment ?

> `// no pragma once, because the header defines alias namespaces`

I thought it was intended to be like that :-)

Do things break if you use `#pragma once` ?

SimeonEhrig:
> @SimeonEhrig do you mean this comment ?
> 
> > `// no pragma once, because the header defines alias namespaces`
> 
> I thought it was intended to be like that :-)
> 
> Do things break if you use `#pragma once` ?

It is intended because of the `using namespace` and the `pragma once` break it. I was simply not sure, if this is a good solution, because I never did this `using namespace` in a more complex context before.

PR #2658, opened 2026-09-10
Add concepts for `Device`, `Queue` and `Platform`

sbaldu:
This PR adds concepts for `Device`, `Queue` and `Platform`, marking the pre-existing traits as deprecated.
It also moves the `View` concept to the public interface.


fwyzard:
@sbaldu thanks for these.

fwyzard (approved):

PR #2654, opened 2026-09-09
Add support for ROCm 7.14

SimeonEhrig:
- extend job-generator to generate ROCm 7.14 jobs
- add method to install ROCm 7.14+
- fix typo in -Wno-lifetime-safety-intra-tu-suggestions

PR #2660, opened 2026-09-11
[do not merge] Test various updates for ROCm 7.14

fwyzard:
Includes
  - #2661 
  - #2662
  - #2654


fwyzard:
Dang.

> 2026-09-11T12:42:02.454770Z 01E update-alternatives: error: alternative path /opt/rocm/lib/llvm/bin/clang doesn't exist

but I have it on my computer after installing ROCm 7.14 locally:
```
$ ls -l /opt/rocm/lib/llvm/bin/clang
lrwxrwxrwx 1 root root 8 Aug 19 04:44 /opt/rocm/lib/llvm/bin/clang -> clang-23
```

fwyzard:
The `amdrocm-core7.14` package seems to be the one that creates the missing symlinks.

SimeonEhrig:
> The `amdrocm-core7.14` package seems to be the one that creates the missing symlinks.

Looks good. I will also implement it in alpaka 3. That was, what I was missing.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -27,6 +27,11 @@
 #            error If ALPAKA_ACC_GPU_HIP_ENABLED is set, the compiler has to support HIP!
 #        endif
 
+#        if ALPAKA_COMP_CLANG >= ALPAKA_VERSION_NUMBER(23, 0, 0)
+#            pragma clang diagnostic push
+#            pragma clang diagnostic ignored "-Wlifetime-safety-intra-tu-suggestions"
+#        endif
```
SimeonEhrig: This warning is enabled by the clang 23 compiler flag `-Weverything` and it is part of a set of warnings. 

https://clang.llvm.org/docs/LifetimeSafety.html

I suggest to disable it in general via compiler flag `-Wno-lifetime-safety-intra-tu-suggestions` because new code needs to be also active supporting this feature.

Lifetime safety looks interesting in general. It follows the Rust model which actual possible with C++. The C++ compiler simply does not blame you if you break the model. We should think about, if we want to support it in alpaka 3.1 or later.

script/install_hip.sh
```diff
@@ -22,22 +22,61 @@ else
...
-    source /etc/os-release
-    wget -q -O - https://repo.radeon.com/rocm/rocm.gpg.key | sudo apt-key add -
-    echo "deb https://repo.radeon.com/rocm/apt/${ALPAKA_CI_HIP_VERSION} ${VERSION_CODENAME} main" | sudo tee -a /etc/apt/sources.list.d/rocm.list
-    travis_retry apt-get -y --quiet update
 
-    ALPAKA_CI_ROCM_VERSION=$ALPAKA_CI_HIP_VERSION
-    # append .0 if no patch level is defined
-    if ! echo $ALPAKA_CI_ROCM_VERSION | grep -Eq '[[:digit:]]+\.[[:digit:]]+\.[[:digit:]]+'; then
-        ALPAKA_CI_ROCM_VERSION="${ALPAKA_CI_ROCM_VERSION}.0"
+    if [ "$(version "${ALPAKA_CI_HIP_VERSION}")" -le "$(version "7.2.4")" ]; then
```
SimeonEhrig:
```suggestion
    if [ "$(version "${ALPAKA_CI_HIP_VERSION}")" -lt "$(version "7.3.0")" ]; then
```

SimeonEhrig:
@fwyzard My suggestion to process with the ROCm 7.14 support. We take this PR, because it includes already my PR. Please apply this single suggestion which based on your suggest, rebase and we merge it. If not automatically done, I will close PR #2654

fwyzard:
@SimeonEhrig I was actually planning to update your PR with the changes :-)

SimeonEhrig:
> @SimeonEhrig I was actually planning to update your PR with the changes :-)

It is also possible. Do what is easier for you.

fwyzard:
I pushed the updates to #2654.

PR #2662, opened 2026-09-13
Prepare the codebase for building with clang 23

fwyzard:
* update built-ins: prefer `__scoped_atomic_load_n` over `__hip_atomic_load` (deprecated).
  * disable `-Wlifetime-safety-intra-tu-suggestions`, that triggers warnings about Lifetime Safety Analysis.


fwyzard:
By the way, [Lifetime Safety Analysis](https://clang.llvm.org/docs/LifetimeSafety.html) looks interesting - but supporting it will require a non trivial amount of work (#2663).

SimeonEhrig:
Like I wrote [here]() already, I would disable the warning via compiler flag and not in source code, because new code is also affected.

fwyzard:
Rebased to re-trigger the readthedocs tests.

fwyzard:
> Like I wrote here already, I would disable the warning via compiler flag and not in source code, because new code is also affected.

OK by me.

Do you mean something like
```diff
+    # Disable warnings about Lifetime Safety Analysis suggestions
+    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 23.0)
+        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-Wnolifetime-safety-intra-tu-suggestions>")
+    endif()
```
in `test/common/devCompileOptions.cmake` ?

Or something else or somewhere else ?

SimeonEhrig:
> > Like I wrote here already, I would disable the warning via compiler flag and not in source code, because new code is also affected.
> 
> OK by me.
> 
> Do you mean something like
> 
> ```diff
> +    # Disable warnings about Lifetime Safety Analysis suggestions
> +    if(CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND CMAKE_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 23.0)
> +        list(APPEND alpaka_DEV_COMPILE_OPTIONS "-Wnolifetime-safety-intra-tu-suggestions>")
> +    endif()
> ```
> 
> in `test/common/devCompileOptions.cmake` ?
> 
> Or something else or somewhere else ?

Yes, this looks good.

SimeonEhrig (approved):

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -77,7 +77,10 @@ namespace alpaka::trait
 // Emulating atomics with atomicCAS is mentioned in the programming guide too.
 // http://docs.nvidia.com/cuda/cuda-c-programming-guide/#atomic-functions
 #        if ALPAKA_LANG_HIP
-#            if __has_builtin(__hip_atomic_load)
+#            if __has_builtin(__scoped_atomic_load_n)
+                EmulatedType old{
+                    __scoped_atomic_load_n(addressAsIntegralType, __ATOMIC_RELAXED, __MEMORY_SCOPE_DEVICE)};
```
psychocoderHPC, line 82: In the old code where `__hip_atomic_load8)` was used it was not possible to use a scope and we always used the device scope in the new function we should map to the scope we need `device -> __MEMORY_SCOPE_DEVICE`, ...
This not not strongly required because we do not support the system scope for now in alpaka but if we add it we will miss changing this line for sure.

fwyzard: Do you mean that we should use one of
- `__MEMORY_SCOPE_SYSTEM` (system)
- `__MEMORY_SCOPE_DEVICE` (device)
- `__MEMORY_SCOPE_WRKGRP` (group == workgroup)
- `__MEMORY_SCOPE_CLUSTR` (?)
- `__MEMORY_SCOPE_WVFRNT` (warp == wavefront)
- `__MEMORY_SCOPE_SINGLE` (single thread)

depending on the value of `THierarchy` ?

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
```diff
@@ -77,7 +77,10 @@ namespace alpaka::trait
...
+#            elif __has_builtin(__hip_atomic_load)
                 EmulatedType old{__hip_atomic_load(addressAsIntegralType, __ATOMIC_RELAXED, __HIP_MEMORY_SCOPE_AGENT)};
```
fwyzard, line 84: This could also use one of
* `__HIP_MEMORY_SCOPE_SINGLETHREAD`
  Represents single-thread memory scope in HIP (value is 1).

* `__HIP_MEMORY_SCOPE_WAVEFRONT`
  Represents wavefront memory scope in HIP (value is 2).

* `__HIP_MEMORY_SCOPE_WORKGROUP`
  Represents workgroup memory scope in HIP (value is 3).

* `__HIP_MEMORY_SCOPE_CLUSTER`
  Represents cluster memory scope in HIP (value is 6).

* `__HIP_MEMORY_SCOPE_AGENT`
  Represents agent memory scope in HIP (value is 4).

* `__HIP_MEMORY_SCOPE_SYSTEM`
  Represents system-wide memory scope in HIP (value is 5).

I guess "agent" behaves the same as "device"?
And "cluster" is still unknown.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
psychocoderHPC: Yes, this is what I mean.

Not required for this release but we need to switch all atomics to built-in atomics anyway: 

https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-extensions.html#built-in-atomic-functions

I think @ikbuibui started with this in the past.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
psychocoderHPC: Yes, the "agent" should be equal as device atomic and "cluster" I have never heard of.

psychocoderHPC (approved):

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
fwyzard: Looks like in some AMD GPUs one can program up to 15 groups to cooperate in a way similar to NVIDIA cooperative groups, and that's called a cluster.

include/alpaka/atomic/AtomicUniformCudaHipBuiltIn.hpp
fwyzard: > Not required for this release but we need to switch all atomics to built-in atomics anyway

That means supporting only CUDA 12.8 and later?

ikbuibui: > > Not required for this release but we need to switch all atomics to built-in atomics anyway
> 
> That means supporting only CUDA 12.8 and later?

Basically yes. In #2602 I started to look into ways to do this for older CUDA with fences based on the [ptx interop guide](https://docs.nvidia.com/cuda/ptx-writers-guide-to-interoperability/atomic-abi.html), basically things like below

``` c++
mem_fence(acc, alpaka::mem_order::release, scope); // calls inline ptx for fences
auto ret = atomic_func(address, value); // actually a non atomic func like add
mem_fence(acc, alpaka::mem_order::acquire, scope);
```
but not knowing if this actually guarantees atomicity and also not knowing what compiler reordering might be allowed is where I got slightly stuck. In principle it might be doable, and I can continue looking into it if someone other than me is interested in the feature as well (else I will add this to alpaka3 and skip mainline alpaka).

PR #2665, opened 2026-09-14
Implement interface for external allocators

sbaldu:
This PR adds to the allocation functions the possibility of passing an externally defined allocator like a caching allocator.
Since there is not a universal convention on the interface of device caching allocators, users are expected to wrap their allocators in a class respecting the `concepts::CachingAllocator` concept.
For example:
```cpp
template <typename TCachingAllocator>
class AllocatorWrapper {
private:
  std::reference_wrapper<TCachingAllocator> m_allocator;
  typename TCachingAllocator::Queue m_queue;

public:
  AllocatorWrapper(TCachingAllocator& allocator, typename TCachingAllocator::Queue queue)
      : m_allocator(allocator), m_queue(std::move(queue)) {}

  void* allocate(std::size_t bytes, std::size_t /* align */) {
    return m_allocator.get().allocate(bytes, m_queue);
  }

  void deallocate(void* ptr) { m_allocator.get().free(ptr); }
};
```
The allocator wrapper can then be constructed and passed to the allocation functions:
```cpp
AllocatorWrapper<Alloc> wrapper(cachingAlloc, queue);
auto buf = alpaka::allocAsyncBuf<float, Idx>(queue, extent, wrapper);
```


fwyzard:
I think the names used reflect a possible intention, but not the interface being implemented:
  - the concept `CachingAllocator` does not require that the allocator is actually caching anything;
  - the structs `AsyncBufAllocCached` and `BufAllocCached` do not require that the allocator does any caching.

Any kind of allocator could fit into these concept and functions, not just a "caching" allocator.

fwyzard:
@sbaldu can you rename the concept and types avoiding to mention "Cache" or "Caching" ?

include/alpaka/mem/buf/cpu/traits/BufCpuTraits.hpp
```diff
@@ -165,12 +165,54 @@ namespace alpaka::trait
...
+
+            static_assert(
+                std::is_same_v<Dev<TQueue>, DevCpu>,
+                "The BufCpu buffer can only be used with a queue on a DevCpu device!");
+            auto const dev = getDev(queue);
+            std::size_t const bytes = static_cast<std::size_t>(getExtentProduct(extent)) * sizeof(TElem);
+            void* const memPtr = allocator.allocate(bytes, alignof(TElem));
+            auto deleter = [l_queue = std::move(queue), alloc = std::move(allocator)](TElem* ptr) mutable
+            {
+                alpaka::wait(l_queue);
```
fwyzard: Wait, this is not an asynchronous operation.
Why do you need to `alpaka::wait` here ?

sbaldu: Yes you're right, for the CPU backend it was not needed, I removed it and enqueued the deallocation to the queue instead.

fwyzard:
OK, but mine was a general comment: you never want to call `alpaka::wait()` in the deallocator, otherwise the deallocations are no longer asynchronous.

PR #2667, opened 2026-09-15
Pin pre-commit hooks with SHA

m-fila:
For hardening, pin the revisions of pre-commit hooks to commit SHA instead of tag since the tags aren't immutable.
The human readable comment uses format compatible with `pre-commit autoupdate --freeze`, `prek update --freeze`, dependabot updates for pre-commit so further updates should automatically update both SHA and comment.


psychocoderHPC (approved):

SimeonEhrig:
@m-fila Please rebase. We merged an fix for the CMake installation in PR #2668.

SimeonEhrig (approved):

PR #2669, opened 2026-09-15
Support `memcpy` with different index types

fwyzard:
Cast the source and destination buffers index types to a common type in order to compare the dimensions correctly.

Add a test for the new functionality.


include/alpaka/mem/view/Traits.hpp
```diff
@@ -337,11 +337,16 @@ namespace alpaka
             std::is_same_v<DstElem, std::remove_const_t<SrcElem>>,
             "The source and destination view must have the same element type!");
 
+        // The extent and the views may use different index types; compare them in a type that can
+        // represent all of the values involved, to avoid both compilation errors and truncation.
+        [[maybe_unused]] auto const extents = getExtents(extent);
+        using SrcCommon = std::common_type_t<Idx<TExtent>, Idx<TViewSrc>>;
```
fwyzard:
```suggestion
        using SrcCommon [[maybe_unused]] = std::common_type_t<Idx<TExtent>, Idx<TViewSrc>>;
```

include/alpaka/mem/view/Traits.hpp
```diff
@@ -337,11 +337,16 @@ namespace alpaka
...
+        using DstCommon = std::common_type_t<Idx<TExtent>, Idx<TViewDst>>;
```
fwyzard:
```suggestion
        using DstCommon [[maybe_unused]] = std::common_type_t<Idx<TExtent>, Idx<TViewDst>>;
```

fwyzard:
@psychocoderHPC is there a better way to do this, without relying on finding a common type?

fwyzard:
By the way, the main use case in the CMS code is to support doing `alpaka::memcpy` from an `std::vector` (that uses `size_t` as the index type) to user GPU buffers (that frequently uses `uint32_t` as the index type).

AuroraPerego (approved):

Issue #2671, opened 2026-09-16
Atomic operations on pinned host memory crash on Intel GPUs


fwyzard:
### Description

On Intel Data Center GPU Max (Ponte Vecchio), calling `alpaka::atomicAdd` on pinned host memory from `alpaka::allocMappedBuf` fails:
```
Segmentation fault from GPU at 0xff80eceed65000, ctx_id: 9 (CCS) type: 2 (AtomicAccessViolation), level: 0 (PTE), access: 2 (Atomic), banned: 1, aborting.
Abort was called at 288 line in file:
.../intel-compute-runtime-25.18.33578.42/shared/source/os_interface/linux/drm_neo.cpp
```

### Environment

- Intel Data Center GPU Max 1550 (`intel_gpu_pvc`), Level Zero backend
- Intel oneAPI 2026.1
- Intel compute-runtime 25.18.33578 (Level Zero driver 1.6.33578+77)
- alpaka `develop` (919a30d3)

### Simple reproducer without alpaka

- `test.cc`
```cpp
#include <sycl/sycl.hpp>

#include <cstdint>
#include <cstring>
#include <iostream>
#include <string>

int main(int argc, char** argv) {
  if (argc != 2 or (std::strcmp(argv[1], "host") != 0 and std::strcmp(argv[1], "shared") != 0 and
                    std::strcmp(argv[1], "device") != 0)) {
    std::cerr << "Usage: " << argv[0] << " host|shared|device\n";
    return 1;
  }
  std::string const type = argv[1];

  sycl::queue q{sycl::gpu_selector_v};
  sycl::device const dev = q.get_device();
  std::cout << "Device: " << dev.get_info<sycl::info::device::name>() << '\n';
  std::cout << "  usm_atomic_host_allocations:   " << dev.has(sycl::aspect::usm_atomic_host_allocations) << '\n';
  std::cout << "  usm_atomic_shared_allocations: " << dev.has(sycl::aspect::usm_atomic_shared_allocations) << '\n';

  int32_t* v = nullptr;
  if (type == "host") {
    v = sycl::malloc_host<int32_t>(1, q);
  } else if (type == "shared") {
    v = sycl::malloc_shared<int32_t>(1, q);
  } else {
    v = sycl::malloc_device<int32_t>(1, q);
  }
  if (v == nullptr) {
    std::cerr << "Failed to allocate " << type << " memory\n";
    return 1;
  }

  // Initialise the value; device memory is not accessible from the host.
[...]

Issue #2641, opened 2026-08-16
Some tests fail on Aurora with Intel Ponte Vecchio GPUs


fwyzard:
When running all tests on Intel Ponte Vecchio GPUs on Aurora, with oneAPI 2025.3.2 and the LevelZero backend, three tests fail:
```
The following tests FAILED:
         41 - execTest (Subprocess aborted)
         43 - memBufTest (SEGFAULT)
         54 - warpTest (Failed)
```

## `execTest`
```
$ ./test/unit/exec/execTest
Randomness seeded to: 2422421902
ALPAKA_CHECK failed because '!(*value == 1)'

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
execTest is a Catch2 v3.8.1 host application.
Run with -? for options

-------------------------------------------------------------------------------
oncePerGrid - alpaka::test::TestAccs - 1
-------------------------------------------------------------------------------
/home/fwyzard/alpaka/test/unit/exec/src/Once.cpp:43
...............................................................................

/home/fwyzard/alpaka/test/unit/exec/src/Once.cpp:77: FAILED:
  REQUIRE( *status == true )
with expansion:
  false == true

Segmentation fault from GPU at 0xff80eceed65000, ctx_id: 9 (CCS) type: 2 (AtomicAccessViolation), level: 0 (PTE), access: 2 (Atomic), banned: 1, aborting.
Segmentation fault from GPU at 0xff80eceed65000, ctx_id: 9 (CCS) type: 2 (AtomicAccessViolation), level: 0 (PTE), access: 2 (Atomic), banned: 1, aborting.
Abort was called at 288 line in file:
/home/ubit/rpmbuild/BUILD/intel-compute-runtime-25.18.33578.42/shared/source/os_interface/linux/drm_neo.cpp
/home/fwyzard/alpaka/test/unit/exec/src/Once.cpp:77: FAILED:
  {Unknown expression after the reported line}
due to a fatal error condition:
  SIGABRT - Abort (abnormal termination) signal

===============================================================================
test cases:      26 |      25 passed | 1 failed
assertions: 1001192 | 1001190 passed | 2 failed

Aborted
```

## `memBufTest`
```
$ ./test/unit/mem/buf/memBufTest
Randomness seeded to: 3254355655
[...]

PR #2672, opened 2026-09-16
Use device memory for `Once.cpp` tests

fwyzard:
Use device memory for the atomic operations: some devices (e.g. Intel GPUs) do not support atomic operations on pinned host memory.


fwyzard:
Works around #2671 to fix one of the crashes reported in #2641.

AuroraPerego (approved):

Issue #2673, opened 2026-09-17
Catch2 tests using managed memory crash on Intel GPUs


fwyzard:
### Description

On Intel Data Center GPU Max (Ponte Vecchio), `memBufManagedTest` in `memBufTest` crashes with a segmentation fault:

```
$ ./test/unit/mem/buf/memBufTest
Randomness seeded to: 3493700263
Testing Kernel with scalar indices with a grid of (123456, 1) blocks x (1, 1) threads x (1, 1) elements...
Testing Kernel with scalar indices with a grid of (192, 1, 1) blocks x (643, 1, 1) threads x (1, 1, 1) elements...

~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
memBufTest is a Catch2 v3.16.0 host application.
Run with -? for options

-------------------------------------------------------------------------------
memBufManagedTest - alpaka::test::TestAccs - 17
-------------------------------------------------------------------------------
.../test/unit/mem/buf/src/ManagedAndMappedMemoryBufTest.cpp:33
...............................................................................

.../test/unit/mem/buf/src/ManagedAndMappedMemoryBufTest.cpp:33: FAILED:
due to a fatal error condition:
  Test if unified memory works in: 
  Intel(R) Data Center GPU Max 1550
  SIGSEGV - Segmentation violation signal

===============================================================================
test cases:  24 |  23 passed | 1 failed
assertions: 118 | 117 passed | 1 failed

Segmentation fault
```

The test itself is correct: the crash is caused by the interaction between the Intel GPU runtime and the Catch2 POSIX signal handlers.

### Environment

- Intel Data Center GPU Max 1550 (`intel_gpu_pvc`), Level Zero backend
- Intel oneAPI 2026.1
- Intel compute-runtime 25.18.33578 (Level Zero driver 1.6.33578+77)
- alpaka `develop` (919a30d3), internal Catch2 v3.16.0

### Analysis

1. On discrete GPUs, the Intel GPU runtime ([compute-runtime](https://github.com/intel/compute-runtime)) implements shared USM allocations (`sycl::malloc_shared`, used by `alpaka::allocManagedBuf`) by migrating the memory between the host and the device.
[...]

PR #2674, opened 2026-09-17
Disable Catch2 POSIX handlers for Intel GPUs

fwyzard:
Catch2's signal handlers are not compatible with the SIGSEGV handler used by the Intel GPU runtime to migrate shared memory.

When building for Intel GPUs (`alpaka_SYCL_ONEAPI_GPU=ON`), alpaka now configures the internal Catch2 with `CATCH_CONFIG_NO_POSIX_SIGNALS`.

See `README_SYCL.md` for more details.


fwyzard:
Addresses #2673 to fix one of the crashes reported in #2641.

SimeonEhrig:
I'm fine with the change. Only one question. What we are missing, if we disable the POSIX handler? I was a little bit digging in the code. It looks like, there will be no stacktrace and maybe other extra error output missing. That's correct and is something missing?

fwyzard:
Not sure.

I guess if Catch2 is no longer able to intercept SIGSEGV we won't get a nice report in case of failures, and we cannot write explicit tests for any SIGSEGV functionality.

The change should be limited to building for Intel GPUs, and should not affect any other builds.

AuroraPerego (approved):

SimeonEhrig:
> Not sure.
> 
> I guess if Catch2 is no longer able to intercept SIGSEGV we won't get a nice report in case of failures, and we cannot write explicit tests for any SIGSEGV functionality.
> 
> The change should be limited to building for Intel GPUs, and should not affect any other builds.

Okay. Yes because it is limited to the Intel GPU, I see good chances that other build configurations will cover potential problems will with signals.

SimeonEhrig (approved):

PR #2675, opened 2026-09-17
SYCL warp operations on Intel GPUs require inlining

fwyzard:
On Intel GPUs, the SYCL group operations used to implement the warp operations (`group_ballot()`, `get_opportunistic_group()` and the group algorithms) give wrong results when some work-items have returned early from the kernel, and the code calling them is not inlined, for example when compiling with `-O0` (as in CMake's Debug builds), `-fno-inline` or `-fno-inline-functions`: the work-items that have returned are still counted as members of the group.

To avoid silently wrong results, add a `static_assert` to the SYCL implementation of `activemask`, `all`, `any`, `ballot`, `shfl`, `shfl_up`, `shfl_down` and `shfl_xor`, that fails the compilation when they are used in an Intel GPU build where `__NO_INLINE__` is defined. The check can be disabled defining `ALPAKA_SYCL_DISABLE_WARP_INLINE_CHECK`, for example if the kernels using the warp operations never return early.

Skip the corresponding SYCL tests in these builds, and document the limitation in `README_SYCL.md`.


fwyzard:
See #2470, #2641, and intel/llvm#23204.

AuroraPerego (approved):

PR #2676, opened 2026-09-17
Limit threads per block to the device and kernel maximum

fwyzard:


fwyzard:
When compiled by clang-cuda in debug mode, some of the kernels cannot run with the device maximum of 1024 threads per block.

psychocoderHPC:
> When compiled by clang-cuda in debug mode, some of the kernels cannot run with the device maximum of 1024 threads per block.

The reason can be that the kernel will than exceed the shared memory limits or is above the register limit. Since in debug some optimizations are not performed thats often happen if you go near to the thread limit per block.

Exactly what you write in the comment for the funcrion. Sry was answering first and than had a quick check of the diff

psychocoderHPC (approved):

